PR #3098, opened 2018-12-06
Fix Traffic Ops Golang wrapper copying *http.Request, breaking status codes

rob05c:
Fixed the Traffic Ops Golang Auth wrapper to not copy the Request.

Wrappers should never copy the Request, because it breaks the context,
set by other wrappers or the real handler (really, the whole reason
the context exists).

Right now, it was specifically breaking the response code context
key; but in the future, it could break anything. Wrappers should
never change the *http.Request pointer, always the value. That is,
`*req = *foo`, not `req = foo`.

#### What does this PR do?

Fixes #(issue_number)

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/api/api.go
+++ b/traffic_ops/traffic_ops_golang/api/api.go
@@ -526,8 +526,8 @@ func GetUserFromReq(w http.ResponseWriter, r *http.Request, secret string) (auth
 	return user, nil, nil, http.StatusOK
 }
 
-func AddUserToReq(r *http.Request, u auth.CurrentUser) *http.Request {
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/2911/
Test FAILed.

PR #3100, opened 2018-12-06
Updated profiles and parameters test to clean up parameters better

moltzaum:
#### What this PR does
Updated profile test to delete parameters it generates. Updated parameters delete test so that it checks whether or not a parameter was deleted by id instead of name and config (If there happens to be a parameter created by a profile there is a conflict).

#### How to verify this PR
Checkout the first commit, build traffic ops, run `go test -run TestP` and check the database to make sure no parameters exist. Repeat for the second commit. The first commit should show the problem, and the second should fix it.

This PR adds data to `tc-fixtures.json`, then modifies the tests so that they still pass. The data in `tc-fixtures.json` I will need to use.

#### Check all that apply

- [x] This PR includes tests (not new ones)
- [x] This PR includes all required license headers (no new files)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/2917/
Test PASSed.

PR #3104, opened 2018-12-10
Add monitoring.json snapshotting

### LINKED ISSUE #1738: Traffic Monitor `monitoring.json` Is Not Synchronized with CRConfig
The Monitor gets data from 2 TO endpoints, `monitoring.json` and the `CRConfig`; the `monitoring.json` is live data, not a snapshot. 

So, for example, if you remove the `health.polling.url` from a Profile, or delete the Profile, or rename the Profile, in preparation to snapshot a version/name change, it disappears from` monitoring.json`, and the Monitor doesn’t know where to poll anymore. Until it gets the updated Snapshot.

The `monitoring.json` data needs Snapshotted in sync with the CRConfig (either put in the CRConfig itself, or the same 'snapshot' button creates a synchronized `monitoring.json` Snapshot).

This will allow users to make large, lengthy changes, and Snapshot atomically, without breaking Monitor polling in unexpected ways in the meantime.

Note the `monitoring.json` data overlaps with the CRConfig a great deal. The monitor necessarily uses the CRConfig Snapshot for all overlapping data. Hence, if the data is combined it really isn't that much to add to the CRConfig, and if it isn't combined, the overlapping data should be removed as it isn't used.

ezelkow1:
These changes are to fix the issue of the monitoring.json and crconfig's being out of sync. Currently it is possible to make a change to a status (or other field that the monitoring.json presents) and have it show up
immediately while the changes are not reflected in a crconfig until it has been snapshotted, causing an inconcsistency across the files.

This will change the column name in the snapshot table from 'content' to 'crconfig' and then add a new column 'monitoring'. When a snapshot is done we then store both the crconfig and monitoring.json in the table at
the same time, maintaining consistency. Monitoring.json is then pulled from the table instead of being generated on the fly like was done previously.

#### What does this PR do?

Fixes #1738  

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [x] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

#### Check all that apply

- [x] This PR includes tests
- [x] This PR includes a database migration (ensure that migration sequence is correct)

### DIFF
--- a/traffic_ops/traffic_ops_golang/crconfig/snapshot.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/snapshot.go
@@ -27,10 +27,12 @@ import (
 
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/monitoring"
 )
 
 // Snapshot takes the CRConfig JSON-serializable object (which may be generated via crconfig.Make), and writes it to the snapshot table.
-func Snapshot(tx *sql.Tx, crc *tc.CRConfig) error {
+// It also takes the monitoring config JSON and writes it to the snapshot table.
+func Snapshot(tx *sql.Tx, crc *tc.CRConfig, monitoringJSON *monitoring.Monitoring) error {
 	log.Debugln("calling Snapshot")
 	bts, err := json.Marshal(crc)
 	if err != nil {
@@ -40,10 +42,16 @@ func Snapshot(tx *sql.Tx, crc *tc.CRConfig) error {
 	if crc.Stats.DateUnixSeconds != nil {
 		date = time.Unix(*crc.Stats.DateUnixSeconds, 0)
 	}
+
+	btstm, err := json.Marshal(monitoringJSON)
+	if err != nil {
+		return errors.New("marshalling JSON: " + err.Error())
+	}
+
 	log.Debugf("calling Snapshot, writing %+v\n", date)
-	q := `insert into snapshot (cdn, content, last_updated) values ($1, $2, $3) on conflict(cdn) do update set content=$2, last_updated=$3`
-	if _, err := tx.Exec(q, crc.Stats.CDNName, bts, date); err != nil {
-		return errors.New("Error inserting the snapshot into database: " + err.Error())
+	q := `insert into snapshot (cdn, crconfig, last_updated, monitoring) values ($1, $2, $3, $4) on conflict(cdn) do update set crconfig=$2, last_updated=$3, monitoring=$4`
+	if _, err := tx.Exec(q, crc.Stats.CDNName, bts, date, btstm); err != nil {
+		return errors.New("Error inserting the crconfig and monitoring snapshot into database: " + err.Error())
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


asfgit:

traffic_ops/traffic_ops_golang/routes.go
```diff
@@ -142,7 +141,8 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.4, http.MethodGet, `cdns/dnsseckeys/refresh/?(\.json)?$`, cdn.RefreshDNSSECKeys, auth.PrivLevelOperations, Authenticated, nil},
 
 		//CDN: Monitoring: Traffic Monitor
-		{1.1, http.MethodGet, `cdns/{cdn}/configs/monitoring(\.json)?$`, monitoring.Get, auth.PrivLevelReadOnly, Authenticated, nil},
+		{1.1, http.MethodGet, `cdns/{cdn}/configs/monitoring(\.json)?$`, crconfig.SnapshotGetMonitoringHandler, auth.PrivLevelReadOnly, Authenticated, nil},
+		{1.2, http.MethodGet, `cdns/{name}/configs/monitoring(\.json)?$`, crconfig.SnapshotGetMonitoringHandler, auth.PrivLevelReadOnly, Authenticated, nil},
```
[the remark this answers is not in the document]
rob05c: Actually, the second one is both broken, and will never get called. Routes are called in the order they're registered. So, a 1.2 route has to be before a 1.1 route with the same path, or it'll never get called. But it's broken anyway, because it's using `{name}`, but the handler itself expected a param named `{cdn}`. So yeah, needs removed.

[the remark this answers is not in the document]
ezelkow1: Ok, I wasnt sure how those worked completely.  Previously there were actually 2 definitions for the monitoring.json but in separate locations in the routes.go. I figured it was time to consolidate them at this point since that was just confusing but I can just move it all to the single one

traffic_ops/traffic_ops_golang/routes.go
```diff
@@ -142,7 +141,7 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.4, http.MethodGet, `cdns/dnsseckeys/refresh/?(\.json)?$`, cdn.RefreshDNSSECKeys, auth.PrivLevelOperations, Authenticated, nil},
 
 		//CDN: Monitoring: Traffic Monitor
-		{1.1, http.MethodGet, `cdns/{cdn}/configs/monitoring(\.json)?$`, monitoring.Get, auth.PrivLevelReadOnly, Authenticated, nil},
+		{1.1, http.MethodGet, `cdns/{cdn}/configs/monitoring(\.json)?$`, crconfig.SnapshotGetMonitoringHandler, auth.PrivLevelReadOnly, Authenticated, nil},
```
dneuman64: Looks like we are just changing the behavior of the end point that TM uses, do you feel that there is value in knowing what the non-snapped Monitor config would look like?  Seems to me like we should have two seperate end points, one for the monitoring snapshot and one for the raw monitoring.config file.

dneuman64: Yes, I think the idea is that at some point we will want to be able to diff what is snapshotted vs what is raw.   There could also be some future use cases where the raw data really is desired, I just want to make sure that we keep it available like it is today.

ezelkow1: @dneuman64 would you be good with us opening up a separate issue for that? We were throwing around some ideas like a diff end point instead of a full raw one and things like that, but its a bit out of the scope of this PR

traffic_ops/traffic_ops_golang/crconfig/snapshot.go
```diff
@@ -68,11 +76,42 @@ where c.name = $1
...
+// An error is only returned on database error, never if the CDN or snapshot does not exist.
+// Because all snapshotting is handled by the crconfig endpoints we have to also do the monitoring one
+// here as well
+func GetSnapshotMonitoring(tx *sql.Tx, cdn string) (string, bool, error) {
+	log.Debugln("calling GetSnapshotMonitoring")
+
+	monitorSnapshot := sql.NullString{}
+	// cdn left join snapshot, so we get a row with null if the CDN exists but the snapshot doesn't, and no rows if the CDN doesn't exist.
+	q := `
+select s.monitoring as snapshot
```
rob05c: Nitpick: Good formatting, but we're trying to move toward capitalizing SQL keywords as well, prefer:
```
SELECT s.monitoring AS snapshot
FROM cdn AS c ...
```

ezelkow1: That was vscode's doing :). Hopefully it will let me capitalize things

rob05c: Yeah, and we still have a lot of SQL in TOGolang that's lowercase and one-line. Not a big deal.

traffic_ops/traffic_ops_golang/routes.go
dneuman64: Yeah, I think a new issue is fine.

PR #3126, opened 2018-12-14
Removed unnecessary http-parser dependency

ocket8888:
## What does this PR do?
Removes Traffic Portal's dependency on an obscure 'http-parser' RPM package that doesn't appear to be used for anything.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Build and test Traffic Portal

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/infrastructure/docker/build/Dockerfile-traffic_portal
+++ b/infrastructure/docker/build/Dockerfile-traffic_portal
@@ -30,8 +30,7 @@ RUN	yum -y install \
 	yum -y clean all
 
 # traffic_portal specific
-RUN	rpm -ivh https://kojipkgs.fedoraproject.org//packages/http-parser/2.7.1/3.el7/x86_64/http-parser-2.7.1-3.el7.x86_64.rpm && \
-	yum -y install \
+RUN	yum -y install \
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/2950/
Test FAILed.

Shihta:
The commit seems caused CIAB error.
Maybe CIAB needs more dependencies settings if you remove this package.
I suggest to revert this patch.

ocket8888:
I tested this by running Traffic Portal in CDN-in-a-Box - can you explain why you think it broke?

dangogh:
@Shihta Please provide more detail -- what error did you see?

Shihta:
I think it's the same issue as #3131 and it has fixed.

My screenshot is shown as follows:


Thanks!

PR #3143, opened 2018-12-18
Fix improper file slurping in Perl

rawlinp:
#### What does this PR do?

The proper way to enable slurp mode is by doing `local $/;`. When you do
`undef $/` it is undefining the global `$/` input record separator which
is bad and can cause a multitude of problems. Using the `local` version
only nullifies the input record separator locally in order to enable
slurp mode in that particular function rather than changing it globally.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Hit the `POST /api/$version/isos?stream=yes` endpoint. This can be done in Traffic Portal, just make sure to set stream=yes in order to hit the changed code. I'm not sure how to really verify `parse_retention_period_in_seconds` since I can't find anywhere in the code that actually uses that function. The GenDbDump change isn't really verifiable anymore since it's a UI endpoint, but this should cover our bases in case it is being used in some obscure way by a non-UI endpoint.

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/app/lib/API/Iso.pm
+++ b/traffic_ops/app/lib/API/Iso.pm
@@ -270,7 +270,7 @@ sub generate_iso {
 			}
 
 			# slurp it in..
-			undef $/;
+			local $/;
 			$data = <$fh>;
 
 			close $fh;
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/2973/
Test FAILed.

PR #3157, opened 2018-12-21
Added error context helper to help with tests. Also wrote unit tests for it.

moltzaum:
#### What does this PR do?

I plan on using this PR in some tests I am writing. It creates a new type of error that has an error code associated with it. There is also an 'ErrorContext', which makes sure you don't accidentally create errors with codes that have no meaning. The ErrorContext also keeps statistics, which gives a nice metric for negative test coverage.

This affects nothing yet. Later it will affect new tests (for traffic ops primarily).
I have no plans to include this with traffic ops directly or other components.

Run the unit test and read the code. I am hoping it is somewhat understandable from reading the code, however it might be a little more obscure since I'm only running negative tests and not showing how I'd use it. Below are a few snippets of snippets + pseudocode showing how I might use it.

```golang
package cache_config

const (
    COMMON_BASE            = 10 + iota
    NOT_ENOUGH_ASSIGNMENTS 
    BAD_ASSIGNMENT_MATCH   
)

const (
    A_BASE                  = 40 + iota
    INVALID_ACTION_LABEL  
    INVALID_ACTION         
    INVALID_RESP_TO_COOKIES
    INVALID_PIN_IN_CACHE    
    INVALID_REVALIDATE      
    INVALID_TTL_IN_CACHE   
)

// scoped to the package name
var ErrorContext *test.ErrorContext

func init() {
    // *_BASE constants are not error codes in themselves, but can be used to
    // determine if an error code exists within a certain 'class' of errors
    errorCodes := []uint{
        INVALID_ACTION_LABEL,
        INVALID_ACTION,
        INVALID_RESP_TO_COOKIES,
        INVALID_PIN_IN_CACHE,
        INVALID_REVALIDATE,
        INVALID_TTL_IN_CACHE,
        NOT_ENOUGH_ASSIGNMENTS,
        BAD_ASSIGNMENT_MATCH,
    }
    // Not many mappings are made since most error messages have additional information
    ErrorContext = test.NewErrorContext("cache config", iterableErrorCodes)
    ErrorContext.AddMapping(NOT_ENOUGH_ASSIGNMENTS, "not enough assignments in rule")
    ErrorContext.TurnPanicOn()
}

```

```golang
package cache_config

func matchAssignmentThenDoXXX(token string) {
    if token does not contain '=' {
        return ErrorContext.NewError(BAD_ASSIGNMENT_MATCH, "could not match yyy assignment")
    }
    // do XXX with match result
}

func matchRuleThenDoXXX(config string) {

    if tokens in config is < 2 {
        return ErrorContext.NewError(NOT_ENOUGH_ASSIGNMENTS)
    }
    matchAssignmentThenDoXXX( config_tokens[0] )
    // continue...
}

```

Currently the only documentation I have is in the comments. To whoever is reading this, do you think it needs a README?

#### Check all that apply

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/test/error_context.go
+++ b/traffic_ops/traffic_ops_golang/test/error_context.go
@@ -40,6 +40,7 @@ func (e errorCode) ErrorCode() int {
 }
 
 // not a pointer receiver, does not modify itself
+// in the new error the original cause is maintained
 func (e errorCode) Prepend(fmtStr string, fmtArgs ...interface{}) ErrorCoder {
 	err := fmt.Errorf(fmtStr, fmtArgs...)
 	e.error = fmt.Errorf("%v %v", err, e.error)
@@ -54,8 +55,13 @@ func AddErrorCode(code int, err error) ErrorCoder {
 	return NewError(code, "%v", err)
 }
 
-// ErrorContext
-// contains a list of all error codes (a whitelist)
+// ErrorContext regulates which error codes can be made and keeps a count
+// of all the errors created through the context
+//
+// ErrorContext.NewError can be used like test.NewError
+//
+// Implementation details:
+// contains a list of all valid error codes
 //		- allows user to make sure they are creating the correct errors
 //		- actually a map
 //			lookup can be done without linear search
@@ -63,7 +69,32 @@ func AddErrorCode(code int, err error) ErrorCoder {
 // contains mapping from error code to name (either for testing metainfo or used in case no args are given)
 //		not required for all error codes, or for any
 //
-// context.NewError was made to improve upon test.NewError
+// Recommended setup:
+//
+//	package some_regex_checker
+//
+//	const (
+//	    COMMON_BASE            = 10 + iota
+//	    NOT_ENOUGH_ASSIGNMENTS
+//	    BAD_ASSIGNMENT_MATCH
+//		...
+//	)
+//
+//	// scoped to the package name
+//	var ErrorContext *test.ErrorContext
+//
+//	func init() {
+//	    errorCodes := []uint{
+//	        NOT_ENOUGH_ASSIGNMENTS,
+//	        BAD_ASSIGNMENT_MATCH,
+//	    }
+//	    ErrorContext = test.NewErrorContext("cache config", errorCodes)
+//		err := ErrorContext.AddMapping(NOT_ENOUGH_ASSIGNMENTS, "not enough assignments in rule")
+//		// check err
+//	    ErrorContext.TurnPanicOn()
+//		// in non-init function create errors with the context
+//	}
+//
 type ErrorContext struct {
 	calledNewError  bool
 	createdNewError bool
@@ -92,10 +123,10 @@ func NewErrorContext(contextName string, errCodes []uint) *ErrorContext {
 	}
 }
 
-// although highly discouraged, panic mode is made as an option
+// although golang panics are highly discouraged, panic mode is made as an option
 // made this decision partially because type assertions and map membership have similar options
 // if a user doesn't have panic mode on, they should still terminate after running into an error
-// panic is off by default, and must be turned on explicitly
+// panic is off by default, and must be turned on explicitly so that the user must make an active decision
 // panic must be turned on before errors are created
 // once turned on the panic mode can't be turned off
 func (ec *ErrorContext) TurnPanicOn() error {
@@ -106,7 +137,18 @@ func (ec *ErrorContext) TurnPanicOn() error {
 	return nil
 }
 
-// msg should be a plain string without special formatting
+// AddMapping gives a default message for a given error code.
+// Mappings should be configured before errors are called for consistency.
+//
+// usage: ErrorContext.AddMapping(404, "not found")
+//		  err := ErrorContext.NewError(404)
+//        err := ErrorContext.NewError(404, "not found: %v", prev_err")
+// the err automatically has a default message
+// the default message can be overridden with something else to add context to the error
+//
+// parameters:
+//	`code` should exist in the error context, only one mapping can exist per error context
+//  `msg` should be a plain string without special formatting
 func (ec *ErrorContext) AddMapping(code uint, msg string) error {
 	if ec.calledNewError {
 		return ec.internalError(BAD_INIT_TIMING, nil)
@@ -121,6 +163,7 @@ func (ec *ErrorContext) AddMapping(code uint, msg string) error {
 	return nil
 }
 
+// AddMap applies the `AddMapping` function for every element in the given map
 // the function argument will not override the current contents of the map, everything is additive
 // an error is returned if a duplicate code is added
 func (ec *ErrorContext) AddMap(mapping map[uint]string) error {
@@ -166,6 +209,7 @@ func (ec ErrorContext) internalError(code int, offender interface{}) ErrorCoder 
 	return AddErrorCode(code, err)
 }
 
+// whitelist is defined by code map membership
 func (ec ErrorContext) whitelisted(code uint) bool {
 	_, ok := ec.codes[code]
 	return ok
@@ -178,10 +222,13 @@ func (ec *ErrorContext) newError(code uint, fmtStr string, fmtArgs ...interface{
 	return NewError(int(code), fmtStr, fmtArgs...)
 }
 
-// primary error creator
-//	 checks whitelist
-//	 handles internal errors
-//	 interprets args
+// ErrorContext.NewError creates an error similar to test.NewError
+// any error created in this manner must have a code that belongs to the error context
+// the args is interpreted as a format string with args, but I'm using a ...interface{} because
+// if no args are specified a lookup will be made to find the default configured string (see `AddMapping`)
+//
+// usage: NewError(404, "not found: %v", prev_err)
+//		  NewError(404) // (mapping must exist otherwise this is an error)
 func (ec *ErrorContext) NewError(code uint, args ...interface{}) ErrorCoder {
 	ec.calledNewError = true
 
@@ -204,10 +251,14 @@ func (ec *ErrorContext) NewError(code uint, args ...interface{}) ErrorCoder {
 	return ec.internalError(BAD_ERROR_CODE, code)
 }
 
+// AddErrorCode takes a regular error and gives it a code
+// since this method is scoped to an error context, the code must exist in the whitelist
 func (ec ErrorContext) AddErrorCode(code uint, err error) ErrorCoder {
 	return ec.NewError(code, "%v", err)
 }
 
+// GetErrorStats returns the map of error codes
+// ec.codes[code] is the number of times an error with the indexed code has been run
 func (ec ErrorContext) GetErrorStats() map[uint]uint {
 	if ec.createdNewError {
 		return ec.codes
--- a/traffic_ops/traffic_ops_golang/test/error_context.go
+++ b/traffic_ops/traffic_ops_golang/test/error_context.go
@@ -67,16 +67,16 @@ func AddErrorCode(code int, err error) ErrorCoder {
 //			lookup can be done without linear search
 //			we can use the map to keep count of which errors are made
 // contains mapping from error code to name (either for testing metainfo or used in case no args are given)
-//		not required for all error codes, or for any
+//		- not required for all error codes, or for any
 //
 // Recommended setup:
 //
 //	package some_regex_checker
 //
 //	const (
-//	    COMMON_BASE            = 10 + iota
-//	    NOT_ENOUGH_ASSIGNMENTS
-//	    BAD_ASSIGNMENT_MATCH
+//		COMMON_BASE            = 10 + iota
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/2985/
Test PASSed.

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// all internal errors
+const BAD_ERROR_CODE = -1  // when bad error code is given in creation of new error
+const BAD_MAP_LOOKUP = -2  // when creating an error with no error message, default message wasn't found
+const BAD_MAP_CREATE = -3  // when bad error code is given in creation of map
+const BAD_DUP_MAPPING = -4 // when a mapping is already made
+const BAD_FMT_STRING = -5  // when the fmt string isn't a string
+const BAD_INIT_TIMING = -6 // when the error context is modifed after having created an error
+
```
dangogh: why not make use of `iota` here?
```
const (
    BAD_ERROR_CODE = -1 * iota
    BAD_MAP_LOOKUP // no assignment needed for the rest...
...
)
```

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,214 @@
...
+// contains mapping from error code to name (either for testing metainfo or used in case no args are given)
+//		not required for all error codes, or for any
+//
+// context.NewError was made to improve upon test.NewError
+type ErrorContext struct {
+	calledNewError  bool
+	createdNewError bool
+	doPanic         bool
+	name            string
+	codes           map[uint]uint
```
dangogh: had to trace thru the code to figure out where the whitelist came from, then saw the comment above.  Should this be called "whitelist" to make that apparent?

moltzaum: `codes` has a double use. It is both a whitelist of codes and a count of which error codes have been called. There is a `whitelisted` method, which I think communicates well enough when the map of error codes is used as a whitelist in member functions. In all other references besides object construction, the variable is used for its other purpose.

Maybe it would be more clear if the test refers to its code list as a whitelist? Or if I have a comment somewhere else? Because the `whitelisted` method is unexported and the primary purpose of the ErrorContext is to be a whitelist, I would think most people would refer to the ErrorContext comment.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3046/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3049/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3050/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3067/
Test PASSed.

traffic_ops/traffic_ops_golang/test/error_context.go
rob05c: >codes has a double use. It is both a whitelist of codes and a count of which error codes have been called.
It would be helpful to add that as a GoDoc comment to the variable, i.e.
```
	// codes is both a whitelist of codes and a count of which error codes have been called.
	codes           map[uint]uint
```

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+		calledNewError:  false,
+		createdNewError: false,
+		doPanic:         false,
+		name:            contextName,
+		codes:           codeMap,
+		description:     descMap,
+	}
+}
+
+// although golang panics are highly discouraged, panic mode is made as an option
```
rob05c: This breaks GoDoc. GoDoc comments need to start with the symbol name, e.g. `// TurnPanicOn enables panic mode. Although...`

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+	var err error
+	switch code {
+	case BAD_ERROR_CODE:
+		err = fmt.Errorf("error code %v not found in whitelist for \"%s\" error context", offender, ec.name)
+	case BAD_MAP_CREATE:
+		err = fmt.Errorf("when creating default error mapping, code %v wasn't found in the code whitelist for the \"%s\" error context", offender, ec.name)
+	case BAD_DUP_MAPPING:
+		err = fmt.Errorf("when creating default error mapping, code %v was already found as a mapping for the \"%s\" error context", offender, ec.name)
+	case BAD_MAP_LOOKUP:
+		err = fmt.Errorf("bad default error lookup for code %v in the \"%s\" error context", offender, ec.name)
```
rob05c: I'm not really clear on what "offender" is? It looks like it's being used as the code in some errors, but `the leading argument \"%v\"` on others? And it's an `interface{}`? 

Would you mind adding a function-level comment explaining the `offender` parameter?

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+		err = fmt.Errorf("tried to modify error context after creating error")
+	}
+
+	if ec.doPanic {
+		panic(err)
+	}
+	return AddErrorCode(code, err)
+}
+
+// whitelist is defined by code map membership
```
rob05c: GoDoc, should be `// whitelisted`

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+func (ec ErrorContext) whitelisted(code uint) bool {
+	_, ok := ec.codes[code]
+	return ok
+}
+
+// makes sure every created error gets counted
```
rob05c: GoDoc, `// newError makes...`

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+func (ec *ErrorContext) newError(code uint, fmtStr string, fmtArgs ...interface{}) ErrorCoder {
+	ec.codes[code]++
+	ec.createdNewError = true
+	return NewError(int(code), fmtStr, fmtArgs...)
+}
+
+// ErrorContext.NewError creates an error similar to test.NewError
```
rob05c: GoDoc, should be `// NewError creates...`

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+// any error created in this manner must have a code that belongs to the error context
+// the args is interpreted as a format string with args, but I'm using a ...interface{} because
+// if no args are specified a lookup will be made to find the default configured string (see `AddMapping`)
+//
+// usage: NewError(404, "not found: %v", prev_err)
```
rob05c: For GoDoc, if you put this on its own line, indented at least 2 spaces, it'll show up correctly as an example, i.e.
```
// usage: 
//  NewError(404, "not found: %v", prev_err)
//  NewError(404) // (mapping must exist otherwise this is an error)
```

traffic_ops/traffic_ops_golang/test/error_context.go
rob05c: Also, GoDoc comments should be a paragraph of complete sentences, with capitalization and punctuation. i.e.
```
// ErrorContext.NewError creates an error similar to test.NewError.
// Any error created in this manner must have a code that belongs to the error context.
// The args is interpreted as a format string with args, but I'm using a ...interface{} because
// if no args are specified a lookup will be made to find the default configured string (see `AddMapping`).
```

See https://blog.golang.org/godoc-documenting-go-code

traffic_ops/traffic_ops_golang/test/error_context_test.go
```diff
@@ -0,0 +1,162 @@
...
+package test
+
+import (
+	"fmt"
+	"testing"
+)
+
+var codes []uint
+
+func init() {
```
rob05c: `init` is dangerous and unintuitive, because people importing your package don't necessarily know what's happening or when. It should be avoided when it isn't necessary. 

It isn't necessary here, you can 
```
var codes =[]uint{
	0, 1, 2,
}
```

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+func NewError(code int, fmtStr string, fmtArgs ...interface{}) ErrorCoder {
+	err := fmt.Errorf(fmtStr, fmtArgs...)
+	return &errorCode{err, err, code}
+}
+
+func (e errorCode) ErrorCode() int {
+	return e.code
+}
+
+// not a pointer receiver, does not modify itself
```
rob05c: GoDoc, should start with the symbol, and be complete sentences, e.g. 
```
// Prepend prepends the string built from fmtStr and fmtArgs to the errorCode and returns it.
// Note it is not a pointer receiver, and does not modify itself.
// In the new error, the original cause is maintained.
```

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+//		COMMON_BASE            = 10 + iota
+//		NOT_ENOUGH_ASSIGNMENTS
+//		BAD_ASSIGNMENT_MATCH
+//		...
+//	)
+//
+//	// scoped to the package name
+//	var ErrorContext *test.ErrorContext
+//
+//	func init() {
```
rob05c: I'm not really a fan of recommending people put things in `init`. In general, `init` should be avoided, it's dangerous and unintuitive. People expect code to execute when they tell it to, people don't expect arbitrary code to execute outside the `main` function, especially just because they imported a package.

Since this is designed for testing, what about recommending people do this initialization in `TestMain`?                                                                                                                  
https://golang.org/pkg/testing/#hdr-Main
https://github.com/apache/trafficcontrol/blob/4d974f0000397016570e60597522f884077e1f26/traffic_ops/testing/api/v14/traffic_ops_test.go#L36

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+		return ec.internalError(BAD_INIT_TIMING, nil)
+	}
+	ec.doPanic = true
+	return nil
+}
+
+// AddMapping gives a default message for a given error code.
+// Mappings should be configured before errors are called for consistency.
+//
+// usage: ErrorContext.AddMapping(404, "not found")
```
rob05c: GoDoc,
```
// usage:
//  ErrorContext.AddMapping(404, "not found")
```

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+type ErrorCoder interface {
+	Error() string
+	ErrorCode() int
+	Cause() error
+	Prepend(string, ...interface{}) ErrorCoder
+}
+
+func NewError(code int, fmtStr string, fmtArgs ...interface{}) ErrorCoder {
```
rob05c: Go `New` functions should generally match the type they're creating, either `NewErrorCoder`, or rename `ErrorCoder` to `Error`.

traffic_ops/traffic_ops_golang/test/error_context.go
```diff
@@ -0,0 +1,267 @@
...
+//	func init() {
+//		errorCodes := []uint{
+//			NOT_ENOUGH_ASSIGNMENTS,
+//			BAD_ASSIGNMENT_MATCH,
+//		}
+//		ErrorContext = test.NewErrorContext("cache config", errorCodes)
+//		err := ErrorContext.AddMapping(NOT_ENOUGH_ASSIGNMENTS, "not enough assignments in rule")
+//		// check err
+//		ErrorContext.TurnPanicOn()
+//		// in non-init function create errors with the context
```
rob05c: >create errors with the context

How? Would you mind adding another example of how to do that? Should that example be in a GoDoc comment for `ErrorCoder`, and here `// in non-init function create errors with the context - see ErrorCoder` maybe?

traffic_ops/traffic_ops_golang/test/error_context.go
moltzaum: I changed it so that it isn't "recommended", but it is still an example. I will keep `TestMain` in mind for the future.

traffic_ops/traffic_ops_golang/test/error_context.go
moltzaum: I expanded the example so that a comment isn't used.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3147/
Test PASSed.

moltzaum:
`godoc -http=":6060"`

http://localhost:6060/pkg/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/test/

PR #2928, opened 2018-10-13
TO Client:  no ids required in input json

dangogh:
#### What does this PR do?

Removes requirement for providing IDs for foreign keys when using traffic ops client.   Example:  using `client.CreateProfile`,  the Cdn field currently must be provided as `cdnID`. With this change,  it can be provided instead as `cdnName`. 

#### Which TC components are affected by this PR?

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

#### What is the best way to verify this PR?

The API tests have also been modified to take advantage of this capability.   From `traffic_ops/testing/api/v14`, run `go test -cfg ../conf/traffic_ops_test.conf` to test.

#### Check all that apply

- [x] This PR includes tests
- [ ] This PR includes documentation updates
- [ ] This PR includes an update to CHANGELOG.md
- [x] This PR includes all required license headers
- [ ] This PR includes a database migration (ensure that migration sequence is correct)
- [ ] This PR fixes a serious security flaw. Read more: [www.apache.org/security](http://www.apache.org/security/)

PR #3163, opened 2018-12-25
Fix Traffic Ops Tenancy and Activity Bugs, Fix TO API Test Framework to work with Tenancy

### LINKED ISSUE #2732: Unable to Edit or Delete a Tenant where Active is false
When a Tenant is made inactive, the user loses the ability to edit or delete it from the system. An error is returned to the user: "not authorized on this tenant." This limitation should be removed.

rob05c:
This fixes major TO Tenancy bugs, as well as fixing the TO API framework to work with Tenancy enabled, and enabling it.

Unfortunately, all the tenancy issues, and the test framework, are intimately related, and there isn't a good way to separate them; so this PR is unfortunately large.

Specifically:

- Fix TO Tenancy checks to consider whether the user's tenant is active, not whether the resource's tenant is active.

- Fix TO Tenancy to prevent access when parent tenants (of the 
  user's tenant) are inactive.

- Add TO API Tests to verify tenancy, verify access is disabled when
  parents or self are inactive.

- Fix TO API Test user panic on test failure.

- Fix TO API Test tenants to automatically delete children before
  parents, rather than hard-coding.

- Fix TO API Test parameters race condition, fixed to delete all
  parameters not only first name+file.

- Fix TO API Tests to enable Tenancy on all tests (by creating
  Parameters, which creates the use_tenancy param).
  The TO API Tests were completely broken with tenancy, due
  to the above bugs. With the bugs fixed, the tests pass
  with tenancy enabled.

Fixes #2732 

#### What does this PR do?

Fixes #2732 

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Run API tests. Run TO with Tenancy enabled, verify resources are accessible as expected.

#### Check all that apply

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/client/deliveryserviceserver.go
+++ b/traffic_ops/client/deliveryserviceserver.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
+	"net/url"
 	"strconv"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
@@ -61,23 +62,21 @@ func (to *Session) DeleteDeliveryServiceServer(dsID int, serverID int) (tc.Alert
 	return resp, reqInf, nil
 }
 
+// GetDeliveryServiceServers gets all delivery service servers, with the default API limit.
 func (to *Session) GetDeliveryServiceServers() (tc.DeliveryServiceServerResponse, ReqInf, error) {
-	route := apiBase + `/deliveryserviceserver`
-	reqResp, remoteAddr, err := to.request(http.MethodGet, route, nil)
-	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
-	if err != nil {
-		return tc.DeliveryServiceServerResponse{}, reqInf, errors.New("requesting from Traffic Ops: " + err.Error())
-	}
-	defer reqResp.Body.Close()
-	resp := tc.DeliveryServiceServerResponse{}
-	if err = json.NewDecoder(reqResp.Body).Decode(&resp); err != nil {
-		return tc.DeliveryServiceServerResponse{}, reqInf, errors.New("decoding response: " + err.Error())
-	}
-	return resp, reqInf, nil
+	return to.getDeliveryServiceServers(url.Values{})
 }
 
+// GetDeliveryServiceServersN gets all delivery service servers, with a limit of n.
 func (to *Session) GetDeliveryServiceServersN(n int) (tc.DeliveryServiceServerResponse, ReqInf, error) {
-	route := apiBase + `/deliveryserviceserver` + `?limit=` + strconv.Itoa(n)
+	return to.getDeliveryServiceServers(url.Values{"limit": []string{strconv.Itoa(n)}})
+}
+
+func (to *Session) getDeliveryServiceServers(urlQuery url.Values) (tc.DeliveryServiceServerResponse, ReqInf, error) {
+	route := apiBase + `/deliveryserviceserver`
+	if qry := urlQuery.Encode(); qry != "" {
+		route += `?` + qry
+	}
 	reqResp, remoteAddr, err := to.request(http.MethodGet, route, nil)
 	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
 	if err != nil {
--- a/traffic_ops/client/deliveryserviceserver.go
+++ b/traffic_ops/client/deliveryserviceserver.go
@@ -17,6 +17,9 @@ package client
 
 import (
 	"encoding/json"
+	"errors"
+	"net/http"
+	"strconv"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/lib/go-util"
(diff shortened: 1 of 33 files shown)


traffic_ops/testing/api/v14/cachegroupsdeliveryservices_test.go
```diff
@@ -122,7 +125,7 @@ func CreateTestCachegroupsDeliveryServices(t *testing.T) {
 func DeleteTestCachegroupsDeliveryServices(t *testing.T) {
 	log.Debugln("DeleteTestCachegroupsDeliveryServices")
 
-	dss, _, err := TOSession.GetDeliveryServiceServers()
+	dss, _, err := TOSession.GetDeliveryServiceServersN(1000000)
```
rob05c: This fixes a race, where tests would randomly fail when DSSes weren't in the 20 returned by default.

traffic_ops/testing/api/v14/parameters_test.go
```diff
@@ -115,9 +115,14 @@ func DeleteTestParameter(t *testing.T, pl tc.Parameter) {
...
-	if len(resp) > 0 {
-		respParameter := resp[0]
 
+	if len(resp) == 0 {
+		t.Errorf("DeleteTestParameter got no params for %+v %+v\n", pl.Name, pl.ConfigFile)
+	} else if len(resp) > 1 {
+		// TODO figure out why this happens, and be more precise about deleting things where created.
+		// t.Errorf("DeleteTestParameter params for %+v %+v expected 1, actual %+v\n", pl.Name, pl.ConfigFile, len(resp))
+	}
+	for _, respParameter := range resp {
```
rob05c: This fixes a race, where certain tests would create additional parameters with the same name+file (which is permissible, as long as the values are different). This fixes it to delete all returned parameters.

traffic_ops/testing/api/v14/tenants_test.go
```diff
@@ -110,27 +114,195 @@ func DeleteTestTenants(t *testing.T) {
...
 	}
+	expectedChildDeleteErrMsg := `Tenant '` + strconv.Itoa(tenant1.ID) + `' has child tenants. Please update these child tenants and retry.`
+	if _, err := TOSession.DeleteTenant(strconv.Itoa(tenant1.ID)); err == nil {
+		t.Fatalf("%s has child tenants -- should not be able to delete", t1)
+	} else if !strings.Contains(err.Error(), expectedChildDeleteErrMsg) {
+		t.Errorf("expected error: %s;  got %s", expectedChildDeleteErrMsg, err.Error())
+	}
+
+	deletedTenants := map[string]struct{}{}
+	for {
```
rob05c: This makes `DeleteTestTenants` delete all tenants starting with children, instead of hard-coding which need deleting first.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/2987/
Test PASSed.

traffic_ops/testing/api/v14/user_test.go
```diff
@@ -61,11 +63,10 @@ func CreateTestUsers(t *testing.T) {
 	for _, user := range testData.Users {
 
 		resp, _, err := TOSession.CreateUser(&user)
-		log.Debugln("Response: ", resp.Alerts)
-
 		if err != nil {
-			t.Errorf("could not CREATE user: %v\n", err)
+			t.Fatalf("could not CREATE user: %v\n", err)
```
moltzaum: @rob05c Nearly two months ago @dangogh changed Fatalf to Errorf across all tests because Fatalf tests from cleaning up (2928/commits/f32c5ec73db68ef9b24caa7ad6a0fb175051eaf0). It seems that other tests would have the same panic issue if their create method failed. I want to make sure we have a consistent understanding of how tests should terminate and why.

moltzaum: Full link: https://github.com/apache/trafficcontrol/pull/2928/commits/f32c5ec73db68ef9b24caa7ad6a0fb175051eaf0

traffic_ops/testing/api/v14/tenants_test.go
dangogh: +1

traffic_ops/testing/api/v14/user_test.go
dangogh: +1 -- if the Fatalf occurs before the test gets cleaned up,  the following test may fail.   Really should only use Fatalf if we can't recover from the situation,  and that should be made very clear.

traffic_ops/traffic_ops_golang/tenant/tenancy.go
```diff
@@ -136,11 +136,28 @@ func GetUserTenantListTx(user auth.CurrentUser, tx *sql.Tx) ([]tc.TenantNullable
...
+  SELECT bool_and(active) as v FROM user_tenant_parents
+),
+user_tenant_children AS (
+  SELECT id, name, active, parent_id
+  FROM tenant WHERE id = (SELECT v FROM user_tenant_id) AND (SELECT v FROM user_tenant_active)
+  UNION
+  SELECT t.id, t.name, t.active, t.parent_id
+  FROM tenant t JOIN user_tenant_children ON user_tenant_children.id = t.parent_id
+)
+SELECT id FROM user_tenant_children;
```
dangogh: +1 -- this is much clearer, although still pretty hairy.   I wonder if it would be worth creating this as an SQL function that can be tested/maintained outside of the code?

traffic_ops/traffic_ops_golang/tenant/tenancy.go
```diff
@@ -171,13 +188,46 @@ func IsTenancyEnabledTx(tx *sql.Tx) (bool, error) {
...
+  (select bool_and(active) from user_tenant_parents) as active,
+  tenancy.value AS use_tenancy
+FROM
+  tenancy,
+  q
+WHERE
+  id = (select v from resource_tenant_id)
+UNION ALL SELECT -1, false, tenancy.value AS use_tenancy FROM tenancy
+FETCH FIRST 1 ROW ONLY;
+`
```
dangogh: We've already decided to eliminate the `use_tenancy` param and have it now always set to 1.    Can we get rid of that portion?

traffic_ops/testing/api/v14/user_test.go
rob05c: >Really should only use Fatalf if we can't recover from the situation

That's exactly what happens here, that's why I changed it. If an error is returned, the following line
```
log.Debugln("Response: ", resp.Alerts)
```

will panic. If an error is returned, the test can't keep going, which is exactly what `Fatalf` is for. Because of the panic, the test won't get cleaned up anyway, and subsequent tests will fail anyway. Using `Fatal` instead of `Error` avoids the panic stacktrace, and makes it easier to see what went wrong to fix it.

We really need to fix the API Test framework to clean up tests even if they panic or `t.Fatalf`. I don't think it'll be hard, I think it's as simple as changing all our tests from e.g.
```
func TestCacheGroups(t *testing.T) {
	CreateTestTypes(t)
	CreateTestCacheGroups(t)
	GetTestCacheGroups(t)
	CheckCacheGroupsAuthentication(t)
	UpdateTestCacheGroups(t)
	DeleteTestCacheGroups(t)
	DeleteTestTypes(t)
}
```
To
```
func TestCacheGroups(t *testing.T) {
	defer DeleteTestTypes(t)
	CreateTestTypes(t)
	defer DeleteTestCacheGroups(t)
	CreateTestCacheGroups(t)
	GetTestCacheGroups(t)
	CheckCacheGroupsAuthentication(t)
	UpdateTestCacheGroups(t)
}
```
(similar to https://github.com/apache/trafficcontrol/blob/d083557e1a59cd4620b7652fd2821bf5d18d1ad3/traffic_ops/testing/api/v14/origins_test.go#L24 but the `defer Delete`s need to come before their `Create`s).

But that should probably be its own PR.

traffic_ops/client/deliveryserviceserver.go
```diff
@@ -42,3 +45,48 @@ func (to *Session) CreateDeliveryServiceServers(dsID int, serverIDs []int, repla
...
+	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
+	if err != nil {
+		return tc.DeliveryServiceServerResponse{}, reqInf, errors.New("requesting from Traffic Ops: " + err.Error())
+	}
+	defer reqResp.Body.Close()
+	resp := tc.DeliveryServiceServerResponse{}
+	if err = json.NewDecoder(reqResp.Body).Decode(&resp); err != nil {
+		return tc.DeliveryServiceServerResponse{}, reqInf, errors.New("decoding response: " + err.Error())
+	}
+	return resp, reqInf, nil
```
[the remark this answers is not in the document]
rob05c: Does `/deliveryserviceserver?limit=-1` work?

traffic_ops/traffic_ops_golang/tenant/tenancy.go
rob05c: I think we made the decision, but the code doesn't currently enforce it, at present users can still leave tenancy disabled.

I think we need to merge https://github.com/apache/trafficcontrol/pull/2791 before removing `use_tenancy` checks.

rob05c:
Fixed merge conflicts.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/2995/
Test PASSed.

traffic_ops/client/deliveryserviceserver.go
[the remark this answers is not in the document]
rob05c: Ok, I changed them to use a common func. IMO the abstraction isn't worth the readability cost. So, I changed it in the safest way I could think of.

A sigil (-1 or 0) would be confusing, since not passing a limit results in the default API limit of 20, not infinite (which is the bug adding a `N` func fixes); likewise passing the API limit would be bug-prone if the API changed; so, passing the query param itself seemed the safest and least confusing.

traffic_ops/testing/api/v14/user_test.go
rob05c: Actually, thinking about this some more, I think we should use `Fatalf` anywhere the test can't continue, irrespective of `defer` cleanups attempts. 

There are always going to be cases where "delete" can't clean up appropriately, after a test fails at some point. Test errors are just like compile errors: you should only ever consider the first failure. Because any one failure can cause a chain of other unrelated failures, or worse, false successes. Any one test failure is a failure of the entire test suite. 

The only right answer is to only consider the first failure, and cleanup is irrelevant, because all subsequent tests, pass or fail, are invalid.

This means all build pipelines (and people) should be using `go test -failfast`.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3014/
Test PASSed.

PR #3171, opened 2018-12-30
Add TO api helper safety to prevent double writes

rob05c:
#### What does this PR do?

Adds a safety to all api helper write funcs, to prevent double writes.

This catches a common programming mistake, of doing `if err != nil { HandleErr() }` instead of `if err != nil { HandleErr(); return}`, and `if foo { WriteResp(foo) } WriteResp(bar)` instead of `if foo { WriteResp(foo); return } WriteResp(bar)`.

If a `return` is missed resulting in a double write, now the api func will not write the second time (which is almost always the correct behavior), and will log an error.

Includes unit tests, to both verify that a double-write doesn't write twice, and to verify single writes do still write correctly.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

`go test`

#### Check all that apply

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/api/api.go
+++ b/traffic_ops/traffic_ops_golang/api/api.go
@@ -46,6 +46,7 @@ import (
 const DBContextKey = "db"
 const ConfigContextKey = "context"
 const ReqIDContextKey = "reqid"
+const APIRespWrittenKey = "respwritten"
 
 type CRUDFactory func(reqInfo *APIInfo) CRUDer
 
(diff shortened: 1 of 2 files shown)


traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -60,6 +61,12 @@ func WriteResp(w http.ResponseWriter, r *http.Request, v interface{}) {
 
 // WriteRespRaw acts like WriteResp, but doesn't wrap the object in a `{"response":` object. This should be used to respond with endpoints which don't wrap their response in a "response" object.
 func WriteRespRaw(w http.ResponseWriter, r *http.Request, v interface{}) {
+	if getWroteResp(r) {
```
rawlinp: My only suggestion would be renaming these `respWritten` and `setRespWritten`. Overall I think this PR looks good.

rob05c: Done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3016/
Test PASSed.

PR #3177, opened 2019-01-02
Added warning SQL query for duplicated origins

ocket8888:
## What does this PR do?
This adds a check to `admin`'s (and `admin.pl`'s) upgrade, migration and status commands for checking for "bad data". That's intended to be expanded on an as-needed basis as more bad data configurations are discovered. Currently checks for Delivery Services that share origin <abbr titel="Fully Qualified Domain Name">FQDN</abbr>s.

The check can also be run on its own by using the 'check' command of `admin` (and `admin.pl`).

- [x] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Try to run the `admin` and `admin.pl` "upgrade", "migration", "status" and "check"  commands against various environments and verify the new "check" step doesn't break anything.

## Check all that apply
- [ ] This PR includes tests
- [ ] This PR fixes a serious security flaw. [Read more](http://www.apache.org/security/)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3002/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3003/
Test PASSed.

traffic_ops/app/db/admin.pl
```diff
@@ -64,7 +64,7 @@
 	. "dropdb  - Execute db 'dropdb' on the database for the current environment.\n"
```
rawlinp: Since this script is deprecated I don't think we should be adding new functionality to it. I think the new functionality should only be added to admin.go unless there is a very good reason it also needs to be in admin.pl. If there is a very good reason for it, then admin.pl should just shell out to the admin binary so that the checks aren't duplicated.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3007/
Test PASSed.

traffic_ops/app/db/admin.pl
rawlinp: If we want to take this even further I was just thinking of having the admin.pl script shell out all of its individual commands to the binary. So the admin.pl script would just be a tiny perl script that basically only does something like this:
```
my $new_admin = dirname(__FILE__) + "/admin";
shift @ARGV;
my $args = join " ", @ARGV;
system("$new_admin $args");
```
We could probably even just convert it into a bash script still named admin.pl, but I'd be concerned about the possible confusion with that.

rob05c: > having the admin.pl script shell out all of its individual commands to the binary

+1

>We could probably even just convert it into a bash script still named admin.pl

-1. Perl is required by the Linux Standard Base (LSB), so there's not really any downside of a basic Perl script that doesn't require any dependencies or external modules (i.e. the absurd install time of TO/CPAN).

ocket8888: Took me a few tries, but I think I got it right now.

ocket8888: I agree, no reason to cause such confusion.
I like the idea of using the binary for everything, but I don't think it belongs in the same commit as added functionality.

rawlinp: > I like the idea of using the binary for everything, but I don't think it belongs in the same commit as added functionality.
+1, would be better to review and merge the PR that makes admin.pl shell everything out to the binary first. Then the new functionality can just be reviewed in admin.go.

ocket8888: #3186

ocket8888: rebased on master after #3186 was merged - this should be good to go now.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3034/
Test PASSed.

rawlinp:
I think this could be a useful check to have in the db/admin tool, but I'm not sure we should be running the check on `upgrade`, `status`, and other commands just to emit a warning since it will most likely be overlooked, especially when some of those commands produce quite a bit of output. Since it will require manual intervention for operators to fix their duplicate origins, we need to really make it noticeable, like a red notification in the TP UI. To be clear, I'm thinking we probably need this `check` command so the data can be checked without having a running TO instance, but we probably also need an API endpoint that basically runs the same check so that the results could be used in Traffic Portal. That endpoint should probably share the sql files that provide the data checks so that we only need one copy of the sql checks.

ocket8888:
My concern is that adding a new mode and requiring users to use it to check their databases not only requires them to read documentation on a tool they already know how to use, but also allows them to perform database upgrades and be completely unaware that the next time they do it the upgrade may fail. I feel strongly that the warning should be emitted on `upgrade`, but less so about the other commands. `create_db` is probably unnecessary in retrospect, because I believe it merely creates the _schema_ and leaves the user with an empty database. I'm not totally sure how `status` works, but it seemed like a good idea for inspecting the status of a database to indicate potentially invalid data - though the command is more about migration status so there're arguments on both sides.

A notification in the TP UI could be beneficial, but in my opinion that kind of thing is most helpful to people who don't deal with the low-level database operations (or shouldn't) and will likely end up being ignored. I'm preparing a separate PR that will cause creation of new, bad data (duplicate origin fqdns on DSes) to fail with an appropriate error message, which _will_ show up as a notification in the TP UI.

PR #3184, opened 2019-01-07
WIP Add CDN-in-a-Box Dynamic IPs

rob05c:
This is an experiment to add dynamic IPs to the CDN-in-a-Box.

I'm not sure it works, I was having trouble bringing up the CiaB with it, but I'm not sure if it was just my environment, because it's setting all the `/etc/resolv.conf` and `/data/bind/etc` zone data, identical to the hardcoding.

It's also not complete. In order to remove the hard-coded IPs from the `docker-compose.yml`, the `traffic_ops_data/profiles` and `traffic_ops_data/servers` will also need dynamically populated (I think that should be straightforward, using the same pattern as the rest of this).

The big motivation for this is to be able to run multiple CiaB on the same machine. This is especially useful for build systems.

#### What does this PR do?

Fixes #(issue_number)

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

#### Check all that apply

- [ ] This PR includes tests


rob05c:
@dangogh @JBevillC Thoughts?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3009/
Test FAILed.

JBevillC:
Rob,

I few things I've noticed when testing this PR:

1) Serial number of all zone files will need to be incremented each time the DNS server is restarted. Otherwise future reload/restart(s) of the bind server will not pick up any new additions.

2) The size of the IPV4 and IPV6 subnets is not known to the DNS server and must be parsed out of $(ip addr list).  By default, docker allocates a class 16 on 172.0.0.0 starting with 172.0.16.0/16 and increments the third octet for each new dynamic network created.  This isn't always the case as a user can configure the default subnet size to be much smaller.  The IPV6 default subnet allocation is a bit more difficult to determine and depends on the subnet configured in /etc/docker/daemon.json. 

3) Using a pull method for discovering IPs of services slows down the start of CIAB.  A push model, imo, would be more suitable for adding stuff like DNSSEC DS records as they change within trafficops.  There would also be no delay without starting up as each container would push their IP configuration and DS record (any any CNAMES) to the DNS server on startup.  My next task is to add DNSSEC support and it will need the ability to push a DS record into the TLD of the CDN.

If you are okay with a push model, I think I can have something implemented in about 16 working hours.

Thoughts?

rob05c:
> Serial number of all zone files will need to be incremented each time the DNS server is restarted. 

That shouldn't be difficult with e.g. sed/awk, right? I haven't tried, but I can try to come up with something when I have time.

> The size of the IPV4 and IPV6 subnets is not known to the DNS server

Why is this an issue? If you mean for the Profiles, e.g. `ip_allow`, they don't need subnets, we can insert each IP into them. Is there another reason we need subnets?

> Using a pull method for discovering IPs of services slows down the start of CIAB. A push model, imo, would be more suitable for adding stuff like DNSSEC DS records as they change within trafficops.

I would really rather not Push. The "Enroller" does "push", and it's constantly having issues, and seems to be the cause of the CiaB taking so long frequently. I literally spent half the day yesterday, from 8am to noon, trying to get the CiaB to start, with cryptic Enroller failures.

> There would also be no delay

That's the theory, but in practice, it doesn't seem to be true. Things (both in CiaB and in the real world) are frequently waiting on pushes to happen, that fail and you never know why, or which side failed, or whether a push is being retried.

Push has myriad downsides, not least of which is the unreliability, and difficulty to debug.

I would much rather see us move away from those models, which to my experience, appears to be causing unreliability and slowdowns.

The script is very lightweight, it should be fine running every 10s. Would that be too slow?

JBevillC:
@rob05c

Understood. I will proceed with using your script and the pull method instead of a push.  I'm going to have to add some additional processing as mentioned above to determine the subnet of each network interface.  The goal is to add as little to your scripts as possible to make this work.

In the future I don't know how I'm going to make DNSSEC work with this solution, but we can cross that bridge at a later time.

rob05c:
> I'm going to have to add some additional processing as mentioned above to determine the subnet of each network interface. 

@JBevillC I'm still not understanding why this is necessary. Where do we need the subnets?

JBevillC:
I'll just remove the reverse zones and see if that is good enough.

rob05c:
Yeah, I was going to ask if we really need them?

If so, maybe we can find a tool to generate them? I found a few, but they're all GPL, still looking for one with a compatible license.

rob05c:
Putting a note here so I don't forget: 

I wonder if we can use DNS UPDATE
https://docstore.mik.ua/orelly/networking_2ndEd/dns/ch10_02.htm
https://tools.ietf.org/html/rfc2136

JBevillC:
@rob05c I don't know if we need reverse DNS or not, but proper forward/reverse dns does exist in production.

PR #3186, opened 2019-01-07
All functionality for the admin program is now located in admin.go

ocket8888:
## What does this PR do?
Makes the deprecated `admin.pl` a small wrapper around the `admin` binary, so that new functionality need only be added in one place.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Other: admin.pl

Run any and all tests for the `admin.pl` script

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/app/db/admin.go
+++ b/traffic_ops/app/db/admin.go
@@ -303,12 +303,13 @@ func loadSchema() {
 }
 
 func reverseSchema() {
-	fmt.Println("WARNING: the '" + CmdReverseSchema + "' command will be removed with Traffic Ops Perl because it will no longer be necessary")
-	cmd := exec.Command("db/admin.pl", "--env="+Environment, CmdReverseSchema)
+	fmt.Fprintf(os.Stderr, "WARNING: the '%s' command will be removed with Traffic Ops Perl because it will no longer be necessary\n", CmdReverseSchema)
+	cmd := exec.Command("db/reverse_schema.pl")
+	cmd.Env = append(os.Environ(), "MOJO_MODE="+Environment)
 	out, err := cmd.CombinedOutput()
 	fmt.Printf("%s", out)
 	if err != nil {
-		die("Can't run `db/admin.pl reverse_schema`: " + err.Error())
+		die("Can't run `db/reverse_schema.pl`: " + err.Error())
 	}
 }
 
--- a/traffic_ops/app/db/reverse_schema.pl
+++ b/traffic_ops/app/db/reverse_schema.pl
@@ -14,10 +14,20 @@
 # limitations under the License.
 #
 use strict;
-use File::Basename;
+use Schema;
+use DBIx::Class::Schema::Loader qw/make_schema_at/;
 
-STDERR->autoflush(1);
-print STDERR "WARNING: this script is deprecated, please use the db/admin binary instead.\n\n";
-my $args = join(" ", @ARGV);
-my $new_admin = dirname(__FILE__) . "/admin";
-exit(system("$new_admin " . $args));
+my $db_info = Schema->get_dbinfo();
+
+my $user    = $db_info->{user};
+my $pass    = $db_info->{password};
+my $dsn     = Schema->get_dsn();
+make_schema_at(
+	'Schema', {
+		debug                   => 1,
+		dump_directory          => './lib',
+		overwrite_modifications => 1,
+		exclude => 'goose_db_version',
+	},
+	[ $dsn, $user, $pass ],
+);


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3011/
Test PASSed.

traffic_ops/app/db/admin.pl
```diff
@@ -14,261 +14,8 @@
...
-	}
-}
-
-sub show_users {
-	if ( system("psql -h $host_ip -p $host_port -U $db_super_user -ec '\\du';") != 0 ) {
-		die "Can't show users";
-	}
-}
-
-sub reverse_schema {
```
rawlinp: Shoot, I just remembered that this function is still needed as long as we have Perl TO API endpoints, and the admin binary actually shells out this command to this admin.pl script. So this function should probably be pulled out into its own little Perl script that the admin binary can shell that command out to. Then admin.pl can just be as you have it now. Otherwise, in this PR if you try running this command, I think it will just create an endless cycle due to the circular dependency.

ocket8888: shouldn't that be implemented in the binary?

ocket8888: I'm concerned though - because this needs to exist as long as the Perl back-end exists, and because we deprecated the script, it sort of seems like we're now locked in to a release schedule that demands that 4.0 not be released until the entire Perl TO codebase can be removed.

rob05c: Can we put `reverse_schema` in its own script file, `admin_reverse_schema.pl`, with only that function, and make Go execute that script?

It's not ideal, but the only dependencies it will have is DBIx, which fortunately is in a CentOS package, so it shouldn't need CPan or Carton.

ocket8888: We could, but idk how much that really buys us. It also has a dependency on `Schema` btw, but I think that's maybe a standard library thing? Not sure. Leaving it in `admin.pl` means adding two dependencies: `GetOpt::Long` (standard?) and `YAML`, which I don't personally think of as a huge deal, but it is two places to maintain the command line argument format. On the other hand separating it means we now have three programs to maintain for the `admin` tool, at least until `admin.pl` vanishes. It's also functionality specific to `admin.pl` since it only deals with Perl data representation and should vanish with the need for that script

In summary, I'm sort of torn on the idea of putting that in a new script, and I'm leaning towards saying that that's something that should maybe take place once `admin.pl` actually gets removed (assuming the Perl back-end still exists at that point).

rob05c: >It also has a dependency on Schema 
`Schema` is DBIx.

>two dependencies: GetOpt::Long (standard?) and YAML

>two places to maintain the command line argument format

I'm not suggesting maintaining the argument format. Make the script take the arguments it needs (environment, user, pass, host, port), and make Go pass them. No YAML or GetOpt.

ocket8888: I think you misunderstood, those dependencies and that maintenance *only apply* if we're leaving it in `admin.pl`, not if we put it in a separate script as you suggest.

rawlinp: We shouldn't really be worried about which dependencies the new reverse_schema.pl script will require, since as long as we have Perl endpoints we will still have the Perl dependencies required for reverse_schema. For now, yes, we will end up with 3 programs to maintain until we get rid of admin.pl, but that shouldn't really be a concern either because reverse_schema.pl will only contain that required functionality and won't require any further updates until it's eventually removed with TO-Perl. And with admin.pl just being a thin Perl wrapper over the admin.go binary, there really won't be much to maintain there either.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3018/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3019/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3020/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3021/
Test PASSed.

traffic_ops/app/db/admin.pl
ocket8888: the alternative I'm suggesting is keep that one function in `admin.pl` and remove everything else. It's not as clean as making a new script, I guess, but that's not super clean either. I suppose I just don't see the advantage of moving this to a new file until we're prepared to get rid of `admin.pl` in 4.0.

rob05c: As long as `admin.pl` is kept as an empty wrapper, having real functionality called by `admin.go` creates a circular dependency. It can work, but circular scripts are perilous. Much better, simpler, safer, easier to understand, to make a third script to avoid the cycle.

ocket8888: Well I was going to have `admin.pl` intercept the `reverse_schema` functionality such that it never calls `admin` in that case. No circularity required.

rob05c: But `admin.go` has to call `admin.pl` for that function as well. The scripts are still calling each-other. Even if the code execution isn't circular, the scripts are. That's confusing and dangerous.

ocket8888: Well in any case I don't feel strongly about it. I can separate it into its own script.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3022/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3025/
Test PASSed.

traffic_ops/app/db/admin.pl
rawlinp: The `get_dbinfo` function checks the environment variables for MOJO_MODE and uses that to determine which database.conf to read the config from. By default it will use conf/development/database.conf. That comes from traffic_ops/app/lib/Schema.pm I believe.

I'll pull this down and take a spin.

traffic_ops/app/db/admin.go
```diff
@@ -303,12 +303,12 @@ func loadSchema() {
 }
 
 func reverseSchema() {
-	fmt.Println("WARNING: the '" + CmdReverseSchema + "' command will be removed with Traffic Ops Perl because it will no longer be necessary")
-	cmd := exec.Command("db/admin.pl", "--env="+Environment, CmdReverseSchema)
+	fmt.Fprintf(os.Stderr, "WARNING: the '%s' command will be removed with Traffic Ops Perl because it will no longer be necessary\n", CmdReverseSchema)
+	cmd := exec.Command("db/reverse_schema.pl")
```
rawlinp: This command needs the MOJO_MODE environment variable injected in order to reverse_schema on a specific environment (rather than just "development"):
```
cmd := exec.Command("db/reverse_schema.pl")
cmd.Env = append(os.Environ(), "MOJO_MODE="+Environment)
...
```

With that fixed I believe this will be good to go.

ocket8888: Really? Wouldn't it inherit the environment of the parent process? It seemed to be working for me, but if you think it needs this I'll add it.

ocket8888: From the Go docs:
```go
        // If Env is nil, the new process uses the current process's
        // environment.
```
However, it looks like the re-write to Go dropped the Perl script's behaviour of setting `MOJO_MODE` - probably because only `reverse_schema` used it and that was calling back to the Perl script anyway.

tl;dr I took your suggestion.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3031/
Test PASSed.

PR #3192, opened 2019-01-07
Fix trafficops python client installation

rawlinp:
#### What does this PR do?

Separate library metadata like `__version__` into `__version__.py` that
`setup.py` pulls from rather than importing version from the package itself.

This fixes an issue with installing this package caused by `setup.py`
importing from the package itself, which causes the package's imports to
run, leading to ModuleNotFoundError.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Other: python trafficops client

Create a python 3 virtualenv, activate it in your shell, and try installing both the current version of master and the version of this PR:
```
mkdir pyclient_test && cd pyclient_test
python3 -m virtualenv venv
source venv/bin/activate
# checkout the current version of TC master
pip install <location of your cloned TC repo>/traffic_control/clients/python # this will fail with ModuleNotFoundError
# checkout this PR and attempt to install again
pip install <location of your cloned TC repo>/traffic_control/clients/python # this should succeed
```

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_control/clients/python/setup.py
+++ b/traffic_control/clients/python/setup.py
@@ -22,14 +22,17 @@ Setuptools build/install script for the Python Traffic Control client
 import sys
 import os
 from setuptools import setup, find_packages
-from trafficops import __version__ as version
 
 HERE = os.path.abspath(os.path.dirname(__file__))
+
+about = {}
+with open(os.path.join(HERE, 'trafficops', '__version__.py'), mode='r', encoding='utf-8') as f:
+	exec(f.read(), about)
 
 with open(os.path.join(HERE, "README.rst")) as fd:
 	setup(
 	       name="Apache-TrafficControl",
-	       version=version,
+	       version=about["__version__"],
 	       author="Apache Software Foundation",
 	       author_email="dev@trafficcontrol.apache.org",
 	       description="Python API Client for Traffic Control",
--- a/traffic_control/clients/python/trafficops/__version__.py
+++ b/traffic_control/clients/python/trafficops/__version__.py
@@ -14,15 +14,4 @@
 # limitations under the License.
 #
 
-from .__version__ import __version__
-
-# Import Local Modules into the 'trafficops' package namespace for convenience.
-from .tosession import TOSession
-from .restapi import InvalidJSONError, LoginError, OperationError, api_request, RestApiSession
-
-__all__ = [u'TOSession',
-           u'InvalidJSONError',
-           u'LoginError',
-           u'OperationError',
-           u'api_request',
-           u'RestApiSession']
+__version__ = '1.0.0'


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3015/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3017/
Test PASSed.

traffic_control/clients/python/trafficops/__init__.py
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 
+from .__version__ import __version__
+
 # Import Local Modules into the 'trafficops' package namespace for convenience.
 from .tosession import TOSession
 from .restapi import InvalidJSONError, LoginError, OperationError, api_request, RestApiSession
```
ocket8888: what might be better is removing these bits. If you do that, it shouldn't require anything to be imported to get to the version, which stays in one place at the package level. The cost is that you can't really get much else into the current namespace when you do `from trafficops import *` which is honestly something you shouldn't do anyway, because it clobbers.

ocket8888: to be clear, by "these bits" I mean the local module imports (and consequently the `__all__` definition). Failing that, the package-level `__version__` should still be defined to be the package's version. Otherwise to get the version from within Python you'd need to do `from trafficops.__version__ import __version__` instead of just `from trafficops import __version__`

rawlinp: I'm not totally following the advantages you're proposing. I can get the module version the same way as it was before: 
```
Python 3.6.7 (default, Jan  4 2019, 09:57:41)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import trafficops
>>> trafficops.__version__
'1.0.0'
>>>
```
as well as:
```
Python 3.6.7 (default, Jan  4 2019, 09:57:41)
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from trafficops import __version__
>>> __version__
'1.0.0'
>>>
```

ocket8888: I missed this line: `from .__version__ import __version__`. I suppose this is fine, though I was really excited to find a reason to specifically disallow `from trafficops import *`... oh well.

rawlinp: Yeah, unfortunately, I think something like that would probably require a deprecation cycle as well so that existing instances of `from trafficops import *` don't break for people without notice.

PR #3197, opened 2019-01-09
Calendar widget for DNSSEC/KSK date selection

### LINKED ISSUE #2722: Change Portal CDN DNSSEC Generate input to a Calendar
Right now, the Traffic Portal `CDNs / over-the-top / DNSSEC / Keys / generate` `Effective Date (UTC) *` is a text input field. That field is a date, and always needs to be a valid timestamp in a particular format.

The input field should be changed to a calendar, so users can't typo it, and also just for ease of use.

### LINKED ISSUE #2734: Fix Traffic Portal cdn dnssec generate to pass UNIX epoch, not string
The  Traffic Portal cdn/dnssec/generate page passes a human-readable string, which catastrophically breaks that CDN's DNSSEC keys without https://github.com/apache/trafficcontrol/pull/2720 (see https://github.com/apache/trafficcontrol/issues/2723). 

With that PR in TO, the Portal's input will be accepted and not break the world, even though it doesn't conform to the API proper. But, the Portal still needs fixed to send the correct API field, _especially_ because this allows us to fix the catastrophic bug if someone updates the Portal but not TO.

The fix should be fairly simple, something like
`traffic_portal/app/src/common/modules/form/cdnDnssecKeys/FormCdnDnssecKeysController.js`
```
                modalInstance.result.then(function() {
-                       generate();
+                       var expirationDateMS = Date.parse($scope.ksk_new.expirationDate); // convert human-readable text back to the UNIX epoch the Traffic Ops API accepts
+                       if( isNaN(expirationDateMS) ) {
+                               messageModel.setMessages([ { level: 'error', text: 'invalid date' } ], false);
+                       } else {
+                               $scope.ksk_new.expirationDate = expirationDateMS / 1000;
+                               generate();
+                       }
```

If someone can test that code in a working TP+TO+Riak setup, and make a PR, that'd be awesome.

chadgilloth:
#### What does this PR do?
This adds a date picker for entering date and time when generating or regenerating DNSSEC keys and KSK which ensures the proper date format is created.

Fixes #2722 
Fixes #2734 

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Load Traffic Portal, got to CDNs, choose or create a CDN, from the menu top right select Manage DNSSEC Keys, click Generate or Regenerate and modify the Effective Date field.

#### Check all that apply

- [ ] This PR includes tests


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3027/
Test FAILed.

ocket8888:
what's wrong with [date-type input fields](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date)?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3028/
Test PASSed.

ocket8888:
If you need date **and** time you could use the [`datetime-local` type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) instead.

mitchell852:
@ocket8888 does have a good point. were these native elements considered and would they do the job and provide a suitable user experience?

rob05c:
Do we only officially support Chrome?

I use Firefox. Even if we only "officially" support Chrome, -1 on making intentional decisions that break other browsers.

ocket8888:
We only test on Chrome, and some features are broken on Firefox. "Officially" is probably a strong word, but "supported" is accurate. We could add testing for Firefox, it has a test suite I think.

I don't want to only officially support one browser, but I still think that even user-agent detection to use `<input type="datetime-local">` vs `<input type="date"><input type="time">` as necessary/appropriate would be better than introducing a new front-end dependency and 240 new lines of code.

rob05c:
>user-agent detection to use <input type="datetime-local"> vs <input type="date"><input type="time"> as necessary

Sure.

>better than introducing a new front-end dependency and 240 new lines of code.

I agree in general, but I know Chad's put a good bit of work into this. With this being functional, I'd say it should depend on how much work it is to change. If it's trivial, sure. If it's a lot of work, it's better to have this than not, I'd say merge this, and we can optimize to not use the dependency later.

@chadgilloth what do you think? How difficult does it sound to change?

chadgilloth:
Personally, I think the widget offers a better UX experience for the user. I do prefer to use native elements most of the time but not here.  This widget allows for quick click selection, has a lot of formatting/appearance options and manual editing if the user chooses.

rob05c:
+1 User experience matters, sometimes the standard elements don't cut it, and this is functional as-is. We can always revisit the dependency in the future.

ocket8888:
I'm gonna rebuild my CDN-in-a-Box to check this out, but fwiw the `<input type="date">` element provides a calendar for quick selection as well as manual editing in a format defined by the user-agent's locale.

ocket8888:
I actually can't get this to work; When I go to 'CDNs' --> '{a CDN}' --> 'More' --> 'Manage DNSSEC Keys' the "KSK Expiration (UTC)" input field still appears to be a normal text input field.

elsloo:
At a minimum this should be the effective date field, not the expiration field. I also support the use of the library over the builtins for the widget if the UX is better, we're not in conflict from a licensing standpoint, and we aren't taking on an inordinate amount of burden in terms of maintenance overhead. In this case it seems like a fairly easy thing to support.

mitchell852:
@chadgilloth - is this "effective date" UTC or local date/time? it seems like everything else in TP is UTC so can you make sure it is UTC and labeled correctly? i.e. Effective Date/Time (UTC)

mitchell852:
@chadgilloth - nm. 

image

but have you verified that it's actually UTC time?

mitchell852:
I get confused when i get to this screen

image

it turns out it's the "seconds" screen. do we really need to get that precise? seems to me like minutes would be good enough

mitchell852:
image

Since you can edit the text field (after using the calendar), it seems like the format should be validated to warn on something like this.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3084/
Test PASSed.

chadgilloth:
@mitchell852 Date now displays as UTC. Seconds removed and defaults to 00. Field validates onBlur

traffic_portal/app/src/common/modules/form/cdnDnssecKeys/generate/form.GenerateCdnDnssecKeys.tpl.html
```diff
@@ -56,8 +56,22 @@
...
+                           name="effectiveDate"
+                           type="text"
+                           ng-model="dnssecKeysRequest.effectiveDate"
+                           ng-model-options="{ updateOn: 'blur' }"
+                           ng-pattern="/^[0-9\:\-\ ]+$/"
+                           change="ctrl.zeroSeconds()"
+                           required
+                           autofocus>
                     <small class="input-error" ng-show="hasPropertyError(cdnGenerateDnssecKeysForm.effectiveDate, 'required')">Required</small>
+                    <small class="input-error" ng-show="hasPropertyError(cdnGenerateDnssecKeysForm.effectiveDate, 'pattern')">Invalid date format</small>
```
mitchell852: can you add this to the message so it's clear what the expected format is? YYYY-MM-DD HH:mm:ss

traffic_portal/app/src/common/modules/form/cdnDnssecKeys/generate/FormGenerateCdnDnssecKeysController.js
```diff
@@ -31,6 +31,13 @@ var FormGenerateCdnDnssecKeysController = function(cdn, dnssecKeysRequest, $scop
 
 	$scope.dnssecKeysRequest = dnssecKeysRequest;
 
+	var ctrl = this;
+	ctrl.zeroSeconds = function () {
```
mitchell852: can't this just be $scope.zeroSeconds or is there something i'm missing?

traffic_portal/app/src/common/modules/form/cdnDnssecKeys/generate/form.GenerateCdnDnssecKeys.tpl.html
chadgilloth: Validation message changed to show format

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3087/
Test PASSed.

mitchell852:
were you going to add a warning if they select a date in the past?

image

chadgilloth:
@mitchell852  Not sure how you are typing "foo." I limited the field to only the chars that are used in the date filed 0-9, space, - and :. And that is what I see when I run this locally. I'll make sure again in the morning but in my earlier testing, pressing the generate button caused the field to be validated and the error message to show up as 'Invalid Date Format'

chadgilloth:
I should clarify, yes, you can type it since I am not doing a restrict, but on exit from the field it is validating to an error state.

mitchell852:
@chadgilloth - can u look into this error in the console?

image

mitchell852:
@chadgilloth - ^^ that error happens in the console when i put in an invalid date format into the field and then leave the field (blur)

image

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3175/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3176/
Test PASSed.

PR #3199, opened 2019-01-09
Fix TO POST cachegroups/id/deliveryservices failure when params exist

rob05c:
Fixes cachegroups/id/deliveryservices failing when one of the location
parameters being created on the newly-assigned servers already exists
(because a different server on the same profile was previously
assigned to the delivery service).

This causes the query "ON CONFLICT DO UPDATE SET id = EXCLUDED.id
RETURNING id" to fail, because changing the id creates a foreign
key conflict on the profile_parameters table.

The query doesn't actually need to update the id, Postgres just
requires _something_ be updated in order to return. This changes it to
set name=name, which is the same, and doesn't cause a query failure
from a fk conflict.

Includes API Tests.

#### What does this PR do?

Fixes #(issue_number)

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

#### Check all that apply

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v14/cachegroupsdeliveryservices_test.go
+++ b/traffic_ops/testing/api/v14/cachegroupsdeliveryservices_test.go
@@ -69,6 +69,15 @@ func CreateTestCachegroupsDeliveryServices(t *testing.T) {
 		t.Errorf("setting cachegroup delivery services returned success, but no servers set\n")
 	}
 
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3032/
Test PASSed.

PR #3211, opened 2019-01-12
Remove mandatory x509v3 Subject/Authority Key Identifiers in certificate(s) (#3181)

### LINKED ISSUE #3181: TO API - DeliveryServices AddSSLKeys Incorrectly stores X509 certificate missing Subject/Authority Key as zero length string.
**Method**: POST
**Endpoint**: api/1.4/deliveryservices/sslkeys/add
**Description**: Importing X509 SSL certificate and corresponding intermediate/root X509 certificates that do not contain X509v3 Subject Key Identifier (SKI) and corresponding Authority Key Identifier (AKI) results in the certificate being stored in Traffic Vault as a zero length string.  Further, all intermediate CA certificates that follow are ignored for X509 certificate validation. Additionally,  the RSA key and certificate-request (csr) do get stored correctly in traffic vault, but the certificate ('crt' field in json) is a zero length string. The response code for this import operation is incorrectly sent back to the client as a 200 indicating a successful import.

**Impact**: Traffic Router Languid SSL connector is impacted (partially and fully) upon receiving incomplete/invalid SSL Key/Certificate data from Traffic Ops SSLKeysAPIs.  This presents a problem in that the incomplete/invalid SSL key/cert pair causes TrafficRouter to throw an exception related to an invalid X509 certificate.  After exception has occurred, any new TLS handshake attempts hang and eventually timeout, thus a HTTPS services have been partially or fully impacted.

**Possible Solution**: All X509 certificates that do not contain Subject/Authority Key Identifier X509v3 should be rejected with proper 4XX HTTP code and corresponding JSON error description.  Traffic Router should also be updated to handle invalid certificate data received from Traffic Ops without partial/full impact to SSL Languid connector.

JBevillC:
Remove mandatory check for x509v3 Subject/Authority Key Identifiers in imported certificate(s)

#### What does this PR do?

Fixes #3181 

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

1. Start up a Traffic Ops environment with at least one HTTP_and_HTTPS delivery service.
2. Import a self-signed CA and a certificate that is signed by this CA.  Make sure that the certificate does NOT contain the X509v3 Subject/Authority Identifer Key(s) extensions.
3. After importing the certs, view the certificate either via the API or in the DS sslkeys page of the Traffic Portal.  The certificate field should contain a valid certificate.

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/keys.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/keys.go
@@ -20,7 +20,6 @@ package deliveryservice
  */
 
 import (
-	"bytes"
 	"crypto/x509"
 	"database/sql"
 	"encoding/base64"
@@ -327,12 +326,14 @@ func verifyCertificate(certificate string, rootCA string) (string, bool, error) 
 	}
 	pemEncodedChain := ""
 	for _, link := range chain[0] {
-		// Only print non-self signed elements of the chain
-		if link.AuthorityKeyId != nil && !bytes.Equal(link.AuthorityKeyId, link.SubjectKeyId) {
-			block := &pem.Block{Type: "CERTIFICATE", Bytes: link.Raw}
-			pemEncodedChain += string(pem.EncodeToMemory(block))
-		}
+		// Include all certificates in the chain, since verification was successful.
+		block := &pem.Block{Type: "CERTIFICATE", Bytes: link.Raw}
+		pemEncodedChain += string(pem.EncodeToMemory(block))
 	}
+   
+  	if len(pemEncodedChain) < 1 {
+		return "", false, errors.New("Invalid empty certicate chain in request")
+  	}
 
 	return pemEncodedChain, false, nil
 }


jhg03a:
This PR works for me

JBevillC:
I've reviewed the existing unit test for this endpoint and it doesn't need to change IMO.  I've also tested the following X509 certificate import scenarios on a test instance and all have been successful:

* X509v1 Single Self-Signed Certificate: CN only
* X509v1 Root CA Self Signed + Server Certificate Signed by Root CA 
* X509v1 Root CA + Intermediate CA + Server Certificate Signed by Intermediate CA
* X509v3 Root CA Self Signed + Server Certificate (Test Both with and w-out SKI/AKI extensions)
* X509v3 Root CA Self Signed + Two intermediate CAs + Server Certificate (Test Both with and w-out SKI/AKI extensions)
* Imporrted various real certificates from a real prod system.  All verify okay.

IMO this PR is ready to be merged.

PR #3218, opened 2019-01-15
Bugfixes for users endpoint

### LINKED ISSUE #3117: Bad regex for PUT method of /api/1.x/users endpoint
Specifically, that endpoint cannot be reached for any API version. One *can* `PUT` to `/api/1.x/users/` or even `/api/1.x/users/.json`. The problem is that the regular expression being matched is `users/{id}?(\.json)?$` but should be `users(/{id})?(\.json)?$`.

### LINKED ISSUE #3118: PUT to /api/1.4/users/{{ID}} is broken
There is no payload the endpoint will consider valid for an ID that matches the logged-in user; `role` cannot be blank, but a user cannot modify their own role. Thus, a user cannot modify itself in any way using the `1.4` endpoint.

### LINKED ISSUE #3214: On username conflict `PUT users/:id` hangs and curl gets empty response
This issue documents a bug with the PUT users/:id endpoint, but is not critical enough to fix in perl IMO. It should be resolved when the go endpoint is re-written.

The problem is this:
I was changing a user to have a username that was already taken. The user that I'm updating is not the current user. When there is a conflict the curl hangs for some time, then curl tells me there was an empty reply from the server.

The bug does not occur when the logged in user is updating themselves or when there is no conflict. I presume that this is the case for all conflicts, but I only tested with the username. I don't think this bug is specific to admin or that this only occurs when the already-taken username belongs to the current user but I didn't test these cases thoroughly. 

To replicate:

POST the following user
```json
{
    "username": "admin7",
    "email": "admin7@comcast.net",
    "fullName": "Comcast Admin",
    "localPasswd": "better_twelve",
    "confirmLocalPasswd": "better_twelve",
    "role": 1,
    "tenantId": 1
}
```

Then, logged in as 'admin', change the username of the POSTed user to admin.

### LINKED ISSUE #3219: TO-golang users API "rolename" regression
I fixed a bug a while back that was caused by wrong capitalization of the `"roleName"` field in the `/api/1.x/user/current` API endpoint: #2535. This bug has come back due to #2846 which set the field back to `"rolename"`.

This is a client-breaking change because clients expect the field in `GET /api/x/user/current` to be `"roleName"` not `"rolename"`. There are _other_ user endpoints that return `"rolename"`, but this is not consistent across *all* user endpoints. If you take a close look at https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/API/User.pm you will see that the different endpoints will return either `"roleName"` or `"rolename"` depending on the endpoint. It appears the API documentation has also been updated to *not* match the Perl implementation.

The field names in the Golang implementation of the users endpoints should match the Perl versions exactly, otherwise clients will be broken. For example, in Traffic Portal you no longer have the option to bypass the Delivery Service Request workflow based upon your role name (e.g. "admin"), since Traffic Portal is expecting `"roleName"` rather than `"rolename"`.

moltzaum:
#### What does this PR do?

Fixes #3117
Fixes #3118
Fixes #3214
Fixes #3219 

This PR does not include documentation updates, so doc updates will follow.
_EDIT:_ Committed documentation updates

_EDIT 2:_ This PR is currently a WIP
There is a bugfix for #3219 that I want to address as well

_EDIT 3:_ Committed fix for #3219

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

#3117
Read the issue

#3118
   1) user can update themselves
   2) user cannot update their own role
   3) user cannot PUT or POST a user whos priv_level is higher than their own

I wrote integration tests for PUT. POST should work. I can add another test, but it has the same privilege check.

#3214 
I did not write an integration test for this. The commit is simple enough to see what the problem was, nor is it likely to change in the future.

#3219 
Post a user and make sure 'roleName' is in the json. Update the user and make sure 'roleName' is in the json, then GET the user and make sure 'rolename' is in the json. This should work for all minor versions 1.1 to 1.4

#### Check all that apply

- [x] This PR includes tests

### DIFF
--- a/docs/source/api/user_current.rst
+++ b/docs/source/api/user_current.rst
@@ -169,28 +169,6 @@ Request Structure
 
 Response Structure
 ------------------
-:addressLine1:     The user's address - including street name and number
-:addressLine2:     An additional address field for e.g. apartment number
-:city:             The name of the city wherein the user resides
-:company:          The name of the company for which the user works
-:country:          The name of the country wherein the user resides
-:email:            The user's email address
-:fullName:         The user's full name, e.g. "John Quincy Adams"
-:gid:              A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user - now it is always ``null``
-:id:               An integral, unique identifier for this user
-:lastUpdated:      The date and time at which the user was last modified, in ISO format
-:newUser:          A meta field with no apparent purpose that is usually ``null`` unless explicitly set during creation or modification of a user via some API endpoint
-:phoneNumber:      The user's phone number
-:postalCode:       The postal code of the area in which the user resides
-:publicSshKey:     The user's public key used for the SSH protocol
-:registrationSent: If the user was created using the :ref:`to-api-users-register` endpoint, this will be the date and time at which the registration email was sent - otherwise it will be ``null``
-:role:             The integral, unique identifier of the highest-privilege role assigned to this user
-:rolename:         The name of the highest-privilege role assigned to this user
-:stateOrProvince:  The name of the state or province where this user resides
-:tenant:           The name of the tenant to which this user belongs
-:tenantId:         The integral, unique identifier of the tenant to which this user belongs
-:uid:              A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX user ID of the user - now it is always ``null``
-:username:         The user's username
 
 .. code-block:: http
 	:caption: Response Example
--- a/traffic_ops/testing/api/v14/user_test.go
+++ b/traffic_ops/testing/api/v14/user_test.go
@@ -15,6 +15,9 @@ package v14
 */
 
 import (
+	"bytes"
+	"encoding/json"
+	"fmt"
 	"strings"
 	"testing"
 	"time"
@@ -27,6 +30,7 @@ import (
 func TestUsers(t *testing.T) {
 	WithObjs(t, []TCObj{CDNs, Types, Tenants, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, DeliveryServices, Users}, func() {
 		UpdateTestUsers(t)
+		RolenameCapitalizationTest(t)
 		OpsUpdateAdminTest(t)
 		UserSelfUpdateTest(t)
 		UserUpdateOwnRoleTest(t)
@@ -42,10 +46,57 @@ func CreateTestUsers(t *testing.T) {
 
 		resp, _, err := TOSession.CreateUser(&user)
 		if err != nil {
-			t.Fatalf("could not CREATE user: %v\n", err)
+			t.Errorf("could not CREATE user: %v\n", err)
 		}
 		log.Debugln("Response: ", resp.Alerts)
 	}
+}
+
+func RolenameCapitalizationTest(t *testing.T) {
+
+	var err error
+	var jsonBytes []byte
+
+	user := tc.User{}
+	blob := []byte(`
+	{
+		"username": "test_user",
+		"email": "cooldude6@comcast.net",
+		"fullName": "full name is required",
+		"localPasswd": "better_twelve",
+		"confirmLocalPasswd": "better_twelve",
+		"role": 1,
+		"tenantId": 1
+	}`)
+	err = json.Unmarshal(blob, &user)
+	if err != nil {
+		t.Errorf("could not unmarshal json into struct: %v\n", err)
+	}
+
+	resp, _, err := TOSession.CreateUser(&user)
+	if err != nil {
+		t.Errorf("could not CREATE user: %v\n", err)
+	}
+	jsonBytes, err = json.Marshal(resp)
+	if err != nil {
+		fmt.Printf("coult not marshal struct to bytes: %v\n", err)
+	}
+	if bytes.Contains(jsonBytes, []byte("rolename")) {
+		t.Errorf("incorrect json was returned for PUT")
+	}
+
+	resp2, _, err := TOSession.GetUserByUsername(*user.Username)
+	if err != nil {
+		t.Errorf("could not GET user: %v\n", err)
+	}
+	jsonBytes, err = json.Marshal(resp2)
+	if err != nil {
+		fmt.Printf("coult not marshal struct to bytes: %v\n", err)
+	}
+	if bytes.Contains(jsonBytes, []byte("roleName")) {
+		t.Errorf("incorrect json was returned for GET")
+	}
+
 }
 
 func OpsUpdateAdminTest(t *testing.T) {
--- a/traffic_ops/testing/api/v14/user_test.go
+++ b/traffic_ops/testing/api/v14/user_test.go
@@ -30,7 +30,7 @@ import (
 func TestUsers(t *testing.T) {
 	WithObjs(t, []TCObj{CDNs, Types, Tenants, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, DeliveryServices, Users}, func() {
 		UpdateTestUsers(t)
-		TestRolenameCapitalization(t)
+		RolenameCapitalizationTest(t)
 		OpsUpdateAdminTest(t)
 		UserSelfUpdateTest(t)
 		UserUpdateOwnRoleTest(t)
@@ -52,7 +52,7 @@ func CreateTestUsers(t *testing.T) {
 	}
 }
 
-func TestRolenameCapitalization(t *testing.T) {
+func RolenameCapitalizationTest(t *testing.T) {
 
 	roles, _, _, err := TOSession.GetRoles()
 	if err != nil {
--- a/traffic_ops/testing/api/v14/user_test.go
+++ b/traffic_ops/testing/api/v14/user_test.go
@@ -17,7 +17,6 @@ package v14
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"strings"
 	"testing"
 	"time"
@@ -79,7 +78,7 @@ func RolenameCapitalizationTest(t *testing.T) {
 	}
 	jsonBytes, err = json.Marshal(resp)
 	if err != nil {
-		fmt.Printf("coult not marshal struct to bytes: %v\n", err)
+		t.Errorf("coult not marshal struct to bytes: %v\n", err)
 	}
 	if bytes.Contains(jsonBytes, []byte("rolename")) {
 		t.Errorf("incorrect json was returned for PUT")
@@ -91,7 +90,7 @@ func RolenameCapitalizationTest(t *testing.T) {
 	}
 	jsonBytes, err = json.Marshal(resp2)
 	if err != nil {
-		fmt.Printf("coult not marshal struct to bytes: %v\n", err)
+		t.Errorf("coult not marshal struct to bytes: %v\n", err)
 	}
 	if bytes.Contains(jsonBytes, []byte("roleName")) {
 		t.Errorf("incorrect json was returned for GET")
--- a/traffic_ops/testing/api/v14/user_test.go
+++ b/traffic_ops/testing/api/v14/user_test.go
@@ -16,7 +16,8 @@ package v14
 
 import (
 	"bytes"
-	"encoding/json"
+	"fmt"
+	"net/http"
 	"strings"
 	"testing"
 	"time"
@@ -29,7 +30,7 @@ import (
 func TestUsers(t *testing.T) {
 	WithObjs(t, []TCObj{CDNs, Types, Tenants, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, DeliveryServices, Users}, func() {
 		UpdateTestUsers(t)
-		RolenameCapitalizationTest(t)
+		TestRolenameCapitalization(t)
 		OpsUpdateAdminTest(t)
 		UserSelfUpdateTest(t)
 		UserUpdateOwnRoleTest(t)
@@ -51,48 +52,60 @@ func CreateTestUsers(t *testing.T) {
 	}
 }
 
-func RolenameCapitalizationTest(t *testing.T) {
+func TestRolenameCapitalization(t *testing.T) {
 
-	var err error
-	var jsonBytes []byte
+	roles, _, _, err := TOSession.GetRoles()
+	if err != nil {
+		t.Errorf("could not get roles: %v", err)
+	}
+	if len(roles) == 0 {
+		t.Fatalf("there should be at least one role to test the user")
+	}
 
-	user := tc.User{}
-	blob := []byte(`
+	tenants, _, err := TOSession.Tenants()
+	if err != nil {
+		t.Errorf("could not get tenants: %v", err)
+	}
+	if len(tenants) == 0 {
+		t.Fatalf("there should be at least one tenant to test the user")
+	}
+
+	// this user never does anything, so the role and tenant aren't important
+	blob := fmt.Sprintf(`
 	{
 		"username": "test_user",
 		"email": "cooldude6@example.com",
 		"fullName": "full name is required",
 		"localPasswd": "better_twelve",
 		"confirmLocalPasswd": "better_twelve",
-		"role": 1,
-		"tenantId": 1
-	}`)
-	err = json.Unmarshal(blob, &user)
+		"role": %d,
+		"tenantId": %d 
+	}`, *roles[0].ID, tenants[0].ID)
+
+	reader := strings.NewReader(blob)
+	request, err := http.NewRequest("POST", fmt.Sprintf("%v/api/1.4/users", TOSession.URL), reader)
 	if err != nil {
-		t.Errorf("could not unmarshal json into struct: %v\n", err)
+		t.Errorf("could not make new request: %v\n", err)
+	}
+	resp, err := TOSession.Client.Do(request)
+	if err != nil {
+		t.Errorf("could not do request: %v\n", err)
 	}
 
-	resp, _, err := TOSession.CreateUser(&user)
-	if err != nil {
-		t.Errorf("could not CREATE user: %v\n", err)
-	}
-	jsonBytes, err = json.Marshal(resp)
-	if err != nil {
-		t.Errorf("coult not marshal struct to bytes: %v\n", err)
-	}
-	if !bytes.Contains(jsonBytes, []byte("roleName")) {
-		t.Errorf("incorrect json was returned for PUT")
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(resp.Body)
+	strResp := buf.String()
+	if !strings.Contains(strResp, "roleName") {
+		t.Errorf("incorrect json was returned for POST")
 	}
 
-	resp2, _, err := TOSession.GetUserByUsername(*user.Username)
-	if err != nil {
-		t.Errorf("could not GET user: %v\n", err)
-	}
-	jsonBytes, err = json.Marshal(resp2)
-	if err != nil {
-		t.Errorf("coult not marshal struct to bytes: %v\n", err)
-	}
-	if !bytes.Contains(jsonBytes, []byte("rolename")) {
+	request, err = http.NewRequest("GET", fmt.Sprintf("%v/api/1.4/users?username=test_user", TOSession.URL), nil)
+	resp, err = TOSession.Client.Do(request)
+
+	buf = new(bytes.Buffer)
+	buf.ReadFrom(resp.Body)
+	strResp = buf.String()
+	if !strings.Contains(strResp, "rolename") {
 		t.Errorf("incorrect json was returned for GET")
 	}
 
--- a/traffic_ops/testing/api/v14/user_test.go
+++ b/traffic_ops/testing/api/v14/user_test.go
@@ -60,7 +60,7 @@ func RolenameCapitalizationTest(t *testing.T) {
 	blob := []byte(`
 	{
 		"username": "test_user",
-		"email": "cooldude6@comcast.net",
+		"email": "cooldude6@example.com",
 		"fullName": "full name is required",
 		"localPasswd": "better_twelve",
 		"confirmLocalPasswd": "better_twelve",
@@ -80,7 +80,7 @@ func RolenameCapitalizationTest(t *testing.T) {
 	if err != nil {
 		t.Errorf("coult not marshal struct to bytes: %v\n", err)
 	}
-	if bytes.Contains(jsonBytes, []byte("rolename")) {
+	if !bytes.Contains(jsonBytes, []byte("roleName")) {
 		t.Errorf("incorrect json was returned for PUT")
 	}
 
@@ -92,7 +92,7 @@ func RolenameCapitalizationTest(t *testing.T) {
 	if err != nil {
 		t.Errorf("coult not marshal struct to bytes: %v\n", err)
 	}
-	if bytes.Contains(jsonBytes, []byte("roleName")) {
+	if !bytes.Contains(jsonBytes, []byte("rolename")) {
 		t.Errorf("incorrect json was returned for GET")
 	}
 
@@ -136,7 +136,7 @@ func UserSelfUpdateTest(t *testing.T) {
 	user := resp[0]
 
 	fullName := "Oops-man"
-	email := "operator@comcast.net"
+	email := "operator@example.com"
 	user.FullName = &fullName
 	user.Email = &email
 
--- a/traffic_ops/testing/api/v14/user_test.go
+++ b/traffic_ops/testing/api/v14/user_test.go
@@ -17,14 +17,19 @@ package v14
 import (
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-tc"
+	toclient "github.com/apache/trafficcontrol/traffic_ops/client"
 )
 
 func TestUsers(t *testing.T) {
 	WithObjs(t, []TCObj{CDNs, Types, Tenants, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, DeliveryServices, Users}, func() {
 		UpdateTestUsers(t)
+		OpsUpdateAdminTest(t)
+		UserSelfUpdateTest(t)
+		UserUpdateOwnRoleTest(t)
 		GetTestUsers(t)
 		GetTestUserCurrent(t)
 	})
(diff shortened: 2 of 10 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3053/
Test PASSed.

ocket8888:
So to be clear does this change let people update users by submitting a `PUT` request to `/api/1.x/users` or does it require an integral, unique identifier path parameter?

moltzaum:
Yes

The way all endpoints work currently for golang endpoints is that both path and query parameters are supported for GET, PUT and DELETE (POST does not have a path parameter equivalent).

Regarding the documentation:

I could add a message like so:
> Use the ``id`` query parameter of a ``PUT`` request to the :ref:`to-api-users` endpoint instead.

Messages like this exist for most or all GET requests that have been re-written.
However I didn't initially add a message because in the docs `PUT /cdns` does not exist despite supporting a query parameter.

moltzaum:
I updated my comment. It does not exist; I misunderstood something based on how GET works.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3068/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3069/
Test PASSed.

traffic_ops/traffic_ops_golang/routes.go
```diff
@@ -162,15 +162,15 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
...
 		{1.4, http.MethodPost, `users/?(\.json)?$`, api.CreateHandler(user.GetTypeSingleton()), auth.PrivLevelOperations, Authenticated, nil},
-		//{1.4, http.MethodDelete, `users/{id}?(\.json)?$`, api.DeleteHandler(user.GetTypeSingleton()), auth.PrivLevelOperations, Authenticated, nil},
+		//{1.4, http.MethodDelete, `users/{id}$`, api.DeleteHandler(user.GetTypeSingleton()), auth.PrivLevelOperations, Authenticated, nil},
 
 		{1.1, http.MethodGet, `users/?(\.json)?$`, handlerToFunc(proxyHandler), auth.PrivLevelReadOnly, Authenticated, []Middleware{}},
 		{1.1, http.MethodGet, `users/{id}?(\.json)?$`, handlerToFunc(proxyHandler), auth.PrivLevelReadOnly, Authenticated, []Middleware{}},
-		{1.1, http.MethodPut, `users/{id}?(\.json)?$`, handlerToFunc(proxyHandler), auth.PrivLevelReadOnly, Authenticated, []Middleware{}},
-		{1.1, http.MethodPost, `users/?(\.json)?$`, handlerToFunc(proxyHandler), auth.PrivLevelReadOnly, Authenticated, []Middleware{}},
+		{1.1, http.MethodPut, `users/{id}?(\.json)?$`, handlerToFunc(proxyHandler), auth.PrivLevelOperations, Authenticated, []Middleware{}},
```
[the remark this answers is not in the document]
mitchell852: From routes.go

```
// 1.1 and 1.2 routes are simply a Go replacement for the equivalent Perl route. They may or may not conform with the API guidelines (https://cwiki.apache.org/confluence/display/TC/API+Guidelines).
// 1.3 routes exist only in a Go. There is NO equivalent Perl route. They should conform with the API guidelines (https://cwiki.apache.org/confluence/display/TC/API+Guidelines).
```

The idea here was that if you were simply rewriting a 1.1 Perl api endpoint in Go, it should be defined as 1.1 in routes.go to truly "override" it. It turns out that 99% of the routes in TrafficOpsRoutes.pm are 1.1 (only 1 is an 1.2 route - it's a weird asn one with a weird response structure).

So, imo, if i were to  look thru routes.go I should see:

- mostly 1.1 definitions because most are simply rewrites of Perl 1.1 endpoints
- one 1.2 definition that rewrites the "weird" asn endpoint
- some 1.3 endpoints that were new at the time
- some 1.4 endpoints that were new at the next time

my understanding is that every time something new is added (that is backwards compatible), we roll the minor version, 1.5, 1.6, etc...

but looking at routes.go, doesn't look like that was followed :(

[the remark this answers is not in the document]
rawlinp: The minor version should only be incremented once per release though, right? E.g. if we release 3.0.0, master becomes "the next release". So if I add a new field to the cachegroups API, I would increment those minor versions. Somebody else comes along and adds yet another new field to the cachegroups API. At that point the version has already been incremented since the 3.0.0 release, so they wouldn't have to increment the version again?

Also, would the new incremented version have to follow the latest incremented version? E.g. cachegroups API is 1.1 right now. If I add a field, should it be incremented to 1.4 (since that's the "latest" version)?

To this day I am still confused about our API versioning and how things should really be handled and how the versioning corresponds to release versions. For instance, why do we need different handlers for v1.2 and v1.3 delivery services but not everything else? Is it because some backwards-incompatible changes were made, in which case it should really be v1.2 and v2.0 (except we hadn't truly followed SemVer yet)?

We do have _some_ info about versioning in the API guidelines on the wiki, but that doesn't really answer any of these questions to common examples. We should consider adding a `VERSIONING.md` to the traffic_ops_golang directory with just the API guidelines w.r.t. versioning requirements, explaining the common cases like adding a new optional field to an existing endpoint, rewriting a Perl endpoint, or implementing an entirely new endpoint. Also, any pre-SemVer exceptions should be documented (if any). Even the `README.md` in that directory seems to give contradictory information about the API versioning because it has an example of rewriting the `cdns` endpoint but uses 1.2 rather than 1.1.

I can take on the `VERSIONING.md` stuff if I can get some answers to these questions. We can't really expect people to follow the versioning requirements if they're not readily available and easy to understand.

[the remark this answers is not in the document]
rob05c: >The minor version should only be incremented once per release though, right?

Right.

>Also, would the new incremented version have to follow the latest incremented version? E.g. cachegroups API is 1.1 right now. If I add a field, should it be incremented to 1.4 (since that's the "latest" version)?

Right. The version is API-wide, not per-endpoint. An addition to `/cachegroups` needs to be at `/api/1.4/cachegroups`, **not** `/api/1.2/cachegroups`.

>We should consider adding a VERSIONING.md to the traffic_ops_golang directory

+1

We've had trouble remembering which is the current/next minor as well, and not adding new API minor versions when we should. I think maybe we should have a file somewhere, with which minor versions were added in which TC versions. Not sure the best way to ensure it's updated though; maybe a note in the Release Management wiki page?

ocket8888:
So wait a minute, our ability to make breaking changes is directly tied to the ATC major version? Why are they separate versioning schemes, then?

rob05c:
>So wait a minute, our ability to make breaking changes is directly tied to the ATC major version? Why are they separate versioning schemes, then?

Sort of. They're separate because they're separate things. We shouldn't be required to roll the TO API version, just because some other component in TC had a breaking change. That's unnecessary and confusing.

That said, adding a new major version to TO isn't a breaking change. We could decide to add `/api/2.0` and we wouldn't need to increment the TC version. Because nothing broke. Now, if we stopped serving `/api/1.x` we would. But we shouldn't do that anyway, people need time to port clients, APIs should always serve one major version back.

This seems to make @rawlinp's case, of the need to put all this in a document.

traffic_ops/traffic_ops_golang/routes.go
```diff
@@ -162,15 +162,15 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
...
 		//Incrementing version for users because change to Nullable struct. Delete is completely new.
 		{1.4, http.MethodGet, `users/?(\.json)?$`, api.ReadHandler(user.GetTypeSingleton()), auth.PrivLevelReadOnly, Authenticated, nil},
-		{1.4, http.MethodGet, `users/{id}?(\.json)?$`, api.ReadHandler(user.GetTypeSingleton()), auth.PrivLevelReadOnly, Authenticated, nil},
-		{1.4, http.MethodPut, `users/{id}?(\.json)?$`, api.UpdateHandler(user.GetTypeSingleton()), auth.PrivLevelOperations, Authenticated, nil},
+		{1.4, http.MethodGet, `users/{id}$`, api.ReadHandler(user.GetTypeSingleton()), auth.PrivLevelReadOnly, Authenticated, nil},
+		{1.4, http.MethodPut, `users/{id}$`, api.UpdateHandler(user.GetTypeSingleton()), auth.PrivLevelOperations, Authenticated, nil},
 		{1.4, http.MethodPost, `users/?(\.json)?$`, api.CreateHandler(user.GetTypeSingleton()), auth.PrivLevelOperations, Authenticated, nil},
-		//{1.4, http.MethodDelete, `users/{id}?(\.json)?$`, api.DeleteHandler(user.GetTypeSingleton()), auth.PrivLevelOperations, Authenticated, nil},
+		//{1.4, http.MethodDelete, `users/{id}$`, api.DeleteHandler(user.GetTypeSingleton()), auth.PrivLevelOperations, Authenticated, nil},
```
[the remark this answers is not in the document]
moltzaum: When I first put it in we were considering adding an implementation of some sort. I remember there being a github issue and possibly something discussed on the apache dev list. It seems that it might be awhile before we actually do something like that. There might be a comment about it someplace else in the code.

I normally wouldn't keep commented out code, but this is a single line and tells the viewer fairly explicitly "we don't have a delete for this endpoint".

traffic_ops/traffic_ops_golang/routes.go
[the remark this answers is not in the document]
moltzaum: > my understanding is that every time something new is added (that is backwards compatible), we roll the minor  version, 1.5, 1.6, etc...
> 
> but looking at routes.go, doesn't look like that was followed :(
@mitchell852 

But if something in made in go, isn't it safer to increment the version since some changes can be made on accident and incrementing the version limits the number of things you need to look for? Breaking changes are easier to notice than minor changes.

If messages aren't the same across golang and perl would that count as a minor change? I'd have to look at it, but I'm thinking of validation messages in particular. I think the minor version should be incremented in cases of uncertainty at the very least.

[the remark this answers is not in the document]
moltzaum: > Is the id portion still supposed to be optional?
> 
> Also, are we supposed to be still proxying 1.1 routes to the Perl even though those endpoints have now been rewritten in Go as 1.4? Should the Go routes be 1.1?  
@rawlinp 
It's not supposed to be optional. I'll update it.

I remember that the reverse proxy worked without an explicit route defined. However, by adding everything explicitly and removing the code that does the automatic reverse proxy we would be able to do something..

I'm forgetting what it was. It might have had something to do with implementing roles and capabilities or something related to priv_level. I thought a PR was made that has a list of all perl routes and if I'm not mistaken it was done by @rob05c?

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -217,16 +221,45 @@ func (user *TOUser) Read() ([]interface{}, error, error, int) {
 			return nil, nil, fmt.Errorf("parsing user rows: %v", err), http.StatusInternalServerError
 		}
 
+		// role is a required field for the endpoint, but not for an item in the database
+		// I doubt a nil check is needed, but I'm including it just in case
+		if user.RoleName == nil {
+			return nil, nil, fmt.Errorf("role name is nil", err), http.StatusInternalServerError
```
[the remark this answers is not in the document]
moltzaum: Looks like I intended to use %v, which would have used the actual formatting.

traffic_ops/traffic_ops_golang/routes.go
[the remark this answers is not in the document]
rawlinp: It's more confusing than informative IMO. When I saw that I thought there was actually a Golang implementation but it wasn't finished or something. In this case there is just a no-op handler to satisfy the CRUDer interface.

traffic_ops/traffic_ops_golang/routes.go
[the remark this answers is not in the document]
rawlinp: If there are Perl routes which need to be explicitly declared for some reason then that's probably fine to do in whatever PR for that new feature would be, but my understanding is that the endpoints rewritten in golang need to replace the Perl endpoints. If there are Perl routes explicitly declared as 1.1 and the golang rewrite at 1.4, then any clients requesting the 1.1 endpoint will still be hitting the Perl instead of the new golang version even though the golang is supposed to be backwards-compatible.

traffic_ops/traffic_ops_golang/user/user.go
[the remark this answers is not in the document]
rawlinp: It would have tried to format a nil value since there isn't an actual value in `err` yet -- copy-pasta mistake I think.

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -217,16 +221,45 @@ func (user *TOUser) Read() ([]interface{}, error, error, int) {
...
+		user.RoleName = nil
+
 		users = append(users, *user)
 	}
 
 	return users, nil, nil, http.StatusOK
 }
 
+func (user *TOUser) privCheck() (error, error, int) {
+	privLevel, _, err := dbhelpers.GetPrivLevelFromRoleID(user.ReqInfo.Tx, *user.Role)
```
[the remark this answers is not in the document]
moltzaum: Not sure how that will make it clearer since I read `requestedPrivLevel` as a getter. `privCheck` makes sure the requested priv level is accessible to the current user. I'll add a comment.

traffic_ops/traffic_ops_golang/routes.go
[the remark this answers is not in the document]
moltzaum: Well, that's kinda what it was. I made it be a no-op handler just in case someone thought "hey, I'll just uncomment this". That would have been a bad idea since the generic delete would not have been aware of tenancy or priv_level.

traffic_ops/traffic_ops_golang/routes.go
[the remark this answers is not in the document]
moltzaum: > Any clients requesting the 1.1 endpoint will still be hitting the Perl instead of the new golang version even though the golang is supposed to be backwards-compatible.

Yes. Unless there is a hotfix that is needed for the current version, I think this is what is supposed to be done. To my knowledge we don't have any version number for patches. Incrementing a minor version should be done for backwards-compatible changes (see https://semver.org/).

traffic_ops/traffic_ops_golang/user/user.go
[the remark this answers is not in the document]
moltzaum: Oh, I thought you meant change the method name.

traffic_ops/traffic_ops_golang/routes.go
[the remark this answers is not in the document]
rawlinp: We kinda-sorta follow SemVer but not exactly (yet). Technically, nearly all of our golang rewritten APIs have made breaking changes by not including the same response fields as the Perl version when doing a POST/PUT, so really they should all be 2.x now until they've been truly fixed.

However, my concern is not really about the proper version number. My concern is that the Perl endpoints are still accessible even though the Golang endpoints are supposed to replace them completely. Old clients requesting 1.1 should be served by the rewritten golang version, since it's supposed to be backwards-compatible. We shouldn't serve Perl to 1.1 requests and Go to 1.2+ requests IMO. If there are no explicit routes to the Perl version, then I don't really care what the api version number is as long as it "makes sense".

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3094/
Test PASSed.

traffic_ops/traffic_ops_golang/routes.go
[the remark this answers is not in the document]
ocket8888: :+1: The proxy handler is - afaik - meant to handle cases where an endpoint hasn't been re-written, but the Go routes handler would shadow it if it were not explicitly declared. The idea is that if an endpoint has been re-written in a way that's consistent with the behavior of a 1.1 endpoint, that 1.1 endpoint should no longer be served by Perl code.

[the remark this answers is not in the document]
moltzaum: Hmm.. interesting.

traffic_ops/testing/api/v14/user_test.go
```diff
@@ -42,12 +45,59 @@ func CreateTestUsers(t *testing.T) {
...
+
+	resp, _, err := TOSession.CreateUser(&user)
+	if err != nil {
+		t.Errorf("could not CREATE user: %v\n", err)
+	}
+	jsonBytes, err = json.Marshal(resp)
+	if err != nil {
+		t.Errorf("coult not marshal struct to bytes: %v\n", err)
+	}
+	if bytes.Contains(jsonBytes, []byte("rolename")) {
```
rawlinp: I'd maybe invert the logic here and fail the test if it does NOT contain `"roleName": "<insert requested role name here>"`. We could probably "fix" this API by just returning the same data in `"rolename"` _and_ `"roleName"`. That might actually be easier than trying to remember which API returns which form of capitalization. I think that would also mean being able to use the same golang struct for all the users endpoints, kind of like how you changed it originally to embed a struct with the common users fields, which we could probably use here rather than checking the jsonBytes.

I might bring that users API change up on the mailing list to be tackled in a different PR. What do you think?

traffic_ops/testing/api/v14/user_test.go
```diff
@@ -42,12 +45,59 @@ func CreateTestUsers(t *testing.T) {
...
+
+	resp2, _, err := TOSession.GetUserByUsername(*user.Username)
+	if err != nil {
+		t.Errorf("could not GET user: %v\n", err)
+	}
+	jsonBytes, err = json.Marshal(resp2)
+	if err != nil {
+		t.Errorf("coult not marshal struct to bytes: %v\n", err)
+	}
+	if bytes.Contains(jsonBytes, []byte("roleName")) {
```
rawlinp: ditto the comment above, would probably be better to fail the test if the response does NOT contain `"rolename": "<insert role name here>"`.

traffic_ops/testing/api/v14/user_test.go
```diff
@@ -42,12 +45,59 @@ func CreateTestUsers(t *testing.T) {
...
+func RolenameCapitalizationTest(t *testing.T) {
+
+	var err error
+	var jsonBytes []byte
+
+	user := tc.User{}
+	blob := []byte(`
+	{
+		"username": "test_user",
+		"email": "cooldude6@comcast.net",
```
rawlinp: probably shouldn't use a real domain here, that could potentially be a real email address. Better to use example.com

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -224,7 +225,7 @@ func (user *TOUser) Read() ([]interface{}, error, error, int) {
 		// role is a required field for the endpoint, but not for an item in the database
 		// I doubt a nil check is needed, but I'm including it just in case
 		if user.RoleName == nil {
-			return nil, nil, fmt.Errorf("role name is nil", err), http.StatusInternalServerError
+			return nil, nil, fmt.Errorf("role name is nil"), http.StatusInternalServerError
```
rawlinp: nit: there's a couple instances of `fmt.Errorf` without any actual formatting and could just be `errors.New` instead

traffic_ops/testing/api/v14/user_test.go
moltzaum: > I think that would also mean being able to use the same golang struct for all the users endpoints, kind of like how you changed it originally to embed a struct with the common users fields, which we could probably use here rather than checking the jsonBytes.

I'm not 100% sure what you're saying here. Currently the same struct is used for all `users/{id}?` endpoints. The `users/current` endpoint isn't affected by the fix and was separate beforehand for other reasons.

> We could probably "fix" this API by just returning the same data in "rolename" and "roleName"
> ...
> I might bring that users API change up on the mailing list to be tackled in a different PR. What do you think?

I think that just makes things potentially more confusing and is unnecessary, but you can ask the mailing list if you want to see other opinions.

traffic_ops/testing/api/v14/user_test.go
moltzaum: I'll change it..

but comcast specifically has a naming convention that wouldn't have 'cooldude6', nor is it used for anything.

ocket8888: Not used for anything today. The naming standard could change in the future.

traffic_ops/testing/api/v14/user_test.go
rawlinp: So we have this `commonUserFields` struct that is embedded in both the structs used for the /users endpoints and the /user/current endpoints. What I'm saying is we could have `roleName` and `roleNameGet` in the `commonUserFields` struct. Then for all the users endpoints we would simply populate both fields with the actual role name. So all the users endpoints would return:
```
...
"rolename": "admin",
"roleName": "admin",
...
```
Then we wouldn't have to special-case the GET endpoint vs the POST/PUT endpoints. Does that make sense?

traffic_ops/testing/api/v14/user_test.go
rob05c: >Same goes for example IP addresses/CIDRs, generally you're supposed to use the designated "example" ranges that are reserved just for that

`192.0.2.0/24` and `2001:DB8::/32` are one of each.

https://tools.ietf.org/html/rfc5737
https://tools.ietf.org/html/rfc3849
https://tools.ietf.org/html/rfc2606

traffic_ops/traffic_ops_golang/routes.go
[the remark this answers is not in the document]
rob05c: >The proxy handler is - afaik - meant to handle cases where an endpoint hasn't been re-written, but the Go routes handler would shadow it if it were not explicitly declared.

Right now, but FYI that will be changing once Capabilities are enforced (https://github.com/apache/trafficcontrol/pull/2791). After that, no Perl route without a Go declaration will be proxied. Just FYI.

traffic_ops/testing/api/v14/user_test.go
rawlinp: > I don't like the idea of having two role names in part because the user might try to set only one. How would we know which role name to use if we needed to reference it? Were you originally thinking of making a getter and setter surrounding both fields?

That's the thing though -- these are read-only fields. The user _can_ set them in the request, but anything the user sets them to should be ignored. The only _write_ field the API cares about is `role` which is the id of the requested role. The `roleName` fields are only populated after reading the role name from the database to return back to the user.

Come to think of it, the PR as it is now would break any golang clients that hit the `users` endpoint, because the struct field has been renamed `RoleName` -> `RoleNameGET`. Existing clients will expect the `RoleName` field to contain the actual role name, but will instead end up with a nil value which would likely cause a panic. They will have no idea about the new `RoleNameGET` field. So I think the only solution we can provide here is to keep a single `RoleName` field but do custom JSON marshalling so that both `"roleName"` and `"rolename"` get populated with the same value in the JSON. Then the JSON unmarshalling would simply take the one `"roleName"` field as the value for `RoleName`.

This would prevent existing golang clients from breaking, and non-golang clients would be able to read either `"roleName"` or `"rolename"` from the JSON since it will be the same value (since they've likely become accustomed to reading both, depending on the method used).

Also, once we actually fix the golang API to return the proper fields on POST/PUT, we won't end up with a second variation of the special-casing to do `roleName if GET but rolename if PUT/POST` in the other two endpoints as well. The JSON marshalling would just populate both fields under the hood.

Does this make sense? I think it will work and will prevent breaking not only non-golang clients but also golang clients as well.

rob05c: >populate both fields with the actual role name. So all the users endpoints would return:
>
>...
>"rolename": "admin",
>"roleName": "admin",

I'm not a fan of sending unnecessary fields in JSON. It's unnecessary and potentially confusing to users. IMO making the backend a little more complex and having a clean interface, is better than a clean backend and dirty interface.

Which is to say, I think we should use different structs for the different endpoints.

rob05c: You're both right about the problems. The `if bytes.Contains(jsonBytes, []byte("rolename")) {` is error-prone, and renaming a struct field to `RoleNameGET` will break clients. Renaming the struct itself will also break, but it will break their compile and it's easy to fix; whereas the field name breaks at runtime and is very subtle and difficult to track down.

Likewise, sending duplicate fields in the JSON is not only a mess, but will lead to problems with no sane solution. Which value is used if someone posts an object with two different values in `rolename` and `roleName`?

The only solution that solves all these problems, is to separate the structs: `UserGet`, `UserPost`, `UserPutReq`, `UserPutResp`, etc. Because some endpoints are the same and some aren't, type aliases wouldn't be a bad idea either, e.g. `type UserPutResp = UserGet`.

That solves all the problems except (1) the code is a little more complex, and (2) clients may fail to compile and have to change e.g. `tc.User` to `tc.UserGet`. IMO (1) is much less bad than all the other problems, and (2) is inevitable, and a compile-time break is much better than the alternatives.

moltzaum: Can't I just keep RoleName as it was and designate it for GET, then make RoleNamePOST instead?

moltzaum: Also, I think if we separated things into structs there would only need to be two. One for GET, and one for the others.

rob05c: >What are the existing clients? Wasn't it made within this same release?

Ah, I thought the client functions had been around for a while. If it's the same release, then nothing breaks, great.

>Also, I think if we separated things into structs there would only need to be two. One for GET, and one for the others.

Sure, if they're all the same. The docs weren't clear and I didn't test. I still think type aliases would make things clearer, but I won't insist on it.

Incidentally, we should fix the docs. The docs say /users/id PUT takes "username", but then the example JSON for the same endpoint uses `"userName"`. Same for /users POST.

moltzaum: Ok, so the current fix for rolenames works then. I will double check the docs for special casing.

docs/source/api/user_current.rst
```diff
@@ -169,28 +169,6 @@ Request Structure
 
 Response Structure
 ------------------
-:addressLine1:     The user's address - including street name and number
```
moltzaum: @ocket8888 I removed this to match the documentation for `user/current/update POST`. The response example is below and doesn't return the actual user information. Do you know if this response is something we want to include in the future for this endpoint?

ocket8888: Yes. All `PUT` requests should print out the resulting object representation in the response.
Did you modify `PUT /api/1.x/user_current` such that it no longer does that?

moltzaum: I did not modify it. That is how it is. The response example shows the current output.

moltzaum: The `PUT /api/1.x/user/current` endpoint is not written in go yet.

ocket8888: okay

rawlinp: If we're going to be consistent with API 2.0, `POST`s and `PUT`s should probably all return the same data that can be retrieved by doing a subsequent `GET`. _Most_ of our Perl endpoints did that, but nearly all of the golang-rewritten versions fail to do that currently, which is technically a breaking-API regression.

Anyways, thanks for cleaning up this lying documentation. 👍

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3151/
Test PASSed.

traffic_ops/traffic_ops_golang/routes.go
```diff
@@ -159,18 +159,12 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `user/{id}/deliveryservices/available/?(\.json)?$`, user.GetAvailableDSes, auth.PrivLevelReadOnly, Authenticated, nil},
 		{1.1, http.MethodPost, `user/login/?$`, login.LoginHandler(d.DB, d.Config), 0, NoAuth, nil},
 
-		//User: CRUD (1.4 then 1.1)
-		//Incrementing version for users because change to Nullable struct. Delete is completely new.
+		//User: CRUD
+		//Incrementing version for users because change to Nullable struct.
 		{1.4, http.MethodGet, `users/?(\.json)?$`, api.ReadHandler(user.GetTypeSingleton()), auth.PrivLevelReadOnly, Authenticated, nil},
```
rawlinp: Thanks for deleting the explicit Perl 1.1 routes, but I think the go routes need to be 1.1 now too so that the Go handler will serve 1.1 and above, not just 1.4.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3154/
Test PASSed.

PR #3222, opened 2019-01-15
ORT.py now uses the official ATC client

ocket8888:
## What does this PR do?
Previously, the Python-based ORT implementation had re-written a bunch of functionality already present in the Python client. This PR aims to change ORT.py to instead make use of the client.

I also took the opportunity to fix some documentation build warnings generated by the python client and/or typos in existing API documentation, as well as add a new endpoint method and fix some existing broken ones.

- [x] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Python ORT
- [x] Python Client

Build CDN-in-a-Box and verify that ORT runs properly, clears server updates, and check that the origin content can be accessed through the default Delivery Service (via `http://video.demo1.mycdn.ciab.test/`) after the dust settles. You may need to take a snapshot manually after startup due to some timing issues.

## Check all that apply
- [ ] This PR includes tests


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3055/
Test PASSed.

traffic_control/clients/python/trafficops/tosession.py
```diff
@@ -701,16 +691,18 @@ def get_cdns_domains(self):
 	#
 	# CDN Topology
 	#
-	@api_request(u'get', u'cdns/{cdn_name:s}/configs', (u'1.2', u'1.3',))
```
rawlinp: This might be a typo for `/api/$version/cdns/configs"`

ocket8888: no, you don't need to add that to the path

Shihta:
Excuse me, I have a question...

Is there any document about these 2 python library?
[traffic_control/clients](https://github.com/apache/trafficcontrol/tree/master/traffic_control/clients/python)
and
[ort](https://github.com/apache/trafficcontrol/tree/master/infrastructure/cdn-in-a-box/ort)

I am not quite clear about the functions and differences of these two packages

traffic_control/clients/python/trafficops/tosession.py
rawlinp: That's not what I mean; we have an API endpoint there so the decorator was probably meant to be `cdns/configs`

Shihta:
Why there are 2 versions of ORT? (perl and python)
Is the python version going to replace perl version?

ocket8888:
That's the general plan, although we always seem to have bigger fish to fry.

Currently the only thing that uses ORT.py is CDN-in-a-Box - because it doesn't share the Perl version's dependency on systemd.

traffic_control/clients/python/trafficops/tosession.py
ocket8888: oh. Yeah, you're probably right

ocket8888:
Uhhh sure. The Perl script uses systemd to enable/disable/start/stop all system daemon services - i.e. `systemctl`. ORT.py makes calls to `trafficserver` directly. See, in a Docker container because we really don't want to be running any more services than are absolutely necessary and basically we don't want to run systemd inside of them if we can help it.

That warning just means it'll ignore services besides `trafficserver`. By default I think TO is also telling it to start/stop/enable/disable `ntpd` and maybe something else like `sshd`. Without systemd, those can't be worked with. ORT.py will _try_ to use systemd just like the Perl script, but _unlike_ the Perl script it can still work with trafficserver when systemd is unavailable. (Although, much of the systemd behaviour is untested, because at the moment it only needs to run/is tested inside of CiaB)

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3082/
Test PASSed.

JBevillC:
I've test this PR successfully. 

Tests procedure performed:
1) Check out PR branch
2) Rebuilt all TC component RPMs.
3) Rebuilt all TC docker containers
4) Verified traffic router started up properly and that delivery service 'demo1' is available via DNS.
5) Verified curl -Lvs http://video.demo1.mycdn.ciab.test/ works.
6) Verified both mid and edge containers appear to be provisioned correctly with updates to ORT client.

PR #3228, opened 2019-01-16
Optimize regex_revalidate.config API endpoint

### LINKED ISSUE #3232: The regex_revalidate.config API endpoint is too slow for large numbers of invalidations
This endpoint can be unbearably slow and CPU-intensive with hundreds of
current invalidation jobs. We should dig into this endpoint (currently still Perl) and see if we can make some easy optimizations.

rawlinp:
#### What does this PR do?

This endpoint can be unbearably slow and CPU-intensive with hundreds of
current invalidation jobs. Do some simple optimization to get this
endpoint more on par with other ATS config file endpoints.

Locally with ~600 current invalidation jobs, I'm seeing about 5-6x speed
improvement with these changes. Because of the added SQL filtering, the
endpoint now returns almost instantly for CDNs that have little or no
current invalidations. Previously, generating this file for ANY CDN
would require iterating over ALL current jobs across ALL CDNs, doing a
bunch of work before throwing away the results from other CDNs.

Fixes #3232

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

1. Create some invalidation jobs for a test DS and compare the before/after output of regex_revalidate.config (with/without this change)
2. Create 500 invalidation jobs and repeat step 1, compare request times

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
+++ b/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
@@ -1767,7 +1767,16 @@ sub regex_revalidate_dot_config {
 	my $max_hours = $max_days * 24;
 	my $min_hours = 1;
 
-	my $rs = $self->db->resultset('Job')->search( { start_time => \$interval }, { prefetch => 'job_deliveryservice' } );
+	my $date = Date::Manip::Date->new();
+
+	my $rs = $self->db->resultset('Job')->search(
+		{
+			start_time => \$interval,
+			"job_deliveryservice.cdn_id" => $cdn_obj->id,
+		},
+		{
+			prefetch => 'job_deliveryservice'
+		} );
 	while ( my $row = $rs->next ) {
 		next unless defined( $row->job_deliveryservice );
 
@@ -1787,7 +1796,6 @@ sub regex_revalidate_dot_config {
 			next;
 		}
 
-		my $date       = new Date::Manip::Date();
 		my $start_time = $row->start_time;
 		my $start_date = ParseDate($start_time);
 		my $end_date   = DateCalc( $start_date, ParseDateDelta( $ttl . ':00:00' ) );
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3060/
Test PASSed.

rob05c:
This endpoint is pretty small. Is there a reason not to rewrite it in Go instead?

Go should be even faster, and has the added benefit of reducing load on Postgres compared to using Mojo/workers. Not to mention insisting new changes be in Go will make the rewrite go faster.

rawlinp:
It should be rewritten in Go, but that would take much longer to implement and test than this low-hanging fruit optimization done in the Perl that is only a slight modification of the current implementation. Essentially a SQL filter has been added, and an expensive object initialization has been moved out of a loop -- super simple to test and low-risk to deploy.

rawlinp:
Most of the work done in this endpoint is date parsing and calculations in pure Perl, so I imagine this would get even faster than the optimized Perl simply by being written in Go. The slowness of the Date::Manip Perl package is even a known issue called out in some of the websites I found when investigating it.

rawlinp:
Definitely agree on putting this near the head of the endpoints to port, because it's almost completely open to self-service users who can easily add invalidation jobs to their delivery services.

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
```diff
@@ -1803,15 +1811,11 @@ sub regex_revalidate_dot_config {
...
-		my $job_cdn_id = $row->job_deliveryservice->cdn_id;
-		if ( $cdn_obj->id == $job_cdn_id ) {
-
-			# if there are multipe with same re, pick the longes lasting.
-			if ( !defined( $regex_time{ $row->asset_url } )
-				|| ( defined( $regex_time{ $row->asset_url } ) && $purge_end > $regex_time{ $row->asset_url } ) )
-			{
-				$regex_time{ $row->asset_url } = $purge_end;
-			}
+		# if there are multipe with same re, pick the longes lasting.
```
dangogh: minor nit -- 2 misspellings

rawlinp: done

mtorluemke:
thanks @rawlinp !

dangogh:
moving the $date setting outside the loop is obviously correct and a welcome change -- haven't looked at the rest of the changes in detail -- someone signing up for testing this?   definitely +90billion on this...

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3062/
Test PASSed.

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
alficles: Yeah, I saw the spellings. They were there in the original before it was unindented. Probably an error in translation from the original Dutch. 👀

Issue #3239, opened 2019-01-17
PUT /api/x/deliveryservices/{id} nil panic if matchList is null in request


rawlinp:
If a `PUT /api/x/deliveryservices/{id}` request contains a null `"matchList"` field (which is not a required field and should generally be null in a PUT request anyways since it's a read-only field), the endpoint will throw a nil dereference panic.

PR #3240, opened 2019-01-17
Return a 500 response rather than an empty 200 response on panic recovery

### LINKED ISSUE #3238: Traffic Ops Golang panic recovery returns an empty 200 response
I recently found a nil dereference panic in TO-go, and instead of getting a 500 back, the client gets an empty 200 response from TO-go. This breaks clients that are expecting a valid response body when given back a 200. TO-go should return a 500 in this case since it was most likely unable to handle the request successfully.

rawlinp:
It doesn't really make sense to return an empty 200 response if the
server recovers from a panic and most likely didn't fulfill the request
successfully.

#### What does this PR do?

Fixes #3238 

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

```
cd traffic_ops/traffic_ops_golang
go test .
```
Until the fix for #3239 is merged, you can follow the instructions there to cause TO-Golang to recover from a nil dereference panic. Before this PR, it returns an empty 200 response. With this PR, it should return a 500 response.

#### Check all that apply

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/wrappers.go
+++ b/traffic_ops/traffic_ops_golang/wrappers.go
@@ -100,7 +100,8 @@ func wrapPanicRecover(h http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
 			if err := recover(); err != nil {
-				log.Errorf("panic: (err: %v) stacktrace:\n%s\n", err, stacktrace())
+				api.HandleErr(w, r, nil, http.StatusInternalServerError, nil, fmt.Errorf("panic: (err: %v) stacktrace:\n%s\n", err, stacktrace()))
+				return
 			}
 		}()
 		h(w, r)
--- a/traffic_ops/traffic_ops_golang/wrappers_test.go
+++ b/traffic_ops/traffic_ops_golang/wrappers_test.go
@@ -83,6 +83,29 @@ func TestWrapHeaders(t *testing.T) {
 	}
 }
 
+// TestWrapPanicRecover checks that a recovered panic returns a 500
+func TestWrapPanicRecover(t *testing.T) {
+	f := wrapPanicRecover(func(w http.ResponseWriter, r *http.Request) {
+		var foo *string
+		bar := *foo // will throw nil dereference panic
+		w.Write([]byte(bar))
+	})
+	f = wrapHeaders(f)
+
+	w := httptest.NewRecorder()
+	r, err := http.NewRequest("", "/", nil)
+	if err != nil {
+		t.Error("Error creating new request")
+	}
+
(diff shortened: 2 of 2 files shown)


traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -134,6 +134,7 @@ func handleSimpleErr(w http.ResponseWriter, r *http.Request, statusCode int, use
 		return
 	}
 	*r = *r.WithContext(context.WithValue(r.Context(), tc.StatusKey, statusCode))
+	w.WriteHeader(statusCode)
```
rob05c: Is this necessary for the 500 to be returned properly?

The wrapper stuff is supposed to require the line immediately above it,
`*r = *r.WithContext(context.WithValue(r.Context(), tc.StatusKey, statusCode))`

(Something I'd like to fix, but it is what it is today). 

If the `WriteHeader` doesn't actually do anything, it should be removed. If it does, we should figure out what changed, and if we can remove the StatusKey Context stuff.

traffic_ops/traffic_ops_golang/wrappers.go
```diff
@@ -100,7 +100,7 @@ func wrapPanicRecover(h http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		defer func() {
 			if err := recover(); err != nil {
-				log.Errorf("panic: (err: %v) stacktrace:\n%s\n", err, stacktrace())
+				api.HandleErr(w, r, nil, http.StatusInternalServerError, nil, fmt.Errorf("panic: (err: %v) stacktrace:\n%s\n", err, stacktrace()))
```
rob05c: Nitpick: a `return` after this would be safer. It doesn't matter right now, but if someone changed the func to do something else, they could miss that it needs to return after `HandleErr`. Safer to add it now.

traffic_ops/traffic_ops_golang/api/api.go
rawlinp: So the `gzipResponse` function appears to handle writing the status header. I added this line because my added unit test was failing, but I've updated the unit test to also wrap with `wrapHeaders` which includes `gzipResponse`.

traffic_ops/traffic_ops_golang/wrappers.go
rawlinp: done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3075/
Test PASSed.

PR #3241, opened 2019-01-17
Fix nil pointer dereference in PUT /deliveryserivces/{id} endpoint

### LINKED ISSUE #3239: PUT /api/x/deliveryservices/{id} nil panic if matchList is null in request

rawlinp:
If this endpoint receives a json request without a "matchList" array
(which is not a required field), the endpoint will panic on a nil
pointer dereference.

Load the delivery service matchList from the DB before dereferencing it.
As a side effect, this also prevents a user from passing in their own
"matchList" array which would potentially cause the wrong hostname to be
updated into the delivery service's ssl keys.

#### What does this PR do?

Fixes #3239 

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Run the API integration tests and/or use curl to make a `PUT /api/x/deliveryservices/{id}` request with a json body that doesn't contain a `"matchList"` field.

#### Check all that apply

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v14/deliveryservices_test.go
+++ b/traffic_ops/testing/api/v14/deliveryservices_test.go
@@ -17,10 +17,10 @@ package v14
 
 import (
 	"strconv"
+	"testing"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-tc"
-	"testing"
 )
 
 func TestDeliveryServices(t *testing.T) {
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3074/
Test PASSed.

PR #3254, opened 2019-01-23
Grove: Advanced cache key with JWT claims

### LINKED ISSUE #3253: Grove: Advanced cache key with JWT claims
Hello,

I'm facing the next problem: I have an web application (API) that responds in a different way (content and formatting) based on the JWT claims on the authorization token and I need cache that responses. After a little research about how Grove works I think that feature could be done over 2 plugins:

1. JWT authorization and claims exporting plugin: Validates token and copies the desired claims to the request header
2. Advanced cache behavior plugin: Modifies the default cache key to add request headers from a whitelist

A config example could be:
```
  "plugins": {
    "advanced_cache_behavior": {
      "whitelist": ["X-Claim-Access", "X-Claim-Device"]
    },
    "jwt_auth": {
      "export_claims": ["access", "device"],
      "secrets": [
        {
          "iss": "api.example.cl",
          "aud": "example.cl"
          "alg": "rs256",
          "value": "/etc/keys/examplecl_key.pem"
        }
   }
```
What do you think?

astroza:
#### What does this PR do?

Fixes #3253

- [ ] Documentation
- [X] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Loading the plugins with a valid configuration and check if it misses with different claims

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/grove/plugin/jwt_auth.go
+++ b/grove/plugin/jwt_auth.go
@@ -87,8 +87,9 @@ func jwtAuth(icfg interface{}, d OnRequestData) bool {
 	if claims, ok := token.Claims.(*jwt.MapClaims); ok {
 		for _, claimName := range jwtConfig.ExportClaims {
 			if value := (*claims)[claimName]; value != nil {
-				// What if value is not string ?
-				d.R.Header.Set(fmt.Sprintf("X-Claim-%s", strcase.ToCamel(claimName)), value.(string))
+				if valueString, ok := value.(string); ok {
+					d.R.Header.Set(fmt.Sprintf("X-Claim-%s", strcase.ToCamel(claimName)), valueString)
+				}
 			}
 		}
 	}
@@ -149,9 +150,9 @@ func jwtLoad(b json.RawMessage) interface{} {
 	jwtConfig := JwtConfig{}
 	err := json.Unmarshal(b, &jwtConfig)
 	if err != nil {
-		log.Errorln("JWT loading config, unmarshalling JSON: " + err.Error())
+		log.Errorln("JWT config loading error: " + err.Error())
 		return nil
 	}
-	log.Debugf("JWT config load success")
+	log.Debugf("JWT config loaded")
 	return &jwtConfig
 }
--- a/grove/plugin/jwt_auth.go
+++ b/grove/plugin/jwt_auth.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 
 	log "github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/dgrijalva/jwt-go"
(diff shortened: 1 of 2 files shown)


asfgit:

astroza:
I don't feel that way.  Plugins are optional and for people using JWT a feature like this could determinate if Grove fits or not.

On the other hand, I think advanced_cache_behavior is generic enough to mimic a feature from a CDN provider

Cheers

rob05c:
@astroza Ok, I don't really agree, but I don't feel that strongly about it. I'll dismiss the review and do a regular code review.

dneuman64:
@astroza Thanks for your contribution!  I am glad you took the time to contribute to our project.

grove/plugin/jwt_auth.go
```diff
@@ -0,0 +1,162 @@
...
+func init() {
+	AddPlugin(10000, Funcs{onRequest: jwtAuth, startup: jwtStartup, load: jwtLoad})
+}
+
+func jwtAuth(icfg interface{}, d OnRequestData) bool {
+	if d.R.Method == http.MethodOptions {
+		return false
+	}
+	jwtContext := (*d.Context).(JwtContext)
+	jwtConfig := icfg.(*JwtConfig)
```
rob05c: Do you mind adding a nil check and return, if this is `nil`?
```
if icfg == nil {
  return
}
```
That will let Grove run instead of crashing, if the config is nil.

grove/plugin/jwt_auth.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/iancoleman/strcase"
+)
+
+type JwtSecret struct {
+	Issuer   string `json:"iss"`
+	Audience string `json:"aud"`
+	Algoritm string `json:"alg"`
+	Kid      string `json:"kid"`
+	Value    string `json:"value"`
+}
```
alficles: Consider using a JWK for the Algorithm, Kid, and Value. That will give you considerably more flexibility in the future as additional algorithms become desirable. There's a good JWK library I've used for this purpose here: https://godoc.org/github.com/lestrrat/go-jwx/jwk . (It's got a good JWT library next to it that does some of the heavy lifting if you wanted to do some heavier refactoring; though I don't think that's necessarily important for this.)

Using the standard format for JWKs will work nicely with the JSON-based config, too, since it's a standard way of formatting the keys that users of JWTs will expect and understand.

grove/plugin/jwt_auth.go
```diff
@@ -0,0 +1,162 @@
...
+func getSecretIndexKey(kid, iss, aud string) string {
+	if kid != "" {
+		return kid
+	}
+	s := ""
+	if iss != "" {
+		s += iss
+	}
+	if aud != "" {
+		s += ":" + aud
```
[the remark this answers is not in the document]
alficles: Yes. Though if we're here, we're already in a fairly strange place, because there really should be a `kid` under typical circumstances and both the `iss` and `aud` claims are mandatory. This is a stopgap for finding the right key when no `kid` is available. Obviously, a `kid` is strongly advised because otherwise, key-rolling is impossible. But you don't always get to pick your clients.

astroza:
I will split it

PR #3264, opened 2019-01-25
CIAB: Dynamically detect server netmask

Shihta:
#### What does this PR do?

Since CIAB change to use Dynamic DNS, the netmask will also change.
Therefore, it should be dynamically detect.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Other CIAB

Start CIAB
Check the netmask of db, trafficvault, enroller, or dns.
Their netmask should be the real netmask (255.255.240.0)

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/infrastructure/cdn-in-a-box/traffic_ops/set-to-ips-from-dns.sh
+++ b/infrastructure/cdn-in-a-box/traffic_ops/set-to-ips-from-dns.sh
@@ -22,7 +22,7 @@ base_data_dir="/traffic_ops_data"
 servers_dir="${base_data_dir}/servers"
 profiles_dir="${base_data_dir}/profiles"
 
-service_names='db trafficops trafficops-perl trafficportal trafficmonitor trafficvault trafficrouter edge mid origin enroller socksproxy client vnc dns'
+service_names='db trafficops trafficops-perl trafficportal trafficmonitor trafficvault trafficrouter enroller dns'
 
 service_domain='infra.ciab.test'
 
@@ -38,11 +38,22 @@ done
 
 service_ips="${gateway_ip}"
 service_ip6s="${gateway_ip6}"
+INTERFACE=$(ip link | awk '/\<UP\>/ && !/LOOPBACK/ {sub(/@.*/, "", $2); print $2}')
+NETMASK=$(route | awk -v INTERFACE=$INTERFACE '$8 ~ INTERFACE && $1 !~ "default"  {print $3}')
+DIG_IP_RETRY=10
 
 for service_name in $service_names; do
 	service_fqdn="${service_name}.${service_domain}"
 
-	service_ip="$(dig +short ${service_fqdn} A)"
+	for (( i=1; i<=DIG_IP_RETRY; i++ )); do
+		service_ip="$(dig +short ${service_fqdn} A)"
+		if [ -z "${service_ip}" ]; then
+			printf "service \"${service_fqdn}\" not found in dns, count=$i, waiting ...\n"
+			sleep 3
+		else
+			break
+		fi
+	done
 
   #
 	# TODO add a way to determine if a service wasn't built in the Compose,
@@ -54,6 +65,7 @@ for service_name in $service_names; do
 	if [ -z "${service_ip}" ]; then
 		# TODO sleep and try again? Up to n times?
 		printf "setting ips from dns: service \"${service_fqdn}\" not found in dns, skipping!\n"
+		continue
 	fi
 
 	service_ip6="$(dig +short $service_name AAAA)"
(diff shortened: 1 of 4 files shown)


asfgit:

Shihta:
@JBevillC This PR is updated today so please make sure you are using the latest code.

rob05c:
@Shihta I'm trying to test this, and it looks like the netmask didn't get set properly in my CiaB. 

Everything started, and I was able to request the delivery service successfully.

But, the dns, db, enroller, and trafficvault netmasks are empty:

```
traffic_ops=# select host_name, ip_netmask from server;                                                                                                                                                                                                                                                 
    host_name    | ip_netmask  
-----------------+-------------
 dns             | 
 db              | 
 enroller        | 
 trafficvault    | 
 trafficportal   | 255.255.0.0
 trafficops      | 255.255.0.0
 trafficops-perl | 255.255.0.0
 mid             | 255.255.0.0
 edge            | 255.255.0.0
 trafficrouter   | 255.255.0.0
 trafficmonitor  | 255.255.0.0
(11 rows)
```

Any ideas?

Shihta:
It's quite weird... I'm curious if your source code is the same as this branch.
There are 2 processes that handle the IP obtaining:

* P1: dns, db and enroller use `set-to-ips-from-dns.sh` to make enrollment. You can check this [folder](https://github.com/apache/trafficcontrol/tree/master/infrastructure/cdn-in-a-box/traffic_ops_data/servers)
* P2: the others call function `to-enroll` to make enrollment. TV belongs to P2 after the PR #3287 is merged

If there is an error in this commit, it should not affect TV...
maybe you should rebuild your container image, especially `trafficops-perl` and `trafficvault`.
everything seems correct in my environment:


So there are something I need your help:

1. Could you send me your trafficops-perl log? You can get it by command: `docker logs CONTAINER_ID_OF_TO_PERL > LOG_FILE_NAME 2>&1`. I want to check if there are some log like this:

```
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'db'
setting ips from dns: trying service file '/traffic_ops_data/servers/020-db_server.json'
setting ips from dns: service file '/traffic_ops_data/servers/020-db_server.json' exists, adding IPs
service "trafficops.infra.ciab.test" not found in dns, count=1, waiting ...
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'trafficops'
setting ips from dns: trying service file ''
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'trafficops-perl'
setting ips from dns: trying service file ''
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'trafficportal'
setting ips from dns: trying service file ''
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'trafficmonitor'
setting ips from dns: trying service file ''
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'trafficvault'
setting ips from dns: trying service file ''
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'trafficrouter'
setting ips from dns: trying service file ''
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'enroller'
setting ips from dns: trying service file '/traffic_ops_data/servers/030-enroller_server.json'
setting ips from dns: service file '/traffic_ops_data/servers/030-enroller_server.json' exists, adding IPs
setting ips from dns: checking file for dir '/traffic_ops_data/servers' service 'dns'
setting ips from dns: trying service file '/traffic_ops_data/servers/010-dns_server.json'
setting ips from dns: service file '/traffic_ops_data/servers/010-dns_server.json' exists, adding IPs
```

The old file `040-trafficvault_server.json` should not appear here.

2. Check my modified command of obtaining netmask can be correctly executed in your environment:

```
# docker exec -ti CONTAINER_ID_OF_TO_PERL bash
[root@trafficops-perl app]# INTERFACE=$(find /sys/class/net -type l '!' -name lo -exec basename '{}' \;)
[root@trafficops-perl app]# echo $INTERFACE
eth0
[root@trafficops-perl app]# NETMASK=$(route |grep -v default |grep $INTERFACE |awk '{print $3}')
[root@trafficops-perl app]# echo $NETMASK
255.255.0.0
```

Thanks!

dangogh:
I see the same problem -- here's the output from your method:   ```
[root@trafficops-perl app]# INTERFACE=$(find /sys/class/net -type l '!' -name lo -exec basename '{}' \;)
[root@trafficops-perl app]# echo $INTERFACE
tunl0 eth0 ip6tnl0
[root@trafficops-perl app]# NETMASK=$(route |grep -v default |grep $INTERFACE |awk '{print $3}')
grep: eth0: No such file or directory
grep: ip6tnl0: No such file or directory
```

**Comment by dangogh** [comment]
So the above is not eliminating the tunnel interfaces `tunl0` and `ip6tnl0`.   Wondering if that's specific to docker on mac?     I found the following works for my case, and seems should be more general:

```
INTERFACE=$(ip link | awk '/\<UP\>/ && !/LOOPBACK/ {sub(/@.*/, "", $2); print $2}')
NETMASK=$(route | awk -v INTERFACE=$INTERFACE '$8 ~ INTERFACE && $1 !~ "default"  {print $3}')
```

Shihta:
the command of getting network mask looks awesome
but I'm confused that is that necessary to detect interface dynamically
docker / k8s seem to use `eth0` as the default name in the major network interface
so hard coding it seems a right and simple way
any idea?

dangogh:
I would think that would depend on the image being used in docker,  and the above seems to be a good way to pick up the right one..  At any rate,  I'd like to get this in, but the `find` method you've used doesn't work for us -- it returns multiple names..

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3960/
Test PASSed.

Shihta:
The 2 commands for `INTERFACE` and `NETMASK` have been updated.
I verified this PR with the latest code as well, it seems okay.

PR #3273, opened 2019-01-28
fixed minor issue preventing package installation in Python2

ocket8888:
## What does this PR do?
Fixes an issue with the Python client `setup.py` not being Python2-compatible - related to getting the version number out of `__version__.py`.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] ATC Python client

Try to build/install the package for Python versions 2.7 and 3.4+ (I know the docs say we only support 3.6+, but 3.4 is the newest version available from elrepo so I think we ought to maintain compatibility with it)

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_control/clients/python/trafficops/__version__.py
+++ b/traffic_control/clients/python/trafficops/__version__.py
@@ -18,13 +18,13 @@ Versioning
 The :mod:`trafficops.__version__` module contains only the :data:`__version__` "constant" which
 gives the version of this *Apache-TrafficControl package* and **not** the version of
 *Apache Traffic Control* for which it was made. The two are versioned separately, to allow the
-client to grow in a version-controlled manner despite the relatively slow release cadence of Apache
+client to grow in a version-controlled manner without being tied to the release cadence of Apache
 Traffic Control as a whole.
 
-We guarantee version 1.0 will work (in general) with Apache Traffic Control version 3.0 (release
-pending at the time of this writing). New functionality will be added as the :ref:`to-api` evolves,
-but changes to this client will remain non-breaking for existing code using it until the next major
-version is released.
+Version 1.0 is supported for use with Apache Traffic Control version 3.0 (release pending at the
+time of this writing). New functionality will be added as the :ref:`to-api` evolves, but changes to
+this client will remain non-breaking for existing code using it until the next major version is
+released.
 
 .. deprecated:: 1.0
 	The v1.0 release of this client deprecates support of Python2. Versions 2.0 and onward will
--- a/traffic_control/clients/python/trafficops/__version__.py
+++ b/traffic_control/clients/python/trafficops/__version__.py
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,4 +12,4 @@
 # limitations under the License.
 #
 
-__version__ = '1.1.1'
+__version__ = '1.1.1-1'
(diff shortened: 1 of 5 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3115/
Test PASSed.

traffic_control/clients/python/trafficops/__version__.py
```diff
@@ -14,4 +12,4 @@
 # limitations under the License.
 #
 
-__version__ = '1.1.1'
+__version__ = '1.1.1-1'
```
rawlinp: This kind of goes back to the versioning discussion we were having on your other PR. If we're going to be following some kind of versioning practice it should probably be documented in a `VERSIONING.md` file in the `clients/python` directory. For instance, I assume you're adding `-1` since it's not really part of the version and wouldn't cause clients already installed at 1.1.1 to upgrade to this version?

ocket8888:
I wish. No, we _deprecated _Python2. As far as I know, the client code all works (*worked) in Python2, and it's used heavily by Comcast SREs who use only Python2. The best commitment I could get out of the mailing list was to drop it for the 4.0 release (although if we don't release 4.0 by the end of the year I intend to ask again about maybe dropping a micro release for it).

traffic_control/clients/python/trafficops/__version__.py
ocket8888: I think `pip` would still do the upgrade if you asked it to. Not totally sure. I added a patch number because this change didn't add functionality or really change anything, it's just a bugfix. You think a micro version is more appropriate? I'm not totally clear on that to be honest.

As far as documenting the versioning scheme, I can certainly get behind that but i'd prefer it to live in our main docs rather than a separate MarkDown file. My suggestion would be to include it in the docstring of the `__version__` module, or maybe in the package docstring. That way, sphinx-autodoc will pick it up and put it in the main docs, and it will live with the package wherever it's hosted and however it's installed.

rawlinp: Bugfixes should be `patch` in the `major.minor.patch` scheme if we're choosing to use SemVer for the client versioning scheme. I'm okay with describing the versioning scheme in a docstring, I'd just be a little concerned that people wouldn't easily notice it. A `VERSIONING.md` file would clearly stick out in a directory listing, and people generally look for things like a `README.md` when learning about a new project.

ocket8888: I added a documented versioning strategy, hope it makes sense.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3118/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3119/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3120/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3121/
Test PASSed.

traffic_control/clients/python/trafficops/__version__.py
```diff
@@ -14,4 +12,26 @@
...
 #
 
-__version__ = '1.1.1'
+"""
+Versioning
+==========
+The :mod:`trafficops.__version__` module contains only the :data:`__version__` "constant" which
+gives the version of this *Apache-TrafficControl package* and **not** the version of
+*Apache Traffic Control* for which it was made. The two are versioned separately, to allow the
+client to grow in a version-controlled manner despite the relatively slow release cadence of Apache
```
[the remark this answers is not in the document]
ocket8888: To be clear, that was just meant to convey that the release cadence of ATC as a whole is slow _relative to the client_ - not a stab at the release cadence in general.

PR #3274, opened 2019-01-28
Fix Traffic Ops CDN DNSSEC generation when no delivery services exist

rob05c:
Fixes an edge case of generating DNSSEC keys on a CDN when no delivery
services exist on that CDN (for example, if generating immediately
after creating the CDN). The edge case caused keys to be generated
with no domain name.

API Tests don't support Riak yet.

Manually tested, verified DNSSEC generation on a new CDN with no DSes works, and generates keys with the CDN's domain. Tested that DNSSEC generation and regeneration on a CDN with DSes still works.

#### What does this PR do?

Fixes #(issue_number)

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Create a new CDN and generate DNSSEC keys. Verify the keys at `/api/1.2/cdns/name/your-new-cdn-name/dnsseckeys` have `name` of the cdn's domain name, and not `.`.

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/cdn/dnssec.go
+++ b/traffic_ops/traffic_ops_golang/cdn/dnssec.go
@@ -259,7 +259,7 @@ type CDNDS struct {
 // getCDNDeliveryServices returns basic data for the delivery services on the given CDN, as well as the CDN name, or any error.
 func GetCDNDeliveryServices(tx *sql.Tx, cdn string) ([]CDNDS, error) {
 	q := `
-SELECT ds.xml_id, ds.protocol, t.name as type, ds.routing_name, cdn.domain_name as cdn_domain
+SELECT ds.xml_id, ds.protocol, t.name as type, ds.routing_name
 FROM deliveryservice as ds
 JOIN cdn ON ds.cdn_id = cdn.id
 JOIN type as t ON ds.type = t.id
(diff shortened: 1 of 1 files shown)


traffic_ops/traffic_ops_golang/cdn/dnssec.go
```diff
@@ -245,7 +257,7 @@ type CDNDS struct {
 }
 
 // getCDNDeliveryServices returns basic data for the delivery services on the given CDN, as well as the CDN name, or any error.
-func getCDNDeliveryServices(tx *sql.Tx, cdn string) ([]CDNDS, string, error) {
+func GetCDNDeliveryServices(tx *sql.Tx, cdn string) ([]CDNDS, error) {
 	q := `
 SELECT ds.xml_id, ds.protocol, t.name as type, ds.routing_name, cdn.domain_name as cdn_domain
```
rawlinp: The `cdn.domain_name as cdn_domain` should be removed from the query if it's not being scanned out from the result, right?

rob05c: Yep. Fixed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3145/
Test PASSed.

PR #3279, opened 2019-01-29
Records.config ds profile parameters

ezelkow1:
#### What does this PR do?

Fixes #(issue_number)

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Add records.config parameters to a DS profile, make sure existing remaps stay the same

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
+++ b/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
@@ -2679,7 +2679,7 @@ sub remap_dot_config {
 
 					#Check that the parameter name matches the records.config syntax
 					if ( $rec_params[0] eq 'CONFIG') {
-						my @rec_values = split(' ', $ds->{'param'}->{'records.config'}->{$rc_entry}, 3);
+						my @rec_values = split(' ', $ds->{'param'}->{'records.config'}->{$rc_entry}, 2);
 
 						#Check parameter value matches records.config syntax
 						if ( {map{$_=>1}('INT','FLOAT','STRING')}->{$rec_values[0]}) {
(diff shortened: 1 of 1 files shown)


asfgit:

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
```diff
@@ -2671,6 +2671,27 @@ sub remap_dot_config {
...
+			if ( defined( $ds->{'param'}->{'records.config'} ) ) {
+				foreach my $rc_entry ( keys %{ $ds->{'param'}->{'records.config'} } ) {
+					my @rec_params = split(' ', $rc_entry, 3);
+
+					#Check that the parameter name matches the records.config syntax
+					if ( $rec_params[0] eq 'CONFIG') {
+						my @rec_values = split(' ', $ds->{'param'}->{'records.config'}->{$rc_entry}, 2);
+
+						#Check parameter value matches records.config syntax
+						if ( {map{$_=>1}('INT','FLOAT','STRING')}->{$rec_values[0]}) {
```
dangogh: this is pretty cluttered and would make more sense to create that hash prior to the loop:
```
my %goodtype = map { $=>1 } qw{INT FLOAT STRING};
...
if ( $goodtype{$rec_values[0]}) {
...
}
```

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
```diff
@@ -2671,6 +2671,27 @@ sub remap_dot_config {
...
+							$mid_remap{ $ds->{org} } .= " \@plugin=conf_remap.so \@pparam=" . $rec_params[1] . "=" . $rec_values[1];
```
dangogh: since you're not interpolating any vars in the string,   use single-quotes and drop the backslashes:
```
... .= ' @plugin=conf_remap.so @pparam=...'
```

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
```diff
@@ -2769,6 +2790,27 @@ sub build_remap_line {
 		}
 	}
 
+	#Add DS profile records.config entries as overrided ATS parameters
+	if ( defined( $remap->{'param'}->{'records.config'} ) ) {
```
dangogh: this repeats the above -- would it make sense to make it a sub?

PR #3281, opened 2019-01-30
CIAB: Fixed the method of obtaining IP

Shihta:
#### What does this PR do?

The sequence of setting DNS record is shown as follows:

1. Set /etc/resolv.conf (infrastructure/cdn-in-a-box/dns/set-dns.sh)
2. Obtaining IP by "dig +short ${my_host}" (infrastructure/cdn-in-a-box/dns/insert-self-into-dns.sh)
3. Set DNS record (infrastructure/cdn-in-a-box/dns/insert-self-into-dns.sh)

It can not obtain IP after the /etc/resolv.conf changed.
This commit provide one way to fix it by changing the method of obtaining IP.
It uses parsing the ifconfig result instead of dig.

This commit also deletes insert-db-into-dns.sh.
DB should use insert-self-into-dns.sh instead of insert-db-into-dns.sh as well.

Beyond that, add trafficvault into the trafficops-perl dependency.
That could avoid the failure of obtaining trafficvault IP when trafficops-perl run set-to-ips-from-dns.sh.

Fixes #(issue_number)

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Other CIAB

This is a bug fix, so just simply start CIAB and then check if `http://video.demo1.mycdn.siab.test` is available.

#### Check all that apply

- [ ] This PR includes tests


asfgit:

infrastructure/cdn-in-a-box/docker-compose.yml
```diff
@@ -87,6 +89,7 @@ services:
     depends_on:
       - db
       - enroller
+      - trafficvault
```
dangogh: this won't fix the problem of trafficvault not being available for trafficops use -- it does not enforce startup order,  only build dependencies.

That conflicts with one of the goals we had:  ability to build and test trafficops in isolation.

Shihta: Sorry that I didn't know it need the ability to build and test trafficops in isolation
I can remove it to meet this requirement.

But on the other hand, It does guarantee startup order of container.
Do you mean that this is not enough?
Should I add more steps in `set-to-ips-from-dns.sh` to guarantee that trafficvault registered its DNS record before `set-to-ips-from-dns.sh` add trafficvault IP to `trafficcontrol/infrastructure/cdn-in-a-box/traffic_ops_data/servers/040-trafficvault_server.json`?

ocket8888: adding TV as a dependency of TO guarantees that it will be _started_ before TO, but that doesn't guarantee it will be _done_ starting by the time TO needs it. "starting" the container is an instantaneous action that spins off the container's `CMD` as an asynchronous procedure. Ordering within that cannot be guaranteed. I think your idea about using the enroller could work, if I understand you right.

Shihta: @ocket8888 You are right. that what I want to say.
@dangogh Since it need more work, I'll remove this line first and then send it in another PR.

Shihta: Please refer to #3287 for this dependency problem

dangogh:
ok to test

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3214/
Test PASSed.

JBevillC:
Hi @Shihta,

Thanks for submitting this PR:

I tested this PR and it appears some of the servers are still being added with the wrong subnet:

Problems I found:
1) The `db`, `dns`, `enroller`, and `trafficvault` servers are still being set with a `255.255.255.0` subnet.  
2) The `db` and `enroller` servers also have empty ipAddress.

My suggestion is to delete all the files at `infrastructure/cdn-in-a-box/traffic_ops_data/servers` and have the server json payload generated dynamically. It doesn't make sense to keep static files around at this point in my opinion.
```

$ cd infrastructure/cdn-in-a-box
$ make very-clean
$ make -j 7
$ alias mydc="docker-compose "` \
    `"-f $CIAB_HOME/docker-compose.yml "` \
    `"-f $CIAB_HOME/docker-compose.expose-ports.yml "` \
    `"-f $CIAB_HOME/optional/docker-compose.socksproxy.yml "` \
    `"-f $CIAB_HOME/optional/docker-compose.socksproxy.expose-ports.yml "` \
    `"-f $CIAB_HOME/optional/docker-compose.vnc.yml "` \
    `"-f $CIAB_HOME/optional/docker-compose.vnc.expose-ports.yml "
$ mydc kill && mydc rm -fv && docker volume prune -f 
$ mydc build
$ mydc up -d 

[root@trafficops app]# to-get 'api/1.4/servers/' | jq '.response[] | { (.hostName):{"ipAddress":.ipAddress, "ipNetmask":.ipNetmask, "ipGateway":.ipGateway}}' | jq -s 'add'
{
  "trafficrouter": {
    "ipAddress": "172.25.0.14",
    "ipNetmask": "255.255.0.0",
    "ipGateway": "172.25.0.1"
  },
  "trafficmonitor": {
    "ipAddress": "172.25.0.9",
    "ipNetmask": "255.255.0.0",
    "ipGateway": "172.25.0.1"
  },
  "trafficops": {
    "ipAddress": "172.25.0.15",
    "ipNetmask": "255.255.0.0",
    "ipGateway": "172.25.0.1"
  },
  "trafficops-perl": {
    "ipAddress": "172.25.0.7",
    "ipNetmask": "255.255.0.0",
    "ipGateway": "172.25.0.1"
  },
  "db": {
    "ipAddress": "",
    "ipNetmask": "255.255.255.0",
    "ipGateway": "172.25.0.1"
  },
  "trafficportal": {
    "ipAddress": "172.25.0.8",
    "ipNetmask": "255.255.0.0",
    "ipGateway": "172.25.0.1"
  },
  "trafficvault": {
    "ipAddress": "172.25.0.11",
    "ipNetmask": "255.255.255.0",
    "ipGateway": "172.25.0.1"
  },
  "dns": {
    "ipAddress": "172.25.0.4",
    "ipNetmask": "255.255.255.0",
    "ipGateway": "172.25.0.1"
  },
  "enroller": {
    "ipAddress": "",
    "ipNetmask": "255.255.255.0",
    "ipGateway": "172.25.0.1"
  },
  "edge": {
    "ipAddress": "172.25.0.12",
    "ipNetmask": "255.255.0.0",
    "ipGateway": "172.25.0.1"
  },
  "mid": {
    "ipAddress": "172.25.0.10",
    "ipNetmask": "255.255.0.0",
    "ipGateway": "172.25.0.1"
  }
}
```

Shihta:
@JBevillC thanks for helping me review it!

The 2 problems you found are not caused by this commit.
As you mentioned that there are some files in `infrastructure/cdn-in-a-box/traffic_ops_data/servers`.
They will be modified by `infrastructure/cdn-in-a-box/traffic_ops/set-to-ips-from-dns.sh` and then be sent by container enroller.

In fact, I have sent a PR #3264 trying to solve these 2 problems.
In PR #3264, it add a retry mechanism in obtaining IP and fix the netmask issue.
I didn't highlight it in slack because I'm not sure it's a right direction.

Should we just fix them by the PR #3264?
or should we let each service goes to enroll by themselves? like your opinion

It seems like the option 2 is better than 1.
If we want to select option 2, I'll close PR #3264 first and work the selection 2 in the future.

------------

Back to this PR
It aim to fix the sequence of registering DNS record, not service enrollment of TO.
I don't know why, but registering DNS record of DB sometimes just fails in my environment.
That why I sent this PR

JBevillC:
@Shihta Thanks for the follow up.  I'm going to re-test PRs #3264 and #3281 together (merged) against the latest master branch.  Testing them both together seems like a better approach since they are highly related to the same area of CiaB DNS IP discovery and assignment.  Thanks again for your contribution.

Issue #3225, opened 2019-01-16
Remove CacheURL Delivery Service Field


jhg03a:
CacheURL was a plugin that was deprecated in ATS6 and removed in ATS7.  ATS6 is now dead.  We should remove all traces of cacheURL from the TO API, TODB Database, Native-client API Libraries, & TP Web UI.

This is previously discussed in #2957 and #1130

jhg03a:
Consensus appears to have been reached and a deprecation notice issued for the 3.x release.

https://lists.apache.org/thread.html/d0a8805315819b5e4d9d9bf7f19bfd2b895924dc4247bec299b3ba15@%3Cdev.trafficcontrol.apache.org%3E

shamrickus:
Working on this, see #4644

mitchell852:
i feel like this should be done by now with 4 merged PRs :)

is it @shamrickus ?

shamrickus:
No. It's now officially deprecated, which it wasn't before. We were returning the field in api v3 so until that's gone we can't get rid of it completely

mitchell852:
> No. It's now officially deprecated, which it wasn't before. We were returning the field in api v3 so until that's gone we can't get rid of it completely

gotcha. thanks for the clarification. hard to get rid of these things, huh?

PR #3284, opened 2019-01-30
Update API Documentation with CacheURL deprecation

jhg03a:
This is a precursor to work being performed on #3225

#### What does this PR do?

This is a precursor to work being performed on #3225

- [X] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Read at the documentation

#### Check all that apply

- [ ] This PR includes tests
- [X] This PR includes documentation updates


docs/source/api/servers_id_deliveryservices.rst
```diff
@@ -50,7 +50,7 @@ Response Structure
 ------------------
 :active:                   ``true`` if the Delivery Service is active, ``false`` otherwise
 :anonymousBlockingEnabled: ``true`` if :ref:`Anonymous Blocking <anonymous_blocking-qht>` has been configured for the Delivery Service, ``false`` otherwise
-:cacheurl:                 A setting for a deprecated feature of now-unsupported Trafficserver versions
+:cacheurl:                 Deprecated. A setting for a deprecated feature of now-unsupported Trafficserver versions
```
ocket8888: You should use the `deprecated` directive. It would look like this:

```rst
:cacheurl:                 A setting for a deprecated feature of now-unsupported Trafficserver versions

    .. deprecated:: 3.0
        This field is deprecated as of Traffic Control 3.0, and will be removed in version 4.0

:ccrDnsTtl:                The Time To Live (TTL) of the DNS response for A or AAAA record queries requesting the IP 
```

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3131/
Test PASSed.

docs/source/api/servers_id_deliveryservices.rst
jhg03a: To say that is removed in 4.0 may or may not be accurate though.  Wouldn't it be better to say that's only supported through 3.0 and subject to removal in 4.0 or later?

jhg03a: Presumably once it's gone for good, it's also stripped from the documentation.  Trying to keep up with when issues are slated for releases in documentation seems presumptuous.

ocket8888: Oh, feel free to word it however you think is best. I was just demonstrating the syntax. The only thing that's truly important is that first line `.. deprecated:: 3.0`

ocket8888:
Actually, I take that back. There is one other important thing. Because this is in the API docs, you should make note of the fact that it's version 3.0 of _Traffic Control_ that deprecates it. Since we version our API separately there could be some confusion otherwise.

docs/source/api/servers_id_deliveryservices.rst
ocket8888: Actually, I take that back. There is one other important thing. Because this is in the API docs, you should make note of the fact that it's version 3.0 of Traffic Control that deprecates it. Since we version our API separately there could be some confusion otherwise.

docs/source/admin/traffic_ops/using.rst
```diff
@@ -441,7 +441,7 @@ The fields in the Delivery Service view are:
 |                                            |                                                                                                                                                                        |
 |                                            | **Note:** you will not be able to save a Regex Remap Expression if you have Query String Handling set to drop query strings at the edge. See :ref:`regex-remap`.       |
 +--------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-| Cache URL expression                       | Deprecated.                                                                                                                                                            |
+| Cache URL expression                       | Deprecated in Traffic Control 3.x and subject to removal in Traffic Control 4.x or later.                                                                                                                                                            |
```
ocket8888: This leaves the table malformed. The table borders and column/row delimiters need to remain intact for it to compile. Should just be a whitespace change.

docs/source/api/deliveryservices.rst
```diff
@@ -51,7 +51,11 @@ Response Structure
 ------------------
 :active:                   ``true`` if the Delivery Service is active, ``false`` otherwise
 :anonymousBlockingEnabled: ``true`` if :ref:`Anonymous Blocking <anonymous_blocking-qht>` has been configured for the Delivery Service, ``false`` otherwise
-:cacheurl:                 Deprecated. A setting for a deprecated feature of now-unsupported Trafficserver versions
+:cacheurl:                 A setting for a deprecated feature of now-unsupported Trafficserver versions
+
+	.. deprecated:: ATCv3.0
+	This field has been deprecated in Traffic Control 3.x and is subject to removal in Traffic Control 4.x or later
```
ocket8888: So I think this may be my fault. The "normal" syntax for the `.. deprecated` directive is 

```rst
.. deprecated:: version A description of the deprecation.
```
... but because it treats the first string (whitespace-delimited) specially, lots of people (myself included) prefer to write it as:

```rst
.. deprecated:: version
	A description of the deprecation.
```

However, that may somewhat obscure the fact that the description must fall within the directive block. Essentially what I'm saying is that you need to indent that second line one more time. That will cause it to be treated as a part of the `.. deprecated` directive block.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3210/
Test PASSed.

PR #3286, opened 2019-01-30
Pattern-based consistent hashing feature for TR/TO/TP

rivasj:
#### What does this PR do?
Pattern-based consistent hashing is a new feature that allows the user to influence cache selection (and delivery service selection for steering ds's) by adjusting the request path before it is passed to the consistent hasher in Traffic Router by applying a regex group matching. This is accomplished via a new delivery service field on HTTP and Steering delivery services called "Consistent Hash Regex" that is propagated to TR through the CRConfig.

Traffic Router parses the new optional regex field from the CRConfig and uses it during cache selection (HTTP delivery services) or delivery service selection (Steering delivery services).

For HTTP delivery services:
- consistent hash regex is applied to request path to extract pieces of the original request path to 'normalize' a string before passing to the consistent hasher for cache selection

For Steering delivery services:
- consistent hash regex is applied to request path to extract pieces of the original request path to 'normalize' a string before passing to the consistent hasher for delivery service selection
- the selected (target) delivery service then inherits the consistent hash regex from the parent/steering delivery service and uses it as an HTTP delivery service does

Additionally, I've included a Traffic Portal "Test Tool" so that users can test a regex against a request path to determine exactly what will be used by Traffic Router's consistent hasher. There is a link to the test tool underneath the form field in Traffic Portal.

- [x] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [x] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

- run Traffic Router unit and integration tests
- run Traffic Ops unit and integration tests
- install the components (TP/TO/TR) and verify the "consistentHashRegex" field makes it into the CrConfig
- verify the TP "Test Tool" works as expected by providing various inputs

#### Check all that apply

- [x] This PR includes tests
- [x] This PR includes a database migration (ensure that migration sequence is correct)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3146/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3150/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3153/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3158/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3171/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3195/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3200/
Test PASSed.

rivasj:
> I was able to test the new DS field in TP and saw that it made its all the way to the cr-config, so that looks good.
> 
> Using the TP Test Regex tool, I was able to get the Traffic Router spinning its wheels on an evil regex.
> Regex: /(a+)+$
> String: /aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaax
> 
> I don't see catastrophic backtracking to be a common occurance with how this will be used, but it is possible to spike the router CPU with bad requests like the one above.
> 
> The Match and Pattern libraries being used do not offer configurable safeguards for this, so a word of warning in the documentation might be useful. Either this or you can use an interrupt to kill off regex processing that is taking too long.

Added restrictions in both TP and the TR API to restrict the length of request paths hitting the endpoint to safeguard TR from inputs like those in this comment.

docs/source/api/consistenthash.rst
```diff
@@ -0,0 +1,50 @@
...
+..
+
+.. _to-api-consistenthash:
+
+***************
+``consistenthash``
+***************
+Test pattern based consistent hashing for a Delivery Service using a regex and a request path
+
+``POST``
```
mitchell852: whenever i think of `POST` i think of create. i guess it's kind of "creating" a consistent hash but not really in the sense that other POSTs create. What about something like:

GET /api/1.4/consistenthash?regex=&requestpath=&cndId=

but not sure if a regex value would make that impossible

traffic_ops/traffic_ops_golang/routes.go
```diff
@@ -407,6 +408,9 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `deliveryservices/{id}/urlkeys/?(\.json)?$`, deliveryservice.GetURLKeysByID, auth.PrivLevelReadOnly, Authenticated, nil},
 		{1.1, http.MethodGet, `riak/bucket/{bucket}/key/{key}/values/?(\.json)?$`, apiriak.GetBucketKey, auth.PrivLevelAdmin, Authenticated, nil},
 
+		//Pattern based consistent hashing endpoint
+		{1.1, http.MethodPost, `consistenthash/?$`, consistenthash.Post, auth.PrivLevelReadOnly, Authenticated, nil},
```
mitchell852: i believe this should be 1.4...or maybe even 1.5...

@rawlinp @rob05c

docs/source/api/consistenthash.rst
rob05c: Yeah, the regex can contain special characters. It'd have to be encoded, which would be awkward and difficult to use.

I'd vote keeping it a POST. It feels odd, but I think it's a pretty standard technique for APIs, for endpoints that require data with special characters or large amounts of data.

traffic_ops/traffic_ops_golang/routes.go
rob05c: Yeah, Release 3.0.0 has 1.4, so this needs to be 1.5.

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,117 @@
...
+	"encoding/json"
+	"errors"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strconv"
+)
+
+type TRConsistentHashResult struct {
```
ocket8888: This package and the types within it could use some comments describing what they are and what they do. Could be as simple as naming the endpoint that a `func` serves, or the concept that a `type` represents. All Go code hosted on GitHub.com is documented using [godoc](https://blog.golang.org/godoc-documenting-go-code), so we should try to make it helpful (there are also some limited formatting options you can read about in the docs)

docs/source/api/consistenthash.rst
rivasj: I started with a `GET` but I had issues encoding the URI components so I went to a `POST`. It's not a typical `GET` either, but I can try to change it if that is the consensus.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservicesv13.go
```diff
@@ -959,6 +959,7 @@ ds.ccr_dns_ttl,
 ds.cdn_id,
 cdn.name as cdnName,
 ds.check_path,
+ds.consistent_hash_regex,
```
ocket8888: I'm not totally sure we can add fields to API v1.3

docs/source/api/deliveryservices.rst
```diff
@@ -176,12 +176,13 @@ Response Structure
 
 	.. warning:: This number will not be correct if keys are manually replaced using the API, as the key generation API does not increment it!
 
-:tenantId:          The integral, unique identifier of the tenant who owns this Delivery Service
-:trRequestHeaders:  If defined, this takes the form of a string of HTTP headers to be included in Traffic Router access logs for requests - it's a template where ``__RETURN__`` translates to a carriage return and line feed (``\r\n``)\ [2]_
-:trResponseHeaders: If defined, this takes the form of a string of HTTP headers to be included in Traffic Router responses - it's a template where ``__RETURN__`` translates to a carriage return and line feed (``\r\n``)\ [2]_
-:type:              The name of the routing type of this Delivery Service e.g. "HTTP"
-:typeId:            The integral, unique identifier of the routing type of this Delivery Service
-:xmlId:             A unique string that describes this Delivery Service - exists for legacy reasons
+:consistentHashRegex: If defined, this is a regex used for the pattern based consistent hashing feature. It is only applicable for HTTP and Steering Delivery Services
```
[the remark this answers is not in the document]
ocket8888: These also need `versionadded` directives to indicate in what API version they were added - presumably:

```rst
.. versionadded:: 1.4
```

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -655,6 +655,19 @@
...
+                        <span uib-popover-html="label('consistentHashRegex', 'desc')" popover-trigger="click" popover-placement="top" popover-append-to-body="true" popover-class="popover-class">{{label('consistentHashRegex', 'title')}}</span>
+                    </label>
+                    <div class="col-md-10 col-sm-10 col-xs-12">
+                        <input name="consistentHashRegex" type="text" class="form-control" ng-model="deliveryService.consistentHashRegex" ng-maxlength="1024" autofocus>
+                        <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.consistentHashRegex, 'maxlength')">Too Long</small>
+                        <small class="input-diff" ng-show="settings.isRequest && open() && deliveryService.consistentHashRegex != dsCurrent.consistentHashRegex">Current Value: [ {{dsCurrent.consistentHashRegex}} ]</small>
+                        <small><a ng-click="encodeRegex(deliveryService.consistentHashRegex)" href="/#!/delivery-services/{{deliveryService.id}}/consistent-hash?pattern={{encodedRegex}}" target="_blank">Test Regex&nbsp;&nbsp;<i class="fa fs-xs fa-external-link"></i></a></small>
```
ocket8888: This honestly probably isn't a problem, but for safety's sake I've adopted the habit of setting `referrerpolicy="no-referrer"` and `rel="noreferrer"` on `<a>` tags, especially when information is encoded in the URL as is the case with our DS forms.

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
```diff
@@ -0,0 +1,60 @@
...
+            <li><a ng-click="navigateToPath('/delivery-services/' + deliveryService.id + '?type=' + deliveryService.type)">{{deliveryService.xmlId}}</a></li>
+            <li class="active">Consistent Hash Regex Test Tool</li>
+        </ol>
+        <div class="clearfix"></div>
+    </div>
+    <div class="x_content">
+        <br>
+        <form name="dsConsistentHashRegexForm" class="form-horizontal form-label-left" novalidate>
+            <div class="form-group" ng-class="{'has-error': hasError(dsConsistentHashRegexForm.pattern), 'has-feedback': hasError(dsConsistentHashRegexForm.pattern)}">
+                <label class="control-label col-md-2 col-sm-2 col-xs-12">Pattern *</label>
```
ocket8888: labels should have a `for` attribute naming the thing they're labeling, e.g. `for="pattern"`.

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
```diff
@@ -0,0 +1,60 @@
...
+                    <small class="input-error" ng-show="hasPropertyError(dsConsistentHashRegexForm.requestPath, 'pattern')">Path Must Start With '/'</small>
+                    <small class="input-error" ng-show="hasPropertyError(dsConsistentHashRegexForm.requestPath, 'maxlength')">Too Long - Request Path length is restricted by test tool</small>
+                    <span ng-show="hasError(dsConsistentHashRegexForm.requestPath)" class="form-control-feedback"><i class="fa fa-times"></i></span>
+                </div>
+            </div>
+
+            <div class="form-group">
+                <label class="control-label col-md-2 col-sm-2 col-xs-12">Resulting Path</label>
+                <div class="col-md-10 col-sm-10 col-xs-12">
+                    <input name="resultingPath" type="text" ng-disabled="true" class="form-control" ng-model="resultingPath" autofocus>
```
ocket8888: consider setting `disabled` instead of `ng-disabled` when the latter would be constant. It gives the front-end compiler (:nauseated_face:) one less thing to do.

docs/source/api/consistenthash.rst
ocket8888: I actually like POST for this - associating POST with object creation is a pitfall of working with the TO API imo. What it really means is "process this data", and what this does is ask the server to perform operations on the data, and not change server state in any way. The results of the request are the results of those operations, not just data stored on the server. POST seems appropriate.

traffic_ops/traffic_ops_golang/routes.go
ocket8888: In that case, the DS structs need to be re-done - iirc this PR would add a field to them starting at 1.3

docs/source/admin/traffic_router.rst
```diff
@@ -595,3 +595,35 @@ set in ``/lib/systemd/system/traffic_router.service``.
   	-XX:+UseG1GC \
   	-XX:+UnlockExperimentalVMOptions \
   	-XX:InitiatingHeapOccupancyPercent=30"
+
+Pattern Based Consistent Hashing Feature
+========================================
+
+Overview
+--------
+Pattern based consistent hashing is a feature to modify the request path given to Traffic Router's consistent hasher for Cache selection (and Delivery Service selection for Steering Delivery Services) using a regex. This new regex field 'Consistent Hash Regex' is applied on a per-Delivery Service basis and is given to Traffic Router via the CRConfig. The purpose of this feature is to increase cache efficiency by directing requests for the same asset, but with varying request paths, to the same cache.
```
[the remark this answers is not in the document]
rivasj: Do you suggest I do a replace for the entire file, break precedence with PBCH, or leave as `CRConfig` for my doc updates to match the rest of the file?

rivasj:
@ocket8888 addressing your feedback now... can you verify the functionality of the feature while you're reviewing?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3211/
Test PASSed.

ocket8888:
sure. Probably tomorrow, my day's almost over

traffic_ops/traffic_ops_golang/routes.go
```diff
@@ -407,6 +408,9 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `deliveryservices/{id}/urlkeys/?(\.json)?$`, deliveryservice.GetURLKeysByID, auth.PrivLevelReadOnly, Authenticated, nil},
 		{1.1, http.MethodGet, `riak/bucket/{bucket}/key/{key}/values/?(\.json)?$`, apiriak.GetBucketKey, auth.PrivLevelAdmin, Authenticated, nil},
 
+		//Pattern based consistent hashing endpoint
+		{1.5, http.MethodPost, `consistenthash/?$`, consistenthash.Post, auth.PrivLevelReadOnly, Authenticated, nil},
```
mitchell852: if you do this make sure you change a couple of things:

1. https://github.com/apache/trafficcontrol/blob/master/traffic_portal/app/src/scripts/config.js#L26
2. add a section for 1.5 in https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/TrafficOpsRoutes.pm#L61

traffic_portal/app/src/common/api/DeliveryServiceConsistentHashRegexService.js
```diff
@@ -0,0 +1,43 @@
...
+
+        var url = ENV.api['root'] + "consistenthash",
+            params = {regex: regex, requestpath: requestPath, cdnid: cdnId};
+
+        var deferred = $q.defer();
+        return $http.post(url, params)
+            .then(
+                function(result) {
+                    deferred.resolve(result);
+                    return result.data;
```
mitchell852: i don't think this return statement does anything. instead pass result.data into the resolve.

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/FormDeliveryServiceConsistentHashRegexController.js
```diff
@@ -0,0 +1,45 @@
...
+
+	$scope.navigateToPath = locationUtils.navigateToPath;
+
+	$scope.hasError = formUtils.hasError;
+
+	$scope.hasPropertyError = formUtils.hasPropertyError;
+
+    $scope.test = function(pattern, requestPath, cdnId) {
+        deliveryServiceConsistentHashRegexService.getConsistentHashResult(pattern, requestPath, cdnId).then(function (response) {
+            if(response != null && response.response != null) {
```
mitchell852: if you add 2 functions to this `then` you can handle a success and a failure. if you passed result.data to your resolve, then you should always have a response.

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
```diff
@@ -0,0 +1,60 @@
...
+                </div>
+            </div>
+
+            <div class="modal-footer">
+                <button type="button" class="btn btn-success" ng-disabled="pattern == null || requestPath == null" ng-click="test(pattern, requestPath, deliveryService.cdnId)">Test</button>
```
mitchell852: you can simplify this by marking these fields as required and then doing something like ng-disabled={{form.$invalid}} or something like that. that's how the other forms do it.

docs/source/admin/traffic_router.rst
```diff
@@ -601,16 +601,16 @@ Pattern Based Consistent Hashing Feature
...
 
 .. Note:: Pattern Based Consistent Hashing is only available for HTTP and Steering Delivery Services
 
 How it Works
 ------------
 
-With Pattern Based Consistent Hashing, a regex (Consistent Hash Regex) is applied to the request path to extract matching elements to build a new string to pass to the consistent hasher. i.e.: using the Consistent Hash Regex ``/.*?(/.*?/).*?(m3u8)`` given the request paths ``/test/path/asset.m3u8`` and ``/other/path/asset.m3u8`` the resulting path to hash will be ``/path/m3u8``
+With Pattern Based Consistent Hashing, a regular expression (Consistent Hash Regex) is applied to the request path to extract matching elements to build a new string to pass to the consistent hasher. i.e.: using the Consistent Hash Regex :regexp:`/.*?(/.*?/).*?(m3u8) given the request paths ``/test/path/asset.m3u8`` and ``/other/path/asset.m3u8`` the resulting path to hash will be ``/path/m3u8``
```
ocket8888: you're missing a closing backtick on that regexp, which is causing compilation errors.

docs/source/admin/traffic_router.rst
```diff
@@ -601,16 +601,16 @@ Pattern Based Consistent Hashing Feature
 
 Overview
 --------
-Pattern based consistent hashing is a feature to modify the request path given to Traffic Router's consistent hasher for Cache selection (and Delivery Service selection for Steering Delivery Services) using a regex. This new regex field 'Consistent Hash Regex' is applied on a per-Delivery Service basis and is given to Traffic Router via the CRConfig. The purpose of this feature is to increase cache efficiency by directing requests for the same asset, but with varying request paths, to the same cache.
+Pattern-Based Consistent Hashing is a feature to modify the request path given to Traffic Router's consistent hasher for Cache selection (and :term:`Delivery Service` selection for Steering Delivery Services) using a regular expression. This new regular expression field 'Consistent Hash Regex' is applied on a per-Delivery Service basis and is given to Traffic Router via the CRConfig. The purpose of this feature is to increase cache efficiency by directing requests for the same asset, but with varying request paths, to the same Cache.
```
ocket8888: It's preferable to always link a term to its definition, but once per section ("page" in an HTML context) is probably enough at a minimum.

traffic_ops/traffic_ops_golang/crconfig/deliveryservice.go
```diff
@@ -308,7 +308,7 @@ and d.active = true
 			}
 		}
 
-		if consistentHashRegex.String != "" {
+		if consistentHashRegex.Valid && consistentHashRegex.String != "" {
```
ocket8888: This is actually better than what I suggested (assuming we don't want to allow empty strings for the consistent hash regular expression), just wanted to point that out.

ocket8888:
Currently, when I make the request directly to the API using version 1.5 I get a `200 OK` response with what (I think?) is the right payload. However, TP is still trying to query it at v1.4, so it's failing.

I also, as a user who really is trying to figure out what this is and how to use it for the first time, would appreciate some help tooltips or instructions on the `/#!//delivery-services/{{Delivery Service ID}}/consistent-hash?pattern={{encoded pattern}}` page.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservicesv13.go
rob05c: We can't. Per Semantic Versioning, which we've agreed to as a project, it needs to be a new minor version. But, right now it's a pain, it's a lot of boilerplate that has to be copy-pasted for a new version.

It's ok for `master` not to be perfect, as long I think I'm inclined to Approve this, with the caveat that we fix it before the next release. I'm willing to commit to fixing it; but, I'm also kind of hoping to propose a better solution.

traffic_ops/traffic_ops_golang/routes.go
rob05c: The structs are easy change, but right now, the endpoint itself needs a lot of copy-paste. It didn't used to be too bad, but it got considerably worse with all the "CRUDer" boilerplate.

The Go language has very limited metaprogramming, which is what we need here. So, there's no ideal solution. But, I'm really hoping to come up with something better than what we have. Or if not, I'm willing to make the new endpoint, rather than requiring Jesse to. As long as we fix it before the next Release, it's ok for `master` to be imperfect between releases.

So, I'm inclined to Approve this, as long as we make a Github issue to fix it before the next release. If you're ok with that @ocket8888 ?

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservicesv13.go
ocket8888: idk how much of a pain this is, I think you'd need to rewrite a bit of the v13 handling code and then totally re-implement it for a different API version (again, no idea how much work that is), but from a struct point of view this should be as easy as `type DeliveryServiceNullable struct {` -> `type DeliveryServiceNullableV13 struct {` and

```go
type DeliveryServiceNullable struct {
    DeliveryServiceNullableV13
    consistent_hash_regex *string `json:"consistentHashRegex,omitempty"`
}
```

traffic_ops/traffic_ops_golang/routes.go
ocket8888: Sure, but it'll mean that something documented as part of v1.3 will later be revised to be part of v1.5 (god I hope the issue resolution doesn't miss that) or we could straight-up lie in the documentation and say it's only available in v1.5, then link to the issue and call the field's appearance in v1.3 and v1.4 a "bug".

ocket8888: To be clear, I'm fine with either, personally. I think the best solution is the latter, since even if the issue resolution misses cleaning that up, people can follow the issue link and see that it actually was resolved.

docs/source/admin/traffic_router.rst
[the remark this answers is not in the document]
ocket8888: "PBCH"?
I think the best solution is to do a replace for the entire file, but that's not technically your responsibility, so if you want to just make the change in your changes specifically, I won't block this PR based solely on that.

docs/source/api/deliveryservices.rst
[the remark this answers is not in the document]
ocket8888: actually this'll be either v1.5 or v1.3 depending on how we want to handle that.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservicesv13.go
rob05c: >from a struct point of view this should be as easy as `type DeliveryServiceNullable struct {` -> `type DeliveryServiceNullableV13 struct {` and

Correct.

>I think you'd need to rewrite a bit of the v13 handling code and then totally re-implement it for a different API version

Also correct. It's less rewriting, and more copy-paste. You basically copy the v12 file, name it v13, and rename v13 to v15. 

>no idea how much work that is

It's not really much work, it'd probably take me 15mins. But it'd take longer for someone who didn't already know how. It's mostly just the copy-paste, which people also (rightly) object to.

This is what I meant on the other comment, about metaprogramming. We really need some form of MP—generics, macros, anything would do—and Go simply doesn't have it. The "idiomatic" solution is `go generate`, which, generating code can be viewed as a form of metaprogramming. But people object to that too, just because it's idiomatic doesn't mean it's good.

Again, I'm planning to propose better solution, and I also know what needs done as-is, so I'm willing to commit to fixing this afterwards.

FWIW I'm also planning to create dev docs, either the existing or new solution, for how to add new versions. Hopefully that will help, too.

ocket8888: It could help to pass the API version into version-independent handlers for each endpoint under `/api/1.x/*`, then adding a new route can be as easy as adding a `case` in a few places for selective inclusion of fields.

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
rivasj: it looks like `for` label attributes are only for referencing an input's `id` but we're using `ng-model` instead

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3218/
Test PASSed.

rivasj:
> Currently, when I make the request directly to the API using version 1.5 I get a `200 OK` response with what (I think?) is the right payload. However, TP is still trying to query it at v1.4, so it's failing.
> 
> I also, as a user who really is trying to figure out what this is and how to use it for the first time, would appreciate some help tooltips or instructions on the `/#!//delivery-services/{{Delivery Service ID}}/consistent-hash?pattern={{encoded pattern}}` page.

@ocket8888 I've pushed some changes so that hitting the TO endpoint from TP should work now. I've addressed much of the other feedback as well -- I'll work on getting some tooltips added on the Test Tool page.

docs/source/api/consistenthash.rst
```diff
@@ -34,17 +34,42 @@ Request Structure
...
 ------------------
 :resultingPathToConsistentHash: The resulting path that Traffic Router will use for consistent hashing
 :consistentHashRegex:           The regex used by Traffic Router derived from POST 'regex' parameter
 :requestPath:                   The request path used by Traffic Router to test regex against
 
-.. code-block:: json
+.. code-block:: http
 	:caption: Response Example
 
+	HTTP/2 200
```
ocket8888: Unfortunately this is causing compilation errors because the Pygments library used for highlighting syntax (like HTTP) does not have support for HTTP/2 (I have [an open issue](https://bitbucket.org/birkenfeld/pygments-main/issues/1486/lexing-support-for-http-2) about it). If you just change `HTTP/2` to `HTTP/1.1` (in both the request and the reply), and add a 'reason' text (i.e. `200 OK` not just `200` - don't ask, I don't know why) the lexer will be happy.

rivasj: okay, made the change

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3219/
Test PASSed.

traffic_ops/traffic_ops_golang/routes.go
mitchell852: Can you add this endpoint to roles/capabilities so it is not lost when we migrate to roles/capabilities?

Off the top of my head probably do this:
1. create a capability called consistenthash-read or consistenthash-generate
2. add the capability to the admin, ops and r/o role
3. add the POST api/consistenthash endpoint to the consistenthash-read capability

you can follow the examples in seeds.sql

traffic_portal/app/src/common/api/DeliveryServiceConsistentHashRegexService.js
rivasj: I removed that return statement, passed result.data into the resolve, and added `return deferred.promise;` after the post

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/FormDeliveryServiceConsistentHashRegexController.js
rivasj: I added a second function and removed the null checking for the success

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3220/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3221/
Test PASSed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -555,6 +559,35 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
 		return routeResult;
 	}
 
+	public String buildPatternBasedHashString(final DeliveryService deliveryService, final String requestPath) {
+		if (deliveryService.getConsistentHashRegex() != null && !deliveryService.getConsistentHashRegex().isEmpty() && !requestPath.isEmpty()) {
+			return buildPatternBasedHashString(deliveryService.getConsistentHashRegex(), requestPath);
+		} else {
```
rawlinp: nit: since the `if` block returns, this `else` block isn't really necessary and could just `return requestPath` outside of the block.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -555,6 +559,35 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+			final Matcher matcher = pattern.matcher(requestPath);
+
+			final StringBuilder sb = new StringBuilder();
+			if (matcher.find() && matcher.groupCount() > 0) {
+				for (int i = 1; i <= matcher.groupCount(); i++) {
+					final String text = matcher.group(i);
+					sb.append(text);
+				}
+				return sb.toString();
+			} else {
```
rawlinp: nit: ditto above comment about the else block

traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/external/ConsistentHashTest.java
```diff
@@ -258,4 +263,172 @@ public void itUsesBypassFiltersWithDeliveryServiceSteering() throws Exception {
...
+			requestPath = URLEncoder.encode("/other/path/other_thing.m3u8", "UTF-8");
+			httpGet = new HttpGet("http://localhost:3333/crs/consistenthash/patternbased/regex?regex=" + encodedConsistentHashRegex + "&requestPath=" + requestPath);
+
+			response = closeableHttpClient.execute(httpGet);
+
+			resp = objectMapper.readTree(EntityUtils.toString(response.getEntity()));
+
+			assertThat(JsonUtils.optString(resp, "resultingPathToConsistentHash"),equalTo(resultingPathToConsistentHash));
+		} finally {
+			if (response != null) response.close();
```
rawlinp: nit/question: I _think_ we can use https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html here since `CloseableHttpResponse` implements `java.io.Closeable`. That might clean these test cases up slightly.

traffic_portal/app/src/modules/private/deliveryServices/consistentHashRegex/index.js
```diff
@@ -0,0 +1,42 @@
...
+	.config(function($stateProvider, $urlRouterProvider) {
+		$stateProvider
+			.state('trafficPortal.private.deliveryServices.consistentHashRegex', {
+				url: '/{deliveryServiceId}/consistent-hash?pattern',
+				views: {
+					deliveryServicesContent: {
+						templateUrl: 'common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html',
+						controller: 'FormDeliveryServiceConsistentHashRegexController',
+                        resolve: {
+                            deliveryService: function($stateParams, deliveryServiceService) {
```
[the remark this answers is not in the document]
ocket8888: all Go source files should be hit with `go fmt`, which will handle consistent indentation

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,117 @@
...
+	if trafficRouter == "" {
+		return nil, errors.New("no eligible Traffic Router found for pattern based consistent hashing with cdn Id: " + strconv.FormatInt(cdnId, 10))
+	}
+	if apiPort == "" {
+		return nil, errors.New("no parameter 'api.port' found for pattern based consistent hashing with cdn Id: " + strconv.FormatInt(cdnId, 10))
+	}
+
+	trafficRouterAPI := "http://" + trafficRouter + ":" + apiPort + "/crs/consistenthash/patternbased/regex?regex=" + url.QueryEscape(regex) + "&requestPath=" + url.QueryEscape(requestPath)
+
+	r, err := http.Get(trafficRouterAPI)
```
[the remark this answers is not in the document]
rivasj: would the proxy_timeout from cdn.conf make sense here?
`Timeout:   time.Duration(d.Config.ProxyTimeout) * time.Second,`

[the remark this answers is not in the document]
rawlinp: hmmm...maybe we should create a `const RouterRequestTimeout = time.Second * 10` similar to the `const MonitorRequestTimeout = time.Second * 10` in cachesstats/cachesstats.go. 10 seconds should be a reasonable time for a TR API request to complete, right? Probably don't need it to be configurable either.

traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/external/ConsistentHashTest.java
rivasj: so I tried adjusting a test case to use:
`try (CloseableHttpResponse response = closeableHttpClient.execute(httpGet)){ 
...
}`
but it looks like a resource declared with a try statement is final, and that's not ideal here since I'm reusing the `CloseableHttpResponse` object for multiple requests in each test case

rivasj: I could add a try-with-resource statement for each request, but I think that's going the opposite way in terms of cleanliness

rawlinp: yeah, thanks for trying it out. If it makes it messier we shouldn't do it.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3224/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3225/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3230/
Test PASSed.

docs/source/api/consistenthash.rst
```diff
@@ -0,0 +1,75 @@
...
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+.. _to-api-consistenthash:
+
+******************
+``consistenthash``
+******************
+Test Pattern-Based Consistent Hashing for a Delivery Service using a regular expression and a request path
```
ocket8888: "Delivery Service" should be a `:term:` at least once per major section ("page" in an HTML context)

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -471,3 +471,65 @@ Request Structure
...
+	| requestPath       | yes      | The (URI encoded) request path to use to test pattern based consistent hashing                               |
+	+-------------------+----------+--------------------------------------------------------------------------------------------------------------+
+	| deliveryServiceId | yes      | The integral, unique identifier?/'xml_id'?/name? of a :term:`Delivery Service` served by this Traffic Router |
+	+-------------------+----------+--------------------------------------------------------------------------------------------------------------+
+	| ip                | yes      | The IP address of a potential client                                                                         |
+	+-------------------+----------+--------------------------------------------------------------------------------------------------------------+
+
+Response Structure
+------------------
+TBD
```
ocket8888: "TBD" in not an acceptable response structure (or response example). It should minimally contain an example - though the routes that omit structure explanations only do so because I don't understand them.

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,131 @@
...
+	api.WriteResp(w, r, consistentHashResult)
+}
+
+const RouterRequestTimeout = time.Second * 10
+
+// queries database for active Traffic Router on the CDN specified by cdnId
+// passes regex and requestPath to the Traffic Router via API request,
+// and returns the response
+func getPatternBasedConsistentHash(tx *sql.Tx, regex string, requestPath string, cdnId int64) ([]byte, error) {
+	q := `
```
ocket8888: I'm not clear based on existing code what the procedure is for things like this, but could this be made into a `const`? The compiler might be smart enough to handle it, but allocating this space on a per-call basis may be a waste of cycles and threaded memory.

docs/source/development/traffic_router/traffic_router_api.rst
ocket8888: "TBD" is not an acceptable response structure. It should minimally have an example - though the only reason any of these don't have field explanations is because I don't understand them (yet), and since you know exactly what they are what they mean, you should document it so others can know too.

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
ocket8888: I'm not sure based on the examples set forth by extant code, but shouldn't this be a `const`? Seems like allocating space for this on a per-call basis might be a waste of threaded stack memory and CPU cycles (though maybe the compiler is smart enough to make that optimization anyway?).

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,131 @@
...
+	r, err := trClient.Get(trafficRouterAPI)
+
+	if err != nil {
+		return nil, errors.New("Error creating request to Traffic Router: " + err.Error())
+	}
+	defer r.Body.Close()
+
+	body, err := ioutil.ReadAll(r.Body)
+
+	return body, nil
```
ocket8888: This is shadowing any errors that might occur while reading the `Body`. Should probably be `return body, err`, but another `if` like you have above could also work, if you think that makes more sense.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -655,6 +655,19 @@
                     </div>
                 </div>
 
+                <div class="form-group" ng-class="{'has-error': hasError(deliveryServiceForm.consistentHashRegex), 'has-feedback': hasError(deliveryServiceForm.consistentHashRegex)}">
+                    <label class="control-label col-md-2 col-sm-2 col-xs-12">
```
ocket8888: A label should have the `for` attribute that tells what it's labeling (valued as the `name` attribute of the `<input>` to which it refers)

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -655,6 +655,19 @@
...
+                        <span uib-popover-html="label('consistentHashRegex', 'desc')" popover-trigger="click" popover-placement="top" popover-append-to-body="true" popover-class="popover-class">{{label('consistentHashRegex', 'title')}}</span>
+                    </label>
+                    <div class="col-md-10 col-sm-10 col-xs-12">
+                        <input name="consistentHashRegex" type="text" class="form-control" ng-model="deliveryService.consistentHashRegex" ng-maxlength="1024" autofocus>
+                        <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.consistentHashRegex, 'maxlength')">Too Long</small>
+                        <small class="input-diff" ng-show="settings.isRequest && open() && deliveryService.consistentHashRegex != dsCurrent.consistentHashRegex">Current Value: [ {{dsCurrent.consistentHashRegex}} ]</small>
```
ocket8888: I know you're just copying existing form control markup, but imo this `ng-show` is a little wacky. The thing will never be displayed if this isn't a ds *Request*, which is a known constant on pageload. So imo it should actually be more like:
```html
<small class="input-diff" ng-if="settings.isRequest" ng-show="open() && deliveryService.consistentHashRegex !== dsCurrent.consistentHashRegex>Current Value: [ {{dsCurrent.consistentHashRegex}} ]</small>
```
... it's all about reducing the work done by that **front-end compiler** (:nauseated_face:). Also note the `!==` instead of `!=`, to account for the fact that `==` does type coercion, such that e.g. `0 == "0"` is `true` but `0 === "0"` is `false`. Things get really crazy when you consider more complex objects. We have an ESLint configuration file that would probably warn about this, but I'm not sure if it checks JS expressions in angular attributes.

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
ocket8888: Again, labels should have a `for`.

docs/source/admin/traffic_router.rst
```diff
@@ -595,3 +595,35 @@ set in ``/lib/systemd/system/traffic_router.service``.
   	-XX:+UseG1GC \
   	-XX:+UnlockExperimentalVMOptions \
   	-XX:InitiatingHeapOccupancyPercent=30"
+
+Pattern-Based Consistent Hashing Feature
+========================================
+
+Overview
+--------
+Pattern-Based Consistent Hashing is a feature to modify the request path given to Traffic Router's consistent hasher for Cache selection (and :term:`Delivery Service` selection for Steering Delivery Services) using a regular expression. This new regular expression field 'Consistent Hash Regex' is applied on a per-Delivery Service basis and is given to Traffic Router via the CDN Snapshot. The purpose of this feature is to increase cache efficiency by directing requests for the same asset, but with varying request paths, to the same Cache.
```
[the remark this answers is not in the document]
rivasj: okay, I'll update it

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
rivasj: this follows the existing procedure for database queries. also, this isn't an endpoint that will be getting much traffic.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
rivasj: see my previous comment about using the `for` attribute when using `ng-model` (it references the `id` attribute of an `<input>`, which we don't use)

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
rivasj: good point, I'll add an error check

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
rivasj: I think that this should be left as is to match the rest of the TP files. you can create an issue if you want it to be addressed in the future

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
rawlinp: Yeah we do this same thing all over the place in traffic_ops_golang. Also, premature optimization is the root of all evil.

docs/source/development/traffic_router/traffic_router_api.rst
ocket8888: accidental duplicate :P

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
rawlinp: +1 on keeping it consistent w/ the rest of the TP files. Fixing all of that would not be in the spirit of this PR.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
ocket8888: `ng-model` assigns the value of a control to a property of its Javascript controller object. `for` is a pure markup attribute that should use the `name` attribute (*not* the `id`) of the control it labels. Otherwise, things that aren't a human brain with working eyes can't tell what it's labeling.

ocket8888: Well, actually, it _can_ use an `id` instead of a `name`, but that's pretty much just for when the same `name` exists on multiple controls.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
ocket8888: fair enough

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
ocket8888: I'm fine leaving it the way it is, but on that note: why are things this way? Is there an advantage to e.g. creating a `func` that does nothing but return a static, constant string?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3232/
Test PASSed.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
ocket8888: Nevermind, I'm very, very wrong about this.

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
ocket8888: disregard as above

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
rob05c: Readability. If it really is a constant string, there's no reason to do a `func` instead of a `const`. Unless maybe ever endpoint does `func selectQuery`, and then the consistency is easier to read. But some things can't be `const` in Go (e.g. `map`). Then, a `func` makes it impossible to modify, and is safer than a global `var`.

Here, it's arguably easier to read:
```
qry := ` 
// seven lines 
`
tx.Exec(qry, cdnID)
```
Than: ```
tx.Exec(`
// seven lines
, cdnID)
```

>allocating this space on a per-call basis may be a waste of cycles and threaded memory.

The compiler is smarter than you and me. Are you certain it's faster to reference a distant heap object every usage, than to copy the memory onto the stack and reference local memory? I'm not. 

Even the compiler isn't smart enough, things should only be optimized when necessary; here it's not. Readability is more important.

"Programs must be written for people to read, and only incidentally for machines to execute." —Abelson, SICP

File: docs/source/development/traffic_router/traffic_router_api.rst (hunk 24 above)
Comment (rivasj) [reply docs/source/development/traffic_router/traffic_router_api.rst]: I've included request/response structures for 2/3 of the routes I've added. I don't currently have the environment to get a successful response for the third (healthy caches), but I can add it at a later time

**Comment by asfgit** [comment]
https://builds.apache.org/job/trafficcontrol-PR/3233/
Test PASSed.

File: traffic_ops/app/db/seeds.sql [hunk 30]
```diff
@@ -317,6 +321,7 @@ INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHER
 INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHERE name = 'operations'), 'cdns-read' WHERE EXISTS (SELECT id FROM role WHERE name = 'operations') ON CONFLICT DO NOTHING;
 INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHERE name = 'operations'), 'cdn-security-keys-read' WHERE EXISTS (SELECT id FROM role WHERE name = 'operations') ON CONFLICT DO NOTHING;
 INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHERE name = 'operations'), 'change-logs-read' WHERE EXISTS (SELECT id FROM role WHERE name = 'operations') ON CONFLICT DO NOTHING;
+INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHERE name = 'operations'), 'consistenthash-read' WHERE EXISTS (SELECT id FROM role WHERE name = 'operations') ON CONFLICT DO NOTHING;
```
mitchell852: oh i forgot. there's one more place you should add this capability and that's in the patches.sql file for the 'portal' role

traffic_portal/app/src/common/api/index.js
```diff
@@ -27,6 +27,7 @@ module.exports = angular.module('trafficPortal.api', [])
     .service('cdnService', require('./CDNService'))
     .service('changeLogService', require('./ChangeLogService'))
     .service('coordinateService', require('./CoordinateService'))
+    .service('deliveryServiceConsistentHashRegexService', require('./DeliveryServiceConsistentHashRegexService'))
```
mitchell852: what do you think about just putting this in the deliveryServiceService? and getting rid of deliveryServiceConsistentHashRegexService.

rivasj: that might make more sense because the consistent hash regex is related to a Delivery Service and but it's not its own object... let me shuffle some code around and see how it works

traffic_portal/app/src/traffic_portal_properties.json
```diff
@@ -189,6 +189,10 @@
             "title": "Check Path",
             "desc": "A path (ex: /crossdomain.xml) to verify the connection to the origin server with. This can be used by Check Extension scripts to do periodic health checks against the delivery service."
           },
+          "consistentHashRegex": {
+            "title": "Consistent Hash Regex",
+            "desc": "Regex used to extract parts of the request path using grouping in order to build a consistent request string to be used for cache selection. If this field is set on a steering delivery service, it will override the regex set on the target delivery services."
```
mitchell852: is there a docs link you can link to kinda like the deepCachingType?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3234/
Test PASSed.

traffic_portal/app/src/traffic_portal_properties.json
rivasj: yeah, I can add that

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3235/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3236/
Test PASSed.

mitchell852:
image

imo that should probably pop a modal dialog rather than opening a new tab/window but maybe that's something that can be revisited in the future.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3242/
Test PASSed.

rivasj:
@mitchell852 barring the modal refactor, does everything else look good with the front-end changes?

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
```diff
@@ -0,0 +1,60 @@
...
+                    <small class="input-error" ng-show="hasPropertyError(dsConsistentHashRegexForm.requestPath, 'pattern')">Path Must Start With '/'</small>
+                    <small class="input-error" ng-show="hasPropertyError(dsConsistentHashRegexForm.requestPath, 'maxlength')">Too Long - Request Path length is restricted by test tool</small>
+                    <span ng-show="hasError(dsConsistentHashRegexForm.requestPath)" class="form-control-feedback"><i class="fa fa-times"></i></span>
+                </div>
+            </div>
+
+            <div class="form-group">
+                <label class="control-label col-md-2 col-sm-2 col-xs-12">Resulting Path</label>
+                <div class="col-md-10 col-sm-10 col-xs-12">
+                    <input name="resultingPath" type="text" disabled class="form-control" ng-model="resultingPath" autofocus>
```
ocket8888: Rather than a read-only input, I think [`<output>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output) is more appropriate. If you don't change the name `name` or `class` attributes, then all you do is get rid of `type`, `ng-model` and `autofocus` and make an ng-bind to the desired scope variable. Basically, this:

```html
<output name="resultingPath" class="form-control">{{resultingPath}}</output>
```

The MDN page to which I linked also has an example on how to do it using the HTMLFormAPI, but I think implementing that would be a big, unnecessary change.

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,134 @@
...
+    JOIN profile_parameter ON ((profile_parameter.profile = profile.id)))
+    JOIN parameter ON ((parameter.id = profile_parameter.parameter)))
+    JOIN status ON ((server.status = status.id))
+ WHERE ((server.type = (select id from type where name = 'CCR')) AND
+ (parameter.name = 'api.port'::text) AND
+ (status.name = 'ONLINE') AND
+ (server.cdn_id = $1))
+`
+	rows, err := tx.Query(q, cdnId)
+	defer rows.Close()
```
rawlinp: I think the `defer` has to come after the `if err != nil` check.

rob05c: Good catch. Even if it's technically permissible, it's always safer to treat all other values as undefined, if an err is non-nil

docs/source/api/consistenthash.rst
```diff
@@ -0,0 +1,75 @@
...
+
+:Auth. Required: Yes
+:Roles Required: None
+:Response Type:  Object
+
+Request Structure
+-----------------
+:regex:       The regex to apply to the request path to get a resulting path that will be used for consistent hashing
+:requestpath: The request path to use to test the regular expression against
+:cdnid:       The unique identifier of a CDN that will be used to query for an active Traffic Router
```
rawlinp: I think these two fields are `requestPath` and `cdnId` now, right?

docs/source/api/consistenthash.rst
```diff
@@ -0,0 +1,75 @@
...
+
+	POST /api/1.5/consistenthash HTTP/1.1
+	Host: trafficops.infra.ciab.test
+	User-Agent: curl/7.54.0
+	Accept: */*
+	Cookie: mojolicious=...
+	Content-Length: 80
+	Content-Type: application/x-www-form-urlencoded
+
+	{"regex":"/.*?(/.*?/).*?(m3u8)","requestpath":"/test/path/asset.m3u8","cdnid":2}
```
rawlinp: I think these two fields are `requestPath` and `cdnId` now, right?

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
rawlinp: actually, since we really only need one random traffic router to try, I think we should add this to the query:
```
ORDER BY RANDOM()
LIMIT 1
```
and use `tx.QueryRow()` instead of `tx.Query()` like:
```
if err := tx.QueryRow(q, cdnId).Scan(&trafficRouter); err != nil {
...
}
```

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
```diff
@@ -0,0 +1,134 @@
...
+
+	trClient := &http.Client{
+		Timeout: RouterRequestTimeout,
+	}
+	r, err := trClient.Get(trafficRouterAPI)
+
+	if err != nil {
+		return nil, errors.New("Error creating request to Traffic Router: " + err.Error())
+	}
+	defer r.Body.Close()
```
rawlinp: We should also check to make sure the TR response is a 200, otherwise return an error like "Traffic Router returned a 404". When I was testing this locally it was making the request to a non-upgraded TR that didn't support the endpoint yet, and in Traffic Portal it wasn't showing an error.

traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
rivasj: using `tx.QueryRow()` makes the code much cleaner

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
```diff
@@ -147,6 +148,7 @@ public DeliveryService(final String id, final JsonNode dsJo) throws JsonUtilsExc
 		}
 		sslEnabled = JsonUtils.optBoolean(dsJo, "sslEnabled");
 		this.anonymousIpEnabled = JsonUtils.optBoolean(dsJo, "anonymousBlockingEnabled");
+		this.consistentHashRegex = JsonUtils.optString(dsJo, "consistentHashRegex");
```
rawlinp: I'm thinking we should probably compile the regex here so that it doesn't need to be compiled on every request. It's probably negligible in terms of the total amount of time that a single request makes, but it's just unnecessary work to do on every request that comes in for this DS.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3277/
Test PASSed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
rivasj: So it's pretty convenient to be able to pass a string around as the regex and let the `buildPatternBasedHashString(..)` method do the regex compilation and try/catch handling, especially for the code paths that don't use a delivery service (unit tests and API endpoints). Do you think that the benefit of your suggested approach outweighs the convenience of what is currently in place?

rawlinp: I don't feel too strongly about it. It's hard to make any judgements about performance without doing any benchmarking or load testing, but just timing a simple regex compilation locally took between 0.05 - 0.1 ms. That seems pretty negligible to me, but if we start caring about fractions of a millisecond in `ttms` this is something we could optimize later down the line.

ocket8888:
b-but my requested changes

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3278/
Test PASSed.

rawlinp:
@ocket8888 sorry, I thought all requested changes had been addressed at this point. If it's just tooltips you're after, that can be addressed in a quick follow-up PR.

ocket8888:
No, he put in the tooltip. I was referring to the most recent changes I requested (which are maybe even simpler, to be fair). The output of the regex testing procedure is being represented as a disabled `<input>` node, and I suggested it use the standard `<output>` element instead.

PR #61, opened 2016-11-07
TC-34 - Traffic Ops can now return ssl certificate information for DSs with periods in the xml_id

dneuman64:
Traffic Ops now returns ssl certificate information for delivery services with periods in the name.  This caused an issue in update_riak_for_search.pl where  the ssl keys for a delivery service were not indexed because the keys could not be gotten from riak.

This fixes TC-34.

PR #3287, opened 2019-01-31
CIAB: Fix Traffic Vault Enrollment

Shihta:
#### What does this PR do?

Since we need to follow the services isolation policy, it should not set dependency in docker-compose.yml.
So if TO-perl run faster than TV, the script `set-to-ips-from-dns.sh` can not obtain TV's IP and supplly it to `040-trafficvault_server.json`

This commit fix TV enrollment so TV can register by itself.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Other CIAB

This is a bug fix, so just simply start CIAB and then check if `http://video.demo1.mycdn.siab.test` is available.

#### Check all that apply

- [ ] This PR includes tests


asfgit:

infrastructure/cdn-in-a-box/traffic_ops/to-access.sh
```diff
@@ -168,85 +172,68 @@ to-enroll() {
...
+	fi
+	if [[ ! -z "$4" ]]; then
+		export MY_TCP_PORT="$4"
+	else
+		export MY_TCP_PORT="80"
+	fi
+	if [[ ! -z "$5" ]]; then
+		export MY_HTTP_PORT="$5"
+	else
+		export MY_HTTP_PORT="443"
```
ocket8888: Shouldn't this be e.g. `MY_HTTPS_PORT`? The field it's going into is `httpsPort`, not `httpPort`. It's confusing that we have `tcpPort` and `httpsPort` instead of `httpPort` or `tlsPort`, but we should at least be consistently confusing in the same way :P

infrastructure/cdn-in-a-box/traffic_ops/to-access.sh
```diff
@@ -168,85 +172,68 @@ to-enroll() {
...
-			else
-				export MY_CACHE_GROUP="CDN_in_a_Box_Edge"
-			fi
 			;;
 		"mid" )
 			export MY_TYPE="MID"
 			export MY_PROFILE="ATS_MID_TIER_CACHE"
 			export MY_STATUS="REPORTED"
-			if [[ ! -z "$3" ]]; then
-				export MY_CACHE_GROUP="$3"
```
ocket8888: I think this will break. The mid's `run.sh` is calling `to-enroll` without specifying a Cache Group. Haven't looked at any of the others, but I suspect they do the same thing - which in their case is fine because it by and large won't matter which Cache Group they are placed into.

Shihta: I don't quite understand what you mean
The `grep to-enroll` results are shown as follows:

```
edge/run.sh:61:to-enroll edge $CDN || (while true; do echo "enroll failed."; sleep 3 ; done)
mid/run.sh:61:to-enroll mid $CDN || (while true; do echo "enroll failed."; sleep 3 ; done)
traffic_monitor/run.sh:63:to-enroll tm $CDN || (while true; do echo "enroll failed."; sleep 3 ; done)
traffic_ops/run.sh:130:to-enroll "to" ALL || (while true; do echo "enroll failed."; sleep 3 ; done)
traffic_ops/run-go.sh:79:to-enroll "to" ALL || (while true; do echo "enroll failed."; sleep 3 ; done)
traffic_portal/run.sh:70:to-enroll "tp" ALL || (while true; do echo "enroll failed."; sleep 3 ; done)
traffic_router/run.sh:100:to-enroll tr || (while true; do echo "enroll failed."; sleep 3 ; done)
```

Except TR, all another services have specified their cache group.
Take mid as an example, it has specified its cache group as `CDN-in-a-Box`.
Please refer to [here](https://github.com/apache/trafficcontrol/blob/master/infrastructure/cdn-in-a-box/mid/run.sh#L48-L61) for details: 

On the other hand, what is the function of cache group?
Could you share the document link with me?

Shihta:
@ocket8888 Since I did not get your reply, I have to fix the problem of HTTPS only.
On the other hand, this commit only aim to fix the TV enrollment.
I think we can make another issue to discuss your opinion.

infrastructure/cdn-in-a-box/traffic_ops/to-access.sh
ocket8888: no, `$CDN` and `ALL` are the names of CDNs, not Cache Groups. Those are all relying on the defaults

Shihta: @ocket8888 I see what you mean. I'm sorry to have kept you waiting so long.

As the above search results, there is no any service supply more than 2 parameters.
It means current design wants to put them all in the same cache group, even that's not a good idea.

This commit add an *empty string* for the third prameter when TV calls to-enroll, in [here](https://github.com/apache/trafficcontrol/pull/3287/commits/2df03f70eba694a671b0046648d621e0da98566f#diff-5e6d2212372d15ad9c4e202c58843b85R29)
So this commit will not break anything because it keep the original architecture and give default value to each new parameter.

How about let me raise a new issue for it first?
Maybe I can fix it after I know more about the cache group.

ocket8888: You can read more about Cache Groups [in their section of our official documentation](https://traffic-control-cdn.readthedocs.io/en/latest/faq/administration.html#cachegroup) (though that'll move soon if I get my PR merged).

In the meantime, lemme take another look at this PR.

ocket8888: Yeah, the problem still exists. You can see the issue in `infrastructure/cdn-in-a-box/mid/run.sh` on line #61 where it's calling `to-enroll`:

```bash
to-enroll mid $CDN || (while true; do echo "enroll failed."; sleep 3 ; done)
```

This means that when `to-enroll` is called, it sees its positional parameters defined as:

```bash
1="mid"
2="$CDN"
3=""
4=""
5=""
# etc.
```

so with your change the Cache Group to which the mid will be assigned will be set to `"CDN_in_a_Box_Edge"`, which is not correct. The Mid-tier cache server needs to be in a separate Cache Group from the Edge-tier cache server, so that the the latter can assign the former as its parent.

Shihta: OMG I found what's my problem! It's all my fault!
I did not see clearly that only mid cache group is set to "CDN_in_a_Box_**Mid**" and all the other are set to "CDN_in_a_Box_**Edge**"!
I'll fix it right away!

Shihta:
@ocket8888 does this PR ok to test?

ocket8888:
I actually don't know if lowly contributors have that power. Let's find out.

ocket8888:
ok to test

infrastructure/cdn-in-a-box/traffic_vault/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN rm -rfv /etc/riak/prestart.d/* /etc/riak/poststart.d/*
 RUN echo 'APT::Install-Recommends 0;' >> /etc/apt/apt.conf.d/01norecommends \
  && echo 'APT::Install-Suggests 0;' >> /etc/apt/apt.conf.d/01norecommends \
  && apt-get update \
- && DEBIAN_FRONTEND=noninteractive apt-get install -y vim.tiny wget net-tools sudo net-tools ca-certificates unzip dnsutils \
+ && DEBIAN_FRONTEND=noninteractive apt-get install -y vim.tiny wget net-tools sudo net-tools ca-certificates unzip dnsutils gettext-base \
```
ocket8888: I know this isn't something you did, but as long as we're touching this line, can you remove `vim.tiny`, `wget`, `sudo` and the extra `net-tools`? That container has no real use for those, and it's making the thing bigger than it needs to be.

Shihta: ok, I'll remove them

rob05c:
ok to test

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3226/
Test PASSed.

Shihta:
@ocket8888 since this PR has passed the test, can you approve this it?

Shihta:
@rob05c can you help to merge it?

PR #3296, opened 2019-02-01
Change TO cdns/dnsseckeys to abstract versioning

rob05c:
TechDebt. Changes the cdns/dnsseckeys api/1.1 and api/1.4+ endpoints
to use the same helper, so the 1.1 func is doing nothing but handling
and casting the latest obj to v11.

#### What does this PR do?

Fixes #(issue_number)

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/lib/go-tc/deliveryservice_ssl_keys.go
+++ b/lib/go-tc/deliveryservice_ssl_keys.go
@@ -224,6 +224,25 @@ type DNSSECKeyDSRecordV11 struct {
 	Digest     string `json:"digest"`
 }
 
+func (keys *DNSSECKeys) ToV11() DNSSECKeysV11 {
+	v11 := DNSSECKeysV11{}
+	for keyName, keySet := range *keys {
+		v11[keyName] = keySet.ToV11()
+	}
+	return v11
+}
+
+func (set *DNSSECKeySet) ToV11() DNSSECKeySetV11 {
+	v11 := DNSSECKeySetV11{}
+	for _, zsk := range set.ZSK {
+		v11.ZSK = append(v11.ZSK, zsk.DNSSECKeyV11)
+	}
+	for _, ksk := range set.KSK {
+		v11.KSK = append(v11.KSK, ksk.DNSSECKeyV11)
+	}
+	return v11
+}
+
 // CDNDNSSECGenerateReqDate is the date accepted by CDNDNSSECGenerateReq.
 // This will unmarshal a UNIX epoch integer, a RFC3339 string, the old format string used by Perl '2018-08-21+14:26:06', and the old format string sent by the Portal '2018-08-21 14:14:42'.
 // This exists to fix a critical bug, see https://github.com/apache/trafficcontrol/issues/2723 - it SHOULD NOT be used by any other endpoint.
--- a/traffic_ops/traffic_ops_golang/cdn/dnssec.go
+++ b/traffic_ops/traffic_ops_golang/cdn/dnssec.go
@@ -90,29 +90,15 @@ func GetDNSSECKeys(w http.ResponseWriter, r *http.Request) {
 	}
 	defer inf.Close()
 
-	cdnName := inf.Params["name"]
-
-	riakKeys, keysExist, err := riaksvc.GetDNSSECKeys(cdnName, inf.Tx.Tx, inf.Config.RiakAuthOptions)
-	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("getting DNSSEC CDN keys: "+err.Error()))
+	cdnName := tc.CDNName(inf.Params["name"])
+	keys, userErr, sysErr, errCode := getDNSSECKeys(inf, cdnName)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
 		return
 	}
-	if !keysExist {
+	if keys == nil {
 		// TODO emulates Perl; change to error, 404?
-		api.WriteRespAlertObj(w, r, tc.SuccessLevel, " - Dnssec keys for "+cdnName+" could not be found. ", struct{}{}) // emulates Perl
-		return
-	}
-
-	dsTTL, err := GetDSRecordTTL(inf.Tx.Tx, cdnName)
-	if err != nil {
-		log.Errorln("Getting DNSSEC Keys: getting DS Record TTL from CRConfig Snapshot: " + err.Error())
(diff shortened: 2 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3156/
Test FAILed.

rob05c:
Hm. This prevents new fields from ever being null, and not omitted. If a new field is null, it will be omitted from the JSON object returned (because this scheme requires all new fields to be `omitempty`). We can define new fields to behave that way, so it's ok from a versioning standpoint. But are we ok with that limitation? Are you sure we won't ever encounter a scenario in the future, where we need to return `"newField": null`?

I'm also not a fan of the `func(interface{}) interface{}`. Empty interfaces throw away all type safety. At the least, this would really need to be `func(interface{}) (interface{}, error)` to avoid panicing. But even then, it's a runtime error, instead of a compile error like it currently is. Sometimes `interface{}` is inevitable, but it'd be really nice to keep these kind of errors at compile time, if we can.

It'd also be ideal if we could avoid Reflection, which is inevitable with struct tags. I'm less worried about the performance, than the readability. If someone who didn't write it comes along in the future, how difficult will it be to climb the learning curve, for the reflecting code that's parsing the struct tags? 

I'm not outright opposed. But these are the limitations I'm seeing, which would be really nice if we could figure out how to avoid.

rob05c:
So, I took a stab at making the versioning better, too:

https://github.com/rob05c/trafficcontrol/commit/4732c665f3d3b1d05bd13154576aa21670e2df77

https://github.com/rob05c/trafficcontrol/blob/4732c665f3d3b1d05bd13154576aa21670e2df77/lib/go-tc/deliveryservices.go

https://github.com/rob05c/trafficcontrol/blob/4732c665f3d3b1d05bd13154576aa21670e2df77/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go

https://github.com/rob05c/trafficcontrol/blob/4732c665f3d3b1d05bd13154576aa21670e2df77/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservicesv11.go

So, this puts all the logic of the latest version in `deliveryservices.go`, and each version `deliveryservicesv11.go` etc is identical to the previous except `V11` -> `V13`.

So, to add a new version, all you have to do is
1. Add the new struct in `lib/go-tc`, e.g. `type DeliveryServiceNullableV15 struct {DeliveryServiceNullableV13; ConsistentHashRegex *string}` with its conversion func, sanitize, and update `type DeliveryServiceNullable DeliveryServiceNullableV15`
2. Add your logic to `traffic_ops_golang/deliveryservices/deliveryservices.go`
3. `cp deliveryservicesv11.go deliveryservicesv15.go && sed -i 's/V11/V15/' deliveryservicesv15.go`

We can and should add those steps to the documentation, and possibly scripts to make the new `deliveryservicesv1x.go` files.

The generated code isn't ideal, but there's no logic there, and whenever Go2 is out, we should be able to use Generics to get rid of it.

(FYI I attempted to write Go ducktyping/interfaces to abstract the generic stuff, I think it's possible, but it'd be a mess.)

But, you don't have to touch old versions, when adding a new version. That lets you add e.g. 1.5 without having to touch anything in 1.4, 1.3, 1.2, 1.1. Only the latest-version `deliveryservices.go`, and things for the version you're adding.

What do you think of that solution?

rawlinp:
> Hm. This prevents new fields from ever being null, and not omitted. If a new field is null, it will be omitted from the JSON object returned (because this scheme requires all new fields to be `omitempty`). We can define new fields to behave that way, so it's ok from a versioning standpoint. But are we ok with that limitation? Are you sure we won't ever encounter a scenario in the future, where we need to return `"newField": null`?

New optional fields are meant to have a sane default in most cases (i.e. "null" should be able to be interpreted as something concrete), in which case I'd think the API should just populate the struct with whatever that sane default is. So I don't think that limitation would be a big deal, but it's hard to say for sure. In terms of API guidelines, maybe returning null where null can be interpreted as a sane default with a concrete value should be prohibited anyways.

> I'm also not a fan of the `func(interface{}) interface{}`. Empty interfaces throw away all type safety. At the least, this would really need to be `func(interface{}) (interface{}, error)` to avoid panicing. But even then, it's a runtime error, instead of a compile error like it currently is. Sometimes `interface{}` is inevitable, but it'd be really nice to keep these kind of errors at compile time, if we can.

I think temporarily dropping down into `interface{}` to nullify fields based on tags then type-asserting back to the concrete value after nullifying would be alright. The runtime error would be easy to avoid since there would only be one concrete struct type for a given endpoint.

> It'd also be ideal if we could avoid Reflection, which is inevitable with struct tags. I'm less worried about the performance, than the readability. If someone who didn't write it comes along in the future, how difficult will it be to climb the learning curve, for the reflecting code that's parsing the struct tags?

I think the function that parses struct tags to nullify fields with versions greater than the given version would be fairly self-contained. As long as you were somewhat familiar with Go reflection, I'd think it wouldn't be too steep of a learning curve. The function would essentially do one thing and do it well, and produce well-defined outputs for a given input, so it would probably be easy to test with unit tests with various structs as input.

> The generated code isn't ideal, but there's no logic there, and whenever Go2 is out, we should be able to use Generics to get rid of it.
>
> (FYI I attempted to write Go ducktyping/interfaces to abstract the generic stuff, I think it's possible, but it'd be a mess.)
>
> But, you don't have to touch old versions, when adding a new version. That lets you add e.g. 1.5 without having to touch anything in 1.4, 1.3, 1.2, 1.1. Only the latest-version deliveryservices.go, and things for the version you're adding.
>
> What do you think of that solution?

I like that solution better than having to update all the previous versions of an endpoint when adding a new version, but the deeper I get into this the more I think jumping through all these hoops just to prevent a 1.1 requester from seeing 1.3 fields in their response isn't worth all the hassle. The API should just honor the major version, and clients should be made aware that if they start seeing new fields in their responses, that means they can start using them. If they don't see new fields, then they can't start using fields they've learned about from other TO API deployments that are running the latest code.

Updating an API to add a new optional field should be as easy as updating the single existing implementation with the new field. We shouldn't have to copy/paste entire files and jump through all the minor-version hoops just to hide things from the user. Plus, promising to handle all minor versions of an endpoint means maintaining and testing all those minor versions forever, when in reality we are only going to be testing the latest version of the API, and clients should always migrate to the latest available version of the API anyways.

If Go 2 gets generics and we find that we can more reasonably handle every single minor version of an endpoint using generics, then we should consider doing so at that point in time. But right now I don't think hiding new, backwards-compatible fields from users is worth duplicating entire files every time a new minor version is required.

Issue #3304, opened 2019-02-05
Golang endpoints should accept "string integers"


moltzaum:
```json
{
        "name": "Manchester",
        "division": 1,
        "divisionName": "England"
}
```

The old perl API accepts both integers and strings since it is a dynamic language. In most go endpoints only an integer is accepted.

For instance, the regions endpoint documentation says that the POST supports strings: https://traffic-control-cdn.readthedocs.io/en/latest/api/regions.html
But actually POSTing an "integer string" yields an error message. 

The go endpoints should be able to accept either integer or "string integers" to reduce the likelihood of breaking things that were once supported.

rob05c:
IMO this isn't a bad thing, following the Robustness Principle, but we shouldn't require it for Go endpoints. Perl accepts both because it's "dynamic," but if the API docs define it one way, and Perl returns that, I think it's sufficient to make Go to follow the docs.

It's also going to be a bit of a mess, if we make all Go endpoints accept these for every endpoint, to change every `int` to a custom `JSONIntStr`. Not only in the Server, but it'll make the structs we share with the client awkward, and client users will have to e.g. `int(cdn.ID)` to use them.

moltzaum:
Do you think that it would be possible to avoid using `JSONIntStr` and instead replace `"[0-9]*"` with the number inside the quotes before decoding? Would that break anything?

Another potential solution is to replace `[0-9]` (excluding quotes) with the number now inside quotes. Then the json struct tag can be `",string"` for that field.

jhg03a:
I'm generally an advocate for strong typing.  It prevents problems like the number `b`.  That said, API in my opinion are contracts.  Even if we adapt TP (which is really just a client) to use strong typing, I don't believe it's fair to require all 3rd parties to also adopt strong typing retroactively.  So I vote that we embrace strong typing on API versions and routes/methods where perl has never existed.  That said if we had a perl endpoint with weak types we should downgrade to weak types in Golang.  Also, this doesn't just stop at integers.  Perl used weak types for booleans and probably floats too.

rawlinp:
I believe this was inadvertently closed automatically; this is still an issue with most of the TO Go API.

zrhoffman:
API 1.x should have preserved backwards compatibility with Perl to ensure that any types accepted in Perl were accepted in the TO Go implementation, too, but in API 2.x and above, this behavior is explicitly no longer supporter, and API 1.x is deprecated, pending #5539 being merged.

[...]

PR #3306, opened 2019-02-05
Use JSONIntStr for user endpoint

moltzaum:
#### What does this PR do?

Has a temporary fix for #3304 and is just for the users endpoint. Thinking a better solution exists for all endpoints, but this is to wrap up the current TC-verification process.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

POST a user with "" for the role or tenant id. I've presumed the gid and uid don't need to use JSONIntStr, nor the main id.

#### Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/client/user.go
+++ b/traffic_ops/client/user.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
+	util "github.com/apache/trafficcontrol/lib/go-util"
 )
 
 // Users gets an array of Users.
@@ -78,6 +79,7 @@ func (to *Session) GetUserCurrent() (*tc.UserCurrent, ReqInf, error) {
 
 // CreateUser creates a user
 func (to *Session) CreateUser(user *tc.User) (*tc.CreateUserResponse, ReqInf, error) {
+
 	if user.TenantID == nil && user.Tenant != nil {
 		tenant, _, err := to.TenantByName(*user.Tenant)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


traffic_ops/client/user.go
```diff
@@ -24,6 +24,7 @@ import (
 	"strconv"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
+	util "github.com/apache/trafficcontrol/lib/go-util"
```
rob05c: Nitpick: the alias isn't necessary, `go-util` package automatically imports as `util`

moltzaum: That was done automatically when I saved. ¯\_(ツ)_/¯

moltzaum: I think it does that  because the go convention is for the package name to match the folder name.

rawlinp:
Doesn't this break the golang client from 2.2.x? https://github.com/apache/trafficcontrol/blob/2.2.x/lib/go-tc/users.go

If so, Go TO client users should be notified somehow.

moltzaum:
It seems to break whatever might be using the `go-tc` structs on an upgrade to 3.0.x. Anyone referencing the Role and TenantID field in particular will need to cast to an int while using it.

rob05c:
I see two options:
1. leave the code, and add a note to the CHANGELOG.md
2. Change `tc.User` back to `int`s, and anonymously embed a `JSONIntStr` in the TO code, e.g.
```
type TOUser struct {
	ReqInfo *api.APIInfo `json:"-"`
	tc.User

	Role            *util.JSONIntStr `json:"role" db:"role"`
	TenantID        *util.JSONIntStr `json:"tenantId" db:"tenant_id"`
}

⋮ 

if user.Role != nil {
  user.User.Role = util.IntPtr(*user.Role)
}
```

I'd lean toward the latter. It's not that much code on our side, and it provides a cleaner user experience. And besides, we're already doing it in other places, e.g. where we need a `pq.Array`.

PR #3313, opened 2019-02-08
Ort.py environment variables

ocket8888:
## What does this PR do?
The Python implementation of ORT will now use the semi-standard environment variables:

TO_URL
    Specifies the URL/hostname of the Traffic Ops server, optionally with a port (must match `(?i:https?://)?\w+(\.\w+)*(:\d+)?` )
TO_USER
   Specifies the name of the user as whom to authenticate with Traffic Ops
TO_PASSWORD
    Specifies the password of the user as whom to authenticate with Traffic Ops

This PR adds a new call signature to ORT.py - in addition to the old, Perl-compatible(ish) call signature - wherein specification of these pieces of information is optional when these environment variables are defined. They can also be overridden when they **are** defined by using the new `--to_url`, `-u`/`--to_user`, and `-p`/`--to_password` options, respectively. This new call signature also makes the log level optional (provided by the `-l`/`--log_level` option), and it will now default to `WARN`.

- [x] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [x] Traffic Ops ORT (Python)
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Define those environment variables and verify that

1. When using the old call signature, they are ignored and operations proceed normally
2. When using the new call signature, they are used and operations proceed normally
3. When using the new call signature, they can be overridden by command-line options and operations proceed normally.

This PR also includes updates to the documentation that should go into greater detail regarding the command line options allowed, so feel free to build and look at that if you need more information

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/docs/source/development/ort/traffic_ops_ort.configuration.rst
+++ b/docs/source/development/ort/traffic_ops_ort.configuration.rst
@@ -13,13 +13,10 @@
 .. limitations under the License.
 ..
 
+**************************************
 traffic\_ops\_ort.configuration module
-======================================
-
-.. autodata:: traffic_ops_ort.configuration.HOSTNAME
-	:annotation:
-
+**************************************
 .. automodule:: traffic_ops_ort.configuration
 	:members:
-    :undoc-members:
-    :show-inheritance:
+	:undoc-members:
+	:show-inheritance:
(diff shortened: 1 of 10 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3194/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3198/
Test PASSed.

PR #3326, opened 2019-02-12
Added logging.yaml support

ezelkow1:
Generate a yaml file if the logging.yaml file is specified as output. Tested with CIAB and it appears to work. Fed the resulting yaml to an ats8 install and it appeared to be happy (once removing things like pqsn that do not exist in ats8)

Fixes #(replace_this_text_with_issue_number) 

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [x] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
+++ b/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
@@ -299,6 +299,7 @@ sub get_profile_config {
 	elsif ( $filename eq "drop_qstring.config" ) { $file_contents = $self->drop_qstring_dot_config( $profile_obj, $filename ); }
 	elsif ( $filename eq "logs_xml.config" ) { $file_contents = $self->logs_xml_dot_config( $profile_obj, $filename ); }
 	elsif ( $filename eq "logging.config" ) { $file_contents = $self->logging_dot_config( $profile_obj, $filename ); }
+	elsif ( $filename eq "logging.yaml" ) { $file_contents = $self->logging_dot_yaml( $profile_obj, $filename ); }
 	elsif ( $filename eq "plugin.config" ) { $file_contents = $self->generic_profile_config( $profile_obj, $filename ); }
 	elsif ( $filename eq "records.config" ) { $file_contents = $self->generic_profile_config( $profile_obj, $filename ); }
 	elsif ( $filename eq "storage.config" ) { $file_contents = $self->storage_dot_config( $profile_obj, $filename ); }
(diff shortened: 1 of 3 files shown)


ezelkow1:
screen shot 2019-02-12 at 3 27 16 pm
screen shot 2019-02-12 at 3 26 43 pm

asfgit:

ezelkow1:
@alficles if you want to take a look

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
```diff
@@ -1424,6 +1426,94 @@ sub drop_qstring_dot_config {
...
+sub logging_dot_yaml {
+	my $self        = shift;
+	my $profile_obj = shift;
+
+	my $data = $self->profile_param_data( $profile_obj->id, "logging.yaml" );
+
+	# This is an YAML file, so we need to massage the header a bit for YAML commenting.
+	my $text = "# " . $self->header_comment( $profile_obj->name );
+	$text =~ s/# //;
+	$text =~ s/\n//;
```
alficles: We appear to be adding a `# ` and then removing it.

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
```diff
@@ -1424,6 +1426,94 @@ sub drop_qstring_dot_config {
...
+
+		my $log_object_filename = $data->{$log_object_field . ".Filename"} || "";
+		if ( length($log_object_filename) > 0 ) {
+			my $log_object_type                 = $data->{$log_object_field . ".Type"}               || "ascii";
+			my $log_object_format               = $data->{$log_object_field . ".Format"}             || "";
+			my $log_object_rolling_enabled      = $data->{$log_object_field . ".RollingEnabled"}     || "";
+			my $log_object_rolling_interval_sec = $data->{$log_object_field . ".RollingIntervalSec"} || "";
+			my $log_object_rolling_offset_hr    = $data->{$log_object_field . ".RollingOffsetHr"}    || "";
+			my $log_object_rolling_size_mb      = $data->{$log_object_field . ".RollingSizeMb"}      || "";
+			my $log_object_filters              = $data->{$log_object_field . ".Filters"}            || "";
```
alficles: Is it possible for `$log_object_filters` to have a return carriage in it? Cause that would cause problems below.

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
ezelkow1: Removed the # stripping and the # usage in the comment. The default header comment has a # in it and so is already a valid YAML comment

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
ezelkow1: tested now adding \n's and \r's to a filters object and they get written exactly as is into the yaml file and not interpreted as newlines. So I think this is ok as is

ezelkow1: added stripping of all types of newline, \v repeating, that appears to work

Shihta:
@ezelkow1 could you share with me an example that how to use/write logging.yaml?

PR #3330, opened 2019-02-15
Fix TM nil pointer panic

rob05c:

Fixes a Monitor nil pointer panic, from when a poll happened to not include any caches from a given cachegroup that was previously polled.

This is a race condition, but, as part of the fix, I added logging to see when the new (correct) default construction, and observed it getting hit after a few days. So I'm certain this fixes the race.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [x] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Verify Monitor runs as expected. This was a race condition, so it can't really be verified. You can run the Monitor with a large dataset for a few days, and see that it doesn't crash.

## Check all that apply

- [x] This PR includes tests

### DIFF
--- a/traffic_monitor/ds/stat.go
+++ b/traffic_monitor/ds/stat.go
@@ -20,6 +20,7 @@ package ds
  */
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3227/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3228/
Test PASSed.

PR #3333, opened 2019-02-18
Add TO deleting old DS certs on snapshot

### LINKED ISSUE #882: [TC-516] Deleting a DS thru the TO API should also delete all SSL keys (if applicable)
<p>There are currently 4 protocols for a DS:</p>

<p>        0 - HTTP<br/>
        1 - HTTPS<br/>
        2 - HTTP AND HTTPS<br/>
        3 - HTTP TO HTTPS</p>

<p>Currently, if you delete a DS thru the API, SSL keys are not deleted. This should occur if ds.protocol &gt; 0.</p>

Author: Jeremy Mitchell
JIRA Link: <a href="https://issues.apache.org/jira/browse/TC-516">https://issues.apache.org/jira/browse/TC-516</a>
Found Version: 2.1.0

rob05c:

Fixes #882

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Riak endpoint, TO API tests don't support Riak yet.

To verify:
1. Enable info logging in Traffic Ops.
2. Delete an HTTPS delivery service with SSL keys. 
3. Snapshot the CRConfig on that DS's CDN.

Traffic Ops should log to the info log `Deleted Riak SSL keys for delivery service which has been deleted in the database 'your-ds' key 'your-ds-latest'.

The endpoint `/api/1.2/cdns/name/your-cdn-name/sslkeys` should no longer contain that delivery service.

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/docs/source/api/cdns_id_snapshot.rst
+++ b/docs/source/api/cdns_id_snapshot.rst
@@ -25,6 +25,8 @@
 =======
 Performs a CDN snapshot. Effectively, this propagates the new *configuration* of the CDN to its *operating state*, which replaces the output of the :ref:`to-api-cdns-name-snapshot` endpoint with the output of the :ref:`to-api-cdns-name-snapshot-new` endpoint.
 
+.. Note:: Snapshotting the CDN also deletes all HTTPS certificates for every :term:`Delivery Service` which has been deleted since the last CDN :term:`Snapshot`.
+
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
 :Response Type:  ``undefined``
--- a/docs/source/api/snapshot_name.rst
+++ b/docs/source/api/snapshot_name.rst
@@ -23,6 +23,8 @@
 =======
 Performs a CDN snapshot. Effectively, this propagates the new *configuration* of the CDN to its *operating state*, which replaces the output of the :ref:`to-api-cdns-name-snapshot` endpoint with the output of the :ref:`to-api-cdns-name-snapshot-new` endpoint.
 
+.. Note:: Snapshotting the CDN also deletes all HTTPS certificates for every :term:`Delivery Service` which has been deleted since the last :term:`Snapshot`.
+
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
 :Response Type:  ``undefined``
--- a/traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
@@ -1,0 +1,197 @@
+package deliveryservice
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/config"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/riaksvc"
+
+	"github.com/basho/riak-go-client"
+)
+
+// DeleteOldCerts asynchronously deletes HTTPS certificates in Riak which have no corresponding delivery service in the database.
+//
+// Note the delivery service may still be in the CRConfig! Therefore, this should only be called immediately after a CRConfig Snapshot.
+//
+// This creates a goroutine, and immediately returns. It returns an error if there was an error preparing the delete routine, such as an error creating a db transaction.
+//
+// Note because it is asynchronous, this may return a nil error, but the asynchronous goroutine may error when fetching or deleting the certificates. If such an error occurs, it will be logged to the error log.
+//
+// If certificate deletion is already being processed by a goroutine, another delete will be queued, and this immediately returns nil. Only one delete will ever be queued.
+//
+func DeleteOldCerts(db *sql.DB, tx *sql.Tx, cfg *config.Config, cdn tc.CDNName) error {
+	if !cfg.RiakEnabled {
+		log.Infoln("deleting old delivery service certificates: Riak is not enabled, returning without cleaning up old certificates.")
+		return nil
+	}
+	if db == nil {
+		return errors.New("nil db")
+	}
+	if cfg == nil {
+		return errors.New("nil config")
+	}
+	startOldCertDeleter(db, tx, time.Duration(cfg.DBQueryTimeoutSeconds)*time.Second, cfg.RiakAuthOptions, cfg.RiakPort, cdn)
+	cleanupOldCertDeleters(tx)
+	return nil
+}
+
+// deleteOldDSCerts deletes the HTTPS certificates in Riak of delivery services which have been deleted in Traffic Ops.
+func deleteOldDSCerts(tx *sql.Tx, authOpts *riak.AuthOptions, riakPort *uint, cdn tc.CDNName) error {
+	dsKeys, err := riaksvc.GetCDNSSLKeysDSNames(tx, authOpts, riakPort, cdn)
+	if err != nil {
+		return errors.New("getting riak ds keys: " + err.Error())
+	}
+
+	dses, err := dbhelpers.GetCDNDSes(tx, cdn)
+	if err != nil {
+		return errors.New("getting ds names: " + err.Error())
+	}
+
+	successes := []string{}
+	failures := []string{}
+	for ds, riakKeys := range dsKeys {
+		if _, ok := dses[ds]; ok {
+			continue
+		}
+		for _, riakKey := range riakKeys {
+			err := riaksvc.DeleteDeliveryServicesSSLKey(tx, authOpts, riakPort, riakKey)
+			if err != nil {
+				log.Errorln("deleting Riak SSL keys for Delivery Service '" + string(ds) + "' key '" + riakKey + "': " + err.Error())
+				failures = append(failures, string(ds))
+			} else {
+				log.Infoln("Deleted Riak SSL keys for delivery service which has been deleted in the database '" + string(ds) + "' key '" + riakKey + "'")
+				successes = append(successes, string(ds))
+			}
+		}
+	}
+	if len(failures) > 0 {
+		return errors.New("successfully deleted Riak SSL keys for deleted dses [" + strings.Join(successes, ", ") + "], but failed to delete Riak SSL keys for [" + strings.Join(failures, ", ") + "]; see the error log for details")
+	}
+	return nil
+}
+
+// deleteOldDSCertsDB takes a db, and creates a transaction to pass to deleteOldDSCerts.
+func deleteOldDSCertsDB(db *sql.DB, dbTimeout time.Duration, riakOpts *riak.AuthOptions, riakPort *uint, cdn tc.CDNName) {
+	dbCtx, cancelTx := context.WithTimeout(context.Background(), dbTimeout)
+	tx, err := db.BeginTx(dbCtx, nil)
+	if err != nil {
+		log.Errorln("Old Cert Deleter Job: beginning tx: " + err.Error())
+		return
+	}
+	defer cancelTx()
+	txCommit := false
+	defer dbhelpers.CommitIf(tx, &txCommit)
+	if err := deleteOldDSCerts(tx, riakOpts, riakPort, cdn); err != nil {
+		log.Errorln("deleting old DS certificates: " + err.Error())
+		return
+	}
+	txCommit = true
+}
+
+var oldCertDeleters = OldCertDeleters{D: map[tc.CDNName]*OldCertDeleter{}}
+
+type OldCertDeleters struct {
+	D map[tc.CDNName]*OldCertDeleter
+	M sync.Mutex
+}
+
+// startOldCertDeleter tells the old cert deleter goroutine to start another delete job, creating the goroutine if it doesn't exist.
+func startOldCertDeleter(db *sql.DB, tx *sql.Tx, dbTimeout time.Duration, riakOpts *riak.AuthOptions, riakPort *uint, cdn tc.CDNName) {
+	oldCertDeleter := getOrCreateOldCertDeleter(cdn)
+	oldCertDeleter.Once.Do(func() {
+		go doOldCertDeleter(oldCertDeleter.Start, oldCertDeleter.Die, db, dbTimeout, riakOpts, riakPort, cdn)
+	})
+
+	select {
+	case oldCertDeleter.Start <- struct{}{}:
+	default:
+	}
+}
+
+func getOrCreateOldCertDeleter(cdn tc.CDNName) *OldCertDeleter {
+	oldCertDeleters.M.Lock()
+	defer oldCertDeleters.M.Unlock()
+	oldCertDeleter, ok := oldCertDeleters.D[cdn]
+	if !ok {
+		oldCertDeleter = newOldCertDeleter()
+		oldCertDeleters.D[cdn] = oldCertDeleter
+	}
+	return oldCertDeleter
+}
+
+// cleanupOldCertDeleters stops all cert deleter goroutines for CDNs which have been deleted in the database.
+// Any error is logged, but not returned.
+// This is designed to be called when starting a new cert deleter job, to clean up any old cert deleters from deleted CDNs.
+// This should only be called immediately when snapshotting, and immediately after startOldCertDeleter, because otherwise a cert deleter may be removed before it can delete old certs for a given CDN.
+func cleanupOldCertDeleters(tx *sql.Tx) {
+	cdns, err := dbhelpers.GetCDNs(tx) // (map[tc.CDNName]struct{}, error) {
+	if err != nil {
+		log.Errorln("cleanupOldCertDeleters: getting cdns: " + err.Error())
+		return
+	}
+
+	oldCertDeleters.M.Lock()
+	defer oldCertDeleters.M.Unlock()
+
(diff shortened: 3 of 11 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3237/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3238/
Test PASSed.

ocket8888:
This should include a `.. note` or something in the snapshot endpoint docs. Or maybe note it somewhere in the Traffic Vault documentation?

On another note, does this change in behavior necessitate a change in API version? (i.e. only 1.5 has this new functionality)

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
```diff
@@ -0,0 +1,136 @@
...
+
+// DeleteOldCerts asychronously deletes HTTPS certificates in Riak which have no corresponding delivery service in the database.
+//
+// Note the delivery service may still be in the CRConfig! Therefore, this should only be called immediately after a CRConfig Snapshot.
+//
+// This creates a goroutine, and immediately returns. It returns an error if there was an error preparing the delete routine, such as an error creating a db transaction.
+//
+// Note because it is asynchronous, this may return a nil error, but the asynchronous goroutine may error when fetching or deleting the certificates. If such an error occurs, it will be logged to the error log.
+//
+// If certificate deletion is already being processed by a goroutine, this immediately returns nil.
```
rawlinp: q: why? What if there is a DS deletion + snapshot shortly after this snapshot while an existing transaction is still in progress. Wouldn't we want that 2nd attempt to simply wait until the current transaction is done being processed, then start its own transaction to delete certs for the latest deleted DS?

rawlinp:
Tested this out, appears to do what it says on the tin.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/config"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/riaksvc"
+
+	"github.com/basho/riak-go-client"
+)
+
+// DeleteOldCerts asychronously deletes HTTPS certificates in Riak which have no corresponding delivery service in the database.
```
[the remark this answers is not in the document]
rob05c: Fixed

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rob05c: Right. But we don't want them to get backed up. If people do 50 snapshots at once, then the first deletion job starts, we don't want to do 49 more, we just want to do the 1 more to fix the "race."

The ideal solution is to have a queue of 1, and if 1 is going, and 49 more come in, the first one waits, and the other 48 say "someone is already waiting, we're done."

That's just a little more complicated, and it seemed like an unlikely race. It'd only be different if someone deleted a DS and snapshotted, all in-between the time the job got DSes from Riak and deleted them.

I'll try to see how much more work that "queue of 1" is, though.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3261/
Test PASSed.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rawlinp: Yeah I agree on it being fairly unlikely, but I figure if it's not too much trouble to avoid it then maybe we should. We could also just push that bit out into a separate PR for later if it's nontrivial.

rob05c: I just realized, it's more important than that (and more difficult), because it has to be per-cdn.

Right now, if snapshots of 2 CDNs come in at once, it'll only run the first. That's a problem.

I'll try to have it done by the end of the day.

rob05c:
@ocket8888 I don't think it needs a version change, this is really a bug, not a feature. It's a bug that it didn't do this before. See #882.

I'll add something to the docs.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rawlinp: Oh! Good catch. That seems like a much more likely race.

ocket8888:
Cool, thanks :+1:

rob05c:
@rawlinp Ok, I added a "queue of 1" per-CDN. Not sure there's a great way to verify the internals, without adding your own debug prints, but FWIW I added debug prints, and verified 
1. Multiple concurrent snapshots work, and don't deadlock.
2. The second concurrent snapshot writes to the job chan `case oldCertDeleter.Start ← struct{}{}:`, entering the "queue of 1" to be processed when the current job finishes.
3. Additional concurrent snapshots hit the `default:` on the channel and don't execute.
4. Deleting a CDN which was previously snapshotted, and snapshotting a different CDN, cleans up the deleted CDN in the `oldCertDeleters` map.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3263/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3264/
Test PASSed.

rob05c:
@ocket8888 Added docs https://github.com/apache/trafficcontrol/pull/3333/commits/30d2c0daac5c20a6e114aaf460c91f330e1e775f

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rob05c: Done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3265/
Test PASSed.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
```diff
@@ -0,0 +1,197 @@
...
+		Start: make(chan struct{}, 1),
+		Die:   make(chan struct{}, 1),
+	}
+}
+
+func doOldCertDeleter(do chan struct{}, die chan struct{}, db *sql.DB, dbTimeout time.Duration, riakOpts *riak.AuthOptions, riakPort *uint, cdn tc.CDNName) {
+	for {
+		select {
+		case <-do:
+			deleteOldDSCertsDB(db, dbTimeout, riakOpts, riakPort, cdn)
```
rawlinp: could this be simplified into the following?
```
for _ := range do {
    ...
}
```
Then this `select` wouldn't be necessary since we would no longer need the `die chan`, and `cleanupOldCertDeleters` could just `close(oldCertDeleter.Start)` instead of `select`ing as well.

This might also prevent the case where there is an outstanding "delete old CDN certs" job in the `do` channel but the CDN was deleted at the same time.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// cleanupOldCertDeleters stops all cert deleter goroutines for CDNs which have been deleted in the database.
+// Any error is logged, but not returned.
+// This is designed to be called when starting a new cert deleter job, to clean up any old cert deleters from deleted CDNs.
+// This should only be called immediately when snapshotting, and immediately after startOldCertDeleter, because otherwise a cert deleter may be removed before it can delete old certs for a given CDN.
+func cleanupOldCertDeleters(tx *sql.Tx) {
+	cdns, err := dbhelpers.GetCDNs(tx) // (map[tc.CDNName]struct{}, error) {
+	if err != nil {
+		log.Errorln("StopOldCertDeleters: getting cdns: " + err.Error())
```
rawlinp: `s/Stop/cleanup/`

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	oldCertDeleters.M.Lock()
+	defer oldCertDeleters.M.Unlock()
+
+	for cdn, oldCertDeleter := range oldCertDeleters.D {
+		if _, ok := cdns[cdn]; ok {
+			continue
+		}
+		delete(oldCertDeleters.D, cdn)
```
rawlinp: Deleting from a thing you're iterating over makes my spidey senses go off, but supposedly this is safe in Go.

docs/source/api/cdns_id_snapshot.rst
```diff
@@ -25,6 +25,8 @@
 =======
 Performs a CDN snapshot. Effectively, this propagates the new *configuration* of the CDN to its *operating state*, which replaces the output of the :ref:`to-api-cdns-name-snapshot` endpoint with the output of the :ref:`to-api-cdns-name-snapshot-new` endpoint.
 
+.. Note:: Snapshotting the CRConfig also deletes all HTTPS certificates for delivery services which have been deleted since the last snapshot.
```
ocket8888: Please use "Snapshot" or "CDN Snapshot" as opposed to "CRConfig" (not a big deal in the TO using docs, because that whole thing is deprecated/disabled. Frankly I wouldn't care if you didn't update it). It's probably less awkward to refer to the action as "taking a CDN Snapshot" rather than "Snapshotting the CDN Snapshot".

Also, "Snapshot" and "Delivery Service" are ATC *terms* and should be linked to the glossary. At a minimum that should be done once per major section ("page" in an HTML context) but ideally it'd be all of them. Regardless of whether you decide to link them here - assuming it is done elsewhere in the same major section - they should be capitalized like that for ease of recognition.

docs/source/api/snapshot_name.rst
```diff
@@ -23,6 +23,8 @@
 =======
 Performs a CDN snapshot. Effectively, this propagates the new *configuration* of the CDN to its *operating state*, which replaces the output of the :ref:`to-api-cdns-name-snapshot` endpoint with the output of the :ref:`to-api-cdns-name-snapshot-new` endpoint.
 
+.. Note:: Snapshotting the CRConfig also deletes all HTTPS certificates for delivery services which have been deleted since the last snapshot.
```
ocket8888: Same as above.

docs/source/api/cdns_id_snapshot.rst
ocket8888: E.g. term usage

```rst
This -> :term:`Delivery Service` will link to the glossary for the definition of a DS.
For the moment, the glossary has a problem with plurals, so to link the plural of
a term to the glossary, link the singular form and place an 's' with an escaped
space after the closing backtick, e.g. :term:`Delivery Service`\ s.
```

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rob05c: Go channels can only be safely closed by a single writer goroutine. If `close` is called on an already closed chan, it panics. Here, any number of client requests goroutines could be executing the cleanup. This would introduce a race condition, if multiple requests happened to be in the block at once.

It's possible to make it safe, by either ensuring the `OldCertDeleters` mutex is always locked when the channel is used, or adding a mutex for each `OldCertDeleter` chan.

There would also be another race when writing to the chan, if another request goroutine closed it. So, either another lock would have to be introduced in the start, or code would have to be added to check if the chan was closed before writing to it. 

Go also has no way to check if a chan is closed without reading from it, so if either the writer or closer checked for a close, it would have to also write back what it just read.

So, it's possible to use `range` and `close`, but it wouldn't be simpler. Further, the map mutex is inevitable (unless a `sync.Map` were used), but mutexing channels is very unexpected. Because `chan` is defined to be safe for multiple goroutines, it seems likely someone might come along in the future, not fully understand the race conditions because "chan is threadsafe", and remove mutexes to simplify things, introducing races.

>This might also prevent the case where there is an outstanding "delete old CDN certs" job in the `do` channel but the CDN was deleted at the same time.

I'm not sure what you mean exactly, but we want the final delete job to go through, even if the channel is closed. For example, if someone snapshots and deletes the cdn at the same time.

Hence, a `range`/`close` change would also have to account for that, and be careful not to introduce races around that, such that a final delete job didn't get executed.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rob05c: Fixed

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3269/
Test PASSed.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rob05c: Yes, it's defined safe in Go. https://golang.org/ref/spec#For_statements

It seems like the clearest thing here, we really do need to delete elements matching a criteria. We could put things to be deleted in a list, and then iterate and delete from that list, but the only advantage would be to slightly ease reading for programmers unfamiliar with Go. 

I tend to think it's worth the readability and simplicity for people who are familiar with Go's relatively small spec.

docs/source/api/cdns_id_snapshot.rst
rob05c: >link the singular form and place an 's' with an escaped space after the closing backtick, e.g.
:term:`Delivery Service`\ s.

Tangent: How terrible would it be to recommend using zero-width spaces, so it looks right?
https://www.fileformat.info/info/unicode/char/200B/index.htm

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rawlinp: Ah, I didn't realize closing a closed channel actually panics (thanks, golang).

I was thinking that this block:
```
	for {
		select {
		case <-do:
			deleteOldDSCertsDB(db, dbTimeout, riakOpts, riakPort, cdn)
		case <-die:
			return
		}
	}
```
would potentially read from the `die` channel when there is an outstanding job in the `do` channel, which would cause the goroutine to exit without handling that final delete job. This is because it reads from a random channel if multiple are ready.

Can we just `defer` a `recover()` in case the channel has already been closed by another goroutine?

docs/source/api/cdns_id_snapshot.rst
rob05c: >use "Snapshot" or "CDN Snapshot" as opposed to "CRConfig"

>"Snapshot" and "Delivery Service" are ATC _terms_ and should be linked

Done.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3270/
Test PASSed.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rob05c: >Can we just `defer` a `recover()` in case the channel has already been closed by another goroutine?

We could, but that's very un-idiomatic. Go chans panicing on double-close is an intentional decision by the Go authors to suggest you should rethink your design, if you need that. I'd really prefer not to write code with a known race, that violates good design principles, handled by `recover`. A `die` chan is clean and obvious what it's doing, and I think is the common and accepted way of doing this in Go.

>would potentially read from the `die` channel when there is an outstanding job in the `do` channel, which would cause the goroutine to exit without handling that final delete job.

You're right. Though I think it would have to be two writes, a second write while a job is already executing, concurrently with a CDN delete and snapshot. The "perfect" way to handle that is:

```
	for {
		select {
		case <-do:
			deleteOldDSCertsDB(db, dbTimeout, riakOpts, riakPort, cdn)
		case <-die:
			select {
			case <-do:
				deleteOldDSCertsDB(db, dbTimeout, riakOpts, riakPort, cdn)
			default:
			}
			return
		}
	}
```

You can see the Monitor doing this in places. I'd decided it wasn't worth the readability/complexity for the odds of that triple race happening. But in retrospect, I can agree, even the unlikeliest races cause outages years later when you least expect. I'll change it.

rob05c: Done.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3271/
Test PASSed.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
rawlinp: Yeah I was thinking of the Python way of "do it now and ask forgiveness later". I think your nested selects get the job done too and are not that complicated to read IMO.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3272/
Test PASSed.

traffic_ops/traffic_ops_golang/deliveryservice/deleteoldcerts.go
ocket8888: I agree that this is fine as-is

docs/source/api/cdns_id_snapshot.rst
ocket8888: A better solution than zero-width spaces is just adding to the list of terms defined by the "Delivery Service" glossary entry. right now we have something like:

```rst
.. glossary::

    Delivery Service
        A :dfn:`Delivery Service` is a thing that ....
```

... but the glossary would support:

```rst
.. glossary::

    Delivery Service
    Delivery Services
        A Delivery Service is a thing that....
```

Once that's done:

```rst
Both :term:`Delivery Service` and :term:`Delivery Services` link to the glossary appropriately.
```

The only reasons that isn't done already are because
1) I didn't know you could do that at the time I added the Delivery Service glossary entry
2) I haven't gotten around to changing it yet. Feel free to roll that into this PR if you want - though preferably as a separate commit.

docs/source/api/snapshot_name.rst
```diff
@@ -10,7 +10,7 @@
 .. distributed under the License is distributed on an "AS IS" BASIS,
 .. WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 .. See the License for the specific language governing permissions and
-.. limitations under the License.
```
ocket8888: Did you mean to do this?

docs/source/api/cdns_id_snapshot.rst
rob05c: Meh, I just reworded it to be singular.

docs/source/api/snapshot_name.rst
rob05c: Ah, no, good catch. Fixed.

docs/source/api/cdns_id_snapshot.rst
ocket8888: I don't blame you.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3274/
Test PASSed.

Issue #3312, opened 2019-02-08
Disallow underscore in xmlId


jhg03a:
The XMLID by default is part of the delivery service example URLs/regex.  We should add a constraint to disallow any characters forbidden in URLs such as `_`.  TrafficOps currently allows you do do this, but you will get back a 400 error back from TR whenever you use the example url generated by default.  Tangentially, this also helps #3267.

ocket8888:
To be maybe a bit more exhaustive, the TO API considers `aB&%=/#@$^*()[]!{}+\\|`~` to be a perfectly valid xml_id, and (I think) only the `a` ought to be allowed. Maybe `B` if we implement casefolding for SNI quirks in SSL implementations. Additionally, the TP validator catches none of that either, and will also let you submit a form with a `.` in it (though that fails because the API rejects it) but I do believe checks for `_`

jhg03a:
I ran into this using the TO API directly rather than TP.

If the XMLID is going to be a separate concept than what goes into the default host regex, we should fail to make a default regex as painful as that is and alert the user.  There's a high likelyhood the user would be just fine with an XMLID that is compliant with a more conservative validation pattern in my opinion.

mitchell852:
https://github.com/apache/trafficcontrol/blob/master/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html#L101

TP only checks that you don't have spaces in the xml id

`ng-pattern="/^\S*$/"`

a. it's great if the API is doing a tighter validation
b. i think it's fine to duplicate that validation in TP but seems like a lower priority as the api does it anyhow.

ocket8888:
No, you're right, that would be an enhancement to TP. It's a bug in TO.

And maybe this is a totally separate conversation, but what are we doing with `ng-pattern` that can't be done with [`pattern`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#pattern)?

jhg03a:
I'm not using TP and this happened.  It needs to be enforced in TO and/or Postgres.

ZugNZwang:
When making a `POST` request to `https://{{TO_BASE_URL}}/api/{{api_version}}/deliveryservices` with an `xmlId` key containing any of these:

- special character
- underscore 
- capital letters
- beginning/ending hyphens

it is allowed on the TO API. The TO API only handles spaces and periods.
Traffic Portal does not allow for any of these.
[...]

Issue #3202, opened 2019-01-10
Better error message when DS form is invalid


mitchell852:
For the 4 DS forms (HTTP*, DNS*, ANY_MAP*, STEERING*), there is a `Show Advanced` button that shows additional (mostly optional) fields for a DS. If there is an invalid entry in one of those fields, the `Update` button will be disabled but it is not clear why it is disabled.

Maybe some type of "This form has an error" message needs to be displayed at the top of the form when a field value is invalid.

mitchell852:
the form has changed since this issue was created

PR #3361, opened 2019-02-23
Added pattern validation for truly valid xmlids

ocket8888:
This PR partially addresses Issue #3312 by performing front-end validation on the XML_ID field of new, changed, or requested Delivery Services. Specifically, the validation checks for a whole-field match against the regular expression: `[a-z0-9][a-z0-9\-]*`. Input that doesn't match this *may still be submitted via the API and stored by the database*. The XML_ID is not always used as part of an <abbr title="Fully Qualified Domain Name">FQDN</abbr>, so such constraints on the server-side are better left until such a time that the future of the XML_ID - regarding whether it should be semantically distinct from a DNS label - is decided.

This PR also includes other quality-of-life improvements for Delivery Service forms.

* All `<label>`s now have an appropriate `for` attribute indicating which `<input>` (or `<select>`) it is actually labeling
* Changed many `ng-pattern` attributes to `pattern`, `ng-maxlength` attributes to `maxlength`.
* Form fields that allow the selection from among alternate `<options>` having a default, initial value of `Select...` will no longer include that value in the list of selectable options.
* ~Added a warning about hazy compatibility when attempting to set a new or existing Delivery Service's `geoProvider` to "Neustar".~ - this wound up being removed
* Fields that use enumerated names like e.g. `qstringIgnore` now have the integral value obscured, containing instead a human-readable description of the selected behavior. (They still pass the appropriate, integral value back to the API)
* Fields that previously allowed selecting either "True" or "False" now have more context-sensitive names - this removes a dependency on enumerating a hard-coded array (`[{"label": "false", "value": false}, {"label": "true", "value": true}]`) for every true/false field in the form. (Incoming PR to change those to checkboxes for sheer simplicity and effective communication of the represented idea)
* Fields that defined numbers and relied on `ng-pattern="/^\d+$/"` to ensure only natural numbers were entered now rely instead on the `min` and `step` attributes
* Latitude and Longitude fields now validate that the entered number is within the allowed range for each ([-90,90] and [-180,180], respectively)
* Fields that contained URLs have been updated to have `type="url"`
* `initalDispersion` is now an `<input type="number" min="1" max="10" step="1"/>` instead of a `<select>` iterating over a hard-coded Javascript array of those numbers.
* "More" menu delimiters are now `<hr/>`s instead of `<li class="divider"></li>`, as they are not actually items in the list but rather horizontal rules that separate them. (Incoming PR to change that `<ul>` to a `<menu>` and do the styling without classes on the `<hr/>` tags)

- [x] Traffic Portal

Run the Traffic Portal UI tests, build and run Traffic Portal and check out the Delivery Service creation, modification and request forms.

### DIFF
--- a/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
+++ b/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
@@ -100,10 +100,10 @@ under the License.
                     <span uib-popover-html="label('xmlId', 'desc')" popover-trigger="click" popover-placement="top" popover-append-to-body="true" popover-class="popover-class">{{label('xmlId', 'title')}}</span> *
                 </label>
                 <div class="col-md-10 col-sm-10 col-xs-12">
-                    <input id="xmlId" name="xmlId" type="text" class="form-control" placeholder="Unique id used for the delivery service" ng-model="deliveryService.xmlId" required maxlength="48" pattern="[a-z][a-z0-9\-]*" ng-readonly="(!settings.isRequest && !settings.isNew) || (settings.isRequest && changeType == 'update')">
+                    <input id="xmlId" name="xmlId" type="text" class="form-control" placeholder="Unique id used for the delivery service" ng-model="deliveryService.xmlId" required maxlength="48" pattern="[a-z0-9]([a-z0-9\-]*[a-z0-9])?" ng-readonly="(!settings.isRequest && !settings.isNew) || (settings.isRequest && changeType == 'update')">
                     <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.xmlId, 'required')">Required</small>
                     <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.xmlId, 'maxlength')">Too Long</small>
-                    <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.xmlId, 'pattern')">Must be a valid DNS label (no special characters, capital letters, periods, underscores, or spaces)</small>
+                    <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.xmlId, 'pattern')">Must be a valid DNS label (no special characters, capital letters, periods, underscores, or spaces and cannot begin or end with a hyphen)</small>
                     <small class="input-diff" ng-show="settings.isRequest && open() && deliveryService.xmlId != dsCurrent.xmlId">Current Value: [ {{dsCurrent.xmlId}} ]</small>
                     <span ng-show="hasError(deliveryServiceForm.xmlId)" class="form-control-feedback"><i class="fa fa-times"></i></span>
                 </div>
@@ -230,10 +230,10 @@ under the License.
                     </label>
                     <div class="col-md-10 col-sm-10 col-xs-12">
                         <small class="input-warning" ng-show="!settings.isNew && deliveryServiceForm.routingName.$dirty">Warning: Changing the routing name may require SSL certificates to be updated.</small>
-                        <input id="routingName" name="routingName" type="text" class="form-control" placeholder="Routing name used for the delivery service resulting in FQDN = <routing name>.<key>.<CDN domain>" ng-model="deliveryService.routingName" maxlength="48" pattern="^[a-z][a-z\-0-9]*$" required>
+                        <input id="routingName" name="routingName" type="text" class="form-control" placeholder="Routing name used for the delivery service resulting in FQDN = <routing name>.<key>.<CDN domain>" ng-model="deliveryService.routingName" maxlength="48" pattern="[a-z0-9]([a-z\-0-9]*[a-z0-9])?" required>
                         <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.routingName, 'required')">Required</small>
                         <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.routingName, 'maxlength')">Too Long</small>
-                        <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.routingName, 'pattern')">Must be a valid DNS label (no special characters, capital letters, periods, underscores, or spaces)</small>
+                        <small class="input-error" ng-show="hasPropertyError(deliveryServiceForm.routingName, 'pattern')">Must be a valid DNS label (no special characters, capital letters, periods, underscores, or spaces and cannot begin or end with a hyphen)</small>
                         <small class="input-diff" ng-show="settings.isRequest && open() && deliveryService.routingName != dsCurrent.routingName">Current Value: [ {{dsCurrent.routingName}} ]</small>
                         <span ng-show="hasError(deliveryServiceForm.routingName)" class="form-control-feedback"><i class="fa fa-times"></i></span>
                     </div>
@@ -653,10 +653,10 @@ under the License.
             </div>
 
             <div class="modal-footer">
-                <button class="btn btn-link" ng-click="advancedShowing = !advancedShowing"><span ng-show="!advancedShowing">Show</span><span ng-show="advancedShowing">Hide</span> Advanced</button>
+                <button type="button" class="btn btn-link" ng-click="advancedShowing = !advancedShowing"><span ng-show="!advancedShowing">Show</span><span ng-show="advancedShowing">Hide</span> Advanced</button>
                 <button type="button" class="btn btn-danger" ng-if="!settings.isNew" ng-disabled="!deletable()" ng-click="confirmDelete(deliveryService)">{{settings.deleteLabel}}</button>
-                <button type="button" class="btn btn-success" ng-disabled="deliveryServiceForm.$pristine || deliveryServiceForm.$invalid || !saveable()" ng-click="save(deliveryService)">{{settings.saveLabel}}</button>
-                <button type="button" class="btn btn-primary" ng-if="settings.isRequest && fulfillable()" ng-disabled="deliveryServiceForm.$invalid" ng-click="fulfillRequest(deliveryService)">Fulfill Request</button>
+                <button class="btn btn-success" ng-disabled="deliveryServiceForm.$pristine || deliveryServiceForm.$invalid || !saveable()" ng-click="save(deliveryService)">{{settings.saveLabel}}</button>
+                <button class="btn btn-primary" ng-if="settings.isRequest && fulfillable()" ng-disabled="deliveryServiceForm.$invalid" ng-click="fulfillRequest(deliveryService)">Fulfill Request</button>
             </div>
         </form>
     </div>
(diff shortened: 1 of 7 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3255/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3273/
Test PASSed.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.Steering.tpl.html
```diff
@@ -296,47 +308,46 @@
...
                 </div>
 
                 <div class="form-group" ng-class="{'has-error': hasError(deliveryServiceForm.infoUrl), 'has-feedback': hasError(deliveryServiceForm.infoUrl)}">
-                    <label class="control-label col-md-2 col-sm-2 col-xs-12">
+                    <label class="control-label col-md-2 col-sm-2 col-xs-12" for="infoUrl">
                         <span uib-popover-html="label('infoUrl', 'desc')" popover-trigger="click" popover-placement="top" popover-append-to-body="true" popover-class="popover-class">{{label('infoUrl', 'title')}}</span>
                     </label>
                     <div class="col-md-10 col-sm-10 col-xs-12">
-                        <input name="infoUrl" type="text" class="form-control" ng-model="deliveryService.infoUrl" ng-maxlength="255" autofocus>
+                        <input id="infoUrl" name="infoUrl" type="url" title="Must be a valid URL" class="form-control" ng-model="deliveryService.infoUrl" maxlength="255">
```
mitchell852: i could be wrong but i think people use this for non url info :)

mitchell852:
image

should routing name have the same validation as xml-id?

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.Steering.tpl.html
mitchell852: actually, i think you should just switch type back to text

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -163,403 +165,413 @@
...
-                    <label class="control-label col-md-2 col-sm-2 col-xs-12">
+                    <label class="control-label col-md-2 col-sm-2 col-xs-12" for="signingAlgorithm">
                         <span uib-popover-html="label('signingAlgorithm', 'desc')" popover-trigger="click" popover-placement="top" popover-append-to-body="true" popover-class="popover-class">{{label('signingAlgorithm', 'title')}}</span>
                     </label>
                     <div class="col-md-10 col-sm-10 col-xs-12">
-                        <select name="signingAlgorithm" class="form-control" ng-change="changeSigningAlgorithm(deliveryService.signingAlgorithm)" ng-model="deliveryService.signingAlgorithm" ng-options="sa.value as sa.label for sa in signingAlgos">
-                            <option value="">Select...</option>
+                        <select id="signingAlgorithm" name="signingAlgorithm" class="form-control" ng-change="changeSigningAlgorithm(deliveryService.signingAlgorithm)" ng-model="deliveryService.signingAlgorithm">
+                            <option ng-value="null">None</option>
+                            <option value="url_sig">URL Signtature Keys</option>
```
mitchell852: typo

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.Steering.tpl.html
ocket8888: The message in the `title` is meant to be sufficient. I could add a red text error underneath, though. Probably should, actually, for consistency's sake.

Are there really people who use the "Info URL" field for things that aren't URLs? With _three unlimited length description fields on the object_? That sounds like a horribly unintuitive misuse of the field.

ocket8888:
> _"should routing name have the same validation as xml-id? i mean, should it have the same error message?"_

yes, it absolutely should.

mitchell852:
is this true?

image

mitchell852:
I"m guessing you didn't change this but seems incorrect

image

image

Seems like when "None" it should hide the redirect url as well....

mitchell852:
do me a favor @ocket8888 - turn on ds requests in traffic_portal_properties.json and rebuild TP, then edit an existing ds (i.e. switch all the drop down values) which will create a ds request and then check out the ds request. your more intuitive dropdown values (i.e. active instead of true) makes the ds request comparison harder to comprehend imo.

image

ocket8888:
> _"Seems like when "None" it should hide the redirect url as well...."_
Yeah, I noticed that, and no I didn't change it. I suppose I can - and it should be changed - but there are a bunch of other fields like that. Actually, there's no reason for separate forms for DS types at all, because you could just hide the irrelevant fields based on the type. Because of that, to make the change in the best way possible is a big re-write; so I didn't do it.

> _"is this [Neustar support may have been dropped] true?"_

It's unclear. Neustar exists as an option, but I have yet to find anyone who uses it, or who even knows how to obtain the database - their website appears to be in a decrepit state of abandon.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.Steering.tpl.html
mitchell852: yeah if i look at our own production data, there are many values that are not a URL. so if you keep type=url, then it will disable the update button for a number of delivery services and it's hard to understand why the update button is disabled because of this - https://github.com/apache/trafficcontrol/issues/3202

mitchell852: and yes, there are 3 long descrip fields (don't get me started) but at comcast, those are intended for certain data (again, don't get me started).

ocket8888:
> _"your more intuitive dropdown values (i.e. active instead of true) makes the ds request comparison harder to comprehend imo."_

Counter position: those hints should use the more intuitive values. Why don't those "current value" prompts appear when editing a DS without a request? That is, I know why they aren't rendered, but why was the decision made not render them?

mitchell852:
> _"Counter position: those hints should use the more intuitive values. Why don't those "current value" prompts appear when editing a DS without a request? That is, I know why they aren't rendered, but why was the decision made not render them?"_

yep, i agree. those "current value" hints should be the more intuitive values. maybe that's pretty simple to do. let me see. 

also, i think that would be nice to show what has changed when simply editing a ds. not sure why that wasn't done. want to create an issue for that?

ocket8888:
Should be easy to manipulate the function `magicNumberLabel` that's already in use to label things like signing algorithm.

I'll create the issue.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3281/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3282/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3285/
Test PASSed.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.Steering.tpl.html
ocket8888: The three values I've seen in Comcast's production environment that are not either an empty string or a valid URL are "TBD", "NA" and a single JIRA ticket number. "TBD" and "NA" should be obliterated - they aren't valid URLs and give us no information. The JIRA ticket number is one of hundreds of linked JIRA tickets, but all of the others are actual URLs instead of just the number. Even if this field wasn't explicitly for URLs, that should be changed for consistency and was likely an oversight/mistake in the first place.

I've also added a message like the other fields have that indicates when the entered value is not a valid URL (although I can't help but interject that validation of constant constraints is best left to the user-agent). The angular attributes of the "advanced" section can be tweaked such that it's open by default when any of the fields have errors. We could even go so far as to disable the "Hide Advanced" button in that case. I suppose I could roll that into this PR, or just make a note on that issue and leave it for another PR. Doesn't matter much to me which one, provided that's an acceptable solution.

elsloo:
> is this true? @elsloo - do you know?
> 
> image

As far as I know and can tell, Neustar support should still be present in Traffic Router. I don't believe we have removed support.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.Steering.tpl.html
mitchell852: expanding the advanced section when there is a form validation error sounds like a good idea but i'm not sure about disabling the hide button (i hate making those forms too complicated) but, anyhow, let's save it for another PR.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3287/
Test PASSed.

ocket8888:
I don't think we _intentionally_ removed support, but it's gone untested for ages, and you're the first person I've ever heard say with any confidence that it should still work. I've removed the warning from the form, but note that the documentation for e.g. DS-related API endpoints has similar warnings. If Neustar is unambiguously supported, we should open an issue to get those warnings removed.

mitchell852:
@ocket8888 - fyi

image

ocket8888:
Ahh, I was limiting it to the traditional, RFC952 restrictions (summarized [here](https://tools.ietf.org/html/rfc1035#section-2.3.1)). Didn't realize that as of [RFC1123](https://tools.ietf.org/html/rfc1123#section-2) the restriction on the first character of a DNS label was relaxed to allow numerics.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3292/
Test PASSed.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -653,10 +653,10 @@
             </div>
 
             <div class="modal-footer">
```
mitchell852: what changed here exactly? looks like you added a type=button on one button and removed it on 2 others?

ocket8888: `type="button"` specifies that the `<button>` has some Javascript action that may or may not be related in any way to the `<form>` containing it. The default `type` for a `<button>` is `type="submit"` meaning that the `<button>` is responsible for submitting it's ancestor `<form>`. The upshot is that the "show/hide advanced" `<button>` would perform form validation, which it shouldn't do because you could want to see advanced features before filling out the basic stuff, and the "create"/"update" and "fulfill request" `<button>`s were not performing validation. (Because the `<form>` has no `action`, allowing `<button>`s to submit the `<form>` has no real consequence beyond input validation. And really that shouldn't matter anyway because the submit buttons are disabled until the `<form>` has no Angular errors, so on validation everything should always check out.)

mitchell852: oh, i see. makes sense.

PR #3366, opened 2019-02-27
Bugfix for self-signed SSL certs

hbeatty:

The get_cookie subroutine was not using the same connection as the rest of the ORT script. This fixes that. A side effect is that it also fixes self-signed certs to work again. I currently need self-signed certs to work because I use those for lab testing.

I do think that we should not make that the default, which the ORT script currently does. I think we should add an option to the ORT script for self-signed certs like curl's -k.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [x] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Best way to test: Run this ORT script against a TO that has a self-signed cert.

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/bin/traffic_ops_ort.pl
+++ b/traffic_ops/bin/traffic_ops_ort.pl
@@ -1628,8 +1628,7 @@ sub get_cookie {
 
 	my $url = $to_host . "/api/1.3/user/login";
 	my $json = qq/{ "u": "$u", "p": "$p"}/;
-	my $lwp = LWP::UserAgent->new;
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3268/
Test PASSed.

PR #3367, opened 2019-02-28
astats increase file buffer size, fix compiler warnings

traeak:

Looking for a reporting bug, determined that the astats buffer for scanning /proc/net/dev is too small, especially for a test host that is running docker with many virtual network interfaces.  Also cleaned up a (valid) compiler warning over a signed/vs unsigned check regarding snprintf.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Other *astats_over_http plugin*

Compile and install onto ATS, run curl on localhost

```
curl -Lv  "http://ats:8080/_astats?application=system&inf.name=<iface>"
```

ensure *proc.net.dev* shows up in the *system:* section

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_server/plugins/astats_over_http/astats_over_http.c
+++ b/traffic_server/plugins/astats_over_http/astats_over_http.c
@@ -220,8 +220,9 @@ stats_process_read(TSCont contp, TSEvent event, stats_state *my_state) {
 #define APPEND(a) my_state->output_bytes += stats_add_data_to_resp_buffer(a, my_state)
 #define APPEND_STAT(a, fmt, v) do { \
 		char b[3048]; \
-		if (snprintf(b, sizeof(b), "   \"%s\": " fmt ",\n", a, v) < sizeof(b)) \
-		APPEND(b); \
+		int nbytes = snprintf(b, sizeof(b), "   \"%s\": " fmt ",\n", a, v); \
+		if (0 < nbytes && nbytes < (int)sizeof(b)) \
+			APPEND(b); \
 } while(0)
 
 static void
@@ -298,19 +299,18 @@ static int getSpeed(char *inf, char *buffer, int bufferSize) {
 
 static void appendSystemState(stats_state *my_state) {
 	char *interface = my_state->interfaceName;
-	char buffer[2024];
-	int bsize = 2024;
+	char buffer[16384];
 	char *str;
 	char *end;
 	int speed = 0;
 
 	APPEND_STAT("inf.name", "\"%s\"", interface);
 
-	speed = getSpeed(interface, buffer, bsize);
+	speed = getSpeed(interface, buffer, sizeof(buffer));
 
 	APPEND_STAT("inf.speed", "%d", speed);
 
-	str = getFile("/proc/net/dev", buffer, bsize);
+	str = getFile("/proc/net/dev", buffer, sizeof(buffer));
 	if (str && interface) {
 		str = strstr(str, interface);
 		if (str) {
@@ -321,7 +321,7 @@ static void appendSystemState(stats_state *my_state) {
 		}
 	}
 
-	str = getFile("/proc/loadavg", buffer, bsize);
+	str = getFile("/proc/loadavg", buffer, sizeof(buffer));
 	if (str) {
 		end = strstr(str, "\n");
 		if (end)


asfgit:

traffic_server/plugins/astats_over_http/Makefile.tsxs
```diff
@@ -0,0 +1,27 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+TSXS?=tsxs
+PLUGIN=astats_over_http
+
+all:	$(PLUGIN).c
```
ocket8888: `make` is for exactly what it sounds like: making things. When you write a rule like that, you're saying to `make` "This is a recipe that makes a file named 'all'. It requires `$(PLUGIN).c` and these are the steps in the recipe:"

When you make a rule like this, you're sort of losing the major advantages of a Makefile. Since this recipe only creates the plugin file, you should tell `make` the truth - that this recipe creates the plugin.

```make
# You could also do `$(shell which tsxs)` if you're worried about it not being in `$PATH`
TSXS?=tsxs

# `:=` sets the variable for calculation once - syntactically meaningless for a constant,
# but for things like `SOURCE` that *aren't* constant it calculates it once as opposed to
# every time the variable is expanded.
PLUGIN := astats_over_http.so
SOURCE := $(PLUGIN:.so=.c) #name substitution shortcut - technically GNU-only I think

# '$@' expands to the recipe's target; as for '$<' you can use '$?' instead in this case,
# I just prefer '$<'.
$(PLUGIN): $(SOURCE)
    $(TSXS) -v -o $@ $<
```

Now, obviously you still want to be able to do `make && make install` and `make clean`. The recipe that gets run when you call `make` with no arguments is simply the first one specified in the file; `all` has no special meaning to `make`. So already you've got one out of three things. Of course, the install process doesn't create a local file named `install`, but you still want to be able to do `make install`. When a target of a rule isn't created by that rule, you should add it as a dependency of the special `.PHONY` target like this:

```make
.PHONY: install clean

install: $(PLUGIN)
    $(TSXS) -i -o $<

# Using `$(RM)` is recommended for POSIX compliance - it will always
# expand to the path to a binary for removing files forcibly (typically `/bin/rm -f`)
clean:
    $(RM) $(PLUGIN)

# Explicitly removing `$(PLUGIN)` avoids destroying objects people may have
# created from other sources.
```

traffic_server/plugins/astats_over_http/astats_over_http.c
```diff
@@ -298,19 +299,18 @@ static int getSpeed(char *inf, char *buffer, int bufferSize) {
 
 static void appendSystemState(stats_state *my_state) {
 	char *interface = my_state->interfaceName;
-	char buffer[2024];
-	int bsize = 2024;
+	char buffer[16384];
```
ocket8888: What's the motivation for this? What problems were the old buffer size causing?

traffic_server/plugins/astats_over_http/Makefile.tsxs
traeak: env TSXS=~/opt/ats8/bin/tsxs make -f Makefile.tsxs
env TSXS=~/opt/ats7/bin/tsxs make -f Makefile.tsxs

This make file is a ripoff of other ones with a slight change.  You could argue that one could just invoke it via command line directly

traffic_server/plugins/astats_over_http/astats_over_http.c
traeak:
```
Inter-|   Receive                                                |  Transmit
 face |bytes    packets errs drop fifo frame compressed multicast|bytes    packets errs drop fifo colls carrier compressed
br-5e66a6784dcd:       0       0    0    0    0     0          0         0      928      12    0    0    0     0       0          0
   em3:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
br-09a0dde5b54a:  244822    8694    0    0    0     0          0         0    58191     542    0    0    0     0       0          0
virbr0:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
br-4d9440de4eba:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
docker0: 77867415 1412703    0    0    0     0          0         0 11219225352 2039122    0    0    0     0       0          0
  p4p1: 17166817434005 2351818698    0    2    0     0          0  14184099 109452120652 879661558    0    0    0     0       0          0
br-cdfbca58ca33:       0       0    0    0    0     0          0         0     5276      60    0    0    0     0       0          0
br-3f46f2f9970a:       0       0    0    0    0     0          0         0     8478     179    0    0    0     0       0          0
virbr0-nic:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
br-12e7ed6b66f3: 2271911   21947    0    0    0     0          0         0 151217170   24506    0    0    0     0       0          0
   em1:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
    lo: 366481874375030 12180805622    0    0    0     0          0         0 366481874375030 12180805622    0    0    0     0       0          0
br-96a55ab3fc5d:  185022    3546    0    0    0     0          0         0 29023485    4014    0    0    0     0       0          0
 bond0: 34395814195367 4714290784    0    0    0     0          0  28368240 221007483586 1752861461    0   12    0     0       0          0
  p4p2: 17228996761982 2362472091    0    3    0     0          0  14184141 111555363182 873199905    0    0    0     0       0          0
   em4:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
br-c61aab48073a: 49841107   90471    0    0    0     0          0         0 31814390  135500    0    0    0     0       0          0
   em2:       0       0    0    0    0     0          0         0        0       0    0    0    0     0       0          0
```

ocket8888: I assume you mean because the interface you're looking for is far enough down in the file that it's outside the length of the buffer, e.g. `em2`.

looking at the plugin itself, I'm not convinced that simply increasing the buffer size actually fixes the problem; it merely mitigates it. The file should be read in in buffer-sized chunks until EOF, not just read in up to the buffer size and the rest discarded. For as long as we read the file improperly, it will be possible for `/proc/net/dev`s to exist that defy interface detection.

traffic_server/plugins/astats_over_http/Makefile.tsxs
ocket8888: I'm not really sure what to make of those first two lines. Of course anything that can be produced by a Makefile could also be produced with the recipe's direct command line steps, but if a Makefile is provided it should be a good one.

traffic_server/plugins/astats_over_http/astats_over_http.c
traeak: Mitigation is fine.  The plugin is so full of holes regarding unchecked conditions that it needs a general review.

PR #3372, opened 2019-03-05
CIAB: remove tccache

Shihta:

This PR let mid and edge refer to the same Dockerfile and remove the tccache container.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Other CIAB

1. Start CIAB
2. check edge and mid status. they should work correctly

## Check all that apply

- [ ] This PR includes tests


asfgit:

infrastructure/cdn-in-a-box/docker-compose.yml
```diff
@@ -209,48 +209,36 @@ services:
...
+      context: ../..
+      dockerfile: infrastructure/cdn-in-a-box/cache/Dockerfile
     domainname: infra.ciab.test
     env_file:
       - variables.env
     hostname: edge
     volumes:
       - ./dns/set-dns.sh:/usr/local/sbin/set-dns.sh
       - ./dns/insert-self-into-dns.sh:/usr/local/sbin/insert-self-into-dns.sh
+      - ./edge/run.sh:/opt/init.d/99-run.sh
```
ocket8888: Why put this in `/opt/init.d` instead of just `/`? It seems like you're setting this up to run a bunch of scripts in a row, but there's only the one. If this was just added to container so that the `CMD` called it directly you could get rid of the whole `cache/run.sh` file you made to shell out calls to the real run script.

infrastructure/cdn-in-a-box/docker-compose.yml
```diff
@@ -209,48 +209,36 @@ services:
...
-    image: tccache:latest
-
   # edge cache
   edge:
     privileged: True
     build:
-      context: .
-      dockerfile: edge/Dockerfile
-    depends_on:
-      - enroller
```
ocket8888: Shouldn't this still depend on the enroller?

Shihta: It's not necessary. Dependency is used to guarantee the startup sequence of container, not the services inside.
You can take a look at:
`infrastructure/cdn-in-a-box/dns/set-dns.sh`: waiting for dns
`to-enroll` of `infrastructure/cdn-in-a-box/traffic_ops/to-access.sh`: waiting for the services enrollment

infrastructure/cdn-in-a-box/docker-compose.yml
Shihta: I want to let this container could be launched without any other injection.
That why I add another `run.sh` to drive another injected startup script.
If we add `CMD /run.sh` but not add a real `run.sh` in the container, we can't run `docker run cdn-in-a-box_edge` directly.
How do you think?

ocket8888: Well, I think normally the way we'd do something like that is with a build argument using `ARG`

Shihta: It's different. If you build different images through different ARG, that means the built images will be different. I mean the image id will be different.
What I want to do is to keep to use the only image. The role it plays is dependent on runtime parameters (file(s) be put in `/opt/init.d`)

infrastructure/cdn-in-a-box/docker-compose.yml
ocket8888: I guess with the cert stuff there's enough going on at the beginning that you don't notice much that the enroller isn't up yet.

Shihta: Both edge and mid wait the cert stuff in `run.sh`
[edge/run.sh](https://github.com/apache/trafficcontrol/blob/master/infrastructure/cdn-in-a-box/edge/run.sh#L30)
[mid/run.sh](https://github.com/apache/trafficcontrol/blob/master/infrastructure/cdn-in-a-box/mid/run.sh#L30)

ocket8888: yeah, exactly.

infrastructure/cdn-in-a-box/docker-compose.yml
ocket8888: Yeah, that's true, they would be different images if we used `ARG`. I guess I just balk at the idea of running a bunch of things in one container, and this sort of seems to be moving in that direction. I think a volume is fine implementation, I just hope this doesn't turn into an orchestration system for a highly complex container.

Shihta: I don't quite understand what you're saying `a volume is fine implementation`. Could you give me a hint?

ocket8888: Nothing, it's fine

PR #3377, opened 2019-03-05
Quality of life improvements to consistent hash regex test form

ocket8888:
Changes the `<input>` element used to display results into the
more appropriate `<output>` element. Also applies proper `for`
attributes to all labels, corrected a broken pattern, changed
angular validation restrictions into their standard counterparts
(angular not needed when the values are constant).

- [x] Traffic Portal

Check out the "Test Regex" page for consistent-hash regular expressions on a Delivery Service

## Check all that apply
N/A

### DIFF
--- a/traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
+++ b/traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
@@ -32,13 +32,13 @@ under the License.
             <div class="form-group" ng-class="{'has-error': hasError(dsConsistentHashRegexForm.pattern), 'has-feedback': hasError(dsConsistentHashRegexForm.pattern)}">
                 <label for="pattern" class="control-label col-md-2 col-sm-2 col-xs-12">Pattern</label>
                 <div class="col-md-10 col-sm-10 col-xs-12">
-                    <input id="pattern" name="pattern" type="text" class="form-control" ng-model="pattern" maxlength="48" required autofocus>
+                    <input id="pattern" name="pattern" type="text" class="form-control" ng-model="pattern" maxlength="48" autofocus>
                 </div>
             </div>
             <div class="form-group" ng-class="{'has-error': hasError(dsConsistentHashRegexForm.requestPath), 'has-feedback': hasError(dsConsistentHashRegexForm.requestPath)}">
                 <label class="control-label col-md-2 col-sm-2 col-xs-12" for="requestPath">Request Path</label>
                 <div class="col-md-10 col-sm-10 col-xs-12">
-                    <input id="requestPath" name="requestPath" type="text" class="form-control" ng-model="requestPath" maxlength="27" pattern="/.*" required>
+                    <input id="requestPath" name="requestPath" type="text" class="form-control" ng-model="requestPath" maxlength="27" pattern="/.*">
                     <small class="input-error" ng-show="hasPropertyError(dsConsistentHashRegexForm.requestPath, 'pattern')">Path Must Start With '/'</small>
                     <span ng-show="hasError(dsConsistentHashRegexForm.requestPath)" class="form-control-feedback"><i class="fa fa-times"></i></span>
                 </div>
@@ -52,7 +52,7 @@ under the License.
             </div>
 
             <div class="modal-footer">
-                <button class="btn btn-success" ng-disabled="dsConsistentHashRegexForm.$invalid" ng-click="test(pattern, requestPath, deliveryService.cdnId)">Test</button>
+                <button class="btn btn-success" ng-click="test(pattern, requestPath, deliveryService.cdnId)">Test</button>
             </div>
         </form>
     </div>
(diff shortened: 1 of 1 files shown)


traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
```diff
@@ -30,25 +30,24 @@
         <br>
         <form name="dsConsistentHashRegexForm" class="form-horizontal form-label-left" novalidate>
             <div class="form-group" ng-class="{'has-error': hasError(dsConsistentHashRegexForm.pattern), 'has-feedback': hasError(dsConsistentHashRegexForm.pattern)}">
-                <label class="control-label col-md-2 col-sm-2 col-xs-12">Pattern *</label>
```
mitchell852: isn't pattern required? why remove the *?

ocket8888: > _"isn't pattern required?"_

I don't think so. Using the API I was able to send a regex containing whitespace and get a 200 response from the test, so even if that may not be the best way to construct a URL it's a syntactically valid input.

> _"why remove the `*`?"_

The `*` just means "This is required", which is best handled by the `required` attribute. Even if we didn't disable the submitting `<button>` until such requirements were met, the user-agent would refuse to submit the `<form>` if all `required` fields aren't filled out. This form _does_ seem to be missing the usual "Required" error messages in `<small class="input-error">` tags that the other forms have. I could add those in if you like.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3298/
Test PASSed.

traffic_portal/app/src/common/modules/form/deliveryServiceConsistentHashRegex/form.deliveryServiceConsistentHashRegex.tpl.html
mitchell852: yeah, can you add the 

`<small class="input-error">`

also, i just think if the form field has the required attribute, it should also have the `*` in the label to provide the visual cue to the user. i can't say with 100% certainty that adding a `*` is the proper web standard but it seems to be a pretty common thing and that's how these all the forms are.

also, if pattern isn't truly required in the api, i guess you can remove both the `*` and the `required` attribute, however, that makes the form a little error prone imo. wonder why it wasn't required in the api.

mitchell852: long story short, i'm just saying if the form field has `required` in it, the label should have `*` in it

ocket8888: > _"...  if pattern isn't truly required in the api ..."_

No, it is, I misunderstood. I thought you were referring to the fact that I deleted the `ng-pattern` attribute. The input is still `required`.

> _"i just think if the form field has the required attribute, it should also have the * in the label to provide the visual cue to the user."_

But doesn't the `<small>` error do exactly that? If we're really going to stick with a `*` on every required input, it should IMO be a part of the presentation, not the content. Meaning, the label for the pattern should say exactly `Pattern` because that's what it is. It's not a `Pattern *`, whatever that is. Instead, a CSS rule like `input:required::before { content: " *"; color: inherit }` should be used so that visually it appears but isn't a part of the label itself.

But a change like that falls outside the scope of the PR, so if you really don't want me get rid of it I'll put it back and leave a change like that for future changes.

ocket8888: Actually, I take that back. The pattern isn't required - the only field the API actually requires is `"cdnId"`.

```http
POST /api/1.5/consistenthash HTTP/1.1
Host: trafficops.infra.ciab.test
User-Agent: curl/7.47.0
Accept: */*
Cookie: mojolicious=...
Content-Length: 11

{"cdnId":2}

HTTP/1.1 200 OK
Set-Cookie: mojolicious=...; Path=/; HttpOnly
Whole-Content-Sha512: L/eE7wg1BHCLOxokLLkoduNo9RKuWILH0fKjZ9K4x9oEbrS3ZTYvfsJPsRA7yjdJ69mmromk4JI7kIjA0HWjUg==
Date: Wed, 06 Mar 2019 18:12:03 GMT
Content-Length: 91

{
	"response": {
		"resultingPathToConsistentHash": "",
		"consistentHashRegex": "",
		"requestPath": ""
	}
}
```

mitchell852: i like the css approach. so can you put the `*` back in the label for now and create an issue for the future work of moving the `*` into a css declaration?

ocket8888: Well actually I removed the `*` because none of those fields are required by the API :P

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3305/
Test FAILed.

mitchell852:
retest this please

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3353/
Test PASSed.

PR #3379, opened 2019-03-05
Static dns entry example documentation

### LINKED ISSUE #3374: Document Static DNS Entries Example
We should create a Static DNS Entries page in the How To Guides, explaining how to create and use them.

Example:

In Traffic Portal
1. Navigate to the Delivery Service’s page.
2. Click the “More” button, in the top-right.
3. In the More dropdown, Click “View Static DNS Entries”
4. On the Static DNS Entries page, click the “+” button in the top-right
5. You should now be at the Static DNS Entries creation form, with the following fields:

**Host** the sub-domain to be served at that Delivery Service’s domain. For the CiaB DS, this is `video.demo1.mycdn.ciab.test`. For example, a host of `foo` will create the DNS record `foo.video.demo1.mycdn.ciab.test`.
**Type** the DNS Record type - A, AAAA, CNAME, and TXT are supported.
**TTL** the TTL of the DNS Record.
**Address** the value of the DNS record to serve.

6. After filling out the form and clicking “Create”, Snapshot the CDN.
7. Once the Snapshot propagates to the Router, from within a container on the CiaB Docker Network, you should be able to request the record you created. For example:
```
$ docker exec cdn-in-a-box_trafficvault_1 dig foo.video.demo1.mycdn.ciab.test

;; ANSWER SECTION:
foo.video.demo1.mycdn.ciab.test. 30 IN      A   192.0.2.1.
```

ocket8888:
Resolves #3374 

Also changes `View X` to a more appropriate `Manage X` in Delivery Service "More" drop-down menus (except for the "View Charts" option)

Also quality-of-life updates to the static DNS entry form including
* `<label>`s now have an appropriate `for` attribute indicating what it is they label
* only one thing on a page can be `autofocus`ed
* numeric inputs properly utilize `min` and `step` instead of `ng-pattern`
* The `Select...` placeholder option is no longer selectable in the option list

- [x] Documentation
- [x] Traffic Portal

Build the documentation and read it to make sure it makes sense, ideally by actually following the outlined procedure.

## Check all that apply


docs/source/admin/quick_howto/static_dns.rst
```diff
@@ -0,0 +1,77 @@
...
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+.. _static-dns-qht:
+
+******************************
+Configuring Static DNS Entries
+******************************
+Any number static DNS entries can be configured on a per-:term:`Delivery Service` basis. In a typical scenario, the :term:`Delivery Service` will have DNS entries automatically generated based on its "xml_id" and "routing name", and the name and subdomain of the CDN to which it belongs. For example, in the :ref:`ciab` default environment, the "demo1" :term:`Delivery Service` has an automatically generated DNS record for ``video.demo1.mycdn.ciab.test``. Configuring a static DNS entry allows for further extension of this, for example, one could create an entry that enforces lookups of the name ``foo.video.demo1.mycdn.ciab.test`` resolve to the IPv4 address ``0.0.0.1``.
```
rawlinp: Example IP addresses should probably be 10.* or 192.168.*

`foo.video.demo1.mycdn.ciab.test` would probably make more sense as `foo.demo1.mycdn.ciab.test` so that it doesn't look like `video` is a subdomain (since `video` is actually the routing name).

"one could create an entry" -> "one could create an A record"

docs/source/admin/quick_howto/static_dns.rst
```diff
@@ -0,0 +1,77 @@
...
+		:alt: The static DNS entries table page
+		:align: center
+
+		The Static DNS Entries Table Page
+
+#. Click on the blue :guilabel:`+` button to add a new static DNS Entry
+#. Fill in all of the fields.
+
+	Host
+		This is the lowest-level DNS label that will be used in the DNS record. In the :ref:`ciab` scenario, for example, entering ``foo`` here will result in a full DNS name of ``foo.video.mycdn.ciab.test``.
```
rawlinp: I think you mean `foo.demo1.mycdn.ciab.test`

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3294/
Test PASSed.

docs/source/admin/quick_howto/static_dns.rst
ocket8888: yes I do

docs/source/admin/quick_howto/static_dns.rst
ocket8888: > _"Example IP addresses should probably be 10.* or 192.168.*"_

Why?

> _"foo.video.demo1.mycdn.ciab.test would probably make more sense as foo.demo1.mycdn.ciab.test so that it doesn't look like video is a subdomain (since video is actually the routing name)."_

seems I misunderstood how that works...

> _"'one could create an entry' -> 'one could create an A record'"_

Then, in general "entry" -> "record"? Semantically, what's the difference between a DNS "entry" and a DNS "record"? Does the former ever exist or is it just something we're making up?

rob05c: Example IPs should be `192.0.2.*` and `2001:DB8::*`. 

Because 
https://tools.ietf.org/html/rfc5737 
https://tools.ietf.org/html/rfc3849

ocket8888: > _"This document is not an Internet Standards Track specification; it is
   published for informational purposes."_

Not that it matters to me what the example IP is, but this doesn't seem to be a real standard. And where does that leave `10.*`?

rob05c: >this doesn't seem to be a real standard

Many RFCs are not Internet Standards Track. These RFCs are authoritative, they reserve these IP subnets, and they are not to be used or assigned on the internet.

Moreover, the IANA is authoritative:
https://www.iana.org/assignments/iana-ipv4-special-registry/iana-ipv4-special-registry.xhtml
https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml

> where does that leave 10.*?

`10.*` is reserved for internal subnets. See https://tools.ietf.org/html/rfc1918. We shouldn't use it for documentation, RFC5737 and 3849 authoritatively specify the correct subnets for docs.

rob05c: >Semantically, what's the difference between a DNS "entry" and a DNS "record"? Does the former ever exist or is it just something we're making up?

"Record" is the canonical term. Older DNS RFCs occasionally use the term "entry," and other people do use it, not just us. But "record" is the canonically correct term.

See:
https://tools.ietf.org/html/rfc1035#section-4.1.3
https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-4
https://en.wikipedia.org/wiki/List_of_DNS_record_types

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3295/
Test FAILed.

docs/source/admin/quick_howto/static_dns.rst
ocket8888: Neat. I'll leave the page name, menu option name, and docs section name alone so that they all match the API route, but I'll prefer "record" to "entry" elsewhere.

ocket8888:
```ERROR: for weasel  Cannot create container for service weasel: open /var/lib/docker/aufs/layers/d1b2bc5e3c6b0bc0b8ce1f5ec1dd91979006103fcf41aa0c4042154152a9733d: no such file or directory```

looks like a driver failure, probably won't repeat if tested again

**Comment by asfgit** [comment]
https://builds.apache.org/job/trafficcontrol-PR/3296/
Test FAILed.

**Comment by asfgit** [comment]
https://builds.apache.org/job/trafficcontrol-PR/3297/
Test PASSed.

File: traffic_portal/app/src/common/modules/form/deliveryServiceStaticDnsEntry/form.deliveryServiceStaticDnsEntry.tpl.html [hunk 3]
```diff
@@ -31,42 +31,42 @@
         <br>
         <form name="dsStaticDnsEntryForm" class="form-horizontal form-label-left" novalidate>
             <div class="form-group" ng-class="{'has-error': hasError(dsStaticDnsEntryForm.host), 'has-feedback': hasError(dsStaticDnsEntryForm.host)}">
-                <label class="control-label col-md-2 col-sm-2 col-xs-12">Host *</label>
```
mitchell852: why removing the * on required fields?

docs/source/admin/quick_howto/static_dns.rst
```diff
@@ -0,0 +1,77 @@
...
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+.. _static-dns-qht:
+
+******************************
+Configuring Static DNS Entries
+******************************
+Any number static DNS records (historically "entries") can be configured on a per-:term:`Delivery Service` basis. In a typical scenario, the :term:`Delivery Service` will have DNS records automatically generated based on its "xml_id" and "routing name", and the name and subdomain of the CDN to which it belongs. For example, in the :ref:`ciab` default environment, the "demo1" :term:`Delivery Service` has an automatically generated DNS record for ``video.demo1.mycdn.ciab.test``. Configuring a static DNS record allows for further extension of this, for example, one could create an ``A`` record that enforces lookups of the name ``foo.demo1.mycdn.ciab.test`` resolve to the IPv4 address ``192.0.2.0``.
```
rawlinp: The `(historically "entries")` bit seems superfluous. I'd also remove the `Any number` part and add `within the Delivery Service's subdomain` to the end of the first sentence.

Also, `*.0` addresses are typically reserved as the "network address", so a better example would probably be something between .1 and .254 (.255 is typically reserved as the "broadcast address").

docs/source/admin/quick_howto/static_dns.rst
```diff
@@ -0,0 +1,77 @@
...
+
+.. note:: It's not possible to alter higher levels of the DNS name using this method. That is, one could **not** create a rule for routing the name ``foo.bar.mycdn.ciab.test`` on the :ref:`ciab` :term:`Delivery Service` "demo1".
```
rawlinp: I would reword this as `It's only possible to create static DNS records within a Delivery Service's subdomain. That is, one could not create an A record for foo.bar.mycdn.ciab.test on the ciab Delivery Service "demo1", since demo1's subdomain is demo1.mycdn.ciab.test.

traffic_portal/app/src/common/modules/form/deliveryServiceStaticDnsEntry/form.deliveryServiceStaticDnsEntry.tpl.html
ocket8888: Because they're already required by virtue of the `required` attribute, as well as the big red "Required" we put under them when they aren't filled in. If the `<form>`s submitting `<button>` were not disabled until these requirements were met, the user-agent would also refuse to submit the form with an appropriate message (e.g. "Please fill out this field" in Firefox - I think Chrome says "This field is required").

rawlinp: yeah the `*` seems to be common throughout TP for required fields

docs/source/admin/quick_howto/static_dns.rst
rawlinp: > 10.* is reserved for internal subnets. See https://tools.ietf.org/html/rfc1918. We shouldn't use it for documentation, RFC5737 and 3849 authoritatively specify the correct subnets for docs.

Yeah, that's right. I was thinking rfc1918 would be better than 0.0.0.1 but was also too lazy to look up the rfc for the recommended documentation CIDRs. Thanks for doing that!

docs/source/admin/quick_howto/static_dns.rst
ocket8888: Thanks, I struggled with that wording.

docs/source/admin/quick_howto/static_dns.rst
ocket8888: ¯\\\_(ツ)\_/¯ if you say so.

I'd like to leave in the `(historically "entries")` bit because otherwise it could be confusing that the reader is in the "Configuring Static DNS **Entries**" section but it just keeps talking about **records**. I think the connection is important to make - after all _I_ didn't know they were the same thing so it seems unfair of me to expect others to know.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3299/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3300/
Test FAILed.

ocket8888:
> _"Note there is also a small amount of static dns entry documentation [here](https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_ops/using.html#static-dns-entries). That should probably link to the quick_howto"_

Disagree. It should be removed as it relates to a UI that no longer exists.

> _"and [this](https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_portal/usingtrafficportal.html#delivery-service) should probably link there as well."_

Agree

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3301/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3302/
Test PASSed.

rawlinp:
>> "Note there is also a small amount of static dns entry documentation here. That should probably link to the quick_howto"

> Disagree. It should be removed as it relates to a UI that no longer exists.

For someone familiar with the legacy UI and finding information about delivery services on that page, it could be useful to link to "how to do this in Traffic Portal now". That page is a hodgepodge of information that doesn't only apply to the old UI. Most of the information there is pertinent regardless of instructions about the old UI.

ocket8888:
> _"For someone familiar with the legacy UI and finding information about delivery services on that page, it could be useful to link to "how to do this in Traffic Portal now". That page is a hodgepodge of information that doesn't only apply to the old UI. Most of the information there is pertinent regardless of instructions about the old UI."_

It certainly is, and really ought to be properly disseminated and expanded upon (I'm working on it, but it's slow going). The idea being that section will evaporate down to instructions for actually using Traffic Ops - functionality that doesn't exist shouldn't be documented. The Administration -> Traffic Ops -> Using Traffic Ops section also does already come with a disclaimer warning that readers should look elsewhere for a UI to use when interacting with ATC.

docs/source/admin/quick_howto/static_dns.rst
```diff
@@ -0,0 +1,77 @@
...
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+.. _static-dns-qht:
+
+******************************
+Configuring Static DNS Entries
+******************************
+Any number static DNS records (historically "entries") can be configured on a per-:term:`Delivery Service` basis. In a typical scenario, the :term:`Delivery Service` will have DNS records automatically generated based on its "xml_id" and "routing name", and the name and subdomain of the CDN to which it belongs. For example, in the :ref:`ciab` default environment, the "demo1" :term:`Delivery Service` has an automatically generated DNS record for ``video.demo1.mycdn.ciab.test``. Configuring a static DNS record allows for further extension of this, for example, one could create an ``A`` record that enforces lookups of the name ``foo.demo1.mycdn.ciab.test`` resolve to the IPv4 address ``192.0.2.1``.
```
rawlinp: Ok, if you'd like to keep the historical bit that's fine by me, but I'd still like my other suggestions.

traffic_portal/app/src/common/modules/form/deliveryServiceStaticDnsEntry/form.deliveryServiceStaticDnsEntry.tpl.html
rawlinp: I think I could be on board with removing the `*` for required fields. However, these changes are conflating the PR which should generally just be a single related change.

I think at one point the required fields weren't outlined in red when you opened the page, so having the `*` was helpful from a UI perspective. Does that sound right @mitchell852? Since it now has a better visual indication of required fields I'm inclined to agree with them being unnecessary now.

docs/source/admin/quick_howto/static_dns.rst
ocket8888: Simply appending `within the Delivery Service's subdomain` to the end seemed awkward, so after removing the `Any number of` part I reworded the end a bit. Does it look okay now?

traffic_portal/app/src/common/modules/form/deliveryServiceStaticDnsEntry/form.deliveryServiceStaticDnsEntry.tpl.html
ocket8888: No, I agree, any changes to those should be uniformly applied across TP forms - and thus doesn't belong in this PR. I've just been waiting for a conversation on a separate PR where I'd done the same thing to come to a conclusion before I did anything on this one.

FYI an extended discussion about the future of ` *` could be had on Issue #3381 that I opened to address my thoughts on the matter.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3304/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3307/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3308/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3310/
Test PASSed.

docs/source/admin/quick_howto/static_dns.rst
rawlinp: Yep, thanks

PR #3380, opened 2019-03-06
Add SSL Certificate Validation to Traffic Router

ajschmidt:
When TR obtains SSL certificates it now checks each cert before accepting it. A cert could get rejected for being expired, a mismatch between the Delivery Service URL and the subjects of the cert or a mismatch between the modulus of the certificate's private key and the public keys. Rejected certs are just dropped and not used but TR continues to operate using the valid certs it can find. 

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [X] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Unit tests have been added to the build process which test valid certs and examples of certs triggering each of the failure conditions.

T1. This can be manually tested by adding invalid certificates to a test CDN through Traffic Portal 
R1. Watch the error logs of Traffic Router when it next retrieves the SSL certs for the expected error logs.  
T2. After seeing the expected errors in the logs then attempt to make and SSL connection with TR for the offending Delivery Service URL. (Use curl with the --resolve option) 
R2. The handshake will return the 'default' certificate. 

## Check all that apply

- [X] This PR includes tests


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3303/
Test PASSed.

infrastructure/docker/traffic_router/run.sh
```diff
@@ -28,7 +28,8 @@
 # ORIGIN_URI # origin server (e.g. hotair), used to create a delivery service
 
 start() {
-	systemctl start traffic_router
+	chmod 777 starttr.sh
```
ocket8888: this should probably just be done once, in the Dockerfile, not every time TR starts.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -17,30 +17,63 @@
...
+import sun.security.rsa.RSAPublicKeyImpl;
 
+import java.math.BigInteger;
 import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.security.cert.CertificateExpiredException;
+import java.security.cert.CertificateNotYetValidException;
 import java.security.cert.X509Certificate;
+import java.util.ArrayList;
 import java.util.List;
```
ocket8888: Now, I'm not a big Java guy, so this could be very wrong, but couldn't you just use a `List` instead of an `ArrayList`? From what I understand, that leaves the list implementation up to the compiler, which in general allows better optimization. (it's also one less import)

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -49,6 +82,80 @@ public HandshakeData toHandshakeData(final CertificateData certificateData) {
...
+	private boolean hostCompare(final String hostAlias, final String subject) {
+		if (hostAlias.contains(subject) || subject.contains(hostAlias)) {
+			return true;
+		}
+		final String[] chopped = subject.split("CN=", 2);
+		String chop = null;
+		if (chopped != null && chopped.length > 1) {
+			chop = chopped[1];
+		}
+		if (chop != null) {
```
ocket8888: Can't the body of this `if` be inside the body of the previous `if` (i.e. right under `chop = chopped[1];`)? `chop` can only not be `null` if that first `if`'s condition is `true`. For that matter, you could also just declare chop inside of that as `String chop = chopped[1];` since it's not going to be used outside of the `if` block.

traffic_router/core/src/test/resources/api/1.3/cdns/name/thecdn/sslkeys-missing-1.json
```diff
@@ -22,7 +22,7 @@
...
...TjdtNGFSNQpTTHhtZ21nVkhrQ0dOZkEybXFVWHA2dTlNL1BXOGx4SjRjNExLRldCbTVJ\nYjFRdlRZM3hOWDNKWVZYZkhTWkNlCitvaDMwaStmcm9oYmNnZzRpaG9QeDU2aGhpbEdlSXZGSS9l\nMDg0M0Z0dU9mWkgrQ0s1dHRKOU1DZ1lFQTRWQnEKTTNUUlhzbUU3ZHQvSFFza2ladGFjTVdkQldx\nOVRXVlBIb3hBK3VjalFoandGVFhBZzhkZGFsbE9TS1hWNjVGawoxMlphU1lTbTd5VnplS3lVRk4w\nazBxNXpRQkRZQ1FNYVA2YjZYbGxiL2VMaWJQQTMxbU5uRGl2WjdkVWM3UUVqClhUUy9nclFLc2Zl\nenRVMXUwNG1SMTZ5THhZSHh3MW5IL3NselFKTUNnWUVBMHF6S2lkR1NxNThFZFhRRnlTS24KZ1dk\nWGgrZ1BlZUV4OExiaE1kODZicEF5VUNWNlM5bjZ0QUswZ3NJRlZzNmJZWVJYa1hjd2pZYSs1ckVq\nNStrOQpab3Q3Wjlsa2VRc0JWMnRDaTZrNnVZS0lKenVEVTFUM3FzZmlQRVdUNks1TFdPL0VXbGo0\nN0dEVS9MLzhQc3RXCm40WFM0MmRGWlBpdHRHSlV6dkhmL2VFQ2dZQXVpR3dXaW5hL0s4RmZXbWly\nTitUbzRvUFFMSS9jVVlvZEZPSTkKUGR3aHRXREx3dGk2bUtwVXpQVFhCUEN0QWtybTV0VTd3ekM4\nWkVBUnZkdFdQZFlyWk95NDhqeHRLODFpTnhqUgpzb3VjdHJuUCttNm03d21wSmtoZlhlRVpSRjAv\nK1c4elRiU0xxdUZXbGdDd1hmaVlpWjNzTy85MTMvdHRTL3FJCi9WUG5Md0tCZ1FDNGQ4OXNUL3E3\nNjZMREt6bEhySzVEUkE2TTRENEhMbjEwTWJQVExNdWNweEJlQkY1YnhvQkEKeHRFcVhBRCtyaWVr\nT3IybjJSRHlxeGt0MS9FY2JhcTI5bmNXVFFwZjZ2NkV3cTFGT3JhUDlpdTJFbERBcUcyMApXbmdG\nOUttM0VETUpBQkd5VHhieFgxZTZMRG1LaFcwWFFxRWptVmpGQi9uMFVQOGxJd2F2aGc9PQotLS0t\nLUVORCBSU0EgUFJJVkFURSBLRVktLS0tLQo=",
```
ocket8888: What's the purpose of these key replacements? Was there something wrong with the old values? Seems odd that the key would change but not the cert.

infrastructure/docker/traffic_router/Dockerfile
```diff
@@ -35,6 +35,8 @@ ARG TC_REPO=traffic-control.repo
 ADD $TMCAT /
 ADD $RPM /
 ADD $TC_REPO /etc/yum.repos.d/
+ADD starttr.sh /
+ADD shutdowntr.sh /
```
ocket8888: How come? Are you `docker exec`ing into the container to start and stop Traffic Router? What do you gain by doing that without making any changes?

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -17,30 +17,63 @@
...
-import java.util.stream.Collectors;
 
+@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.AvoidDeeplyNestedIfStmts", "PMD.NPathComplexity"})
 public class CertificateDataConverter {
 	private static final Logger log = Logger.getLogger(CertificateDataConverter.class);
 
 	private PrivateKeyDecoder privateKeyDecoder = new PrivateKeyDecoder();
 	private CertificateDecoder certificateDecoder = new CertificateDecoder();
 
+	@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.AvoidDeeplyNestedIfStmts", "PMD.NPathComplexity"})
```
ocket8888: Again, I'm not a Java guy, but this strikes me as a potential codesmell (especially since it's done at the class and method levels - does the class-level one not apply to the methods?). If the cyclomatic complexity is high enough to issue a warning, maybe it should be reduced by breaking the method into parts?

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
rawlinp: That's not how Java works. `List` is an interface of which you need a concrete implementation for, hence the `ArrayList`. Declaring it as an interface and assigning it a concrete implementation is generally best practice, as it allows you to swap in a different concrete implementation more easily without having to change every single place the `List` is used. For example, if we decided a `LinkedList` would work better for some reason, we'd only have to change it on this one line:
```
final List<X509Certificate> x509Chain = new ArrayList<>();
```

ocket8888: Ah, I see.

infrastructure/docker/traffic_router/Dockerfile
ajschmidt: systemd has been disabled in the centos 7 docker container so we can no longer use the unit files to start/stop processes. The starttr.sh gets called by the run.sh, and yes, I do `docker exec` to run the shutdown. I also copy in new TR rpms so I can test code changes without having to rebuild the whole Docker image. We have done something similar in CIAB for the TR and ATS containers.

ocket8888: Well typically you're not supposed to do that with containers, but I'd like the opinion of someone who actually uses these things - @rob05c , thoughts?

infrastructure/docker/traffic_router/run.sh
ajschmidt: Done in new commits.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
ajschmidt: I moved this down to where it belongs. I accidentally left it at the top while I was developing the solution.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
ajschmidt: You are correct. It turns out that the split actually returns an empty string if there is no token. So in the new commit I have also added and extra check to make sure 'chop' has length.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3331/
Test PASSed.

traffic_router/core/src/test/resources/api/1.3/cdns/name/thecdn/sslkeys-missing-1.json
ajschmidt: These keys have been replaced specifically to cause the modulus validation check to fail in the integration tests. It sounds negative, but its actually a positive test.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -49,6 +81,76 @@ public HandshakeData toHandshakeData(final CertificateData certificateData) {
...
+			}
+		}
+		return false;
+	}
+
+	public boolean verifyModulus(final PrivateKey privateKey, final X509Certificate certificate) {
+		BigInteger privModulus = null;
+		if (privateKey instanceof BCRSAPrivateCrtKey) {
+			privModulus = ((BCRSAPrivateCrtKey) privateKey).getModulus();
+		} else if (privateKey instanceof RSAPrivateCrtKeyImpl) {
```
rawlinp: Does this if-else chain need a default case? Otherwise, `privModulus` could end up being null and throw a NPE on L148, right? I suppose that would be captured on L76, but maybe it should be handled with a better error message here?

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -49,6 +81,76 @@ public HandshakeData toHandshakeData(final CertificateData certificateData) {
...
+				final int start = part.indexOf("modulus: ") + 9;
+				if (start < 9) {
+					continue;
+				} else {
+					pubModulus = new BigInteger(part.substring(start));
+					break;
+				}
+			}
+		}
+		if (privModulus.equals(pubModulus)) {
```
rawlinp: similar to my comment above, should we log a better error here if pubModulus is null?

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
```diff
@@ -103,7 +105,7 @@ synchronized public void importCertificateDataList(final List<CertificateData> c
 		previousData.clear();
 		for (final CertificateData certificateData : certificateDataList) {
 			final String alias = certificateData.alias();
-			if (!previousData.containsKey(alias)) {
+			if (!previousData.containsKey(alias) && master.containsKey(alias)) {
```
rawlinp: Since I'm not all that familiar with this part of TR, why is this change necessary? What is this code doing with `master` and `previousData`?

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
```diff
@@ -19,6 +19,8 @@
 import com.comcast.cdn.traffic_control.traffic_router.secure.HandshakeData;
 import com.comcast.cdn.traffic_control.traffic_router.secure.KeyManager;
 import org.apache.log4j.Logger;
+//import org.apache.tomcat.util.modeler.Registry;
```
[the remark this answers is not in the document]
ajschmidt: done in latest commit

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -17,30 +17,62 @@
...
-			final List<X509Certificate> x509Chain = encodedCertificates.stream()
-				.map(encodedCertificate -> certificateDecoder.toCertificate(encodedCertificate))
-				.collect(Collectors.toList());
-
-			return new HandshakeData(certificateData.getDeliveryservice(), certificateData.getHostname(),
-				x509Chain.toArray(new X509Certificate[x509Chain.size()]), privateKey);
+			final List<X509Certificate> x509Chain = new ArrayList<>();
+			boolean hostMatch = false;
+			boolean modMatch = false;
+			for ( final String encodedCertificate : encodedCertificates) {
```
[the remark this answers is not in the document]
ajschmidt: done

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -17,30 +17,62 @@
...
+				certificate.checkValidity();
+				if (!hostMatch && verifySubject(certificate, certificateData.alias())) {
+					hostMatch = true;
+				}
+				if (!modMatch && verifyModulus(privateKey, certificate)) {
+					modMatch = true;
+				}
+				x509Chain.add(certificate);
+			}
+			if ( hostMatch && modMatch) {
```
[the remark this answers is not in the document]
ajschmidt: done

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -17,30 +17,62 @@
...
+						x509Chain.toArray(new X509Certificate[x509Chain.size()]), privateKey);
+			}
+			else if(!hostMatch){
+				log.warn("Service name doesn't match the subject of the certificate = "+certificateData.getHostname());
+			}
+			else if (!modMatch) {
+				log.error("Modulus not == for host: "+certificateData.getHostname());
+			}
 
+		} catch ( CertificateNotYetValidException er) {
```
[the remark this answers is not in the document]
ajschmidt: done

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -17,30 +17,62 @@
...
+				if (!modMatch && verifyModulus(privateKey, certificate)) {
+					modMatch = true;
+				}
+				x509Chain.add(certificate);
+			}
+			if ( hostMatch && modMatch) {
+				return new HandshakeData(certificateData.getDeliveryservice(), certificateData.getHostname(),
+						x509Chain.toArray(new X509Certificate[x509Chain.size()]), privateKey);
+			}
+			else if(!hostMatch){
```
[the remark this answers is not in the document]
ajschmidt: done

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -17,30 +17,62 @@
...
+				log.warn("Service name doesn't match the subject of the certificate = "+certificateData.getHostname());
+			}
+			else if (!modMatch) {
+				log.error("Modulus not == for host: "+certificateData.getHostname());
```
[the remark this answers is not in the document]
ajschmidt: done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3338/
Test FAILed.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
```diff
@@ -17,30 +17,62 @@
 
 import com.comcast.cdn.traffic_control.traffic_router.shared.CertificateData;
 import org.apache.log4j.Logger;
+import org.bouncycastle.jcajce.provider.asymmetric.rsa.BCRSAPrivateCrtKey;
+import sun.security.rsa.RSAPrivateCrtKeyImpl;
```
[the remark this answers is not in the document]
ajschmidt: No. I don't think we will be able to get rid of bouncycastle unless we write our own implementation. I verified that the RSAPrivateCrtKeyImpl class is in both the Oracle JDK and the OpenJDK. The PrivateKey interface in the standard java package does not have the API to get to the modulus so we have to down cast to the implementation.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
ajschmidt: Yes. I corrected it in the latest commit.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateDataConverter.java
ajschmidt: I corrected it in the latest commit. This case will fall through the error message that is output by the calling method.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
ajschmidt: `PreviousData` is the current map of `CertificateData` which is used to compare with the input list (`certificateDataList`) to find differences which need to be updated or deleted. `master` is the subset of valid `HandshakeData` which were successfully converted from the `CertificateData` in the input `certificateDataList`. The added condition in the `if` statement is necessary because of the new certificate validation feature could have filtered some of the entries out of the input `certificateDataList'.

rawlinp: Cool, makes sense. Thanks for helping me learn!

rawlinp:
I approved the code but still want to pull this down to test a little before merging.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3339/
Test PASSed.

PR #3392, opened 2019-03-07
Tr default cert

ajschmidt:

This PR sets and explicit self-signed certificate as the 'default' certificate which gets returned during an SSL handshake if the requesting host does not match any of the other valid certificates TR hosts. The default certificate will not match any hostname and should be rejected by the client. 

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [X] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

There are unit tests and integration tests built into the TR build which test this PR. To manually test:
T1. use curl -kv --resolve to connect with an SNI host that matches the domain of an existing DS, but not the regular expression. For instance try https://cdn.someds.somecdn.toplevel for a DS that responds to https://edge.someds.somecdn.toplevel
R1. Inspect the curl output and you should see that the TLS handshake was successful but the certificate returned was for domain : Invalid

## Check all that apply

- [X] This PR includes tests


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3322/
Test PASSed.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
```diff
@@ -69,23 +68,11 @@ synchronized private void replaceSSLHosts(final Map<String, HandshakeData> sslHo
...
-                firstAlias = false;
             }
         }
-
-    }
-
-    synchronized public void reloadSSLHosts(final Map<String, HandshakeData> cr) {
-        replaceSSLHosts(cr);
-
-        for (final HandshakeData data : cr.values()) {
```
rawlinp: So this is unnecessary now since TR just always sets the default cert and will never swap in a "default" cert from the new sslkeys.json?

rawlinp:
Also, can you add a line in CHANGELOG.md about TR's new default cert behavior?

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
ajschmidt: It was never necessary. All of those same steps happen in the `replaceSSLHosts` method and are for some reason repeated again. I just changed the `replaceSSLHosts` to be public and changed the callback to call it.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
```diff
@@ -37,9 +42,31 @@ private CertificateRegistry() {
 	}
 
 	public static CertificateRegistry getInstance() {
+		synchronized (CertificateRegistryHolder.DELIVERY_SERVICE_CERTIFICATES) {
```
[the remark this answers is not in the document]
ajschmidt: Yes. The `importCertificateData` also synchronizes on the `CertificateRegistry` instance.  Also, the `getInstance()` method could be called by multiple threads.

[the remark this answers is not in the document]
rawlinp: `importCertificateDataList` is a synchronized method, which I think is the same as if the method body was synchronized on `CertificateRegistry.class`, right?

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
```diff
@@ -37,9 +42,31 @@ private CertificateRegistry() {
...
+			final Map<String, HandshakeData> handshakeDataMap =
+					CertificateRegistryHolder.DELIVERY_SERVICE_CERTIFICATES.getHandshakeData();
+			handshakeDataMap.putIfAbsent(DEFAULT_SSL_KEY, createDefaultSsl());
```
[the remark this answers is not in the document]
ajschmidt: Yes. I've updated the PR. Now it returns a null CertificateRegistry if it can't create the cert.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3340/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3443/
Test PASSed.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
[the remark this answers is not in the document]
ajschmidt: Yes. I placed this Synchronize here to protect the data structure in the event that more than one thread calls the 'getInstance' method during startup. There are in fact three endpoints being initialized using the LanguidNioProtocol at startup. Once initialized this method never gets called again so they synchronized doesn't have any impact on performance.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
```diff
@@ -61,61 +66,57 @@ synchronized private void replaceSSLHosts(final Map<String, HandshakeData> sslHo
...
+
+        /* This override has been added as a temporary hack to resolve an issue in Tomcat.
+           Once the issue has been corrected in Tomcat then this can be removed. The
+           'SSL.getLastErrorNumber()' removes an unwanted error condition from the error stack
+           in those cases where some error condition has caused the socket to get closed and
+           then the processor was put back on the processor stack for reuse in a future connection.
+         */
+        @Override
+        protected void doRun() {
+	        final SocketWrapperBase<NioChannel> localWrapper = socketWrapper;
```
rawlinp: this line got mixed up w/ tabs and spaces

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
```diff
@@ -17,14 +17,29 @@
 
 import com.comcast.cdn.traffic_control.traffic_router.protocol.RouterNioEndpoint;
 import com.comcast.cdn.traffic_control.traffic_router.shared.CertificateData;
+//import com.fasterxml.jackson.core.type.TypeReference;
```
rawlinp: these imports can be removed if they're usused

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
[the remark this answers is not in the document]
rawlinp: The data structure would only be protected from race conditions if the other code that accesses it also synchronized on the same object. `synchronized (CertificateRegistryHolder.DELIVERY_SERVICE_CERTIFICATES)` and `synchronized (CertificateRegistry.class)` are locking on two different things, so both of those methods could overlap with each other.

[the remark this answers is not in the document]
rawlinp: Basically, I'm trying to understand why this method isn't just `synchronized public void` like `importCertificateDataList` if it's also meant to not overlap with `importCertificateDataList`. That method is also mutating `handshakeDataMap`, so I'd think `getInstance` and `importCertificateDataList` would have to be sync'd on the same object (i.e. use the same mutex).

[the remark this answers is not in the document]
ajschmidt: I've tested with your suggestion and it works. I was concerned that locking the entire class, (since getInstance is a static method) might prevent the CertificateRegistryHolder from creating the new CertificateRegistry, but it looks like the class loader is getting to the CertificateRegistryHolder first.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
ajschmidt: done

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
ajschmidt: done

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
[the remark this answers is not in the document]
ajschmidt: this is done in new commits

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
```diff
@@ -61,61 +66,57 @@ synchronized private void replaceSSLHosts(final Map<String, HandshakeData> sslHo
...
+
+        /* This override has been added as a temporary hack to resolve an issue in Tomcat.
+           Once the issue has been corrected in Tomcat then this can be removed. The
+           'SSL.getLastErrorNumber()' removes an unwanted error condition from the error stack
+           in those cases where some error condition has caused the socket to get closed and
+           then the processor was put back on the processor stack for reuse in a future connection.
+         */
+        @Override
+        protected void doRun() {
+			final SocketWrapperBase<NioChannel> localWrapper = socketWrapper;
```
rawlinp: this line is still way over-indented -- its using tabs and it looks like the other lines use spaces

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3475/
Test FAILed.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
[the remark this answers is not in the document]
rawlinp: Ah yeah I just realized that `getInstance` is now `static synchronized` while `importCertificateDataList` is non-static `synchronized`. So those methods aren't actually locking on the same object yet (`static` locks on the class while non-static locks on a single instance).

So if one thread calls `getInstance` and another thread calls `importCertificateDataList` at the same time, both methods will interleave and mutate `handshakeDataMap` at the same time. That's what you're trying to prevent with the synchronization here, right?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3477/
Test FAILed.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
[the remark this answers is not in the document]
ajschmidt: I'm more concerned about simultaneous calls to 'getInstance' during startup. The poller that calls 'importCertificateDataList' launches after initialization has succeeded.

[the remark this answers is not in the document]
ajschmidt: This discussion has caused me to realize that there is a bug in `getInstance`. I've just committed another change.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3478/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3492/
Test PASSed.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
```diff
@@ -169,7 +156,22 @@ synchronized public void importCertificateDataList(final List<CertificateData> c
...
+			// Check to see if a Default cert has been provided/created previously
+			if (handshakeDataMap.containsKey(DEFAULT_SSL_KEY)) {
+				master.put(DEFAULT_SSL_KEY, handshakeDataMap.get(DEFAULT_SSL_KEY));
+			}else{
+				// create a new default certificate
+				final HandshakeData defaultHd = createDefaultSsl();
+				if (defaultHd == null){
+					log.error("Failed to initialize the CertificateRegistry because of a problem with the 'default' " +
+							"certificate.  Returning the Certificate Registry without a default.");
+					return;
```
rawlinp: If this ends up being the case, would we still want to do `handshakeDataMap = master` before returning so that all the newly-processed certs would still be loaded (just without a default cert), or would it never make sense to keep the rest of the certs without the default cert in that case?

ajschmidt: In both situation I think it will fail in either case when it goes to `RouterNioEndpoint` because there has to be a default `SSLHost`. This condition should only occur when TR is being initiated. I did perform a test and made sure that these lines only get called once. After that they are skipped because here is always a default cert.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
ajschmidt: Okay. It should be better now. There seems to be an issue with one of my code templates.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3502/
Test PASSed.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
ajschmidt: In other words. I think this is the appropriate behavior. If it can't create the default certificate, something bigger is wrong and TR should not start.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
rawlinp: Thanks, it looks much better now.

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/secure/CertificateRegistry.java
rawlinp: Thanks, you make a fair point.

PR #3394, opened 2019-03-08
Fix interface ptr problem

### LINKED ISSUE #3373: Unable to create a delivery service request when count is zero
1. Ensure that there are no rows in the deliveryservice_request table
2. Set dsRequest.enabled=true in traffic_portal_properties.json
3. Update an existing DS in TP

doing this will make a call to:

POST api/1.5/deliveryservice_requests

which returns this error:

`{"alerts":[{"text":"sql: no rows in result set","level":"error"}]}`

### LINKED ISSUE #3378: Unable to create multiple delivery service requests
1. Set dsRequest.enabled=true in traffic_portal_properties.json
2. Update an existing DS in TP
3. fulfill/complete the ds request
4. update the same DS a 2nd time

doing No. 4 will attempt to create another DS request and will make a call to:

POST api/1.5/deliveryservice_requests

which returns this error:

`{"alerts":[{"text":"'status' complete request cannot be changed","level":"error"}]}`

in fact, even if  your 2nd ds change is another ds change you get the same error.

it seems like "creating a ds request" is somehow invoking the "update a ds request" logic...

moltzaum:

Fixes #3378
Fixes #3373

Fixes potentially other bugs that would have used the last posted data
in the handler.

- Traffic Ops

Follow the steps in the mentioned issues and try to reproduce the bugs.

There aren't tests. Errors should appear at compile time.

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/api/shared_handlers.go
+++ b/traffic_ops/traffic_ops_golang/api/shared_handlers.go
@@ -25,6 +25,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strconv"
 	"strings"
 
@@ -101,6 +102,17 @@ func GetCombinedParams(r *http.Request) (map[string]string, error) {
 	return combinedParams, nil
 }
 
+// reset will set the zero value of the struct pointer
+func reset(v interface{}) error {
+	if reflect.ValueOf(v).Kind() != reflect.Ptr {
+		return errors.New("can only reset pointer values")
+	}
+
+	p := reflect.ValueOf(v).Elem()
+	p.Set(reflect.Zero(p.Type()))
+	return nil
+}
+
 // decodeAndValidateRequestBody decodes and validates a pointer to a struct implementing the Validator interface
 func decodeAndValidateRequestBody(r *http.Request, v Validator) error {
 	defer r.Body.Close()
@@ -125,6 +137,7 @@ func ReadHandler(obj Reader) http.HandlerFunc {
 		}
 		defer inf.Close()
 
+		reset(obj)
 		obj.SetInfo(inf)
 		results, userErr, sysErr, errCode := obj.Read()
 		if userErr != nil || sysErr != nil {
@@ -151,6 +164,7 @@ func UpdateHandler(obj Updater) http.HandlerFunc {
 		}
 		defer inf.Close()
 
+		reset(obj)
 		obj.SetInfo(inf)
 		if err := decodeAndValidateRequestBody(r, obj); err != nil {
 			HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, err, nil)
@@ -230,6 +244,7 @@ func DeleteHandler(obj Deleter) http.HandlerFunc {
 		}
 		defer inf.Close()
 
+		reset(obj)
 		obj.SetInfo(inf)
 		keyFields := obj.GetKeyFieldsInfo() // expecting a slice of the key fields info which is a struct with the field name and a function to convert a string into a interface{} of the right type. in most that will be [{Field:"id",Func: func(s string)(interface{},error){return strconv.Atoi(s)}}]
 		keys := make(map[string]interface{})
@@ -291,6 +306,7 @@ func CreateHandler(obj Creator) http.HandlerFunc {
 		}
 		defer inf.Close()
 
+		reset(obj)
 		obj.SetInfo(inf)
 		err := decodeAndValidateRequestBody(r, obj)
 		if err != nil {
--- a/traffic_ops/traffic_ops_golang/api/shared_handlers.go
+++ b/traffic_ops/traffic_ops_golang/api/shared_handlers.go
@@ -102,17 +102,6 @@ func GetCombinedParams(r *http.Request) (map[string]string, error) {
 	return combinedParams, nil
 }
 
-// reset will set the zero value of the struct pointer
-func reset(v interface{}) error {
-	if reflect.ValueOf(v).Kind() != reflect.Ptr {
-		return errors.New("can only reset pointer values")
-	}
-
-	p := reflect.ValueOf(v).Elem()
-	p.Set(reflect.Zero(p.Type()))
-	return nil
-}
-
 // decodeAndValidateRequestBody decodes and validates a pointer to a struct implementing the Validator interface
 func decodeAndValidateRequestBody(r *http.Request, v Validator) error {
 	defer r.Body.Close()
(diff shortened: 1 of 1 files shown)


moltzaum:
@mitchell852

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3327/
Test FAILed.

moltzaum:
retest this please

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3332/
Test FAILed.

rawlinp:
This seems very similar to the issue we had a while back with inadvertently sharing the server-side request object between requests. Basically, the issue was that the API was reusing the same struct for consecutive requests. So if I made a request with all the fields set, followed by another request with some fields null, the null fields in the 2nd request were getting set to the values used in the 1st request. I think this is why the `NewReadObj()` function was added to the `GenericReader` interface. @rob05c does this issue sound familiar? @moltzaum maybe the `Reader` interface needs something like that now?

moltzaum:
I don't think it does. For each CRUD operation the struct is zeroed out, which should functionally be equivalent to returning a new object.

I think the struct scan also wants to scan into every field, so the TOType that is passed wouldn't work, and because the `GenericReader` is defined on an interface, you can't use the dot operator to get the embedded struct.

api/crud.go:103
```
v := val.NewReadObj()
if err = rows.StructScan(v); err != nil {
  return nil, nil, errors.New("scanning " + val.GetType() + ": " + err.Error()), http.StatusInternalServerError
}
vals = append(vals, v)
```

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3341/
Test PASSed.

mitchell852:
this doesn't seem to fix #3378 - i'm still unable to create subsequent ds requests for the same ds.

mitchell852:
seems to fix #3378. nice job!

mitchell852:
seems to fix #3373. again, nice job!

dangogh:
I'm concerned that this actually does not fix the problem, but hides it.  The obj parameter being passed in is always a default object created on the call (e,g, &tc.,  and so always has the zero value.   Can you explain why the fix actually works?

dangogh:
on further reflection,  I think this *almost* solves the problem...    When the handler generator (e.g. `ReadHandler(obj ...)` is called, the func returned is a closure on `obj`.  Each call to that function (when that endpoint is hit) will use that same object to store the value.  zeroing out that object may solve the problem if there can only be a single concurrent call to that function,  but will fail in unpredictable ways if there are 2 concurrent calls which would both use the object at the same address.

This could be fixed by making a copy of `obj` upon entry to the closure and using the copied object instead.   In this case, zeroing out the object would not be necessary....

rawlinp:
Yeah, I'm thinking we need some sort of solution here that involves a unique object per request. We got into some trouble in the CRUDer stuff by sharing the same request object between requests. Since every request is handled in its own goroutine, really we shouldn't be sharing _any_ mutable state between goroutines without proper safeguards. The zeroing out thing does sound like a data race, which you could try to invoke manually by making the handler for request A sleep after deserializing the request and letting another handler run without sleeping for a different request B (e.g. `if obj.foo == "A": time.sleep(30)`). In theory, the handler for request B would have zeroed out/modified the same object as request A while it was sleeping. When the handler for request A wakes up, the data would match the data for request B if there is a data race.

moltzaum:
Good catch @dangogh. I'll copy from `obj`. By only ever reading from it there won't be a race.

moltzaum:
Does anyone mind if I remove the `if interfacePtr.Kind() != reflect.Ptr {`? 
If a pointer is not provided the following message should appear:

```
# github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing
routing/routes.go:379:81: cannot use apitenant.TOTenant literal (type apitenant.TOTenant) as type api.Deleter in argument to api.DeleteHandler:
	apitenant.TOTenant does not implement api.Deleter (Delete method has pointer receiver)
```

rob05c:
> Does anyone mind if I remove the if interfacePtr.Kind() != reflect.Ptr {?` check

@moltzaum Yes, please keep it. It's still possible to fulfill the interface with a non-pointer and compile successfully, if something implements the functions with non-pointer receivers, e.g. `func (dv TODivision) Delete() (error, error, int) { return api.GenericDelete(dv) }`. In which case, removing the `!= reflect.Ptr` will cause a runtime panic.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3365/
Test PASSed.

mitchell852:
actually, i'm getting one failed unit test

```# github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/status [github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/status.test]
status/statuses_test.go:82:20: too few values in TOStatus literal```

can you see if you can replicate that @moltzaum

**Comment by moltzaum** [comment]
@mitchell852 I ran the whole integration test suite. I did not get an error for statuses, however I got an authorization error.
```
--- FAIL: TestFederations (4.81s)
	userdeliveryservices_test.go:69: failed to set delivery service users: 401 Unauthorized[401] - Error requesting Traffic Ops https://localhost:6443/api/1.4/deliveryservice_user {"alerts":[{"level":"error","text":"Unauthorized, please log in."}]}
	userdeliveryservices_test.go:78: get user delivery services expected 4 actual 0
	userdeliveryservices_test.go:90: get user delivery services expected 33 actual missing
	userdeliveryservices_test.go:90: get user delivery services expected 30 actual missing
	userdeliveryservices_test.go:90: get user delivery services expected 31 actual missing
	userdeliveryservices_test.go:90: get user delivery services expected 32 actual missing
	userdeliveryservices_test.go:182: get user delivery services expected >0 actual 0
FAIL
exit status 1
```

I got one other fail for user delivery services
`--- FAIL: TestUserDeliveryServices (4.44s)`

Neither of these endpoints use the CRUDer framework (hence are not affected by the fix), and I reproduced these errors on master.

rawlinp:
> all api tests passed but i'm getting one failed unit test
> 
> `github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/status [github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/status.test] status/statuses_test.go:82:20: too few values in TOStatus literal`
> 
> can you see if you can replicate that @moltzaum

I believe that `too few values in TOStatus literal` error has been fixed in master and should go away after rebasing this PR onto master.

moltzaum:
> I believe that too few values in TOStatus literal error has been fixed in master and should go away after rebasing this PR onto master.  

I've rebased and confirmed this.

moltzaum:
@rawlinp Is your concern addressed?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3464/
Test FAILed.

rawlinp:
Yes, it appears that each request gets unmarshalled into a fresh struct now.

moltzaum:
retest this please

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3471/
Test PASSed.

PR #3402, opened 2019-03-14
Added to CIAB README.md to document running a limited stack

moltzaum:

ORIGINAL MESSAGE:
Added variable "ADD_DEMO1_SSLKEYS", which determines whether traffic ops is configured with riak and whether the ssl key demo (requiring riak) is tried.

This essentially allows developers to run in a limited environment that just includes trafficops without spamming the logs.

EDIT:
I was not able to reproduce the original issue I had that prompted this PR. Instead, I've documented running TO and TP by themselves, and I would like verification the commands work on others' machines.

- CDN in a Box

ORIGINAL MESSAGE:
The best way to verify this PR is to run CIAB after updating `variables.env` and looking at the logs. 
`docker-compose trafficops-perl trafficops`

With all default values, the logs will repeated messages about SSL keys for the demo1 deliveryservice.
With the new variable false, the logs will repeat messages about AUTO-SNAPQUEUE.
With both the new variable and AUTO_SNAPQUEUE_ENABLED false, the logs will not repeat. You may see an exec tail.

It is not necessary to toggle the auto snapqueue variable, but imo it makes it easier to spot the difference.

EDIT:
Try to reproduce running the commands in the `README.md` successfully.

## Check all that apply

- [ ] This PR includes tests
- [x] This PR is not related to security

### DIFF
--- a/infrastructure/cdn-in-a-box/traffic_ops/run-go.sh
+++ b/infrastructure/cdn-in-a-box/traffic_ops/run-go.sh
@@ -74,11 +74,7 @@ touch /var/log/traffic_ops/traffic_ops.log
 # enroll in the background so traffic_ops_golang can run in foreground
 TO_USER=$TO_ADMIN_USER TO_PASSWORD=$TO_ADMIN_PASSWORD to-enroll $(hostname -s) &
 
-if [[ -z "$TV_HOST" ]]; then
-    ./bin/traffic_ops_golang -cfg $CDNCONF -dbcfg $DBCONF & 
-else 
-    ./bin/traffic_ops_golang -cfg $CDNCONF -dbcfg $DBCONF -riakcfg $RIAKCONF &
-fi
+./bin/traffic_ops_golang -cfg $CDNCONF -dbcfg $DBCONF -riakcfg $RIAKCONF &
 
 to-enroll "to" ALL || (while true; do echo "enroll failed."; sleep 3 ; done)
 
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3342/
Test PASSed.

Shihta:
Why we have to set a special variable to run `trafficops` individually?
I think the goal should be running `trafficops` individually without any environment variable.

moltzaum:
You have a good point. The trouble is, I'm not sure how to tell from the run.sh file whether or not traffic vault is running.

moltzaum:
The common case is to build CIAB all at once, so the variable is true by default. If the variable is set to false the demo wouldn't work the same. What would you name the environment variable?

infrastructure/cdn-in-a-box/traffic_ops/run-go.sh
```diff
@@ -74,7 +74,11 @@ touch /var/log/traffic_ops/traffic_ops.log
 # enroll in the background so traffic_ops_golang can run in foreground
 TO_USER=$TO_ADMIN_USER TO_PASSWORD=$TO_ADMIN_PASSWORD to-enroll $(hostname -s) &
 
-./bin/traffic_ops_golang -cfg $CDNCONF -dbcfg $DBCONF -riakcfg $RIAKCONF &
+if [[ "$ADD_DEMO1_SSLKEYS" = false ]]; then
+    ./bin/traffic_ops_golang -cfg $CDNCONF -dbcfg $DBCONF & 
+else 
+    ./bin/traffic_ops_golang -cfg $CDNCONF -dbcfg $DBCONF -riakcfg $RIAKCONF &
```
Shihta: I think the point is that the variable name is `ADD_DEMO1_SSLKEYS` but it's used to remove `raikcfg`.
If I hadn't read this PR, I would have thought it was simply removing SSLKEY of this Delivery Service.
DEMO1 should correctly provide HTTP service and I can manually add SSLKEY for it as well.

Shihta:
My personal opinion is that let `run.sh` check TV-related variables is better than adding another variables to disable TV.
e.g.
If TO can't detect [these variables](https://github.com/apache/trafficcontrol/blob/master/infrastructure/cdn-in-a-box/variables.env#L84-L94), it run as `NO-TV-Mode`.
If TV-realted variables exists, TO run `TV-Mode`.

How about raise this topic on slack to get others opinion?

ocket8888:
I agree, I like that better than introducing another variable. Just switch on `TV_HOST` (best choice imo) and make note of that in the docs.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3387/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3388/
Test PASSed.

Shihta:
I found there is a runtime error in `enroller` but I believe it will not affect `trafficops`
螢幕快照 2019-03-25 上午11 12 09

I hope you can also copy this section to `docs/source/admin/quick_howto/ciab.rst`, let this section can be easily accessed by others

moltzaum:
Hmm, I don't recall it doing that for me. I'll try to verify.

moltzaum:
Closing this. I don't really have time to look at this, nor do I think I will get to it later.

PR #3417, opened 2019-03-18
Fix add sslkeys endpoint to always use the input certificate

### LINKED ISSUE #3398: TO is mangling SSL server cert on update
Version: 3.0.0-9925.bd61e646.el7
When adding in a new server cert or updating an existing server cert, after pasting in the server cert + intermediate CA Chain into the server cert field and hitting update, the version number increments. TO logs show a POST action with a 200 response and no errors are thrown.

Leave the SSL page and come back to it for that DS, and the server cert field will only contain the server cert and last cert from the intermediate chain. The other certs from the intermediate chain have been removed. 

Validated that this mangled cert is pushed to ATS servers and sometimes causes TR to do Null Pointer Exceptions when attempting to process the cert.

rawlinp:

Sometimes, the Certificate.Verify method will remove elements from a
valid certificate chain. Instead of using the output of that method,
always use the user input certificate and return a warning if the input
certificate does not match the output of the Certificate.Verify method.

Fixes #3398

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Includes changes to the unit tests, but there is currently an unrelated unit test failure in statuses_test.go which should be addressed in a separate PR:
```
# github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/status [github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/status.test]
status/statuses_test.go:82:20: too few values in TOStatus literal
```
Manual verification can be done by adding an sslkey to a delivery service that would normally get mangled by the API and checking that the stored cert matches the input cert.
If you don't have a cert like that on hand, you can still check by adding a non-self-signed cert and verifying that the stored cert matches the input cert.

## Check all that apply

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/keys.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/keys.go
@@ -59,7 +59,7 @@ func AddSSLKeys(w http.ResponseWriter, r *http.Request) {
 		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
 		return
 	}
-	certChain, isUnknownAuth, err := verifyCertificate(req.Certificate.Crt, "")
+	certChain, isUnknownAuth, isVerifiedChainNotEqual, err := verifyCertificate(req.Certificate.Crt, "")
 	if err != nil {
 		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("verifying certificate: "+err.Error()), nil)
 		return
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3354/
Test PASSed.

rawlinp:
This PR might be getting rolled into #3382 since that PR touches the same code.

rawlinp:
This has been addressed in #3382

PR #3413, opened 2019-03-18
CIAB: Add CDN_NAME environment variable

Shihta:
## Which issue is fixed by this PR? If not related to an existing issue, what does this PR do?

Add a new environment variable: `CDN_NAME`
It's used to modify the CDN name in CIAB.

This commit also fix a `AUTO_SNAPQUEUE` issue.
If there are more than 1 deliveryservices in the folder `/infrastructure/cdn-in-a-box/traffic_ops_data/deliveryservices`, the feature `AUTO_SNAPQUEUE` will be failed.

## Which TC components are affected by this PR?

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] Other CIAB

### (If no tests are part of this PR, please provide explanation as to why no tests are included.)

1. Modify the `CDN_NAME`. e.g. `TESTCDN`
2. Start CIAB
3. CIAB will work correctly and it's CDN name should be `TESTCDN`

## Check all that apply

- [ ] This PR includes tests
- [ ] This PR includes documentation updates
- [ ] This PR includes an update to CHANGELOG.md
- [x] This PR includes all required license headers
- [ ] This PR includes a database migration (ensure that migration sequence is correct)
- [ ] This PR fixes a serious security flaw. Read more: [www.apache.org/security](http://www.apache.org/security/)

PR #3424, opened 2019-03-20
CIAB: Dynamically choose the first mid server

bakerolls:

This PR would enable renaming the default Mid Server by removing the hard coded name `CDN_in_a_Box_Mid`. It now calls the API until a cachegroup exists. If there are more than one cachegroups, the first is selected.

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [x] CDN in a Box

Change the `name` in [060-ciabMid.json](infrastructure/cdn-in-a-box/traffic_ops_data/cachegroups/060-ciabMid.json), `parentCachegroupName` in [070-ciabEdge.json](infrastructure/cdn-in-a-box/traffic_ops_data/cachegroups/070-ciabEdge.json) and start CIAB. The default Mid server should have the new name.

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/infrastructure/cdn-in-a-box/mid/run.sh
+++ b/infrastructure/cdn-in-a-box/mid/run.sh
@@ -58,7 +58,13 @@ while [[ -z $found ]]; do
     found=$(to-get api/1.3/cdns?name="$CDN" | jq -r '.response[].name')
 done
 
-to-enroll mid $CDN "CDN_in_a_Box_Mid" || (while true; do echo "enroll failed."; sleep 3 ; done)
+cachegroup=
+while [[ -z $cachegroup ]] || [[ "$cachegroup" == "null" ]]; do
+    echo 'waiting for enroller server setup'
+    sleep 3
+    cachegroup=$(to-get api/1.3/servers | jq -r '.response[0].cachegroup')
+done
+to-enroll mid $CDN $cachegroup || (while true; do echo "enroll failed."; sleep 3 ; done)
 
 while [[ -z "$(testenrolled)" ]]; do
 	echo "waiting on enrollment"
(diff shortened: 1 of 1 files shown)


asfgit:

bakerolls:
After #3413 gets merged, `$CDN` has to be renamed to `$CDN_NAME `.

Shihta:
mid server should use `MID_LOC` cachegroup so you can not randomly choose one.

ocket8888:
I agree with @Shihta , I don't think assigning the mid server to the cache group of the first server randomly enrolled/returned by the API is a good idea.

bakerolls:
Sorry, I honestly did not check the full response of `/api/1.3/servers`. Do you think a random mid server would be sufficient? If yes, we could just add `?type=MID`. If no, should the default be defined inside `variables.env`?

ocket8888:
That won't work at all. Before the mid the server is enrolled, you shouldn't be able to find a mid server, so it'll just loop endlessly.

Shihta:
For me, I would like adding a parameter rather than random choice.
Even you specified the type, directly picking the first one is still a random selection method.
If you want to reuse the same mid container for larger system (more than 1 mid server/cachegroup), uncontrollable random selection method should not be a good idea.

bakerolls:
It worked once, but there were probably some volumes not removed or containers still running. In the end, I used a variable that defined the name. Thank you for your comments.

Issue #2494, opened 2018-07-02
Multiple Go API endpoints don't return "joined" fields on PUT/POST


rawlinp:
This is a regression from the Perl API, where keys like `parentCachegroupName` are returned with the name of the cachegroup rather than `null`. This requires following up `INSERT` or `UPDATE` SQL statements with a `SELECT` that returns the proper fields. This is essentially what the Perl API did through the DBIx ORM for these endpoints, but currently these APIs don't follow up with a `SELECT` to return the same data as the Perl API and return a `null` value in the response:

- POST, PUT /api/1.1/cachegroups
  - parentCachegroupName, secondaryParentCachegroupName, typeName
- POST, PUT /api/1.1/asns
  - cachegroup
- POST, PUT /api/1.3/deliveryservices
  - cdnName, profileName, profileDescription
- POST, PUT /api/1.1/regions
  - divisionName
- POST, PUT /api/1.1/phys_locations
  - region
- POST, PUT /api/1.1/servers
  - cachegroup, cdnName, physLocation, profile, profileDesc, status, type
- POST, PUT /api/1.1/tenants
  - parentName

Example PUT /cachegroups/{id} response from traffic_ops_golang:
```
{
    "alerts": [
        {
            "level": "success",
            "text": "cachegroup was updated."
        }
    ],
    "response": {
        "fallbackToClosest": null,
        "id": 531,
        "lastUpdated": "2018-07-02 17:07:59+00",
        "latitude": 39.739167,
        "longitude": -104.984722,
        "name": "rawl-cg",
        "parentCachegroupId": 3,
        "parentCachegroupName": null,
        "secondaryParentCachegroupId": null,
        "secondaryParentCachegroupName": null,
        "shortName": "rawl-cg",
        "typeId": 6,
        "typeName": null
    }
}
```
Example PUT /cachegroups/{id} response from Perl TO with the same request payload:
```
{
    "alerts": [
        {
            "level": "success",
            "text": "Cachegroup update was successful."
        }
    ],
    "response": {
        "fallbackToClosest": 1,
        "id": 532,
        "lastUpdated": "2018-07-02 21:42:40.062963+00",
        "latitude": 39.739167,
        "longitude": -104.984722,
        "name": "rawl-cg",
        "parentCachegroupId": 3,
        "parentCachegroupName": "mid-west",
        "secondaryParentCachegroupId": null,
        "secondaryParentCachegroupName": null,
        "shortName": "rawl-cg",
[...]

PR #3429, opened 2019-03-22
Added ScanIntoItems and ScanIntoItem functions

moltzaum:

First I created a ScanIntoItems function that basically unpacks a list
into sql's scan function.

Then I updated the generic cruder to use the function, although it is meant
more for endpoints that don't already have a generic function. I updated
the generic case since it could be used as a model and so that the error
messages won't look too different from implementors that use ScanIntoItems.

ScanIntoItems is especially convienent when returning a number of items.
It could potentially be used for: #2494

Here is an example that is applicable for cachegroups:
```
items := []interface{}{
    &cg.ID,
    &cg.ParentName,
    &cg.SecondaryParentName,
}
if err = api.ScanIntoItems(resultRows, items); err != nil {
    if err == sql.ErrNoRows {
	return errors.New("no cachegroup found with this id"), nil, http.StatusNotFound
    }
    return nil, err, http.StatusInternalServerError
}
```

Mayber later down the line, ScanIntoItems would be prefered over
ScanStruct since ScanStruct belongs to sqlx. I don't think we use sqlx
for anything other than ScanStruct, so we could potentially get rid of
a dependency.

- Traffic Ops

Run the traffic ops integration tests.

This PR doesn't introduce new tests, but probably should. The function advises function callers that they don't need to close sql.Rows. The ability of the function to close the rows properly should be tested, but I'm not sure how it should be done yet. I found the mockdb, but I haven't found an example of a query running against it yet.

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/api/generic_crud.go
+++ b/traffic_ops/traffic_ops_golang/api/generic_crud.go
@@ -20,6 +20,7 @@ package api
  */
 
 import (
+	"database/sql"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3398/
Test PASSed.

moltzaum:
Update:
As I've written it now, `ScanIntoItems` uses `sql.Rows`, but an update or create should use `QueryRow`, instead of `NamedQuery` or `Query`.

The generic create uses `NamedQuery` only because it allows us to map fields from an interface using their names. Note: is possible to have a simple `NamedQueryRow` function (https://github.com/jmoiron/sqlx/issues/102), or to make a function that uses reflection to read from the struct in a linear fashion (instead of using names). Both of these solutions would eliminate the need for `ScanIntoIItems`

TL;DR:
I don't think `ScanIntoItems` should be used as it currently exists, so I'm closing this PR.

PR #3442, opened 2019-03-28
Fixes internal server error for `cachegroups/:id/queue_update` blank field

### LINKED ISSUE #2933: /api/1.x/cachegroups/{{ID}}/queue_update gives incorrect error message when request doesn't specify a CDN
The easiest way to see this is just send a request to queue updates on a cachegroup without specifying enough information e.g. `curl -sb cookie.jar -d '{"action": "queue"}' "$TO_URL/api/1.1/cachegroups/7/queue_update"`. This gives the response:
```json
{
  "alerts": [
    {
      "text": "cdn does not exist",
      "level": "error"
    }
  ]
}
```
... which may be referring to the fact that the actual `"cdn"` request key doesn't exist, but the error message could be more clear on that - also, based on other error messages for similar situations, the message `"cdn" is a required field` seems more consistent.

moltzaum:

Fixes #2933 
Also adds error message for when cdn is blank, preventing an internal server error that is caused by a panic.

- Traffic Ops

Post to the endpoint using a blank cdn and then with no cdn (as it is mentioned in the issue)
`curl --cookie $mc -d '{"action": "queue", "cdn": ""}' -Lvsk "https://localhost:6443/api/1.1/cachegroups/7/queue_update"`

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go
+++ b/traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go
@@ -49,10 +49,14 @@ func QueueUpdates(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if reqObj.CDN == nil && reqObj.CDNID == nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("cdn does not exist"), nil)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("cdn is a required field"), nil)
 		return
 	}
-	if reqObj.CDN == nil || *reqObj.CDN == "" {
+	if *reqObj.CDN == "" {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("cdn cannot be blank"), nil)
+		return
+	}
+	if reqObj.CDNID != nil && reqObj.CDN == nil || *reqObj.CDN == "" {
 		cdn, ok, err := dbhelpers.GetCDNNameFromID(inf.Tx.Tx, int64(*reqObj.CDNID))
 		if err != nil {
 			api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("getting CDN name from ID '"+strconv.Itoa(int(*reqObj.CDNID))+"': "+err.Error()))
--- a/traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go
+++ b/traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go
@@ -48,15 +48,11 @@ func QueueUpdates(w http.ResponseWriter, r *http.Request) {
 		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("action must be 'queue' or 'dequeue'"), nil)
 		return
 	}
-	if reqObj.CDN == nil && reqObj.CDNID == nil {
+	if reqObj.CDNID == nil && (reqObj.CDN == nil || *reqObj.CDN == "") {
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3411/
Test FAILed.

ocket8888:
As a mere, lowly contributor I cannot assign issues to anyone, but I'll check this out tomorrow.

moltzaum:
That was an accident. I'll remove it.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3413/
Test PASSed.

traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go
```diff
@@ -49,10 +49,14 @@ func QueueUpdates(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if reqObj.CDN == nil && reqObj.CDNID == nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("cdn does not exist"), nil)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("cdn is a required field"), nil)
 		return
 	}
-	if reqObj.CDN == nil || *reqObj.CDN == "" {
+	if *reqObj.CDN == "" {
```
rob05c: This doesn't seem to make sense, with the check below. This will panic if `reqobj.CDN` is nil, but `reqobj.CDNID` is not nil. 

It seems like, with the check below, that this should just be removed. Right? If not, it still needs to check `reqObj.CDN == nil` to avoid panicing.

traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go
```diff
@@ -49,10 +49,14 @@ func QueueUpdates(w http.ResponseWriter, r *http.Request) {
...
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("cdn cannot be blank"), nil)
+		return
+	}
+	if reqObj.CDNID != nil && reqObj.CDN == nil || *reqObj.CDN == "" {
```
rob05c: `&&` has higher precedence than `||` in Go, so it binds first. This is equivalent to `if (reqObj.CDNID != nil && reqObj.CDN == nil) || *reqObj.CDN == "" {`. Which panics on the dereference immediately below if the CDNID is nil, and `CDN` is not nil and the empty string. 

Needs to be `if reqObj.CDNID != nil && (reqObj.CDN == nil || *reqObj.CDN == "") { `.

Even if `&&` and `||` had equal precedence evaluated from the left, IMO it's still always better to parenthesize different logical operators, to make it obviously correct.

See: https://golang.org/ref/spec#Operator_precedence
Example: https://play.golang.org/p/L3FkgKISgFZ

traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go
rob05c: Actually, I was mistaken: this does need to exist, even with the check below. Because below doesn't account for the case if `CDNID == nil && *reqobj.CDN == ""`. 

I think this should be removed, _and_ the above check `if reqObj.CDN == nil && reqObj.CDNID == nil {` should be changed to `if reqObj.CDNID == nil && (reqObj.CDN == nil || *reqObj.CDN == "" {`.

I think that handles all the cases, and IMO makes it obvious, "if the ID is nil _and_ the name is either nil or empty, return an error".

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3468/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3472/
Test PASSed.

PR #3445, opened 2019-03-28
TO: Give a better error message when deleting an object being referenced

### LINKED ISSUE #2729: Improve Error Message when attempting to delete a Division that has assigned Regions
If you create a new Division, then assign a Region to it and then attempt to delete the Division, you get a _database access error_. We should provide a better error message in the client to improve the user experience. Something like, "Division <xyz> has Regions associated with it and can not be deleted."

### LINKED ISSUE #3410: Deleting a profile assigned to a server results in internal server error
ERROR:  update or delete on table "profile" violates foreign key constraint "fk_contentserver_atsprofile1" on table "server"

instead a 400 should be returned with a message like "can't delete profiles that are currently in use"

moltzaum:

Known Examples:
* deleting coordinate with a cachegroup
* deleting profile with a server (Fixes #3410)
* deleting division with assigned region (Fixes #2729)

Any endpoint that uses the generic cruder should be fixed (though, the disclaimer comment at the top of the function should also be read). An endpoint that doesn't use the generic cruder can use api.ParseDBError to get the same result.

-  Traffic Ops

* Test the specific cases mentioned in the issues.
* Run the integration tests.
* Analyze the comment and understand the assumptions it makes.
If the current assumptions are not reasonable the logic can instead be handled outside `api.ParseDBError`, even being privatized to the generic delete function.

I made a test for the divisions-regions issue. Divisions use the generic delete, so the test pretty much tests the generic function.

## Check all that apply

- [x] This PR includes new tests
- [x] This PR does not need to include documentation updates

### DIFF
--- a/traffic_ops/testing/api/v14/divisions_test.go
+++ b/traffic_ops/testing/api/v14/divisions_test.go
@@ -33,7 +33,13 @@ func TestDivisions(t *testing.T) {
 
 func TryToDeleteDivision(t *testing.T) {
 	division := testData.Divisions[0]
-	_, _, err := TOSession.DeleteDivisionByName(division.Name)
+
+	resp, _, err := TOSession.GetDivisionByName(division.Name)
+	if err != nil {
+		t.Errorf("cannot GET Division by name: %v - %v\n", division.Name, err)
+	}
+	division = resp[0]
+	_, _, err = TOSession.DeleteDivisionByID(division.ID)
 
 	if err == nil {
 		t.Errorf("should not be able to delete a division prematurely")
@@ -45,7 +51,7 @@ func TryToDeleteDivision(t *testing.T) {
 		return
 	}
 
-	if strings.Contains(err.Error(), "This division is currently used by regions.") {
+	if strings.Contains(err.Error(), "cannot delete division because it is being used by a region") {
 		return
 	}
 
--- a/traffic_ops/testing/api/v14/divisions_test.go
+++ b/traffic_ops/testing/api/v14/divisions_test.go
@@ -16,6 +16,7 @@ package v14
 */
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3412/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3414/
Test PASSed.

traffic_ops/testing/api/v14/divisions_test.go
```diff
@@ -16,19 +16,42 @@ package v14
...
+	WithObjs(t, []TCObj{Parameters, Divisions, Regions}, func() {
+		TryToDeleteDivision(t)
 		UpdateTestDivisions(t)
 		GetTestDivisions(t)
 	})
 }
 
+func TryToDeleteDivision(t *testing.T) {
+	division := testData.Divisions[0]
+	resp, _, err := TOSession.DeleteDivisionByName(division.Name)
```
rob05c: Doesn't compile, `resp` isn't used, needs to be `_, _, err := TOSession.DeleteDivisionByName(division.Name)`

traffic_ops/testing/api/v14/divisions_test.go
```diff
@@ -16,19 +16,42 @@ package v14
...
+		t.Errorf("should not be able to delete a division prematurely")
+		return
+	}
+
+	if strings.Contains(err.Error(), "Resource not found.") {
+		t.Errorf("division with name %v does not exist", division.Name)
+		return
+	}
+
+	if strings.Contains(err.Error(), "This division is currently used by regions.") {
```
rob05c: The test doen't compile, but it looks like this won't pass, needs to be `cannot delete division because it is being used by a region`.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3469/
Test FAILed.

traffic_ops/testing/api/v14/divisions_test.go
moltzaum: Looks like I need to use `DeleteRegionByID`. Should I use a status code instead of looking at the error message so that this is less brittle? A bad request does not necessarily indicate the error I'm actually testing for, so I'm not sure what the best way is.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3470/
Test PASSed.

PR #3452, opened 2019-04-01
Make cachekey plugin responsible for range header when caching_range_…

### LINKED ISSUE #3444: Range Requests are cached incorrectly when ignoring query strings and caching exact ranges
When a delivery service is configured with qstring_ignore = 1 and cache_range_request plugin enabled, byte range requests will be cached incorrectly. 

ATS only allows the cachekey to be set once per transaction, and both the cachekey plugin (for qstring_ignore=1) and the cache_range_request plugin are trying to modify the cachekey. As cachekey.so runs first, it determines the cachekey without respect for the byte range requested. When cache_range_requests.so runs, it tries to set the cache key a second time, but fails silently. 

Example:
```
GET http://customer.com/master.m3u8
Range: bytes=0-100
```

Results in a cache key of `http://customer.com/master.m3u8` which is missing the critical byte range value. When a viewer requests a different byte range from the same URL (ie. bytes=101-300), their request will match the existing cache key and they will be delivered the wrong content. 

Strangely enough, this includes the response headers from the original request so you can see a request for 101-300 answered with an actual range of 0-100.

See also #3393 and #1085 for some related issues in this area.

limited:
…requests

Fixes #3444

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] Other _________

Manual verification, set `qstring_ignore` to `1` and `cache_range_requests` to `2`. Verify in `remap.config` that the cachekey.so plugin has an option to cache the Range header
There are currently no automated tests for the contents of config file generation

## Check all that apply

- [ ] This PR includes tests

### DIFF
--- a/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
+++ b/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
@@ -2774,7 +2774,7 @@ sub remap_dot_config {
 				$mid_remap{ $ds->{org} } .= " \@plugin=header_rewrite.so \@pparam=" . $ds->{mid_hdr_rw_file};
 			}
 			if ( $ds->{qstring_ignore} == 1 ) {
-				$mid_remap{ $ds->{org} } .= UI::DeliveryService::get_qstring_ignore_remap(UI::DeliveryService::get_ats_major_version( $self, $server_obj ));
+			    	$mid_remap{ $ds->{org} } .= UI::DeliveryService::get_qstring_ignore_remap(UI::DeliveryService::get_ats_major_version( $self, $server_obj ), $ds->{range_request_handling});
 			}
 			if ( defined( $ds->{cacheurl} ) && $ds->{cacheurl} ne "" ) {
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3436/
Test PASSed.

mitchell852:
@limited - you were probably trying to be thorough by making changes in both the UI namespace and the API namespace but I "think" the UI namespace for config file generation is pretty much deprecated. so i'm wondering if that change you made in `traffic_ops/app/lib/UI/DeliveryService.pm` needs to be made in the api namespace in the appropriate place...

limited:
:-) 

From our internal code review:
From @PeterRyder: "For the record, I despise the API calling a UI function but this is better than adding that check to 3 places"

PR #3460, opened 2019-04-03
removes a ds type (DNSLive) that is not considered a steering type

mitchell852:

This PR removes a ds type (DNSLive) from the `IsSteering` convenience method. DNSLive is not considered a steering type. 

Fortunately, `isSteering` was only used in 2 places and because it was used in conjunction with isDNS(), this bug has never caused a problem...yet...

https://github.com/apache/trafficcontrol/blob/master/lib/go-tc/enum.go#L523
https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go#L195

- Traffic Ops

Running the current tests should ensure that this change does not cause a problem.

Run the TO unit tests -

cd trafficcontrol/traffic_ops/traffic_ops_golang
go test $(go list ./...)

Run the TO API tests - https://github.com/apache/trafficcontrol/blob/master/traffic_ops/testing/api/README.md

Documentation is not necessary for this bug fix.

- master (55690d7)
- 3.0.0
- 3.0.1 (RC2)

### DIFF
--- a/lib/go-tc/enum.go
+++ b/lib/go-tc/enum.go
@@ -511,8 +511,6 @@ func (t DSType) IsSteering() bool {
 	case DSTypeSteering:
 		fallthrough
 	case DSTypeClientSteering:
-		fallthrough
-	case DSTypeDNSLive:
 		return true
 	}
 	return false


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3456/
Test PASSed.

PR #3470, opened 2019-04-05
Add support for servercheck updates to go client

guzzijason:
Adding support for the `/api/1.3/servercheck` route to the go client, which was previously missing.

- This PR is not related to any Issue
- Documentation via godoc

- Traffic Control Client (go)
- Traffic Ops servercheck scripts (WIP, new go version to replace legacy perl)

Client test code used:
```
        var statusData tc.ServercheckNullable
        statusData.ID = server.ID
        statusData.Name = "DSCP"
        statusData.Value = 0
        _, err := session.UpdateCheckStatus(statusData)
        if err != nil {
            panic(err)
        }
```
No tests were included because no API changes were made.

### DIFF
--- a/traffic_ops/client/servercheck.go
+++ b/traffic_ops/client/servercheck.go
@@ -20,9 +20,10 @@ import (
 )
 
 const API_V13_SERVERCHECK = "/api/1.3/servercheck"
+const API_V13_SERVERCHECK_GET = "/api/1.3/servers/checks"
 
 // Update a Server Check Status
-func (to *Session) UpdateCheckStatus(status tc.ServercheckNullable) (*tc.ServercheckPostResponse, ReqInf, error) {
+func (to *Session) UpdateCheckStatus(status tc.ServercheckPostNullable) (*tc.ServercheckPostResponse, ReqInf, error) {
 	uri := API_V13_SERVERCHECK
 	var remoteAddr net.Addr
 	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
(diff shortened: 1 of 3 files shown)


asfgit:

dangogh:
add to whitelist

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3483/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3484/
Test FAILed.

dangogh:
sorry -- only had a cursory look at it before..   The code looks good,  but I'd like to see a test for it. It should be fairly straightforward:  in `traffic_ops/testing/api/v14`,  add a new `servercheck_test.go` and fill in json in `tc-fixtures.json`.   Simple test to make sure you can read the data when it goes into the table as well as any tests to check for incorrect input.  Plenty of other examples there.  I can help out with that if needed.

guzzijason:
I think I'll need to add a client function to read the data then - will probably be helpful for the test :)

traffic_ops/client/servercheck.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"encoding/json"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+)
+
+const API_V13_SERVERCHECK = "/api/1.3/servercheck"
+
+// Update a Server Check Status
+func (to *Session) UpdateCheckStatus(status tc.ServercheckNullable) (*tc.ServercheckPostResponse, error) {
```
rob05c: Mind changing this to return `ReqInf`? 

Should just be 

`func (to *Session) UpdateCheckStatus(status tc.ServercheckNullable) (*tc.ServercheckPostResponse, ReqInf, error) {`

```
reqInf, err = post(to, uri, jsonReq, &resp)
if err != nil {
		return nil, reqInf, err
	}
	return &resp, reqInf, nil
```

We're working to change all client funcs to return `ReqInf`. It has data that helps clients debug and return better error messages.

guzzijason: Sure, will do.

guzzijason:
So, I added a test, but I'm not 100% sure it's what is needed. Currently, the only use case is to update the check status, so it's pretty simplistic, I think. Also, I'm not sure the best way to handle the fact that the "extension" use is required to be used for the `servercheck` API route. I added that user to the test, but I doubt it will authenticate as-is.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3487/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3667/
Test PASSed.

ocket8888:
The API test for this route doesn't compile, load the data it needs to run, or run as the proper user. It also causes a runtime panic when run.

PR #3475, opened 2019-04-09
CIAB: Let auto-snapqueue and add-sslkeys be functions

Shihta:

auto-snapqueue and add-sslkeys are good features and worth reusing.
This commit changes them into separate functions.

This PR doesn't change the CIAB usage so I didn't attach the document.

- CDN in a Box

1. Start CIAB
2. Check the snapshot can be automatically took
3. Check the demo site can be accessed via `HTTPS` protocol.

### DIFF
--- a/infrastructure/cdn-in-a-box/traffic_ops/run-go.sh
+++ b/infrastructure/cdn-in-a-box/traffic_ops/run-go.sh
@@ -87,94 +87,12 @@ while true; do
 done
 
 ### Add SSL keys for demo1 delivery service
-demo1_sslkeys_verified=false
-demo1_version=1
-while [[ "$demo1_sslkeys_verified" = false ]]; do
-   while true; do
-     sslkeys_response=$(to-get "api/1.4/deliveryservices/xmlId/$ds_name/sslkeys?decode=true")
-     echo "CDN SSLKeys=$sslkeys_response"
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3495/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3497/
Test PASSed.

ocket8888:
You know, looking at the auto snap/queue code I'm starting to wonder if maybe we're pushing the limits of what Bash should be doing. Every iteration parses these JSON responses seven times and I can't help but think how much cleaner this would be if we could parse those into objects immediately and iterate over them with `for x in y` instead of a while loop with condition checking

Shihta:
I can try to improve that.
However, do you mean we should use another method instead of continually improving this piece of bash code?

ocket8888:
Yeah, I think maybe we're approaching a level of complexity that might be better handled by a Python script. Or a Perl script, I guess, but I think we're moving away from Perl as a project.

Shihta:
I think you are right and I have raised a discussion on slack.
I need more suggestion.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3557/
Test PASSed.

PR #3478, opened 2019-04-10
Mock Traffic Ops Server

ocket8888:

Adds a "Mock" Traffic Ops server that provides static, immutable, and deterministic responses to the same routes as the actual Traffic Ops API. The purpose of this is for testing API clients (e.g. ORT, which is the client I, personally ,am specifically interested in testing with this), since the API itself cannot be guaranteed to produce the same response to subsequent requests made to any given endpoint.

- Documentation
- Traffic Control Client (Go)

Please build and run the mock Traffic Ops server (see the provided README in `traffic_ops/client_tests`), and build and view the documentation (which can be found under "Tools"). No tests are provided because that feels like a dependency recursion... But more seriously because the intention is for this to BE a testing tool.

N/A


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3503/
Test PASSed.

traffic_ops/client_tests/cachegroups.go
```diff
@@ -0,0 +1,99 @@
...
+var MID_CACHEGROUP_LATITUDE = 0.0
+var MID_CACHEGROUP_LONGITUDE = 0.0
+var MID_CACHEGROUP_FALLBACK_TO_CLOSEST = true
+var MID_CACHEGROUP_LOCALIZATION_METHODS = []tc.LocalizationMethod{
+	tc.LocalizationMethodCZ,
+	tc.LocalizationMethodDeepCZ,
+	tc.LocalizationMethodGeo,
+}
+
+var CACHEGROUPS = []tc.CacheGroupNullable{
```
rawlinp: These would be the static test cases, but the issue here would be that they would have to be kept up to date with any new/deleted fields in the structs. I was kind of envisioning dynamic test cases too that would automatically stay up to date by using reflection to enumerate all of the current fields and populate them with randomly chosen (or statically-chosen value for strings, ints, floats, etc) values to emit from the API.

For example, I'm envisioning iterating over the struct fields and populating them based on what type the field is. We could just say "if it's a string type, use 'foo'. If it's a pointer to a string type, use a pointer to the string 'foo'. If it's an integer type, use 42." Or, we could say "if it's a string type, randomly generate a 8-char string to populate the field with. Or, generate a random float value for float fields". I'm not sure how difficult that would be or if it's even possible with Golang's reflection capabilities, but I think it would be really valuable to keep this Mock TO up to date automatically.

traffic_ops/client_tests/mock.go
```diff
@@ -0,0 +1,247 @@
...
+
+	server := &http.Server{
+		Addr:      fmt.Sprintf("%s:%d", listenAddr, port),
+		TLSConfig: &tls.Config{InsecureSkipVerify: false},
+	}
+
+	for i := API_MIN_MINOR_VERSION; i <= API_MAX_MINOR_VERSION; i += 1 {
+		v := "1." + strconv.Itoa(i)
+		log.Printf("Loading API v%s\n", v)
+		http.HandleFunc("/api/"+v+"/ping", ping)
```
rawlinp: I was also hoping this set of routes could be taken from the traffic_ops_golang code if possible, so that new routes that get added would also be automatically picked up by Mock TO. That's probably also another fairly difficult thing to wire into this, but I think it would also be pretty valuable.

traffic_ops/client_tests/cachegroups.go
ocket8888: Some of that behavior can lead to invalid or self-inconsistent fields, like `Type: "asdfg",` and `Latitude: 400.6`. We could make constraints by field, but at some point that becomes more complex than defining static defaults - and with essentially the same effect. Plus, making it dynamic/random could harm testability. At the least, I want this thing to be able to be used for tests of ORT, and that means it's beneficial to have static configfile output. But if profile/server/cdn names/ids and associated parameter names/ids/values/locations are random/dynamic then they could conflict with the static output of the configfiles. I don't think that's a problem today, but setting up a self-inconsistent testing environment seems like a bad idea to me. In order to make that deterministic, we'd need to build configfile generation logic into the mock server to be maintained separately from TO's configfile generation logic, and probably do the same for any tests using the mock server.

It's unfortunate because I share your concern about maintaining structure twice, but this seems like the lesser of the two evils, and less work to maintain.

traffic_ops/client_tests/mock.go
ocket8888: I have a branch somewhere where I was working on being able to do that kind of reflection - at least pulling the names of the routes if not their associated handlers. It's part of a PR for adding an `OPTIONS` method handler to `/api/` itself, but I don't believe the reviewer and I could ever agree on the appropriate response format.

docs/source/tools/mock_trafficops.rst
```diff
@@ -0,0 +1,99 @@
...
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+.. _mock-to:
+
+***************************
+The Mock Traffic Ops Server
+***************************
+.. versionadded:: 4.0.0
```
limited: Still on v3, probably 3.1 at this point

docs/source/tools/mock_trafficops.rst
```diff
@@ -0,0 +1,99 @@
...
+* The :ref:`to-api` produces non-deterministic output, especially for configuration files\ [#configfiles_issue]_. Because of this, no two requests to the same endpoint with the same data in the Traffic Ops Database can be guaranteed to produce the same output.
+
+If these factors are important to the tests for a client of the :ref:`to-api`, it is suggested that they instead utilize the Traffic Ops Mock server. The source code for this binary can be found in :file:`traffic_ops/client_tests`.
+
+Behavior
+========
+The Traffic Ops mock server **SHALL**
+
+* Serve all of the API routes supported by the :ref:`to-api`\ [#implemented_routes_disclaimer]_
+* Serve static responses that never change in structure or content
```
limited: One of the most valuable properties of a mock is the ability to easily simulate failures. Could we consider that a potential use for the mock TO? Specifically, can we have some routes that are hardcoded to specific types of failures?

traffic_ops/client_tests/configfiles.go
```diff
@@ -0,0 +1,365 @@
...
+CONFIG proxy.config.http.slow.log.threshold INT 10000
+CONFIG proxy.config.http.transaction_active_timeout_in INT 0
+CONFIG proxy.config.log.logfile_dir STRING /var/log/trafficserver
+CONFIG proxy.config.proxy_name STRING __FULL_HOSTNAME__
+CONFIG proxy.config.reverse_proxy.enabled INT 0
+CONFIG proxy.config.url_remap.remap_required INT 0
+`
+
+func edgeRecordsConfig(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Server", SERVER_STRING)
```
limited: Theres an awful lot of boilerplate repeated in these Config get functions (and the API functions as well). 

Could the duplication here be reduced?

traffic_ops/client_tests/cachegroups.go
limited: I like the suggestion. The more we can decouple the need to modify the mock TO, the easier (and more likely) it will be kept up to date. 

To keep the responses consistent across instances/versions of the mock server, instead of randomly generating perhaps we could derive the value of the field from properties of that field like name/type.

docs/source/tools/mock_trafficops.rst
ocket8888: So, I've been told conflicting things about this. Some people have told me that `master` is set to track the next major rev, some say the next minor.

docs/source/tools/mock_trafficops.rst
ocket8888: Could you describe a planned failure? I'm not sure what you mean by that.

traffic_ops/client_tests/configfiles.go
ocket8888: Probably. You know, originally `SERVER_STRING` didn't exist and I just had it as a string literal everywhere :^). The thing is, I've got this `common` function that sets `Content-Type: application/json` and `Server: Traffic Ops/{{VERSION}} (Mock)`. But sometimes you're sending plain text, so you only need the `Server` header, so I found it hard to justify replacing this one line that's everywhere with a different one line that just resolves to a one-line function. 

What'd be great is to put in some middleware, which I'm not really sure how to do. My Go-fu is weak.

docs/source/tools/mock_trafficops.rst
limited: Sure, ask there. I'll chime in on that thread

ocket8888:
> _"To keep the responses consistent across instances/versions of the mock server, instead of randomly generating perhaps we could derive the value of the field from properties of that field like name/type."_

The problem with that is it's going to have to look something like:
```go
if field.Name == "Latitude" {
    field.Value = 0.0
}
```
and I'm just not sure how that's meaningfully different than 
```go
Latitude: 0.0
```
Besides that the former is less likely to cause compilation errors. But imo, it'd be better to get a compilation error, so you immediately know what's wrong, because either way you need to change something if the field name changes from `Latitude` to `Foo`.

(I'm not sure why, but I couldn't respond to you directly, so I had to make a new comment)

traffic_ops/client_tests/cachegroups.go
ocket8888: > _"To keep the responses consistent across instances/versions of the mock server, instead of randomly generating perhaps we could derive the value of the field from properties of that field like name/type."_

The problem with that is it's going to have to look something like:
```go
if field.Name == "Latitude" {
    field.Value = 0.0
}
```
and I'm just not sure how that's meaningfully different than 
```go
Latitude: 0.0
```
Besides that the former is less likely to cause compilation errors. But imo, it'd be better to get a compilation error, so you immediately know what's wrong, because either way you need to change something if the field name changes from `Latitude` to `Foo`.

docs/source/tools/mock_trafficops.rst
limited: Something like a route which would return a 404 or malformed response (not JSON). I'm not totally sure how(unit test, integration test, etc...) you are planning to use this mock server, so it might not make sense. Here's an example of some negative test cases we could run against the ORT script:

* How does ORT handle a TO timeout?
* How does ORT handle getting a 404 from a route it thinks should exist?
* What happens if ORT gets not JSON (or not the expected format) somewhere (we've seen garbage end up in some of the ATS config files....) 
* What if ORT gets a remap activating a plugin that isn't present (something like this would be best handled in a unit test for the remap.config or plugin.config functions. I don't know if I'd put something this specific into an integration test).

limited:
> > _"To keep the responses consistent across instances/versions of the mock server, instead of randomly generating perhaps we could derive the value of the field from properties of that field like name/type."_
> 
> The problem with that is it's going to have to look something like:
> 
> ```go
> if field.Name == "Latitude" {
>     field.Value = 0.0
> }
> ```
> 
> and I'm just not sure how that's meaningfully different than
> 
> ```go
> Latitude: 0.0
> ```
> 
> Besides that the former is less likely to cause compilation errors. But imo, it'd be better to get a compilation error, so you immediately know what's wrong, because either way you need to change something if the field name changes from `Latitude` to `Foo`.
> 
> (I'm not sure why, but I couldn't respond to you directly, so I had to make a new comment)

What if it was instead (with equally weak Go-fu  reflection)
```
  for field: range fields {
    field.Name: CreateValue(field.Name, field.Type)
  }
```

traffic_ops/client_tests/configfiles.go
limited: This would cut way down on size of the code and should hold us for a fairly long time. I don't see ORT sending PUTS/POSTS to the config generation or API routes anytime soon. 

I was thinking something like: 
```
func edgeRecordsConfig(w, r) {
  sendRaw(w, r, []byte(EDGE_PROFILE_HEADER + EDGE_RECORDS_CONFIG))
}

func sendRaw(w, r, contents) {
  w.Header().Set("Server")
  if r.Method == http.MethodGet {
		w.Write(contents)
  } else {
		w.Header().Set("Allow", http.MethodGet)
		w.WriteHeader(http.StatusNotImplemented)
	}
}
```

traffic_ops/client_tests/cachegroups.go
limited: What if it was instead (with equally weak Go-fu reflection)

  for field: range fields {
    field.Name: CreateValue(field.Name, field.Type)
  }

docs/source/tools/mock_trafficops.rst
ocket8888: done

traffic_ops/client_tests/mock.go
ocket8888: Actually wait, part of the prerequisite work for that change has been merged, so I actually think that might be possible.

The next question, though, is how to implement that. Specific routes need specific handlers, so unless we want to implement another CRUDer (to which I vehemently object, because we have a huge amount of endpoints that don't use it) idk how much use that's gonna be.

ocket8888:
> What if it was instead (with equally weak Go-fu reflection)
> 
> ```go
>   for field: range fields {
>     field.Name: CreateValue(field.Name, field.Type)
>   }
> ```

Sure, the example `if` statement/block would be inside the `CreateValue` function in that case. Because you can't just give all numerics random values, or a specific, constant value. Because Ids won't line up, or types that aren't totally captured by their field types  can be wrong - e.g. QStringIgnore needs to be in the range [0,2], initialDispersion can't be 0 (or > 15?) etc.

traffic_ops/client_tests/cachegroups.go
rawlinp: That's a good point, I was mostly just handwaving away invalid values based upon the types, but I suppose we do have clients that would expect to get a valid value to parse into specific types (thinking about the Golang client here). Clients like the Python TO client and TP don't really parse the values into specific types that are different than the json types, so it doesn't really matter what value they get back (as long as the "primitive" types are the same).

I guess, what would the value in Mock TO be if it was just all 100% statically returned values? At that point, we don't really need to use the go-tc structs; we can just create static JSON files that are served up by a simple http server in a simple directory structure:
```
api/
    1.2/
        cachegroups <---static json file
        asns <--- static json file
```
where `curl http://localhost:30003/api/1.2/cachegroups` would return that static json. It would essentially be accomplishing the same thing, except for the go struct's marshalling/unmarshalling code (which I think is already well-tested elsewhere).

If Mock TO isn't integrated enough with Real TO to not require maintenance every time real TO changes something, I think it will lose its value quickly as it gets more and more out of date.

limited: Right, its an impossible line to walk- trying to build a server that both understands the context of a specific field, but does not require updates when new fields are added. 

I'm still fuzzy on the purpose of this code. Have you considered writing a UT suite against ORT and then mocking out the specific calls it will make on a function by function basis? When new functionality gets added to ORT, a new test gets added as well.

ocket8888: > _"we can just create static JSON files that are served up by a simple http server in a simple directory structure"_

Yeah we could. My concern with that is double-maintenance of values. Like, right now, I have the parent of the edge cache group set to a pointer to the name of the mid cache group. That way you only have to change it once. If they're in JSON, we lose the ability to make those connections, unless we're going to do text replacements which once again more or less ends with us having all the values statically defined in the Go code anyway.

> _"Have you considered writing a UT suite against ORT and then mocking out the specific calls it will make on a function by function basis?"_
That's more or less what I'm doing. Thing is, we have ORT.pl and my baby, ORT.py. I want the tests to be independent of implementation, i.e. "after a SYNCDS run of an ORT implementation pointed at the Mock TO, the exact file system and process table changes should be this:". Building tests into the Perl code limits my ability to to test Python, and vice-versa. Basically the entire reason I wanted a mock TO in the first place was to figure out what gaps there still are between the two implementations.

What I've done here is build a fake HTTP service that ORT requires to run, which is the first step in writing the actual tests. By separating them, I can make a more general-purpose testing tool that could be used by more general-purpose clients, like the Apache-TrafficControl Python package and the Golang client @rawlinp mentioned.

docs/source/tools/mock_trafficops.rst
ocket8888: I had designed this with only positive tests in mind. Negative tests are, of course, valuable, but I'm not sure how to intertwine the two while the server is separate from the test suite. How does the test suite know which routes are supposed to 404? If we build tests into the mock server, it destroys the ability to use it generically, because now it's only for testing something specific.

traffic_ops/client_tests/configfiles.go
ocket8888: that's a good idea. That should actually probably be `http.StatusMethodNotAllowed`, now that I think about it. Because yeah, nothing should be doing anything to a config file except GET(/HEAD/OPTIONS ?).

traffic_ops/client_tests/cachegroups.go
rawlinp: > If they're in JSON, we lose the ability to make those connections, unless we're going to do text replacements which once again more or less ends with us having all the values statically defined in the Go code anyway.

We'd only need to make the connection once when creating the static files, and I'm not sure why the connections would ever need to be updated after that. An upside of just using static json files instead would be that you could just query a running TO instance, find/replace stuff you think should be redacted, and the static files would be ready to serve to mock the entire API (instead of writing a bunch of code). If you wanted to test specific changes to the static responses, all you'd have to do is manually update the static files instead of changing hard-coded values in the server, recompiling, and restarting the server.

ocket8888: I'm not super worried about restarting the server, since this should never be any part of production.

As far as updates, you _probably_ won't need to update them _after an exhaustive set_ is determined. Like we probably want a ton more cache servers so that our tests cover a bunch of mutually-exclusive cases. Which means updating cache groups and parents and profiles and parameters etc. etc. Then also, whenever something changes like "oh this thing that used to allow arbitrary text now has a restricted format" you would need to grep/sed through all the JSON. Which may or may not be a big deal. I think it is, but I could be convinced otherwise.

rawlinp: > I'm not super worried about restarting the server, since this should never be any part of production.

That's not what I'm worried about either. I was just pointing out that it's easier to just modify a file than it is to modify some source code, recompile, and restart a server for your changes to be realized.

With Mock TO I was mostly interested in the idea of it being able to stay up to date automatically with Real TO by reusing the same Go structs and other traffic_ops_golang code, but if it's all going to be static then I don't see a really compelling reason why it has to be Go when it can just be static json files and a simple http server.

traffic_ops/client_tests/mock.go
rawlinp: Automatically picking up routes from traffic_ops_golang wouldn't really make sense for the "static response" use case, since each "static response" would require its own handler. If there was a single "generic" handler function that was capable of just taking any struct, using reflection to dynamically populate all of the struct fields, and marshalling it into json for the response, then maybe we would just have to include an instance of the struct type in the `Route` structs that traffic_ops_golang populates in the `Routes` function.

traffic_ops/client_tests/cachegroups.go
ocket8888: The only reasons I really have are because it will allow us to to always provide the same "alerts" objects as a real instance - without ever needing to update -, and re-use of values that makes adding new endpoints easier and is hard to replicate with JSON. I don't really feel strongly about JSON vs Go structs (because secretly I hate all of those global declarations), so if you feel the benefits of using JSON outweigh the benefits of the static Go structs I'll change it.

limited: If there will likely be repetition between the endpoints (for example if you have 3 DS'), then staying with the structs will reduce duplication. You could declare a base struct then copy it and modify as needed. When we add more DS fields, only the base needs to change. In contrast, if we had 3 separate DS JSON files, each one of those would need an update- tedious.

ocket8888: I'm gonna put this on the mailing list to see if we can reach some kind of consensus.

rawlinp: There's a reason why I haven't explicitly -1'd this PR and have just been commenting on it. I was hoping to get some features out of it that I had originally proposed, but that doesn't necessarily have to hold up the statically-defined implementation. I do wish we had a representative dataset like Kabletown 2.0 that we could leverage for the static-Mock TO use case, but maybe we'll get there eventually.

ocket8888: Kabletown 2.0 would be great - but for it to be useful the API responses need to be deterministic IMO. Until things like The Duplicate Origin Problem<sup>TM</sup> are solved, a static dataset isn't enough because you can't predict what the result of a request will be.

rawlinp: Since we'd control the static Kabletown 2.0 dataset, we could simply keep the origins unique, which would mean that the API responses would be deterministic.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3596/
Test PASSed.

ocket8888:
I think the components that would benefit from this already have better solutions.

Issue #3107, opened 2018-12-11
/api/1.3/origins has bad interactions with STEERING Delivery Services


ocket8888:
Despite that an origin on a STEERING (or CLIENT_STEERING) Delivery Service is a more or less meaningless notion, it is perfectly acceptable to create an origin for those Delivery Services. Worse, creating an origin with (or modifying an existing one such that) `"isPrimary": true` on said Delivery Services - making an origin the 'primary' origin of that Delivery Service - will not only succeed nonsensically, but will also erroneously reply with a response body that shows `"isPrimary": true` is set on the origin, when it in fact is not. Further `GET`s to the `api/1.4/origins` endpoint will show `"isPrimary": false"`.

The API should return a `400 BAD REQUEST` - or better yet a `422 UNPROCESSABLE ENTITY` - when a user attempts to create an origin assigned to a STEERING Delivery Service.

rawlinp:
This is not really a problem with just the Origins API. For any endpoint that returns read-only fields (usually name fields like `profileName`, `fooName`, etc.), you can pass a value for those in a PUT/POST request, and the API will return back the value you requested in the response. However, a subsequent GET will contain the _correct_ things in those read-only fields because they will actually be populated with a SQL SELECT statement. The problem is that most of our traffic_ops_golang APIs don't properly populate those read-only fields to return in POST/PUT response by doing a SQL SELECT after doing the SQL INSERT/UPDATE. Ideally this would be fixable in the framework rather than requiring changes on a per-handler basis.

rawlinp:
related to https://github.com/apache/trafficcontrol/issues/2494

mitchell852:
since 1.x will soon be removed. can this be closed @ocket8888 ?

ocket8888:
No, the problem is not version-specific

mitchell852:
> No, the problem is not version-specific

can u change issue title?

PR #3484, opened 2019-04-12
TO: "info" struct tag example for better generic endpoint handling

moltzaum:
## What does this PR do?

This PR is related to the joined-fields issue: #2494

The issue of joined-fields was also briefly mentioned in #3107, where it is said that "ideally this would be fixable in the framework rather than requiring changes on a per-handler basis".

This PR implements joined fields for ASNs in a generic fashion. The reason I have only implemented the ASN handler is because replacing the current generic implementation with the one I did requires a change in struct tags for every endpoint it affects. I don't mind doing this, but I want to more or less present what I have now for potential feedback.

**How would the struct tags change?**

```
type ASNNullable struct {
	ASN          *int       `json:"asn"          db:"asn"`
	Cachegroup   *string    `json:"cachegroup"   db:"cachegroup"`
	CachegroupID *int       `json:"cachegroupId" db:"cachegroup_id"`
	ID           *int       `json:"id"           db:"id"`
	LastUpdated  *TimeNoMod `json:"lastUpdated"  db:"last_updated"`
}
```
```
type ASNNullable struct {
	ASN          *int       `json:"asn"`
	Cachegroup   *string    `json:"cachegroup"   info:"join"`
	CachegroupID *int       `json:"cachegroupId"`
	ID           *int       `json:"id"           info:"key"`
	LastUpdated  *TimeNoMod `json:"lastUpdated"  info:"lastUpdated"`
}
```

**What does the info tag do?**

It allows us to say `fields["join"]` and get all fields in the struct that have the info struct tag of "join".
Any value may be used, and for now I just specify "join", "key", and "lastUpdated".

**What advantage do we have with info tags over db tags?**

db struct tags are used with `NamedQuery`, `NamedExec`, and `StructScan`. In sql queries we may use syntax of the form `:field` to reference a field with `db:"field"`. The db tags use reflection to find the fields.

For a single endpoint implementation there is no need to use a named query. I believe _some_ endpoints have db tags because of copy-pasta, then used the `StructScan` for convenience. However, the main user of the named exec are the generic methods that greatly benefit from the reflection because the cruder only gets to work with interfaces.

The info tag gives us more control over the result of the reflection. Instead of just using the result to replace a name for a value in query, we can also use it to scan items into, since the sql database functions use interfaces (holding pointers of any type).

It is relevant to note that the named functions we are currently using is from `sqlx`. If we no longer use these functions anymore there is potential to remove a dependency that we don't need.

**What endpoints could this affect**

1) All endpoints in `asns`, `deliveryservices`, `regions`, `phys_locations`, `servers`, `tenants` from #2494  
2) Endpoints that did not use the generic methods in order to include join fields properly (I'm not sure specifically what endpoints). That being said, where tenancy is needed there is still no generic method created yet (methods exist, but the `GenericCreate`, `GenericUpdate`, etc. does not know how to check tenancy).

**Additional benefits beyond just replacing the db tag**

In our code we have methods like `SetLastUpdated`, `GetKeyFieldInfo`, etc. `SetLastUpdated` is no longer needed because with the info tag, we have recieved the pointer and scan into it. Some instances of `SetKeys` will also be removed for the same reason. I think in the future it would be beneficial to consider the complete removal of a struct's implementation for how it handles keys since this can simplify the logic and reduce boilerplate.

- Traffic Ops

This bug has been persistent since the affected endpoints were rewritten in go.

Although there are some tests, I haven't tested odd cases yet.
The documentation I've included is GoDoc

  - I'm not _fully_ sure when a changelog is necessary.
  - These changes are mostly internal, but also affect the visibility of certain joined fields.

### DIFF
--- a/lib/go-tc/asns.go
+++ b/lib/go-tc/asns.go
@@ -71,7 +71,7 @@ type ASNNullable struct {
 
 	// Related cachegroup name
 	//
-	Cachegroup *string `json:"cachegroup" db:"cachegroup" info:"joinz"`
+	Cachegroup *string `json:"cachegroup" db:"cachegroup" info:"join"`
 
 	// Related cachegroup id
 	//
--- a/lib/go-tc/asns.go
+++ b/lib/go-tc/asns.go
@@ -71,7 +71,7 @@ type ASNNullable struct {
 
 	// Related cachegroup name
 	//
-	Cachegroup *string `json:"cachegroup" db:"cachegroup"`
+	Cachegroup *string `json:"cachegroup" db:"cachegroup" info:"joinz"`
 
 	// Related cachegroup id
 	//
@@ -80,11 +80,11 @@ type ASNNullable struct {
 	// ID of the ASN
 	//
 	// required: true
-	ID *int `json:"id" db:"id"`
+	ID *int `json:"id" db:"id" info:"key"`
 
 	// LastUpdated
 	//
-	LastUpdated *TimeNoMod `json:"lastUpdated" db:"last_updated"`
+	LastUpdated *TimeNoMod `json:"lastUpdated" db:"last_updated" info:"lastUpdated"`
 }
 
 type ASNsV11 struct {
--- a/traffic_ops/traffic_ops_golang/api/api.go
+++ b/traffic_ops/traffic_ops_golang/api/api.go
@@ -284,6 +284,7 @@ func Parse(r io.Reader, tx *sql.Tx, v ParseValidator) error {
 type APIInfo struct {
 	Params    map[string]string
 	IntParams map[string]int
+	Fields    StructFields
 	User      *auth.CurrentUser
 	ReqID     uint64
 	Tx        *sqlx.Tx
--- a/traffic_ops/traffic_ops_golang/api/shared_handlers.go
+++ b/traffic_ops/traffic_ops_golang/api/shared_handlers.go
@@ -160,6 +160,7 @@ func UpdateHandler(updater Updater) http.HandlerFunc {
 		}
 		defer inf.Close()
 
+		// Every object contains APIInfo as a package of data that is often helpful while implementing endpoints.
 		interfacePtr := reflect.ValueOf(updater)
 		if interfacePtr.Kind() != reflect.Ptr {
 			HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("reflect: can only indirect from a pointer"))
@@ -167,9 +168,17 @@ func UpdateHandler(updater Updater) http.HandlerFunc {
 		}
 		objectType := reflect.Indirect(interfacePtr).Type()
 		obj := reflect.New(objectType).Interface().(Updater)
+
+		var err error
+		inf.Fields, err = GetFieldInfo(obj)
+		if err != nil {
+			HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, err)
+			return
+		}
 		obj.SetInfo(inf)
 
-		if err := decodeAndValidateRequestBody(r, obj); err != nil {
+		// Verifies JSON syntax is correct and calls Validate
+		if err = decodeAndValidateRequestBody(r, obj); err != nil {
 			HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, err, nil)
 			return
 		}
@@ -325,7 +334,16 @@ func CreateHandler(creator Creator) http.HandlerFunc {
 		obj := reflect.New(objectType).Interface().(Creator)
 		obj.SetInfo(inf)
 
-		err := decodeAndValidateRequestBody(r, obj)
+		var err error
+		inf.Fields, err = GetFieldInfo(obj)
+		if err != nil {
+			HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, err)
+			return
+		}
+		obj.SetInfo(inf)
+
+		// Verifies JSON syntax is correct and calls Validate
+		err = decodeAndValidateRequestBody(r, obj)
 		if err != nil {
 			HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, err, nil)
 			return
--- a/traffic_ops/traffic_ops_golang/api/struct_reflect.go
+++ b/traffic_ops/traffic_ops_golang/api/struct_reflect.go
@@ -1,7 +1,7 @@
 package api
 
 import (
-	"bufio"
+	"database/sql"
 	"errors"
 	"reflect"
 )
@@ -17,8 +17,14 @@ func allFieldsExported(obj interface{}) bool {
 	return true
 }
 
+// NotMarked is a key value for the `StructFields` map returned.
+// It refers to fields that had no "info" tag
+const NotMarked = ""
+
 // AllFields is a key value for the `StructFields` map returned.
-const AllFields = ""
+// It refers to all fields, regardless of whether or not they
+// have an info tag.
+const AllFields = "all"
 
 // StructFields is a map from an "info" struct tag to a list of pointers.
 // It can be obtained by using the `GetFieldInfo` function.
@@ -73,21 +79,25 @@ func GetFieldInfo(obj interface{}) (StructFields, error) {
 
 	// infoTag is captured by the append function, and
 	// can be updated in the main loop.
-	var infoTag string = ""
+	var infoTag string = NotMarked
 
 	// AppendToResult appends the passed argument (presumably a pointer)
 	// to the list of fields. Any info tag will create a new key, value
 	// pair in the map if it doesn't exist already.
 	AppendToResult := func(ptr ...interface{}) StructFields {
-		if infoTag != "" {
-			fields[infoTag] = append(fields[infoTag], ptr...)
+		if infoTag == AllFields {
+			// error?
+			// just don't count it?
 		}
+		fields[infoTag] = append(fields[infoTag], ptr...)
 		fields[AllFields] = append(fields[AllFields], ptr...)
 		return fields
 	}
 
 	// if obj implements scanner, read obj as atomic value
-	if _, ok := obj.(bufio.Scanner); ok {
+	// I'm unsure if this is needed, or what scanner to do. TODO: Update after testing
+	//if _, ok := obj.(bufio.Scanner); ok {
+	if _, ok := obj.(sql.Scanner); ok {
 		return AppendToResult(obj), nil
 	}
 
@@ -141,6 +151,9 @@ func GetFieldInfo(obj interface{}) (StructFields, error) {
 		field := val.Field(i)
 
 		infoTag = Struct.Field(i).Tag.Get("info")
+		if infoTag == "-" { // skip!
+			continue
+		}
 
 		switch field.Kind() {
 		case reflect.Struct:
--- a/traffic_ops/traffic_ops_golang/api/struct_reflect.go
+++ b/traffic_ops/traffic_ops_golang/api/struct_reflect.go
@@ -1,0 +1,170 @@
+package api
+
+import (
+	"bufio"
+	"errors"
+	"reflect"
+)
+
+// allExportedFields returns true if ALL fields are exported.
+// The input is expected to be a struct pointer.
+func allFieldsExported(obj interface{}) bool {
+	for i := 0; i < reflect.ValueOf(obj).Elem().NumField(); i++ {
+		if !reflect.ValueOf(obj).Elem().Field(i).CanInterface() {
+			return false
+		}
+	}
+	return true
+}
+
+// AllFields is a key value for the `StructFields` map returned.
+const AllFields = ""
+
+// StructFields is a map from an "info" struct tag to a list of pointers.
+// It can be obtained by using the `GetFieldInfo` function.
+type StructFields map[string][]interface{}
+
+// GetFieldInfo returns a list of pointers to fields in the struct.
+//
+// An error is returned for structs that have non-exported fields or if
+// a non-pointer is passed. It is not an error to pass a pointer to a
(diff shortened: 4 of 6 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3514/
Test FAILed.

moltzaum:
retest this please

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3515/
Test PASSed.

rawlinp:
So, my initial thoughts on this are that it's using a lot of reflection to mainly just get a set of fields from the struct to scan into when doing a SQL `RETURNING ...` statement. Ideally, we should be trying to use reflection only as a last ditch effort when the problem can't reasonably be solved without it. Originally I was thinking along the lines of something similar to 
```
func (v *TOASNV11) NewReadObj() interface{} { return &tc.ASNNullable{} }
```
where `TOASNV11` has a method like `NewReturningFields` that returns pointers to the fields that would be scanned into from the `RETURNING` clause.

However, I'm not sure we can utilize the `RETURNING` clause in all cases, since IIUC the expression can only contain columns from the table that was just inserted/updated/deleted. So, we should most likely be following up `INSERT`/`UPDATE`s with `SELECT`s, ideally just reusing the main `SelectQuery` that each struct already implements. TC is a read-heavy system, so I wouldn't be worried about doing extra `SELECT`s after every `INSERT`/`UPDATE`. Utilizing the `SelectQuery` methods like this would save us from having to do reflection or adding more methods like `NewReturningFields` that I described above. What do you think about doing it that way instead?

Here is some relevant info about `RETURNING` clauses (which I might be misunderstanding) https://www.postgresql.org/docs/9.6/sql-insert.html:
> The optional RETURNING clause causes INSERT to compute and return value(s) based on each row actually inserted (or updated, if an ON CONFLICT DO UPDATE clause was used). This is primarily useful for obtaining values that were supplied by defaults, such as a serial sequence number. However, any expression using the table's columns is allowed.

> output_expression
An expression to be computed and returned by the INSERT command after each row is inserted or updated. The expression can use any column names of the table named by table_name. Write * to return all columns of the inserted or updated row(s).

moltzaum:
> So, my initial thoughts on this are that it's using a lot of reflection to mainly just get a set of fields from the struct to scan into when doing a SQL `RETURNING` statement.

I agree. Doing all of this just for `LastUpdated` and join fields is too much overhead. However, this method works for _all_ fields. We can use it to read into an entire object, we can use it to specify PUT and DELETE ids, etc;. If we use it well, it can simultaneously replace the `db` tag and boilerplate CRUDer functions together. It also is not difficult to add other categories of fields if the need arises. For instance, we could make a new tag specifically for non-numeric ids.

We already have cost of reflection for the `db` tag, but sometimes its use has been overlooked because the named queries are easier to view as general methods with little to no difference in performance cost. I'm essentially hoping to rework the use of the `db` tag in a way that suits our needs better since the `db` tags only allow us to associate a field with something we put into a query -- we can't write back to a field using `db`.

> I'm not sure we can utilize the RETURNING clause in all cases, since IIUC the expression can only contain columns from the table that was just inserted/updated/deleted.  

Its true that in some cases, having another select statement may be unavoidable. However you can return a subquery that isn't related to the main table at all. I'm not really sure what you're getting at though since I don't think how we return items (within sql) relates to the main question: "how do we update certain fields 'generically' in go, in a way that can work easily for every/most endpoints". 

I'm not fond of creating a method `NewReturningFields` because I think it adds to a framework in  a way that makes it more difficult to maintain. Between all the methods we already have, it is not immediately obvious how they are called. I like json tags because if you understand the semantics behind each tag, you only have to look in one place (the struct definition) to understand the majority of the behavior of the endpoint.

rob05c:
>I'm not sure we can utilize the RETURNING clause in all cases, since IIUC the expression can only contain columns from the table that was just inserted/updated/deleted

We can, it's a little ugly, but it's doable with a CTE:

```
WITH i as (
INSERT INTO t (c) VALUES ('v') RETURNING id
)
SELECT c2 from t2 where t2.t_id = i.id FROM rows
```

rob05c:
I'll say, I like the general idea of this. I'm not a fan of Reflection in general, but we're already using it via the massive `sqlx` dependency, we just don't have to look at it. I think this gives us more control, and it looks more powerful than what we use in `sqlx`.

It's clearly a WIP, but I think it has potential. The biggest benefit I see, is if we can reduce some of the CRUDer boilerplate for each endpoint, and encapsulate it, either in the CRUDer (shared_handler) stuff or the GenericX stuff. Right now, it looks like it's involved in the endpoints themselves, but I think we could put it in the framework funcs. Then maybe with a little self-contained Reflection, we could greatly simplify the API framework boilerplate. I could be mistaken. But IMO it has potential and is worth investigating. My 2¢.

rawlinp:
> I'm not fond of creating a method NewReturningFields because I think it adds to a framework in a way that makes it more difficult to maintain. Between all the methods we already have, it is not immediately obvious how they are called. I like json tags because if you understand the semantics behind each tag, you only have to look in one place (the struct definition) to understand the majority of the behavior of the endpoint.
EDIT: I also don't like NewReturningFields because I think it is odd for a concrete function to return an interface for the purpose of being used in a 'generic' function. It just seems a little backwards to me. The database Scan function accepts a list of interfaces representing points, which makes me think making a function simply to get a list of interfaces representing pointers is the way to approach this problem.

Sorry if I wasn't clear, but that is **not** what I was proposing. This is what I said:
> Utilizing the SelectQuery methods like this would save us from having to do reflection or adding more methods like NewReturningFields that I described above

Basically, what I'm saying is that instead of trying to do an `UPDATE...RETURNING` sql statement, we would just always follow up the `UPDATE` with a separate `SELECT` using the `SelectQuery` methods that already exist -- the same queries used when doing a GET. `NewReturningFields` would not be required. It would just read the results of the `SelectQuery()` into the `NewReadObj()` result.

moltzaum:
@rob05c 
> Right now, it looks like it's involved in the endpoints themselves, but I think we could put it in the framework funcs.

Yes, to start I implemented it in a single endpoint. The code I made can be copied to the `GenericCreate` and related functions to yield the same result for all endpoints (after adjusting json tags for affected endpoints).

@rawlinp 
> It would just read the results of the SelectQuery() into the NewReadObj() result.
Oh, that makes a lot of sense actually. Would you be against doing it this way though? I think it still has a lot of benefits.

rawlinp:
Let's take the `Identifier` interface for example. That's an interface for a struct to implement that tells the "generic handlers" or whatever how to get/set the ID of the object. If I'm adding a new API with a new struct, I know I have to implement that interface in order to use some of the "generic handler" functionality. If I tried to use the generic handlers without implementing that interface, my code probably wouldn't even compile. If we somehow moved that information into the struct tags, then I'd have to know how to implement the struct tags for my new struct in order for everything to work, and I wouldn't know if I was using stuff incorrectly until it panicked at runtime. It would seem like moving the boilerplate information into the struct tags instead of just keeping it in boilerplate methods.

That's kinda why I'm leaning towards just having the `GenericUpdate` and `GenericCreate` functions check if the value implements `GenericReader`. If it implements `GenericReader`, go ahead and `SELECT` the result after doing the `UPDATE`. That would probably even allow us to get rid of `SetLastUpdated` and `SetKeys`, since we'd just always be reading the full result back into the struct.

moltzaum:
> Let's take the Identifier interface for example. That's an interface for a struct to implement that tells the "generic handlers" or whatever how to get/set the ID of the object. If I'm adding a new API with a new struct, I know I have to implement that interface in order to use some of the "generic handler" functionality. If I tried to use the generic handlers without implementing that interface, my code probably wouldn't even compile. If we somehow moved that information into the struct tags, then I'd have to know how to implement the struct tags for my new struct in order for everything to work, and I wouldn't know if I was using stuff incorrectly until it panicked at runtime. It would seem like moving the boilerplate information into the struct tags instead of just keeping it in boilerplate methods.

> That's kinda why I'm leaning towards just having the GenericUpdate and GenericCreate functions check if the value implements GenericReader. If it implements GenericReader, go ahead and SELECT the result after doing the UPDATE. That would probably even allow us to get rid of SetLastUpdated and SetKeys, since we'd just always be reading the full result back into the struct.

I meant to respond to this sooner, but anyway..

In my opinion learning the meaning of struct tags is much, _much_ simpler than learning how to implement the methods. I like the idea of struct tags because you can view them all in one place, whereas with the implementing methods the functions might be scattered around the file. With struct tags each implementor can decrease the lines of code that need to be added, making reviews easier as well. 

Although I can see where in some cases there might be some additional compile-time checks, a lot of the implementing methods use the `interface{}` struct, which in itself causes similar issues. If struct tags are used incorrectly we'd have to adjust the sql query. Even with the `db` tags you might make a typo. The `ParamColumns` function also works in a similar way, where mistakes get caught at runtime.

rawlinp:
These are some of the general cons I see with using reflection with struct tags to generate/operate SQL queries:
- potential for typos in the struct tags
- potential for not having the _right_ struct tags
- encoding SQL statements/operations into the struct tags, instead of being able to write and read pure SQL
- might work ok for simple structs that map nicely to the DB tables, but won't work for complex structs that require updates across multiple tables that aren't straightforward
- due to ^ that, I think it would incentivize developers to write APIs that map very closely to the DB tables, and we'll end up with a clunkier API because of it
- using reflection with struct tags will become yet another path for API implementation, furthering inconsistency in implementations across the API

Rob Pike, co-designer of Go, writing about Go reflection:
> It's a powerful tool that should be used with care and avoided unless strictly necessary.

Do you think this reflection is _strictly necessary_? I think we can avoid it here.

moltzaum:
To clarify, I am not suggesting generating SQL. The SQL operations we do generate won't be affected:

`where, orderBy, queryValues, errs := dbhelpers.BuildWhereAndOrderBy(val.APIInfo().Params, val.ParamColumns())`

Beyond that, the implementor would still write pure sql. The implementor would need to make sure the order works, but it is an order that we already follow for the vast majority of our sql statements when we write them. And this doesn't just work with simple structs. For instance, the `key` is not specifically tied to a particular type (which might be helpful if we switch to natural keys), nor is it required to be the only `key`.

Here is an instance where GetKeys returns multiple keys:
```
func (st TOSteeringTargetV11) GetKeys() (map[string]interface{}, bool) {
	keys := map[string]interface{}{}
	valid := true
	if st.DeliveryServiceID == nil {
		keys["deliveryservice"] = 0
		valid = false
	} else {
		keys["deliveryservice"] = int(*st.DeliveryServiceID)
	}
	if st.TargetID == nil {
		keys["target"] = 0
		valid = false
	} else {
		keys["target"] = int(*st.TargetID)
	}
	return keys, valid
}
```
Featuring the cons you mentioned of runtime checking and misspelling.

vs

```
type SteeringTargetNullable struct {
	DeliveryService   *DeliveryServiceName `json:"deliveryService"`
	DeliveryServiceID *uint64              `json:"deliveryServiceId" info:"key"`
	Target            *DeliveryServiceName `json:"target"`
	TargetID          *uint64              `json:"targetId" info:"key"`
	Type              *string              `json:"type"`
	TypeID            *int                 `json:"typeId"`
	Value             *util.JSONIntStr     `json:"value"`
}
```
I think the difference from the implementor's view speaks for itself.
But, as a brief reminder, _this_ PR in particular only affects joined fields.

> using reflection with struct tags will become yet another path for API implementation, furthering inconsistency in implementations across the API

I know.. but I think this PR has the potential to join all the affected endpoints with the endpoints that currently use the generic methods. In a way, we are simultaneously removing inconsistencies as well. I'm not just making a new method that each implementor calls, but some implementors can fully replace their own code by calling the generic functions.

> Do you think this reflection is strictly necessary? I think we can avoid it here.

Reflection is never _really_ necessary, but I think it fits well here. This is a question that we can also ask about the `db` tag. I think it is more of a question of whether or not we find it worth it to have our own internal implementation that can do 2 things, or an external implementation that only does 1.

rawlinp:
> I think the difference from the implementor's view speaks for itself.

In terms of what? Lines of code? The `info` tags hide all the added reflection complexity behind a curtain, whereas the GetKeys() method (and it's corresponding interface) is clear and easy to follow. If I just see non-standard tags on a struct, I have no way to figure out how those tags are even used or how they are _supposed_ to be used.

I don't really see things like `GetKeys()` and `NewReadObj()` as a problem that needs fixing. Those methods are straightforward to implement in order to get your structs to work in the framework, and they don't really require much code at all. For solving this "joined field" problem, I don't think we need to upend the current framework. Rather than trying to do an update+read in the same query, I think it's acceptable to just reuse the existing "read" functionality to follow up an update with a read. I believe that could be easily implemented in the framework as-is.

moltzaum:
Yes, in terms of lines of code.

You expressed some concern for some of the cons that come with reflection, and these are cons that propagate with each implementation method. With the GetKeys implementation in particular we have the con of empty interfaces, potentially mixing up `int` with `*int`, or making a typo in the map. If we made one method and make unit tests for it, we'd have better coverage than testing GetKeys inside the shared handler like we currently do: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/api/shared_handlers.go#L200-L206

It isn't just reducing the lines of code, but being able to make it common enough at a high-enough level that it applies to everything.

If you understand `GetKeys`, you already understand what I want to do with `info:"keys"`. Each is the same topic given a different name. The info tag is more or less the declarative way to do it. I don't think it adds an additional learning curve. I planned on adding documentation for every info tag.

____

I'm trying to get an example of reusing the read functionality for the create. As of now the create method returns `[]interface{}` like the Read function, and I need to find out how to filter the Read so that I only get one.

moltzaum:
I have made a separate attempt here: https://github.com/apache/trafficcontrol/pull/3580
Since it solves the same main issue as this PR, I'm closing this.

PR #3493, opened 2019-04-16
Move TO DS nonversioned logic to separate file.

rob05c:

Puts all the logic in deliveryservices.go, and makes v12.go, v13.go,
v14.go purely boilerplate.

This should make it significantly easier to add new versions: you'll
only have to copy the previous v13.go file, rename v13->v14, and
add your logic to the main deliveryservices.go

Also adds a section on adding new versions to the README.md.

No documentation because no interface change, just moving code symbols around. It does add new documentation in the form of the README explaining how to increment the version.

No tests, no new functionality, just moving symbols. Covered by existing tests, existing unit and API tests pass.

- Traffic Ops

Build TO, run TO, test changing a DS, run tests.

Not a bug fix.


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3530/
Test PASSed.

rawlinp:
This code reorganization probably makes sense, but I still don't think we should be moving forward with minor versioning taking this approach.

The whole process of receiving a 1.2 request, "upgrading" it into a 1.4 request by defaulting any 1.4 fields, and handling it from then on as if it had been a 1.4 request all along is flawed.

For example, a 1.2 client will wipe out (default) any non-default 1.4 fields on the resource, which can cause some pretty serious data loss and potential production outages in some cases.

rob05c:
>This code reorganization probably makes sense, but I still don't think we should be moving forward with minor versioning taking this approach.

Well, we don't seem to have consensus on changing the versioning for now, and this doesn't change any methodology, it just makes it easier to work with. There isn't really a reason not to do this, to make  we have immediately better.

>For example, a 1.2 client will wipe out (default) any non-default 1.4 fields on the resource, which can cause some pretty serious data loss and potential production outages in some cases.

Agreed, we should fix that. But this doesn't address that, one way or another. It just makes what we have easier to work with.

mitchell852:
i do think making the current way easier is a big plus as i just went thru adding a field to a DS and it was painful. 

the data loss is concerning, however. just so i understand the problem @rawlinp @rob05c 

1.2 client makes an api request to PUT /api/1.2/foos/1 with { a: 1, b: 2, c: 3 }

in 1.3, d was added with a default of 0
in 1.4, e was added with a default of 0

so basically that call to PUT /api/1.2/foos/1 will result in
a=1
b=2
c=3
d=0
e=0

when there might have been some real values in d and e for foo=1...

rob05c:
@mitchell852 Correct. Data overwriting from an old version should be a separate Github issue, if there isn't one.

Incidentally, the tags/reflection solution (https://github.com/rob05c/apiver) implicitly solves this. There is a function which transforms the real struct into a struct with only the given version (https://github.com/rob05c/apiver/blob/master/apiver.go#L99). That function can be used for database as well as JSON operations, automatically giving us the ability to SQL `update` only the fields for that version, and not overwrite any newer fields.

rawlinp:
Adding instructions on how to add new minor versions to APIs and making statements like this just perpetuates the issue which I don't think we should be moving forward with, especially considering how it's currently leading to inadvertent data loss between clients using different minor versions:
> Traffic Control implements [Semantic Versioning](https://semver.org). When adding new fields to the API, we must increase the minor version. If you're the first one adding a new field to a particular object in a particular release, you'll need to do this.

@mitchell852 your example is correct. I just tried this out with a 1.3 deliveryservice field (deepCachingType), and the field was reverted to the default after updating the DS with a 1.2 client. I imagine this is currently affecting all fields from different minor versions, because to fix this issue you'd basically have to "upgrade" a 1.2 request into a 1.4 request by reading all the existing values for 1.4 fields from the DB into the 1.4 struct, _in addition to_ ignoring 1.4 fields passed in a 1.2 request. Right now the code is just sanitizing 1.4 fields without taking into account the existing values in those fields. I haven't opened an issue for this yet because this PR reminded me to try out that scenario yesterday.

This minor versioning code structure is dangerous and is already causing major issues. I think we have mostly been covered so far just because most delivery service changes are made using TP which is supposed to specify the latest minor version, but the whole idea behind minor versioning is that it's supposed to make it safe/stable to request different minor versions. But right now, it's actually more unsafe to request minor versions that aren't the latest version. We can't continue down this current path.

rawlinp:
I opened #3497 per my findings here.

rob05c:
>This minor versioning code structure is dangerous and is already causing major issues.

Removing minor versioning doesn't fix this bug. It just hides it. 

If we remove minor versions, and someone does an UPDATE from an old client, what happens?

The same thing: new fields are overridden with defaults. The only difference is, neither the client nor server operators know what real version the request came from, making it much more difficult to debug the issue.

We can't simply reject requests with null fields, either. Even if null were required to be invalid for all new fields, rejecting nulls would break all old clients, it would be impossible to ever use a client older than the newest server field.

The bug will need to be fixed, irrespective of versioning.

Moreover, removing versioning adds another bug in the other direction: someone with a newer client posting to an older server will lose data from all new fields posted.

>Adding instructions on how to add new minor versions to APIs and making statements like this just perpetuates the issue

This fix isn't a political statement. It just makes life easier for people. Those instructions are correct, and will help people with our current process. If we decide as a project to get rid of minor versions, it's trivial to remove that readme section. In fact, moving the logic to a single file will make it easier to remove minor versioning, if we decide on that as a project.

rawlinp:
> If we remove minor versions, and someone does an UPDATE from an old client, what happens?
The same thing: new fields are overridden with defaults. The only difference is, neither the client nor server operators know what real version the request came from, making it much more difficult to debug the issue.

This is only an issue with minor versioning. If we just support major versions, then all requests are treated as v1. The issue is very easy to debug at that point.

Clients should not be blindly making PUT requests to resources without actually doing a GET for the resource first before making updates. And if an "old" client drops unknown fields between the GET and the PUT, then yes, I would expect it to end up overwriting those unknown fields with the defaults. That is a _client_ bug not necessarily an _API_ bug IMO.

rob05c:
>This is only an issue with minor versioning. If we just support major versions, then all requests are treated as v1. The issue is very easy to debug at that point.

It's an issue without versioning. It's very difficult to debug, because all requests are treated as "v1," we have no idea what actual fields were in the client version they're using. That's one of the primary reasons versioning exists as a concept.

>Clients should not be blindly making PUT requests to resources without actually doing a GET for the resource first

Clients should be allowed to PUT from an object they constructed, or received in the past. The server should allow that, and not lose data.

>if an "old" client drops unknown fields between the GET and the PUT, then yes, I would expect it to end up overwriting those unknown fields with the defaults.

Most clients do that. Only dynamic languages like Perl and Python that deserialize into a hash, rather than a structure, will keep unknown fields.

One of our primary supported clients is Go. We can't just declare "we lose your data if you're using a strongly typed language." Even someone doing a manual HTTP request might lose data, for example if they use `jq` to filter out the fields they know about and do a PUT again.

It's technically possible to retain unknown fields in a typed language like Go, but it would be very verbose and ugly code, and Go in particular would be very slow, and involve polluting the API structs with something like "UnknownFields map[string]string", as well as custom JSON marshallers and unmarshallers.

This isn't a Go problem. Adopting such a policy will be requiring all clients in strongly-typed languages to implement verbose, unwieldy, awkward client code; or risk data loss.

> would expect it to end up overwriting those unknown fields with the defaults. That is a client bug not necessarily an API bug IMO.

Poor usability and data loss is an API bug. Redefining bugs as not our problem doesn't make them any less painful to users. Our goal should be to provide a good API, not an API that declares poor usability and data loss to be the client's fault.

rawlinp:
> Most clients do that. Only dynamic languages like Perl and Python that deserialize into a hash, rather than a structure, will keep unknown fields.

> This isn't a Go problem. Adopting such a policy will be requiring all clients in strongly-typed languages to implement verbose, unwieldy, awkward client code; or risk data loss.

This _is_ a Go problem. It's all about how a client is implemented. Java is a statically-typed language, yet our Java client handles json responses with unknown fields just fine. It's implemented in such a way that new "unknown" fields can be added all the time without the client causing any inadvertent data loss. We should not have to gut our API server just to adapt to the Go client which currently doesn't handle unknown fields well.

mitchell852:
I'm ok merging this as it makes the current way (ability to support minor versions) easier although the current way is still under debate.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3566/
Test PASSed.

mitchell852:
TO unit and api tests pass. Also did some manual creating/updating in TP and all worked well.

lib/go-tc/deliveryservices.go
```diff
@@ -78,7 +78,7 @@ type DeliveryService struct {
 }
 
 type DeliveryServiceV13 struct {
-	DeliveryServiceV12
+	DeliveryServiceV11
 	DeepCachingType   DeepCachingType `json:"deepCachingType"`
 	FQPacingRate      int             `json:"fqPacingRate,omitempty"`
```
mitchell852: do you think in the nullable structs you can get rid of the omitempty's? not sure how these got in here. probably just a misunderstanding of how the api works.

rob05c: Done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3569/
Test PASSed.

PR #3500, opened 2019-04-18
Fix Traffic Ops Delivery Service Update of Old Versions to Omit New Fields

### LINKED ISSUE #3497: TO API clients that don't specify the latest minor version will overwrite/default any fields introduced in later versions
For example:
1. POST a new deliveryservice using the **1.4** API with `deepCachingType = "ALWAYS"`
2. GET the new deliveryservice using the **1.4** API and observe that `deepCachingType == "ALWAYS"`
3. Using the resulting deliveryservice json from the GET request without making any changes, PUT the new deliveryservice using the **1.2** API.
4. GET the updated deliveryservice using the **1.4** API and observe that the `deepCachingType` field has been inadvertently _overwritten_ to `"NEVER"` (the default value).

This example was for `deepCachingType`, but this issue most likely affects all other deliveryservice fields between the different minor versions too.

rob05c:

Fix Traffic Ops Delivery Service Update of Old Versions to Omit New Fields.

Based on #3493, recommend merging that first. (Would have created merge conflicts otherwise.)

- [x] This PR fixes #3497 OR is not related to any Issue

Includes unit tests, verified by existing API Tests.

No documentation necessary, does not change interface.

Includes CHANGELOG.md entry.

- Traffic Ops

PUT a delivery service to a prior version, which has a non-default value in a newer version (for example, `trRequestHeaders` and `/api/1.2`). Verify field is not overwritten.

Run unit tests, run integration tests.

Appears to be in 

- master (21e50eedf)
- 3.0.0
- 3.0.1


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3544/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3545/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3546/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3565/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3570/
Test PASSed.

rawlinp:
Instead of using reflection to determine which struct fields map to database columns and only updating those columns, why not just change the "sanitize" process to just read what's in the database already for those fields instead of setting them to the default?

For example:
```
Foo { // 1.4
  FooV11
  C
}

FooV11 {
  A
  B
}
```

1. the server gets a `PUT /1.1/foo/1` request
2. the server needs to "upgrade" that request into a 1.4 request
3. the server reads the "C" column for id = 1 from the database and populates `Foo.C` with the value.
4. the server then continues handling the `UPDATE` request as if it had always been a 1.4 request.

That would fix the data loss issue without requiring reflection and dynamically generating the set of `UPDATE` fields based upon "db" tags in the struct. Some fields would still need updated separately and don't nicely match up with DB columns in a single table.

rob05c:
> why not just change the "sanitize" process to just read what's in the database already

We could. That would be much, much slower, to incur another DB round-trip to load all the data unnecessarily. A DB roundtrip is at least 50ms, maybe more in a production environment, where the reflection is likely <2ms.

I'm not a fan of reflection, but this was a relatively small function, and we're already doing it in the CRUDer framework; and the `sqlx` calls all have a huge amount of it, it's just hidden from us. And it's by-far the most performant option (unless we did some very ugly things in the code).

rob05c:
It also occurs to me, we can use `dbhelpers.GetUpdateQueryFields` to compute all the update queries for all our objects, to remove hundreds of lines of duplicate update queries. Pretty easy to do the same for create,insert,delete too.

rawlinp:
I don't think we should be worrying about write performance when the system is mostly read-heavy.

The problem with "generic reflection" functions is that it assumes/enforces certain constraints between the struct and the DB schema. I'm trying to imagine how I might use a function like that for `cachegroup.LocalizationMethods`. That's not really a sub-object that gets inserted. It's a list of strings that gets updated in the database by first deleting all the rows in the table with the cachegroup ID at hand, then reinserting rows for each string in the request. I'm not sure how that would work with a function that checks "db" tags for columns to update. It seems like the code would end up with:
- use this special reflection function to update fields that map nicely to the database columns
- use this other specialized logic to update the rest of the fields that don't map nicely to database columns

I also don't really like the whole "well we're already using a little bit of reflection here and there, so we should be good to go crazy with it all over the place" idea. We should only be using it as a last resort IMO.

> It also occurs to me, we can use dbhelpers.GetUpdateQueryFields to compute all the update queries for all our objects, to remove hundreds of lines of duplicate update queries. Pretty easy to do the same for create,insert,delete too.

We shouldn't really have duplicate update queries to begin with IMO. If we can `SELECT` 1.4 values in order to "upgrade" a 1.1 request into a 1.4 request, there would only ever have to be one unique set of `UDPATE` queries per endpoint.

rob05c:
>I also don't really like the whole "well we're already using a little bit of reflection here and there, so we should be good to go crazy with it all over the place" idea.

That is not what I said. You're creating a strawman and exaggerating my position to make it easier to attack. I suggested this small bit of reflection is worth the cost, for the benefit it provides. I do not support "going crazy all over the place" with reflection. 

In this particular case, we're talking about 50 lines of one public function (and one helper), which has the potential to remove huge numbers of duplicate database field strings.

>We shouldn't really have duplicate update queries to begin with IMO. If we can SELECT 1.4 values in order to "upgrade" a 1.1 request into a 1.4 request, there would only ever have to be one unique set of UDPATE queries per endpoint.

Versioning is irrelevant here. The benefit of this function applies, even without versioning.

Currently, every single database field is duplicated in both the struct tag, and the manual update query. Three queries, in fact - create, read, update - for nearly every object we have. For example:
https://github.com/apache/trafficcontrol/blob/master/lib/go-tc/deliveryservices.go#L171
https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go#L1024
https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go#L1099
https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go#L1160

This one-time 50 lines of reflection gives us the ability to remove all those duplicate database field strings, and only store the data in one place. DRY.

>We should only be using it as a last resort

It is a last resort here. Reflection is the only way to get tags in Go. In fact, if you look at the code, it's about as simple as reflection can be. That's all it's doing, iterating over the object and getting the tags. There's nothing magical or complex about the function.

rawlinp:
> That is not what I said. You're creating a strawman and exaggerating my position to make it easier to attack. I suggested this small bit of reflection is worth the cost, for the benefit it provides. I do not support "going crazy all over the place" with reflection.

You're right, I might've been exaggerating, but recently you've gone from vehemently objecting any and all reflection to the point where you're actively reaching for reflection to solve all our problems with it. This problem even has a solution that doesn't require reflection on the go structs, but "one extra sql roundtrip is too slow" for a write operation.

rob05c:
>You're right, I might've been exaggerating, but recently you've gone from vehemently objecting any and all reflection to the point where you're actively reaching for reflection to solve all our problems with it. 

>You're creating a strawman and exaggerating my position to make it easier to attack

We have a lot of problems. I've suggested Reflection for two of them, as a compromise, because you made it clear code generation is unacceptable to you. Code generation and reflection are the only forms of metaprogramming Go has. I agree, metaprogramming should be avoided unless it's necessary, but sometimes there's no way to accomplish something without it.

If you're going to veto every form of metaprogramming, it's going to be much more difficult for us to work with Go. Consider all the places we're already using it in libraries - sql, json, sqlx. When we hit similar situations, primarily places where the type system is limited, it's going to be very difficult, potentially impossible, to write a good, readable, performant application.

I don't really agree with the "write heavy" argument. It's not uncommon for Ops to write a script to change one field on many delivery services. It feels like we've been making many decisions to ignore performance, and I fear the Go is going to end up with many of the same problems as the Perl app, if we continue down this road.

That said, I'm willing to concede here, in the interest of working together. I don't feel like it's that critical of a problem today, to keep fighting over. It does feel like I've been making a lot of concessions lately. I do think it's important for us all to work together on Traffic Control, and the project will be better if we try to see other people's concerns, and are willing to consider them. I'd hope we all value the professional opinions of each other, even when we don't always agree.

I understand why reflection is undesirable in general. But, can you help me understand why this particular function is so bad?

```go
func GetUpdateQueryFields(obj interface{}, omit []string) (string, error) {
	objVal := reflect.ValueOf(obj)
	for objVal.Kind() == reflect.Ptr {
		objVal = reflect.Indirect(objVal)
	}
	if objVal.Kind() != reflect.Struct {
		return "", errors.New("obj must be a struct")
	}
	objType := objVal.Type()
	omitMap := make(map[string]struct{}, len(omit))
	for _, omitVal := range omit {
		omitMap[omitVal] = struct{}{}
	}
	qryFieldStrs, err := getUpdateQueryFieldsType(objType, omitMap)
	return strings.Join(qryFieldStrs, ",\n"), err
}
func getUpdateQueryFieldsType(objType reflect.Type, omit map[string]struct{}) ([]string, error) {
	qryFieldStrs := []string{}
	for i := 0; i < objType.NumField(); i++ {
		objField := objType.Field(i)
		if objField.Type.Kind() == reflect.Struct && objField.Anonymous {
			embeddedQryFieldStrs, err := getUpdateQueryFieldsType(objField.Type, omit)
			if err != nil {
				return nil, errors.New("subfield '" + objField.Name + "': " + err.Error())
			}
			qryFieldStrs = append(qryFieldStrs, embeddedQryFieldStrs...)
			continue
		}
		objDBTag := objField.Tag.Get("db")
		if _, ok := omit[objDBTag]; objDBTag == "" || ok {
			continue
		}
		qryFieldStrs = append(qryFieldStrs, objDBTag + "=:" + objDBTag)
	}
	return qryFieldStrs, nil
}
```

This helper function would also let us fix the duplicate database column names. Do you object to this function being added for that purpose? Do you feel that this function is worse than having those column names duplicated everywhere? If so, can you help me understand why this function is so bad?

rawlinp:
I don't really have a problem with the function itself; it probably does exactly what it's meant to do. My concern is about how it will be used and its long-term effects on the codebase. It would really only apply to struct fields that map nicely to the database columns, and for fields that don't map nicely to a database column in the same table as the rest of the `UDPATE` fields, this function could not be used. I think it would lead to people building structs that always map nicely to the database, which is not always a good thing. I think it would also lead to inconsistency in API implementations because there would be two ways to update tables: one way for fields that map nicely to the DB, and another way for fields that don't. We also wouldn't be able to read the code and see the exact SQL query that is meant to be run (at least in terms of which columns are involved), which would make it a little harder to troubleshoot. It would also add to the mental overhead of defining structs by having to make sure that you have the proper tags defined on only specific fields.

IMO I think those could be worse problems than having the database column names duplicated across the SELECT, INSERT, and UPDATE queries for the same object.

PR #3501, opened 2019-04-18
[Ready for Review] Use of AngularJS markup in URL-valued attribute

MatheusRV:
Using AngularJS markup in an HTML attribute that references a URL (such as 'href' or 'src') may cause the browser to send a request with an invalid URL.

### DIFF
--- a/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
+++ b/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
@@ -88,7 +88,7 @@ under the License.
                         DNS is the standard routing type for most CDN services. HTTP Redirect is a specialty routing service that is primarily used for video and large file downloads where localization and latency are significant concerns. A "Live" routing type should be used for all live services.
                         <br>
                         <br>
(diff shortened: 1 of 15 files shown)


asfgit:

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -356,7 +356,7 @@ <h6>Note:</h6>
                             Enables/disables the Multi-Site Origin feature for this Delivery Service.
                             <br>
                             <br>
-                            <a href="https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_ops/using.html#multi-site-origin" target="_blank">See Multi-Site Origin.</a>
+                            <a ng-href="https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_ops/using.html#multi-site-origin" target="_blank">See Multi-Site Origin.</a>
```
ocket8888: This - and several others at first glance - does not appear to be using any AngularJS string interpolation (which is I think what you mean by "markup"?) so this is an unnecessary change.

PR #3502, opened 2019-04-18
Duplicate parameter names

MatheusRV:
If a function has two parameters with the same name, the second parameter shadows the first one, which makes the code hard to understand and error-prone.

### DIFF
--- a/traffic_portal/app/src/common/api/AuthService.js
+++ b/traffic_portal/app/src/common/api/AuthService.js
@@ -76,5 +76,5 @@ var AuthService = function($rootScope, $http, $state, $location, $q, httpService
 
 };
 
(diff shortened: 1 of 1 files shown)


asfgit:

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-var AuthService = function($rootScope, $http, $state, $location, $q, $state, httpService, userModel, messageModel, ENV) {
+var AuthService = function($rootScope, $http, $state, $location, $q, httpService, userModel, messageModel, ENV) {
```
mitchell852: good catch, however, this change probably won't compile or will cause strange errors because you didn't update line 79 as well.

MatheusRV:
Done.

dangogh:
> Please fill out the Pull Request template to the best of your ability.

+1

ocket8888:
This was resolved in #3601

PR #3503, opened 2019-04-18
Comparison using is when operands support __eq__

MatheusRV:
Comparison using 'is' when equivalence is not the same as identity

### DIFF
--- a/misc/kickstart_create_network_line.py
+++ b/misc/kickstart_create_network_line.py
@@ -156,7 +156,7 @@ def useable_interfaces(net_devs, nc, iface_speed):
                 else:
                     iface_list = [nc['BOND_DEVICE']]
                     notes = "{0} did not have carrier at install time, and may not work".format(nc['BOND_DEVICE'])
-    elif iface_speed is not 'auto':
+    elif iface_speed != 'auto':
         if len(net_devs[iface_speed]) > 0:
             iface_list = net_devs[iface_speed]
         else:


asfgit:

ocket8888:
That is a good change, but an interesting tidbit: in CPython this actually works properly because the interpreter will cache numeric constants and short string constants at compile time. So assigning a variable to the value of a small string actually just points it at the existing object. Since strings and numbers are immutable, this is totally safe. But it's wrong to rely on implementation quirks.

MatheusRV:
as nornagon says in https://github.com/electron/electron/pull/17864#issuecomment-484582353 "Python deduplicates small integers (<= 256) so is works on them, but larger integers get their own instance and don't compare equal with is any more."

ocket8888:
I didn't realize there was a size limit on integers. Wonder how it handles negative integers?

rawlinp:
I also agree with this change, and IMO this change is so simple and obvious that I'd let it bypass filling out the PR template. Next time, please fill it out though, it helps reviewers a lot.

ocket8888:
I disagree. One of the big things pushed for in the new template was "what versions of TC are affected by this bug?". This bug almost certainly exists in 3.0.x, and it'd be nice to have a record of that.

Most sections can easily be filled in with N/A or a short description for such a small change. Deleting the PR template is a bad habit that I'm loathe to encourage.

MatheusRV:
I Will edit adding PR template 🐶

ocket8888:
Obsolete as of #3691

PR #3505, opened 2019-04-18
Oauth integration

mattjackson220:
…rtal Login page updates, Traffic Ops API updates, and documentation updates

rebase got messed up. please reference new PR here:
https://github.com/apache/trafficcontrol/pull/3763

Sets up integration with OAuth providers to allow Single Sign On instead of username/password login to Traffic Portal.  It defaults to disabled.  Once OAuth is set up, login will follow these steps:
- User will go to TP login page as usual, for example at tp.domain.com.  If OAuth is enabled, a single button will be visible to login with SSO.
- When button is clicked, it will redirect to the OAuth provider URL from the traffic_portal_properties.json.
- OAuth provider will authenticate the user using SSO.  If user is not logged in, they will be redirected to SSO login page.  If user is logged in, they will be redirected to tp.domain.com/#!/sso?auth_token=encryptedTokenFromOAuthProvider
- the /sso page will parse the token and POST to the API /user/login/oauth endpoint
- the API /user/login/oauth endpoint will decode and validate the token, cross reference the Json Key Set URL against the whitelisted URLs in the cdn.conf file, query the database to get the user's role, and if all of that is successful, return a cookie.
- If the login is successful, the user will be redirected to the page they were trying to see.  If login was unsuccessful, a 401 error will be returned and the user will be redirected back to the login page.

- Documentation
- Traffic Ops
- Traffic Portal

To test using CDN In A Box:
- Update config.sh to : 
        "whitelisted_oauth_urls": [
    		"insert domain for your expected Json Key Set returned by oAuth provider"
	]
- Update traffic_portal_properties.json to: 
         "oAuth": {
  		"_comment": "Opt-in OAuth properties for SSO login",
 		 "enabled": true,
 		 "oAuthUrl": "insert your oAuth provider URL",
  		"oAuthTokenQueryParam": "insert your oAuth provider's token query parameter"
	}
- Run CDN in a box
- Verify new login page only shows button to login with SSO
- Click button with Network panel open and verify that it makes call to OAuth provider
- Verify that login failed
- Login to db container: 
        docker exec -it cdn-in-a-box_db_1 /bin/bash
- Login to postgres 
        psql -d traffic_ops -U traffic_ops
- Insert your user into db
         insert into tm_user (username, role, tenant_id) values ('yourUserId', (SELECT id FROM role WHERE name = 'admin'), (SELECT id FROM tenant WHERE name='root'));
- Click Login button again, Verify call to OAuth provider, Verify that login succeeds
- Log out

- Try to go directly to /#!/servers endpoint
- Verify it requires login and shows SSO login page
- Click Login
- Verify call to OAuth provider, verify login successful, verify it redirects to /#!/servers endpoint

- Update config.sh to : 
        "whitelisted_oauth_urls": []
- Restart Traffic Ops
- Try to log in and verify it returns the following error:
        "Key URL from token is not included in the whitelisted urls. Received: //url"


asfgit:

ocket8888:
I'll check this out in a minute here

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -164,6 +164,7 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `users/{id}/deliveryservices/?(\.json)?$`, user.GetDSes, auth.PrivLevelReadOnly, Authenticated, nil},
 		{1.1, http.MethodGet, `user/{id}/deliveryservices/available/?(\.json)?$`, user.GetAvailableDSes, auth.PrivLevelReadOnly, Authenticated, nil},
 		{1.1, http.MethodPost, `user/login/?$`, login.LoginHandler(d.DB, d.Config), 0, NoAuth, nil},
+		{1.4, http.MethodPost, `user/login/oauth$`, login.OauthLoginHandler(d.DB, d.Config), 0, NoAuth, nil},
```
ocket8888: most other endpoints support an optional trailing slash. It's probably not a huge deal, but it's also only a 2-character change to support it.

traffic_portal/app/src/app.js
```diff
@@ -474,7 +475,7 @@ trafficPortal.factory('authInterceptor', function ($rootScope, $q, $window, $loc
             if (rejection.status === 401) {
                 $rootScope.$broadcast('trafficPortal::exit');
                 userModel.resetUser();
-                if (url == '/login' || $location.search().redirect) {
+                if (url == '/login' || url =='/sso' || $location.search().redirect) {
```
ocket8888: nitpick: I know we aren't doing this already (as evidenced by `url == '/login'`), but, in general, it's better to compare JS values with "triple-equals" (`===`/`!==`) .

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -57,6 +57,25 @@ var AuthService = function($rootScope, $http, $state, $location, $q, $state, htt
         return deferred.promise;
     };
 
+    this.oauthLogin = function(token) {
+        return httpService.post(ENV.api['root'] + 'user/login/oauth', { t: token})
+            .then(
+                function(result) {
+                    $rootScope.$broadcast('authService::oauthLogin');
+                    var redirect = decodeURIComponent($location.search().redirect);
+                    if (redirect !== 'undefined') {
```
ocket8888: Is this meant to check if `redirect` is `undefined`? because it's checking for equality against the *string* `'undefined'`.

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -57,6 +57,25 @@ var AuthService = function($rootScope, $http, $state, $location, $q, $state, htt
         return deferred.promise;
     };
 
+    this.oauthLogin = function(token) {
+        return httpService.post(ENV.api['root'] + 'user/login/oauth', { t: token})
+            .then(
+                function(result) {
+                    $rootScope.$broadcast('authService::oauthLogin');
+                    var redirect = decodeURIComponent($location.search().redirect);
```
ocket8888: I'm finding myself saying this a lot: I know that we use this throughout the TP codebase, but for ease of debugging would you mind using `let`/`const` (`const` is appropriate in this case, I think) instead of `var`?

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -57,6 +57,25 @@ var AuthService = function($rootScope, $http, $state, $location, $q, $state, htt
...
+                    if (redirect !== 'undefined') {
+                        $location.search('redirect', null); // remove the redirect query param
+                        $location.url(redirect);
+                    } else {
+                        $location.url('/');
+                    }
+                },
+                function(fault) {
+                    // do nothing
```
ocket8888: I know that this is how things are done throughout the code-base, but honestly we shouldn't be silently ignoring errors. Can you `console.error` out the `fault`?

traffic_portal/app/src/common/service/utils/UrlUtils.js
```diff
@@ -0,0 +1,34 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+var UrlUtils = function() {
+    this.getUrlQueryParams = function(url) {
+        var vars = {};
+        var hashes = url.split('?')[1];
+        var hash = hashes != null ? hashes.split('&') : '';
```
ocket8888: same as above RE: `let`/`const`

also, though, you could instead use [the `URL` object](https://developer.mozilla.org/en-US/docs/Web/API/URL), which has a `searchParams` property that implements [the `URLSearchParams` interface](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) and provides exactly the parsing you're implementing.

traffic_portal/app/src/common/service/utils/UrlUtils.js
```diff
@@ -0,0 +1,34 @@
...
+
+        for (var i = 0; i < hash.length; i++) {
+            var params=hash[i].split('=');
+            vars[params[0]] = params[1];
+        }
+        return vars;
+    };
+};
+UrlUtils.$inject = [];
+module.exports = UrlUtils;
```
ocket8888: newlines at the end of files for POSIX-compliance, please

traffic_portal/app/src/modules/public/login/LoginController.js
```diff
@@ -48,9 +53,31 @@ var LoginController = function($scope, $log, $uibModal, authService, userService
...
+        var returnParams = '?redirect=' + queryParams['redirect'];
+        // keeps query parameters that were added to the goToUrl
+        for (var i in goToQueryParams) {
+            returnParams = returnParams + '&' + i + '=' + goToQueryParams[i];
+        }
+
+        if (returnParams) {
+            continueURL = '?continue=' + $scope.returnProtocol + '//' + $scope.returnUrl + ($scope.returnPort ? ':' + $scope.returnPort : '') + '/' + encodeURIComponent('#!') + '/sso' + returnParams;
+        } else {
+            continueURL = '?continue=' + $scope.returnProtocol + '//' + $scope.returnUrl + ($scope.returnPort ? ':' + $scope.returnPort : '') + '/' + encodeURIComponent('#!') + '/sso';
```
ocket8888: instead of separating the parts of `window.location`, why not store the full `Location` object so you could just manipulate the `hash` property thereof? That's essentially what you're doing, but currently storing the information in three different places and doing inline collapsing and ternaries and whatnot (that will also handle URI encoding for you).

traffic_portal/app/src/modules/public/login/LoginController.js
```diff
@@ -48,9 +53,31 @@ var LoginController = function($scope, $log, $uibModal, authService, userService
         });
     };
 
+    $scope.loginOauth = function() {
+        var continueURL = '';
+        const gotoUrl = propertiesModel.properties.oAuth.oAuthUrl;
+        const gotoUrlWithoutParams = gotoUrl.split('?')[0];
```
ocket8888: though it's odd, a `?` could conceivably appear in the path of a URL and not necessarily signify the beginning of query params - if you use a `URL` object you can get what you're looking for with `urlObject.origin + urlObject.pathname` (optionally support document fragments by appending `# + urlObject.hash` if desired/if it exists).

traffic_portal/app/src/modules/public/login/login.tpl.html
```diff
@@ -21,12 +21,13 @@
     <div class="col-md-12 col-sm-12 col-xs-12">
         <div class="x_panel">
             <div class="x_title">
-                <h2>Login</h2>
+                <h2 ng-show="!oAuthEnabled">Login</h2>
+                <h2 ng-show="oAuthEnabled">Login With Single Sign On</h2>
```
ocket8888: instead of `ng-show`, consider `ng-if` since this should never change for a single pageload.

traffic_portal/app/src/modules/public/login/login.tpl.html
```diff
@@ -21,12 +21,13 @@
...
                 <div class="clearfix"></div>
             </div>
             <div class="x_content">
                 <br>
-                <form name="loginForm" role="form" class="form-horizontal form-label-left" novalidate>
+                <form name="loginForm" role="form" class="form-horizontal form-label-left" novalidate ng-show="!oAuthEnabled">
```
ocket8888: Same as above RE: `ng-if` vs `ng-show`.

Also, why are we doing `novalidate` on these forms? Your change is consistent with what we have already, but we aren't doing any angular validations on these forms, so `novalidate` only disables validations that aren't being overridden anyway.

traffic_portal/app/src/modules/public/login/login.tpl.html
```diff
@@ -47,6 +48,13 @@ <h2>Login</h2>
                         </div>
                     </div>
                 </form>
+                <form name="loginForm" role="form" class="form-horizontal form-label-left" novalidate ng-show="oAuthEnabled">
```
ocket8888: Same as above RE: `ng-if` vs `ng-show`

traffic_portal/app/src/modules/public/sso/SsoController.js
```diff
@@ -0,0 +1,37 @@
...
+var SsoController = function($scope, $state, $interval, urlUtils, authService, propertiesModel) {
+
+	var init = function () {
+        const queryParams = urlUtils.getUrlQueryParams(window.location.href);
+        const tokenParameter = propertiesModel.properties.oAuth.oAuthTokenQueryParam;
+        const token = queryParams[tokenParameter];
+
+        authService.oauthLogin(token)
+            .then(
+                function() {}
```
ocket8888: If you're not doing anything, do you need a `.then`? Maybe just log a success message or something?

traffic_portal/app/src/modules/public/sso/sso.tpl.html
```diff
@@ -0,0 +1,20 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<div id="ssoContainer"/>
```
ocket8888: Same as above RE: terminating newlines

also, as far as I can tell, this is never styled or accessed anywhere, so could it just have a blank template instead of this file? Or maybe I'm wrong and Angular will complain about blank templates. In that case, I still don't think this `id` is ever used for anything.

docs/source/api/user_login_oauth.rst
```diff
@@ -0,0 +1,71 @@
...
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+.. _to-api-user-login-oauth:
+
+********************
+``user/login/oauth``
+********************
+
```
ocket8888: You should use a `.. versionadded::` directive to clarify in what API version(s) this endpoint is available.

traffic_portal/app/src/modules/public/sso/sso.tpl.html
ocket8888: Also, according to [the HTML spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element) the DIV element may not omit a closing tag.

mitchell852:
@mattjackson220 - since you created a new api endpoint. can you create a new "capability" called sso_auth or oath or whatever you think is best and and add it to every role in seeds.sql and patches.sql? let me know if you have questions about that.

docs/source/admin/quick_howto/index.rst
```diff
@@ -27,6 +27,7 @@ Traffic Control is a complicated system, and documenting it is not trivial. Some
 	ds_requests
 	federations
 	multi_site
+	oauth-login
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,77 @@
...
+.. limitations under the License.
+..
+.. _oauth-login:
+
+*********************
+Configure OAuth Login
+*********************
+
+An opt-in configuration for SSO using OAuth is supported and can be configured through the :file:`/opt/traffic_portal/public/traffic_portal_properties.json` and :file:`/opt/traffic_ops/app/conf/cdn.conf` files.
+OAuth uses a third party provider to authenticate the user. Once enabled, the Traffic Portal Login page will no longer accept username and password but instead will authenticate using OAuth.
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,77 @@
...
+This will redirect to the ``oAuthUrl`` from :file:`/opt/traffic_portal/public/traffic_portal_properties.json` which will authenticate the user then redirect to the new ``/sso`` page with a Json Web Token added as a query parameter.
+The new ``/sso`` page will parse the token from the URL and post this information to the ``/api/1.4/user/login/oauth`` API endpoint. See :ref:`to-api-user-login-oauth`.
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,77 @@
...
+The ``/api/1.4/user/login/oauth`` API endpoint will decode the token, validate that it is between the issued time and the expiration time, and validate that the public key set URL is allowed by the list of whitelisted URLs read from :file:`/opt/traffic_ops/app/conf/cdn.conf`. It will then authorize the user from the database and return a mojolicious cookie as per the normal login workflow.
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,77 @@
...
+
+.. Note:: OAuth providers sometimes do not return the public key set URL but instead require a locally stored key. This functionality is not currently supported and will require further development.
+
+To configure OAuth login:
+
+Set up authentication with a third party OAuth provider.
+
+- Update :file:`/opt/traffic_portal/public/traffic_portal_properties.json` and ensure the following properties are set up correctly:
+
+.. table:: OAuth Configuration Property Definitions In traffic_portal_properties.json
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,77 @@
...
+.. code-block:: text
+    :caption: Example OAuth Configuration Properties In traffic_portal_properties.json
+
+    "oAuth": {
+        "enabled": true,
+        "oAuthUrl": "example.oauth.com",
+        "oAuthTokenQueryParam": "access_token",
+    }
+
+- Update :file:`/opt/traffic_ops/app/conf/cdn.conf` property traffic_ops_golang.whitelisted_oauth_urls to contain all allowed domains for the Json key set (Use `*` for wildcard):
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,77 @@
...
+
+.. table:: OAuth Configuration Property Definitions In cdn.conf
+
+	+--------------------------+------------+---------------------------------------------------------------------------------------------------------------+
+	| Name                     | Type       | Description                                                                                                   |
+	+==========================+============+===============================================================================================================+
+	| whitelisted_oauth_urls   | []string   | List of whitelisted URLs for the Json public key set returned by OAuth provider.  Can contain * wildcards.    |
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,77 @@
...
+
+    "oAuth": {
+        "enabled": true,
+        "oAuthUrl": "example.oauth.com",
+        "oAuthTokenQueryParam": "access_token",
+    }
+
+- Update :file:`/opt/traffic_ops/app/conf/cdn.conf` property traffic_ops_golang.whitelisted_oauth_urls to contain all allowed domains for the Json key set (Use `*` for wildcard):
+
+.. table:: OAuth Configuration Property Definitions In cdn.conf
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,77 @@
...
+
+	+--------------------------+------------+---------------------------------------------------------------------------------------------------------------+
+	| Name                     | Type       | Description                                                                                                   |
+	+==========================+============+===============================================================================================================+
+	| whitelisted_oauth_urls   | []string   | List of whitelisted URLs for the Json public key set returned by OAuth provider.  Can contain * wildcards.    |
+	+--------------------------+------------+---------------------------------------------------------------------------------------------------------------+
+
+
+.. code-block:: text
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +108,119 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
 		fmt.Fprintf(w, "%s", respBts)
 	}
 }
+
+func OauthLoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -21,8 +21,12 @@ package login
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"github.com/dgrijalva/jwt-go"
+	"github.com/lestrrat/go-jwx/jwk"
```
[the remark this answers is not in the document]
mattjackson220: These libraries have more easily used parsing and validating functions for jwts and for fetching the keys from the json key set URL. do they not seem good to use?

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +108,119 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+				return nil, errors.New("Key URL from token is not included in the whitelisted urls. Received: " + publicKeyUrl)
+			}
+
+			keys, err := jwk.FetchHTTP(publicKeyUrl)
+			if err != nil {
+				return nil, err
+			}
+
+			keyById := keys.LookupKeyID(publicKeyId)
+			selectedKey, err := keyById[0].Materialize()
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
ocket8888: The libraries seem fine, it's just that I don't like pulling in more external dependencies if the standard library would work just as well. I'm especially hesitant about `jwk`, because that doesn't really appear to have any docs.

But after digging through the standard library's Oath2 package ( and its sub-packages ) I wasn't really able to find a way to simply parse a JWT - which is sort of exactly what we want to do here. 

Maybe I'm just an idiot and it can be done, but since I couldn't figure out how I'm fine with just using these I guess. It'd be nice if we could just pick one or the other though (preferably `jwt-go`).

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +108,119 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+
+			if err != nil {
+				return nil, err
+			}
+
+			return selectedKey, nil
+		})
+		if err != nil {
+			handleErrs(http.StatusInternalServerError, errors.New("Error decoding token with message: "+err.Error()))
```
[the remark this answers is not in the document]
mattjackson220: I was torn on this.  Since the KeyFunc is passed into the Parse method, this line will catch all of the errors above, including errors from Materialize, bad key set URLs, bad public key, etc.  So I'm open to suggestions if InternalServerError is not appropriate, but this error is a catch-all for the other errors

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +108,119 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+		form.Username = userId
+
+		userAllowed, err, blockingErr := auth.CheckLocalUserIsAllowed(form, db, time.Duration(cfg.DBQueryTimeoutSeconds)*time.Second)
+		if blockingErr != nil {
+			api.HandleErr(w, r, nil, http.StatusServiceUnavailable, nil, fmt.Errorf("error checking local user password: %s\n", blockingErr.Error()))
+			return
+		}
+		if err != nil {
+			log.Errorf("checking local user: %s\n", err.Error())
+			return
```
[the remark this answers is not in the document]
mattjackson220: For this one, I followed what is done in the original LoginHandler.  If this error is hit, then it logs the error and returns an "Invalid username or password." message because it failed to authorize and authenticate the user.  i think a call to api.HandleErr would be confusing because it wouldnt display the message that the user was invalid

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +108,119 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+		}
+
+		authenticated = decodedToken.Valid
+
+		userId := decodedToken.Claims.(jwt.MapClaims)["sub"].(string)
+		form.Username = userId
+
+		userAllowed, err, blockingErr := auth.CheckLocalUserIsAllowed(form, db, time.Duration(cfg.DBQueryTimeoutSeconds)*time.Second)
+		if blockingErr != nil {
+			api.HandleErr(w, r, nil, http.StatusServiceUnavailable, nil, fmt.Errorf("error checking local user password: %s\n", blockingErr.Error()))
```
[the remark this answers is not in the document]
mattjackson220: I followed the LoginHandler for which to use in each case.  I believe the api.HandleErrs should be used for system errors and just the handleErrs is used for errors with the call to the API or within the API endpoint processing.

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +108,119 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+		respBts, err := json.Marshal(resp)
+		if err != nil {
+			handleErrs(http.StatusInternalServerError, err)
+			return
+		}
+		w.Header().Set(tc.ContentType, tc.ApplicationJson)
+		if !authenticated {
+			w.WriteHeader(http.StatusUnauthorized)
+		}
+		fmt.Fprintf(w, "%s", respBts)
```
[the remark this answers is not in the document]
mattjackson220: Also followed the LoginHandler for consistency.  I'm not sure the reasoning though

traffic_ops/traffic_ops_golang/login/login_test.go
```diff
@@ -0,0 +1,34 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+func TestVerifyUrlOnWhiteList(t *testing.T) {
+	url := "test.right"
+	whitelistedUrls := [][]string{[]string{}, []string{""}, []string{"*"}, []string{"test.wrong"}, []string{"test.right"}, []string{"*.right"}, []string{"test.wrong", "test.right"}}
+	expected := []bool{false, false, true, false, true, true, true}
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_ops/traffic_ops_golang/login/login_test.go
```diff
@@ -0,0 +1,34 @@
+package login
+
+import "testing"
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_ops/traffic_ops_golang/routing/routes.go
mattjackson220: done

traffic_portal/app/src/app.js
mattjackson220: done

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
ocket8888: Alright :+1: sounds good

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
ocket8888: Well I still think it's weird, but if it's established practice there's no reason to worry any more about it on this PR.

traffic_portal/app/src/common/api/AuthService.js
mattjackson220: done. and ill keep my eye out for more things like this as i continue to work in TP

traffic_portal/app/src/modules/public/login/login.tpl.html
mattjackson220: done

traffic_portal/app/src/modules/public/login/login.tpl.html
mattjackson220: done

traffic_portal/app/src/modules/public/login/login.tpl.html
mattjackson220: done for the ng-if
as for the novalidate, is this worth updating now or should i stay with going for consistency with the other code?

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
ocket8888: It doesn't really matter I suppose. It's just weird. Especially because it's using `fmt.Fprintf` when the format specifier is just `%s` - so why not use `fmt.Print`?

traffic_portal/app/src/modules/public/sso/SsoController.js
mattjackson220: done

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
rawlinp: Yeah, historically, new project dependencies are something that should really be discussed on the mailing list, primarily because we can get into hot water if we bring in a project dependency that is against the ASF licensing policy. These are both MIT-licensed, which means we should be able to bring them in, but discussing it on the mailing list also lets people evaluate the proposed libraries and give their input on whether or not we should really include them (or if they know of better alternatives -- stuff like that).

docs/source/api/user_login_oauth.rst
mattjackson220: done

traffic_portal/app/src/modules/public/login/login.tpl.html
ocket8888: unless you wanna change both - which you absolutely do not have to do -, don't bother

traffic_portal/app/src/modules/public/sso/sso.tpl.html
mattjackson220: done.  you were right, i didnt need it at all and could just use a blank template.

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
mattjackson220: good to know! I'll send out an email to the mailing list.
And I couldnt figure it out either.  I'll dig more into jwt-go and see if i can get rid of the jwk dependency

traffic_portal/app/src/common/api/AuthService.js
mattjackson220: done. looks like it was just a mistake that didnt matter before because if there isnt a redirect parameter then it just goes to the main page anyway.  probably dont even need this if statement at all actually, but ill update to check for undefined instead of the string

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
ocket8888: Yeah, looks like I misunderstood what was happening here. If you want, you could call `handleErrs` _within_ the anonymous `func` and set a boolean variable local to the caller to indicate if a returned `error` was internal to the anonymous `func`. Something like:

```go
internal := false
decodedToken, err := jwt.Parse(encodedToken, func(unverifiedToken *jwt.Token) (interface{}, error) {
    thing, err := funcThatCouldErr()
    if err != nil {
        internal = true
        handleErrs(http.ApropriateStatus, errors.New("Error description"))
        return nil, err
    }
    // ...
}
if internal {
    return
} else if err != nil {
    handleErrs(http.StatusInternalServerError, errors.New("Error parsing JWT: "+err.Error())
    return
}
```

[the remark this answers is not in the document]
ocket8888: But idk how big a deal that is - I don't know of any ATC client sophisticated enough to care about the appropriateness of the status code. Or any human (besides me) that cares. If you want to leave it the way you have it, that's fine with me.

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
rob05c: Yeah, `handleErrs`/`tc.GetHandleErrorsFunc` is legacy, it’s awkward to pre-declare and it isn’t as obvious what’s happening. We’re moving to `api.HandleErr(w, r, tx.Tx, httpErrCode, userErr, sysErr)` `https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/api/api.go#L103`

I know you're following existing code, and it's not a big deal. Just for future use, and if you get a chance.

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
rob05c: This has to return an error to the client. Right now, by returning, this is just dropping the connection to the client, with no return code or body. If it was doing that before, that was a bug.

That's what `handleErrs` and `api.HandleErr` do. You could manually `w.WriteHeader(http.StatusInternalServerError)` here, but it needs to return _something_ to the client.

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +108,119 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+				tc.Alerts
+			}{tc.CreateAlerts(tc.ErrorLevel, "Invalid username or password.")}
+		}
+
+		respBts, err := json.Marshal(resp)
+		if err != nil {
+			handleErrs(http.StatusInternalServerError, err)
+			return
+		}
+		w.Header().Set(tc.ContentType, tc.ApplicationJson)
```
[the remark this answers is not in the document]
rob05c: Symbols instead of duplicate text prevent typos. `tc.ContemtType` won't compile, but `"Contemt-Type"` will. Of course, enum.go `const ContentType = "Content-Type"` could have a typo, but by making it a symbol, we limit the potential bugs to 1 place instead of the 500 places we need the string "Content-Type"

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +108,119 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+func VerifyUrlOnWhiteList(url string, whiteListedUrls []string) bool {
+
+	for _, listing := range whiteListedUrls {
+		if listing == "" {
+			continue
+		}
+		if listing == "*" || strings.Contains(url, listing) {
+			return true
+		} else if strings.Contains(listing, "*") {
+			abbrvListing := strings.SplitAfter(listing, "*.")[1]
```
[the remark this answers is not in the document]
mattjackson220: looks like that works great! thank you! 
done

traffic_portal/app/src/common/api/AuthService.js
mattjackson220: done

traffic_portal/app/src/common/service/utils/UrlUtils.js
mattjackson220: deleted this utils file and moved it all to just use the URL object
done

traffic_portal/app/src/common/service/utils/UrlUtils.js
mattjackson220: deleted this utils file and moved it all to just use the URL object
done

traffic_portal/app/src/modules/public/login/LoginController.js
mattjackson220: done

traffic_portal/app/src/modules/public/login/LoginController.js
mattjackson220: done

traffic_portal/app/src/traffic_portal_properties.json
```diff
@@ -407,6 +407,12 @@
           "url": "http://trafficcontrol.apache.org/"
         }
       ]
+    },
+    "oAuth": {
+      "_comment": "Opt-in OAuth properties for SSO login",
+      "enabled": false,
+      "oAuthUrl": "",
```
mitchell852: can you put some kind of fake example value in oAuthUrl and oAuthTokenQueryParam just to give us an idea of what to put in here. also, can you maybe explain a bit more in the _comment field?

traffic_portal/app/src/modules/public/login/LoginController.js
```diff
@@ -17,7 +17,12 @@
  * under the License.
  */
 
-var LoginController = function($scope, $log, $uibModal, authService, userService) {
+var LoginController = function($scope, $log, $uibModal, authService, userService, propertiesModel) {
+
+    $scope.returnUrl = window.location.hostname;
```
mitchell852: I don't see where returnUrl, returnPort or returnProtocol are being used

traffic_portal/app/src/modules/public/login/LoginController.js
```diff
@@ -48,9 +53,28 @@ var LoginController = function($scope, $log, $uibModal, authService, userService
...
+        const gotoUrl = new URL(propertiesModel.properties.oAuth.oAuthUrl);
+        const currentUrl = new URL(window.location.href);
+        const afterHash = currentUrl.hash.substring(2);
+        const urlNoHash = new URL(currentUrl.protocol + '//' + currentUrl.hostname + currentUrl.pathname + afterHash);
+
+        const redirectParam = urlNoHash.searchParams.get('redirect') !== null ? urlNoHash.searchParams.get('redirect') : '';
+        var continueParam = new URL(window.location.href);
+        continueParam.hash = '#!/sso?redirect=' + redirectParam;
+
+        var continueURL = new URL(gotoUrl.protocol + '//' + gotoUrl.hostname + gotoUrl.pathname);
```
mitchell852: this is all pretty confusing. can you add some comments to explain the format of the url you are building to send them to?

traffic_portal/app/src/modules/public/login/login.tpl.html
```diff
@@ -47,6 +48,13 @@ <h2>Login</h2>
                         </div>
                     </div>
                 </form>
+                <form name="loginForm" role="form" class="form-horizontal form-label-left" novalidate ng-if="oAuthEnabled">
```
mitchell852: can you rename this to name="ssoLoginForm" or something so there's not 2 forms of the same name?

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -57,6 +57,25 @@ var AuthService = function($rootScope, $http, $state, $location, $q, $state, htt
         return deferred.promise;
     };
 
+    this.oauthLogin = function(token) {
+        return httpService.post(ENV.api['root'] + 'user/login/oauth', { t: token})
+            .then(
+                function(result) {
+                    $rootScope.$broadcast('authService::oauthLogin');
```
mitchell852: you can change this to 

`$rootScope.$broadcast('authService::login');`

if you do a search for authService::login you'll see what dispatching that event does.

traffic_portal/app/src/modules/public/login/login.tpl.html
```diff
@@ -47,6 +48,13 @@ <h2>Login</h2>
...
+                    <div class="form-group">
+                        <div class="col-md-6 col-sm-6 col-xs-12 col-md-offset-1">
+                            <button name="loginOauthSubmit" type="submit" class="btn btn-primary" ng-click="loginOauth()">Log in &nbsp;&nbsp;<i class="fa fa-chevron-circle-right"></i></button>
```
mitchell852: i think this should be changed to type=button

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -57,6 +57,25 @@ var AuthService = function($rootScope, $http, $state, $location, $q, $state, htt
...
+                    const redirect = decodeURIComponent($location.search().redirect);
+                    if (redirect !== undefined) {
+                        $location.search('redirect', null); // remove the redirect query param
+                        $location.url(redirect);
+                    } else {
+                        $location.url('/');
+                    }
+                },
+                function(fault) {
```
mitchell852: any failure will be a 401, right? i.e. if it turns out that the token is bogus or username doesn't exist in the database is a 401 returned? if so, you can kill this 2nd function as 401s are captured globally

traffic_portal/app/src/modules/public/login/login.tpl.html
ocket8888: Either that or `ng-click` on the button could be `ng-submit` on the form

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,72 @@
...
+
+.. Note:: OAuth providers sometimes do not return the public key set URL but instead require a locally stored key. This functionality is not currently supported and will require further development.
+
+To configure OAuth login:
+
+- Set up authentication with a third party OAuth provider.
+
+- Update :file:`/opt/traffic_portal/public/traffic_portal_properties.json` and ensure the following properties are set up correctly:
+
+        .. table:: OAuth Configuration Property Definitions In traffic_portal_properties.json
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/api/user_login_oauth.rst
```diff
@@ -0,0 +1,73 @@
...
+``POST``
+========
+.. versionadded:: 1.4
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_portal/app/src/traffic_portal_properties.json
mattjackson220: done

traffic_portal/app/src/modules/public/sso/SsoController.js
```diff
@@ -0,0 +1,37 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+var SsoController = function($scope, $state, $interval, authService, propertiesModel) {
+
+	var init = function () {
+	    const currentUrl = new URL(window.location.href);
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_portal/app/src/modules/public/login/LoginController.js
mattjackson220: done

traffic_portal/app/src/modules/public/login/login.tpl.html
mattjackson220: done

traffic_portal/app/src/common/api/AuthService.js
mattjackson220: done

traffic_portal/app/src/modules/public/login/login.tpl.html
mattjackson220: done

traffic_portal/app/src/common/api/AuthService.js
mattjackson220: done

traffic_portal/app/src/modules/public/login/LoginController.js
mattjackson220: done

mitchell852:
image

That button looks funny. Can you kill the `col-md-offset-1`

traffic_portal/app/src/traffic_portal_properties.json
```diff
@@ -407,6 +407,14 @@
           "url": "http://trafficcontrol.apache.org/"
         }
       ]
+    },
+    "oAuth": {
+      "_comment": "Opt-in OAuth properties for SSO login.  oAuthUrl is the url to redirect to for your oAuth provider.  oAuthTokenQueryParam is the query parameter key that the oAuth provider will use for the token (defaults to access_token).  redirectUriParameterOverride is to override if the oAuth provider requires a different key for the redirect_uri parameter (defaults to redirect_uri if empty). clientId is the client to be sent to the oAuth provider in the client_id field.",
```
mitchell852: Now that i think about it more, maybe it makes more sense to just have a link to http://traffic-control-cdn.readthedocs.io/en/latest/admin/quick_howto/oauth_login.html in here. what do you think?

traffic_ops/app/conf/cdn.conf
```diff
@@ -31,7 +31,8 @@
         "backend_max_connections": {
             "mojolicious": 4
         },
-	"profiling_enabled": false
+        "whitelisted_oauth_urls": [],
```
mitchell852: i can't decide if this should go inside of traffic_ops_golang or outside of it. maybe @dangogh has some ideas. not sure if the traffic_ops_golang node was reserved for certain properties or if they are arbitrary.

mitchell852:
If i forget to put the oauth url in the whitelisted_oauth_urls, i get a 500 ISE:

image

can you kick back a 401 instead?

docs/source/admin/traffic_portal/usingtrafficportal.rst
```diff
@@ -602,6 +602,8 @@ User management includes the ability to (where applicable):
 - update an existing user
 - view :term:`Delivery Service`\ s visible to a user
 
+.. Note:: If OAuth is enabled, creating/updating a user here will update the user's roles but the user needs to be created/updated with the OAuth provider as well.
```
mitchell852: a user just has one role

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
mitchell852: i kinda think if anything fails in the parsing of this token, just return a 401?

401 means "sorry, don't know you you are." you currently get that if:

1 - username and/or password is not found so TO doesn't know who you are
2 - mojo cookie is expired so TO doesn't know who you are

this kinda looks like #2

traffic_portal/app/src/traffic_portal_properties.json
mattjackson220: i like that! ill update it. i might keep the defaults in there too so that its a quick and easy thing to reference

ocket8888: In general linking to the latest docs is dangerous - someone who deploys 4.0 can get directed to the 5.0 docs.

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,80 @@
...
```
ocket8888: This still appears to be spaces... check your editor setting, it might be changing them automatically on save.

docs/source/admin/traffic_portal/usingtrafficportal.rst
ocket8888: Also, `Role` is a `term`

traffic_portal/app/src/traffic_portal_properties.json
mattjackson220: good call.  done

docs/source/admin/traffic_portal/usingtrafficportal.rst
mattjackson220: done. also changing 'updating' to 'deleting' since right now oauth only deals with user authentication and changing the user's role will not have any effect on it, so only creates/deletes need to be done in the oauth provider too

docs/source/admin/quick_howto/oauth_login.rst
mattjackson220: dang it thats exactly what's happening. i think i got it right this time. sorry about that

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -21,8 +21,13 @@ package login
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
+	"github.com/dgrijalva/jwt-go"
```
rawlinp: Did you get a chance to look into whether or not we could use one of these dependencies over the other, or do you think we still need them both?

For whatever we choose, we should probably run them by the mailing list just to make sure nobody has any issues with including those dependencies in the project. Once they're agreed upon, we typically vendor our go dependencies within our repo (probably in the traffic_ops_golang/vendor dir).

traffic_ops/app/conf/cdn.conf
dangogh: I think if it's something that's directly related to the execution of `traffic_ops_golang`,  it should be inside.   This looks like something outside of that.   But,  scanning the list, it doesn't look like that was adhered to,  so that's not a strong opinion....

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
dangogh: +1 - `fmt.Fprint` or `w.Write` would be (slightly) more efficient, but good practice...  we should revisit the LoginHandler code if it's doing that...

traffic_ops/traffic_ops_golang/login/login.go
dangogh: Also, do these have Apache-compatible licenses?   This shouldn't be merged until we know that for sure.

traffic_portal/app/src/modules/public/sso/SsoController.js
```diff
@@ -0,0 +1,33 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+var SsoController = function($scope, $state, $interval, $location, authService, propertiesModel) {
+
+	var init = function () {
+        const tokenParameter = propertiesModel.properties.oAuth.oAuthTokenQueryParam !== '' ? propertiesModel.properties.oAuth.oAuthTokenQueryParam : 'access_token';
+		const token = $location.search()[tokenParameter];
```
ocket8888: Looking at this implementation, you seem to be setting up an "Implicit Grant" ([RFC6749 Section 4.2](https://tools.ietf.org/html/rfc6749#section-4.2)) type of OAuth. However, the Implicit Grant specifies that the [Access Token Response](https://tools.ietf.org/html/rfc6749#section-4.2.2) will be found within parameters in

> _"the fragment component of the redirection URI using the 'application/x-www-form-urlencoded' format"_

Whereas [the documentation of `$location.search`](https://docs.angularjs.org/api/ng/service/$location#search) indicates it will pull out the "search part" (example shows that this is the query string) of a URL - **not** the fragment. So will this work for OAuth2-compliant servers?

traffic_ops/traffic_ops_golang/login/login.go
ocket8888: I believe they are Apache-compatible. They're both licensed under the MIT License, which - if I'm reading [the 3<sup>rd</sup>-party guidelines correctly](https://apache.org/legal/resolved.html#category-a) - are "similar" to Apache 2.0

mattjackson220: I tried to cut it down to just one of them but its not trivial / maybe not possible. i can respond to my email in the distro and point out that i added them in.

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
ocket8888: So I've just confirmed - if you login with an OAuth SSO user that doesn't exist in the database, this will return a 200 response. In Traffic Portal, this results in it trying to load the dashboard (or redirect page, I think) but then failing to load resources with 401 errors and the user is redirected to the login page and asked to log in again.

The proper behavior here is to return a 403 with some appropriate message, and having looked at the normal user login code I don't see it bailing without setting an error response in this case. It _does_ do the error log the same way as this, but does not immediately return afterward, which is the problem here.

traffic_ops/traffic_ops_golang/login/login.go
rawlinp: Ok, once the mailing list gives the thumbs up on these new dependencies, we will have to vendor them in our project. I would recommend doing that in a separate PR that just vendors these dependencies. Once that's merged, this PR can merge. Basically, we just `git clone` their repo into one of our vendor directories, and remove their `.git` directory so that git doesn't think it's a sub-repo.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3887/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3889/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3902/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3913/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3950/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3961/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3963/
Test PASSed.

docs/source/api/user_login_oauth.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Authentication of a user using an encrypted Json Web Token from an OAuth service. Traffic Ops will decode and validate the token, validate the key set domain, and send back a session cookie.
+
+:Auth. Required: No
+:Roles Required: None
+:Response Type:  ``undefined``
+
+Request Structure
+-----------------
+:t: Token
```
ocket8888: this part now needs to be updated to reflect the new expected request format

traffic_ops/traffic_ops_golang/server/servers_assignment.go
```diff
@@ -68,8 +68,8 @@ func AssignDeliveryServicesToServerHandler(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	if err := api.CreateChangeLogRawErr(api.ApiChange, "Assigned "+strconv.Itoa(len(assignedDSes))+" delivery services to server id: "+strconv.Itoa(server) , inf.User, inf.Tx.Tx); err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("error writing to change log: " + err.Error()))
+	if err := api.CreateChangeLogRawErr(api.ApiChange, "Assigned "+strconv.Itoa(len(assignedDSes))+" delivery services to server id: "+strconv.Itoa(server), inf.User, inf.Tx.Tx); err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("error writing to change log: "+err.Error()))
```
ocket8888: Did `go fmt` do this?

traffic_portal/app/src/modules/public/sso/SsoController.js
```diff
@@ -0,0 +1,41 @@
...
+        } else {
+            code = $location.search()['code'];
+        }
+        authService.oauthLogin(authCodeTokenUrl, code, clientId, clientSecret, redirectUri);
+	};
+	init();
+
+};
+
+SsoController.$inject = ['$scope', '$state', '$interval', '$location', 'authService', 'propertiesModel'];
```
ocket8888: Are you using `$interval` or `$state` anymore?

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,88 @@
...
+
+.. Note:: Ensure that the user names in the Traffic Ops database match the value returned in the `sub` field in the response from the OAuth provider when setting up with the OAuth provider.  The `sub` field is used to reference the roles in the Traffic Ops database in order to authorize the user.
```
ocket8888: Should mention that the part of the response it checks is the `access_token` - probably also wanna make mention of an RFC-compliant `application/json` response with `access_token` and other fields, vs the `application/octet-stream` for checking the entire body as a JWK

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
mattjackson220: done

docs/source/api/user_login_oauth.rst
mattjackson220: done

traffic_portal/app/src/modules/public/sso/SsoController.js
mattjackson220: nope. removed them

traffic_ops/traffic_ops_golang/server/servers_assignment.go
mattjackson220: i did to make the formatting consistent

docs/source/admin/quick_howto/oauth_login.rst
mattjackson220: done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3971/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3972/
Test FAILed.

infrastructure/cdn-in-a-box/traffic_ops/config.sh
```diff
@@ -98,7 +98,8 @@ cat <<-EOF >/opt/traffic_ops/app/conf/cdn.conf
         "max_db_connections": 20,
         "backend_max_connections": {
             "mojolicious": 4
-        }
+        },
+        "whitelisted_oauth_urls": ["*.xfinity.tv"]
```
jhg03a: I don't think you meant for this to be here.

mattjackson220: oh shoot good catch. deleted it

docs/source/api/user_login_oauth.rst
ocket8888: You updated the example, but the structure explanation still has just `t`

ocket8888: It looks like you just updated the example, but the structure explanation still just lists `t`

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
ocket8888: ... is it though? At least here on GitHub I'm still seeing it just return after the `log.Errorf` call. I've refreshed a few times to be sure I was looking at the right thing.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3973/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3974/
Test PASSed.

traffic_ops/traffic_ops_golang/login/login.go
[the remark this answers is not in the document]
mattjackson220: deleted that return.  now it will return a 403 with "Invalid user" if the user is either not in the db or doesnt have permission

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3975/
Test PASSed.

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +110,184 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+		}
+
+		data := url.Values{}
+		data.Add("code", parameters.Code)
+		data.Add("client_id", parameters.ClientId)
+		data.Add("client_secret", parameters.ClientSecret)
+		data.Add("grant_type", "authorization_code") // Required by RFC6749 section 4.1.3
+		data.Add("redirect_uri", parameters.RedirectUri)
+
+		req, err := http.NewRequest("POST", parameters.AuthCodeTokenUrl, bytes.NewBufferString(data.Encode()))
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +110,184 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		if err != nil {
+			log.Errorf("obtaining token using code from oauth provider\n%s", err.Error())
+			return
+		}
+
+		client := http.Client{}
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -104,3 +110,184 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+		data.Add("code", parameters.Code)
+		data.Add("client_id", parameters.ClientId)
+		data.Add("client_secret", parameters.ClientSecret)
+		data.Add("grant_type", "authorization_code") // Required by RFC6749 section 4.1.3
+		data.Add("redirect_uri", parameters.RedirectUri)
+
+		req, err := http.NewRequest("POST", parameters.AuthCodeTokenUrl, bytes.NewBufferString(data.Encode()))
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		if err != nil {
+			log.Errorf("obtaining token using code from oauth provider\n%s", err.Error())
```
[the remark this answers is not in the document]
mattjackson220: done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4054/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4055/
Test FAILed.

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,90 @@
...
+		+--------------------------+--------------------+-----------------------------------------------------------------------------------------------------------------+
+
+
+	.. code-block:: json
+		:caption: Example OAuth Configuration Properties In cdn.conf
+
+		{
+			"traffic_ops_golang": {
+				"whitelisted_oauth_urls": [
+					"example.oauth.com",
```
jhg03a: Generally one of the reserved RFC example URLs are specified rather than a well-known one.

docs/source/admin/traffic_portal/usingtrafficportal.rst
```diff
@@ -603,6 +621,8 @@ User management includes the ability to (where applicable):
 - update an existing user
 - view :term:`Delivery Service`\ s visible to a user
 
+.. Note:: If OAuth is enabled, creating/deleting a user here will update the user's :term:`role` but the user needs to be created/deleted with the OAuth provider as well.
```
jhg03a: This could probably be reworded for clarity that a given username must exist in both authentication systems, but the rights of a TrafficOps user are defined by the role assigned to the TrafficOps user.

docs/source/admin/quick_howto/oauth_login.rst
```diff
@@ -0,0 +1,90 @@
...
+
+.. Note:: Ensure that the user names in the Traffic Ops database match the value returned in the `sub` field in the response from the OAuth provider when setting up with the OAuth provider.  The `sub` field is used to reference the roles in the Traffic Ops database in order to authorize the user.
+
+.. Note:: OAuth providers sometimes do not return the public key set URL but instead require a locally stored key. This functionality is not currently supported and will require further development.
+
+.. Note:: The ``POST`` from the API to the OAuth provider to exchange the code for a token expects the response to have the token in JSON format with `access_token` as the desired field (and can include other fields).  It also supports a response with just the token itself as the body.  Further development work will need to be done to allow other resposne forms or other response fields.
+
```
jhg03a: There should be a note that a given username must exist in both authentication systems, but the rights of a TrafficOps user are defined by the role assigned to the TrafficOps user.

docs/source/api/user_login_oauth.rst
```diff
@@ -0,0 +1,81 @@
...
+	POST /api/1.4/user/login/oauth HTTP/1.1
+	Host: trafficops.infra.ciab.test
+	User-Agent: curl/7.47.0
+	Accept: */*
+	Cookie: mojolicious=...
+	Content-Length: 26
+	Content-Type: application/json
+
+	{
+		"authCodeTokenUrl": "https://url-to-convert-code-to-token.com",
```
jhg03a: Use a properly defined example URL.

docs/source/api/user_login_oauth.rst
```diff
@@ -0,0 +1,81 @@
...
+		"code": "AbCd123",
+		"clientId": "oauthClientId",
+		"clientSecret": "clientSecret",
```
jhg03a: Isn't the`clientSecret` privileged information generally held by the application rather than the client?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4056/
Test FAILed.

docs/source/admin/quick_howto/oauth_login.rst
mattjackson220: done

docs/source/admin/traffic_portal/usingtrafficportal.rst
mattjackson220: done

docs/source/admin/quick_howto/oauth_login.rst
mattjackson220: done

docs/source/api/user_login_oauth.rst
mattjackson220: done

docs/source/api/user_login_oauth.rst
mattjackson220: moved client secret to TO config

mattjackson220:
Rebase got messed up so here is a new version of this PR:
https://github.com/apache/trafficcontrol/pull/3763

PR #3522, opened 2019-04-25
TP: removes hard coding of 'Edge' from clone ds functionality because TO …

### LINKED ISSUE #2963: When cloning DS assignments from one cache to another, target cache list is incomplete
https://tp.domain.net/#!/servers/:id/delivery-services under `More` provides the ability to clone ds assignments from one cache to another, however, the target cache list is incomplete as the dropdown in this view:

image

is populated via this API call:

`GET api/1.3/servers?type=EDGE`

This is problematic because the system allow the addition of caches types such as:

`EDGE_FOO`

so currently, the targets list is limited to type=EDGE and you can't clone ds assignments from one EDGE_FOO cache to another EDGE_FOO cache.

mitchell852:
…allows you to create servers of type EDGE_* for example and those types were excluded from the clone ds assignment feature in TP. So after this change you can clone ds assignments from an EDGE_FOO server to another EDGE_FOO server instead of the target servers only being of type=EDGE

image

note: this only allows you to clone ds assignments from a source and target server where the type is the same. EDGE-->EDGE, EDGE_FOO --> EDGE_FOO, etc.

also, added a change log entry (not to be confused with CHANGELOG.md) when a bulk assignment of ds's to a server occurs.

- [x] This PR fixes #2963 

- Traffic Ops
- Traffic Portal

Navigate to a server's delivery services - i.e. https://localhost:8442/#!/servers/2591/delivery-services

under "More" you'll see:

image

in the list of target servers, you'll see those of the same type as the source server. clone and ensure, the ds's are assigned to the target server and check the change log as well: https://localhost:8442/#!/change-logs

Mostly UI code, no UI tests were added for this. No documentation changes required as this functionality was never documented in the first place.

- master (488039c)
- 3.0.1

### DIFF
--- a/traffic_portal/app/src/common/modules/table/serverDeliveryServices/TableServerDeliveryServicesController.js
+++ b/traffic_portal/app/src/common/modules/table/serverDeliveryServices/TableServerDeliveryServicesController.js
@@ -48,10 +48,10 @@ var TableServerDeliveryServicesController = function(server, deliveryServices, $
 
 	$scope.isEdge = serverUtils.isEdge;
 
-	$scope.cloneDsAssignments = function() {
+	$scope.cloneDsAssignments = function(server) {
 		var params = {
 			title: 'Clone Delivery Service Assignments',
-			message: "Please select an edge cache to assign these " + deliveryServices.length + " delivery services to.<br><br>Warning - Any delivery services currently assigned to the selected edge cache will be lost and replaced with these delivery service assignments...",
+			message: "Please select an " + server.type + " cache to assign these " + deliveryServices.length + " delivery services to.<br><br>Warning - Any delivery services currently assigned to the target cache will be lost and replaced with these delivery service assignments...",
 			labelFunction: function(item) { return item['hostName'] + '.' + item['domainName'] }
 		};
 		var modalInstance = $uibModal.open({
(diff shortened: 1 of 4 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3583/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3606/
Test PASSed.

traffic_portal/app/src/common/modules/table/serverDeliveryServices/TableServerDeliveryServicesController.js
```diff
@@ -63,7 +63,7 @@ var TableServerDeliveryServicesController = function(server, deliveryServices, $
 					return params;
 				},
 				collection: function(serverService) {
-					return serverService.getServers({ type: 'EDGE', orderby: 'hostName' });
+					return serverService.getServers({ type: server.type, orderby: 'hostName' });
```
ocket8888: This doesn't filter for the server itself - meaning that if I try to clone ds assignments of server A, server A itself will appear in the dropdown.

mitchell852: good point

mitchell852: i'll circle back on this one after restangular is removed

PR #3545, opened 2019-05-03
Delivery Service Overview Documentation

### LINKED ISSUE #3432: Documentation for STEERING_GEO and origins missing
Looks like the documentation related to `STEERING_GEO_*` is missing. Looks like there might be some dependency on the Origins configuration to get geo steering working. This origins configuration is not documented as well.

Please read the steering documentation as well as I believe it needs updated (we can set steering through UI): 
 
https://traffic-control-cdn.readthedocs.io/en/latest/admin/quick_howto/steering.html?highlight=steering

ocket8888:
- [x] This PR closes #3432 

Adds an entire page for Delivery Services to the "Overview" section. Also removes similar information from the TO administration pages that was TO UI-centric. 

Also added the "Origins" section to the TP using section, since the DS docs needed to reference it.

Added documentation for strings treated specially by ORT (was previously covered somewhat incompletely in the TO Using docs - removed)

Adds plural terms to the glossary (you no longer should need to use backslashes to make glossary terms plural)

Added a plugin to the documentation build system that provides the following:
```rst
.. impl-detail:: This is an implementation detail admonition. It's like ``note``, but specifically for implementation details.

:issue:`1` - this creates an inline link to ATC GitHub Issue #1 
:pr:`1` - this creates an inline link to ATC GitHub Pull Request #1
:atc-file:`LICENSE` - links to the LICENSE file at the root of the master branch on GitHub (editor, not raw)
```

- Documentation

Build the documentation and read it to verify that it's accurate, grammatically and syntactically correct, thorough, and well-written.

Unit tests for the sphinx plugin would be nice, but the only way to verify it renders properly is to have a human look at it. I've done that, but it can't be automated.

This would likely break in half if you tried to graft just it onto 3.x


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3617/
Test PASSed.

docs/source/glossary.rst
```diff
@@ -63,10 +65,15 @@ Glossary
...
+
+	Deep Coverage Zone File
+	Deep Coverage Zone Map
+		The :abbr:`DCZF (Deep Coverage Zone File)` or :abbr:`DCZM (Deep Coverage Zone Map)` maps network prefixes to "locations" - almost like the :term:`Coverage Zone File`. Location names must be unique, and within the file are simply used to group :term:`Edge-tier cache servers`. When a mapping is performed by Traffic Router, it will only look in the :abbr:`DCZF (Deep Coverage Zone File)` if the :term:`Delivery Service` to which a client is being directed makes use of :ref:`ds-deep-caching`. If the client's IP address cannot be matched by entries in this file, Traffic Router will fall back to geographic mapping using a database provided by the :term:`Delivery Service`'s :ref:`ds-geo-provider`.
```
rawlinp: First the DCZF is checked, then the CZF, then the GeoIP DB.

docs/source/overview/delivery_services.rst
```diff
@@ -0,0 +1,755 @@
...
+	| Name                  | Use(s)               | Type(s)                                              |
+	+=======================+======================+======================================================+
+	| deliveryservice_regex | Traffic Ops database | unique, integral identifier for a regular expression |
+	+-----------------------+----------------------+------------------------------------------------------+
+
+Max DNS Answers
+---------------
+The maximum number of :term:`Edge-tier cache server` IP addresses that the Traffic Router will include in responses to DNS requests for DNS-:ref:`Routed <ds-types>` Delivery Services. The :ref:`to-api` restricts this value to the range [1, 15], but no matching restraints are placed on the actual data as stored in the Traffic Ops Database. When provided, the :term:`cache server` IP addresses included are rotated in each response to spread traffic evenly. Ideally this number will reflect the amount of traffic - e.g. ``1`` for a trial Delivery Service with very little traffic, ``2`` for a small production Delivery Service. Add 1 for every 20 :abbr:`Gbps (Gigabits per second)` of traffic you expect at peak.
```
rawlinp: The 20Gbps number is highly dependent upon the cache's networking capabilities. I don't think we should be making recommendations like this in the documentation. I think it's sufficient enough to say that this number should be scaled with the amount of traffic you expect from this delivery service.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3645/
Test PASSed.

docs/source/overview/delivery_services.rst
```diff
@@ -0,0 +1,755 @@
...
+--------------------
+A boolean value that controls whether or not clients using IPv6 can be routed to this Delivery Service by Traffic Router. When creating a Delivery Service in Traffic Portal, this will default to "true".
+
+Info URL
+--------
+This should be a URL (though neither the :ref:`to-api` nor the Traffic Ops Database in any way enforce the validity of said URL) to which administrators or others may refer for further information regarding a Delivery Service - e.g. a related JIRA ticket.
+
+Initial Dispersion
+------------------
+The number of :term:`cache servers` between which traffic requesting the same object will be randomly split - meaning that if 4 clients all request the same object (one after another), then if this is above 4 there is a possibility that all 4 are cache misses, necessitating a fresh pull of the content from the next-highest level in the CDN. For most use-cases, this should be ``1``.
```
rawlinp: The information about cache misses seems irrelevant to this field. We should not this only applies to HTTP-type delivery services. Similar to the max_dns_answers field, we should just note that it should be scaled up depending on the amount of traffic you expect for assets of this delivery service.

docs/source/overview/delivery_services.rst
```diff
@@ -0,0 +1,755 @@
...
+
+Regional Geoblocking
+--------------------
+A boolean value that defines whether or not :ref:`Regional Geoblocking <regionalgeo-qht>` is active on this Delivery Service. The actual configuration of :ref:`Regional Geoblocking <regionalgeo-qht>` is done in the :term:`Profile` used by the Traffic Router serving the Delivery Service. Rules for this Delivery Service may exist, but they will not actually be used unless this field is ``true``.
+
+.. tip:: :ref:`Regional Geoblocking <regionalgeo-qht>` is configured primarily with respect to Canadian postal codes, so unless specifically Canadian regions should be allowed/disallowed to access content, `Geo Limit`_ is probably a better setting for controlling access to content according to geographic location.
+
+Routing Name
+------------
+The smallest DNS zone used to create an :abbr:`FQDN (Fully Qualified Domain Name)` used by clients to request content. All together, the constructed :abbr:`FQDN (Fully Qualified Domain Name)` looks like: :file:`{Delivery Service Routing Name}.{Delivery Service xml_id}.{CDN Subdomain}.{CDN Domain}.{Top-Level Domain}`\ [#xmlValid]_.
```
rawlinp: I wouldn't call it a zone, it's really a label. "A label in the Delivery Service's domain that forms the FQDN that is used by clients to request content."

docs/source/overview/delivery_services.rst
```diff
@@ -0,0 +1,755 @@
...
+DNS
+	Uses DNS content routing. Delivers content normally. This is the recommended Type for delivering smaller objects like web page images.
+DNS_LIVE\ [#dupOrigin]_
+	Uses DNS Content routing, but optimizes caching for live video streaming. Specifically, the configuration generated for :term:`cache servers` responsible for serving content for this Delivery Service will not cache that content on storage disks. Instead, they will make use of RAM block devices dedicated to ATS - as specified by the special ``RAM_Drive_Prefix`` and ``RAM_Drive_Letters`` :term:`Parameters`. Also, any :term:`Mid-tier` of caching is bypassed.
+DNS_LIVE_NATNL
+	Works exactly the same as DNS_LIVE, but is optimized for delivery of live video content across a wide physical area. What this means is that the :term:`Mi
```
rawlinp: TR doesn't really "eliminate all but the closest location to the client as possibilities." It still includes other locations as possibilities, ordered by proximity to the client's location. This is the same for STEERING_GEO_WEIGHT.

docs/source/overview/delivery_services.rst
ocket8888: I agree - that's just a copy of what was under "Using".

docs/source/overview/delivery_services.rst
ocket8888: I included the "cache misses" stuff because I don't think just that first part is enough to grasp what this is and why it matters. I know for sure it wasn't for me. It's not very well worded, though, and if an improvement can be made I'd love to do it.

docs/source/overview/delivery_services.rst
ocket8888: Uh, I think you put that on the wrong line. You're talking about STEERING_GEO_ORDER, right?

rawlinp: Yeah, not sure what happened there.

docs/source/overview/delivery_services.rst
rawlinp: "The number of edge caches that a particular asset will be distributed across within a cachegroup. For most use-cases, this should be 1, meaning that all clients requesting a particular asset will be directed to a 1 cache. Depending on the popularity and size of assets, you should consider increasing this number in order to spread the request load across more than 1 cache. The larger this number, the more copies of a particular asset are stored in a cachegroup, which pollutes caches (if load distribution is unnecessary) and decreases caching efficiency (due to cache misses if the asset is not requested enough to stay fresh in all the caches)."

How does that sound?

docs/source/overview/delivery_services.rst
```diff
@@ -0,0 +1,755 @@
...
+
+	+----------+----------------------------------------------+--------------------------------------------------------+
+	| Name     | Use(s)                                       | Type(s)                                                |
+	+==========+==============================================+========================================================+
+	| TenantID | Go code and :ref:`to-api` requests/responses | Integral, unique identifier (``bigint``, ``int`` etc.) |
+	+----------+----------------------------------------------+--------------------------------------------------------+
+
+Traffic Router Additional Response Headers
+------------------------------------------
+List of HTTP header ``{{name}}:{{value}}`` pairs separated by ``__RETURN__`` or simply on separate lines. Listed pairs will be included in all HTTP responses from Traffic Router for HTTP-:ref:`Routed <ds-types>` Delivery Services.
```
mitchell852: `__RETURN__` is still supported but new lines work as well. can you mention that anywhere where `__RETURN__` is mentioned.

docs/source/admin/traffic_ops/using.rst
```diff
@@ -286,485 +274,110 @@ These are the types of servers that can be managed in Traffic Ops:
...
+.. _asn-czf:
 
-+-----------------------+---------------+---------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-|      Name             |    Filename   | Default |                                                                                    Description                                                                    |
-+=======================+===============+=========+===================================================================================================================================================================+
-| psel.qstring_handling | parent.config | -       | Sets qstring handling without the use of regex remap for
```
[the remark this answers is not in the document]
ocket8888: yeah, the content of that section needs to be relocated to more appropriate places. This PR  is but one step in that direction.

Also I'm not sure what you mean; this line was deleted.

docs/source/overview/delivery_services.rst
ocket8888: Didn't I? I definitely did here, at least - did I miss it somewhere else?

docs/source/admin/traffic_router.rst
```diff
@@ -447,16 +417,16 @@ Steering Feature
...
 
 Some other points of interest
 """""""""""""""""""""""""""""
 - Steering is currently only available for HTTP :term:`Delivery Service`\ s that are a part of the same CDN.
 - A new role called STEERING has been added to the Traffic Ops database. Only users with the Steering :term:`Role` or higher can modify steering assignments for a :term:`Delivery Service`.
-- A new API has been created in Traffic Ops under ``/internal``. A Steering user can either directly access this API to modify assignments, or use the Traffic Portal UI (:menuselection:`More --> View Targets` on a Steering :term:`Delivery Service`'s details page), however a filter can only be created via the API.
-- Traffic Router uses the steering API in Traffic Ops to poll for steering assignments, the assignments are then used when routing traffic.
+- A new API has been created in Traffic Ops under ``/internal``. A Steering user can either directly access this API to modify assignments, or use the Traffic Portal UI (:menuselection:`More --> View Targets` on a Steering :term:`Delivery Service`'s details page), however a filter can only be created via the :ref:`to-api`.
```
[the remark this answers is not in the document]
ocket8888: Honestly, details about the TO-API and TP-UI don't belong in the TR section anyway.

docs/source/overview/delivery_services.rst
ocket8888: how's it look now?

docs/source/admin/traffic_ops/using.rst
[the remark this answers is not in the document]
ocket8888: Okay, I think I understand now why you commented that. In the "Conversation" tab it was showing this as attached to a different, deleted line. The one with profile parameter for MSO.

This section wasn't _created_ by me, I just _moved_ it. I can't delete it because things reference it and replacing the content is beyond the scope of what I'm trying to accomplish.

docs/source/admin/traffic_router.rst
```diff
@@ -386,40 +386,10 @@ DNS Specifics
 	|      |                                                                     | NXDOMAIN (the domain/name requested does not exist) |
 	+------+---------------------------------------------------------------------+-----------------------------------------------------+
 
-.. _tr-ngb:
-
-GeoLimit Failure Redirect Feature
```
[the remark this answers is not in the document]
ocket8888: Because it has nothing to do with the administration of Traffic Router; it only deals with the configuration of Delivery Service objects.

docs/source/admin/traffic_router.rst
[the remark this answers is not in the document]
ocket8888: I just removed that list item - does that work?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3654/
Test FAILed.

docs/source/overview/delivery_services.rst
```diff
@@ -0,0 +1,755 @@
...
+DNS
+	Uses DNS content routing. Delivers content normally. This is the recommended Type for delivering smaller objects like web page images.
+DNS_LIVE\ [#dupOrigin]_
+	Uses DNS Content routing, but optimizes caching for live video streaming. Specifically, the configuration generated for :term:`cache servers` responsible for serving content for this Delivery Service will not cache that content on storage disks. Instead, they will make use of RAM block devices dedicated to ATS - as specified by the special ``RAM_Drive_Prefix`` and ``RAM_Drive_Letters`` :term:`Parameters`. Also, a
```
rawlinp: This sounds better than before, but "choices ~by~ according to these groupings ~, and ordered according to~ and order them by physical distance from the client (closest to farthest)"

I think that would read better.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3656/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3657/
Test PASSed.

PR #3552, opened 2019-05-06
TP/TO/TR Consistent Hash with Query Parameters

dangogh:
Adds capability for a deliveryservice to use the presence of query parameters on 

- Documentation
- Traffic Portal
- Traffic Ops
- Traffic Router

Upgrade the Traffic Ops DB with the new migration to create the new table.

### Traffic Ops:
Using Traffic Ops API,  create a new deliveryservice (DS) using JSON with an extra key: `consistentHashQueryParams: ["paramZ", "paramA"]` (CHQP).   Again using the TO API, retrieve that DS and ensure that key with those parameter names (in lexicographical order) is present.  Ensure other DS's have the key with an empty array `[]` as the value (and not `null`).   Ensure that DS with a type other than one of the `HTTP.*` will not allow `consistentHashQueryParams` to be created either on creation or update.

Snapshot and examine the DS's section of the crconfig.   For the new DS, that same key/value pair should be present.  The key should not be present at all in DS's without that key set.

### Traffic Router:

Different combinations should be tested to ensure the consistent hash is being formed properly from the client url:
- a DS without an associated CHQP should map the url with and without query parameters to the same cache:
   e.g. `"consistentHashQueryParams": []`
          `http://x.y.z?a=0` and e.g. `http://x.y.z?a=1` should map to different caches.

- a DS with CHQP should map the url with a query param that is not in the CHQP list to the same cache as one without any query params at all:
   e.g. `"consistentHashQueryParams": [ "a", "b" ]`
          `http://x.y.z` and `http://x.y.z?c=1` should map to the same cache

- a DS with CHQP should map the url with a query param that is in the CHQP list and the same url with no query param to different caches (depending on number of servers assigned to the DS):
   e.g. `"consistentHashQueryParams": [ "a", "b" ]`
          `http://x.y.z` and `http://x.y.z?b=1` should map to different caches

- a DS with CHQP should map the url with a query param that is in the CHQP list with different values to different caches (depending on number of servers assigned to the DS):
   e.g. `"consistentHashQueryParams": [ "a", "b" ]`
          `http://x.y.z?b=0` and `http://x.y.z?b=1` should map to different caches


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3625/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3631/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3632/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3633/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3638/
Test PASSed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
```diff
@@ -125,6 +128,21 @@ public DeliveryService(final String id, final JsonNode dsJo) throws JsonUtilsExc
 		this.soa = dsJo.get("soa");
 		this.shouldAppendQueryString = JsonUtils.optBoolean(dsJo, "appendQueryString", true);
 
+		this.consistentHashQueryParams = new TreeSet<String>();
+		if (dsJo.has("consistentHashQueryParams")) {
+			final JsonNode cqpNode = dsJo.get("consistentHashQueryParams");
+			if (!cqpNode.isArray()) {
+				LOGGER.error("Delivery Service '" + id + "' has malformed consistentHashQueryParams. Disregarding.");
+			} else {
+				for (int i=0; i<cqpNode.size(); ++i) {
```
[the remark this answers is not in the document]
ocket8888: Probably. I just didn't know they were iterable like that - the fact that they return `null` for indices greater than the array size rather than throwing an exception spooked me.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -559,9 +560,17 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
 		return routeResult;
 	}
 
-	public String buildPatternBasedHashString(final DeliveryService deliveryService, final String requestPath) {
+	public String buildPatternBasedHashString(final DeliveryService deliveryService, final HTTPRequest request) {
+		final String requestPath = request.getPath();
 		if (deliveryService.getConsistentHashRegex() != null && !deliveryService.getConsistentHashRegex().isEmpty() && !requestPath.isEmpty()) {
-			return buildPatternBasedHashString(deliveryService.getConsistentHashRegex(), requestPath);
+			final StringBuilder hashString = new StringBuilder(request.getPath());
+			for (String qkey : request.getQueryString().split("&")) {
```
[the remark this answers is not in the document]
ocket8888: I don't think so, but does that matter? shouldn't affect anything afaik.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -559,9 +560,17 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+				qkey = qkey.split("=")[0];
+				if (deliveryService.consistentHashQueryParams.contains(qkey)) {
+					hashString.append(qkey);
```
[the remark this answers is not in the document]
ocket8888: > _"I thought we needed to include the query param value in the hash string as well, not just the key?"_

... yep. :man_facepalming: oops

> _"Also, we need to sort the query string key/values somehow..."_

I'm pretty sure `SortedSet<E>`s iterate in sorted order. I'd need to check the docs again to be sure (since you've just shaken my faith on that), but that's why I chose that interface.

> _"Also, how should we handle the case of /foo?a=b&a=c -- two keys with the same value? Just include both key/value pairs in the hash string? Later values overwrite earlier values? I'm thinking the former, which is along the lines of what this code is already doing."_

iirc we (@mitchell852, @dangogh, and myself) had decided on the former. Seems the most proper handling, IMO. Lemme know if I'm mis-remembering, guys.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
```diff
@@ -96,6 +98,7 @@
 	private final boolean redirectToHttps;
 	private final DeepCachingType deepCache;
 	private String consistentHashRegex;
+	public SortedSet<String> consistentHashQueryParams;
```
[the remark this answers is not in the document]
ocket8888: ugh, do I _havveeee tooo_? Getters/setters that only have return statements make me die inside.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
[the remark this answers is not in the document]
rawlinp: Well, we should probably figure that out. If it returned the decoded query string, then splitting on `&` won't work as expected if someone passes an encoded ampersand in the query string. There is also the question of what we should be hashing on -- decoded vs encoded? I'd think decoded, because `foo=%61` and `foo=a` should decode to the same thing, so we shouldn't hash them differently, right?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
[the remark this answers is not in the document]
rawlinp: Publicly mutable state is worse. Welcome to Javatown :)

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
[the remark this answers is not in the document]
ocket8888: I know for a fact that when initialized with a Java-standard `URL` object, it won't be decoded. I needed to check into that third party `HTTPRequest` to know what it does. It doesn't decode either.

> _"I'd think decoded..."_

I think that assumes that the origin is going to treat `%61` and `a` the same - which is probably true, maybe even _should_ be true, but doesn't _have_ to be true. I'm not sure if that's something we want to care about, though.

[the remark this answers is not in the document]
rob05c: >I think that assumes that the origin is going to treat %61 and a the same - which is probably true, maybe even should be true, but doesn't have to be true. 

I believe it is an RFC violation if it isn't.

https://tools.ietf.org/html/rfc3986#section-7.3
>Percent-encoded octets must be decoded at some point during the dereference process.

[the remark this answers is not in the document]
ocket8888: That the origin needs to decode is enough reason for me.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3641/
Test PASSed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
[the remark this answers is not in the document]
rawlinp: > I'm pretty sure SortedSet<E>s iterate in sorted order. I'd need to check the docs again to be sure (since you've just shaken my faith on that), but that's why I chose that interface.

Here we are iterating over the tokens of the query string, not the `deliveryService.consistentHashQueryParams` SortedSet, so I think we have to sort the tokens before iterating over them and adding them to the hashString.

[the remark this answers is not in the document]
ocket8888: What's wrong with me?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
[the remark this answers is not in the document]
ocket8888: Well I don't like it, but I did it.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3646/
Test PASSed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -560,16 +564,50 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+				for (final String qparam : request.getQueryString().split("&")) {
+					if (qparam.isEmpty()) {
+						continue;
+					}
+
+					String[] parts = qparam.split("=");
+					for (short i = 0; i<parts.length; ++i) {
+						parts[i] = URLDecoder.decode(parts[i]);
+					}
+					parts[0] = parts[0].toLowerCase(); // I truly have no idea why this triggers `PMD.UseStringBufferForStringAppends`
```
rawlinp: I'm not sure we should be lowercasing the input. That would make it case-insensitive, and I'm not sure query strings are generally case-insensitive.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -560,16 +564,50 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+
+					if (deliveryService.getConsistentHashQueryParams().contains(parts[0])) {
+						if (parts.length == 2) {
+							parts[1] = URLDecoder.decode(parts[1]);
```
rawlinp: this should've already been decoded up on L592

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -560,16 +564,50 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+						}
+						qparams.add(URLEncoder.encode(String.join("=", parts)));
```
rawlinp: Why are we re-encoding before appending to the hashString? I thought we wanted to hash on the decoded qparams.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -560,16 +564,50 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+					}
 				}
+
+				for (final String qparam : qparams) {
+					hashString.append(URLEncoder.encode(qparam));
```
rawlinp: ditto my comment on L600

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3660/
Test PASSed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: I'm pretty sure the database constraint on uniqueness is case-insensitive (or at least it was at one point - maybe that's changed?) so this is necessary for consistency.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: The point was to compare things in an encoding-consistent way. But I guess that's accomplished by just doing the decoding.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: After discussion, the db constraint _was_ case-insensitive, but isn't anymore. should be changed in my latest.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3663/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3664/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3666/
Test PASSed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -559,9 +563,47 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+			final SortedSet<String> qparams = new TreeSet<String>();
+			try {
+				for (final String qparam : request.getQueryString().split("&")) {
+					if (qparam.isEmpty()) {
+						continue;
+					}
+
+					String[] parts = qparam.split("=");
+					for (short i = 0; i<parts.length; ++i) {
+						parts[i] = URLDecoder.decode(parts[i]);
```
rawlinp: Apparently this method is deprecated because the method that specified the encoding should be used instead ("UTF-8" I believe?): https://docs.oracle.com/javase/8/docs/api/java/net/URLDecoder.html#decode-java.lang.String-java.lang.String-

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -559,9 +563,47 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+
+					if (deliveryService.getConsistentHashQueryParams().contains(parts[0])) {
+						qparams.add(String.join("=", parts));
+					}
+				}
+
+				for (final String qparam : qparams) {
+					hashString.append(qparam);
+				}
+			} catch (Exception e) {
```
rawlinp: We should probably log an error here as well instead of silently ignoring whatever it was that caused this exception.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -846,11 +888,39 @@ protected CacheLocation getCoverageZoneCacheLocation(final String ip, final Deli
...
+	 * @return A cache object chosen to serve the client's request
+	 */
 	public Cache consistentHashForCoverageZone(final String ip, final String deliveryServiceId, final String requestPath, final boolean useDeep) {
+		final HTTPRequest r = new HTTPRequest();
+		r.setPath(requestPath);
+		r.setQueryString("");
+		return consistentHashForCoverageZone(ip, deliveryServiceId, r, useDeep);
+	}
+
+	public Cache consistentHashForCoverageZone(final String ip, final String deliveryServiceId, final HTTPRequest request) {
```
rawlinp: My IDE is flagging this method as never used. Can it be deleted?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -893,12 +978,19 @@ public Cache consistentHashForGeolocation(final String ip, final String delivery
...
 
 	public String buildPatternBasedHashStringDeliveryService(final String deliveryServiceId, final String requestPath) {
-		return buildPatternBasedHashString(cacheRegister.getDeliveryService(deliveryServiceId), requestPath);
+		final HTTPRequest r = new HTTPRequest();
+		r.setPath(requestPath);
+		r.setQueryString("");
+		return buildPatternBasedHashString(cacheRegister.getDeliveryService(deliveryServiceId), r);
+	}
+
+	public String buildPatternBasedHashStringDeliveryService(final String deliveryServiceId, final HTTPRequest request) {
```
rawlinp: My IDE is flagging this method as never used. Can it be deleted?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -559,9 +563,47 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
...
+				}
+
+				for (final String qparam : qparams) {
+					hashString.append(qparam);
+				}
+			} catch (Exception e) {
+				return requestPath;
+			}
+
+			return buildPatternBasedHashString(deliveryService.getConsistentHashRegex(), hashString.toString());
```
rawlinp: So I think the pattern-based consistent hashing feature was meant to only be for the request path, but now we're adding on the consistent hash query parameters. Is that going to be a problem? Should we build the pattern-based consistent hash string first (using just the path), _then_ append any consistent hash query params onto the path?

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
```diff
@@ -96,6 +98,7 @@
 	private final boolean redirectToHttps;
 	private final DeepCachingType deepCache;
 	private String consistentHashRegex;
+	private final SortedSet<String> consistentHashQueryParams;
```
rawlinp: This doesn't really need to be a sorted set since we don't have to iterate over the set in a sorted manner. Right now it's only being used for `contains()` so maybe a HashSet makes more sense.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: Oh yeah, I remember reading that. I don't think you can just give it literally `"UTF-8"` though, because it failed to compile I think... You're supposed to give it some kind of `Charset` object I think. I'll try to figure that out.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: but it's used literally 3 lines above its declaration...

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: same as above

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: I thought it'd be fine to do the regex on the whole path+query params. I don't _think_ it would bother things only meant to match on paths  - unless maybe you have a path in a param value? 

It sort of expands the functionality of the regex consistent hashing, which I thought would be neat - like `.*\?test=\d+.*` to consistent hash based on the `test` param but only if it's a number or something.

Idk, do *you* think it'll cause problems? We haven't issued a release with CHR in it, so idk if breaking that behavior is a problem in itself.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
ocket8888: I thought about changing this, but the intent is to match the TO API declared representation - it's technically an array because JSON doesn't have proper sets but it _does_ sort the output.

It's probably not a big deal either way, and you have way more experience with Traffic Router (and probably just Java in general), so having said that if you want me to change it I don't have a problem doing that.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: nvm - this actually is getting bypassed. that other one though...

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: No, I'm mistaken. Must've misspelled it or something else was wrong.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
rawlinp: I don't think we should expand the functionality of the regex consistent hashing to apply to the consistent hash query params (CHQPs). If we can keep the two features separate, they're easier to reason about. With CHQPs we're saying "these are the query parameters that change the content of a request path". Since those query parameters produce different content for the same path, we should hash them differently. I'm not sure we should take that any further by being able to configure something like "only consider this query param if its value matches this pattern". Trying to come up with a single regex that extracts a hash string from the path *and* the qstring seems overly complicated.

It just seems like keeping the two features separate for now would make them both easier to use on their own. If we have a requirement in the future to only consider certain query parameters if they match a pattern, then I'd think each qparam that had to match a pattern would get its own regex instead of trying to configure patterns for all of those qparams in the one consistent hash regex.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
rawlinp: It just seems weird to use a TreeSet when it makes more sense to use a HashSet based on how it's used in TR. TreeSets are less optimal for `contains()` compared to HashSet, and we're never iterating over the set in a sorted manner.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
ocket8888: If you say so ¯\_(ツ)_/¯. Done (pending branch merge).

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
ocket8888: Done (pending branch merge).

dangogh: only reason TO sorts them is so they don't come out differently for 2 dses with the same set.  Makes comparison more easily automated.

ocket8888: It's probably best not to tie ourselves to sorting it anyway.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3673/
Test PASSed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
dangogh: I disagree.   no reason for TR to attempt to maintain the order,  but TO should.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3678/
Test FAILed.

ocket8888:
Idk why it doesn't show me the snippet you're commenting on, so imma just do this:

> _"TR still performs consistent hashing for DNS-routed delivery services to determine which caches to return in the DNS response, it's just not at the "request path" level."_

Thanks, I was very unsure about that.

> _"why was this marked http only? I think this primarily applies to DNS delivery services"_

Not sure. I did a lot of copy/pasting those huge DS descriptions around (now that there's a dedicated overview page I probably wanna go through and clean that up to just describe the type and link to the dedicated section - time permitting) and I wrote that before I knew it was possible to name footnotes in reST.

> _"I'm not entirely positive that this regex isn't used for matching the name in a DNS request (e.g. `.*\.myds\..*` would match a `dig foo.myds.mycdn.example.com A` DNS request). I think it applies to both types."_

You're not? I am! I'm positive that they *are* used for matching the name in a DNS request. Those footnotes need to get shifted down one - they should apply to `HEADER` and `PATH`, not `HOST` and `HEADER`. Not sure how/when that happened...

> _"Is there any way we can _not_ duplicate this same information across multiple files? For any deliveryservice endpoint, this information should be the same. Can we have one shared common file with all this information that can then be `#include` in the various places that need it?"_

Yeah I sort of mentioned that above. These should (probably) all just be links to the dedicated section, now that it exists. It's outside the scope of this PR, but it's on my radar - and it's why I chose the DS section of the "Using Traffic Ops" page to pull out first, instead of just going top-down. I really hate the way that was all copy/pasted sloppily around, and very nearly wrote them into a macro instead. But at the time getting api routes accurately, consistently, and thoroughly documented took precedence over best practices.

tl;dr I was lazy, will fix.

> _"it would be good to include some more context around the error here, like the delivery service xml_id, so that it could be more easily tracked down to a bad delivery service config or client (ditto for L627)"_

Does Java do 'reprs'? can I just log the object or will that log a pointer value or something and I need to pick an identifying property like the xmlid?

ocket8888:
> _"Does Java do 'reprs'? can I just log the object or will that log a pointer value or something and I need to pick an identifying property like the xmlid?"_

nvm, I found `DeliveryService.toString()`

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3680/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3686/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3692/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3694/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3716/
Test FAILed.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/DeliveryService.java
```diff
@@ -658,4 +682,61 @@ public boolean isAcceptHttp() {
...
+	/**
+	 * Extracts the significant parts of a request's query string based on this
+	 * Delivery Service's Consistent Hashing Query Parameters
+	 * @param r The request from which to extract query parameters
+	 * @return The parts of the request's query string relevant to consistent
+	 *	hashing. The result is URI-decoded - if decoding fails it will return
+	 *	a blank string instead.
+	 */
+	public String extractSignificantQueryParams(HTTPRequest r) {
+		if (r.getQueryString() == null || r.getQueryString().isEmpty()) {
```
rawlinp: We should probably also add `|| this.getConsistentHashQueryParams().isEmpty()` to this conditional -- if it's empty the result should always be `""`.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3718/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3719/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3720/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3722/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3724/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3725/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3727/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3730/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3754/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3755/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3756/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3777/
Test PASSed.

PR #3556, opened 2019-05-08
Default unavailable_server_retry_responses

smalenfant:

This avoid warnings in Traffic Server logs by setting the default value explicitly

```
[May  7 18:48:34.899] Server {0x2b8caf4ddd40} WARNING: [ParentSelection] initializing UnavailableServerResponseCodes on line 2 to 503 default.
[May  7 18:48:34.899] Server {0x2b8caf4ddd40} WARNING: UnavailableServerResponseCodes - unavailable_server_retry_responses is null loading default 503 code.
[May  7 18:48:34.901] Server {0x2b8caf4ddd40} WARNING: [ParentSelection] initializing UnavailableServerResponseCodes on line 4 to 503 default.
[May  7 18:48:34.901] Server {0x2b8caf4ddd40} WARNING: UnavailableServerResponseCodes - unavailable_server_retry_responses is null loading default 503 cod
```

- Traffic Ops
- Traffic Ops ORT
- Documentation: Not required, just avoid a warning.

Run ORT against Traffic Server and verify not warning or errors displayed in diags.log

### DIFF
--- a/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
+++ b/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
@@ -2526,7 +2526,7 @@ sub parent_dot_config { #fix qstring - should be ignore for quika
 			my $multi_site_origin                  = $ds->{multi_site_origin} || 0;
 			my $mso_algorithm                      = $ds->{'param'}->{'parent.config'}->{'mso.algorithm'} || 'consistent_hash';
 			my $parent_retry                       = $ds->{'param'}->{'parent.config'}->{'mso.parent_retry'} || 'both';
-			my $unavailable_server_retry_responses = $ds->{'param'}->{'parent.config'}->{'mso.unavailable_server_retry_responses'} || '';
+			my $unavailable_server_retry_responses = $ds->{'param'}->{'parent.config'}->{'mso.unavailable_server_retry_responses'} || '"503"';
 			my $max_simple_retries                 = $ds->{'param'}->{'parent.config'}->{'mso.max_simple_retries'} || 1;
 			my $max_unavailable_server_retries     = $ds->{'param'}->{'parent.config'}->{'mso.max_unavailable_server_retries'} || 1;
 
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3640/
Test PASSed.

traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm
```diff
@@ -2617,13 +2617,8 @@ sub parent_dot_config { #fix qstring - should be ignore for quika
 				}
 				$text .= "$parents round_robin=$mso_algorithm qstring=$parent_qstring go_direct=false parent_is_proxy=false";
 
-				if ( $ats_major_version >= 6 && $parent_retry ne "" ) {
-					if ( $unavailable_server_retry_responses ne "" && $unavailable_server_retry_responses =~ /^"(?:\d{3},)+\d{3}"\s*$/) {
```
ocket8888: You removed the validation this does on the param - it used to check for a valid 3-digit response code.

smalenfant: @ocket8888 Would you know what happens if not validated? Traffic Server crash? If so, should have been documented. 

I removed the validation because there was now defaults for all these params. 

What should be the behavior if I add the validation? Add the "503" default and not fail?

ocket8888: Well the old behavior was to fall back on just `parent_retry=$parent_retry`, so you could do that. But you could also just fall back on the default. In fact, that's probably more appropriate since it used to fall back on `parent_retry=$parent_retry` if it failed validation _or_ if it wasn't defined. So the most consistent thing to do would be adding the 503 and not fail - but you may want to log a warning.

As far as what would happen to ATS, I believe that in general parsing failures would cause either the field or the line to fail, depending on the ATS version and the file in question. Meaning it wouldn't crash, but the configuration won't be applied and it will log an error.

elsloo:
@smalenfant wouldn't it be better to keep these error messages as an indicator that the configuration is missing? One could set up monitoring to look for this, which would allow them to reconfigure MSO accordingly. This seems like it's a little magical and someone might not realize that they're only failing over on 503s. Swallowing the error via forcing the default in TO gives the user no indication that they have an incomplete configuration.

Thoughts?

alficles:
Yeah, I tend to agree here. :/ ATS is warning you that it loaded a default value. Setting a default here also means that you're not inheriting a new default if ATS ever changes it, which also seems bad. I think ATS is probably doing the right thing and we shouldn't generally suppress it. People already have several solutions to this problem: set their log parsers to ignore the message, set the value to 503 explicitly, and so on, depending on their exact goals.

smalenfant:
I will add/keep the defaults to 503, this will match other parameters that have defaults.
I will put validation back in and modify as current validation requires quotes, quotes will be stripped. ATS DOES NOT requires the quotes. 

Question. Should I add validation and warning about non 5xx codes?

In the end, I don't want to see those WARNINGs all the time. There was no warning in 5.3.2 and defaults were taken.

jrushford:
I think you should validate, ATS will reject anything other than 5xx codes and not load that config line.

PR #3565, opened 2019-05-10
TO: Fix 404 errors

### LINKED ISSUE #3448: TO: 404 StatusNotFound not implemented correctly for a few endpoints
In cachegroups the delete will say "deleted cg" even if the cachegroup isn't found. It is caused by the line `return nil, nil, http.StatusNotFound`. I will be making a PR soon for this case, however the same line exists in a few other places that also need to be fixed.

Steering targets:
`steeringtargets/steeringtargets.go:237 return nil, nil, http.StatusNotFound`
`steeringtargets/steeringtargets.go:259 return nil, nil, http.StatusNotFound`

Origins:
`origin/origins.go:316 return nil, errors.New("origin update: no rows returned"), http.StatusInternalServerError`

The origins delete also doesn't implement the 404 case well either. It has `return nil, nil, http.StatusNotFound`, but it actually returns an internal server error earlier in the function.

In all cases the use of `NamedQuery` is unnecessary. A `QueryRow` should be used instead and then check for sql.ErrNoRows.

moltzaum:

This PR fixes #3448

The origins and steering targets did not return 404 errors for updating or deleting.

- Traffic Ops

Try to update an origin or steering target that doesn't exist.

As far as I know, this has been an issue since each endpoint got rewritten.

    - No new features or significant changes
    - No new features or significant changes

### DIFF
--- a/traffic_ops/testing/api/v14/origins_test.go
+++ b/traffic_ops/testing/api/v14/origins_test.go
@@ -16,6 +16,7 @@ package v14
 */
 
 import (
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3655/
Test PASSed.

PR #3573, opened 2019-05-13
Client steering forced diversity

rawlinp:

Add a Traffic Router profile parameter that enables the Client Steering Forced Diversity feature. With this feature enabled, Traffic Router will modify its default consistent-hashing behavior w.r.t. CLIENT_STEERING in order to force a more diverse result (one that includes more unique edge caches). This is beneficial for clients to have a better chance at recovery in case of edge cache failures, because they are more likely to retry a different edge cache. With the default behavior, a client can sometimes receive a result with the same edge multiple times, causing a client to potentially retry a failed node multiple times.

- Documentation
- Traffic Router

1. Run the Traffic Router tests: ```mvn clean verify -Djava.library.path=/usr/local/opt/tomcat-native/lib # replace path with your system's path to tc-native```
2. Manually: create a parameter called `client.steering.forced.diversity` with file = CRConfig.json and value = "false", link it it your Traffic Router profile, and snapshot. Make requests to a CLIENT_STEERING delivery service and observe the default (non-diverse) responses. Now, set the value to "true", snapshot, and make the same requests. Observe that the responses contain a more diverse set of edges for all requests.

### DIFF
--- a/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
+++ b/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
@@ -547,14 +547,18 @@ public class TrafficRouter {
 		for (final SteeringResult steeringResult : steeringResults) {
 			final DeliveryService ds = steeringResult.getDeliveryService();
 
-			final List<Cache> caches = selectCaches(request, ds, track);
+			List<Cache> caches = selectCaches(request, ds, track);
 
 			// child Delivery Services can use their query parameters
 			final String pathToHash = steeringHash + ds.extractSignificantQueryParams(request);
 
 			if (caches != null && !caches.isEmpty()) {
-				if (isClientSteeringDiversityEnabled() && caches.size() > selectedCaches.size()) {
-					caches.removeAll(selectedCaches);
+				if (isClientSteeringDiversityEnabled()) {
+				    final List<Cache> tryCaches = new ArrayList<>(caches);
+				    tryCaches.removeAll(selectedCaches);
+				    if (!tryCaches.isEmpty()) {
+						caches = tryCaches;
+					}
 				}
 				final Cache cache = consistentHasher.selectHashable(caches, ds.getDispersion(), pathToHash);
 				steeringResult.setCache(cache);
--- a/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
+++ b/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
@@ -566,6 +566,7 @@ public class TrafficRouter {
 					} else if (track.result == ResultType.DEEP_CZ) {
 						// deep caches have been selected already, try non-deep selection
 						tryCaches = selectCaches(request, ds, track, false);
+						track.setResult(ResultType.DEEP_CZ); // request should still be tracked as a DEEP_CZ hit
 						tryCaches.removeAll(selectedCaches);
 						if (!tryCaches.isEmpty()) {
 							caches = tryCaches;
--- a/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
+++ b/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
@@ -81,6 +81,7 @@ import com.comcast.cdn.traffic_control.traffic_router.core.loc.AnonymousIpDataba
 public class TrafficRouter {
 	public static final Logger LOGGER = Logger.getLogger(TrafficRouter.class);
 	public static final String XTC_STEERING_OPTION = "x-tc-steering-option";
+	public static final String CLIENT_STEERING_FORCED_DIVERSITY = "client.steering.forced.diversity";
 
 	private final CacheRegister cacheRegister;
 	private final ZoneManager zoneManager;
@@ -89,6 +90,7 @@ public class TrafficRouter {
 	private final AnonymousIpDatabaseService anonymousIpService;
 	private final FederationRegistry federationRegistry;
 	private final boolean consistentDNSRouting;
+	private final boolean clientSteeringForcedDiversityEnabled;
 
 	private final Random random = new Random(System.nanoTime());
 	private Set<String> requestHeaders = new HashSet<String>();
@@ -114,6 +116,7 @@ public class TrafficRouter {
 		this.anonymousIpService = anonymousIpService;
 		this.federationRegistry = federationRegistry;
 		this.consistentDNSRouting = JsonUtils.optBoolean(cr.getConfig(), "consistent.dns.routing");
+		this.clientSteeringForcedDiversityEnabled = JsonUtils.optBoolean(cr.getConfig(), CLIENT_STEERING_FORCED_DIVERSITY);
 		this.zoneManager = new ZoneManager(this, statTracker, trafficOpsUtils, trafficRouterManager);
 
 		if (cr.getConfig() != null) {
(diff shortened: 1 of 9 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3661/
Test FAILed.

mitchell852:
nice work! any reason why this doesn't just override the default behavior. In other words, why does it have to be enabled via a parameter?

rawlinp:
Thanks! I was mostly for just changing the behavior without requiring a new parameter, but if we don't like the behavior for some reason, this allows us to turn the feature off without having to downgrade. If in the future we think the new behavior should be standard, I guess we can bring it up on the mailing list to remove the parameter and just go forward with the new behavior.

traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/external/SteeringTest.java
```diff
@@ -553,4 +556,54 @@ public void itUsesMultiLocationFormatWithMoreThanTwoEntries() throws Exception {
...
+				String[] parts = l.split("\\.");
+				parts[0] = parts[0].replaceFirst("http://", "");
+				actualEdges.add(parts[0]);
+				actualTargets.add(parts[1]);
+			}
+
+			// with client.steering.forced.diversity = true, expect to get all 3 edges back in the locations
+			// by default (client.steering.forced.diversity = false), only one edge would be returned in the locations
+
+			String[] expectedEdgesArray = {"edge-cache-csd-1", "edge-cache-csd-2", "edge-cache-csd-3"};
```
Daisykh: Would it be beneficial to add some more testing on some edge cases? Here's what my initial thoughts were:
- Test on no locations (empty failure case)
- Test on 2 locations available only (I think this one is pretty important)
- Test on different DS weighting values

traffic_router/core/src/test/resources/publish/CrStates
```diff
@@ -146,6 +149,30 @@
       "disabledLocations": [],
       "isAvailable": true
     },
+    "client-steering-diversity-test": {
```
Daisykh: Not sure if it would be of value to test CS diversity with a case where disabled locations for a target is populated.

rawlinp: I think the `disabledLocations` list is mostly just an optimization (an unnecessary one IMO but that's a different issue) for Traffic Monitor to tell Traffic Router that entire cachegroups are currently unavailable for a delivery service so that TR can eliminate those cachegroups from its "cache selection process". The CSD stuff isn't really changing that cache selection process since CSD is operating on the _result_ of that process to diversify the end result. So, I'm thinking that test case would add more coverage around the "cache selection process" but not really CSD per se.

traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/external/SteeringTest.java
rawlinp: So this integration test basically just verifies that CSD provides the expected level of diversity. With 3 available caches and 5 targets, we should expect to see all 3 caches at least once and all 5 targets in the response.

Can you elaborate on what functionality we would expect/verify for those three bullets? They seem to mostly apply to client-steering in general rather than to the expected results of CSD.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -537,8 +542,12 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
 			final List<Cache> caches = selectCaches(request, ds, track);
 
 			if (caches != null && !caches.isEmpty()) {
+				if (isClientSteeringDiversityEnabled() && caches.size() > selectedCaches.size()) {
+					caches.removeAll(selectedCaches);
+				}
 				final Cache cache = consistentHasher.selectHashable(caches, ds.getDispersion(), pathToHash);
 				steeringResult.setCache(cache);
+				selectedCaches.add(cache);
```
ajschmidt: Once all of the caches in the coverage zone have been added to the selectedCaches it looks like forced diversity becomes disabled and we revert to the standard consistent hash over all the caches. Is this the desired behavior? I would guess that you would want to clear the selectedCaches and start over.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -537,8 +542,12 @@ public HTTPRouteResult multiRoute(final HTTPRequest request, final Track track)
 			final List<Cache> caches = selectCaches(request, ds, track);
 
 			if (caches != null && !caches.isEmpty()) {
+				if (isClientSteeringDiversityEnabled() && caches.size() > selectedCaches.size()) {
+					caches.removeAll(selectedCaches);
```
ajschmidt: Is it safe to assume that the same cache won't appear in the 'caches' list more than once? If so then it seems like you could end up with an empty 'caches' list here.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
rawlinp: Yeah, that is the desired behavior that I wanted. We can't really assume that the set of caches assigned to each target is the same (even though in practice that might be true), otherwise I'd think we would do what you said and clear the selectedCaches to start over.

The effect this has is diversifying the result list "enough". E.g. with 3 available caches and 5 targets:
```
edge2.target1
edge1.target2
edge3.target3
<edge 1-3>.target4
<edge 1-3>.target5
```
The last two targets could be any of edges 1 through 3, but by that point we know the result already contains all the possible edges, so by the time we have failed over to target4, either all 3 edges were down, the first 3 target origins were down, or some mix of both. So we've at least covered our bases by trying all 3 possible caches before reaching targets 4 and 5 which could contain duplicates.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
rawlinp: Yes, I believe that is safe to assume. `selectCaches` should never return a list with duplicate caches.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3677/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3696/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3706/
Test FAILed.

JBevillC:
I was trying to test this in CiaB and it appears that the both the CZF and deep CZF json file will need to be updated for the integration test to work properly (at least for the correct code path(s) to be tested).  Both the CZF and dCZF json files in CiaB are completely invalid, so all results are based on the maxmind dataset.  Given that CiaB contains almost zero cache group data , it is a lot of work to test this PR in CiaB without writing a script to add a ton of data to TO.

For local testing (outside of CiaB), I haven't tested the latest commits from yesterday, but it does seem necessary to update the deep CZF json file to make sure all code paths are tested with respect to the changes of this PR.  In my testing, it seems just the CZF is being hit and not the dCZF which suggests that the data in the integration test `dczmap.json` file may not be valid.

For example the IPs in `$src/trafficcontrol/traffic_router/core/src/test/resources/dczmap.json` seem to be completely ignored by the traffic router integration test and the fail back is to only return a CZF result (which then gets filtered by the diversity code).

```
{
	"deepCoverageZones":
	{
		"cache-group-01":
		{
			"network6":
			[
				"1234:5678::/64",
				"1234:5679::/64"
			],
			"network":
			[
				"192.168.8.0/24",
				"192.168.9.0/24"
			],
			"caches":
			[
				"host1",
				"host2",
				"host3",
				"host4"
			]
		}
	}
}
```

The only way to test this in CiaB is add a ton of cache groups and associated data with a script and also populate the deep CZF with randomly allocated IP/subnets for each edge cache once they have been created via this same script.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3731/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3815/
Test PASSed.

PR #3580, opened 2019-05-16
TO: Added ReadBack function to fix "joined fields" issue

moltzaum:

This PR should solve #2494 for the majority of reported cases, although there might be a few cases where the issue isn't fixed. The cachegroups case is fixed in a separate PR.

- Traffic Ops

POST and PUT json objects mentioned in the issue. All of these should be fixed:

- [x] POST, PUT /api/1.1/asns
    - cachegroup
- [x] POST, PUT /api/1.3/deliveryservices
    - cdnName, profileName, profileDescription
- [x] POST, PUT /api/1.1/regions
    - divisionName
- [x] POST, PUT /api/1.1/phys_locations
    - region
- [x] POST, PUT /api/1.1/servers
    - cachegroup, cdnName, physLocation, profile, profileDesc, status, type
- [x] POST, PUT /api/1.1/tenants
    - parentName

EDIT: It seems deliveryservices and servers is not fixed atm. I'm looking into it. The `profileparameters` case is not fixed either.

EDIT 2: I fixed the servers case. The deliveryservices endpoint does not use the `api.CreateHandler`, so the endpoint must be edited directly.

EDIT 3: The `profileparameters` endpoint uses the `TOProfileParameter` object (not the `TOProfileParameters` one), so even after writing the ReadBack function to work with non `id` keys, I had type errors between the `TOProfileParameter` and `TOProfileParameters` struct. I think I'm going to leave it for now. I was thinking of moving the readback logic to the shared `api.CreateHandler`, but since it doesn't help with the deliveryservices or profileparameters, I'm leaving it for now as well.

This bug has existed since each endpoint was originally written. That was awhile ago.

    - In the client the object itself is not returned back, but just the `tc.Alert`
    - For the test case performing a followup GET does not make sense
    - I don't see an easy way to make a test with our current api test setup
    - I can make a test with curl easily I think, but I'd rather not write a test completely differently
    - Updated asns, phys_locations, servers, and tenants
    - Documentation for deliveryservices and regions looked fine
    - I don't think an update to CHANGELOG.md is necessary


moltzaum:
The servers endpoint isn't fixed because the server does not implement the `GenericReader`. Previously I had a case working that used `Read` instead of the `GenericReader`, but I had to change the function signature for Create to be: `Create() ([]interface{}, error, error, int)` like the Read function. I didn't like that approach since I don't think it makes sense intuitively for a Create method to return an array of interfaces, especially since it would not benefit endpoints that don't use the generic create method.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3670/
Test PASSed.

moltzaum:
I made two functions. One called `ReadBack` and another called `GenericReadBack`. In the `ReadBack` function I convert an interface to the object of our choice by marshaling and unmarshaling json. I originally had both methods because the `GenericReadBack` is the common case and is faster since it doesn't marshal and unmarshal, then I had the `ReadBack` function that calls the `Read` function. The `ReadBack` function was supposed to be the backup and not the main method since I perceived it to be slower, but I just now deleted the `GenericReadBack` function since both functions do the same thing.

Should I use both methods?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3672/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3675/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3676/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3685/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3687/
Test PASSed.

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -80,8 +81,33 @@ func GenericCreate(val GenericCreator) (error, error, int) {
...
+
+	return nil, nil, http.StatusOK
+}
+
+func ReadBack(val Reader) (error, error, int) {
+
+	i := val.(Identifier)
+	keys, _ := i.GetKeys()
+	inf := val.APIInfo()
+	inf.Params["id"] = strconv.Itoa(keys["id"].(int))
```
ocket8888: This'll panic if the object doesn't have an `id`, which sort of defeats the purpose of `APIInfo.IntParams` and `Reader.GetKeys`.

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -80,8 +81,33 @@ func GenericCreate(val GenericCreator) (error, error, int) {
...
+
+	results, usrErr, sysErr, errCode := val.Read()
+	if usrErr != nil || sysErr != nil {
+		return usrErr, sysErr, errCode
+	}
+	if len(results) != 1 {
+		return nil, errors.New("bad result in reading back a POST"), http.StatusInternalServerError
+	}
+	ConvertFrom(&val, results[0])
```
ocket8888: This seems like a lot of work to get at the underlying data - maybe instead find a way to expose that?

moltzaum: What do you mean `expose`? We are given an interface so we don't have many options.

ocket8888: You can always just make them [define a setter for the underlying data](https://play.golang.org/p/02kL8msfpNj)

PR #3583, opened 2019-05-17
Added IDImpl struct

moltzaum:

IDImpl is a struct that implements GetKeys and SetKeys for the common case where the only key is a single numerical ID. The idea is to make a single, common implementation that can be used across all endpoints that can use it. As a bonus these endpoints don't need to make a GetKeys or SetKeys function, nor do they have to provide the json and db tags and we know the behavior we want is implemented.

The downside is that current swagger comments may be broken (although we weren't exactly using it anyway).

This is mostly a test to get feedback. I've had an idea like this for awhile, but didn't implement it because I had doubts. I think it is relatively useful, although it is debatable whether or not this works well as a layer of abstraction since "every abstraction layer has a cost". I was considering other ways to do the same thing in a better way. Also, some might perfer to keep the tc structs 'clean'.

For now, I've only made the CDN struct use the IDImpl. IDImpl only works for Nullable structs (the structs that we define endpoints on).

The main affected component is Traffic Ops, but since I've edited a TC struct, it could affect any component that uses the structs. go is smart enough so that `StructType.EmbeddedType.Field` can be called as `StructType.Field` as long as that field name doesn't exist higher up in the embedded struct hierarchy, so moving the field to an embedded type shouldn't affect much.

Tests already exist for CDNs and should pass.

- [x] Tests already exist for affected pieces of code

### DIFF
--- a/lib/go-tc/cdns.go
+++ b/lib/go-tc/cdns.go
@@ -1,5 +1,7 @@
 package tc
 
+import "github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api/impl"
+
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
@@ -83,7 +85,7 @@ type CDNNullable struct {
 	// ID of the CDN
 	//
 	// required: true
-	ID *int `json:"id" db:"id"`
+	impl.ID
 
 	// LastUpdated
 	//
--- a/traffic_ops/traffic_ops_golang/cdn/cdns.go
+++ b/traffic_ops/traffic_ops_golang/cdn/cdns.go
@@ -58,14 +58,6 @@ func (cdn TOCDN) GetKeyFieldsInfo() []api.KeyFieldInfo {
 	return []api.KeyFieldInfo{{"id", api.GetIntKey}}
 }
 
-//Implementation of the Identifier, Validator interface functions
-func (cdn TOCDN) GetKeys() (map[string]interface{}, bool) {
-	if cdn.ID == nil {
-		return map[string]interface{}{"id": 0}, false
-	}
-	return map[string]interface{}{"id": *cdn.ID}, true
-}
-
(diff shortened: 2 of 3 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3681/
Test FAILed.

moltzaum:
retest this please

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3682/
Test FAILed.

lib/go-tc/cdns.go
```diff
@@ -83,7 +85,7 @@ type CDNNullable struct {
 	// ID of the CDN
 	//
 	// required: true
-	ID *int `json:"id" db:"id"`
+	impl.ID
```
ocket8888: `lib/go-tc/cdns.go:88:2: error: reference to undefined identifier 'impl.ID'`
I think either this should be `impl.ImplID` or you need to rename the struct to `ID`. Either will work, but I'm partial to the latter, myself; `impl.ImplID` feels redundant.

traffic_ops/traffic_ops_golang/cdn/cdns.go
```diff
@@ -58,14 +58,6 @@ func (cdn TOCDN) GetKeyFieldsInfo() []api.KeyFieldInfo {
...
 
-//Implementation of the Identifier, Validator interface functions
-func (cdn TOCDN) GetKeys() (map[string]interface{}, bool) {
-	if cdn.ID == nil {
-		return map[string]interface{}{"id": 0}, false
-	}
-	return map[string]interface{}{"id": *cdn.ID}, true
-}
-
 func (cdn TOCDN) GetAuditName() string {
```
ocket8888: This `func` accesses the `cdn.ID` field, which no longer has the right type, causing compilation issues.

moltzaum:
Closing, because I don't think this will get merged and it was experimental anyway.

PR #3591, opened 2019-05-21
Finally removed all of that trailing whitespace from the license header once and for all

ocket8888:
Just removes trailing whitespace from license headers. Also adds a single newline character at EOF for one file that was missing it.

- Traffic Portal

Make sure that all it does is remove whitespace from license headers. It should still pass the TP tests.


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3701/
Test PASSed.

PR #3595, opened 2019-05-22
Removing restangular

### LINKED ISSUE #3571: TP: remove Restangular dependency
At one point, I added Restangular (https://github.com/mgonto/restangular) to help simplify the TP services (traffic_portal/app/src/common/api), however, many of the services use Restangular and others do not. It might make sense to just remove the dependency on Restangular altogether so the service code can be consistent and one less dependency.

### LINKED ISSUE #4194: TP: Generate ISO view won't resolve
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Clicking on the `Generate ISO` menu item does nothing.

image

## Expected / new behavior:
Clicking on the `Generate ISO` menu item should display the form required to generate an ISO.

1. Click on the `Generate ISO` menu item

ocket8888:
Removes a dependency that Traffic Portal had on Restangular.

- [x] This PR resolves #3571 and fixes #4194 

- Traffic Portal

Traffic Portal dependencies are not individually documented, and so no changes are necessary.

No functionality should have changed (except that errors will now be logged instead of ignored in many cases), so the existing tests should all pass.

Jeremy: actually, the only way to truly verify that the functionality did not change is to:
1. run the UI tests (which is a very small subset of all TP functionality)
2. manually verify the functionality associated with ALL the services functions that changed just to be sure all is well.

### DIFF
--- a/traffic_portal/app/src/common/api/CapabilityService.js
+++ b/traffic_portal/app/src/common/api/CapabilityService.js
@@ -25,7 +25,6 @@ var CapabilityService = function($http, messageModel, ENV) {
 				return result.data.response;
 			},
 			function(err) {
-				console.error(err);
 				throw err;
 			}
 		);
@@ -37,7 +36,6 @@ var CapabilityService = function($http, messageModel, ENV) {
 				return result.data.response[0];
 			},
 			function(err) {
-				console.error(err);
 				throw err;
 			}
 		);
--- a/traffic_portal/app/src/common/api/CapabilityService.js
+++ b/traffic_portal/app/src/common/api/CapabilityService.js
@@ -25,6 +25,7 @@ var CapabilityService = function($http, messageModel, ENV) {
 				return result.data.response;
 			},
 			function(err) {
+				console.error(err);
 				throw err;
 			}
 		);
@@ -36,6 +37,7 @@ var CapabilityService = function($http, messageModel, ENV) {
 				return result.data.response[0];
 			},
 			function(err) {
+				console.error(err);
 				throw err;
 			}
 		);
--- a/traffic_portal/app/src/common/api/ChangeLogService.js
+++ b/traffic_portal/app/src/common/api/ChangeLogService.js
@@ -34,7 +34,7 @@ var ChangeLogService = function($http, $rootScope, ENV) {
 		$rootScope.$broadcast('changeLogService::getChangeLogs');
 		return $http.get(ENV.api['root'] + 'logs', {params: queryParams}).then(
 			function(result) {
-				return result;
+				return result.data.response;
 			},
 			function(err) {
 				throw err;
--- a/traffic_portal/app/src/common/api/ChangeLogService.js
+++ b/traffic_portal/app/src/common/api/ChangeLogService.js
@@ -34,7 +34,7 @@ var ChangeLogService = function($http, $rootScope, ENV) {
 		$rootScope.$broadcast('changeLogService::getChangeLogs');
 		return $http.get(ENV.api['root'] + 'logs', {params: queryParams}).then(
 			function(result) {
-				return result.data.response;
+				return result;
 			},
 			function(err) {
 				throw err;
--- a/traffic_portal/app/src/common/api/ProfileParameterService.js
+++ b/traffic_portal/app/src/common/api/ProfileParameterService.js
@@ -32,8 +32,8 @@ var ProfileParameterService = function($http, messageModel, ENV) {
 			);
 	};
 
-	this.linkProfileParameters = function(profile, params) {
-		return $http.post(ENV.api['root'] + 'profileparameter', { profileId: profile.id, paramIds: params, replace: true }).then(
+	this.linkProfileParameters = function(profileId, params) {
+		return $http.post(ENV.api['root'] + 'profileparameter', { profileId: profileId, paramIds: params, replace: true }).then(
 			function(result) {
 				messageModel.setMessages([ { level: 'success', text: 'Parameters linked to profile' } ], false);
 				return result;
--- a/traffic_portal/app/src/common/api/ProfileParameterService.js
+++ b/traffic_portal/app/src/common/api/ProfileParameterService.js
@@ -35,7 +35,6 @@ var ProfileParameterService = function($http, messageModel, ENV) {
 	this.linkProfileParameters = function(profile, params) {
 		return $http.post(ENV.api['root'] + 'profileparameter', { profileId: profile.id, paramIds: params, replace: true }).then(
 			function(result) {
-				messageModel.setMessages([ { level: 'success', text: 'Parameters linked to profile' } ], false);
 				return result;
 			},
 			function(err) {
--- a/traffic_portal/app/src/common/api/ProfileParameterService.js
+++ b/traffic_portal/app/src/common/api/ProfileParameterService.js
@@ -35,6 +35,7 @@ var ProfileParameterService = function($http, messageModel, ENV) {
 	this.linkProfileParameters = function(profile, params) {
 		return $http.post(ENV.api['root'] + 'profileparameter', { profileId: profile.id, paramIds: params, replace: true }).then(
 			function(result) {
+				messageModel.setMessages([ { level: 'success', text: 'Parameters linked to profile' } ], false);
 				return result;
 			},
 			function(err) {
--- a/traffic_portal/app/src/common/api/ProfileParameterService.js
+++ b/traffic_portal/app/src/common/api/ProfileParameterService.js
@@ -32,8 +32,8 @@ var ProfileParameterService = function($http, messageModel, ENV) {
 			);
 	};
 
-	this.linkProfileParameters = function(profileId, params) {
-		return $http.post(ENV.api['root'] + 'profileparameter', { profileId: profileId, paramIds: params, replace: true }).then(
+	this.linkProfileParameters = function(profile, params) {
+		return $http.post(ENV.api['root'] + 'profileparameter', { profileId: profile.id, paramIds: params, replace: true }).then(
 			function(result) {
 				messageModel.setMessages([ { level: 'success', text: 'Parameters linked to profile' } ], false);
 				return result;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed Traffic Ops Golang POST servers/id/deliveryservice double-logging errors.
 - Issue #4131 - The "Clone Delivery Service Assignments" menu item is hidden on a cache when the cache has zero delivery service assignments to clone.
 - Traffic Portal - Turn off TLSv1
+- Removed Traffic Portal dependency on Restangular
 
 ### Deprecated/Removed
 - The TO API `cachegroup_fallbacks` endpoint is now deprecated
--- a/traffic_portal/app/src/common/api/CacheGroupService.js
+++ b/traffic_portal/app/src/common/api/CacheGroupService.js
@@ -36,6 +36,7 @@ var CacheGroupService = function($http, locationUtils, messageModel, ENV) {
                 return result.data.response[0];
             },
             function (err) {
+                console.error(err);
                 throw err;
             }
         );
@@ -113,6 +114,7 @@ var CacheGroupService = function($http, locationUtils, messageModel, ENV) {
                 return result.data.response;
             },
             function(err) {
+                console.error(err);
                 throw err;
             }
         );
--- a/traffic_portal/app/src/common/api/CacheGroupService.js
+++ b/traffic_portal/app/src/common/api/CacheGroupService.js
@@ -36,7 +36,6 @@ var CacheGroupService = function($http, locationUtils, messageModel, ENV) {
                 return result.data.response[0];
             },
             function (err) {
-                console.error(err);
                 throw err;
             }
         );
@@ -114,7 +113,6 @@ var CacheGroupService = function($http, locationUtils, messageModel, ENV) {
                 return result.data.response;
             },
             function(err) {
-                console.error(err);
                 throw err;
             }
         );
--- a/traffic_portal/app/src/common/api/HttpService.js
+++ b/traffic_portal/app/src/common/api/HttpService.js
@@ -22,7 +22,7 @@ var HttpService = function($http, $q) {
     this.get = function(resource) {
         const deferred = $q.defer();
 
-        return $http.get(resource)
+        $http.get(resource)
             .then(
                 function(result) {
                     deferred.resolve(result);
@@ -38,7 +38,7 @@ var HttpService = function($http, $q) {
     this.post = function(resource, payload) {
         const deferred = $q.defer();
 
-        return $http.post(resource, payload)
+        $http.post(resource, payload)
             .then(
                 function(result) {
                     deferred.resolve(result);
@@ -54,7 +54,7 @@ var HttpService = function($http, $q) {
     this.put = function(resource, payload) {
         const deferred = $q.defer();
 
-        return $http.put(resource, payload)
+        $http.put(resource, payload)
             .then(
                 function(result) {
                     deferred.resolve(result.response);
@@ -70,7 +70,7 @@ var HttpService = function($http, $q) {
     this.delete = function(resource) {
         const deferred = $q.defer();
 
-        return $http.delete(resource)
+        $http.delete(resource)
             .then(
                 function(result) {
                     deferred.resolve(result.response);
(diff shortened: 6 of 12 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3717/
Test PASSed.

traffic_portal/app/src/common/api/ASNService.js
```diff
@@ -17,54 +17,75 @@
...
-                },
-                function(fault) {
-                    messageModel.setMessages(fault.data.alerts, false);
-                }
-            );
+        return $http.post(ENV.api['root'] + 'asns', asn).then(
+            function(result) {
+                messageModel.setMessages([{level: 'success', text: 'ASN created' }], true);
+                console.info("created new ASN: ", result.data.response);
+                return result;
```
mitchell852: this isn't navigating back to the asns table after create. probably because your `return` is before the navigatetoPath

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -21,56 +21,48 @@ var AuthService = function($rootScope, $http, $state, $location, $q, $state, htt
...
                 }
-            );
+            },
+            function(err) {
+               console.error(err);
+            }
+        );
     };
 
     this.tokenLogin = function(token) {
```
mitchell852: i would just leave the tokenLogin function alone unless your sure this doesn't change behavior. seems like it would.

traffic_portal/app/src/common/api/CDNService.js
```diff
@@ -17,252 +17,199 @@
...
+            function(result) {
+                return result.data.response;
+            },
+            function(err) {
+                console.error(err);
+            }
+        );
     };
 
     this.generateDNSSECKeys = function(dnssecKeysRequest) {
```
mitchell852: this doesn't seem to work

https://localhost:8442/#!/cdns/2/dnssec-keys/generate and hit regenerate:

```
shared-libs.js?built=1558556756347:66530 TypeError: Cannot read property 'then' of undefined
    at generate (FormGenerateCdnDnssecKeysController.js:25)
```

traffic_portal/app/src/common/api/CDNService.js
```diff
@@ -17,252 +17,199 @@
...
+            function(result) {
+                return result;
+            },
+            function(err) {
+                messageModel.setMessages(err.data.alerts, false);
+            }
+        );
     };
 
 	this.regenerateKSK = function(kskRequest, cdnKey) {
```
mitchell852: this is not working either. same sort of error:

```
shared-libs.js?built=1558556756347:66530 TypeError: Cannot read property 'then' of undefined
    at generate (FormRegenerateKskController.js:25)
```

traffic_portal/app/src/common/api/CapabilityService.js
```diff
@@ -17,68 +17,69 @@
...
+			function(err) {
+				console.error(err);
+				return err;
+			}
+		);
 	};
 
 	this.getCapability = function(name) {
-		return Restangular.one("capabilities", name).get();
+		return $http.get(ENV.api['root'] + 'capabilities/' + cap).then(
```
mitchell852: this is broken. i think replace cap with name.

traffic_portal/app/src/common/api/CoordinateService.js
```diff
@@ -17,66 +17,60 @@
  * under the License.
  */
 
-var CoordinateService = function($http, $q, Restangular, locationUtils, messageModel, ENV) {
+var CoordinateService = function($http, locationUtils, messageModel, ENV) {
 
     this.getCoordinates = function(queryParams) {
```
mitchell852: not sure what's up here but https://localhost:8442/#!/coordinates looks like

image

traffic_portal/app/src/common/api/AuthService.js
mitchell852: actually,  nm. i think this is fine

traffic_portal/app/src/common/api/CoordinateService.js
ocket8888: I think I fixed that - will check in a second

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3721/
Test PASSed.

traffic_portal/app/src/common/api/CoordinateService.js
ocket8888: yep, fixed.

traffic_portal/app/src/common/api/ChangeLogService.js
```diff
@@ -17,19 +17,25 @@
...
 		return httpService.get(ENV.api['root'] + 'logs/newcount', { ignoreLoadingBar: true });
 	};
 
 	this.getChangeLogs = function(queryParams) {
-		// broadcast an event that will zero out the change log count immediately
 		$rootScope.$broadcast('changeLogService::getChangeLogs');
-		return Restangular.all('logs').getList(queryParams);
+		return $http.get(ENV.api['root'] + 'logs', {params: queryParams}).then(
+			function(result) {
+				return result.data.response;
```
mitchell852: shared-libs.js?built=1558639738692:66530 TypeError: Cannot read property 'data' of undefined
    at ChangeLogModel.js:45

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -21,56 +21,48 @@ var AuthService = function($rootScope, $http, $state, $location, $q, $state, htt
...
-                    $rootScope.$broadcast('trafficPortal::exit');
-                    if ($state.current.name == 'trafficPortal.public.login') {
-                        messageModel.setMessages(result.alerts, false);
-                    } else {
-                        messageModel.setMessages(result.alerts, true);
-                        $state.go('trafficPortal.public.login');
-                    }
-                    return result;
+        httpService.post(ENV.api['root'] + 'user/logout').then(
+            function(result) {
```
mitchell852: keep getting result is undefined on logout

traffic_portal/app/src/common/api/DeliveryServiceService.js
```diff
@@ -17,168 +17,184 @@
...
-            .then(
-                function() {
-                    messageModel.setMessages([ { level: 'success', text: 'Steering target deleted' } ], true);
-                    locationUtils.navigateToPath('/delivery-services/' + dsId + '/targets');
-                },
-                function(fault) {
-                    messageModel.setMessages(fault.data.alerts, true);
-                }
-            );
+        return $http.delete(ENV.api['root'] + 'steering/', dsId + '/targets/' + targetId).then(
```
mitchell852: deleting a ds target is not working

traffic_portal/app/src/common/api/DeliveryServiceService.js
```diff
@@ -17,168 +17,184 @@
...
-            .then(
-                function() {
-                    messageModel.setMessages([ { level: 'success', text: 'Delivery service and server were unlinked.' } ], false);
-                },
-                function(fault) {
-                    messageModel.setMessages(fault.data.alerts, true);
-                }
-            );
+        return httpService.delete(ENV.api['root'] + 'deliveryservice_server/' + dsId + '/' + serverId).then(
+            function(result) {
```
mitchell852: shared-libs.js?built=1558641564542:66530 TypeError: Cannot read property 'data' of undefined
    at DeliveryServiceService.js:163

traffic_portal/app/src/common/api/DeliveryServiceSslKeysService.js
```diff
@@ -17,10 +17,10 @@
  * under the License.
  */
 
-var DeliveryServiceSslKeysService = function($http, $q, locationUtils, messageModel, ENV) {
+var DeliveryServiceSslKeysService = function($http, locationUtils, messageModel, ENV) {
 	this.generateSslKeys = function(deliveryService, sslKeys, generateSslKeyForm) {
 		if (sslKeys.hasOwnProperty('version')){
-			generateSslKeyForm.version = parseInt(sslKeys.version) + 1;
+			generateSslKeyForm.version = parseInt(sslKeys.version, 10) + 1;
```
mitchell852: seems unrelated to restangular

traffic_portal/app/src/common/api/DeliveryServiceUriSigningKeysService.js
```diff
@@ -17,17 +17,31 @@
  * under the License.
  */
 
-var DeliveryServiceUriSigningKeysService = function(Restangular) {
+var DeliveryServiceUriSigningKeysService = function($http, ENV) {
 
 	this.getKeys = function(id) {
-		return Restangular.one("deliveryservices", id).one("urisignkeys").get();
```
mitchell852: TypeError: keys.plain is not a function
    at new DeliveryServiceUriSigningKeysController (DeliveryServiceUriSigningKeysController.js:23)

traffic_portal/app/src/common/api/DeliveryServiceSslKeysService.js
ocket8888: It is, but [using `parseInt` without passing an explicit radix is dangerous](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Octal_interpretations_with_no_radix) so I wanted to fix it asap.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3728/
Test PASSed.

traffic_portal/app/src/common/api/CDNService.js
```diff
@@ -17,252 +17,207 @@
...
-                    } else {
-                        response = _.filter(result.data.response, function(cdn) {
-                            return cdn.name != 'ALL';
-                        });
-                    }
-                    request.resolve(response);
-                },
-                function(fault) {
-                    request.reject();
+        return $http.get(ENV.api['root'] + 'cdns').then(
```
mitchell852: so i'm seeing some weird errors and i think it's because of the way the angular-ui-router's "resolves" work. for example, if you try to go here (when you are unauthorized/logged in) - https://localhost:8442/#!/cdns - the way I built TP with the help of angular-ui-router is it works its way down the resolve chain.

it first tries to resolve these resolves:

https://github.com/apache/trafficcontrol/blob/master/traffic_portal/app/src/modules/private/index.js#L45

specifically, it first tries to fetch "current user" and if that fails, the next resolve should NOT even be attempted:

https://github.com/apache/trafficcontrol/blob/master/traffic_portal/app/src/modules/private/cdns/list/index.js#L29

but with your changes (removing $q.defer()), it's not stopping after the first resolve failure...instead it just keeps going down the resolve chain. maybe that's not a problem...but maybe it is. will need to do some more testing. 

but i do think this format works and it has no restangular (your goal of this PR)

`    this.getCDNs = function(all) {
        var request = $q.defer();
        $http.get(ENV.api['root'] + "cdns")
            .then(
                function(result) {
                    var response;
                    if (all) { // there is a CDN called "ALL" that is not really a CDN but you might want it...
                        response = result.data.response;
                    } else {
                        response = _.filter(result.data.response, function(cdn) {
                            return cdn.name != 'ALL';
                        });
                    }
                    request.resolve(response);
                },
                function(fault) {
                    request.reject(fault);
                }
            );

        return request.promise;
    };
`

mitchell852: UserService.js#getCurrentUser() needs to be implemented like this (or something with similar behavior) so if it fails, the remaining resolves are not attempted at all:

```
    this.getCurrentUser = function() {
        const deferred = $q.defer();
        $http.get(ENV.api['root'] + "user/current")
            .then(
                function(result) {
                    userModel.setUser(result.data.response);
                    deferred.resolve(result.data.response);
                },
                function(fault) {
                    deferred.reject(fault);
                }
            );
        return deferred.promise;
    };
```

traffic_portal/app/src/common/api/UserService.js
```diff
@@ -17,143 +17,148 @@
  * under the License.
  */
 
-var UserService = function(Restangular, $http, $location, $q, authService, httpService, locationUtils, userModel, messageModel, ENV) {
+var UserService = function($http, locationUtils, userModel, messageModel, ENV) {
 
-    var service = this;
+    const service = this;
```
mitchell852: you can get rid of this. it's not used anywhere.

traffic_portal/app/src/common/api/CDNService.js
mitchell852: ^^ you can disregard all of that. see the comment attached to UserService.js

traffic_portal/app/src/common/api/UserService.js
```diff
@@ -17,143 +17,148 @@
...
 
     this.getCurrentUser = function() {
```
mitchell852: getCurrentUser() needs to be implemented like this (or something with similar behavior) so if it fails, the remaining resolves are not attempted at all:

```
    this.getCurrentUser = function() {
        return $http.get(ENV.api['root'] + "user/current").then(
            function(result) {
                userModel.setUser(result.data.response);
                return result.data.response;
            },
            function(err) {
                console.error(err);
                return err;
            }
        );
    };
```

ocket8888: TBH I'd consider asynchronous calls being done synchronously much weirder. But the problem is easily fixed - the interceptor will catch it immediately if you throw the error instead of returning it. Either way the promise is "rejected", but it seems like throwing resolves a call stack faster or something.

Regardless, I've got it no longer making other API calls on a page if it hits a 401. I had to decrease the `$timeout` on the redirect, otherwise the messages wouldn't show up. Not sure what that's about.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3734/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3735/
Test PASSed.

traffic_portal/app/src/app.js
```diff
@@ -464,7 +464,7 @@ trafficPortal.factory('authInterceptor', function ($rootScope, $q, $window, $loc
                         messageModel.setMessages(alerts, true);
                         // forward the to the login page with ?redirect=page/they/were/trying/to/reach
                         $location.url('/login').search({ redirect: encodeURIComponent(url) });
-                    }, 200);
+                    }, 100);
```
mitchell852: this whole timeout thing is weird. maybe it can be removed. i'll test it out. can't remember why i put it in here.

traffic_portal/app/src/common/api/UserService.js
mitchell852: resolves aren't synchronous per se. they are a way to ensure that a state (or a nested state) does not resolve until certain dependencies are fetched. in this case i never want the "private" state (or any state under that) to resolve until the currentUser dependency resolves. you are making a lot of fundamental / architectural changes to this PR that is making me very hesitant  to merge it without a LOT of testing.

mitchell852:
the more I think about it, I think this PR is just way too big / risky. it requires a lot of positive and negative testing to ensure it does not change/break functionality. imo you should tackle one "service" at a time. i.e. remove the use of restangular from cdnservice.js. and move down the list of "services" with a seperate PR until nothing uses restangular and the dependency can be removed altogether. otherwise, the risk is much too high given that TP doesn't have comprehensive regression tests in place.

traffic_portal/app/src/app.js
ocket8888: When I remove it it also doesn't show up messages

ocket8888:
Alright man, you asked for it

I'm gonna leave this open as a draft - I'll rebase once the others are all merged and this will be the one that actually removes restangular entirely.

mitchell852:
@ocket8888 to rebase

traffic_portal/app/src/common/api/ProfileParameterService.js
```diff
@@ -35,6 +35,7 @@ var ProfileParameterService = function($http, messageModel, ENV) {
 	this.linkProfileParameters = function(profile, params) {
 		return $http.post(ENV.api['root'] + 'profileparameter', { profileId: profile.id, paramIds: params, replace: true }).then(
 			function(result) {
+				messageModel.setMessages([ { level: 'success', text: 'Parameters linked to profile' } ], false);
```
mitchell852: i dont think this is supposed to be in here as every call to linkProfileParameters takes care of the messaging.

plus, obviously not restangular related

ocket8888: I actually couldn't remember why I did that. Thanks for catching that.

mitchell852: yep, no worries. remove that and i'll merge this.

PR #3675, opened 2019-06-10
Fix "OURCE" displaying in TP build date string

### LINKED ISSUE #1517: TP Build date field has OURCE in it
image

This dialog is populated from the traffic_portal_release.json file that is created when the traffic_portal rpm is created.

alexluckerman:

Fixes a bug that causes the TP build date string to end in "OURCE" by deleting the seconds part of the date string
- [x] This PR fixes #1517 

- Traffic Portal

Build and start Traffic Portal, open it and click the info icon, and verify that the build date displayed does not contain "OURCE"

### DIFF
--- a/traffic_portal/build/traffic_portal.spec
+++ b/traffic_portal/build/traffic_portal.spec
@@ -61,7 +61,7 @@ tar -xzvf $RPM_SOURCE_DIR/traffic_portal-%{version}.tgz
 
 	# creates dynamic json file needed at runtime for traffic portal to display release info
 	VERSION=%{version}-%{build_number}
-	BUILD_DATE=$(date +'%Y-%m-%d %H:%M:%S')
+	BUILD_DATE=$(date +'%Y-%m-%d %H:%M')
 	VERSION="\"Version\":\"$VERSION\""
 	BUILD_DATE="\"Build Date\":\"$BUILD_DATE\""
 	JSON_VERSION="{\n$VERSION,\n$BUILD_DATE\n}"


asfgit:

traffic_portal/build/traffic_portal.spec
```diff
@@ -61,7 +61,7 @@ tar -xzvf $RPM_SOURCE_DIR/traffic_portal-%{version}.tgz
 
 	# creates dynamic json file needed at runtime for traffic portal to display release info
 	VERSION=%{version}-%{build_number}
-	BUILD_DATE=$(date +'%Y-%m-%d %H:%M:%S')
+	BUILD_DATE=$(date +'%Y-%m-%d %H:%M')
```
ocket8888: Doesn't that just remove the seconds part of the build date?

mitchell852: you are correct @ocket8888 but deleting the seconds also deletes the `OURCE` for some reason. /shrug

not sure the seconds are really needed anyhow. this field is purely for display purposes in TP.

dewrich: Odd, seems like this would have been a problem since the beginning.

mitchell852: oh hi @dewrich - it has been. just never much of a priority.

ocket8888: I bet `%S` is getting expanded as an `rpmbuild` macro

mitchell852: probably something like that

PR #3680, opened 2019-06-14
TP: adds the ability to rearrange table columns as well as toggle column visibility

mitchell852:
This PR adds the ability to show/hide table columns in TP as well as reorder the visible columns of a table. The changes made by the user will be persisted in the browsers local storage.

I tackled an easy table (the types table) to start and prove out the concept.

- Traffic Portal

2. Manually verify
- Launch TP
- navigate to Configure > Types
- reorder columns via drag/drop on the column header
- show/hide columns by using the following drop down

image

Note: I'm not going to document this functionality or add it to the changelog until it is added to other tables in subsequent PRs.


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3860/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3861/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3867/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3868/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3869/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3870/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3871/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3872/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3873/
Test FAILed.

traffic_portal/app/src/common/modules/table/types/TableTypesController.js
```diff
@@ -33,15 +37,43 @@ var TableTypesController = function(types, $scope, $state, locationUtils) {
...
+                    { "name": "description", "visible": true, "searchable": true },
+                    { "name": "lastUpdated", "visible": false, "searchable": false },
+                    { "name": "useInTable", "visible": true, "searchable": true }
+                ],
+                "colReorder": {
+                    realtime: false
+                },
+                "initComplete": function(settings, json) {
+                    // need to bind the show/hide column checkboxes to the saved visibility
+                    $scope.columns = JSON.parse($window.localStorage['DataTables_typesTable_/'])['columns'];
```
ocket8888: The standard [Storage](https://developer.mozilla.org/en-US/docs/Web/API/Storage) interface doesn't appear to support indexing. It does work in Firefox, and I assume you tested it in Chrome, so maybe we don't care, but indexing seems to be non-standard and implementation-dependent. I'd recommend using `Storage.getItem()` instead.

This line also has an uncaught `SyntaxError` from that call to `JSON.parse`, which I guess is fine if it's somehow implicitly handled by the dataTables? Seems unlikely though.

You're also accessing a property of an object by indexing it with a string, which isn't exactly a problem in this case because of how the Storage API works, but it's a bad habit to get into because it forces type coercion on the indexed value.

Finally, you should check for the availability of the localStorage object ([MDN Guide](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Feature-detecting_localStorage)). This is because some browsers don't have localStorage (though I'm sure we don't support them), but also because sometimes it may just not be _available_ even if it is _supported_. As an example, Safari's private browsing mode gives you an empty object with 0 quota on every access. It's possible that dataTablesSort or whatever it's called does this already, but you should be sure; I'm not.

traffic_portal/app/src/common/modules/table/types/table.types.tpl.html
```diff
@@ -34,18 +47,19 @@
             <thead>
                 <tr class="headings">
                     <th>Name</th>
-                    <th>Use In Table</th>
                     <th>Description</th>
+                    <th>Last Updated</th>
+                    <th>Use In Table</th>
```
ocket8888: I hate this name, because it's tying our UI directly to object representation i nthe database, which is meant to be opaque to Traffic Portal. But I don't have a better suggestion, so if you don't either, I guess it's fine.

traffic_portal/app/src/common/modules/table/types/table.types.tpl.html
```diff
@@ -25,6 +25,19 @@
         <div class="pull-right">
             <button class="btn btn-primary" title="Create Type" ng-click="createType()"><i class="fa fa-plus"></i></button>
             <button class="btn btn-default" title="Refresh" ng-click="refresh()"><i class="fa fa-refresh"></i></button>
+            <div class="btn-group" role="group" title="Select Table Columns" uib-dropdown is-open="columnSettings.isopen">
+                <button type="button" class="btn btn-default dropdown-toggle" uib-dropdown-toggle aria-haspopup="true" aria-expanded="false">
+                    <i class="fa fa-columns"></i>&nbsp;
+                    <span class="caret"></span>
+                </button>
+                <ul ng-click="$event.stopPropagation()" class="column-settings dropdown-menu-right dropdown-menu" uib-dropdown-menu>
+                    <li role="menuitem" ng-repeat="c in columns">
```
ocket8888: a `[role="menuitem"] must be the direct child of a `[role="menu"]` or `[role="menubar"]` - you can add these roles to the `<ul>`, or you can get that implicitly by using a [`<menu>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu) instead.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3874/
Test FAILed.

traffic_portal/app/src/common/modules/table/types/table.types.tpl.html
mitchell852: yes, i'm not a fan either. it's a bad practice imo that bubbled up to the UI

traffic_portal/app/src/common/modules/table/types/TableTypesController.js
mitchell852: ok, i'll see what i can clean up here. thanks

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3875/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3876/
Test FAILed.

traffic_portal/app/src/app.js
```diff
@@ -457,7 +457,6 @@ var trafficPortal = angular.module('trafficPortal', [
 
         .run(function($log, applicationService) {
             $log.debug("Application run...");
-            applicationService.startup();
```
ocket8888: wasn't this important?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3878/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3879/
Test FAILed.

traffic_portal/app/src/app.js
mitchell852: no. i thought it might be useful but i never used it. there's an init function in applicationService you can use instead if you need to do something at startup.

mitchell852:
> Dude you gotta [minify](https://duckduckgo.com/?q=online+javascript+minifier&t=ffab&atb=v159-1&ia=answer) that new library - it's more than 10 times bigger than the previous version

I have a few js files i need to minify so i'll get them all in a subsequent PR

mitchell852:
retest this please

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3880/
Test FAILed.

traffic_portal/app/src/app.js
ocket8888: :+1: that sounds like a more proper use of Angular anyway.

traffic_portal/app/src/common/modules/table/types/TableTypesController.js
```diff
@@ -33,15 +37,34 @@ var TableTypesController = function(types, $scope, $state, locationUtils) {
...
+                    { "name": "description", "visible": true, "searchable": true },
+                    { "name": "lastUpdated", "visible": false, "searchable": false },
+                    { "name": "useInTable", "visible": true, "searchable": true }
+                ],
+                "colReorder": {
+                    realtime: false
+                },
+                "initComplete": function(settings, json) {
+                    // need to create the show/hide column checkboxes and bind to the current visibility
+                    $scope.columns = JSON.parse(localStorage.getItem('DataTables_typesTable_/')).columns;
```
ocket8888: You should still handle this error, because an error thrown by `JSON.parse` doesn't necessarily mean that `localStorage` isn't working - the locally stored data could have been corrupted somehow etc. Stack traces from an angular compile cycle are very hard to follow, so even if all you do with the error is log it to the console it could aid in debugging.

mitchell852: handle with a try/catch?

ocket8888: Yeah, a simple

```javascript
try {
    doThing()
} catch e {
    console.error("Failed to load columns from localStorage:", e)
    // optional fallback column assignment here (not sure if that's necessary)
}
```

would be fine IMO. Esp. since that error could propagate upward and mess up the whole UI, not sure what that would look like.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3882/
Test PASSed.

traffic_portal/app/src/common/modules/table/types/TableTypesController.js
mitchell852: actually, failure to parse will just result in an empty dropdown

image

but i like the idea of providing some sort of info as to the problem.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3896/
Test PASSed.

mitchell852:
retest this please

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3897/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3898/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3899/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3901/
Test PASSed.

traffic_portal/app/src/common/modules/table/types/table.types.tpl.html
```diff
@@ -34,18 +47,19 @@
             <thead>
```
ocket8888: So somehow this table is getting an explicit size. I'm not sure if it can be overridden with SCSS, but I think ideally you want `width: 100%;` to fill up that `div.x_content` which I'm guessing is resized by angular/bootstrap. Currently (i.e. in master) this is never an issue because the explicit size appears to fit the available space, and is updated on window resize (to a point). However, if you remove all columns and thus the table does not display, then re-add some columns it seems to get stuck at a much smaller width (519px on my monitor, ymmv).

Idk if you can fix that, it might be obscured within jQuery datatables or inside of the column re-order library, but hopefully adding
```scss
.x_content {
    table {
        width: 100%; // !important ? hopefully not...
    }
}
```
would work.

mitchell852: good catch. worked...but i had to use `!important`

ocket8888: eww. But that's libraries for you.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3907/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3909/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3911/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3984/
Test PASSed.

PR #3681, opened 2019-06-14
Add Python TO client support for updating a user by ID

rawlinp:

Add Python TO client support for updating a user by ID.

- no documentation required

We don't currently have any tests for the Python TO client for specific endpoints.

Manually install the Python TO client and try something like this (passing `TO_URL`, `TO_USER`, and `TO_PASS` as environment variables to this script:
```
#!/usr/bin/env python

from urllib.parse import urlparse

import trafficops

TO_URL = urlparse(os.environ['TO_URL'])
TO_HOST = TO_URL.hostname
TO_PORT = TO_URL.port if TO_URL.port else 443 if TO_URL.scheme == "https" else 80
TO_USE_SSL = TO_URL.scheme == 'https'

TO = trafficops.TOSession(
    host_ip   = TO_HOST,
    host_port = TO_PORT,
    ssl       = TO_USE_SSL)
TO.login(
    os.environ['TO_USER'],
    os.environ['TO_PASS'])

user, _ = TO.get_user_by_id(user_id=1234) # replace with your test user ID
u = user[0]

# you could update something in the user like this:
u.fullName = 'Joe Dirt'

u_res, _ = TO.update_user_by_id(user_id=1234, data=u) # replace with your test user ID
```

### DIFF
--- a/traffic_control/clients/python/trafficops/tosession.py
+++ b/traffic_control/clients/python/trafficops/tosession.py
@@ -2405,7 +2405,7 @@ class TOSession(RestApiSession):
 		:raises: Union[LoginError, OperationError]
 		"""
 
-	@api_request(u'put', u'users/{user_id:d}', (u'1.1', u'1.2', u'1.3',))
+	@api_request(u'put', u'users/{user_id:d}', (u'1.1', u'1.2', u'1.3', u'1.4',))
 	def update_user_by_id(self, user_id=None, data=None):
 		"""
 		Update a user.
--- a/traffic_control/clients/python/trafficops/tosession.py
+++ b/traffic_control/clients/python/trafficops/tosession.py
@@ -2405,6 +2405,17 @@ class TOSession(RestApiSession):
 		:raises: Union[LoginError, OperationError]
 		"""
 
+	@api_request(u'put', u'users/{user_id:d}', (u'1.1', u'1.2', u'1.3',))
+	def update_user_by_id(self, user_id=None, data=None):
+		"""
+		Update a user.
+		:ref:`to-api-users`
+		:param data: The user update data payload.
+		:type data: Dict[str, Any]
+		:rtype: Tuple[Union[Dict[str, Any], List[Dict[str, Any]]], requests.Response]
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3862/
Test PASSed.

traffic_control/clients/python/trafficops/tosession.py
```diff
@@ -2405,6 +2405,17 @@ def create_user(self, data=None):
 		:raises: Union[LoginError, OperationError]
 		"""
 
+	@api_request(u'put', u'users/{user_id:d}', (u'1.1', u'1.2', u'1.3',))
```
ocket8888: you should add a `'1.4'` to this. Or maybe that should be done separately to _all_ the methods? idk, either way there ought to be an issue for adding it to the other methods that support 1.4 (or maybe just reworking it so you only need to specify a minimum version)

rawlinp: done

Although, it appears to still be requesting 1.3. Off-hand, I don't know what the underlying mechanism is for which API version to request -- maybe the latest version common to all methods? Ideally we should only have to specify the lowest minor version that the method supports, but the client would make all API requests at the latest minor version (1.4) where possible.

rawlinp: done

Although, it appears to still be requesting 1.3. Off-hand, I don't know what the underlying mechanism is for which API version to request -- maybe the latest version common to all methods? Ideally we should only have to specify the lowest minor version that the method supports, but the client would make all API requests at the latest minor version (1.4) where possible.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3863/
Test PASSed.

traffic_control/clients/python/trafficops/tosession.py
ocket8888: The api version is set on the TOSession object at init, I think it currently defaults to `1.3`. It can probably be bumped to 1.4 at this point.

Issue #3288, opened 2019-01-31
Fix GoDoc Comments


rob05c:
Our GoDoc is a mess. GoDoc isn't something we can avoid, it exists for all Github projects.

https://godoc.org/github.com/apache/trafficcontrol

We need to clean up our comments, not only to technically conform, but to look good.

This probably isn't one Pull Request, it's probably a bunch of small ones, just cleaning things up and making it look good.

The biggest concerns are the packages intended to be public APIs:
https://godoc.org/github.com/apache/trafficcontrol/lib/go-tc
https://godoc.org/github.com/apache/trafficcontrol/traffic_ops/client

Examples of good docs:
https://golang.org/pkg/time
https://godoc.org/github.com/miekg/dns

This would also be a great first Pull Request for a new committer, cleaning up a few bad GoDoc comments.

zrhoffman:
I can work on this. Would someone please assign #3288 to me?

ocket8888:
You can't be assigned to things, only committers can.

Also, this document may prove helpful: https://blog.golang.org/godoc-documenting-go-code

ocket8888:
well whatdya know. I've always been told it was impossible, but decided to just try it.

mitchell852:
@rob05c @ocket8888 @zrhoffman - are they still a mess? do they still need work?

ocket8888:
I was working on it in #6040 and #6099 - pending merge of #6099 I'd probably keep working on it. Which is to say: "yes, and yes"

mitchell852:
this could probably be broken into sub issues to reduce the effort level.

PR #3682, opened 2019-06-17
Add sort by privilege level to TO and use it in TP

### LINKED ISSUE #3226: TP: roles should be sorted in order of priv level
https://tp.domain.net/#!/roles

roles used to be sorted in descending order based on priv level. this was also true for all the role drop downs (i.e. on the user forms) but it's possible the default sort order of GET /api/roles changed when it was rewritten to Go.

this might be as easy as adding a ?orderby=privLevel to the api call or if that query param is not supported, do the sorting on the UI side.

alexluckerman:
* Add support for sorting roles by priv_level
* Change TP to request roles ordered by priv_level
* Add support to TO Go to order by descending order
* Change TP to request roles in descending order
* **DOES NOT INCLUDE DOCS - requires PR #3740**

- [x] This PR fixes #3226

- Traffic Ops
- Traffic Portal

- Run the TO API tests and ensure they pass
- Build and start TO and TP, open the Roles section of TP, and ensure the roles are listed in descending order by privilege level


asfgit:

traffic_portal/app/src/modules/private/roles/list/index.js
```diff
@@ -28,7 +28,7 @@ module.exports = angular.module('trafficPortal.private.roles.list', [])
 						controller: 'TableRolesController',
 						resolve: {
 							roles: function(roleService) {
-								return roleService.getRoles({ orderby: 'priv_level DESC' });
+								return roleService.getRoles({ orderby: 'privLevel', orderbydown : '' });
```
mitchell852: for some reason a query param with no value feels odd to me (although it might be technically ok). what about a query parameter called `sortOrder` or something with 2 values: asc or desc and if sortOrder is not provided with an orderby, then asc is the default.

alexluckerman:
> So now that this has been done, it'll make any endpoints that use dbhelpers support ordering, right? That'll need to be documented on the affected API endpoints.

I didn't see ordering anywhere in the docs, but would I just add that under parameters to each of the endpoint documentation pages?

ocket8888:
yeah, exactly

mitchell852:
this PR should probably have a changelog.md entry as it introduces a  new api feature.

alexluckerman:
Since documentation for this and #3701 are both covered in #3740 due to weird RST tables things, I removed documentation from this PR

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -59,10 +59,21 @@ func BuildWhereAndOrderBy(parameters map[string]string, queryParamsToSQLCols map
...
+			if sortOrder, exists := parameters["sortOrder"]; exists {
+				log.Debugln("sortOrder: ", sortOrder)
+				if sortOrder == "desc" {
+					orderBy += " DESC"
+				} else if sortOrder != "asc" {
+					log.Debugln("Incorrect name for sortOrder: ", sortOrder)
+				}
+			}
 		} else {
 			log.Debugln("Incorrect name for orderby: ", orderby)
```
mitchell852: I know you didn't write this part but can it say something like "This column is not configured to support orderby: orderby"

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -59,10 +59,21 @@ func BuildWhereAndOrderBy(parameters map[string]string, queryParamsToSQLCols map
...
 			log.Debugln("orderby column ", colInfo)
 			orderBy += " " + colInfo.Column
+
+			// if orderby is specified and valid, also check for sortOrder
+			if sortOrder, exists := parameters["sortOrder"]; exists {
+				log.Debugln("sortOrder: ", sortOrder)
+				if sortOrder == "desc" {
+					orderBy += " DESC"
+				} else if sortOrder != "asc" {
+					log.Debugln("Incorrect name for sortOrder: ", sortOrder)
```
mitchell852: how about something like

```log.Debugln("sortOrder value must be desc or asc. Invalid value provided: ", sortOrder)```

File: traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go [hunk 4]
```diff
@@ -59,10 +59,21 @@ func BuildWhereAndOrderBy(parameters map[string]string, queryParamsToSQLCols map
 		if colInfo, ok := queryParamsToSQLCols[orderby]; ok {
 			log.Debugln("orderby column ", colInfo)
 			orderBy += " " + colInfo.Column
+
+			// if orderby is specified and valid, also check for sortOrder
+			if sortOrder, exists := parameters["sortOrder"]; exists {
+				log.Debugln("sortOrder: ", sortOrder)
```
ocket8888: Is this going to be compatible with #3701 ?

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - In Traffic Portal, provides the ability to clone delivery service assignments from one cache to another cache of the same type. Issue #2963.
 - Traffic Ops now allows each delivery service to have a set of query parameter keys to be retained for consistent hash generation by Traffic Router.
 - In Traffic Portal, delivery service table columns can now be rearranged and their visibility toggled on/off as desired by the user. Hidden table columns are excluded from the table search. These settings are persisted in the browser.
+- Traffic Ops now supports a query parameter to return API responses in descending order
```
mitchell852: probably a good idea to say what that query parameter is?

traffic_ops/client/role.go
```diff
@@ -107,8 +107,28 @@ func (to *Session) GetRoleByID(id int) ([]tc.Role, ReqInf, int, error) {
 
 // GET a Role by the Role name
```
mitchell852: this will break GoDocs. Need to look like this:

// GetRoleByName gets a a role by the role name.

traffic_ops/client/role.go
```diff
@@ -107,8 +107,28 @@ func (to *Session) GetRoleByID(id int) ([]tc.Role, ReqInf, int, error) {
...
+		var data tc.RolesResponse
+		if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+			return data.Response, reqInf, resp.StatusCode, err
+		}
+		return data.Response, reqInf, resp.StatusCode, errClient
+	}
+	return []tc.Role{}, reqInf, 0, errClient
+}
+
+// GET a Role by the Role name
```
mitchell852: incorrect comment

traffic_ops/client/role.go
```diff
@@ -107,8 +107,28 @@ func (to *Session) GetRoleByID(id int) ([]tc.Role, ReqInf, int, error) {
 
 // GET a Role by the Role name
 func (to *Session) GetRoleByName(name string) ([]tc.Role, ReqInf, int, error) {
-	url := fmt.Sprintf("%s?name=%s", API_v13_ROLES, url.QueryEscape(name))
-	resp, remoteAddr, errClient := to.rawRequest(http.MethodGet, url, nil)
+	route := fmt.Sprintf("%s?name=%s", API_v13_ROLES, url.QueryEscape(name))
```
mitchell852: can we change this to 1.4 since we are on Api version 1.4 now?

traffic_ops/client/role.go
```diff
@@ -107,8 +107,28 @@ func (to *Session) GetRoleByID(id int) ([]tc.Role, ReqInf, int, error) {
...
+			return data.Response, reqInf, resp.StatusCode, err
+		}
+		return data.Response, reqInf, resp.StatusCode, errClient
+	}
+	return []tc.Role{}, reqInf, 0, errClient
+}
+
+// GET a Role by the Role name
+func (to *Session) GetRoleByQueryParams(queryParams map[string]string) ([]tc.Role, ReqInf, int, error) {
+	route := fmt.Sprintf("%s?", API_v13_ROLES)
```
mitchell852: change to 1.4

traffic_ops/client/role.go
ocket8888: I don't think that's true. Plus, this is the pattern set by all other `func`s in that file.

mitchell852: maybe i'm wrong i was just going by something @rob05c told me.

traffic_ops/testing/api/v14/roles_test.go
```diff
@@ -105,6 +106,34 @@ func GetTestRoles(t *testing.T) {
...
+	log.Debugln("Status Code: ", status)
+	if err != nil {
+		t.Errorf("cannot GET Role by role: %v - %v\n", err, ascResp)
+	}
+
+	// reverse the descending-sorted response and compare it to the ascending-sorted one
+	for start, end := 0, len(descResp)-1; start < end; start, end = start+1, end-1 {
+		descResp[start], descResp[end] = descResp[end], descResp[start]
+	}
+	if !reflect.DeepEqual(descResp, ascResp) {
```
alexluckerman: I tested this with the sortOrder parameter not implemented and it doesn't fail, so something is definitely wrong but I can't seem to figure out what

traffic_ops/client/role.go
rob05c: @mitchell852 is correct. Symbol (function, variable, package, etc) comments should be a full sentence, starting with the name of the symbol. 

>this is the pattern set by all other funcs in that file.

The comments on the other funcs in the file are broken.

See https://blog.golang.org/godoc-documenting-go-code

traffic_ops/testing/api/v14/roles_test.go
ocket8888: if sortOrder isn't implemented wouldn't descResp be exactly ascResp?

traffic_ops/client/role.go
ocket8888: I don't see anything in there about "starting with the name of the symbol". All it says is

> _"The convention is simple: to document a type, variable, constant, function, or even a package, write a regular comment directly preceding its declaration, with no intervening blank line."_

And that the first sentence appears on the package list. You can see in [the generated documentation](https://godoc.org/github.com/apache/trafficcontrol/traffic_ops/client#Session.GetRoleByID) that those comments don't break anything. They're not great, but I think "broken" is a stretch.

rob05c: >Notice this comment is a complete sentence that begins with the name of the element it describes. This important convention allows us to generate documentation in a variety of formats, from plain text to HTML to UNIX man pages, and makes it read better when tools truncate it for brevity, such as when they extract the first line or sentence. 

They don't break code, no. But they're broken English, and they're broken according to the language-recommended format. GoDoc should be complete sentences, starting with the symbol name. All standard library documentation follows this format, as do all projects with well-formed GoDoc (e.g. https://godoc.org/github.com/miekg/dns).

Would you not object if our readthedocs docs were filled with sentence fragments and missing periods?

traffic_ops/testing/api/v14/roles_test.go
alexluckerman: Yep, but if descResp is then reversed, they shouldn't be equal. And yet the test passes anyway...

traffic_ops/client/role.go
ocket8888: No, it should be a complete sentence, but I didn't see anything about starting with the entity it describes (wasn't looking for 'element').

We don't support third-party GoDoc parsers, and it's clear that what he's doing works - even if not well. It's certainly much better to use a complete sentence, but I think to call something that works "broken" is a bit harsh.

Established practice for things like this - afaik from examples like "use `let`/`const` instead of `var`" or "label-able elements should have labels" - is to file an issue to fix it globally rather than worry about breaking the established pattern. Even if it is absolutely better within the scope of this code to do so.

traffic_ops/testing/api/v14/roles_test.go
ocket8888: Oh, I misunderstood.

traffic_ops/client/role.go
rob05c: It's not only third party parsers. There are many first-party parsers, not just HTML. Including the command-line `go doc`, which it categorically breaks.

>Established practice for things like this - afaik from examples like "use let/const instead of var" or "label-able elements should have labels" - is to file an issue to fix it globally rather than worry about breaking the established pattern.

This isn't just "not the ideal best practice," it categorically breaks GoDoc parsing. We should not merge code into the repository that breaks GoDoc. "Following the surrounding convention" should be followed when it's a "best practice" issue, not when something is categorically broken. This is broken. It breaks GoDoc parsers, including first-party language parsers. The fact that we already have broken documentation does not make it acceptable to add more broken documentation. Certainly not considering it's trivial to fix.

>I think to call something that works "broken" is a bit harsh.

It is broken. It does not work. 

Notwithstanding, it would be broken even if `go doc` didn't break. Just because something coincidentally produces the right output does not mean it isn't "broken". If I write a function to compute powers, which actually squares, but I only ever call it with 2, the function is still broken.

https://golang.org/doc/effective_go.html#commentary

```
 Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared.

// Compile parses a regular expression and returns, if successful,
// a Regexp that can be used to match against text.
func Compile(str string) (*Regexp, error) {

If every doc comment begins with the name of the item it describes, you can use the doc subcommand of the go tool and run the output through grep. Imagine you couldn't remember the name "Compile" but were looking for the parsing function for regular expressions, so you ran the command,

$ go doc -all regexp | grep -i parse

If all the doc comments in the package began, "This function...", grep wouldn't help you remember the name. But because the package starts each doc comment with the name, you'd see something like this, which recalls the word you're looking for.

$ go doc -all regexp | grep -i parse
    Compile parses a regular expression and returns, if successful, a Regexp
    MustCompile is like Compile but panics if the expression cannot be parsed.
    parsed. It simplifies safe initialization of global variables holding
$
```

Running their example on our code:

```
2019-08-06 15:31:11 pc ~/lang/go/src/github.com/apache/trafficcontrol/traffic_ops/client
$ cat role.go | grep -B 1 DeleteRoleByID
// DELETE a Role by ID
func (to *Session) DeleteRoleByID(id int) (tc.Alerts, ReqInf, int, error) {
2019-08-06 15:31:13 pc ~/lang/go/src/github.com/apache/trafficcontrol/traffic_ops/client
$ go doc | grep -i delete
2019-08-06 15:31:15 pc ~/lang/go/src/github.com/apache/trafficcontrol/traffic_ops/client
$ 
```

ocket8888: but...
```shell
$ go doc client.DeleteRoleByID
func (to *Session) DeleteRoleByID(id int) (tc.Alerts, ReqInf, int, error)
    DELETE a Role by ID
```
Still though, the only part of GoDoc that factors into our documentation is the html parser, and this doesn't break that.

mitchell852: @alexluckerman - can you please change to 

// GetRoleByName gets a a role by the role name.

traffic_ops/testing/api/v14/roles_test.go
alexluckerman: Fixed it. Shouldn't have had the `name` in the query parameters, which was limiting the server response size to one

traffic_ops/client/role.go
mitchell852: confused on this whole thread. seems pretty clear to me:

https://blog.golang.org/godoc-documenting-go-code

image

mitchell852: actually, looks like there's already one open - #3288

PR #3686, opened 2019-06-18
Update cdn-in-a-box Documentation with more pitfalls

dsouza93:

Updates the cdn-in-a-box documentation to include some pitfalls I ran into when starting it up. They include having to snap the cdn to get an initial CRconfig and clearing the volumes when the self signed certs are failing.

- CDN in a Box
- Documentation

Does the new doc make sense?


asfgit:

PR #3688, opened 2019-06-19
Add TO Go remap.config

rob05c:

Based on https://github.com/apache/trafficcontrol/pull/3762 - highly recommend merging that first.

Converts remap.config to Traffic Ops Golang.

As with the other ATS configs, the code isn't ideal in some places, because it transliterates the Perl as closely as possible, for safety. I'm hoping we can convert these endpoints in two phases, transliterating, then refactoring, to be as safe as possible.

I've parity tested, against the old Perl for every production edge and mid we have, and they all diff identical (after sorting).

No documentation change, doesn't change the interface.

- [x] This PR is not related to any other issue

- Traffic Ops

Request the remap.config endpoint for a server, request it from an older TO, make sure they're identical after sorting (for ATS, map lines can be in any order, and on each line, plugin directives can be in any order as long as the associated pparam directives following them are the same).

Not a bug fix.

### DIFF
--- a/lib/go-atscfg/remapdotconfig.go
+++ b/lib/go-atscfg/remapdotconfig.go
@@ -152,7 +152,7 @@ func GetServerConfigRemapDotConfigForMid(
 	for originFQDN, midRemap := range midRemaps {
 		textLines = append(textLines, "map "+originFQDN+" "+originFQDN+midRemap+"\n")
 	}
-	sort.Sort(sort.StringSlice(textLines))
+	sort.Strings(textLines)
 
 	text := header
 	for _, line := range textLines {
--- a/lib/go-atscfg/remapdotconfig.go
+++ b/lib/go-atscfg/remapdotconfig.go
@@ -1,0 +1,369 @@
+package atscfg
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import (
+	"errors"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+)
+
+type RemapConfigDSData struct {
+	ID                       int
+	Type                     tc.DSType
+	OriginFQDN               *string
+	MidHeaderRewrite         *string
+	CacheURL                 *string
+	RangeRequestHandling     *int
+	CacheKeyConfigParams     map[string]string
+	RemapText                *string
+	EdgeHeaderRewrite        *string
+	SigningAlgorithm         *string
+	Name                     string
+	QStringIgnore            *int
+	RegexRemap               *string
+	FQPacingRate             *int
+	DSCP                     int
+	RoutingName              *string
+	MultiSiteOrigin          *string
+	Pattern                  *string
+	RegexType                *string
+	Domain                   *string
+	RegexSetNumber           *string
+	OriginShield             *string
+	ProfileID                *int
+	Protocol                 *int
+	AnonymousBlockingEnabled *bool
+	Active                   bool
+}
+
+func MakeRemapDotConfig(
+	serverName tc.CacheName,
+	toToolName string, // tm.toolname global parameter (TODO: cache itself?)
+	toURL string, // tm.url global parameter (TODO: cache itself?)
+	atsMajorVersion int,
+	cacheURLConfigParams map[string]string, // map[name]value for this server's profile and config file 'cacheurl.config'
+	dsProfilesCacheKeyConfigParams map[int]map[string]string, // map[profileID][paramName]paramVal for this server's profile, config file 'cachekey.config'
+	serverPackageParamData map[string]string, // map[paramName]paramVal for this server, config file 'package'
+	serverInfo *ServerInfo, // ServerInfo for this server
+	remapDSData []RemapConfigDSData,
+) string {
+	hdr := GenericHeaderComment(string(serverName), toToolName, toURL)
+	text := ""
+	if tc.CacheTypeFromString(serverInfo.Type) == tc.CacheTypeMid {
+		text = GetServerConfigRemapDotConfigForMid(atsMajorVersion, dsProfilesCacheKeyConfigParams, serverInfo, remapDSData, hdr)
+	} else {
+		text = GetServerConfigRemapDotConfigForEdge(cacheURLConfigParams, dsProfilesCacheKeyConfigParams, serverPackageParamData, serverInfo, remapDSData, atsMajorVersion, hdr)
+	}
+	return text
+}
+
+func GetServerConfigRemapDotConfigForMid(
+	atsMajorVersion int,
+	profilesCacheKeyConfigParams map[int]map[string]string,
+	server *ServerInfo,
+	dses []RemapConfigDSData,
+	header string,
+) string {
+	midRemaps := map[string]string{}
+	for _, ds := range dses {
+		if ds.Type.IsLive() && !ds.Type.IsNational() {
+			continue // Live local delivery services skip mids
+		}
+
+		if ds.OriginFQDN == nil || *ds.OriginFQDN == "" {
+			log.Warnf("GetServerConfigRemapDotConfigForMid ds '" + ds.Name + "' has no origin fqdn, skipping!") // TODO confirm - Perl uses without checking!
+			continue
+		}
+
+		if midRemaps[*ds.OriginFQDN] != "" {
+			continue // skip remap rules from extra HOST_REGEXP entries
+		}
+
+		midRemap := ""
+		if ds.MidHeaderRewrite != nil && *ds.MidHeaderRewrite != "" {
+			midRemap += ` @plugin=header_rewrite.so @pparam=` + MidHeaderRewriteConfigFileName(ds.Name)
+		}
+		if ds.QStringIgnore != nil && *ds.QStringIgnore == tc.QueryStringIgnoreIgnoreInCacheKeyAndPassUp {
+			midRemap += GetQStringIgnoreRemap(atsMajorVersion)
+		}
+		if ds.CacheURL != nil && *ds.CacheURL != "" {
+			midRemap += ` @plugin=cacheurl.so @pparam=` + CacheURLConfigFileName(ds.Name)
+		}
+
+		if ds.ProfileID != nil && len(profilesCacheKeyConfigParams[*ds.ProfileID]) > 0 {
+			midRemap += ` @plugin=cachekey.so`
+			for name, val := range profilesCacheKeyConfigParams[*ds.ProfileID] {
+				midRemap += ` @pparam=--` + name + "=" + val
+			}
+		}
+		if ds.RangeRequestHandling != nil && *ds.RangeRequestHandling == tc.RangeRequestHandlingCacheRangeRequest {
+			midRemap += ` @plugin=cache_range_requests.so`
+		}
+
+		if midRemap != "" {
+			midRemaps[*ds.OriginFQDN] = midRemap
+		}
+	}
+
+	textLines := []string{}
+	for originFQDN, midRemap := range midRemaps {
+		textLines = append(textLines, "map "+originFQDN+" "+originFQDN+midRemap+"\n")
+	}
+	sort.Sort(sort.StringSlice(textLines))
+
+	text := header
+	for _, line := range textLines {
+		text += line
+	}
+	return text
+}
+
+func GetServerConfigRemapDotConfigForEdge(
+	cacheURLConfigParams map[string]string,
+	profilesCacheKeyConfigParams map[int]map[string]string,
+	serverPackageParamData map[string]string, // map[paramName]paramVal for this server, config file 'package'
+	server *ServerInfo,
+	dses []RemapConfigDSData,
+	atsMajorVersion int,
+	header string,
+) string {
+	for name, val := range serverPackageParamData {
+		serverPackageParamData[name] = strings.Replace(val, "__HOSTNAME__", server.HostName+"."+server.DomainName, -1)
+	}
+
+	textLines := []string{}
+
+	for _, ds := range dses {
+		remapText := ""
+		if ds.Type == tc.DSTypeAnyMap {
+			if ds.RemapText == nil {
+				log.Errorln("ds '" + ds.Name + "' is ANY_MAP, but has no remap text - skipping")
+				continue
+			}
+			remapText = *ds.RemapText + "\n"
+			textLines = append(textLines, remapText)
+			continue
+		}
+
+		remapLines, err := MakeEdgeDSDataRemapLines(ds, server)
+		if err != nil {
+			log.Errorln("making remap lines for DS '" + ds.Name + "' - skipping! : " + err.Error())
+			continue
+		}
+
+		for _, line := range remapLines {
+			profilecacheKeyConfigParams := (map[string]string)(nil)
+			if ds.ProfileID != nil {
+				profilecacheKeyConfigParams = profilesCacheKeyConfigParams[*ds.ProfileID]
+			}
+			remapText = BuildRemapLine(cacheURLConfigParams, atsMajorVersion, server, serverPackageParamData, remapText, ds, line.From, line.To, profilecacheKeyConfigParams)
+		}
+		textLines = append(textLines, remapText)
+	}
+
+	text := header
+	sort.Sort(sort.StringSlice(textLines))
+	for _, line := range textLines {
+		text += line
+	}
+	return text
+}
+
+// BuildRemapLine builds the remap line for the given server and delivery service.
+// The cacheKeyConfigParams map may be nil, if this ds profile had no cache key config params.
+func BuildRemapLine(cacheURLConfigParams map[string]string, atsMajorVersion int, server *ServerInfo, pData map[string]string, text string, ds RemapConfigDSData, mapFrom string, mapTo string, cacheKeyConfigParams map[string]string) string {
+	// ds = 'remap' in perl
+
+	mapFrom = strings.Replace(mapFrom, `__http__`, server.HostName, -1)
+
+	if _, hasDSCPRemap := pData["dscp_remap"]; hasDSCPRemap {
+		text += "map	" + mapFrom + "     " + mapTo + ` @plugin=dscp_remap.so @pparam=` + strconv.Itoa(ds.DSCP)
+	} else {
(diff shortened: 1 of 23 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3877/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3924/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3925/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3935/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3936/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3937/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3938/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3986/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4105/
Test PASSed.

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,369 @@
...
+		midRemap := ""
+		if ds.MidHeaderRewrite != nil && *ds.MidHeaderRewrite != "" {
+			midRemap += ` @plugin=header_rewrite.so @pparam=` + MidHeaderRewriteConfigFileName(ds.Name)
+		}
+		if ds.QStringIgnore != nil && *ds.QStringIgnore == tc.QueryStringIgnoreIgnoreInCacheKeyAndPassUp {
+			midRemap += GetQStringIgnoreRemap(atsMajorVersion)
+		}
+		if ds.CacheURL != nil && *ds.CacheURL != "" {
+			midRemap += ` @plugin=cacheurl.so @pparam=` + CacheURLConfigFileName(ds.Name)
+		}
```
alficles: Right now, multiple uses of `cacheurl` do not do what one might expect. It's probably worth actually enhancing this to make these duplications findable. We should probably fix `cacheurl` as well, but it's basically never right to have two of these on a line at the same time.

It looks like we handle it for the edge case, but it is just as problematic at the mid.

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+
+	if ds.QStringIgnore != nil {
+		if *ds.QStringIgnore == tc.QueryStringIgnoreDropAtEdge {
+			dqsFile := "drop_qstring.config"
+			text += ` @plugin=regex_remap.so @pparam=` + dqsFile
+		} else if *ds.QStringIgnore == tc.QueryStringIgnoreIgnoreInCacheKeyAndPassUp {
+			if _, globalExists := cacheURLConfigParams["location"]; globalExists {
+				log.Debugln("qstring_ignore == 1, but global cacheurl.config param exists, so skipping remap rename config_file=cacheurl.config parameter") // TODO warn? Perl was a debug
```
alficles: Is debug the right place to log this? This is a pretty serious configuration snafu. The DS is configured to ignore the query string, but this is ignoring the request to ignore it. I worry that a debug log will wind up ignored.

lib/go-atscfg/remapdotconfig.go
rob05c: Ok, I added error logging any time the config gen adds multiple cachekey or cacheurl entries on the same line.

lib/go-atscfg/remapdotconfig.go
rob05c: Changed to a Warning.

traffic_ops/traffic_ops_golang/ats/db.go
```diff
@@ -197,3 +346,428 @@ WHERE
...
+		}
+		return "", false, errors.New("querying: " + err.Error())
+	}
+	return name, true, nil
+}
+
+// getServerNameFromNameOrID returns the server name from a parameter which may be the name or ID.
+// This also checks and verifies the existence of the given server, and returns an appropriate user error if it doesn't exist.
+// Returns the name, any user error, any system error, and any error code.
+func GetServerNameFromNameOrID(tx *sql.Tx, serverNameOrID string) (tc.CacheName, error, error, int) {
```
alficles: Capitalization of the name in the comment should match the name of the function.

alficles: Not commenting everywhere, but there are a bunch of mismatches.

traffic_ops/traffic_ops_golang/ats/db.go
```diff
@@ -197,3 +346,428 @@ WHERE
...
+func GetATSMajorVersion(tx *sql.Tx, serverProfileID atscfg.ProfileID) (int, error) {
+	atsVersion, _, err := GetProfileParamValue(tx, int(serverProfileID), "package", "trafficserver")
+	if err != nil {
+		return 0, errors.New("getting profile param value: " + err.Error())
+	}
+	if len(atsVersion) == 0 {
+		atsVersion = atscfg.DefaultATSVersion
+		log.Warnln("Parameter package.trafficserver missing for profile " + strconv.Itoa(int(serverProfileID)) + ". Assuming version " + atsVersion)
+	}
+	atsMajorVer, err := strconv.Atoi(atsVersion[:1])
```
alficles: I was hoping this would be fixed somewhere. Is here the right place to fix this? This is going to break on ATS 10.

traffic_ops/traffic_ops_golang/ats/db.go
rob05c: Yeah, funcs I renamed and didn't fix the comment. I'll fix.

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,413 @@
...
+		if midRemap != "" {
+			midRemaps[*ds.OriginFQDN] = midRemap
+		}
+	}
+
+	textLines := []string{}
+	for originFQDN, midRemap := range midRemaps {
+		textLines = append(textLines, "map "+originFQDN+" "+originFQDN+midRemap+"\n")
+	}
+	sort.Sort(sort.StringSlice(textLines))
```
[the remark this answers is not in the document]
rob05c: Done.

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,413 @@
...
+	remapLines := []RemapLine{}
+	hostRegex := *ds.Pattern
+	mapTo := *ds.OriginFQDN + "/"
+
+	mapFromHTTP := "http://" + hostRegex + "/"
+	mapFromHTTPS := "https://" + hostRegex + "/"
+	if strings.HasSuffix(hostRegex, `.*`) {
+		re := hostRegex
+		re = strings.Replace(re, `\`, ``, -1)
+		re = strings.Replace(re, `.*`, ``, -1)
```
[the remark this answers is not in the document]
rob05c: Yes, Perl did the exact same thing: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm#L700

```
			if ( $host_re =~ /\.\*$/ ) {
				my $re = $host_re;
				$re =~ s/\\//g;
				$re =~ s/\.\*//g;
```
IMO we should fix in a subsequent release, and make sure the transliteration of the existing code behaves as before first, for safety.

traffic_ops/traffic_ops_golang/ats/db.go
rob05c: Fixed.

traffic_ops/traffic_ops_golang/ats/db.go
rob05c: Done. Put it in a func, including unit tests verifying correctness.

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,413 @@
...
+		if ds.ProfileID != nil && len(profilesCacheKeyConfigParams[*ds.ProfileID]) > 0 {
+			if hasCacheKey {
+				log.Errorln("DeliveryService qstring_ignore and cachekey params both add cachekey, but ATS cachekey doesn't work correctly with multiple entries! Adding anyway!")
+			}
+			midRemap += ` @plugin=cachekey.so`
+			for name, val := range profilesCacheKeyConfigParams[*ds.ProfileID] {
+				midRemap += ` @pparam=--` + name + "=" + val
+			}
+		}
+		if ds.RangeRequestHandling != nil && *ds.RangeRequestHandling == tc.RangeRequestHandlingCacheRangeRequest {
```
ocket8888: What about background fetch?

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,413 @@
...
+	}
+	return text
+}
+
+// BuildRemapLine builds the remap line for the given server and delivery service.
+// The cacheKeyConfigParams map may be nil, if this ds profile had no cache key config params.
+func BuildRemapLine(cacheURLConfigParams map[string]string, atsMajorVersion int, server *ServerInfo, pData map[string]string, text string, ds RemapConfigDSData, mapFrom string, mapTo string, cacheKeyConfigParams map[string]string) string {
+	// ds = 'remap' in perl
+
+	mapFrom = strings.Replace(mapFrom, `__http__`, server.HostName, -1)
```
ocket8888: `__http__` resolves to the hostname? If this isn't a typo, it's undocumented.

lib/go-util/num.go
```diff
@@ -106,3 +109,17 @@ func BytesLenSplit(s []byte, n int) [][]byte {
 	}
 	return ss
 }
+
+func HashInts(ints []int) string {
```
ocket8888: GoDoc?

traffic_ops/ort/atstccfg/remapdotconfig.go
```diff
@@ -0,0 +1,372 @@
...
+		if ds.ID == nil {
+			// TODO log error?
+			continue
+		}
+		dsIDs = append(dsIDs, *ds.ID)
+	}
+
+	isMid := strings.HasPrefix(server.Type, string(tc.CacheTypeMid))
+
+	serverIDs := ([]int)(nil)
```
ocket8888: Why not just `var serverIDs []int`? A lot easier to read IMO

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,413 @@
...
+		}
+
+		remapLines, err := MakeEdgeDSDataRemapLines(ds, server)
+		if err != nil {
+			log.Errorln("making remap lines for DS '" + ds.Name + "' - skipping! : " + err.Error())
+			continue
+		}
+
+		for _, line := range remapLines {
+			profilecacheKeyConfigParams := (map[string]string)(nil)
```
ocket8888: Why not just `var profilecacheKeyConfigParams map[string]string`? A lot easier to read IMO.

traffic_ops/ort/atstccfg/remapdotconfig.go
```diff
@@ -0,0 +1,372 @@
...
+			dssMap[*dss.DeliveryService] = map[int]struct{}{}
+		}
+		dssMap[*dss.DeliveryService][*dss.Server] = struct{}{}
+	}
+
+	useInactive := false
+	if !isMid {
+		// mids get inactive DSes, edges don't. This is how it's always behaved, not necessarily how it should.
+		useInactive = true
+	}
```
ocket8888: I think this could all just be one line: `useInactive := !isMid`

traffic_ops/ort/atstccfg/remapdotconfig.go
```diff
@@ -0,0 +1,372 @@
...
+
+		paramName := param.Name
+		// some files have multiple lines with the same key... handle that with param id.
+		if _, ok := serverPackageParamData[param.Name]; ok {
+			paramName += "__" + strconv.Itoa(param.ID)
+		}
+		paramValue := param.Value
+		if paramValue == "STRING __HOSTNAME__" {
+			paramValue = server.HostName + "." + server.DomainName // TODO strings.Replace to replace all anywhere, instead of just an exact match?
+		}
```
ocket8888: Are you sure this shouldn't be just checking for `__HOSTNAME__`? Does _all_ of `STRING __HOSTNAME__` get turned into the server's hostname? If that's true, it's undocumented.

traffic_ops/traffic_ops_golang/ats/atsserver/parentdotconfig.go
```diff
@@ -1,4 +1,4 @@
-package ats
+package atsserver
```
ocket8888: Apache Traffic Server... server?

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,413 @@
...
+	textLines := []string{}
+	for originFQDN, midRemap := range midRemaps {
+		textLines = append(textLines, "map "+originFQDN+" "+originFQDN+midRemap+"\n")
+	}
+	sort.Strings(textLines)
+
+	text := header
+	for _, line := range textLines {
+		text += line
+	}
```
ocket8888: I think the standard way to do this is to use `strings.Join`.

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,413 @@
...
+		}
+		textLines = append(textLines, remapText)
+	}
+
+	text := header
+	sort.Sort(sort.StringSlice(textLines))
+	for _, line := range textLines {
+		text += line
+	}
+	return text
```
ocket8888: same as above RE `strings.Join`

lib/go-atscfg/remapdotconfig.go
```diff
@@ -0,0 +1,413 @@
...
+	remapDSData []RemapConfigDSData,
+) string {
+	hdr := GenericHeaderComment(string(serverName), toToolName, toURL)
+	text := ""
+	if tc.CacheTypeFromString(serverInfo.Type) == tc.CacheTypeMid {
+		text = GetServerConfigRemapDotConfigForMid(atsMajorVersion, dsProfilesCacheKeyConfigParams, serverInfo, remapDSData, hdr)
+	} else {
+		text = GetServerConfigRemapDotConfigForEdge(cacheURLConfigParams, dsProfilesCacheKeyConfigParams, serverPackageParamData, serverInfo, remapDSData, atsMajorVersion, hdr)
+	}
+	return text
```
ocket8888: Does this spit out trailing newlines? I don't think it does.

ocket8888: actually, after looking through the rest of it, I think it might.

traffic_ops/ort/atstccfg/remapdotconfig.go
rob05c: I don't agree. I prefer to always use `:=` for two reasons. 

For one, `var` doesn't need to exist, you can do everything with `:=`. IMO it's better to have one way to do things. Any time there are multiple ways to do the same thing, it's that much more to learn, to deal with, to understand the exact behavior of, and easy for especially a new programmer to misunderstand the exact behavior of one of the many ways, and write a bug.

Second, it's explicit. Explicit is better than Implicit. The `:=` syntax makes it abundantly clear that `serverIDs` is `nil` until it's set to something else. Where the implicit zero-initialization isn't always obvious at first glance. It's easy to skim over `var serverIDs []int` and not mentally register that it's `nil`. It's much harder to see the word `nil` and not immediately realize that it's nil, and hopefully catch bugs like "oh, I see `myPtr := (*int)(nil)` is dereferenced on the next line, that's going to panic!

lib/go-atscfg/remapdotconfig.go
rob05c: This is transliterating Perl: https://github.com/apache/trafficcontrol/blob/56ef95e1783360b4afb600b06cf138255fb34c89/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm#L2864

lib/go-util/num.go
rob05c: Done.

traffic_ops/ort/atstccfg/remapdotconfig.go
rob05c: It could. But I'd prefer to leave it within an if-statement for now, because as the preceding comment says, I'm not convinced it should behave this way, and an if-block is easier to modify to something more complex in the future, or to remove outright.

traffic_ops/ort/atstccfg/remapdotconfig.go
rob05c: This is transliterating existing Perl: https://github.com/apache/trafficcontrol/blob/56ef95e1783360b4afb600b06cf138255fb34c89/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm#L465

traffic_ops/traffic_ops_golang/ats/atsserver/parentdotconfig.go
rob05c: Apache Traffic Server, server configs. As opposed to Apache Traffic Server, profile configs.

lib/go-atscfg/remapdotconfig.go
rob05c: Changed.

lib/go-atscfg/remapdotconfig.go
rob05c: Changed.

lib/go-atscfg/remapdotconfig.go
rob05c: Yes.

lib/go-atscfg/remapdotconfig.go
ocket8888: *sigh* - well then it'll need to get documented. I'll open an issue.

traffic_ops/ort/atstccfg/remapdotconfig.go
ocket8888: It actually looks like this would never matter, since the only package it's interested in is something to do with dscp on which it does an exact string match. Bizarre, but as I understand it you're reluctant to make logic changes during translation.

traffic_ops/traffic_ops_golang/ats/atsserver/parentdotconfig.go
ocket8888: ah

traffic_ops/ort/atstccfg/remapdotconfig.go
ocket8888: If you say so. It'll likely be you doing that, anyway.

traffic_ops/ort/atstccfg/remapdotconfig.go
ocket8888: I still like `var`, but it's a stylistic change so there's no point arguing about it.

traffic_ops/traffic_ops_golang/ats/server.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	defer inf.Close()
+
+	serverName, userErr, sysErr, errCode := getServerNameFromNameOrID(inf.Tx.Tx, inf.Params["id-or-host"])
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
+		return
+	}
+
+	serverInfo, ok, err := getServerInfoByHost(inf.Tx.Tx, serverName)
```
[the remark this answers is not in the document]
ocket8888: This comment was pending when I submitted my review, I no longer remember what it means - or even making it - so feel free to ignore it.

traffic_ops/ort/atstccfg/remapdotconfig.go
rob05c: Yeah, we need to refactor all the config generation, they're all an overly-complicated mess, and I'm not sure half the logic is right.

Just hoping to do that in a separate PR, for safety.

PR #3689, opened 2019-06-20
Fix TO update_status to exclude ADMIN_DOWN

### LINKED ISSUE #3687: Ignoring ADMIN_DOWN/PRE_PROD Cache Servers while setting parent_pending flag
**Test setup:**
Edge Group - one edge(EDGE-1) and its parent group has two Mids (MID-1 and MID-2)
Assign this EDGE-1 to a delivery service DS-1.

**Test Case:** 
ADMIN-DOWN (or PRE_PROD) one of the MIDs (MID-1); Execute ORT in EDGE-1
ORT run fails complaining that its parent is pending.
 
**Reason:**
During ORT run, EDGE-1 uses "/api/1.3/servers/<host-name>/update_status" to deduce whether its parent has executed ORT successfully (upd_pending -> false).This API (traffic_ops_golang's getServerUpdateStatus function in servers_update_status.go) uses a DB query which ignores only the OFFLINE Mids and not the ADMIN_DOWN Mids and hence ORT run in Edge fails saying 
 
Fix:
Rewrite the db query to ignore ADMIN_DOWN MIDS.

rob05c:
Specifically, changed to only include ONLINE and REPORTED, and thus
also exclude PREPROD and any other similar future states.

Fixes #3687

Still WIP, haven't tested.

- CDN in a Box
- Documentation
- Grove
- Traffic Control Client 
- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Traffic Portal
- Traffic Router
- Traffic Stats
- Traffic Vault

### DIFF
--- a/traffic_ops/traffic_ops_golang/server/servers_update_status.go
+++ b/traffic_ops/traffic_ops_golang/server/servers_update_status.go
@@ -49,7 +49,7 @@ func getServerUpdateStatus(tx *sql.Tx, cfg *config.Config, hostName string) ([]t
 	baseSelectStatement :=
 		`WITH parentservers AS (SELECT ps.id, ps.cachegroup, ps.cdn_id, ps.upd_pending, ps.reval_pending FROM server ps
          LEFT JOIN status AS pstatus ON pstatus.id = ps.status
-         WHERE pstatus.name != 'OFFLINE' ),
+         WHERE (pstatus.name = '` + string(tc.CacheStatusOnline) + `' OR pstatus.name = '` + string(tc.CacheStatusReported) + `') ),
          use_reval_pending AS (SELECT value::boolean FROM parameter WHERE name = 'use_reval_pending' AND config_file = 'global' UNION ALL SELECT FALSE FETCH FIRST 1 ROW ONLY)
          SELECT s.id, s.host_name, type.name AS type, (s.reval_pending::boolean) as server_reval_pending, use_reval_pending.value, s.upd_pending, status.name AS status, COALESCE(bool_or(ps.upd_pending), FALSE) AS parent_upd_pending, COALESCE(bool_or(ps.reval_pending), FALSE) AS parent_reval_pending FROM use_reval_pending, server s
          LEFT JOIN status ON s.status = status.id


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3883/
Test PASSed.

traffic_ops/traffic_ops_golang/server/servers_update_status.go
```diff
@@ -49,7 +49,7 @@ func getServerUpdateStatus(tx *sql.Tx, cfg *config.Config, hostName string) ([]t
 	baseSelectStatement :=
 		`WITH parentservers AS (SELECT ps.id, ps.cachegroup, ps.cdn_id, ps.upd_pending, ps.reval_pending FROM server ps
          LEFT JOIN status AS pstatus ON pstatus.id = ps.status
-         WHERE pstatus.name != 'OFFLINE' ),
+         WHERE (pstatus.name = '` + string(tc.CacheStatusOnline) + `' OR pstatus.name = '` + string(tc.CacheStatusReported) + `') ),
```
mitchell852: this seems like it could cause other problems. i.e. does an admin_down mid still need to be marked as updated from a upd_pending / reval_pending perspective? if it is excluded and an admin_down mid is flipped to reported suddenly, it might not have the required regex_revalidate.config file that it needs to invalidate content correctly.

my point is, since the perl endpoint only excluded OFFLINE, maybe that was done for a very valid reason.

https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/UI/Server.pm#L917

rob05c: You're right, the ideal solution would handle that. But the alternative is Edges never being able to finish a Reval, if any Mid is ADMIN_DOWN and not running ORT. Until we can get the ideal solution in place, this seems like the less bad option.

mitchell852: @rob05c - do you know off the top of your head what happens when a cache is moved from admin_down/offline/pre_prod to online/reported? i'm guessing nothing happens (except queue updates is triggered on all "child" caches"). 

my point is - if a mid is moved from admin_down/offline/pre_prod to online/reported, it could end up with an out of date regex_revalidate.config file (which could screw up content invalidation) but maybe it's standard operating procedure to "badass" a cache when moving it to online/reported status?

SaravanakumarSivasankaran: > Adding my thoughts, belive it may help some way.
> 
> I assume, the ADMIN_DOWN for the cache server will be set by the admin person through trafficops UI( ie., OFFLINE the server)
> Consider the below setup and scenario:
> 
> SETUP:
>     mid_cache_group1:  mid1 and mid2
>     edge_cache_group1: edge1 (Its parent is mid_cache_group1)
> 
> SCENARIO:
> If operator ADMIN_DOWN mid1 for some reason(say mid1 is dead).  
> Then upd_pending should set to true for mid1,mid2 and edge1. Hope this will done by the operator through trafficops Queue update.
> But immediate ORT run on edge1 should wait only for mid2 (for to complete upd_pending) NOT for mid1 (since mid1 is ADMIN_DOWN).
> Though mid1 is ADMIN_DOWN, on its cron schedule ORT job, it will update its required configuration since upd_pending is set to True.
> 
> SUGGESTION:
> Like OFFLINE in that query, can we handle something like <<<...WHERE pstatus.name != 'OFFLINE' AND pstatus.name != 'ADMIN_DOWN')..>>
>

PR #3691, opened 2019-06-20
Some best-practices updates to the kickstart script

ocket8888:
- [x] This PR obsoletes #3503
This PR makes some best-practices updates to the script responsible for generating the "network line" for kickstart ISO creation. Puts parenthesis around `print` statement args, better implementation of `str.format` and `str.join`, replaces a check for identity equality with the proper check for value equality, removed an unused variable, checking for presence of a member of a `set`/`dict` with `in`/`not in` instead of deprecated methods and/or weird argument nesting, etc.

- Traffic Ops

Generate an ISO through a Traffic Ops instance that has the new version of the script, verify that it comes out correctly. The script has no tests, but you could also just check the output using Docker or a VM.

### DIFF
--- a/misc/kickstart_create_network_line.py
+++ b/misc/kickstart_create_network_line.py
@@ -14,32 +14,28 @@
 #
 
 
-''' 
+'''
     This reads a configuration file and checks for functioning
-    network links in /sys/class/net/*, then emits a ks.cfg network line. 
+    network links in /sys/class/net/*, then emits a ks.cfg network line.
     '''
 
 import os
 import re
 
-# Global bad: 
-global DEBUG
-DEBUG = True
-
 global TO_LOG
 # This "logs" to stdout which is captured during kickstart
 TO_LOG = True
 
-# This is the standard interface we install to. It is set to a speed value of 
-# 5 (vice 100,1000 or 10000) later on and any other interface will override it 
-# if you've got something faster installed. 
+# This is the standard interface we install to. It is set to a speed value of
+# 5 (vice 100,1000 or 10000) later on and any other interface will override it
+# if you've got something faster installed.
 standard_interface=['p4p1']
 
 ## These are configuration settings:
 # Name of Configuration file:
 cfg_file = "network.cfg"
 
-# Where linux is putting the interface stuff: 
+# Where linux is putting the interface stuff:
 iface_dir = '/sys/class/net/'
 
 ignore_interfaces = ['lo','bond']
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3886/
Test PASSed.

misc/kickstart_create_network_line.py
```diff
@@ -14,60 +14,56 @@
...
 # Or if there's not 2 interfaces at the same speed you want the fastest.
-# auto is expected to be a string, otherwise use integers: 
-# Speed is in megs. 1000 is 1 gig, 10000 is 10G. 
+# auto is expected to be a string, otherwise use integers:
+# Speed is in megs. 1000 is 1 gig, 10000 is 10G.
 
 iface_speed = 'auto'
 
-restring="%(iface_dir)s(?P<iface>.*)/speed" %vars()
+restring="".join([iface_dir, "(?P<iface>.*)/speed"])
```
rawlinp: I wouldn't really say string joining like this is a "best practice". It arguably makes the statement harder to read and more error prone. I do agree with not using `vars()` though, but that can be changed like this: 
```
restring="%s(?P<iface>.*)/speed" % iface_dir
```
or
```
restring = iface_dir + "(?P<iface>.*)/speed"
```
What do you think? If 3.6 becomes the new default, I'd recommend using f-strings.

Ditto this comment to the other changes like this below.

misc/kickstart_create_network_line.py
```diff
@@ -109,7 +105,8 @@ def find_usable_net_devs(location):
                 add = False
         if speed  <= 0:
             add = False
-        if TO_LOG: print add, dev
+        if TO_LOG:
+            print (add, dev)
```
rawlinp: In python 2 this changes `(add, dev)` into a tuple which is then printed as a tuple. Why not just add `from __future__ import print_function` as the first line of code in the script and make it actually use the print function from python3? Ditto this comment to the other changes like this below.

misc/kickstart_create_network_line.py
```diff
@@ -215,12 +222,12 @@ def useable_interfaces(net_devs, nc, iface_speed):
...
-if not nc.has_key('DHCP'):
+# This should be set to no in the config file, but that could change:
+if "DHCP" not in nc:
     nc['DHCP']='no'
 
-# This should be set to no in the config file, but that could change: 
-if not nc.has_key('DHCP'):
+# This should be set to no in the config file, but that could change:
+if "DHCP" not in nc:
     nc['DHCP']='no'
```
rawlinp: This `if` block seems to be duplicated.

misc/kickstart_create_network_line.py
```diff
@@ -231,43 +238,46 @@ def useable_interfaces(net_devs, nc, iface_speed):
...
             iface_problems = "IPv6 enabled but no address provided"
-        else: 
+        else:
             iface_problems = "{0} and IPv6 enabled but no address provided".format(iface_problems)
     if re.search(":",nc['IPV6ADDR']):
         if iface_problems is False:
             iface_problems = "IPv6 is disabled, but IPV6ADDR was set to {0}".format(nc['IPV6ADDR'])
         else:
-            iface_problems = "{0} and IPv6 is disabled, but IPV6ADDR was set to {1}".format(iface_problems,nc['IPV6ADDR'])
+            iface_problems = " ".join([iface_problems, "and IPv6 is disabled, but IPV6ADDR was set to", nc['IPV6ADDR']])
```
rawlinp: I would leave the original line as-is

misc/kickstart_create_network_line.py
ocket8888: I remember reading somewhere that the recommended method of string concatenation was to use `join`. I just looked it up, though, and it appears that what I was thinking of was this line from PEP8:

> _" In performance sensitive parts of the library, the ''.join() form should be used instead. This will ensure that concatenation occurs in linear time across various implementations."_

So it's probably fine to use `+`/`+=` here since that's not exactly _"performance sensitive"_.

As to "fstring"s, I'd rather not break 2.7 compatibility immediately - though technically it should be fine because 3.x includes a deprecation notice for Python 2. I don't _think_ 2.7 has those.

misc/kickstart_create_network_line.py
ocket8888: This again just came down to string concatenation - I wanted to be consistent. What about, instead of the original line, this was made to use `+` as above?

misc/kickstart_create_network_line.py
rawlinp: Yeah 2.7 definitely doesn't have f-strings. I'm just saying if/when 3.6 becomes the new default on centOS, we could/should use f-strings. But they should be disregarded for now since IMO this seems like something that depends on the default python installation which is still 2.7 AFAIK. So probably best to maintain 2/3 compatibility for the time being (otherwise it will probably break people).

misc/kickstart_create_network_line.py
rawlinp: I don't think _that_ will work here because `iface_problems` appears to be a bool

misc/kickstart_create_network_line.py
ocket8888: odd

misc/kickstart_create_network_line.py
ocket8888: Well I could just cast it to a string.

ocket8888: Actually, `format`'s way better in that case. I'll just change it back.

PR #3699, opened 2019-06-25
terminate disk check regex with a digit

### LINKED ISSUE #3698: ORT has problem with disk enumeration
On systems with many disks, if the OS root disk ends up being on a device name that "wraps around" past `/dev/sdz`, then ORT mistakenly flags a cache disk as one with active partitions. For example, if the root disk device is `/dev/sdac` and `/dev/sda` is a device specified as a cache disk, when `sda` is checked against the mount table, it will actually get a false positive, due to unqualified regex. Updating the regex to be qualified by a digit, fixes this.

I have a patch, and PR is incoming...

guzzijason:
Fixes #3698

- [x] This PR fixes #3698

Trivial fix to correct an unqualified regex - no new functionality introduced.

- Traffic Ops ORT

Running ORT on a host with many cache disks, where the OS "root" device name shares the same prefix as one of the cache disks. For instance, a case where OS root disk `/dev/sdac` exists on a server where `/dev/sda` is a valid cache disk. In previous version of ORT, this would cause a conflict causing the ORT to believe that the cache disk has mounted filesystems.

- master (61a985ce4)
- 3.0.0
- 3.0.1
- (probably others)

- [ x] This PR includes tests OR I have explained why tests are unnecessary
- [ x] This PR includes documentation OR I have explained why documentation is unnecessary
- [ x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x ] This PR includes any and all required license headers
- [ x] This PR ensures that database migration sequence is correct OR this PR does not include a database migration
- [x ] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

### DIFF
--- a/traffic_ops/bin/traffic_ops_ort.pl
+++ b/traffic_ops/bin/traffic_ops_ort.pl
@@ -2945,7 +2945,7 @@ sub adv_processing_udev {
 				}
 				( my @df_lines ) = split( /\n/, `/bin/df` );
 				foreach my $l (@df_lines) {
-					if ( $l =~ m/$dev_path/ ) {
+					if ( $l =~ m/$dev_path\d/ ) {
 						( $log_level >> $FATAL ) && print "FATAL Device /dev/$dev has an active partition and a file system!!\n";
 					}
 				}


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3910/
Test PASSed.

guzzijason:
PR details updated with template info.

PR #3701, opened 2019-06-25
Add pagination support for TO API endpoints that use BuildWhereAndOrderBy

### LINKED ISSUE #1913: Traffic OPs API in Golang needs to support pagination
limit=X (The number of results returned)
page=X (Which page of those results you are on)

alexluckerman:
Currently uses `limit` and `offset` query parameters, but `page` might also be a good option. **This PR depends on #3740 for documentation**

- [x] This PR fixes #1913

- Traffic Ops

Build TO Golang, run db_helpers_test.go, start Traffic Ops, make a few queries to endpoints that use BuildWhereAndOrderBy to make sure that the `limit` and `offset` parameters are working properly


asfgit:

mitchell852:
this will also require an addition to the changelog.md file

mitchell852:
also, looks like this PR adds 2 things: pagination and sortBy? can you update the title and description of the PR to reflect that?

alexluckerman:
Oops, the sortOrder part wasn't supposed to be included here. Also, the new query parameters affect all the same endpoints as PR [3682](https://github.com/apache/trafficcontrol/pull/3682), so since these each involve changing the same tables in the docs I was thinking I'd just make a separate PR for all of the documentation

ocket8888:
I, personally, am against merging new features without documentation. Especially because of how we do releases. For example, suppose this PR gets cherry-picked into a release. Now, if the docs get pulled in as well, they will be wrong because PR 3682 wasn't also cherry-picked. And that's if someone remembers to cherry-pick the docs at all.

rawlinp:
IMO documentation-only PRs are the way to go. Separate "components" should be in separate PRs, and I consider documentation to be a separate component. Different reviewers have different areas of expertise, and it slows us down to have one reviewer reviewing only the TP changes, one reviewing only the TO changes, and one only reviewing docs changes. In order for that PR to merge it would need separate approval from all three reviewers, instead of just being able to merge individual components by the reviewers that have expertise in that area.

This also helps individual components remain backwards-compatible. I.e. you should generally be able to merge TR changes before corresponding TO changes because TR should always be backwards-compatible w/ older versions of the CRConfig. You don't necessarily _have_ to have TO and TR changes in the same PR.

That said, the new features should still be documented regardless; I'm just saying I think it's easier when separate things are done in separate PRs.

ocket8888:
I disagree. The documentation is not a separate component, it is integral to *all* components. If I hand someone Traffic Ops, they are utterly unable to use it without the documentation. Similarly, if you add a feature and don't document it, then you've essentially done nothing to the eyes of the rest of the world.

I honestly don't think it takes much expertise to review docs changes, if you check that they follow the [documentation guidelines](https://traffic-control-cdn.readthedocs.io/en/latest/development/documentation_guidelines.html#writing), and can run `pkg docs` and read the output to make sure it all makes sense, then you're done. If something doesn't come out right, or if the compiler throws a warning/error, that's all you gotta say. You don't need to know why - that's the writer's problem :P.

If writing/reviewing the documentation really is "expert" level, then something is seriously wrong with our documentation. It's my personal opinion that reST isn't terribly difficult, but many complaints I hear are about it being obscure or arcane - and I think that's because, to be fair, that format itself doesn't come with much of a guide. It probably wouldn't take too long for me to write up an example that showcases everything a prospective writer would need to use, if that would be helpful?

rawlinp:
You are disagreeing as if I had said nothing needs documented. That's not what I said. It's not about the level of expertise required to review docs; it's about keeping separate things separate. Focused and concise PRs are much easier to review and merge, and requiring the documentation and code to be in the same PR just slows down the process.

alexluckerman:
How about if the description of the PRs that required the separate documentation indicated that they **required** that documentation PR to be put into a release so that the cherry pick mishap would be much less likely to happen?

rawlinp:
PRs referencing other related PRs (e.g. a PR with the TO changes referencing its corresponding PR with docs changes) is encouraged. I think the backport case is rare because new features should never be backported (only bug fixes) unless there is very good reason to, but if the new feature _did_ require backporting, we would be able to see that the desired PR also referenced its corresponding docs PR.

There can also be varying degrees of documentation for any given feature. Some things should be documented _before_ coding even begins (e.g. the RFE). Once a very high-level, basic RFE has been documented, it might make sense to get started on a blueprint that the developer will use to explain the new feature's implementation. That should be mostly done before coding begins, but the blueprint may change once the developer has started the implementation and has found a better way, so they may have to update the blueprint along the way. The user-facing documentation should be developed last IMO, after the code is complete and using the blueprint as a reference, because if you write documentation first, you will have to constantly change the documentation _as well as_ the code while working on the new feature. Sometimes you could make a change to the code but forget to also update the docs. Then we end up with _incorrect_ documentation which can sometimes be worse than _no_ documentation. If we wait until the code is complete before starting the user-facing documentation, you should only have to write them once and not worry about details changing in parallel with the code.

rob05c:
My biggest concern with merging code without docs, and I suspect @ocket8888 's as well, is that it never gets done. But I see @rawlinp 's point as well, about docs being outdated before we even release.

If we merge code without docs, IMO it should be a requirement before merging that a Github Issue be created for making those docs, and put in the milestone of the next release, to be sure don't release without adding them.

In fact, I'd propose adding it to the PR template, "This PR includes documentation OR I have explained why documentation is unnecessary OR I have created a Github Issue for the documentation and worked with a Committer to place it in the next Milestone."

How does that sound?

rob05c:
I also like @alexluckerman's suggestion of finding a way to associate them, to ensure they're cherry-picked together. But that's much harder. Linking the PRs is good, but often cherry-pickers only see git, and don't even look at Github. But putting it in the changeset message is hard. I'm not sure what the best solution is. Anyone have any better ideas?

That said, it's not my biggest concern, as  @rawlinp said, feature backports should be very rare.

ocket8888:
> This PR includes documentation
> OR I have explained why documentation is unnecessary
> OR I have created a Github Issue for the documentation and worked with a Committer to place it in the > next Milestone.

This would be fine with me, but I think it's something that should be brought to the mailing list's attention, since it has the potential to annoy a lot of people.

> _"You are disagreeing as if I had said nothing needs documented. That's not what I said. It's not about the level of expertise required to review docs; it's about keeping separate things separate."_

No, that's what I mean, the docs and the code _aren't_ separate, and you're not actually done until they're both in there. Separating them is taking a feature and cutting it in half, which will lead to what we had in v3.0 where the docs lie to you because things were included without documentation. That makes the end result unusable by anyone except the person who wrote the code.

> _"I think the backport case is rare because new features should never be backported (only bug fixes) unless there is very good reason to"_

I agree, but historically that hasn't been true. Backports happen - not directly into a release, but into the release candidate branches.

> _"The user-facing documentation should be developed last IMO, after the code is complete and using the blueprint as a reference, because if you write documentation first, you will have to constantly change the documentation as well as the code while working on the new feature."_

These are all good points. But the way the repository is set up, we put 1 feature in 1 PR and 1 PR becomes 1 commit. Not including the documentation violates that, because it *is* part of the feature. I think it'd be neat to use milestones to track the commits that go into a feature and the progress towards its completion, but that's not possible for contributors (without doing what rob05c said and have a committer set it up). When an RC4 is made, nobody is going to look at all the PRs they cherry-pick to see if they reference others and resolve all of that - I think a milestone would make it easier to tell at a glance what went where - and that means docs will get left behind.

rawlinp:
We should leave it up to the contributor to decide how they would like to contribute documentation and should be happy with any documentation contributed at all. Undocumented features aren't ideal, but discouraging contributors from contributing anything at all by raising the bar too high is way less ideal IMO. If the contributor doesn't feel comfortable updating the docs for whatever reason, someone else should take that on if necessary. The more options a contributor has and the easier it is to do, the more likely they will be happy to write the documentation. 

How about instead of declaring that such and such needs to be documented, we simply point them to the existing documentation that should be updated (so they don't need to go find it themselves), and leave it up to them to contribute however they like.

mitchell852:
Here are the things I like:
- small PRs
- complete PRs

trying to decide which I like more :)

but I think at the end of the day, I do think this project suffers from poor test coverage and undocumented/poorly documented features and I am guilty of both of these. Since our PR template has this checklist in it:

image

I understand that to mean that the PR will not be reviewed until all the checkmarks are checked and if documentation is required, then it should be included with the PR. same with tests, if tests make sense, they should be included. otherwise, like @rob05c said, it just won't get done.

it raises the bar to contribute but at the same time it raises the quality bar of the project.

of course, that checklist is just a guideline and if you can convince the reviewer to overlook documentation for now because you PROMISE to have it done/merged before the next open source release, that's between you and the reviewer imo.

alexluckerman:
@ocket8888 I did not mean to push those three files at all - I believe I just managed to undo that

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -63,14 +66,26 @@ func BuildWhereAndOrderBy(parameters map[string]string, queryParamsToSQLCols map
 			log.Debugln("Incorrect name for orderby: ", orderby)
 		}
 	}
+	// needs error checking
+	if limit, exists := parameters["limit"]; exists {
+		log.Debugln("limit: ", limit)
+		paginationClause += " " + limit
+		if offset, exists := parameters["offset"]; exists {
+			paginationClause += BaseOffset + " " + offset
```
ocket8888: This type of query building is unsafe. In testing I was able to run arbitrary code in the `limit` clause by submitting a request like e.g.: `toget '1.4/servers?limit=1%3B%20DROP%20TABLE%20server%20CASCADE%3B%20--&orderBy=host_name'` which is something that even a read-only user can do.

PR #3706, opened 2019-06-28
Fix TO-Go's internal shared riak cluster finalizer

rawlinp:
Stop the old riak cluster at garbage-collection time instead of
immediately after creating a new riak cluster. Requests might be
currently in flight using the old cluster, and stopping the old cluster
with requests in flight can cause those requests to fail. Instead, all
in-flight requests using the old cluster will be handled normally, and
once there are no in-flight requests to reference the old riak cluster,
it will be stopped and garbage-collected. Once the new riak cluster is
started, any new requests that arrive will be handled by the new riak
cluster, so the old riak cluster should be stopped and garbage-collected
shortly after that point.

- Traffic Ops
- Traffic Vault

1. Create a riak cluster in TO (make sure riak servers are set to ONLINE).
2. Start making some repeated, simultaneous requests to TO that require riak (e.g. `/api/1.3/cdns/name/:name/sslkeys.json`). This will start up a new riak cluster in TO -- check the logs for `New riak cluster started: ` followed by a memory address.
3. Change the riak cluster in TO (e.g. set one of the servers to OFFLINE, leaving at least one other ONLINE). This will create a new riak cluster in TO -- check the logs for `New riak cluster started: ` followed by a memory address.
4. Shortly after the log in step 3 (but necessarily _instantly_ after), there should be another log line that says `running finalizer for riak sharedcluster` followed by the memory address from the log line in step 2. Verify that these memory addresses match.
5. Verify that in-flight TO requests started in step 2 were not 5XX'd due to the old cluster being stopped.

- master (729de83dcb289e4ccfb43cce838aa07b481ec4d1)

Small, targeted patch -- no automated tests or docs are necessary IMO.

### DIFF
--- a/traffic_ops/traffic_ops_golang/riaksvc/riak_services.go
+++ b/traffic_ops/traffic_ops_golang/riaksvc/riak_services.go
@@ -321,10 +321,15 @@ func GetPooledCluster(tx *sql.Tx, authOptions *riak.AuthOptions, riakPort *uint)
 		newcluster, err := GetRiakCluster(newservers, authOptions)
 		if err == nil {
 			if err := newcluster.Start(); err == nil {
-				log.Infoln("New cluster started")
+				log.Infof("New riak cluster started: %p\n", newcluster)
 
 				if sharedCluster != nil {
-					runtime.SetFinalizer(sharedCluster, sharedCluster.Stop())
+					runtime.SetFinalizer(sharedCluster, func(c *riak.Cluster) {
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3917/
Test PASSed.

Daisykh:
Ran through everything, looks good to me and was able to verify in-flight TO requests weren't 5XX'd.

dangogh:
tested and approved by @Daisykh

PR #3711, opened 2019-07-02
Update/add TO Golang audit logs on non-DS endpoints

### LINKED ISSUE #3575: Audit, complete and improve TO change log
We have X api routes written in Go: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/routing/routes.go

We have Y api routes written in Perl (that have yet to be rewritten in Go): https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/TrafficOpsRoutes.pm#L394

1. Ensure that all the api routes that modify data (for example, create, update or delete), generate a TO change log (which gets stored in the log table).

2. Ensure that all the change log entries are comprehensive and readable. For example, `Server updates queued for cdn 2` could be better.

(one could argue that we should not touch the Perl routes but i believe this is a pretty easy change to them that will add value. especially, if we find any perl routes that don't create a change log at all. and also, the completion of the perl->go rewrite is still a ways off)

alexluckerman:

- [x] This PR (partly) fixes #3575

- Traffic Ops

I believe since this only affects our logs it doesn't need tests, and I believe the same is true for documentation and a changelog update - but I could be wrong.

There might need to be some test changes for this PR, but otherwise this can be verified by starting CIAB and ensuring the change log messages are formatted correctly in TP, and also by performing a few create/update/delete actions in TP to verify that they are also being logged properly.

### DIFF
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -22,6 +22,7 @@ package crconfig
 import (
 	"database/sql"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"time"
@@ -175,7 +176,7 @@ func SnapshotHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig and Monitor performed for "+cdn, inf.User, inf.Tx.Tx)
+	api.CreateChangeLogRawTx(api.ApiChange, fmt.Sprintf("CDN: %v, ID: %v, ACTION: Snapshot of CRConfig and Monitor", cdn, inf.IntParams["id"]), inf.User, inf.Tx.Tx)
 	api.WriteResp(w, r, "SUCCESS")
 }
 
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -22,7 +22,6 @@ package crconfig
 import (
 	"database/sql"
 	"errors"
-	"fmt"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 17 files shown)


asfgit:

alexluckerman:
For some reason, trying to handle the id key not existing in a `map[string]interface{}` isn't working, and the id just prints out as `<nil>`:

```
id, ok := keys["id"]
if !ok {
	id = "N/A"
}
msg := fmt.Sprintf("%v: %v, ID: %v, ACTION: %v %v, keys: %v", strings.ToTitle(objType), auditName, id, strings.Title(action), objType, keyStr)
```

mitchell852:
looking at the PR title, does this mean ALL Golang endpoints (excluding GETs) now have a change long entry?

alexluckerman:
Yes, all the ones aside from DeliveryService-related endpoints since Daisy is handling those

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -175,7 +176,7 @@ func SnapshotHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig and Monitor performed for "+cdn, inf.User, inf.Tx.Tx)
+	api.CreateChangeLogRawTx(api.ApiChange, fmt.Sprintf("CDN: %v, ID: %v, ACTION: Snapshot of CRConfig and Monitor", cdn, inf.IntParams["id"]), inf.User, inf.Tx.Tx)
```
dangogh: probably want to avoid using fmt in this situation and especially `%v`.  It has to use reflection to fill in the string and and Sprintf doesn't add that much clarity...

mitchell852:
@alexluckerman - the idea with the PR template is that ALL checkboxes should be checked to indicate that the PR is ready for review. The language has been written in a way that should allow you to check them all.

PR #3725, opened 2019-07-15
Make entire row clickable in TP assignment popups

### LINKED ISSUE #3454: Make entire checkbox rows clickable in TP
For table rows in Traffic Portal that contain checkboxes (e.g. linking profiles to a parameter, linking delivery services to servers, etc), the entire row should be clickable, not just the checkbox itself. This would make it easier to "check" the checkbox for the row rather than having to precisely click inside the tiny checkboxes.

alexluckerman:
- [x] This PR fixes #3454 

- Traffic Portal

This is a very intuitive UI change, so I don't think it needs documentation or a changelog update, but I'm happy to add either if I'm wrong.

Perform the following tasks in TP and ensure you can click both the checkbox and rest of the row:
- link a parameter to a cache group
- assign DSes, users, and resolvers to a federation
- assign DSes to servers (and vice versa)
- assign profiles to parameter (and vice versa)
- assign capabilities to a role

Since there aren't existing tests of these dialogs I could add to, I opted not to create new UI tests for this as that doesn't really seem like it's in the scope of what this PR is trying to do.

### DIFF
--- a/traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
+++ b/traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
@@ -44,6 +44,10 @@ var TableCacheGroupParamsUnassignedController = function(cg, parameters, $scope,
 		}
 	};
 
+	$scope.handleRowClick = function($index) {
+		$('#checkbox-' + $index).trigger('click');
+	};
+
 	$scope.submit = function() {
 		$uibModalInstance.close(selectedParams);
 	};
--- a/traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
+++ b/traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
@@ -36,7 +36,7 @@ var TableCacheGroupParamsUnassignedController = function(cg, parameters, $scope,
 	};
 
 	$scope.updateParams = function($event, paramId) {
-		var checkbox = $event.currentTarget.firstChild.firstChild; // finds the input element from the table row which was clicked
+		var checkbox = $event.currentTarget.firstElementChild.firstElementChild; // finds the input element from the table row which was clicked
 		if (!checkbox.checked) {
 			addParam(paramId);
 			checkbox.checked = true;
--- a/traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
+++ b/traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-var TableProfileParamsUnassignedController = function(cg, parameters, $scope, $uibModalInstance) {
+var TableCacheGroupParamsUnassignedController = function(cg, parameters, $scope, $uibModalInstance) {
 
 	var selectedParams = [];
 
@@ -36,11 +36,13 @@ var TableProfileParamsUnassignedController = function(cg, parameters, $scope, $u
 	};
 
 	$scope.updateParams = function($event, paramId) {
-		var checkbox = $event.target;
-		if (checkbox.checked) {
+		var checkbox = $event.currentTarget.firstChild.firstChild; // finds the input element from the table row which was clicked
+		if (!checkbox.checked) {
 			addParam(paramId);
+			checkbox.checked = true;
 		} else {
 			removeParam(paramId);
+			checkbox.checked = false;
 		}
 	};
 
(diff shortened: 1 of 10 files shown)


asfgit:

traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
```diff
@@ -36,11 +36,13 @@ var TableProfileParamsUnassignedController = function(cg, parameters, $scope, $u
 	};
 
 	$scope.updateParams = function($event, paramId) {
-		var checkbox = $event.target;
-		if (checkbox.checked) {
+		var checkbox = $event.currentTarget.firstElementChild.firstElementChild; // finds the input element from the table row which was clicked
```
mitchell852: this feels hacky. you are digging quite deep to get the checkbox

mitchell852: i'm pretty sure there a click event on that checkbox you can simply catch or something

traffic_portal/app/src/common/modules/table/cacheGroupParameters/table.cacheGroupParamsUnassigned.tpl.html
```diff
@@ -34,8 +34,8 @@ <h3 class="modal-title">Link Parameters to {{::cg.name}}</h3>
         </tr>
         </thead>
         <tbody>
-        <tr ng-repeat="p in ::unassignedParams">
-            <td><input type="checkbox" ng-click="updateParams($event, p.id)"></td>
+        <tr ng-repeat="p in ::unassignedParams" ng-click="updateParams($event, p.id)"">
+            <td><input type="checkbox" ng-click="$event.preventDefault()"></td>
```
mitchell852: why prevent the default. i would suggest you let the browser do what it does rather than doing it yourself.

rawlinp:
Yay, thanks for addressing one of my open TP issues 😄

traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
mitchell852: can you do something like this instead? on tr click, force a click on the nested checkbox and then all the existing code can stay the same?

traffic_portal/app/src/common/modules/table/cacheGroupParameters/table.cacheGroupParamsUnassigned.tpl.html
ocket8888: I think this will actually break the ability to click on the checkbox itself:
```html
<!DOCTYPE html>
<html lang="en-US">
<head>
	<meta charset="utf-8"/>
	<title>Test</title>
</head>
<body>
	<label id="label">Test <input id="checkbox" type="checkbox"></label>

	<script type="text/javascript">
		const label = document.getElementById('label');
		const checkbox = document.getElementById('checkbox');

		label.onclick = e => {
			e.originalTarget.firstElementChild.checked = !e.originalTarget.firstElementChild.checked;
		}

		checkbox.onclick = e => {
			e.preventDefault();
		}
	</script>
</body>
</html>
```
When I checked out the above code, it dumps an error to the console every time the `<label>` is clicked - because of event bubbling. When you click on the checkbox, you'll notice that it doesn't toggle its `checked` state - and an error is dumped to the console. What you want is - I think - to stop the event bubbling in the handler for the click event on the whole `<tr>` - but then also check that it's not a sub-element. This code:
```html
<!DOCTYPE html>
<html lang="en-US">
<head>
	<meta charset="utf-8"/>
	<title>Test</title>
</head>
<body>
	<label id="label">Test <input id="checkbox" type="checkbox"></label>

	<script type="text/javascript">
		const label = document.getElementById('label');
		const checkbox = document.getElementById('checkbox');

		label.onclick = e => {
			e.stopPropagation();

			const target = e.explicitOriginalTarget;

			if (target === label) {
				target.firstElementChild.checked = !target.firstElementChild.checked;
			}
		}
	</script>
</body>
</html>
```
produces no errors and works exactly as expected.

alexluckerman: explicitOriginalTarget appears to be Mozilla-specific

ocket8888: whoops :P pretty sure `target` works fine, just a bad habit I've picked up because I only use Firefox.

mitchell852:
@alexluckerman - in your PR description you said

> assign capabilities to a user

how exactly do you do that? I don't recall that being a feature of TP.

ocket8888:
I just assumed he meant "role" instead of "user".

It's not a page you can navigate to through the UI, but I believe the controller exists. So theoretically you could go to a path that would show it

mitchell852:
> I just assumed he meant "role" instead of "user".

yep, that makes more sense. capabilities are assigned to roles and a role is assigned to a user so indirectly you can "assign capabilities to a user"  but not explicitly.

alexluckerman:
I did mean "role" - my bad

PR #3726, opened 2019-07-15
Add redirect to link profiles page on TP parameter creation

### LINKED ISSUE #3453: After creating a new parameter in Traffic Portal, redirect to the "link profiles" page
What I want to do:
- create a profile parameter and assign it to some existing profiles

What this currently requires in TP:
1. create the profile parameter (page is then redirected to the parameters table)
3. search for parameter that I just created
4. click on parameter that I just created
5. click "View Profiles"
6. click "link profiles to parameter"
7. select profiles that I want to link to

What I think TP should require instead instead:
1. create the profile parameter (page is then redirected to the "link profiles to parameter" view)
2. select profiles that I want to link the new parameter to

This would really streamline the process of creating a parameter and assigning it to one or more profiles, which I think is the most common use case of creating a new parameter.

alexluckerman:
In order to get this to work, I had to change the Restangular response interceptor. It passed all of the tests, but since test coverage isn't perfect it could cause problems on some part of TP that I haven't attempted to use yet.

- [x] This PR fixes #3453

- Traffic Portal

Create a new parameter and ensure you are immediately redirected to the link profiles page


asfgit:

traffic_portal/app/src/app.js
```diff
@@ -425,13 +425,13 @@ var trafficPortal = angular.module('trafficPortal', [
 
             RestangularProvider.setBaseUrl(ENV.api['root']);
 
-            RestangularProvider.setResponseInterceptor(function(data, operation, what) {
+            RestangularProvider.setResponseInterceptor(function(data, operation) {
 
-                if (angular.isDefined(data.response)) { // todo: this should not be needed. need better solution.
-                    if (operation == 'getList') {
-                        return data.response;
+                if (angular.isDefined(data.response)) {
```
mitchell852: not sure how i feel about this change. this is making a very global change to add a simple redirect for one use case as i understand it. this would require a lot of testing of TP.

mitchell852: i'll give it some more thought.

traffic_portal/app/src/common/api/ParameterService.js
```diff
@@ -30,9 +30,9 @@ var ParameterService = function(Restangular, $http, $q, locationUtils, messageMo
     this.createParameter = function(parameter) {
         return Restangular.service('parameters').post(parameter)
             .then(
-            function() {
+            function(response) {
                 messageModel.setMessages([ { level: 'success', text: 'Parameter created' } ], true);
-                locationUtils.navigateToPath('/parameters');
```
ocket8888: Couldn't this be accomplished by just changing this path to `'/parameters/' + String(parameter.id) + '/profiles'` ?

mitchell852: maybe? i need to read up on the  issue. the PR doesn't say exactly what the change is trying to accomplish.

mitchell852: @ocket8888 - not quite. the issue created by @rawlinp (#3453) doesn't want a simple redirect to `'/parameters/' + String(new_parameter.id) + '/profiles' ` after creating a parameter (although that would be an improvement imo) but instead, wants the parameter/profile assignment modal opened after creating a parameter.

alexluckerman:
> plus, restangular is the the process of being removed so depending on it to deliver this feature will not work for the long term.

If that's the case, then how about I just rewrite this Restangular request to use whatever we're planning to use once Restangular is removed?

traffic_portal/app/src/common/api/ParameterService.js
rawlinp: Yeah, @mitchell852 is right. The idea is that if you just created a parameter, you most likely want to then assign that parameter to a profile.
So it seems like there would be two steps:

1. After creating a parameter, instead of navigating back to the "parameters table" view, navigate to the "profiles linked to this parameter" view.
2. Automatically open the "assign parameter to profiles" modal.
When you're done assigning the parameter to profiles, you should end up on the "profiles linked to this parameter" view.

Orphaned parameters don't make much sense, so the easier we can make it to assign parameters to profiles right after creating them, the better.

ocket8888:
We're moving to just using `$http` since it's nearly identical to Restangular in syntax.

mitchell852:
@alexluckerman - you can base your PR off of these PR's if you want to get restangular out of the way #3625 #3623

traffic_portal/app/src/common/api/ParameterService.js
mitchell852: I do feel like this could be simplified a lot however (code-wise) if the requirement changed to this:

- On successful parameter creation, redirect TP to /parameters/new_parameter.id/profiles where the user is required (if they want) to click the link button to link profiles to their new parameter

image

this would turn all these code changes into a one line change :)

mitchell852: ....and if the above is OK, then the same could be done on successful profile creation (with another one line change)....

- On successful profile creation, redirect TP to /profiles/new_profile.id/parameters where the user is required (if they want) to click the link button to link parameters to their new profile.

2 lines of code == 2 new timesaving behaviors :)

rawlinp: Well when you put it that way, I can get on board with that. If it obfuscates the code too much to auto-open the assignment modal, I'd be +1 on that trade-off.

ocket8888:
Won't those need to get merged before this, then?

mitchell852:
> Won't those need to get merged before this, then?

probably unless he rebased off your PR branches or something. anyhow, might be moot as i changed the requirements a bit. see discussion between me and rawlin.

Issue #3512, opened 2019-04-20
Order of operations issue with DS Requests


jhg03a:
I believe there is an order of operations issue in DS Requests.

Perform the following sequence of steps:
1. User A creates a DS request to change LongDesc2
2. Before the DS request is fullfilled, User B updates LongDesc3 and fulfills immediately
3. User B sees an error as well as their change
4. User A sees that their DS Request now includes changes they didn't ask for
5. User B fulfills DS Request from User A
6. User B now sees that their change from 2 is now gone

I believe the correct action is to check for an existing DSR before updating the DS.  The current behavior could lead to accidental deployment/reversion of changes.

mitchell852:
This was fixed with PR #3727 and i confirmed that you cannot create a 2nd ds request when one is currently open.

PR #3727, opened 2019-07-15
Fix bug that causes DS requests to fulfill immediately while other requests are still open

### LINKED ISSUE #2595: trying to delete a ds that has an open ds request is successful
With ds requests turned on in traffic_portal_properties.json, all changes to a ds (create, update or delete) result in the creation of a ds request.

however, if you are an admin, you can select "fulfill immediately" for the ds request. this will result in:

1. creating a ds request to delete the ds and immediately marking it as complete
2. deleting the ds

#1 should fail if an existing ds request is open (status != complete) for the ds which it does but #2 still succeeds (which it should not).

UPDATE:
@dsouza93 found this issue as well (#3717), the underlying issue sounds the same:

When there is an open delivery service request for a delivery service, and a user attempts to change a value on the delivery service and "fulfill immediately", Traffic Portal displays an error that there is an active delivery service request open for said delivery service. The delivery service data is then updated anyway. I am not sure if this is intended behavior and is to serve as a warning or if it supposed to prevent updates, but thought I'd bring it up.

My specific use case was when there was a delivery service request in the "pending" status with the only action left was to "complete" the request. I then set the delivery service itself to "active"/"fulfill immediately" without completing the delivery service request.

I am unsure if this will occur when other fields are altered or if it will occur when DS requests are in other states.

alexluckerman:
I also made a small tweak to the DS requests TP test as it seemed to be failing locally for me because it wasn't waiting long enough for TP to change its URL

- [x] This PR fixes #2595

- Traffic Portal

This is a small bug that should not require documentation or a changelog entry

Enable DS requests in TP and run the DS requests TP test, then manually verify that updating or deleting a DS while at least one DS request for that same DS is open fails

### DIFF
--- a/traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
+++ b/traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
@@ -59,8 +59,8 @@ var FormEditDeliveryServiceController = function(deliveryService, origin, type, 
 				deliveryService: deliveryService
 			};
 
-			// if the user chooses to complete/fulfill the delete request immediately, the ds will be deleted and behind the
-			// scenes a delivery service request will be created and marked as complete
+			// if the user chooses to complete/fulfill the delete request immediately, a delivery service request will be made and marked as complete, 
+			// then if that is successful, the DS will be deleted
 			if (options.status.id === $scope.COMPLETE) {
 				// first create the ds request
 				deliveryServiceRequestService.createDeliveryServiceRequest(dsRequest)
@@ -194,8 +194,8 @@ var FormEditDeliveryServiceController = function(deliveryService, origin, type, 
 					status: status,
 					deliveryService: deliveryService
 				};
-				// if the user chooses to complete/fulfill the update request immediately, the ds will be updated and behind the
-				// scenes a delivery service request will be created and marked as complete
+				// if the user chooses to complete/fulfill the update request immediately, a delivery service request will be made and marked as complete, 
+				// then if that is successful, the DS will be updated
 				if (options.status.id == $scope.COMPLETE) {
 					createDeliveryServiceUpdateRequest(dsRequest, options.comment, true).then(
 						function() {
--- a/traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
+++ b/traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
@@ -22,21 +22,6 @@ var FormEditDeliveryServiceController = function(deliveryService, origin, type, 
 	// extends the FormDeliveryServiceController to inherit common methods
 	angular.extend(this, $controller('FormDeliveryServiceController', { deliveryService: deliveryService, dsCurrent: deliveryService, origin: origin, type: type, types: types, $scope: $scope }));
 
-	var confirmNoOutstandingRequests = function(deliveryService) {
-		deliveryServiceRequestService.getDeliveryServiceRequests()
-		.then(
-			function(dsRequests) {
-				// search all requests for those that are not completed and share the same deliveryservice id
-				angular.forEach(dsRequests, function(value) {
-					if (value.status != 'complete' && value.deliveryService.id == deliveryService.id) {
-						return false;
-					}
-				});
-				return true;
-			}
-		);
-	};
-
 	var createDeliveryServiceDeleteRequest = function(deliveryService) {
 		var params = {
 			title: "Delivery Service Delete Request",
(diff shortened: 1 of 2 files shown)


asfgit:

traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
```diff
@@ -22,6 +22,21 @@ var FormEditDeliveryServiceController = function(deliveryService, origin, type,
 	// extends the FormDeliveryServiceController to inherit common methods
 	angular.extend(this, $controller('FormDeliveryServiceController', { deliveryService: deliveryService, dsCurrent: deliveryService, origin: origin, type: type, types: types, $scope: $scope }));
 
+	var confirmNoOutstandingRequests = function(deliveryService) {
+		deliveryServiceRequestService.getDeliveryServiceRequests()
+		.then(
+			function(dsRequests) {
+				// search all requests for those that are not completed and share the same deliveryservice id
+				angular.forEach(dsRequests, function(value) {
+					if (value.status != 'complete' && value.deliveryService.id == deliveryService.id) {
```
mitchell852: you should short circuit this condition by putting this as the first check as it will fail in most instances and then the 2nd check is never evaluated (might get you a little performance for free)

`if (value.deliveryService.id == deliveryService.id && value.status != 'complete') {`

mitchell852: also, if you rename the function something like `hasOpenRequest` and check for the statuses we know to be open (draft, submitted or pending), i think your other condition will read a little nicer

`if (options.status.id == $scope.COMPLETE && !hasOpenRequest(deliveryService))`

the reason i mention this is because this ds request code is a bit of a mess already as i'm sure you noticed, so easy to understand naming will make everyone's life easier.

mitchell852: so i've been thinking about this a bit more. the way you are performing this check is rather expensive. you are fetching all the DS requests (open or closed) and then looping thru them to determine if there is an open one for this ds when instead this problem could easily be solved by fixing the order of operations. check out the current code:

`				if (options.status.id == $scope.COMPLETE) {
					deliveryServiceService.updateDeliveryService(deliveryService).
						then(
							function() {
								$state.reload(); // reloads all the resolves for the view
								messageModel.setMessages([ { level: 'success', text: 'Delivery Service [ ' + deliveryService.xmlId + ' ] updated' } ], false);
								createDeliveryServiceUpdateRequest(dsRequest, options.comment, true);
							},
							function(fault) {
								$anchorScroll(); // scrolls window to top
								messageModel.setMessages(fault.data.alerts, false);
							}
						);
				} else {
					createDeliveryServiceUpdateRequest(dsRequest, options.comment, false);
				}
`

mitchell852: so i've been thinking about this a bit more. the way you are performing this check is rather expensive. you are fetching all the DS requests (open or closed) and then looping thru them to determine if there is an open one for this ds when instead this problem could easily be solved by fixing the order of operations. check out the current code:

```if (options.status.id == $scope.COMPLETE) {
					deliveryServiceService.updateDeliveryService(deliveryService).
						then(
							function() {
								$state.reload(); // reloads all the resolves for the view
								messageModel.setMessages([ { level: 'success', text: 'Delivery Service [ ' + deliveryService.xmlId + ' ] updated' } ], false);
								createDeliveryServiceUpdateRequest(dsRequest, options.comment, true);
							},
							function(fault) {
								$anchorScroll(); // scrolls window to top
								messageModel.setMessages(fault.data.alerts, false);
							}
						);
				} else {
					createDeliveryServiceUpdateRequest(dsRequest, options.comment, false);
				}```

The order is simply backwards. Currently, it tries to update the ds (which succeeds) and THEN it tries to create a ds request (which fails because there is already one open).

So if  you simply flip the order around and only make it update the ds IF the creation of the ds request succeeds, then problem solved.

File: traffic_portal/test/end_to_end/DeliveryServiceRequests/delivery-service-requests-spec.js [hunk 2]
```diff
@@ -144,6 +144,7 @@ describe('Traffic Portal Delivery Service Requests', function() {
 		pageData.deleteButton.click();
 		pageData.confirmWithNameInput.sendKeys(mockVals.xmlId + ' request');
 		pageData.deletePermanentlyButton.click();
+		browser.sleep(250);
 		expect(browser.getCurrentUrl().then(commonFunctions.urlPath)).toEqual(commonFunctions.urlPath(browser.baseUrl)+"#!/delivery-service-requests");
```
ocket8888: What's the problem being solved here? A comparison before the promise resolves? Or is `browser.getCurrentUrl()` returning the wrong value if you don't wait a bit first? If it's the former I'm gonna open an issue to get this stuff converted to `async`/`await`, because that'd make waiting for all these things much easier.

alexluckerman: Based off the test I was working on this morning, it seems to me that `expect()` will wait until promises are resolved before performing its comparisons, so the issue is just `browser.getCurrentUrl()` needs a bit of wait before it returns the correct value. But I couldn't find anything to corroborate that online, so ¯\\\__(ツ)__/¯

traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
alexluckerman: Thanks, that makes things a lot easier

traffic_portal/test/end_to_end/DeliveryServiceRequests/delivery-service-requests-spec.js
ocket8888: As far as I could tell from their terrible docs, it doesn't do anything special with Promises. I guess I can always just try it, and if I'm wrong things should randomly break.

traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
```diff
@@ -201,18 +197,16 @@ var FormEditDeliveryServiceController = function(deliveryService, origin, type,
 				// if the user chooses to complete/fulfill the update request immediately, the ds will be updated and behind the
```
mitchell852: can you rewrite this comment? it's not true anymore.

traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
```diff
@@ -61,44 +61,40 @@ var FormEditDeliveryServiceController = function(deliveryService, origin, type,
 
 			// if the user chooses to complete/fulfill the delete request immediately, the ds will be deleted and behind the
```
mitchell852: can you update this comment as well. it's backwards.

mitchell852:
@alexluckerman - is this the command you are running to run the single ds request UI test?

`protractor conf.js --specs DeliveryServiceRequests/delivery-service-requests-spec.js` 

because i'm getting

`11 specs, 11 failures`

alexluckerman:
@mitchell852 I've been using `protractor conf-delivery-service-requests.js`, I believe it's in a separate file since DS requests are an optional feature in TP

mitchell852:
This change also fixed #3512

PR #3733, opened 2019-07-17
CDN-in-a-Box now caches Carton dependencies by default

ocket8888:

This moves some stuff around in the CDN-in-a-Box Traffic Ops (Perl) Dockerfile. It copies the source `traffic_ops/app/cpanfile` file into the container before importing the RPM, and will download Traffic Ops's Carton dependencies before installing it. This means that changes to the Traffic Ops RPM won't trigger downloading Carton dependencies, thus significantly reducing the time it takes to rebuild the image after making changes. It also pre-installs the dependencies currently listed in the Traffic Ops spec file to make that installation a little faster.

This _does_ make the cached layers of the build bigger, but the resulting image should not actually be any larger. Hopefully, for frequent re-builders, the cache space can be easily reclaimed with `docker prune` or similar.

Previously, something similar could be accomplished by keeping a local copy of the dependencies tarballed and gzipped in a special place (`infrastructure/cdn-in-a-box/traffic_ops/local/tmp.tgz`, I think?) and the build process would load that data if it exists. However, in order to make use of that feature, the user must first have knowledge of the install process of Traffic Ops, as well as requiring them to have both a Perl interpreter and Carton and their associated dependencies. This runs counter to the philosophy of self-containment that Docker promotes, and could be difficult for, say, Windows users who want to run CDN-in-a-Box. It also runs somewhat counter to the philosophy of CDN-in-a-Box, in my opinion, in that a user should be able to just "press a button" and have a Traffic Control ecosystem spring to life. So I do believe that changing the system in this way is a worthwhile improvement.

- CDN in a Box

Feature change is an implementation detail - shouldn't be documented. I looked for documentation on the "old" way of using locally-cached dependencies, but I couldn't find anything.

Build and run CDN-in-a-Box to verify that Traffic Ops is properly built/installed and operates as normal. No tests because it only affects CDN-in-a-Box, which has no testing.

Here's what I would do (and did) to test this:

```bash
sudo docker-compose down -v --remove-orphans
sudo docker system prune -af
make
sudo docker-compose build
sudo docker-compose up

# At this point you wanna stop and verify that CiaB is working,
# do a curl of the demo1 DS, look at TP to see if anything stands out, etc.

# Now change some file in `traffic_ops/`; add a newline to a
# Perl script or something. Just to change the RPM hash.

sudo docker-compose down -v --remove-orphans
rm traffic_ops/*.rpm
rm ../../dist/*traffic_ops*.rpm
make traffic_ops/traffic_ops.rpm
sudo docker-compose build trafficops && trafficops-perl # this step should be waaayyy faster than before
sudo docker-compose up

# And again verify that everything works as expected in the resulting environment.
```

### DIFF
--- a/infrastructure/cdn-in-a-box/docker-compose.yml
+++ b/infrastructure/cdn-in-a-box/docker-compose.yml
@@ -78,10 +78,10 @@ services:
   # in place until all API endpoints have been rewritten in Go
   trafficops-perl:
     build:
-      context: .
-      dockerfile: traffic_ops/Dockerfile
+      context: ../..
+      dockerfile: infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
       args:
-        TRAFFIC_OPS_RPM: traffic_ops/traffic_ops.rpm
+        TRAFFIC_OPS_RPM: infrastructure/cdn-in-a-box/traffic_ops/traffic_ops.rpm
     depends_on:
       - db
       - enroller
--- a/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
+++ b/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
@@ -23,6 +23,9 @@
 
 FROM centos:7
 
+EXPOSE 443
+ENV MOJO_MODE production
+
 RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 && \
     rpm --import https://dl.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-7 && \
     rpm --import https://download.postgresql.org/pub/repos/yum/RPM-GPG-KEY-PGDG && \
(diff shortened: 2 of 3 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3978/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3987/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3990/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3991/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4030/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4038/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4059/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4104/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4139/
Test PASSed.

rob05c:
I don't know how I missed this when it came in, but I just want to say, this is awesome. Should make builds drastically faster.

We should considering the Carton vendoring in TO's RPM, if it's possible, so production deployments would get it for free too. Maybe even vendoring it in Git, so even the RPM build is that much faster.

dangogh:
From what I understand, including the Perl modules was considered before this project became open source,  but the legal dept nixed it due to various licenses, etc needing to be considered.  This happened before I joined the project, but that's the history....

PR #3735, opened 2019-07-17
Add TP test for DNSSEC keys

alexluckerman:
I'm not that sure that the way I've written/organized the tests is in accordance with how we want to continue to expand the TP test suite, but this is what I've come up with so far.

In addition to adding DNSSEC / KSK (re)generation tests, I made a small fix to one of the other TP CDN tests.

- Traffic Portal

This is only an addition to existing TP tests, so no documentation or changelog entry is needed

Run the end-to-end tests for Traffic Portal


asfgit:

alexluckerman:
@mitchell852 It could be a result of the wait time not being long enough to allow the link to show up? I just added async/await to the tests, so hopefully that helps

PR #3737, opened 2019-07-17
Change redirect location after creating a parameter or profile

alexluckerman:
A simpler way of fixing the issue than my last PR. However, it does still depend on Restangular being removed

- [x] This PR fixes ##3453

- Traffic Portal

As this is a small UI tweak, I don't believe it needs documentation/changelog update, but please correct me if I'm wrong

Create a new parameter and ensure you are immediately redirected to the link profiles page, and vice versa


asfgit:

mitchell852:
also, can you write 2 UI tests that verifies that the creation of a parameter and a profile takes them to the right page/view?

alexluckerman:
> this change should not depend on the removal of restangular, right?

I thought it wasn't dependent on restangular's removal, but it needs to get the ID from the response in order to navigate to that page, which is the thing restangular was giving me so much trouble with

PR #3742, opened 2019-07-22
Restricted job creation to Portal and above

### LINKED ISSUE #3116: Read-Only users can create invalidation jobs
The endpoint at `api/1.x/user/current/jobs` doesn't check roles if tenancy is enabled - thus any user to whom a Delivery Service is assigned and is also a member of the tenancy which owns said Delivery Service is permitted to create content invalidation jobs. This means that technically, Portal, Steering, Extension, and Monitor users may also create content invalidation jobs, though I think it's more distressing that the ability is given to a user permission group literally named "Read-Only".

ocket8888:
- [x] This PR fixes #3116 

Raises the required permissions level to submit new jobs to "Portal", regardless of tenancy and/or DS-to-user assignment(s).

- Documentation
- Traffic Ops

Build TO, try to create a job as a user with "read-only" Role, both with and without DS assigned to you, both with and without the proper tenant, and verify that the operation is forbidden. Then repeat with Portal Role and verify that under the right conditions, the operation is allowed.

- master (@ 95366ccaf333332990f1b4fe9d10ffd6f603d4ed)
- 3.0.1

### DIFF
--- a/docs/source/api/user_current_jobs.rst
+++ b/docs/source/api/user_current_jobs.rst
@@ -122,7 +122,11 @@ Creates a new content revalidation job.
 .. Note:: This method forces a HTTP *revalidation* of the content, and not a new ``GET`` - the origin needs to support revalidation according to the HTTP/1.1 specification, and send a ``200 OK`` or ``304 Not Modified`` HTTP response as appropriate.
 
 :Auth. Required: Yes
-:Roles Required: "admin" or "operations"\ [1]_
+:Roles Required: "portal"
+
+	.. versionchanged:: ATCv3.0.2
+		For security reasons, the endpoint was reworked so that regardless of tenancy, the "portal" :term:`Role` or higher is required.
+
 :Response Type:  ``undefined``
 
 Request Structure
@@ -186,4 +190,3 @@ Response Structure
 		}
 	]}
 
-.. [1] A role is only required if tenancy is not used; if tenancy is used by Traffic Control, then the user will be able to create the content revalidation job on :term:`Delivery Service`\ s scoped to his or her tenancy regardless of role. This means that **even read-only users can create content invalidation jobs for :term:`Delivery Service`\ s scoped to their tenancy**. This behavior is considered a bug, and it is tracked by `GitHub Issue #3116 <https://github.com/apache/trafficcontrol/issues/3116>`_.
--- a/docs/source/api/user_current_jobs.rst
+++ b/docs/source/api/user_current_jobs.rst
@@ -124,7 +124,7 @@ Creates a new content revalidation job.
 :Auth. Required: Yes
 :Roles Required: "portal"
 
-	.. versionchanged:: ATCv3.0.2
+	.. versionchanged:: ATCv3.1.0
 		For security reasons, the endpoint was reworked so that regardless of tenancy, the "portal" :term:`Role` or higher is required.
 
 :Response Type:  ``undefined``
--- a/traffic_ops/app/lib/API/Job.pm
+++ b/traffic_ops/app/lib/API/Job.pm
@@ -154,6 +154,10 @@ sub get_current_user_jobs {
 sub create_current_user_job {
 	my $self = shift;
 
+	if (!&is_portal($self)) {
+		return $self->forbidden();
+	}
+
 	my $ds_id      = $self->req->json->{dsId};
 	my $regex      = $self->req->json->{regex};
 	my $ttl        = $self->req->json->{ttl};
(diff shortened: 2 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/3997/
Test PASSed.

traffic_ops/app/lib/API/Job.pm
```diff
@@ -181,7 +185,7 @@ sub create_current_user_job {
 			return $self->forbidden("Forbidden. Delivery-service tenant is not available to the user.");
 		}
 	} else {
-		if ( !&is_oper($self) && !$self->is_delivery_service_assigned($ds_id) ) {
+		if ( !&is_portal($self) && !$self->is_delivery_service_assigned($ds_id) ) {
```
mitchell852: this would never resolve to true given the condition on line 157, right? i think i'd rather see this rewritten:

```
if ( $tenant_utils->use_tenancy() ) {
		if ( !$tenant_utils->is_ds_resource_accessible($tenants_data, $ds->tenant_id) ) {
			return $self->forbidden("Forbidden. Delivery-service tenant is not available to the user.");
		}
	} else {
		if ( !&is_oper($self) && !$self->is_delivery_service_assigned($ds_id) ) {
			return $self->forbidden();
		}
	}
```

to

```
		if ( !$tenant_utils->is_ds_resource_accessible($tenants_data, $ds->tenant_id) ) {
			return $self->forbidden("Forbidden. Delivery-service tenant is not available to the user.");
		}
```

docs/source/api/user_current_jobs.rst
```diff
@@ -122,7 +122,11 @@ Creates a new content revalidation job.
 .. Note:: This method forces a HTTP *revalidation* of the content, and not a new ``GET`` - the origin needs to support revalidation according to the HTTP/1.1 specification, and send a ``200 OK`` or ``304 Not Modified`` HTTP response as appropriate.
 
 :Auth. Required: Yes
-:Roles Required: "admin" or "operations"\ [1]_
+:Roles Required: "portal"
+
+	.. versionchanged:: ATCv3.0.2
```
mitchell852: need to figure out if this is going in 3.0.2, 3.1 or 4.0

traffic_ops/app/lib/API/Job.pm
ocket8888: I can't do that, since it would change the API behavior beyond the scope of this security fix. I shouldn't be checking for Role twice, though.

docs/source/api/user_current_jobs.rst
ocket8888: Throwing this in as a patch and bumping the patch number seems easiest IMO.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4037/
Test PASSed.

traffic_ops/app/lib/API/Job.pm
mitchell852: i feel like you can simply live with this change alone:

```
if (!&is_portal($self)) {
	return $self->forbidden();
}
```

because `$tenant_utils->use_tenancy()` never evaluates to false anymore so that change to line 188 does nothing.

ocket8888: That's true... so why is it still a mutable, global parameter?

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4040/
Test PASSed.

traffic_ops/app/lib/API/Job.pm
mitchell852: i thought there was some work done to completely remove the ability to turn off tenancy but i guess there were many pieces that were missed. so imo i'd take this opportunity to remove that `$tenant_utils->use_tenancy()`

rawlinp: The `use_tenancy` parameter might still exist in the global profile, but there shouldn't be any code that actually checks that parameter anymore. Nobody should be adding code that checks that parameter, either. In the Perl, pretty much all the code that checked tenancy used that `$tenant_utils->use_tenancy()` parameter, so I decided to just make that function always return true instead of removing all the places in the Perl that the `use_tenancy()` function was called. I would be -1 on going back and removing all the code in the Perl that checks the `use_tenancy()` function currently, since that would not be worth doing now.

ocket8888: Either way, I think it falls outside the scope of this bug fix

mitchell852: > I would be -1 on going back and removing all the code in the Perl that checks the use_tenancy() function currently, since that would not be worth doing now.

sounds good

docs/source/api/user_current_jobs.rst
mitchell852: looks like we are going to do a 3.1. want to change this to that?

ocket8888: Yep, I'm on it

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4043/
Test PASSed.

PR #3743, opened 2019-07-22
astats plugin memory leak fix for intercept handler continuation mutex.

traeak:

This PR fixes a memory leak in the astats plugin where a new mutex created by a continuation
that is passed to TSHttpTxnIntercept is never cleaned up.  Solution is to pass a handle to the
source transaction's mutex.  Patch is ported from ATS stats_over_http, inspired by the tls_bridge plugin which uses this technique (and was written by AMC).

- Traffic Stats

Build this astats_over_http.so plugin, deploy to ATS.  Run ATS as normal.  Making a request against the _astats endpoint exercises the fix.  I used valgrind against ATS to verify that a leak existed before and that the leak no longer persists afterwards.

Currently all current astat plugin versions of Traffic Ops contains this bug.  The bug exists in other plugins in ATS 7, 8 and master.


asfgit:

PR #3747, opened 2019-07-23
Adds the configuration parameter 'wrap_counters' to the  astats_over_http plugin.

jrushford:
This PR adds a configuration parameter to the traffic server astats_over_http plugin.  The parameter 'wrap_counters' may be set to 'true' in the astats.config to insure that int64 type counters and signed integers do not present negative values once they increment beyond the INT64_MAX range.  The impetus for this this is to help in addressing the remap stats bug in traffic server.

When traffic server core dumps, the remap stats indexes used to lookup particular counters in trafficserver become associated with the wrong stats due to a race condition that is triggered when traffic server is slammed with requests after it is restarted following a crash.  To mitigate this problem so that it does not happen, the remap stats plugin can be configured to create persistent statistics.  When this is done, all remap stats are loaded and created from persistent store in a controlled manner which avoids the race.  remap stats may be configured to create persistent stats using it's -p option in the plugin.config file.  When using this option, remap statistics will eventually wrap and show negative values once the INT64_MAX value is surpassed.  To prevent these negative values, the 'wrap_counters=true' parameter will cause the counters values to begin again at 0 and resume counting from there keeping the values positive in the INT64_MAX range.

traffic_server astats_over_http plugin

The astats.config.example file has been updated as part of this PR to document the purpose and provides a 'wrap_counters' example.

- CDN in a Box
- Documentation
- Grove
- Traffic Control Client 
- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Traffic Portal
- Traffic Router
X Traffic Stats
- Traffic Vault

Add the 'wrap_counters=2' parameter in the astats.config file.  When a counter increments beyond INT64_MAX (9223372036854775807), the counter will wrap and start a 0 again rather than becoming negative.

### DIFF
--- a/traffic_server/plugins/astats_over_http/astats.config.example
+++ b/traffic_server/plugins/astats_over_http/astats.config.example
@@ -10,3 +10,10 @@ allow_ip6=::1/64
 # hexidecimal int (without 0x), each bit (1 through 7) representing a record type to return.  0x100 is for the built in system stats.  
 # This can be overriden from the url, e.g.: http://localhost/_astats?record.types=100
 record_types=144
+
+# wrap_counters
+# wrap_counters when set to true, will keep the signed int64_t values within the range of INT64_MAX so that they remain positive
+# and do not wrap to become negative in value.  Instead when they exceed INT64_MAX, they will start at 0 and begin counting
+# from there.  This affects the TS_RECORDDATATYPE_COUNTER and TS_RECORDDATATYPE_COUNTER
+# By default, wrap_counters is 'false'
+wrap_counters = true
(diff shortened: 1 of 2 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4029/
Test PASSed.

ocket8888:
How are we getting negative integers printed when formatting a `uint64_t` with PRIu64? that should be impossible

jrushford:
@ocket8888 you are right, TS_RECORDDATATYPE_COUNTER and TS_RECORDDATATYPE_INT are signed int64_t but I missed the formatting with the PRIu64 which takes care of it.  So, this PR is unnecessary.  Thanks for pointing it out.

PR #3748, opened 2019-07-23
Add initial origins and users TP tests

alexluckerman:
Looking for feedback on new TP tests before I push ahead with creating more of these tests

- CDN in a Box
- Documentation
- Grove
- Traffic Control Client 
- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Traffic Portal
- Traffic Router
- Traffic Stats
- Traffic Vault


asfgit:

traffic_portal/test/end_to_end/users/users-spec.js
```diff
@@ -0,0 +1,73 @@
...
+        };
+    };
+    const myNewUser = newUser();
+    const myOtherNewUser = newUser();
+
+    it('should register a new user', async () => {
+        await browser.setLocation('users');
+        await pageData.registerUserButton.click();
+        await pageData.email.sendKeys(myNewUser.email);
+        commonFunctions.selectDropdownbyNum(pageData.role, 3);
```
mitchell852: can you select by tenant name / role name instead? by id always makes me nervous because if for some reason the id is different (or not there), the test will break. by name makes me think the tests will be less brittle...and brittle tests are never good.

traffic_portal/test/end_to_end/origins/origins-spec.js
```diff
@@ -0,0 +1,65 @@
...
+    const myNewOrigin = {
+        name: "origin-" + commonFunctions.shuffle('abcdefghijklmnopqrstuvwxyz0123456789'),
+        fdqn: "fake.origin.example.com"
+    }
+
+    it('should create a new origin', async () => {
+        browser.setLocation('origins');
+        await pageData.createOriginButton.click();
+        await pageData.name.sendKeys(myNewOrigin.name);
+        commonFunctions.selectDropdownbyNum(pageData.tenant, 1);
```
lbathina: instead of index/number - selecting by option value will be good. So when the element index changes we do not have to worry about anything to fix in script

traffic_portal/test/end_to_end/origins/origins-spec.js
```diff
@@ -0,0 +1,65 @@
...
+        commonFunctions.selectDropdownbyNum(pageData.ds, 1);
+        await pageData.createButton.click();
+        expect(element(by.css('.alert-success')).isPresent()).toBe(true);
+        expect(browser.getCurrentUrl().then(commonFunctions.urlPath)).toEqual(commonFunctions.urlPath(browser.baseUrl) + "#!/origins");
+    });
+
+    it('should update an existing origin', async () => {
+        browser.setLocation('origins');
+        await pageData.searchFilter.clear().sendKeys(myNewOrigin.name);
+        await element.all(by.repeater('o in ::origins')).get(0).click();
```
lbathina: may be a comment would make it clear what is done in this statement.

traffic_portal/test/end_to_end/origins/origins-spec.js
```diff
@@ -0,0 +1,65 @@
...
+        await pageData.fqdn.clear().sendKeys('updated.' + myNewOrigin.fdqn);
+        await pageData.updateButton.click();
+        expect(element(by.css('.alert-success')).isPresent()).toBe(true);
```
lbathina: is element defined in pd? its easier for maintenance when its in page object.

traffic_portal/test/end_to_end/origins/origins-spec.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+var pd = require('./pageData.js');
+var cfunc = require('../common/commonFunctions.js');
+
+describe('Traffic Portal Origins Test Suite', function() {
+    const pageData = new pd();
+    const commonFunctions = new cfunc();
+    const myNewOrigin = {
+        name: "origin-" + commonFunctions.shuffle('abcdefghijklmnopqrstuvwxyz0123456789'),
```
lbathina: we could simply use uuid ?? `uuid.New().String()`  from `github.com/google/uuid`

traffic_portal/test/end_to_end/users/users-spec.js
lbathina: agree with jeremy. selecting by index isn't recommended unless it is a requirement to have specific label tied to specific index.

traffic_portal/test/end_to_end/users/users-spec.js
```diff
@@ -0,0 +1,73 @@
...
+        commonFunctions.selectDropdownbyNum(pageData.tenant, 1);
+        await pageData.sendRegistration.click();
+        //expect(browser.getCurrentUrl().then(commonFunctions.urlPath)).toEqual(commonFunctions.urlPath(browser.baseUrl) + "#!/users");
```
lbathina: we still need to validate expected after the action is complete

traffic_portal/test/end_to_end/CDNs/cdns-spec.js
```diff
@@ -23,96 +23,93 @@ var cfunc = require('../common/commonFunctions.js');
...
 		expect(pageData.expirationDate.getAttribute('value')).toEqual('');
 		await pageData.generateDnssecKeysButton.click();
 		await pageData.regenerateButton.click();
 		expect(pageData.confirmButton.isEnabled()).toBe(false);
-		await pageData.confirmInput.sendKeys(myNewCDN);
+		await pageData.confirmInput.sendKeys(myNewCDN.name);
 		expect(pageData.confirmButton.isEnabled()).toBe(true);
 		await pageData.confirmButton.click();
 		const expirationDate = pageData.expirationDate.getAttribute('value').then((expir) => {return Date.parse(expir + ' UTC');});
 		const calculatedExpirationDate = Date.now() + 365*24*60*60*1000;
```
lbathina: A comment on this logic would be nice

traffic_portal/test/end_to_end/DeliveryServiceRequests/delivery-service-requests-spec.js
```diff
@@ -58,13 +58,13 @@ describe('Traffic Portal Delivery Service Requests', function() {
 		expect(pageData.createButton.isEnabled()).toBe(false);
 		pageData.active.click();
 		pageData.active.sendKeys(mockVals.active);
-		commonFunctions.selectDropdownbyNum(pageData.type, 1);
+		await commonFunctions.selectDropdownByNum(pageData.type, 1);
```
lbathina: this still needs to be changed to select by text

traffic_portal/test/end_to_end/physLocations/phys-locations-spec.js
```diff
@@ -31,30 +31,32 @@ describe('Traffic Portal Phys Locations Test Suite', function() {
...
-		pageData.state.sendKeys(myNewPhysLoc.state);
-		pageData.zip.sendKeys(myNewPhysLoc.zip);
-		commonFunctions.selectDropdownbyNum(pageData.region, 1);
+		await pageData.name.sendKeys(myNewPhysLoc.name);
+		await pageData.shortName.sendKeys(myNewPhysLoc.name);
+		await pageData.address.sendKeys(myNewPhysLoc.address);
+		await pageData.city.sendKeys(myNewPhysLoc.city);
+		await pageData.state.sendKeys(myNewPhysLoc.state);
+		await pageData.zip.sendKeys(myNewPhysLoc.zip);
+		await commonFunctions.selectDropdownByNum(pageData.region, 1);
```
lbathina: select by label /text

alexluckerman: The region dropdown depends on what regions are already in the system when the test runs, so we want to keep this as selecting by num rather than label

traffic_portal/test/end_to_end/users/users-spec.js
```diff
@@ -0,0 +1,83 @@
...
+
+    it('should register a new user', async () => {
+        console.log('Registering new user')
+        await browser.setLocation('users');
+        await pageData.registerUserButton.click();
+        await pageData.email.sendKeys(myNewUser.email);
+        await commonFunctions.selectDropdownByLabel(pageData.role, myNewUser.role); // note: this creates a new user with admin permissions
+        await commonFunctions.selectDropdownByLabel(pageData.tenant, myNewUser.tenant);
+        await pageData.sendRegistration.click();
+        // these statements are commented out because I kept getting an internal server error when running this test locally
```
lbathina: is there a bug raised for this issue?

alexluckerman: I'm not really sure what the issue is, could someone uncomment that line and see if the test still works for them?

PR #3758, opened 2019-07-26
Rewrote deliveryservice_stats to Go

### LINKED ISSUE #3187: Tenancy is not being consulted in the deliveryservice_stats endpoint
$r->get("/api/$version/deliveryservice_stats"

https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/Extensions/TrafficStats/API/DeliveryServiceStats.pm#L38

when use_tenancy parameter = 1:

Tenancy should determine one's access to the deliveryservice_stats endpoint  

when use_tenancy parameter = 0:

one's access to the deliveryservice_stats endpoint should be determined by:

a) is this ds assigned to the user? or
b) does the user have the ops or higher role

currently, this endpoint does not consult the use_tenancy parameter and enforce tenancy accordingly if use_tenancy=1.

### LINKED ISSUE #3416: deliverservice_stats endpoint cannot parse ISO date/times with milliseconds
If you make a request to the `deliveryservice_stats` endpoint with `startDate` and/or `endDate` set to ISO 8601 Extended Format-compliant strings with non-zero milliseconds, it will generate an internal server error and barf up a legacy UI page complaining about `cannot set 'id' of undefined` (paraphrased). For example, the URL
```
https://trafficops.domain.test/api/1.4/deliveryservice_stats?interval=60s&metricType=kbps&serverType=edge&deliveryServiceName=dsXML_ID&startDate=2019-03-18T06:00:00.000Z&endDate=2019-03-18T15:02:09.529Z
```
will fail miserably whereas
```
https://trafficops.domain.test/api/1.4/deliveryservice_stats?interval=60s&metricType=kbps&serverType=edge&deliveryServiceName=dsXML_ID&startDate=2019-03-18T06:00:00.000Z&endDate=2019-03-18T15:02:09.000Z
```
works just fine. 

My guess is the server is failing to parse the non-zero millisecond values for some reason, which is truly bizarre because it's perfectly happy to _accept_ millisecond values - as long as they're exactly zero. Expected behavior is that arbitrary ISO 8601-format strings should parse without breaking.

ocket8888:

- [x] This PR fixes #3416 and fixes #3187 

Rewrote the `/deliveryservice_stats` endpoint in Go. Also adds a set of functionality in the new go-rfc library for dealing with MIME Types, hopefully to aid with content negotiation in other endpoints.

This has some differences in behavior with the old Perl route, however the things that changed were mainly undocumented.

### Differences
- `deliveryServiceName` is no longer a required parameter. Instead, it is required that the client submit *either* `deliveryServiceName` - which must be an XMLID - *or* `deliveryService` - which may be either an integral, unique identifier or an XMLID.
- The `interval` query parameter is now restricted to strings matching `^\d+[mhdw]$`.
- The `startDate` and `endDate` query parameters now accept their inputs strictly in RFC3339 format (with optional sub-second precision), as a number of nanoseconds since the Unix Epoch, or in the special `tc.TimeLayout` format. Previously they accepted anything compliant with ISO 8601 - which included RFC3339 format (but **only** without sub-second precision, which was a bug) - and the `endDate` parameter could accept the special value `now`. It no longer does.
- Timestamps in data series as output can now be in one of two formats, as specified by the client. By default, the output is in RFC3339 format (without sub-second precision), but optionally they can be presented as the number of nanoseconds since the Unix Epoch. The distinction is made by adding a non-standard parameter to the `application/json` media type in the `Accept` header. The parameter is `timestamp`, and it can have one of two values:
    - `rfc` - indicates that the timestamps should be returned in RFC3339 format. This isn't required, since it matches the default behavior.
    - `unix` - indicates that the timestamps should be returned as a number of nanoseconds since the Unix Epoch.
- Previously, `metricType` values could include the undocumented `out_bytes`, `status_4xx`, and `status_5xx`. These didn't actually work, but they returned a success. They are no longer allowed.
- The undocumented response field `language` is no longer returned.
- The endpoint now respects tenancy, and will feign ignorance of a Delivery Service to which the client is forbidden access on that basis.
- The endpoint also no longer considers Delivery Service-to-user assignments when checking permissions.

- CDN in a Box
- Documentation
- Traffic Ops
- Traffic Portal

Check the output of the `deliveryservice_stats` endpoint against the old Perl version. It won't match exactly even for the same requested range, since I've made an attempt to fix an issue with Influx where it will return data points outside of said range. It doesn't always work, but should improve the reliability.

- [x] This PR includes tests


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4041/
Test PASSed.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -104,6 +105,9 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodPost, `asns/?$`, api.CreateHandler(&asn.TOASNV11{}), auth.PrivLevelOperations, Authenticated, nil},
 		{1.1, http.MethodDelete, `asns/{id}$`, api.DeleteHandler(&asn.TOASNV11{}), auth.PrivLevelOperations, Authenticated, nil},
 
+		// Traffic Stats access
+		{1.2, http.MethodGet, `deliveryservice_stats`, trafficstats.GetDSStats, auth.PrivLevelOperations, Authenticated, nil},
```
mitchell852: this should be 1.1 since you are overwriting the perl entirely. otherwise, 1.2 would serve Go and 1.1 would continue to serve Perl.

mitchell852: is operations really the role enforced in Perl? seems high.

mitchell852: also, at one time this routes.go file was alphabetically organized. not sure what happened to that.

rawlinp: Alphabetical organization might be dangerous since the order matters w.r.t. the API version and regex matches. I.e. the same route for different minor versions needs to have the minor versions in descending order (since a 1.1 request will not match a 1.3 or 1.2 route, but a 1.3 request will match a 1.2 or 1.1 route). There may be some regexes that are overlapping as well, so the _longer_ regexes should come before the shorter regexes so that the best match is taken.

ocket8888: The route doesn't exist prior to 1.2 - if you request it today at 1.1 you'll get a 404.

I'm fairly certain the Perl used Ops, but I agree that that's too high. However, I think I'm changing a lot as it is.

mitchell852: fairly certain? better check. :)

mitchell852: > Alphabetical organization might be dangerous 

oh yeah, i forgot. there are some in there that are very order dependent. but if adding a `d` one, might as well put it  next to the `d`s imo. kinda sucks that they can't be alphabetized somehow, gives me a headache searching thru for stuff that i forgot the name to.

mitchell852: @ocket8888 - this is the current check i guess from DeliveryServiceStats.pm

```
			if ( $self->is_delivery_service_name_assigned($ds_name) || &is_admin($self) || &is_oper($self) ) {
```
if it's assigned to your or you're admin or ops. (not sure why it doesn't just check ops...but anyhow)

so this is broken and i think you should fix it. the real check should be:
1. do you pass the tenancy check?
2. are you readonly or above?

mitchell852: so basically the way it is implemented now it is asking "is this ds assigned to you" and if so, you are good to go. but that behavior should have been changed a long time ago to use tenancy instead.

in which case NO role was checked but i think R/O makes perfect sense here.

ocket8888: I agree, but the goal of a rewrite is usually to change as little as possible, so if anything I almost feel like I should check for a user-ds assignment instead.

rawlinp: I smell a tech debt issue in your future :)

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4044/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4046/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4081/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4167/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4168/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4172/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4173/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4174/
Test FAILed.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+
+	if ex, ok := i.Params["exclude"]; ok {
+		switch ex {
+		case "series":
+			c.ExcludeSeries = true
+		case "summary":
+			c.ExcludeSummary = true
+		default:
+			e = errors.New("Invalid exclude! Must be 'series' or 'summary'")
```
mhoppa: should this have been a function like orderableFromString? one used switch in place and returns options to user and the other is func call and doesnt return options to user

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+	var c config
+	if c, userErr, errCode = configFromRequest(r, inf); userErr != nil {
+		sysErr = fmt.Errorf("Unable to process deliveryservice_stats request: %v", userErr)
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+
+	client, err := inf.CreateInfluxClient()
+	if err != nil {
+		if err == sql.ErrNoRows {
```
mhoppa: tbh from a caller point of view it is really weird that sql.ErrNoRows represents no servers from a function called `CreateInfluxClient` and also should this be a 500 ? 503 to me means the api is no available not another service the service relies on

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+			errCode = http.StatusInternalServerError
+			userErr = nil
+			sysErr = err
+		}
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	} else if client == nil {
+		userErr = errors.New("Traffic Stats is not configured!")
+		sysErr = userErr
+		errCode = http.StatusServiceUnavailable
```
mhoppa: again i think this should be a 500

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+	defer (*client).Close()
+
+	var dsTenant uint
+	if dsTenant, err = dsTenantIDFromXMLID(c.DeliveryService, tx); err != nil {
+		sysErr = err
+
+		if err == sql.ErrNoRows {
```
mhoppa: I wonder if dsTenantIDFromXMLID should return `exists, dsTenant, err` instead of bubbling up a mysql error

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+				errBuilder.WriteRune(':')
+				errBuilder.WriteString(m.Text)
+				errBuilder.Write([]byte(", "))
+			}
+			errBuilder.WriteRune(']')
+		}
+
+		if err != nil {
+			sysErr = fmt.Errorf("Getting summary response from Influx: %v - %s", err, errBuilder.String())
+			api.HandleErr(w, r, tx, http.StatusBadGateway, nil, sysErr)
```
mhoppa: I think this should be 500

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
```
mhoppa: 500

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+		log.Debugf("Messages from summary query: %s", errBuilder.String())
+		errBuilder.Reset()
+		messages = nil
+
+		resp.Response.Summary = &summary
+	}
+
+	if !c.ExcludeSeries {
+		series, messages, err := getSeries(client, &c, inf.Config.ConfigInflux.DSDBName)
+		if messages != nil && len(messages) > 0 {
```
mhoppa: could this be moved to an util being used more then once

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+			}
+		}
+	}
+
+	code = http.StatusOK
+	return
+}
+
+func GetDSStats(w http.ResponseWriter, r *http.Request) {
+	// Perl didn't require "interval", but it would only return summary data if it was not given
```
mhoppa: nit but this method is pretty large I wonder if it should be split up?

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
+package trafficstats
```
mhoppa: Is there any unit tests that can be written for this file? 

Like validation on the config or authorized to query endpoint?

lib/go-tc/traffic_stats.go
```diff
@@ -0,0 +1,148 @@
...
+	case OneMonth:
+		return OneMonth
+	}
+	return InvalidDuration
+}
+
+// Seconds returns the number of seconds to which a TrafficStatsDuration is equivalent.
+// For invalid objects, it returns -1 - otherwise it will always, of course be > 0.
+func (d TrafficStatsDuration) Seconds() int64 {
+	switch d {
```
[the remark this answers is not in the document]
ocket8888: I could, but why's a map better, though?

lib/go-tc/traffic_stats.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// TrafficStatsResponse represents a response from one of the "Traffic Stats endpoints" of the
+// Traffic Ops API, e.g. `/deliveryservice_stats`.
+type TrafficStatsResponse struct {
+	// Series holds the actual data - it is NOT in general the same as a github.com/influxdata/influxdb1-client/models.Row
+	Series *TrafficStatsSeries `json:"series,omitempty"`
+	// Source has an unknown purpose. I believe this is supposed to name the "plugin" that provided
+	// the data - kept for compatibility with the Perl version(s) of the "Traffic Stats endpoints".
+	// Deprecated: this'll be removed or reworked to make more sense in the future
```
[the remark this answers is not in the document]
ocket8888: I think "Deprecated" is standard in GoDocs, but I could add a TODO under it or above it or something. Just that it'd say the same thing twice. IMO, it'd be better to open an issue about it, as that's the standard way of tracking things that need to be changed or fixed.

lib/go-tc/traffic_stats.go
```diff
@@ -0,0 +1,148 @@
...
+	Count                  uint    `json:"count"`
+	FifthPercentile        float64 `json:"fifthPercentile"`
+	Max                    float64 `json:"max"`
+	Min                    float64 `json:"min"`
+	NinetyEighthPercentile float64 `json:"ninetyEighthPercentile"`
+	NinetyFifthPercentile  float64 `json:"ninetyFifthPercentile"`
+	// TotalBytes is the total number of bytes served when the "metric type" requested is "kbps"
+	// (or actually just contains "kbps"). If this is not nil, TotalTransactions *should* always be
+	// nil.
+	TotalBytes *float64 `json:"totalBytes"`
```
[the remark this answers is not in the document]
ocket8888: whenever the requested metric type doesn't contain "kbps".

lib/go-tc/traffic_stats.go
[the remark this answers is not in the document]
mhoppa: not better I personally just like it better for readability and adding new values to going forward. But that is just an opinion hence the nit/question :) leave as is

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -369,6 +382,78 @@ func (inf *APIInfo) Close() {
...
+//
+// If Influx connections are not enabled, this will return `nil` - but also no error. It is expected
+// that the caller will handle this situation appropriately.
+func (inf *APIInfo) CreateInfluxClient() (c *influx.Client, e error) {
+	if !inf.Config.InfluxEnabled {
+		return
+	}
+
+	var fqdn string
+	var TCPPort uint
```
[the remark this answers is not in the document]
ocket8888: Why? `go fmt` doesn't care

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -369,6 +382,78 @@ func (inf *APIInfo) Close() {
...
+		if !HTTPSPort.Valid {
+			log.Warnf("INFLUXDB Server %s has no secure ports, assuming default of 8086!", fqdn)
+			HTTPSPort = sql.NullInt64{8086, true}
+		}
+		host = fmt.Sprintf(host, "s", fqdn)
+	} else {
+		host = fmt.Sprintf(host, "", fqdn)
+	}
+
+	if useSSL {
```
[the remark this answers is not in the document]
ocket8888: yeah, I just thought it might get ugly. But I guess the proper solution to that is to break the logic into smaller `func`s, not checking the same condition twice.

traffic_ops/traffic_ops_golang/api/api.go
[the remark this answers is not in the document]
mhoppa: just the convention I have seen for variable names in golang

traffic_ops/traffic_ops_golang/config/config.go
```diff
@@ -122,6 +125,14 @@ type ConfigLDAP struct {
 	LDAPTimeoutSecs int    `json:"ldap_timeout_secs"`
 }
 
+type ConfigInflux struct {
+	User string `json:"user"`
+	Password string `json:"password"`
+	DSDBName string `json:"deliveryservice_stats_db_name"`
+	CacheDBName string `json:"cache_stats_db_name"`
+	Secure *bool
```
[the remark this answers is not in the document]
ocket8888: Not the way it's currently written :P I'm on it, though.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+import "github.com/apache/trafficcontrol/lib/go-rfc"
+import "github.com/apache/trafficcontrol/lib/go-log"
+import "github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
+import "github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/tenant"
+
+import influx "github.com/influxdata/influxdb1-client/v2"
+
+const DEFAULT_INTERVAL = tc.TrafficStatsDuration("1m")
+
+var metricTypes = map[string]interface{}{
```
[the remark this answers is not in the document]
ocket8888: Eh. I think grouping constants/variables when they take up more than one line individually makes them collectively hard to read. However, since they're all internal and don't warrant either group or individual documentation I don't feel strongly about it.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+	"application/json",
+	map[string]string{"timestamp": "rfc"},
+}
+
+var jsonWithUnixTimestamps = rfc.MimeType{
+	"application/json",
+	map[string]string{"timestamp": "unix"},
+}
+
+type APIResponse struct {
```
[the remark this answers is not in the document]
ocket8888: I thought about doing something like that, but do you think it's better inside of, say, `traffic_ops/traffic_ops_golang/api` or in `lib/go-tc`?

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
```diff
@@ -0,0 +1,691 @@
...
+		o = timeOrder
+	case "value":
+		o = valueOrder
+	default:
+		return nil
+	}
+	return &o
+}
+
+type config struct {
```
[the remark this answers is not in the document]
ocket8888: That's a good idea.

There probably should be a client implementation, but that's not strictly part of a rewrite. Normally I do that, but this is already a large, complex rewrite that's been awaiting review for nearly two months. I didn't want to pile on.

lib/go-rfc/mimetype_test.go
```diff
@@ -0,0 +1,359 @@
+package tc
```
mhoppa: package rfc - this is what is causing the build failure i believe

traffic_ops/traffic_ops_golang/api/api.go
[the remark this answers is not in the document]
rob05c: Yeah, local variables should start with lowercase, and be consistent-cased: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

lib/go-rfc/mimetype_test.go
ocket8888: whoops

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -369,6 +382,78 @@ func (inf *APIInfo) Close() {
 	}
 }
 
+// CreateInfluxClient onstructs and returns an InfluxDB HTTP client, if enabled and when possible.
+// The error this returns should not be exposed to the user; it's for logging purposes only.
+//
+// If Influx connections are not enabled, this will return `nil` - but also no error. It is expected
+// that the caller will handle this situation appropriately.
+func (inf *APIInfo) CreateInfluxClient() (c *influx.Client, e error) {
```
rob05c: I don't know that we have an official stance as a project, but I think the majority of the Go community discourages named returns. It makes it non-obvious what's being returned where. You could set `e` anywhere, and `return` anywhere else, and then it's hard for a reader to follow that and figure out what `e` and `c` are being returned at any given `return`.

I won't -1, if you feel strongly about it, but I do think it hinders readability, and it's better to avoid, so future readers can easily see exactly what's being returned at every point.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: Yes, that should be consistent, and they should both return options.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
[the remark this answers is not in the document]
rob05c: Two other options to consider:

I think we started leaning toward anonymous structs for responses, in the Client. It's just so repetitive to define them everywhere, for everything.

```
	resp := struct{Response tc.TrafficStatsResponse `json:"response"`}{
		tc.TrafficStatsResponse{
			Source:  tc.TRAFFIC_STATS_SOURCE,
			Version: tc.TRAFFIC_STATS_VERSION,
			Series:  nil,
			Summary: nil,
		},
	}
```

Or, I'm not sure if we already have one, but a helper func in `lib/go-tc` might be good, that we could then use everywhere:

```
func MakeResponse(v interface{}) interface{} {
  return struct{Response tc.TrafficStatsResponse `json:"response"`}{v}
}
```

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: The HTTP spec seems to indicate that 503 errors are for scheduled maintenance and unplanned overloads, so I agree it's not ideal. I just hesitate to send back a 500 because that tends to make people think "there's something wrong with this endpoint, a 500 shouldn't happen". But I don't have a better idea, I suppose.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: that would be more consistent with other code in the codebase.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: Again I think a 502 is more appropriate here:

> _"The 502 (Bad Gateway) status code indicates that the server, while acting as a gateway or proxy, received an invalid response from an inbound server it accessed while attempting to fulfill the request."_

That perfectly describes this situation

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: Sure

traffic_ops/traffic_ops_golang/api/api.go
ocket8888: I don't really care at all, just seems like the thing to do when your first action would otherwise be to declare the things you're returning.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: I could write some negative tests for the format of the request, but I don't handle user permissions in the route handler, and a unit test can't handle a connection to an external InfluxDB server; that's an integration test.

traffic_ops/traffic_ops_golang/api/api.go
mitchell852: just something i read: "Naked return statements should be used only in short functions. They can harm readability in longer functions."

this function doesn't look that short...

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
mhoppa: is it ? do we return 502 on mysql errors? The user is not really asking you to proxy the request anywhere or understand if you are or not. Per the API contract the user is just asking for stats on delivery services and expects a certain return structure.

I have not worked with or on an API when it returns a 502 when it gets a bad response or can not connect to a backend db

ocket8888: The Traffic Ops database is a part of traffic ops, so failure to communicate with it is considered a failure of the Traffic Ops server. But Traffic Stats is not just an implementation detail of Traffic Ops, it's a documented Traffic Control component in its own right, with its own API. The user of the API does, in fact, know that the data is retrieved this way because the response body says so, and because that fact is documented.

mhoppa: I still think 502 is very very weird here and disagree with it. But Ill defer to the majority of the team though what we should do. end of the day something broke and its a 5xx error

rob05c: I think any server error in TO needs to be a 500 with no additional information, for security reasons. We don't currently have Self Service and random people on the internet logging in and making DSes, but as we move to Self Service, this will be a necessity. Better to do it now, than have to go back and change it later.

When there's a server-side error (not caused by client input, i.e. a 4xx), attackers can use that information to better craft an attack. For example, suppose Influx is dying (alas, not uncommon). If an attacker sees a 502 from this endpoint, they can guess that, and DOS just this endpoint, to overwhelm Influx even worse. Or, they can DOS this endpoint when it's working correctly until they get a 502, and discover exactly how many requests/second are required to DOS Influx.

The important thing is that the right person can find the information to fix the problem. Which they can: it's in the log. For a user error, we give the user as much information as possible about what they did wrong. For a server error, we give the user as little information as possible in case they're an attacker, and log as much as possible, so a good user can call Ops who can then find and fix the issue. If an Ops person encounters the error, they immediately look in the log.

ocket8888:
It could, and I'm leaning away from doing that - someone else suggested it. I really think a regex would suffice - though nobody on the mailing list seemed to care. Figured I'd wait to see if anyone asked about it during the review.

traffic_ops/traffic_ops_golang/api/api.go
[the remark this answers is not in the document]
ocket8888: That section indicates initialisms and abbreviations can be uppercased. I'll change `fqdn` to match.

If conventions like this are supposed to be followed, we should have a mechanism for checking them: some kind of linter.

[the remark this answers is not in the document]
mhoppa: yes we should definitely be running a linter 💯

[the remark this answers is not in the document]
rob05c: >That section indicates initialisms and abbreviations can be uppercased. I'll change fqdn to match.

They can, but local variables should always start with lowercase, hence public and global symbols starting with initialisms should be e.g. `URLFoo`, and private and local should be `urlFoo`.

Local variables should always start with lowercase, because Go exports symbols starting with uppercase, and does not export lowercase. Of course, a local variable in a function can't be viewed from an outside scope. So it makes it very odd and confusing for a local var to be uppercase="exported." If nothing else, if a future reader sees that symbol uses further down in the func, and doesn't notice it's local, they're going to assume it's an exported global variable and be confused.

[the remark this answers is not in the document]
ocket8888: I'll change it because I don't really care that much - though for my money "url" is nonsense that doesn't mean anything while "URL" stands for something - but I don't think it's fair to require a convention not agreed-upon by or enforced mechanically by the project. If local variables should start with a lowercase letters, there should be a linter configuration introduced to the repository and approved on the mailing list.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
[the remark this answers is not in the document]
ocket8888: I made it an anonymous struct.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: _Any_ server error needs to be a 500? Or just things that I'd otherwise want to be 502s? Either way, there'll need to be an issue created for making the rest of TO compliant with that.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: I've moved some of the logic out into utilities, how's it look now?

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
rob05c: Any. And yes, TO used to be compliant, but I'm sure people have added things since.
And not just 500, but specifically not sending any info. I'm sure we also have endpoints returning info about server errors to users, which also need fixed.

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
mhoppa: much better!!!

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
ocket8888: #3946

traffic_ops/traffic_ops_golang/trafficstats/trafficstats.go
mhoppa: sure! 👍

docs/source/api/deliveryservice_stats.rst
```diff
@@ -37,140 +33,184 @@ Request Structure
...
-	+----------------------------+---------------+-----------------------------------------------------------------------------------------+
-	|``>>time``                  | string        |                                                                                         |
-	+----------------------------+---------------+-----------------------------------------------------------------------------------------+
-	|``>>value``                 | float         |                                                                                         |
-	+----------------------------+---------------+-----------------------------------------------------------------------------------------+
-
-.. code-block:: json
+:series: An object containing the actual data series and information necessary for working with it.
+
+	:columns: This is an array of names of the columns of the data contained in the "values" array - should always be ``["time", "value"]``
```
mhoppa: I am getting sum_count returned from the perl version instead of value

ocket8888: Yeah, that's how it used to be in my version too. I decided to make it `value` instead, since `sum_count` is both puzzling to the user, as well as an exposure of implementation detail.

At the time it didn't occur to me that that'd be a breaking API change. But it would be...

ocket8888:
> _"3. In golang no query section is returned but in perl there is"_

That exposed a harmful amount of implementation detail to the user, and was totally undocumented. It serves no purpose, and its removal is deliberate.

ocket8888:
retest this please

ocket8888:
retest this please

PR #3763, opened 2019-07-30
Oauth authentication rebase

mattjackson220:
Sets up integration with OAuth providers to allow Single Sign On instead of username/password login to Traffic Portal.  It defaults to disabled.  Once OAuth is set up, login will follow these steps:
- User will go to TP login page as usual, for example at tp.domain.com.  If OAuth is enabled, a single button will be visible to login with SSO.
- When button is clicked, it will redirect to the OAuth provider URL from the traffic_portal_properties.json.
- OAuth provider will authenticate the user using SSO.  If user is not logged in, they will be redirected to SSO login page.  If user is logged in, they will be redirected to tp.domain.com/#!/sso?auth_token=encryptedTokenFromOAuthProvider
- the /sso page will parse the token and POST to the API /user/login/oauth endpoint
- the API /user/login/oauth endpoint will decode and validate the token, cross reference the Json Key Set URL against the whitelisted URLs in the cdn.conf file, query the database to get the user's role, and if all of that is successful, return a cookie.
- If the login is successful, the user will be redirected to the page they were trying to see.  If login was unsuccessful, a 401 error will be returned and the user will be redirected back to the login page.

- Documentation
- Traffic Ops
- Traffic Portal

To test using CDN In A Box:
- Update config.sh to : 
        "whitelisted_oauth_urls": [
    		"insert domain for your expected Json Key Set returned by oAuth provider"
	]
- Update traffic_portal_properties.json to: 
         "oAuth": {
  		"_comment": "Opt-in OAuth properties for SSO login",
 		 "enabled": true,
 		 "oAuthUrl": "insert your oAuth provider URL",
  		"oAuthTokenQueryParam": "insert your oAuth provider's token query parameter"
	}
- Run CDN in a box
- Verify new login page only shows button to login with SSO
- Click button with Network panel open and verify that it makes call to OAuth provider
- Verify that login failed
- Login to db container: 
        docker exec -it cdn-in-a-box_db_1 /bin/bash
- Login to postgres 
        psql -d traffic_ops -U traffic_ops
- Insert your user into db
         insert into tm_user (username, role, tenant_id) values ('yourUserId', (SELECT id FROM role WHERE name = 'admin'), (SELECT id FROM tenant WHERE name='root'));
- Click Login button again, Verify call to OAuth provider, Verify that login succeeds
- Log out

- Try to go directly to /#!/servers endpoint
- Verify it requires login and shows SSO login page
- Click Login
- Verify call to OAuth provider, verify login successful, verify it redirects to /#!/servers endpoint

- Update config.sh to : 
        "whitelisted_oauth_urls": []
- Restart Traffic Ops
- Try to log in and verify it returns the following error:
        "Key URL from token is not included in the whitelisted urls. Received: //url"


traffic_portal/app/src/traffic_portal_properties.json
```diff
@@ -201,6 +201,16 @@
...
+    },
+    "oAuth": {
+      "_comment": "Opt-in OAuth properties for SSO login. See http://traffic-control-cdn.readthedocs.io/en/release-4.0.0/admin/quick_howto/oauth_login.html for more details. redirectUriParameterOverride defaults to redirect_uri if left blank.",
+      "enabled": false,
+      "oAuthUrl": "https://oauthProvider.example.com/auth",
+      "oAuthTokenQueryParam": "example_token_key",
+      "redirectUriParameterOverride": "example_redirect_url_key",
+      "clientId": "exampleClient",
+      "oAuthCodeTokenUrl": "https://oauthProvider.example.com/auth/token",
+      "clientSecret": "AbCd123"
```
mitchell852: @jhg03a brought up a good point to me regarding storing secrets in this file. probably not a good idea as this file is "publicly" accessible - you might want to look into adding this secret to https://github.com/apache/trafficcontrol/blob/master/traffic_portal/conf/config.js (which is not publicly accessible) which gets read by server.js when  the server starts up.

ocket8888: I actually think that only TO needs to know the client_secret, so you could just turn `whitelisted_oauth_urls` from an array into a map of the URLs to their associated client_secret. Then some parsing logic would need to change, and just modify how it's accessed in the route handler and you're done.

LICENSE
```diff
@@ -432,3 +432,11 @@ The modern-go/concurrent component is used under the Apache 2.0 license:
 The modern-go/reflect2 component is used under the Apache 2.0 license:
 @vendor/github.com/modern-go/reflect2/*
 ./vendor/github.com/modern-go/reflect2/LICENSE
+
+For the lestrrat-go/jwx component:
+@traffic_ops/traffic_ops_golang/vendor/github.com/lestrrat-go/jwx/*
```
ocket8888: I'm not sure if this is the right place, but it'd be nice to make a note somewhere of the git hash at which the repository was vendored. Ideally they have versioned releases, so you can just say that, but if you cloned them from master, then we probably want to know the actual hash.

docs/source/api/user_login_oauth.rst
```diff
@@ -0,0 +1,81 @@
...
+	Cookie: mojolicious=...
+	Content-Length: 26
+	Content-Type: application/json
+
+	{
+		"authCodeTokenUrl": "https://url-to-convert-code-to-token.com",
+		"code": "AbCd123",
+		"clientId": "oauthClientId",
+		"clientSecret": "clientSecret",
+		"redirectUri": "https://traffic-portal.com/sso",
```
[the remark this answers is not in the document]
mitchell852: no, i think TP might be grumpy about it.

traffic_portal/app/src/traffic_portal_properties.json
mattjackson220: i like the idea of moving the client secret into TO but the whitelisted urls are referencing the public key url so not really related to the client id.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4057/
Test FAILed.

docs/source/api/user_login_oauth.rst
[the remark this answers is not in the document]
ocket8888: I don't think TP ever has to worry about it, though.

traffic_portal/app/src/traffic_portal_properties.json
mattjackson220: moved client secret to TO config

docs/source/api/user_login_oauth.rst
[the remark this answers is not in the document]
mattjackson220: just a typo in the docs, took it out

LICENSE
mattjackson220: done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4067/
Test PASSed.

docs/source/api/user_login_oauth.rst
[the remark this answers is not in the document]
mitchell852: oh, didn't realize this was in the docs...

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4091/
Test PASSed.

traffic_portal/app/src/common/api/AuthService.js
```diff
@@ -51,6 +51,30 @@ var AuthService = function($rootScope, $http, $state, $location, userModel, mess
...
+                    if (redirect !== undefined) {
+                        $location.search('redirect', null); // remove the redirect query param
+                        $location.url(redirect);
+                    } else {
+                        $location.url('/');
+                    }
+                },
+                function(fault) {
+                    messageModel.setMessages(fault.data.alerts, true);
+                    locationUtils.navigateToPath('/');
```
mitchell852: if sso fails due to missing user in TO (for example), a 403 is returned (invalid username or password) but line 73 sends them to the dashboard (/) which is a "private" page so TP makes a call to fetch current authenticated user which results in a 401 so you end up with this error message in the UI:

image

so basically that 401 squashed the previous 403 which is the more appropriate message. to fix this, i think you should change this line to 

`locationUtils.navigateToPath('/login');`

^^ the login page is a "public" page that won't attempt to make a call to the api to fetch current authenticated user.

mattjackson220: done

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4092/
Test PASSed.

mitchell852:
i think @ocket8888 did most of the review of this so i'll leave it to him to merge

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4096/
Test PASSed.

PR #3767, opened 2019-08-01
Created a Windows RPM generation script

ocket8888:

Adds a Powershell script equivalent to `pkg` allowing RPMs to be built on Windows

None

This doesn't include documentation because I don't want to commit to supporting this script into the future, and we don't officially support Windows as a development environment or even for running CiaB - although the CiaB docs contain some references to Windows (though I've found it impossible thus far to run CiaB on Windows without a subsystem compatibility layer). It's just something that's there and maybe works, and if it does it makes Windows users' lives a little easier.

Run `pkg.ps1` on a Windows PC with Docker Desktop installed and make sure that the RPMs get generated. If you feel like it - like I said I don't wanna commit to this being guaranteed to work. Tested on Windows 10 with Powershell>=3.0

### DIFF
--- a/pkg.ps1
+++ b/pkg.ps1
@@ -1,0 +1,117 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4077/
Test PASSed.

PR #3768, opened 2019-08-01
Codecov

lbathina:
Adds unit test coverage generation for TO and TM

This doesn't affect the functionality  of the any component. Just exposes the unit test coverage for TO and TM components. Testing, documentation, changelog is not required.

Jenkins job should show coverage information by func
Screen Shot 2019-08-01 at 1 50 12 PM

- [X] This PR does not include a database migration
- [X] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY**

### DIFF
--- a/traffic_monitor/tests/Dockerfile-golangtest
+++ b/traffic_monitor/tests/Dockerfile-golangtest
@@ -23,6 +23,6 @@ WORKDIR /go/src/$DIR/traffic_monitor
 
 RUN bash -c "go get -u github.com/jstemmer/go-junit-report"
 RUN bash -c "go get -v"
-CMD bash -c 'go test -v ./... ../lib/go-tc/... 2>&1 | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml && chmod 777 -R /junit && cat /junit/golang.test.tm.xml'
+CMD bash -c 'go test -coverprofile=tmcoverage.out -v ./... ../lib/go-tc/... 2>&1 | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml && chmod 777 -R /junit && cat /junit/golang.test.tm.xml'
 #
 # vi:syntax=Dockerfile
--- a/traffic_monitor/tests/Dockerfile-golangtest
+++ b/traffic_monitor/tests/Dockerfile-golangtest
@@ -24,5 +24,5 @@ WORKDIR /go/src/$DIR/traffic_monitor
 RUN bash -c "go get -u github.com/jstemmer/go-junit-report"
 RUN bash -c "go get -v"
 CMD bash -c 'go test -coverprofile=tmcoverage.out -v ./... ../lib/go-tc/... 2>&1 | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml && chmod 777 -R /junit && cat /junit/golang.test.tm.xml'
-#
+CMD bash -c 'go tool cover -func=tmcoverage.out'
 # vi:syntax=Dockerfile
--- a/traffic_monitor/tests/Dockerfile-golangtest
+++ b/traffic_monitor/tests/Dockerfile-golangtest
@@ -22,6 +22,7 @@ VOLUME ["/junit"]
 WORKDIR /go/src/$DIR/traffic_monitor
 
 RUN bash -c "go get -u github.com/jstemmer/go-junit-report"
+RUN bash go get "github.com/wadey/gocovmerge"
 RUN bash -c "go get -v"
 ENTRYPOINT ["./tests/gocover.bash"]
 CMD ["./..." , "../lib/go-tc/..."]
--- a/traffic_monitor/tests/Dockerfile-golangtest
+++ b/traffic_monitor/tests/Dockerfile-golangtest
@@ -23,6 +23,6 @@ WORKDIR /go/src/$DIR/traffic_monitor
 
 RUN bash -c "go get -u github.com/jstemmer/go-junit-report"
 RUN bash -c "go get -v"
-CMD bash -c 'go test -coverprofile=tmcoverage.out -v ./... ../lib/go-tc/... 2>&1 | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml && chmod 777 -R /junit && cat /junit/golang.test.tm.xml'
-CMD bash -c 'go tool cover -func=tmcoverage.out'
+ENTRYPOINT ["./tests/gocover.bash"]
+CMD ["./..." , "../lib/go-tc/..."]
 # vi:syntax=Dockerfile
--- a/traffic_monitor/tests/Dockerfile-golangtest
+++ b/traffic_monitor/tests/Dockerfile-golangtest
@@ -22,7 +22,7 @@ VOLUME ["/junit"]
 WORKDIR /go/src/$DIR/traffic_monitor
 
 RUN bash -c "go get -u github.com/jstemmer/go-junit-report"
-RUN bash go get "github.com/wadey/gocovmerge"
+RUN bash -c "go get github.com/wadey/gocovmerge"
 RUN bash -c "go get -v"
 ENTRYPOINT ["./tests/gocover.bash"]
 CMD ["./..." , "../lib/go-tc/..."]
--- a/traffic_monitor/tests/gocover.bash
+++ b/traffic_monitor/tests/gocover.bash
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+set -e
+go get "github.com/wadey/gocovmerge"
+echo "" > result.txt
+packages=( "$@" )
+coverage_out=()
+i=1
+for pkg in ${packages[@]} ; do
+    for d in $(go list $pkg | grep -v vendor); do
+        file="$i.out"
+        go test -v -coverprofile=$file $d >> result.txt
+        cat result.txt
+        if [ -f $file ]; then
+            coverage_out+=( $file )
+        fi
+        ((i++))
+    done
+done 
+gocovmerge ${coverage_out[*]} > coverage.out
+go tool cover -func=coverage.out
+cat result.txt | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml 
+chmod 777 -R /junit && cat /junit/golang.test.tm.xml
--- a/traffic_monitor/tests/gocover.bash
+++ b/traffic_monitor/tests/gocover.bash
@@ -14,20 +14,25 @@
 
 #----------------------------------------
 set -e
-touch result.txt
-packages=( "$@" )
+
 coverage_out=()
 i=1
-for pkg in ${packages[@]} ; do
-    for d in $(go list $pkg | grep -v vendor); do
-        file="$i.out"
-        go test -v -coverprofile=$file $d | tee -a result.txt
-        if [ -f $file ]; then
-            coverage_out+=( $file )
-        fi
-        ((i++))
-    done
+
+for d in $(go list ./... ../lib/... | grep -v vendor); do
+	file="$i.out"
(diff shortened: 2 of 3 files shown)


asfgit:

jhg03a:
ok to test

traffic_monitor/tests/Dockerfile-golangtest
```diff
@@ -23,6 +23,6 @@ WORKDIR /go/src/$DIR/traffic_monitor
 
 RUN bash -c "go get -u github.com/jstemmer/go-junit-report"
 RUN bash -c "go get -v"
-CMD bash -c 'go test -v ./... ../lib/go-tc/... 2>&1 | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml && chmod 777 -R /junit && cat /junit/golang.test.tm.xml'
-#
+CMD bash -c 'go test -coverprofile=tmcoverage.out -v ./... ../lib/go-tc/... 2>&1 | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml && chmod 777 -R /junit && cat /junit/golang.test.tm.xml'
+CMD bash -c 'go tool cover -func=tmcoverage.out'
```
ocket8888: Doing this disables the junit reporting, and actually the testing itself (including generation of `tmcoverage.out`). From [the Dockerfile Reference on `CMD`](https://docs.docker.com/engine/reference/builder/#cmd):

> _"There can only be one CMD instruction in a Dockerfile. If you list more than one CMD then only the last CMD will take effect."_

So what you probably want is just to add this to the original `CMD` with an `&&` or something. If that's getting too complex to read/write/parse then perhaps we need a separate script to use as an `ENTRYPOINT` instead.

traffic_monitor/tests/Dockerfile-golangtest
```diff
@@ -23,6 +23,6 @@ WORKDIR /go/src/$DIR/traffic_monitor
 
 RUN bash -c "go get -u github.com/jstemmer/go-junit-report"
 RUN bash -c "go get -v"
-CMD bash -c 'go test -v ./... ../lib/go-tc/... 2>&1 | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml && chmod 777 -R /junit && cat /junit/golang.test.tm.xml'
-#
```
ocket8888: I think you wanna leave this in. I don't use vim regularly enough to know for sure, but I think this empty comment line may be necessary for it to pick up the following syntax line.

traffic_ops/app/bin/tests/Dockerfile-golangtest
```diff
@@ -20,6 +20,8 @@ ADD lib /go/src/$DIR/lib
 
 WORKDIR /go/src/$DIR/traffic_ops/traffic_ops_golang
 
-CMD bash -c 'go get -v && go test -cover -v ./... ../../lib/go-tc/...'
+CMD bash -c 'go get -v && go test -coverprofile=tocoverage.out  -v ./... ../../lib/go-tc/...'
+CMD bash -c 'go tool cover -func=tocoverage.out'
```
ocket8888: Same as above RE the number of `CMD`s in a Dockerfile

traffic_monitor/tests/Dockerfile-golangtest
lbathina: my bad. Yes that's right.

traffic_ops/app/bin/tests/Dockerfile-golangtest
```diff
@@ -18,8 +18,13 @@ ARG DIR=github.com/apache/trafficcontrol
 ADD traffic_ops /go/src/$DIR/traffic_ops
 ADD lib /go/src/$DIR/lib
 
+COPY traffic_monitor/tests/gocover.bash /go/src/$DIR/traffic_ops/traffic_ops_golang/gocover.bash
+RUN chmod +x /go/src/$DIR/traffic_ops/traffic_ops_golang/gocover.bash
 WORKDIR /go/src/$DIR/traffic_ops/traffic_ops_golang
+RUN bash -c "go get -u github.com/jstemmer/go-junit-report"
+RUN bash -c "go get -v"
```
ocket8888: You're not using any extended features of `bash` here, so there's no reason to force the fork to a shell that may not be in the `PATH` if the image's base is changed (unlikely though that is). You can do all of this in one step after the `WORKDIR` directive to reduce caching as well:

```Dockerfile
WORKDIR /go/src/$DIR/traffic_ops/traffic_ops_golang
RUN chmod a+x gocover.bash && \
    go get -u github.com/jstemmer/go-junit-report && \
    go get -v
```

traffic_monitor/tests/gocover.bash
```diff
@@ -0,0 +1,36 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#----------------------------------------
+set -e
+go get "github.com/wadey/gocovmerge"
```
ocket8888: You should be fetching dependencies that exist for the image's entire run-time during the build step.

But also, has this dependency been approved on a license basis?

traffic_monitor/tests/gocover.bash
```diff
@@ -0,0 +1,36 @@
...
+echo "" > result.txt
```
ocket8888: You don't need to change this, but for future reference I think you're looking for [`touch(1)`](https://linux.die.net/man/1/touch).

traffic_monitor/tests/gocover.bash
```diff
@@ -0,0 +1,36 @@
...
+packages=( "$@" )
+coverage_out=()
+i=1
+for pkg in ${packages[@]} ; do
+    for d in $(go list $pkg | grep -v vendor); do
+        file="$i.out"
+        go test -v -coverprofile=$file $d >> result.txt
+        cat result.txt
```
ocket8888: So now, for every iteration of each loop you're printing the entire results file, which means e.g. you're going to see the first line again on every iteration. Correct me if I'm wrong, but I think what would be better is to print out each line as you go. To do that, you can use [`tee(1)`](http://man7.org/linux/man-pages/man1/tee.1.html) like so:

```bash
go test -v --coverprofile "$file" "$d" | tee -a result.txt # note the quotes around shell variables to prevent interpretation
```

traffic_monitor/tests/gocover.bash
```diff
@@ -0,0 +1,36 @@
...
+        if [ -f $file ]; then
+            coverage_out+=( $file )
+        fi
+        ((i++))
+    done
+done 
+gocovmerge ${coverage_out[*]} > coverage.out
+go tool cover -func=coverage.out
+cat result.txt | go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml 
+chmod 777 -R /junit && cat /junit/golang.test.tm.xml
```
ocket8888: Why is this chmod necessary? I know this is running in a container, but I still balk a little at any time I see a `chmod` with more permissions than it needs - especially with `-R`.

traffic_monitor/tests/gocover.bash
lbathina: I might need some enlightenment here

traffic_monitor/tests/gocover.bash
lbathina: followed how things were already in there. THought it was a special requirement for doing things in open source way!!

traffic_monitor/tests/gocover.bash
lbathina: I just moved the command that was already in there. may be this question was missed in previous review/approval when that statement made to master

traffic_monitor/tests/gocover.bash
ocket8888: lol no

traffic_monitor/tests/gocover.bash
ocket8888: You can ask the mailing list about using this library in this way.

traffic_monitor/tests/gocover.bash
ocket8888: You're probably right. It might be worth checking, though, if everything works if you delete everything before `cat` on that line.

rob05c:
retest this please

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4170/
Test PASSed.

mhoppa:
retest this please

ocket8888:
retest this please

ocket8888:
retest this please

traffic_monitor/tests/gocover.bash
ocket8888: How come you can't do this in the dockerfile when you're `go get`-ing the other stuff you need? That way a built image could be used even offline.

traffic_monitor/tests/gocover.bash
```diff
@@ -0,0 +1,36 @@
...
+i=1
+for pkg in ${packages[@]} ; do
+    for d in $(go list $pkg | grep -v vendor); do
+        file="$i.out"
+        go test -v -coverprofile=$file $d | tee -a result.txt
+        cat result.txt
+        if [ -f $file ]; then
+            coverage_out+=( $file )
+        fi
+        ((i++))
```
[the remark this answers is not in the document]
lbathina: nope it didn't work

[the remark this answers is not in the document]
ocket8888: Yes, besides that fact that `${@[@]}` should just be `$@` you'd still need a temporary file to avoid clobbering yourself. Could do
```bash
tmp="$(mktemp)"
gocovmerge coverage_out "$file" > "$tmp"
mv -f "$tmp" ./coverage_out
```
Which isn't a *whole* lot better - although it is just one extra file per package instead of N for N packages -, so if you don't think it's worth rewriting for that I'm good leaving it as-is.

This'd be a lot easier if `gocovmerge` could take input on stdin.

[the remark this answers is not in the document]
lbathina: gocovermerge doesn't do merge incrementally or from results appended in one file. I tried them first and then got down to what is there now.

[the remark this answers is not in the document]
ocket8888: Yes it does:
```bash
#!/usr/bin/bash
# $GOPATH/src/github.com/apache/trafficcontrol/traffic_monitor/gotestcov.sh
coverage_out=()
i=1

for d in $(go list ./... ../lib/... | grep -v vendor); do
	file="$i.out"
	go test -v -coverprofile="$file" "$d"
	if [ -f "$file" ]; then
		coverage_out+=($file)
	fi
	((i++))
done

gocovmerge ${coverage_out[*]} > coverage.out

touch coverage_out
for pkg in $(go list ./... ../lib/... | grep -v ); do
	tmp="$(mktemp)"
	go test -v -coverprofile covProfile "$pkg" > "$tmp"
	mv -f "$tmp" coverage_out
done
```
then:
```shell
$ chmod a+x ./gotestcov.sh
$ ./gotestcov.sh > /dev/null
$ diff coverage_out coverage.out
$ wc -l coverage.out
3454 coverage.out
```

traffic_monitor/tests/gocover.bash
```diff
@@ -0,0 +1,38 @@
...
+
+gocovmerge ${coverage_out[*]} > coverage.out
+go tool cover -func=coverage.out
+go-junit-report --package-name=golang.test.tm --set-exit-code > /junit/golang.test.tm.xml
+touch coverage_out1
+for pkg in $(go list ./... ../lib/... | grep -v ); do
+	tmp="$(mktemp)"
+	go test -v -coverprofile covProfile "$pkg" > "$tmp"
+	mv -f "$tmp" coverage_out1
+done
```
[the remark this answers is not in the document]
lbathina: @ocket8888 i better leave it to you.

PR #3862, opened 2019-08-12
Add Traffic Ops /deliveryservices, /servers If-Modified-Since support

rob05c:

Adds Traffic Ops /deliveryservices, /servers If-Modified-Since support.

Also adds ETags and If-None-Match support.

Includes API tests.
Includes Changelog.
No documentation; it does technically change the interface, but we don't document headers in the API.

Some notes:
- this is extremely valuable for ORT and cache-side-configs, which will now be able to avoid unnecessary large requests.
- I noticed this doesn't actually work in the Portal. It works in a browser for actual TO, but TP doesn't. It looks like either Node isn't making the request correctly, or Express is stripping the headers somewhere, probably the latter.
- This also fixes some other issues, like the DS query doing a heinously expensive `FULL OUTER JOIN` when it only needs to `LEFT JOIN` (unless I'm mistaken?)
- Also changes the Fixtures to include a DS which uses the mids. Until now, all "fixtures" were DS types that bypassed the mids. This is a critical part of TC that we need more tests for.

- Traffic Ops

Run API tests.

Make a request to a /deliveryservices or /servers endpoint, get the ETag or Last-Modified header, make another request with a If-Modified-Since or If-None-Match header, verify a 304. Make a change to a server or deliveryservice in the request, make a request with a IMS or INM header and verify it does NOT 304 but responds with a 200 and the full object.

Not a bug fix.

### DIFF
--- a/lib/go-rfc/cachecontrol.go
+++ b/lib/go-rfc/cachecontrol.go
@@ -1,0 +1,110 @@
+package rfc
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import (
+	"errors"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+const HdrETag = "ETag"
+const HdrIfNoneMatch = "If-None-Match"
+const HdrIfModifiedSince = "If-Modified-Since"
+const HdrLastModified = "Last-Modified"
+const HdrCacheControl = "Cache-Control"
+
+const ETagVersion = 0
+
+// NotModified is a helper type for users of this library.
+// It allows users to return a NotModified instead of a bool, to make it obvious what the value is.
+type NotModified bool
+
+const IsModified = NotModified(false)
+const IsNotModified = NotModified(true)
+
+// AddModifiedHdrs adds the ETag and Last-Modified headers to the response of w.
+// This must be called before any body is written, of course, or the headers won't be.
+// The lastModified should be the last time this resource was modified.
+func AddModifiedHdrs(w http.ResponseWriter, lastModified time.Time) {
+	w.Header().Set(HdrCacheControl, `max-age=0, must-revalidate`)
+	w.Header().Set(HdrLastModified, FormatHTTPDate(lastModified))
+	w.Header().Set(HdrETag, ETag(lastModified))
+}
+
+// GetModifiedHdr gets the modified time from the ETag or Last-Modified.
+// Returns the modified time, and whether a time was found.
+func GetModifiedHdr(r *http.Request) (time.Time, bool) {
+	// Note we intentionally ignore errors. Logging client errs would be spammy and a potential attack vector, and we want to just return the real object to the client, not an error, if their headers were malformed.
+	if eTag := r.Header.Get(HdrIfNoneMatch); eTag != "" {
+		if t, err := ParseETag(eTag); err == nil {
+			return t, true
+		}
+	}
+	if lastModified := r.Header.Get(HdrIfModifiedSince); lastModified != "" {
+		if t, ok := ParseHTTPDate(lastModified); ok {
+			return t, true
+		}
+	}
+	return time.Time{}, false
+}
+
+// FormatHTTPDate formats t as an RFC7231§7.1.1 HTTP-date.
+func FormatHTTPDate(t time.Time) string {
+	return t.Format(time.RFC1123)
+}
+
+// ETag takes the last time the object was modified, and returns an ETag string. Note the string is the complete header value, including quotes. ETags must be quoted strings.
+func ETag(t time.Time) string {
+	return `"v` + strconv.Itoa(ETagVersion) + `-` + strconv.FormatInt(t.UnixNano(), 36) + `"`
+}
+
+// ParseETag takes a complete ETag header string, including the quotes (if the client correctly set them), and returns the last modified time encoded in the ETag.
+func ParseETag(e string) (time.Time, error) {
+	if len(e) < 2 || e[0] != '"' || e[len(e)-1] != '"' {
+		return time.Time{}, errors.New("unquoted string, value must be quoted")
+	}
+	e = e[1 : len(e)-1] // strip quotes
+
+	prefix := `v` + strconv.Itoa(ETagVersion) + `-`
+	if len(e) < len(prefix) || !strings.HasPrefix(e, prefix) {
+		return time.Time{}, errors.New("malformed, no version prefix")
+	}
+
+	timeStr := e[len(prefix):]
+
+	i, err := strconv.ParseInt(timeStr, 36, 64)
+	if err != nil {
+		return time.Time{}, errors.New("malformed")
+	}
+
+	t := time.Unix(0, i)
+
+	const year = time.Hour * 24 * 365
+
+	// sanity check - if the time isn't +/- 20 years, error. This catches overflows and near-zero errors
+	if t.After(time.Now().Add(20*year)) || t.Before(time.Now().Add(-20*year)) {
+		return time.Time{}, errors.New("malformed, out of range")
+	}
+
+	return t, nil
+}
--- a/traffic_ops/testing/api/v14/deliveryservice_ims_test.go
+++ b/traffic_ops/testing/api/v14/deliveryservice_ims_test.go
@@ -1,0 +1,277 @@
+package v14
+
+/*
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+import (
+	"net/http"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/apache/trafficcontrol/lib/go-rfc"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+)
+
+func TestDeliveryServiceIMS(t *testing.T) {
+	WithObjs(t, []TCObj{CDNs, Types, Tenants, Users, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, DeliveryServices}, func() {
+		GetTestDeliveryServiceIMSAll(t)
+		GetTestDeliveryServiceIMSSingle(t)
+	})
+}
+
+func GetTestDeliveryServiceIMSAll(t *testing.T) {
+	req, err := http.NewRequest(http.MethodGet, TOSession.URL+"/api/1.4/deliveryservices", nil)
+	if err != nil {
+		t.Fatalf("failed to create request: %s", err.Error())
+	}
+
+	resp, err := TOSession.Client.Do(req)
+	if err != nil {
+		t.Fatalf("running request: %s", err.Error())
+	}
+	resp.Body.Close()
+
+	lastModified := resp.Header.Get(rfc.HdrLastModified)
+	if lastModified == "" {
+		t.Fatalf("deliveryservices request expected: " + rfc.HdrLastModified + " header, actual: missing")
+	}
+
+	etag := resp.Header.Get(rfc.HdrETag)
+	if etag == "" {
+		t.Fatalf("deliveryservices request expected: " + rfc.HdrETag + " header, actual: missing")
+	}
+
+	{
+		req, err := http.NewRequest(http.MethodGet, TOSession.URL+"/api/1.4/deliveryservices", nil)
+		if err != nil {
+			t.Fatalf("failed to create request: %s", err.Error())
+		}
+		req.Header.Add(rfc.HdrIfModifiedSince, lastModified)
+
+		resp, err := TOSession.Client.Do(req)
+		if err != nil {
+			t.Fatalf("running request: %s", err.Error())
+		}
+		resp.Body.Close()
+
+		if resp.StatusCode != http.StatusNotModified {
+			t.Errorf("deliveryservices request with " + rfc.HdrIfModifiedSince + " expected: 304, actual: " + strconv.Itoa(resp.StatusCode))
+		}
+	}
+
+	{
+		req, err := http.NewRequest(http.MethodGet, TOSession.URL+"/api/1.4/deliveryservices", nil)
+		if err != nil {
+			t.Fatalf("failed to create request: %s", err.Error())
+		}
+		req.Header.Add(rfc.HdrIfNoneMatch, etag)
+
+		resp, err := TOSession.Client.Do(req)
+		if err != nil {
+			t.Fatalf("running request: %s", err.Error())
+		}
+		resp.Body.Close()
+
+		if resp.StatusCode != http.StatusNotModified {
+			t.Errorf("deliveryservices request with " + rfc.HdrIfNoneMatch + " expected: 304, actual: " + strconv.Itoa(resp.StatusCode))
+		}
+	}
+}
+
+func GetTestDeliveryServiceIMSSingle(t *testing.T) {
+	dses, _, err := TOSession.GetDeliveryServices()
+	if err != nil {
+		t.Errorf("cannot GET DeliveryServices: %v\n", err)
+	}
+	ds := tc.DeliveryService{}
+	for _, tods := range dses {
+		if !tods.Type.IsHTTP() && !tods.Type.IsDNS() {
+			continue
+		}
+		ds = tods
+		break
+	}
+	if ds.ID == 0 {
+		t.Fatalf("GET DeliveryServices returned no DNS or HTTP dses, must have at least 1 to test")
+	}
+
+	req, err := http.NewRequest(http.MethodGet, TOSession.URL+"/api/1.4/deliveryservices?id="+strconv.Itoa(ds.ID), nil)
+	if err != nil {
+		t.Fatalf("failed to create request: %s", err.Error())
+	}
+
+	resp, err := TOSession.Client.Do(req)
+	if err != nil {
+		t.Fatalf("running request: %s", err.Error())
+	}
+	resp.Body.Close()
+
+	lastModified := resp.Header.Get(rfc.HdrLastModified)
+	if lastModified == "" {
+		t.Fatalf("deliveryservices request expected: " + rfc.HdrLastModified + " header, actual: missing")
+	}
+
+	etag := resp.Header.Get(rfc.HdrETag)
+	if etag == "" {
+		t.Fatalf("deliveryservices request expected: " + rfc.HdrETag + " header, actual: missing")
+	}
+
+	{
+		// test a single DS with INM works
+
+		req, err := http.NewRequest(http.MethodGet, TOSession.URL+"/api/1.4/deliveryservices?id="+strconv.Itoa(ds.ID), nil)
+		if err != nil {
+			t.Fatalf("failed to create request: %s", err.Error())
+		}
+		req.Header.Add(rfc.HdrIfNoneMatch, etag)
+
+		resp, err := TOSession.Client.Do(req)
+		if err != nil {
+			t.Fatalf("running request: %s", err.Error())
+		}
+		resp.Body.Close()
+
+		if resp.StatusCode != http.StatusNotModified {
+			t.Errorf("deliveryservices request with " + rfc.HdrIfNoneMatch + " expected: 304, actual: " + strconv.Itoa(resp.StatusCode))
+		}
+	}
+
+	{
+		// test modifying the DS with a field on the same table, and verify an IMS/INM do NOT return a 304
+
+		ds.MaxOriginConnections += 50
+		time.Sleep(time.Second) // sleep for 1s, because IMS is 1-second resolution. Otherwise, it may not be modified.
+		if _, err := TOSession.UpdateDeliveryService(strconv.Itoa(ds.ID), &ds); err != nil {
+			t.Fatalf("cannot UPDATE DeliveryService by ID: %v\n", err)
+		}
+
+		{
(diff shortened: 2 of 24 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4113/
Test FAILed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4114/
Test FAILed.

rob05c:
This implements RFC 7234 Validation. See https://tools.ietf.org/html/rfc7234

Most web browsers will automatically set the headers and use `304 Not Modified` responses appropriately.

In a nutshell, if a server sends a `Last-Modified` or `ETag` header, the client may cache the body of that response. Then, if the client needs to request a new version of that resource, the client may set a request header `If-Modified-Since` or `If-None-Match` with the value it previously received in `Last-Modified` or `ETag`, respectively. Then, the server will inspect the `If-Modified-Since` or `If-None-Match`, and use it to determine if the resource has been modified since that time (the ETag isn't necessarily a time, but it happens to be in this PR). If the resource hasn't been modified, the server sends a `304` code with no body. The client, on receiving a `304`, can then use its previous object knowing it has not been changed on the server. If the object has been modified, the Server sends an ordinary `200` response with the new object, as if no `If-Not-Modified` or `If-None-Match` had been sent.

To answer your questions:

1. No, clients don't need to send anything special, ETag and Last-Modified are always sent. They are not expensive to generate in addition to the object, and the client should ignore them if it doesn't understand them or perform caching.

2. To generate the `Last-Modified` and `ETag`, the DB query now selects an additional `last_modified_any` (see https://github.com/apache/trafficcontrol/pull/3862/files#diff-5e12a051d74ce577019ade5ec8744d93R471 ), with no additional query. 

If the client has sent a `IMS` or `INM`, the server does to an additional, inexpensive query to get the last modified time of any table queried to create the object. If the client didn't send a Validation header, this query isn't done. So, for requests where the object did change, this adds a tiny performance cost. But it makes it drastically faster for unchanged objects.

Alternatively, we could always do the full query every time, and use its `last_modified_any` to return a `304`. That would avoid the tiny performance cost for updated objects where the client has sent an IMS, but it would significantly reduce the performance gain of a successful validation. We're expecting more requests to validate successfully than not, as clients do many more reads than writes; especially as ATS configs are moved client-side and ORT is changed to check for updates more frequently.

3. On successful validation, no headers are sent, only a `304` code with no body. Technically, the RFC permits returning headers, which the client is required to update on the object (though we should be aware, a huge number of clients don't). But we don't have a need to send any header updates.

4. If the IMS/INM validation fails because the object has changed, no additional headers are sent, the 200 code is sent with the full object body, as if no IMS or INM had been received. This is in accordance with RFC 7234.

5. No, in no case do the headers depend on the client request, except the standard IMS/INM header triggering a 304. 

6. It's valuable and trivial to support both. Some clients only understand Last-Modified and not ETag, but ETag is more precise. And it's trivial to support both, just a few more lines of code to check, which only exist once, not for every endpoint. See:
https://github.com/apache/trafficcontrol/pull/3862/files#diff-8db8230918d4903a6a296404e384e8f3R51 
https://github.com/apache/trafficcontrol/pull/3862/files#diff-8db8230918d4903a6a296404e384e8f3R56

rawlinp:
Ok, thank you for the information. I think TO running two DB queries -- one to validate a given ETag and a subsequent one to actually retrieve the full set of resources along with the new ETag -- is fine. Like you said, resources are typically read much more often than they are written, and the extra validation query won't be run for clients that don't send the required headers.

I am curious how this approach handles tenancy w.r.t. delivery services and (in the future) other tenant-able resources. Did you have to do anything special there? What happens if two different tenants update a deliveryservice in their respective tenant at the exact same time and end up with the same `last_updated_any` result?

rob05c:
>I am curious how this approach handles tenancy w.r.t. delivery services and (in the future) other tenant-able resources. Did you have to do anything special there? 

It adds the tenancy query part: https://github.com/apache/trafficcontrol/pull/3862/files#diff-015e9d82ec0d477e602ef7bde3c8b008R841

>What happens if two different tenants update a deliveryservice in their respective tenant at the exact same time and end up with the same last_updated_any result?

Then a If-Modified-Since request could theoretically get a slightly-out-of-date object, as if they had requested a microsecond earlier. 

But in practice, that isn't a problem, because if you're requesting, it doesn't matter in practical terms whether you got one person's edit or another - if the second person had been first, the result would be the same to your client. Postgres has microsecond-precision, which is on the order of clock ticks. So, two database requests would have to come in within a few clock ticks of each-other (probably on different CPU cores). It's theoretically possible, but extremely unlikely.

That said, I do think we should revisit this when Self Service is in place. Right now, TC can't really support a lot of changes at once, just because they aren't safe to synchronize. Further, ORT can't really request more than every 10mins or so, or less if you have fewer caches.

But with SS, we'll have a lot more people making a lot more changes more quickly. Further, we're hoping to lower the ORT pull time, once it can take advantage of HTTP Caching. Once those are in place, the odds of a collision greatly increase (though it's still a very tiny chance, again on the order of clock ticks).

So, the other way to do change tracking is with an auto-incrementing number, essentially just like the `last_updated` field but an integer instead of a time. But, that has a similar problem, with overflow. We could very well overflow an int64 if that many changes were happening. So, I think the best solution is both: a timestamp, and an auto-incrementing number which is 0 for the first microsecond of the same timestamp, and becomes 1 if another transaction happens in the same microsecond, etc.

That solution isn't perfect - it can still theoretically overflow. But a perfect solution requires arbitrary-precision integers which are untenably expensive, and the odds of overflowing an int64 of changes in the same microsecond are astronomically unlikely, certainly on existing hardware.

I don't think the development cost of that is unreasonable, but it isn't trivial either. So again, I vote we do that as soon as Self Service starts to become real, but not right now, because TC just can't handle the number of transactions to make it statistically possible right now, and even if it did happen, it'll just result in ORT or whatever the client is getting a change that's just slightly missing one change, as if it had requested one microsecond older, until the next change/request/ORT run.

But all of that is my opinion, obviously these things are statistical. Does that seem reasonable to you?

rawlinp:
Ok, I'm just trying to understand how the query [here](https://github.com/apache/trafficcontrol/pull/3862/files#diff-015e9d82ec0d477e602ef7bde3c8b008R1564) works w.r.t. tenancy. Let's say my tenancy tree is this (`Ta` is the parent of `Tb`):
```
Tenants:         DSes:
  Ta              Da (last_updated = 1)
  /   
Tb                Db (last_updated = 2)
```
And let's say there are deliveryservices `Da` (which is only visible to tenant `Ta`) and `Db` (which is visible to tenants `Ta` and `Tb`).

When a user in tenant `Ta` requests all delivery services, they'll get back a result with an Etag based on the timestamp of delivery service `Db` (since it is the most recently modified DS) which gets cached somewhere. When a user in tenant `Tb` requests all delivery services, they will be served the result from tenant `Ta`'s request because the most recently modified DS is the same.

Is that correct, or am I missing something? It also seems like the result of that `last_updated_any` subquery is based on _all_ delivery services and related tables, rather than just the things that particular tenant can see. Is that right?

rob05c:
>When a user in tenant Ta requests all delivery services, they'll get back a result with an Etag based on the timestamp of delivery service Db (since it is the most recently modified DS) which gets cached somewhere.

Sort of. The ETag time might happen to come from one DS, but the timestamp doesn't "belong" to one DS, it's the latest time of any row on any table on any DS in the response.

>When a user in tenant Tb requests all delivery services, they will be served the result from tenant Ta's request because the most recently modified DS is the same.

No, the query to get the `last_modified_any` to create the ETag+LM, and the query to compare an ETag to determine if it was modified, both include the tenancy subquery.

So, when tenant `Tb` requests all delivery services with an IMS/INM, it will first check `dsModified`, which includes a tenancy check, and thus only selects the DSes `Tb` has access to. If they're unmodified, we send a 304. If they are modified, we then select the data, along with the `last_updated_any` to build the ETag+LM, which is also selected with the rest of the data, and thus includes the Tenancy subquery, and only gets the last updated time of any DS (or any joined rows from other tables) which is part of the response of DSes `Tb`'s tenant has access to.

So, the ETag/LM for any request isn't all DSes, but only the DSes that client has access to.

The only place this could be a problem, would be if one client made a request, saved the ETag, logged out, logged in a different user, and made an IMS request with the previously saved ETag. Which would be wrong, the ETag/LM is specific to the user. 

Though now that I think about it, it doesn't violate the HTTP spec to do that. I think the proper way to handle that, would be to put the username in the ETag as well. I can add that, or make an issue for it. But I can't imagine anyone doing it in practice.

rawlinp:
> No, the query to get the last_modified_any to create the ETag+LM, and the query to compare an ETag to determine if it was modified, both include the tenancy subquery.

I'm not sure if I'm missing something, but looking at  [the inner last_updated_any query](https://github.com/apache/trafficcontrol/pull/3862/files#diff-015e9d82ec0d477e602ef7bde3c8b008R1564), it doesn't seem to account for tenancy. The `AddTenancyCheck` function appends the check to the given query, which would make the _main_ outer part of the query adhere to tenancy, but I don't know how it would ever add it to that inner subquery. So what it looks like to me is that the result will only include DSes that the tenant has access to, but the `last_updated_any` query takes into account _all_ DSes in the database (because tenancy is not checked). Is that correct? Basically, a cached response for one tenant could be invalidated by changes to delivery services outside of that tenant (because the ETag would change).

> But I can't imagine anyone doing it in practice.

For testing purposes I log out and log into a different user every now and then, and if my browser has cached responses then I would potentially see things I shouldn't have access to as a certain user. I shouldn't have to be aware of that as a tester and have to work around that. I don't think we can make assumptions on what people will do in practice. If the hole is there, there's a chance for someone to step in it. If we're aware of a hole, we should fill it in.

rob05c:
>I'm not sure if I'm missing something, but looking at the inner last_updated_any query, it doesn't seem to account for tenancy. The AddTenancyCheck function appends the check to the given query, which would make the main outer part of the query adhere to tenancy, but I don't know how it would ever add it to that inner subquery.

It does an inner join on the ds id of the outer query:
https://github.com/apache/trafficcontrol/pull/3862/files#diff-015e9d82ec0d477e602ef7bde3c8b008R1583

rawlinp:
Ah, that is what I was missing. So when I query for all DSes, that subquery will be run for every single DS I have access to? As the number of DSes continues to grow, will it become a problem for clients that don't use IMS/INM? What is the change in performance when including the new subquery?

rob05c:
> What is the change in performance when including the new subquery?

```
traffic_ops=> SELECT
ds.active,
ds.anonymous_blocking_enabled,
ds.cacheurl,
ds.ccr_dns_ttl,
ds.cdn_id,
cdn.name as cdnName,
ds.check_path,

CAST(ds.deep_caching_type AS text) as deep_caching_type,
ds.display_name,
ds.dns_bypass_cname,
ds.dns_bypass_ip,
ds.dns_bypass_ip6,
ds.dns_bypass_ttl,
ds.dscp,
ds.edge_header_rewrite,
ds.geolimit_redirect_url,
ds.geo_limit,
ds.geo_limit_countries,
ds.geo_provider,
ds.global_max_mbps,
ds.global_max_tps,
ds.fq_pacing_rate,
ds.http_bypass_fqdn,
ds.id,
ds.info_url,
ds.initial_dispersion,
ds.ipv6_routing_enabled,
ds.last_updated,

ds.logs_enabled,
ds.long_desc,
ds.long_desc_1,
ds.long_desc_2,
ds.max_dns_answers,

ds.mid_header_rewrite,
COALESCE(ds.miss_lat, 0.0),
COALESCE(ds.miss_long, 0.0),
ds.multi_site_origin,
(SELECT o.protocol::text || '://' || o.fqdn || rtrim(concat(':', o.port::text), ':')
FROM origin o
WHERE o.deliveryservice = ds.id
AND o.is_primary) as org_server_fqdn,
ds.origin_shield,
ds.profile as profileID,
profile.name as profile_name,
profile.description  as profile_description,
ds.protocol,
ds.qstring_ignore,
ds.range_request_handling,
ds.regex_remap,
ds.regional_geo_blocking,
ds.remap_text,
ds.routing_name,
ds.signing_algorithm,
ds.ssl_key_version,
ds.tenant_id,
tenant.name,
ds.tr_request_headers,
ds.tr_response_headers,
type.name,
ds.type as type_id,
ds.xml_id,
cdn.domain_name as cdn_domain
from deliveryservice as ds
JOIN type ON ds.type = type.id
JOIN cdn ON ds.cdn_id = cdn.id
LEFT JOIN profile ON ds.profile = profile.id
LEFT JOIN tenant ON ds.tenant_id = tenant.id;
Time: 355.726 ms
```
```
traffic_ops=> SELECT
ds.active,
ds.anonymous_blocking_enabled,
ds.cacheurl,
ds.ccr_dns_ttl,
ds.cdn_id,
cdn.name as cdnName,
ds.check_path,

CAST(ds.deep_caching_type AS text) as deep_caching_type,
ds.display_name,
ds.dns_bypass_cname,
ds.dns_bypass_ip,
ds.dns_bypass_ip6,
ds.dns_bypass_ttl,
ds.dscp,
ds.edge_header_rewrite,
ds.geolimit_redirect_url,
ds.geo_limit,
ds.geo_limit_countries,
ds.geo_provider,
ds.global_max_mbps,
ds.global_max_tps,
ds.fq_pacing_rate,
ds.http_bypass_fqdn,
ds.id,
ds.info_url,
ds.initial_dispersion,
ds.ipv6_routing_enabled,
ds.last_updated,
(SELECT MAX(last_updated) from (
  SELECT
    UNNEST(ARRAY[
      dsu.last_updated,
      tp.last_updated,
      cdn.last_updated,
      pr.last_updated,
      tn.last_updated,
      og.last_updated
    ]) as last_updated
  FROM
    deliveryservice dsu
    JOIN type tp on tp.id = dsu.type
    JOIN cdn ON dsu.cdn_id = cdn.id
    LEFT JOIN origin og ON og.deliveryservice = dsu.id

    LEFT JOIN profile pr ON pr.id = dsu.profile
    LEFT JOIN tenant tn ON tn.id = dsu.tenant_id
  WHERE dsu.id = ds.id
) as updateds
WHERE last_updated IS NOT NULL) as last_updated_any,
ds.logs_enabled,
ds.long_desc,
ds.long_desc_1,
ds.long_desc_2,
ds.max_dns_answers,

ds.mid_header_rewrite,
COALESCE(ds.miss_lat, 0.0),
COALESCE(ds.miss_long, 0.0),
ds.multi_site_origin,
(SELECT o.protocol::text || '://' || o.fqdn || rtrim(concat(':', o.port::text), ':')
FROM origin o
WHERE o.deliveryservice = ds.id
AND o.is_primary) as org_server_fqdn,
ds.origin_shield,
ds.profile as profileID,
profile.name as profile_name,
profile.description  as profile_description,
ds.protocol,
ds.qstring_ignore,
ds.range_request_handling,
ds.regex_remap,
ds.regional_geo_blocking,
ds.remap_text,
ds.routing_name,
ds.signing_algorithm,
ds.ssl_key_version,
ds.tenant_id,
tenant.name,
ds.tr_request_headers,
ds.tr_response_headers,
type.name,
ds.type as type_id,
ds.xml_id,
cdn.domain_name as cdn_domain
from deliveryservice as ds
JOIN type ON ds.type = type.id
JOIN cdn ON ds.cdn_id = cdn.id
LEFT JOIN profile ON ds.profile = profile.id
LEFT JOIN tenant ON ds.tenant_id = tenant.id;
Time: 357.186 ms
```

Looks like it's negligible.

Postgres' query optimizer should be optimizing away the separate identical joins. In fact, we could rewrite it to not be separate joins, I just thought it was more readable that way.

rawlinp:
Yeah, that does seem negligible -- thanks for timing those queries. That new subquery does seem pretty confusing due to `select max -> select unnest array`, but I'm not sure there's a way around that necessary complexity. If we did part of that query in Go (e.g. selecting the max of all the different tables' last_updated columns per row), could we simplify the sql query a bit? Or do you think it would still be optimal to handle that in the sql query itself?

rob05c:
>If we did part of that query in Go

I haven't timed it, but I'd guess it'd add a significant amount of time. Each DB query is probably 50ms minimum, plus latency. IMO it's worth the time to do in one query, especially for reads.

```
SELECT
ds.last_updated,
GREATEST(
  ds.last_updated,
  type.last_updated,
  cdn.last_updated,
  profile.last_updated,
  tenant.last_updated,
  o.last_updated
) as last_updated_any,
(o.protocol::text || '://' || o.fqdn || rtrim(concat(':', o.port::text), ':')) as org_server_fqdn,
ds.xml_id
from deliveryservice as ds
JOIN type ON ds.type = type.id
JOIN cdn ON ds.cdn_id = cdn.id
LEFT JOIN profile ON ds.profile = profile.id
LEFT JOIN tenant ON ds.tenant_id = tenant.id
LEFT JOIN origin o ON (o.deliveryservice = ds.id AND o.is_primary)
```

What about that?

rawlinp:
Oh I meant still just running a single DB query but possibly doing something like iterating over all the result rows in Go and determining the max of the various `last_updated` columns. Your `GREATEST` version of the query definitely cleans it up a bit -- now it doesn't even have to be a subquery at all anymore.

traffic_ops/traffic_ops_golang/cachegroupparameter/parameters.go
```diff
@@ -54,6 +54,32 @@ func (cgparam *TOCacheGroupParameter) GetType() string {
...
+}
+
+func (cgparam TOCacheGroupParameter) GetKeys() (map[string]interface{}, bool) {
+	if cgparam.ID == nil {
+		return map[string]interface{}{"id": 0}, false
+	}
+	return map[string]interface{}{"id": *cgparam.ID}, true
+}
+
+func (cgparam *TOCacheGroupParameter) SetKeys(keys map[string]interface{}) {
```
ocket8888: Is all of this related to Etag/IMS support for DSes and servers?

traffic_ops/traffic_ops_golang/cachegroupparameter/unassigned_parameters.go
```diff
@@ -51,6 +51,32 @@ func (cgunparam *TOCacheGroupUnassignedParameter) GetType() string {
...
+}
+
+func (cgunparam TOCacheGroupUnassignedParameter) GetKeys() (map[string]interface{}, bool) {
+	if cgunparam.ID == nil {
+		return map[string]interface{}{"id": 0}, false
+	}
+	return map[string]interface{}{"id": *cgunparam.ID}, true
+}
+
+func (cgunparam *TOCacheGroupUnassignedParameter) SetKeys(keys map[string]interface{}) {
```
ocket8888: Same as above RE relatedness

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -181,6 +181,34 @@ RETURNING job.asset_url,
...
+}
+
+func (job InvalidationJob) GetKeys() (map[string]interface{}, bool) {
+	if job.ID == nil {
+		return map[string]interface{}{"id": 0}, false
+	}
+	return map[string]interface{}{"id": *job.ID}, true
+}
+
+func (job *InvalidationJob) SetKeys(keys map[string]interface{}) {
```
ocket8888: Same as above RE relatedness

traffic_ops/testing/api/v14/deliveryservice_ims_test.go
```diff
@@ -0,0 +1,277 @@
...
+	if lastModified == "" {
+		t.Fatalf("deliveryservices request expected: " + rfc.HdrLastModified + " header, actual: missing")
+	}
+
+	etag := resp.Header.Get(rfc.HdrETag)
+	if etag == "" {
+		t.Fatalf("deliveryservices request expected: " + rfc.HdrETag + " header, actual: missing")
+	}
+
+	{
```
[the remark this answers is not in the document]
rob05c: It makes `req`, `resp`, and `err` new variables, rather than reusing existing ones. The point is to make copy-paste bugs impossible, where a variable was re-used accidentally. The above should really be in its own block, too, to make it a compile error to use the earlier variables in the lower block.

lib/go-rfc/cachecontrol.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// FormatHTTPDate formats t as an RFC7231§7.1.1 HTTP-date.
+func FormatHTTPDate(t time.Time) string {
+	return t.Format(time.RFC1123)
+}
+
+// ETag takes the last time the object was modified, and returns an ETag string. Note the string is the complete header value, including quotes. ETags must be quoted strings.
+func ETag(t time.Time) string {
+	return `"v` + strconv.Itoa(ETagVersion) + `-` + strconv.FormatInt(t.UnixNano(), 36) + `"`
```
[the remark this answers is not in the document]
rob05c: You're correct. I considered that. Right now, with the scale of Traffic Control deployments and lack of self-service, it's extremely unlikely.

As we move towards more Self-Service and automation, IMO the ideal solution is to use a timestamp + incrementing integer, where the integer is incremented for each operation sharing the same timestamp.

There is no perfect solution. With just integers, there's some number of operations that will overflow. But a Postgres datetime + bigint changes it from "extremely unlikely today, and somewhat likely over time with scripted self-service deployments" to "approaching statistically impossible."

But, those integers would be a much larger change to our database schema and queries.

So, I'd vote we do this for now, and add the integers when we're closer with Self Service.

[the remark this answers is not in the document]
ocket8888: Well you could add the object's uniquely identifying property to the Etag - that'd guarantee that even if there was a collision it'd never be for the same type of object.

traffic_ops/traffic_ops_golang/cachegroupparameter/parameters.go
rob05c: Yes, these had to be added to make TOCacheGroupParameter fulfill the CRUDer interfaces using Modifieder.

ocket8888: But the PR title seems to indicate this should only affect the `/deliveryservices` and `/servers` endpoints? Is this necessary to do that or does this also add IMS support to `/cachegroupparameters`?

rob05c: No, it doesn't add IMS support to `/cachegroupparameters`; but it prepares all CRUDer endpoints for it to be easily added in the future.

If you remove that code, the compiler gives the error
```
routing/routes.go:136:84: cannot use &cachegroupparameter.TOCacheGroupParameter literal (type *cachegroupparameter.TOCacheGroupParameter) as type api.Reader in argument to api.ReadHandler:
        *cachegroupparameter.TOCacheGroupParameter does not implement api.Reader (missing GetAuditName method)
```

Because this PR adds `Identifier` to the `api.Read` interface. That isn't strictly necessary in this PR. But we'll likely want to add unique integers to the ETag in the future, for better uniqueness than the timestamp. 

I think we should also add the unique components of the URL to the ETag, e.g. that it's a deliveryservice and its ID, to catch user errors, if the user sends the ETag without respect to the query params. For example, a user's client might automatically request `/deliveryservices`, cache the ETag, and use it again to request `/deliveryservices?42`. Which would be wrong; but the RFCs don't make that clear, they say little about query strings because it's application-specific whether the query string affects the ETag. It's an easy mistake to make, and we can easily prevent it.

Adding `Identifier` to `Reader` enables that. We could put it in `Modifieder`, but because that's currently a dangerous type assertion, it's easy to miss the Identifier and accidentally not fulfill the interface. Really, I don't see a reason why `Reader`s shouldn't be required to also be `Identifier`s. All objects should be uniquely identifiable.

ocket8888:
Both of these endpoints - and many others - support If-(Un)Modified-Since and/or If-Match/Etag, so should this be closed?

PR #3870, opened 2019-08-14
Rewrite /capabilities to Go

### LINKED ISSUE #3821: Rewrite /capabilities to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/capabilities` is handled by Perl

## Expected / new behavior:
`/capabilities` should be handled by Go.

### Minimum API Versions
- [ ] `GET`
- [ ] `POST`

N/A

ocket8888:
- [x] This PR fixes #3821

Rewrites the `/capabilities` endpoint to Go. Also adds handlers for the `PUT` and `DELETE` request methods.

- Documentation
- Traffic Control Client (Python and Go)
- Traffic Ops
- Traffic Portal

Run the associated tests in `traffic_ops/testing/api/v14`.

- [x] This PR includes tests


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4143/
Test PASSed.

rawlinp:
So, I have a number of questions about capabilities:
1. what is the difference between a capability and an api_capability?
2. do we need both?
3. do we really need the ability to CRUD capabilities in general? It seems like capabilities are statically defined in terms of API endpoints/"things you can do". Why do we need the ability to dynamically CRUD those things? 

It seems like "roles" are what really actually to be dynamic -- not capabilities. If you have a user you want to give a limited set of capabilities, you create a new role, add in the capabilities you want to give them, and assign the new role to the user.

It doesn't make sense to create a new capability via the API or update or delete existing capabilities. IMO it really only makes sense to be able to _read_ the statically defined capabilities, so that you have the full set to choose from.

ocket8888:
1) `api_capability` links an API route to a particular capability.
2) Yes and no. We need both _concepts_, but the routes governed by a capability could probably be dealt with under the `/capabilities` route itself.
3) new capabilities can exist because arbitrary endpoints can be added via extensions. You can also group api routes under a one or more capabilities in arbitrary ways. Not sure that second one is a _good_ reason, but it is a true one.

In any case, such deprecations probably need to go to the mailing list for discussion. I don't disagree that this could be handled better - possibly by requiring extensions to define a governing capability? - but I do want to just get the route into Go, since deprecation takes a full major release cycle anyway.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4147/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4152/
Test PASSed.

rawlinp:
Ah, I see now that `capability.name` is used as a FK in the `api_capabilities` table. If you update the name of a capability does it automatically update all the FK references in the `api_capabilities` table? That was one concern I had about name-based keys in general. If it doesn't auto-update, then it seems like we should make `capability.name` an immutable column.

Good point about the arbitrary endpoints via extensions. I don't really like the idea of making all the built-in capabilities mutable though. It has all the same issues as making built-in `types` mutable today. Maybe we need the concept of extension-specific capabilities, and if extensions want to take advantage of roles/capabilities, they should load their extension-specific capabilities into the DB at startup time. Do extensions have access to TODB? I can't tell from looking through the `traffic_ops_golang/plugin` dir. Either way, it seems like we wouldn't want the ability to create/update/delete any kind of capabilities dynamically -- only the ability to list current capabilities that we could assign to a custom role.

rob05c:
>If you update the name of a capability does it automatically update all the FK references in the api_capabilities table? 

No, but it should. 

https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/db/create_tables.sql#L3243

We should add a migration to change `fk_capability` to `ON UPDATE CASCADE`.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4153/
Test PASSed.

ocket8888:
> _"If you update the name of a capability does it automatically update all the FK references in the api_capabilities table?"_

It can, depends if the foreign key constraint was defined with `ON UPDATE CASCADE`. Which it should have been. But looking at `traffic_ops/app/db/create_tables.sql` it was not. Adding that clause is simple enough - and should be done, IMO -, but it's a database migration, so I'm not sure it's proper to include it in what's already a pretty expansive PR.

Extensions certainly have access to the database at runtime, but idk if they do at startup.

ocket8888:
Oh, rob05c beat me to it. But also of note while we're looking, the Role/Capability relationship has the same problem.

rob05c:
>Do extensions have access to TODB?
Yes. TO Plugins can access anything the main TO app can. If a hook can't access something that's needed, it can be added to the `Data` struct for that hook, e.g. `OnRequestData`.

rawlinp:
Well, if we make the built-in capabilities immutable the `ON UPDATE CASCADE` is unnecessary. Since TO plugins can access the TODB at startup, I think that would be the "proper" way of supporting capabilities for arbitrary extension routes. An extension route shouldn't require an operator to manually add its capabilities via the API. The extension should load them itself on startup time, and they should be immutable from an API perspective.

Since capabilities aren't really used in the code yet, I think we have some freedom to design them properly before we get stuck in the rut of having to support unnecessary API endpoints.

ocket8888:
The best we could do in that direction is return an error without breaking the "API version promise". And actually, that might qualify; not sure.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4154/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4155/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4156/
Test PASSed.

asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4166/
Test PASSed.

traffic_ops/client/capability.go
```diff
@@ -0,0 +1,121 @@
...
+ */
+
+import "encoding/json"
+import "fmt"
+import "net"
+import "net/http"
+
+import "github.com/apache/trafficcontrol/lib/go-tc"
+
+const API_v4_CAPABILITIES = "/api/1.4/capabilities"
```
mhoppa: should this be v14?

traffic_ops/client/capability.go
```diff
@@ -0,0 +1,121 @@
...
+func (to *Session) ReplaceCapabilityByName(name string, c tc.Capability) (tc.Alerts, ReqInf, error) {
+	var remoteAddr net.Addr
+	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
+
+	reqBody, err := json.Marshal(c)
+	if err != nil {
+		return tc.Alerts{}, reqInf, err
+	}
+
+	endpoint := fmt.Sprintf("%s?name=%s", API_v4_CAPABILITIES, name)
```
mhoppa: I know we do not do this across any of our client code so can ignore this commentbut we really should leverage the net/url lib to build up our query parameters so they are url encoded.

traffic_ops/testing/api/v14/capabilities_test.go
```diff
@@ -0,0 +1,159 @@
...
+	for _,c := range staticCapabilities {
+		capMap[c.Name] = c.Description
+	}
+
+
+	caps, _, err := TOSession.GetCapabilities()
+	if err != nil {
+		t.Fatalf("could not get all capabilities: %v", err)
+	}
+	log.Debugln("capacities:", caps)
```
mhoppa: need this?

traffic_ops/testing/api/v14/capabilities_test.go
```diff
@@ -0,0 +1,159 @@
...
+	if len(caps) != testDataLen {
+		t.Errorf("response returned different number of capabilities than those that exist; got %d, want %d", len(caps), testDataLen)
+		return // we can't FATAL because this testing.T object is shared
```
mhoppa: huh we have been doing it a lot of other places

traffic_ops/testing/api/v14/capabilities_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	for c,_ := range capMap {
+		t.Errorf("Capability '%s' existed in the test data but didn't appear in the response!", c)
+	}
+}
+
+func ReplaceTestCapability(t *testing.T) {
+	if len(testData.Capabilities) < 1 {
+		t.Skip("There aren't any test capabilities, skipping (this should probably be rectified!)")
```
mhoppa: I wonder if we should fail the test is this case? if this happens something is causing it and we want to address it and I worry a skip message will just get lost in the big output

traffic_ops/traffic_ops_golang/capabilities/capabilities.go
```diff
@@ -0,0 +1,305 @@
...
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+	defer inf.Close()
+
+
+	var rows *sql.Rows
+	var err error
+	if name, ok := inf.Params["name"]; ok {
+		rows, err = tx.Query(readQuery + "WHERE name=$1", name)
```
mhoppa: could we instead use https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go#L45?

traffic_ops/traffic_ops_golang/capabilities/capabilities.go
```diff
@@ -0,0 +1,305 @@
...
+			errCode = http.StatusInternalServerError
+			sysErr = fmt.Errorf("Parsing database response: %v", err)
+			api.HandleErr(w, r, tx, errCode, nil, sysErr)
+			return
+		}
+
+		caps = append(caps, cap)
+	}
+
+	respBts, err := json.Marshal(struct{R []tc.Capability `json:"response"`}{caps})
```
mhoppa: isnt there an API util function that wraps an interface in a response that you can call to instead of this?

traffic_ops/traffic_ops_golang/capabilities/capabilities.go
```diff
@@ -0,0 +1,305 @@
...
+	row := tx.QueryRow(createQuery, cap.Name, cap.Description)
+	if err := row.Scan(&cap.Description, &cap.LastUpdated, &cap.Name); err != nil {
+		sysErr = fmt.Errorf("Inserting capability: %v", err)
+		errCode = http.StatusInternalServerError
+		api.HandleErr(w, r, tx, errCode, nil, sysErr)
+		return
+	}
+
+	alert := []tc.Alert{tc.Alert{Level: tc.SuccessLevel.String(), Text: "Capability created."}}
+	responseObj := struct{A []tc.Alert `json:"alerts"`; R tc.Capability `json:"response"`}{alert, cap}
```
mhoppa: same comment here I think there is an API util function that does this as well as the other handlers below

traffic_ops/traffic_ops_golang/capabilities/capabilities.go
```diff
@@ -0,0 +1,305 @@
...
+	var cap tc.Capability
+	if err := decoder.Decode(&cap); err != nil {
+		userErr = fmt.Errorf("Couldn't parse capacity: %v", err)
+		errCode = http.StatusBadRequest
+		api.HandleErr(w, r, tx, errCode, userErr, nil)
+		return
+	}
+
+	if cap.Name != inf.Params["name"] {
+		if ok, err := capabilityNameExists(cap.Name, tx); err != nil {
```
mhoppa: second time Ive seen this logic. move to common function?

mhoppa:
Also some merge conflicts to fix

traffic_ops/testing/api/v14/capabilities_test.go
ocket8888: nope

traffic_ops/client/capability.go
ocket8888: yes, we really should

traffic_ops/client/capability.go
ocket8888: well it oughta be v1.4 😏

traffic_ops/testing/api/v14/capabilities_test.go
ocket8888: yeah, that's wrong. Should be fatal

traffic_ops/traffic_ops_golang/capabilities/capabilities.go
ocket8888: is there?

mhoppa: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/api/api.go#L55

ocket8888:
That thread - though it was swiftly derailed - was only about deprecating `/capabilities/{name}`, and while deprecating methods other than GET was suggested it didn't get any +1s, whereas the `/capabilities/{name}` deprecation recieved unanimous support.

Furthermore, we never settled in that thread on whether or not deprecated endpoints ought to be rewritten - there were proponents for both sides.

traffic_ops/traffic_ops_golang/capabilities/capabilities.go
ocket8888: oh yeah. I don't like using that because it doesn't add trailing newlines. But I can fix that...

lib/go-tc/capabilities.go
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+// Capability reflects the ability of a user in ATC to perform some operation.
+//
+// In practice, they are assigned to relevant Traffic Ops API endpoints - to describe the
+// capabilites of said endpoint - and to user permission Roles - to describe the capabilities
+// afforded by said Role. Note that enforcement of Capability-based permisions is not currently
+// implemented.
+type Capability struct {
+	Description string `json:"description"`
```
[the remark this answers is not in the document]
ocket8888: I don't think I need them, do I?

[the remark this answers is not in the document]
ocket8888: I added them anyway

traffic_ops/traffic_ops_golang/capabilities/capabilities.go
ocket8888: where else do you see that logic?

ocket8888:
retest this please

ocket8888:
retest this please

ocket8888:
retest this please

ocket8888:
retest this please

ocket8888:
retest this please

ocket8888:
retest this please

ocket8888:
retest this please

ocket8888:
If I don't add a DELETE handler then I have to remove the API/integration tests because TO Go will be incapable of deleting capabilities that it creates. Are we cool with no tests for this?

mhoppa:
could https://github.com/apache/trafficcontrol/blob/master/traffic_ops/testing/api/v14/todb_test.go#L280 be leveraged? it would be after all tests are run so unsure if that would suffice.

you could also create a connection to the db to clean up after yourself

ocket8888:
As long as no one else tries to create capabilities during the tests' runtime then that should be fine

docs/source/api/capabilities.rst
```diff
@@ -99,34 +122,34 @@ Request Structure
...
+	Whole-Content-Sha512: A1rjpDy+O+oooYeer2j09pCEDpPEFk/nt8/AaJye2sLkfy93MtquCsB/Rlgz7sCYputd/EPOPDyi2WkN8UB1Rw==
+	X-Server-Name: traffic_ops_golang/
+	Date: Thu, 15 Aug 2019 17:18:03 GMT
+	Content-Length: 219
 
 	{ "alerts": [
 		{
-			"level": "success",
-			"text": "Capability was created."
+			"text": "Capability created.",
```
mhoppa: add deprecation alert here?

traffic_ops/testing/api/v14/withobjs_test.go
```diff
@@ -79,6 +80,7 @@ var withFuncs = map[TCObj]TCObjFuncs{
 	CacheGroups:                          {CreateTestCacheGroups, DeleteTestCacheGroups},
 	CacheGroupsDeliveryServices:          {CreateTestCachegroupsDeliveryServices, DeleteTestCachegroupsDeliveryServices},
 	CacheGroupParameters:                 {CreateTestCacheGroupParameters, DeleteTestCacheGroupParameters},
+	Capabilities:                         {CreateTestCapabilities, DeleteTestCapabilities},
```
mhoppa: no DeleteTestCapabilities causing build failures

ocket8888: Fixed. That was harder than I thought - turns out API capabilities aren't managed by WithObjs and some new tests that have been added since this PR opened were broken because of that.

docs/source/api/capabilities_name.rst
```diff
@@ -19,6 +19,9 @@
 ``capabilities/{{name}}``
 *************************
 
+.. deprecated:: 1.4
+	All of the functionality provided by this endpoint can be accomplished using :ref:`to-api-capabilities` instead. This endpoint will be removed in a future version of :abbr:`ATC (Apache Traffic Control)`.
```
mhoppa: This PR does not add deprecation notices to this path I wonder if this docs change should be added at that point

ocket8888: That's fair.

PR #3883, opened 2019-08-20
Fixed double-build of TR RPMs in CiaB makefile

ocket8888:

When building packages for CDN-in-a-Box using `make` with multiple "jobs" (i.e. with `-j`), there is a race condition building the Traffic Router and Tomcat RPMs, since one build process creates both and they both depend on changes to the TR source. In cases where this manifested - which is often - the build will either fail or simply do double the work it needs to do when building those RPMs. This PR eliminates that by changing the rule to a pattern-based one, which `make` will now properly recognize as a single output (consisting of multiple files) of a single build process.

- CDN in a Box

No docs because it's a bugfix

Build the RPMs for CDN-in-a-Box using `make` with the `-j` flag, observe that it consistently has no issue creating RPMs for both Tomcat and Traffic Router exactly once (note that in order to do that, the `build_default` Docker network must already exist - which is accomplished by just having built an RPM on the same machine before - as otherwise duplicate networks will be created).

No tests because this is just the build system for a toy environment.

N/A - CDN-in-a-Box is not a supported component of Traffic Control, and no such support is purported to exist in any existing releases.

### DIFF
--- a/infrastructure/cdn-in-a-box/Makefile
+++ b/infrastructure/cdn-in-a-box/Makefile
@@ -57,8 +57,6 @@ traffic_monitor/traffic_monitor.rpm: ../../dist/traffic_monitor-$(SPECIAL_SAUCE)
 	cp -f $? $@
 traffic_ops/traffic_ops.rpm: ../../dist/traffic_ops-$(SPECIAL_SAUCE)
 	cp -f $? $@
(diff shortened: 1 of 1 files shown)


asfgit:
https://builds.apache.org/job/trafficcontrol-PR/4175/
Test PASSed.

mitchell852:
maybe @mhoppa can review this?

mhoppa:
It looks great - let me pull down and test thought just to make sure it doesn't do anything funky it shouldnt

PR #3900, opened 2019-09-03
Rewrite /cachegroups/{{id}}/parameters to Go

### LINKED ISSUE #3806: Rewrite /cachegroups/{{id}}/parameters to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/cachegroups/{{id}}/parameters` is handled by Perl

## Expected / new behavior:
`/cachegroups/{{id}}/parameters` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`

N/A

mhoppa:

This PR rewrites the  /cachegroups/{{id}}/parameters endpoint to Go

- [x] This PR fixes #3806 

- Documentation
- Traffic Ops

Bring up TO locally and hit the endpoint to ensure parity between perl implementation and newly written Golang implementation 

The API tests and/or unit tests for TO can be run to verify the PR.

- [x] This PR includes tests

### DIFF
--- a/lib/go-tc/cachegroup_parameters.go
+++ b/lib/go-tc/cachegroup_parameters.go
@@ -1,5 +1,3 @@
-package tc
-
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
@@ -18,6 +16,8 @@ package tc
  * specific language governing permissions and limitations
  * under the License.
  */
+
+package tc
 
 // CacheGroupParameter Cache Group Parameter association
 type CacheGroupParameter struct {
--- a/lib/go-tc/cachegroup_parameters.go
+++ b/lib/go-tc/cachegroup_parameters.go
@@ -19,14 +19,40 @@
 
 package tc
 
-// CacheGroupParameter Cache Group Parameter association
-type CacheGroupParameter struct {
+// CacheGroupParameterRequest Cache Group Parameter request body
+type CacheGroupParameterRequest struct {
 	CacheGroupID int `json:"cacheGroupId"`
 	ParameterID  int `json:"parameterId"`
 }
 
-// CacheGroupParameterResponse Response body for associating parameters with cache groups
-type CacheGroupParameterResponse struct {
+// CacheGroupParameterPostResponse Response body when Posting to associate a Parameter with a Cache Group
+type CacheGroupParametersPostResponse struct {
+	Response []CacheGroupParameterRequest `json:"response"`
+	Alerts
+}
+
+// CacheGroupParameterResponse Cache Group Parameter response body
+type CacheGroupParametersResponse struct {
 	Response []CacheGroupParameter `json:"response"`
 	Alerts
+}
+
+// CacheGroupParameter ...
+type CacheGroupParameter struct {
+	ConfigFile  string    `json:"configFile"`
+	ID          int       `json:"id"`
+	LastUpdated TimeNoMod `json:"lastUpdated"`
+	Name        string    `json:"name"`
+	Secure      bool      `json:"secure"`
+	Value       string    `json:"value"`
+}
+
+// CacheGroupParameterNullable ...
+type CacheGroupParameterNullable struct {
+	ConfigFile  *string    `json:"configFile" db:"config_file"`
+	ID          *int       `json:"id" db:"id"`
+	LastUpdated *TimeNoMod `json:"lastUpdated" db:"last_updated"`
+	Name        *string    `json:"name" db:"name"`
+	Secure      *bool      `json:"secure" db:"secure"`
+	Value       *string    `json:"value" db:"value"`
 }
--- a/lib/go-tc/cachegroup_parameters.go
+++ b/lib/go-tc/cachegroup_parameters.go
@@ -21,8 +21,8 @@ package tc
 
 // CacheGroupParameter Cache Group Parameter association
 type CacheGroupParameter struct {
-	ID          int `json:"cachegroupId"`
-	ParameterID int `json:"parameterId"`
+	CacheGroupID int `json:"cachegroupId"`
+	ParameterID  int `json:"parameterId"`
 }
 
 // CacheGroupParameterResponse Response body for associating parameters with cache groups
--- a/lib/go-tc/cachegroup_parameters.go
+++ b/lib/go-tc/cachegroup_parameters.go
@@ -1,0 +1,32 @@
+package tc
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// CacheGroupParameter Cache Group Parameter association
+type CacheGroupParameter struct {
+	ID          int `json:"cachegroupId"`
+	ParameterID int `json:"parameterId"`
+}
+
+// CacheGroupParameterResponse Response body for associating parameters with cache groups
+type CacheGroupParameterResponse struct {
+	Response []CacheGroupParameter `json:"response"`
+	Alerts
+}
--- a/lib/go-tc/cachegroup_parameters.go
+++ b/lib/go-tc/cachegroup_parameters.go
@@ -21,7 +21,7 @@ package tc
 
 // CacheGroupParameter Cache Group Parameter association
 type CacheGroupParameter struct {
-	CacheGroupID int `json:"cachegroupId"`
+	CacheGroupID int `json:"cacheGroupId"`
 	ParameterID  int `json:"parameterId"`
 }
 
--- a/traffic_ops/testing/api/v14/cachegroups_parameters_test.go
+++ b/traffic_ops/testing/api/v14/cachegroups_parameters_test.go
@@ -1,0 +1,91 @@
+// /*
+
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+
+//  http://www.apache.org/licenses/LICENSE-2.0
+
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+// */
+
+package v14
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+
+	"github.com/apache/trafficcontrol/lib/go-log"
+)
+
+func TestCacheGroupParameters(t *testing.T) {
+	WithObjs(t, []TCObj{Types, Parameters, CacheGroups}, func() {
+		GetTestCacheGroupParameters(t)
+	})
+}
+
+func CreateTestCacheGroupParameters(t *testing.T) {
+
+	firstCacheGroup := testData.CacheGroups[0]
+	cacheGroupResp, _, err := TOSession.GetCacheGroupByName(*firstCacheGroup.Name)
+	if err != nil {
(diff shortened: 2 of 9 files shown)


lib/go-tc/cachegroup_parameters.go
```diff
@@ -0,0 +1,32 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// CacheGroupParameter Cache Group Parameter association
+type CacheGroupParameter struct {
+	ID          int `json:"cachegroupId"`
```
ocket8888: Why not `CacheGroupID` to "match" the JSON?

traffic_ops/testing/api/v14/cachegroups_parameters_test.go
```diff
@@ -0,0 +1,91 @@
...
+func TestCacheGroupParameters(t *testing.T) {
+	WithObjs(t, []TCObj{Types, Parameters, CacheGroups}, func() {
+		GetTestCacheGroupParameters(t)
+	})
+}
+
+func CreateTestCacheGroupParameters(t *testing.T) {
+
+	firstCacheGroup := testData.CacheGroups[0]
+	cacheGroupResp, _, err := TOSession.GetCacheGroupByName(*firstCacheGroup.Name)
```
ocket8888: `cacheGroupResp` can be `nil` here, so maybe you wanna check for that if there's an error and just `Fatalf` out. Or actually maybe just return, because that might mess up object cleanup (citation needed).

Also, `Session.GetCacheGroupByName` is deprecated, instead use [`Session.GetCacheGroupNullableByName`](https://godoc.org/github.com/apache/trafficcontrol/traffic_ops/client#Session.GetCacheGroupNullableByName)

traffic_ops/testing/api/v14/cachegroups_parameters_test.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		t.Errorf("cannot GET Cache Group by name: %v - %v\n", firstCacheGroup.Name, err)
+	}
+
+	firstParameter := testData.Parameters[0]
+	paramResp, _, err := TOSession.GetParameterByName(firstParameter.Name)
```
ocket8888: Same as above RE `nil` response

traffic_ops/testing/api/v14/cachegroups_parameters_test.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		t.Errorf("cannot GET Parameter by name: %v - %v\n", firstParameter.Name, err)
+	}
+
+	cacheGroupID := cacheGroupResp[0].ID
+	parameterID := paramResp[0].ID
+
+	resp, _, err := TOSession.CreateCacheGroupParameter(cacheGroupID, parameterID)
```
ocket8888: Same as above RE nil response

traffic_ops/testing/api/v14/cachegroups_parameters_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	delResp, _, err := TOSession.DeleteCacheGroupParameter(cgp.ID, cgp.ParameterID)
+	if err != nil {
+		t.Errorf("cannot DELETE Parameter by cache group: %v - %v\n", err, delResp)
+	}
+
+	// Retrieve the Cache Group Parameter to see if it got deleted
+	queryParams := fmt.Sprintf("parameterId=%d", cgp.ParameterID)
+
+	parameters, _, err := TOSession.GetCacheGroupParametersByQueryParams(cgp.ID, queryParams)
```
ocket8888: Same as above RE `nil` response.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -119,6 +119,8 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodPost, `cachegroups/{id}/queue_update$`, cachegroup.QueueUpdates, auth.PrivLevelOperations, Authenticated, nil},
 		{1.1, http.MethodPost, `cachegroups/{id}/deliveryservices/?$`, cachegroup.DSPostHandler, auth.PrivLevelOperations, Authenticated, nil},
 
+		{1.1, http.MethodGet, `cachegroups/{id}/parameters/?$`, api.ReadHandler(&cachegroup.TOCacheGroupParameter{}), auth.PrivLevelReadOnly, Authenticated, nil},
```
ocket8888: I think this needs to support an optional `.json` suffix. In Perl this was silently accepted by the server, and clients may have depended on that behavior (they shouldn't have IMO, but it was fairly common). In fact, it looks like Perl would accept

- `/cachegroups/{{ID}}/parameters.json`
- `/cachegroups/{{ID}}/parameters`
- `/cachegroups/{{ID}}/parameters/`
- `/cachegroups/{{ID}}/parameters/.json`
- and even `/cachegroups/{{ID}}/parameters.sntoaheusntha`

Pretty sure we don't support that last one, but a lot of handlers (see the `cdns/name/{name}/sslkeys` route directly below for an example) support all the others.

traffic_ops/traffic_ops_golang/cachegroup/parameters.go
```diff
@@ -0,0 +1,97 @@
...
+
+const (
+	CacheGroupIDQueryParam = "id"
+	ParameterIDQueryParam  = "parameterId"
+)
+
+//we need a type alias to define functions on
+type TOCacheGroupParameter struct {
+	api.APIInfoImpl `json:"-"`
+	tc.ParameterNullable
```
ocket8888: Using this inserts a new field into the response that didn't exist in the Perl implementation: `profiles`. That's fine for a v1.4 route, but to be a generic rewrite of 1.x it shouldn't add things. It also omits sub-second precision to the `lastUpdated` timestamp, which is likely less of a problem, since compliant parsers (probably?) shouldn't care. Might be nice, though.

traffic_ops/traffic_ops_golang/cachegroup/parameters.go
```diff
@@ -0,0 +1,97 @@
...
+		if err = rows.StructScan(&p); err != nil {
+			return nil, nil, errors.New("scanning " + cgparam.GetType() + ": " + err.Error()), http.StatusInternalServerError
+		}
+		if p.Secure != nil && *p.Secure && cgparam.ReqInfo.User.PrivLevel < auth.PrivLevelAdmin {
+			p.Value = &parameter.HiddenField
+		}
+		params = append(params, p)
+	}
+
+	return params, nil, nil, http.StatusOK
```
ocket8888: This implementation doesn't check for existence of the requested Cache Group.

In Perl:
```http
GET /api/1.3/cachegroups/-1/parameters HTTP/1.1
User-Agent: python-requests/2.20.1
Accept: */*
Connection: keep-alive
Cookie: mojolicious=...

HTTP/1.1 404 Not Found
Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Content-Encoding: gzip
Content-Length: 77
Date: Wed, 04 Sep 2019 16:08:48 GMT
Server: Mojolicious (Perl)
Set-Cookie: mojolicious=...; expires=Wed, 04 Sep 2019 20:08:48 GMT; path=/; HttpOnly
Vary: Accept-Encoding
Whole-Content-Sha512: ...

{
	"alerts": [
		{
			"level": "error",
			"text": "Resource not found."
		}
	]
}
```
In your rewrite:
```http
GET /api/1.3/cachegroups/-1/parameters HTTP/1.1
User-Agent: python-requests/2.20.1
Accept: */*
Connection: keep-alive
Cookie: mojolicious=...

HTTP/1.1 200 OK
Content-Encoding: gzip
Set-Cookie: mojolicious=...; Path=/; HttpOnly
Whole-Content-Sha512: ...
Date: Wed, 04 Sep 2019 16:09:01 GMT
Content-Length: 39

{
	"response": []
}
```

rawlinp:
add to whitelist

traffic_ops/traffic_ops_golang/cachegroup/parameters.go
mhoppa: The Perl implementation does not check for existence of the cache group its only checking if you give it a negative ID on the path

Would you like me to match parity to that or actually checking if the cache group exists?

ocket8888: That's true... odd. I guess it's just a consequence of the regex used to extract the ID. I would think actually checking for the Cache Group's existence would be best. Hopefully that's not seen as a breaking API change.

rawlinp: +1 on checking for the cachegroup's existence as added validation. If the cachegroup doesn't exist, it should 404 IMO.

mhoppa: done -> 

```
{
    "alerts": [
        {
            "text": "cachegroup does not exist",
            "level": "error"
        }
    ]
}
```

traffic_ops/traffic_ops_golang/cachegroup/parameters.go
mhoppa: done on using a different type to omit profiles. Did not look into lastUpdated as that looks like the format we consistently across our structs

traffic_ops/traffic_ops_golang/cachegroup/parameters.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	cgID, err := strconv.Atoi(parameters[CacheGroupIDQueryParam])
+	if err != nil {
+		return nil, nil, errors.New("cache group id must be an integer"), http.StatusInternalServerError
+	}
+
+	_, ok, err := getCGNameFromID(cgparam.ReqInfo.Tx.Tx, int64(cgID))
+	if err != nil {
+		return nil, nil, errors.New("getting cachegroup from id"), http.StatusInternalServerError
```
ocket8888: I think if you check here for `err == sql.ErrNoRows` you can give a better message back to the client. If the Cache Group doesn't exist, then the resource they're trying to access doesn't exist, and they should get a 404 back. If that's not the case, then probably throw your hands up and say something wicked happened on the server.

rob05c: It also needs to include the error context, so it's easier to debug when it happens: `		return nil, nil, errors.New("getting cachegroup from id: " + err.Error()), http.StatusInternalServerError`

mhoppa: the getCGNameFromID does that for me https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go#L105 

The behavior you are outlining is the behavior implemented. If you look at the unit tests if there is a storage error it returns internal server error if it returns empty from db (no rows) it returns 404

mhoppa: @rob05c  good call out I will do that

ocket8888: Oh I missed that. Sorry, reviewing on my phone atm.

mhoppa: Actually the method I am calling already formats error with context https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/cachegroup/queueupdate.go#L108 so just bubbled that up

traffic_ops/traffic_ops_golang/cachegroup/parameters.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (cgparam *TOCacheGroupParameter) Read() ([]interface{}, error, error, int) {
+	queryParamsToQueryCols := cgparam.ParamColumns()
+	parameters := cgparam.APIInfo().Params
+	where, orderBy, pagination, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(parameters, queryParamsToQueryCols)
+	if len(errs) > 0 {
+		return nil, util.JoinErrs(errs), nil, http.StatusBadRequest
+	}
+
+	cgID, err := strconv.Atoi(parameters[CacheGroupIDQueryParam])
```
ocket8888: This is actually handled for you - the "id" in the path is parsed into the `APIInfo` structure's `IntParams` map. You can test that by just requesting `/cachegroups/notanumber/parameters`, which gives back the error "id cannot parse to integer"; it never hits this error handling.

mhoppa: Not in the generic case when using ReadHandler -> https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/api/shared_handlers.go#L122 it passes in nil into NewAPI for int params

ocket8888: You're right, I just assumed it was doing that because it was parsing it. That's actually happening above in `dbhelpers.BuildWhereAndOrderByAndPagination`, which doesn't appear to store the converted value anywhere... It's a real shame we can't just do that parse once.

Issue #3443, opened 2019-03-28
Undocumented TrafficOps endpoints and fields


moltzaum:
At the time of creating this issue, all the following endpoints were undocumented.
I can update the tickbox to be checked for each endpoint that is updated. Note: This may not be a comprehensive list.

**specific fields**
- [ ] user does not document lastUpdated for POST
- [ ] regions does not document lastUpdated for POST

**whole endpoints**
- [ ] GET steering
- [ ] POST statuses
- [ ] POST type
- [ ] PUT type
- [ ] DELETE type
- [ ] POST roles
- [ ] PUT roles
- [ ] DELETE roles
- [ ] DELETE divisions
- [ ] DELETE regions
- [ ] GET/POST/PUT/DELETE deliveryservice_requests
    - Note: A POST method is documented for `deliveryservices/request` (implemented in perl)
    - I'm not sure whether or not this supposed to be a different endpoint.

Note: most cases where an id is used I mean a path parameter because it is understood, but there are exceptions like `roles`, where only query parameters are used.

____

**Other incomplete documentation**
- [ ] `deliveryservices/dnsseckeys/generate`
    There is no example request structure

ocket8888:
Related: #2726

ocket8888:
I think this is now more fully captured and is just a duplicate of #4290 . Hopefully that one will get resolved before a year goes by and it's out of date.

PR #3901, opened 2019-09-05
Fixed runtime panic when creating roles, added docs

### LINKED ISSUE #3891: Internal Server Error message when create a new role
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When I created a new role, with the 30 Privilege Level, it displayed the danger message "Internal Server Error". The role still created.
## Expected / new behavior:

It should displays "Role created" message instead.

ocket8888:
- [x] This PR fixes #3891 and partially addresses #3443 

Fixes runtime panics caused by requests with invalid payloads - the endpoint handler required a `capabilities` array but didn't check for it until after a role had already been created/replaced, thus returning an error message while still seemingly completing the operation successfully. To solve this, the `capabilities` array was made optional in both `POST` (where its absence will result in simply no capabilities added to the Role) and `PUT` requests (where its absence will result in capabilities not being updated on the Role - **NOT** deleting all existing capabilities, which is what would happen if an empty array was submitted).

As the `POST`, `PUT` and `DELETE` methods of the endpoint were undocumented, I also took the opportunity to fill those in - thereby documenting the new behavior. I also added some missing query parameters to the documentation for the `GET` method.

- Documentation
- Traffic Ops

Verify existing tests pass. Build and run Traffic Ops/DB/Portal (CDN-in-a-Box recommended) and verify that admin users can create roles of privilege levels less than or equal to 30 (admin's privilege level) successfully, and no erroneous error message is reported. Build and read the documentation to ensure that it is accurate, comprehensible, and free of guideline violations and/or spelling/grammar mistakes.

- master (c18d535)
- 3.0.1
- 3.0.2 (RC2)
- 3.1.0 (pre-release candidate)

- [x] This PR adds no additional tests, but tests for the endpoint already exist

### DIFF
--- a/docs/source/api/roles.rst
+++ b/docs/source/api/roles.rst
@@ -21,7 +21,7 @@
 
 ``GET``
 =======
-Retrieves all user roles.
+Retrieves all user :term:`Roles`.
 
 :Auth. Required: Yes
 :Roles Required: None
@@ -34,6 +34,12 @@ Request Structure
 	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
 	| Name      | Required | Description                                                                                                   |
 	+===========+==========+===============================================================================================================+
+	| id        | no       | Return only the :term:`Role` identified by this integral, unique identifier                                   |
+	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
+	| name      | no       | Return only the :term:`Role` with this name                                                                   |
+	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
+	| privLevel | no       | Return only those :term:`Roles` that have this privilege level                                                |
+	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
 	| orderby   | no       | Choose the ordering of the results - must be the name of one of the fields of the objects in the ``response`` |
 	|           |          | array                                                                                                         |
 	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
@@ -48,12 +54,25 @@ Request Structure
 	|           |          | defined to make use of ``page``.                                                                              |
 	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
 
+.. code-block:: http
+	:caption: Request Example
+
+	GET /api/1.4/roles?name=admin HTTP/1.1
+	Host: trafficops.infra.ciab.test
+	User-Agent: curl/7.47.0
+	Accept: */*
+	Cookie: mojolicious=...
+
 Response Structure
 ------------------
-:description: A description of the role
-:id:          The integral, unique identifier for this role
-:name:        The name of the role
-:privLevel:   An integer that allows for comparison between roles
+:capabilities: An array of the names of the Capabilities given to this :term:`Role`
+
+	.. versionadded:: 1.3
+
+:description: A description of the :term:`Role`
+:id:          The integral, unique identifier for this :term:`Role`
+:name:        The name of the :term:`Role`
+:privLevel:   An integer that allows for comparison between :term:`Roles`
 
 .. code-block:: http
 	:caption: Response Example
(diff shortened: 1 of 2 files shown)


traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -122,9 +122,11 @@ func (role *TORole) Create() (error, error, int) {
 	}
 
 	//after we have role ID we can associate the capabilities:
-	userErr, sysErr, errCode = role.createRoleCapabilityAssociations(role.ReqInfo.Tx)
-	if userErr != nil || sysErr != nil {
-		return userErr, sysErr, errCode
+	if role.Capabilities != nil && *role.Capabilities != nil && len(*role.Capabilities) > 0 {
```
mhoppa: Nit but this can be simplified to `role.Capabilities != nil && len(*role.Capabilities) > 0`

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -177,12 +179,16 @@ func (role *TORole) Update() (error, error, int) {
 	if userErr != nil || sysErr != nil {
 		return userErr, sysErr, errCode
 	}
+
 	// TODO cascade delete, to automatically do this in SQL?
-	userErr, sysErr, errCode = role.deleteRoleCapabilityAssociations(role.ReqInfo.Tx)
-	if userErr != nil || sysErr != nil {
-		return userErr, sysErr, errCode
+	if role.Capabilities != nil && *role.Capabilities != nil {
```
mhoppa: do you need to check for len of *role.Capabilities?

docs/source/api/roles.rst
```diff
@@ -65,19 +84,236 @@ Response Structure
...
+				"text": "role was updated.",
+				"level": "success"
+			}
+		],
+		"response": {
+			"id": 5,
+			"name": "test",
+			"description": "quest",
+			"privLevel": 29,
+			"capabilities": null
```
mhoppa: do we return null responses to our users instead of empty throughout our API? I usually dont like doing that as then client have to code against it but if we use it many other places then this is fine

traffic_ops/traffic_ops_golang/role/roles.go
ocket8888: Yeah, so I've heard. Never knew that about Go, makes life much easier

traffic_ops/traffic_ops_golang/role/roles.go
ocket8888: Not to fix the bug.

Sounds like it might be a good idea, though, as it's probably doing extra work otherwise

docs/source/api/roles.rst
ocket8888: No idea. Generally I think an empty list would be better, but this documentation is accurate to how the endpoint behaves (which I don't *believe* I changed in this PR) and that's all I was attempting to accomplish.

mhoppa: fine to leave - might be good to do a sweep through all our endpoint for 2.0 and cleanup any null responses

ocket8888: The 2.0 todo list is long

traffic_ops/traffic_ops_golang/role/roles.go
ocket8888: Oh, no actually, you still need to do that. If the Role had associated capabilities and the user updated it to have none, you still need to delete those associations.

mhoppa: ahh - makes sense 👍

ocket8888:
retest this please

PR #3903, opened 2019-09-05
Rewrote /user/logout to Go

### LINKED ISSUE #3841: Rewrite /user/logout to Go
## I'm submitting a ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/user/logout` is handled by Perl

## Expected / new behavior:
`/user/logout` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `POST`

N/A

ocket8888:

- [x] This PR fixes #3841 

Rewrote /user/logout to Go

- Traffic Ops

functionality is not changing, so no docs

Authenticate as normal, then just logout. Normally, this would be done in Traffic Portal through the button on the sidebar. Then just observe that you're not authenticated - no endpoints (except `/ping` ) should respond with anything other than an "Unauthenticated" error

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/login/logout_test.go
+++ b/traffic_ops/traffic_ops_golang/login/logout_test.go
@@ -25,8 +25,6 @@ import "net/http/httptest"
 import "testing"
 import "time"
 
-// import "github.com/apache/trafficcontrol/lib/go-log"
-
 import "github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 import "github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
 import "github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/config"
--- a/traffic_ops/traffic_ops_golang/login/logout_test.go
+++ b/traffic_ops/traffic_ops_golang/login/logout_test.go
@@ -1,0 +1,159 @@
+package login
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


traffic_ops/traffic_ops_golang/login/logout_test.go
```diff
@@ -0,0 +1,159 @@
...
+ * under the License.
+ */
+
+import "context"
+import "net/http"
+import "net/http/httptest"
+import "testing"
+import "time"
+
+// import "github.com/apache/trafficcontrol/lib/go-log"
```
mhoppa: nit remove ?

Also another nit should imports be in import block here and above?

traffic_ops/traffic_ops_golang/login/logout_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+		if parsedCookie.AuthData != testUser.UserName {
+			t.Errorf("Incorrect user parsed from cookie; expected '%s' but got: %s", testUser.UserName, parsedCookie.AuthData)
+		}
+	}
+
+	if !cookieFound {
+		t.Errorf("Expected handler to set the '%s' cookie, but it didn't", tocookie.Name)
+	}
+}
```
mhoppa: great test case!

traffic_ops/traffic_ops_golang/login/logout_test.go
ocket8888: I don't like setting precedents for arbitrary stylistic changes without a style guide. As a Python programmer this style is most pleasing to me. I'm not saying I'd refuse to put them in blocks, just explaining my choice. If and when (hopefully when) there _is_ a style guide, I'll be more than happy to conform to it.
That comment should absolutely be removed, though.

mhoppa: 👍 completely agree on the style guide hence the comment being a nit/question and still approval of the PR.

FWIW per golangs _style guide_ they have the imports as a block https://github.com/golang/go/wiki/CodeReviewComments#imports

PR #3906, opened 2019-09-06
Fix ordering in overview docs section

mhoppa:

Fix ordering in overview docs section

Introduction was not first which makes it confusing to the end user as they come in 

- Documentation

Build docs locally and look at ordering of overview section

- [x] This PR does not include tests
- [x] This PR does not include an update to CHANGELOG.md as such an update is not necessary

### DIFF
--- a/docs/source/overview/index.rst
+++ b/docs/source/overview/index.rst
@@ -19,15 +19,14 @@ Traffic Control Overview
 Introduces the Traffic Control architecture, components, and their integration.
 
 .. toctree::
-	:maxdepth: 2
-	:glob:
+	:maxdepth: 3
 
 	introduction
-	delivery_services
-	profiles_and_parameters
 	traffic_ops
 	traffic_router
 	traffic_monitor
 	traffic_stats
 	traffic_portal
 	traffic_vault
+	delivery_services
+	profiles_and_parameters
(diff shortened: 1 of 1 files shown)


docs/source/overview/index.rst
```diff
@@ -22,4 +22,12 @@ Introduces the Traffic Control architecture, components, and their integration.
 	:maxdepth: 2
 	:glob:
 
-	*
+	introduction
+	delivery_services
+	profiles_and_parameters
```
ocket8888: I think these two ought to go after the component descriptions. The introduction doesn't mention them, so it's a weird break in the flow.

docs/source/overview/index.rst
```diff
@@ -22,4 +22,12 @@ Introduces the Traffic Control architecture, components, and their integration.
 	:maxdepth: 2
 	:glob:
```
ocket8888: This isn't globbing anymore, so you can get rid of this option.

docs/source/overview/index.rst
```diff
@@ -22,4 +22,12 @@ Introduces the Traffic Control architecture, components, and their integration.
 	:maxdepth: 2
```
ocket8888: This isn't exactly related to what you're trying to fix, but if you change this to `3` I think the properties of Parameters and Profiles would appear in the sidebar. Which could be nice.

PR #3908, opened 2019-09-06
Add deprecated message to cachegroup/:parameter_id/parameter resp

mhoppa:

- Traffic Ops

There is not a change to documentation as it is already mark deprecated in there.

Start up Perl TO locally with change and make a request to it. On success you should receive back a response that has an alert section with deprecate messaging

- [x] This PR does not includes tests
- [x] This PR does include an update to CHANGELOG.md as such an update is not necessary

### DIFF
--- a/traffic_ops/app/lib/API/Cachegroup.pm
+++ b/traffic_ops/app/lib/API/Cachegroup.pm
@@ -380,7 +380,7 @@ sub by_parameter_id {
 		}
 	}
 
-	$self->success( { cachegroups => \@data } );
+	$self->success( { cachegroups => \@data }, "API path is deprecated use /cachegroupparameters instead");
 }
 
 sub available_for_parameter {


ocket8888:
I believe deprecations normally have to go through the mailing list. You probably wanna ask about Cache Group Parameters in general.

mhoppa:
@ocket8888 I noticed it was already marked as such in the documentation https://traffic-control-cdn.readthedocs.io/en/latest/api/cachegroup_parameterID_parameter.html

Does that mean it has already gone through mailing list?

rawlinp:
Cache Group Params in general probably can't be deprecated since I'm pretty sure we actually use and depend on them (despite the absurdity of associating parameters to cachegroups).

About the mailing list, we might want to group up all the endpoints we think we should deprecate in 1.1 and discuss the entire set at once on the mailing list. It will probably get pretty messy to discuss the deprecation of API endpoints one at a time on the mailing list. We might get better feedback with a full list too, because users can just take the whole list and check their usage of all the deprecated endpoints at once.

ocket8888:
RE: the docs deprecation stuff: I did that and I shouldn't have. When I was writing that endpoint I'm sure I asked in an internal Slack channel "what is this?" and was told "we don't use that anymore", so I added the deprecation notice without going through the proper procedure.

I like @rawlinp's idea about a bulk deprecation thread. Seems it'll save a lot of time.

traffic_ops/app/lib/API/Cachegroup.pm
```diff
@@ -380,7 +380,7 @@ sub by_parameter_id {
 		}
 	}
 
-	$self->success( { cachegroups => \@data } );
+	$self->success( { cachegroups => \@data }, "API path is deprecated use /cachegroupparameters instead");
```
ocket8888: "Success" is a bit of a weird type of message for something that's more of a warning about a deprecation. This should use the "warning" level instead, imo.

mitchell852: what are the alert "levels"? success, error, warning and info?

mitchell852: i thought you could return an array of alerts. if so, what about 1 success message (for the successful assignment of a param to a cg) and 1 warning message with the deprecation message? or is that weird?

mhoppa: No I think that is what we want to do! THat would match the work that was done #3954

mhoppa:
Closing pull request as how we are approaching deprecation is being decided in the mailing list and even if we did decide to deprecate in perl this PR would need some work

PR #3914, opened 2019-09-09
TM: Report time for full body download for _astats

### LINKED ISSUE #1923: Health and Stats delay different than Traffic Monitor Java
In previous version of Traffic Monitor (2.1, java). The "query_time" was an indication on how long the _astats took to download. 

In the Go version of Traffic Monitor, the Health and Stats is the delay to receive the header response since it is chunked.

We have cache servers that have bad NIC's which Traffic Monitor would detect even at idle speed. This doesn't show anymore.

The Health and Stats query time should include time finishing downloading the body.

smalenfant:

Reverting back to Traffic Monitor <= 2.1 behavior regarding health/stats query reported time (full body download) 

- [x] This PR fixes #1923 

- Traffic Monitor

You should see increased latency for both Health and Stats query time. Mostly on the stats side.

This is a small regression from Traffic Monitor 2.2 and up.

- master
- 3.x

### DIFF
--- a/traffic_monitor/poller/poller_type_http.go
+++ b/traffic_monitor/poller/poller_type_http.go
@@ -107,21 +107,26 @@ func httpPoll(ctxI interface{}, url string, host string, pollID uint64) ([]byte,
 	req.Host = host
 	startReq := time.Now()
 	resp, err := ctx.Client.Do(req)
-	reqEnd := time.Now()
-	reqTime := reqEnd.Sub(startReq) // note this is essentially the roundtrip, not the body transfer time.
 	if err != nil {
+		reqEnd := time.Now()
+		reqTime := reqEnd.Sub(startReq) // note this is the time to transfer the entire body, not just the roundtrip
 		return nil, reqEnd, reqTime, fmt.Errorf("id %v url %v fetch error: %v", ctx.PollerID, url, err)
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return nil, reqEnd, reqTime, fmt.Errorf("id %v url %v fetch error: bad HTTP status: %v", ctx.PollerID, url, resp.StatusCode)
+		reqEnd := time.Now()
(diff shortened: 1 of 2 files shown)


smalenfant:
Thanks @rob05c for assistance with this fix.

ocket8888:
Does this PR fix a security vulnerability or include a database migration?

smalenfant:
@ocket8888 Sorry, it doesn't. Checked the box...

smalenfant:
I just tested this in production on 3.1. Health time did not increase (expected) and Stat time increased (expected since it contains DS information).

traffic_monitor/poller/poller_type_http.go
```diff
@@ -107,21 +107,26 @@ func httpPoll(ctxI interface{}, url string, host string, pollID uint64) ([]byte,
...
+		reqTime := reqEnd.Sub(startReq) // note this is the time to transfer the entire body, not just the roundtrip
 		return nil, reqEnd, reqTime, fmt.Errorf("id %v url %v fetch error: %v", ctx.PollerID, url, err)
 	}
 	defer resp.Body.Close()
 
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return nil, reqEnd, reqTime, fmt.Errorf("id %v url %v fetch error: bad HTTP status: %v", ctx.PollerID, url, resp.StatusCode)
+		reqEnd := time.Now()
+		reqTime := reqEnd.Sub(startReq) // note this is the time to transfer the entire body, not just the roundtrip
+    return nil, reqEnd, reqTime, fmt.Errorf("id %v url %v fetch error: bad HTTP status: %v", ctx.PollerID, url, resp.StatusCode)
```
ocket8888: this looks like it needs to be run through `go fmt`, but the logic looks solid as far as I can tell at a glance. Once it's formatted I'll stand up a CDN-in-a-Box and take a look at it in full.

smalenfant: Ok, I believe this is fixed.

PR #3919, opened 2019-09-11
TP: adds a better view for comparing/managing params of 2 profiles

### LINKED ISSUE #2391: TP - Improve profile diffing
Currently a profile diff will just compare a sorted version of the json of 2 profiles and then display them side by side highlighting the differences.  I find the old TO diff much more useful where

1) Anything that is the same is stripped in the diff, showing only differences
2) The differences shown are just 'param name' - 'value' - 'value', so you can easily see the name of the param and the value differences between the 2 profiles
3) Doesnt show the full json, I dont think this is particularly useful anyway

mitchell852:

Provides the ability to easily compare parameters across 2 profiles and easily add or remove parameters from each profile.

A blueprint is available for this new feature: https://github.com/apache/trafficcontrol/blob/master/blueprints/profile-param-compare-manage.md

- [x] This PR fixes #2391 

- Traffic Portal

- Launch TP
- Navigate to https://tp.domain.com/#!/profiles
- Select "Compare Profiles" under the More button
- Select 2 profiles to compare
- Ensure each profile has the proper parameters "checked". (you can double check by checking the parameters for the profile at https://tp.domain.com/#!/profiles/{profile_id}/parameters)
- add or remove parameters from each profile; click "Update"
- ensure the proper parameters are assigned to each profile after confirmation.


traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
```diff
@@ -0,0 +1,213 @@
...
+									promises.push(profileParameterService.linkProfileParameters(profile2, _.pluck($scope.selectedProfile2Params, 'id')));
+								}
+
+								if (promises.length > 0) {
+									$q.all(promises)
+										.then(
+											function(result) {
+												let messages = [];
+												for (let i = 0; i < result.length; i++) {
+													messages = _.union(messages, result[i].data.alerts);
```
ocket8888: Instead of using `_.union`, consider using the built-in [`Set` class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) like so:

```javascript
const messages = new Set(); // 'messages' isn't being reassigned, so it's constant
for (let r of result) {
	for (let a of r.data.alerts) {
		messages.add(a);
	}
}
// You can submit it as an array using Array.from:
messageModel.setMessages(Array.from(messages), false);
```
For bonus points, you can use [`Array.prototype.flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) to flatten it into a one-dimensional array prior to iteration:
```javascript
const messages = new Set();
// This loop directly iterates every alert of every response
for (let a of result.flatMap(function(x){return x.data.alerts})) {
	messages.add(a);
}
messageModel.setMessages(messages, false);
```

traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
```diff
@@ -0,0 +1,213 @@
...
+	$scope.dirty = false;
+
+	$scope.profile1 = profile1;
+	$scope.profile2 = profile2;
+	$scope.profilesParams = profilesParams;
+
+	$scope.selectedProfile1Params = [];
+	$scope.selectedProfile2Params = [];
+
+	$scope.selectedParams = _.map(profilesParams, function(param) {
```
ocket8888: instead of using `_.map` you can just directly use the [`Array.prototype.map` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).

traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
```diff
@@ -0,0 +1,213 @@
...
+		let isAssignedToProfile1 = _.find(profile1.params, function(profile1param) { return profile1param.id == param.id }),
+			isAssignedToProfile2 = _.find(profile2.params, function(profile2param) { return profile2param.id == param.id });
```
ocket8888: I don't think `_.find` does what you're looking for. It returns the first value in the array that yields a truthy value from the predicate function. I think what you want to know is whether or not such a value *exists* - you don't care about the value. Instead, consider using [the built-in `Array.prototype.some` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) like so:
```javascript
// also note the use of triple-equals (===)
const isAssignedToProfile1 = profile1.params.some(function(p){return p.id === param.id});
//...
```

traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
```diff
@@ -0,0 +1,213 @@
...
+		} else {
+			param['origSelected2'] = false;
+			param['selected2'] = false;
+		}
+
+		return param;
+	});
+
+	$scope.updateSelectedCount = function(profNum) {
+		$scope['selectedProfile' + profNum + 'Params'] = _.filter($scope.selectedParams, function(param) { return param['selected' + profNum] == true; } );
```
ocket8888: Instead of using `_.filter`, consider using [the built-in `Array.prototype.filter` method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) like so:
```javascript
// assuming 'profNum' is a Number, it should be parsed to avoid type coercion mishaps
const num = parseInt(profNum, 10);
$scope['selectedProfile'+num+'Params'] = $scope.selectedParams.filter(function(p){return param['selected'+num];});
```

traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
```diff
@@ -0,0 +1,213 @@
...
+				}
+			}
+		});
+		modalInstance.result.then(function() {
+			$scope.refresh();
+		});
+	};
+
+	$scope.update = function() {
+		confirmUpdateProfiles();
```
ocket8888: I think you can actually just do
```javascript
$scope.update = confirmUpdateProfiles;
```

traffic_portal/app/src/common/modules/table/profilesParamsCompare/table.profilesParamsCompare.tpl.html
```diff
@@ -0,0 +1,58 @@
...
+    <div class="x_title">
+        <ol class="breadcrumb pull-left">
+            <li><a ng-click="navigateToPath('/profiles')">Profiles</a></li>
+            <li><a ng-click="navigateToPath('/profiles/' + profile1.id)">{{::profile1.name}}</a></li>
+            <li><a ng-click="navigateToPath('/profiles/' + profile2.id)">{{::profile2.name}}</a></li>
+            <li class="active">Compare Parameters</li>
+        </ol>
+        <div class="pull-right">
+            <button class="btn btn-default" title="Restore" ng-show="dirty" ng-click="restore()">Restore</button>
+            <button class="btn btn-primary" title="Update" ng-disabled="!dirty" ng-click="update()">Update</button>
```
ocket8888: These buttons implicitly have `type="submit"`, which is fine for "Update", but the "Restore" button doesn't effect changes - it resets the view. It should have `type="reset"`.

traffic_portal/app/src/styles/main.scss
```diff
@@ -183,6 +183,13 @@ table {
   margin-right: 5px;
 }
 
+input[type="checkbox"].dirty {
+  -moz-box-shadow:    inset 0 0 10px red;
+  -webkit-box-shadow: inset 0 0 10px red;
+  box-shadow:         inset 0 0 10px red;
```
ocket8888: Red to me means "invalid"/"wrong" - I think a light blue glow would be better. You could set the "dirty" class on the entire row to make it glow blue, if you're concerned about a clash with host OS theming colors.

traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
mitchell852: it should have the same effect but i don't mind getting rid of the underscore method.

traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
mitchell852: good point. i think i had more stuff in there at one time.

traffic_portal/app/src/styles/main.scss
mitchell852: yeah, understood. i'll play around with colors more.

traffic_portal/app/src/common/modules/table/_table.scss
```diff
@@ -96,3 +96,10 @@ td span {
     padding: 10px 20px;
     overflow-y: auto;
 }
+table.dataTable tbody td {
+    word-break: break-word;
+    vertical-align: top;
+}
```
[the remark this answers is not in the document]
mitchell852: i'm going to limit this css to only parameter tables.

traffic_portal/app/src/common/modules/table/profilesParamsCompare/table.profilesParamsCompare.tpl.html
mitchell852: i really don't want any implicit meaning on either so i think best to just set to `type=button`

traffic_portal/app/src/styles/main.scss
mitchell852: changed to green

traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
mitchell852: Error: Parsing file /traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js

i guess it didn't like the `of` loop or something so just stuck with boring old for loop

traffic_portal/app/src/common/modules/table/profilesParamsCompare/TableProfilesParamsCompareController.js
ocket8888: there should be an issue for that, you wanna open one?

mitchell852: https://github.com/apache/trafficcontrol/issues/3928

PR #3921, opened 2019-09-13
Rewrite profiles/{id}/export to golang

### LINKED ISSUE #3814: Rewrite /profiles/{{id}}/export to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/profiles/{{id}}/export` is handled by Perl

## Expected / new behavior:
`/profiles/{{id}}/export` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`

N/A

mhoppa:

This PR rewrites the /cachegroups/{id}/unassigned_parameters endpoint to Go

- [x] This PR fixes #3814 

- Documentation
- Traffic Ops

Bring up TO locally and hit the endpoint to ensure parity between perl implementation and newly written Golang implementation 

The API tests and/or unit tests for TO can be run to verify the PR.

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/profile/profile_export.go
+++ b/traffic_ops/traffic_ops_golang/profile/profile_export.go
@@ -1,0 +1,144 @@
+package profile
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/jmoiron/sqlx"
+
+	"github.com/apache/trafficcontrol/lib/go-log"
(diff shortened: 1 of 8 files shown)


traffic_ops/traffic_ops_golang/profile/profile_export.go
```diff
@@ -0,0 +1,144 @@
...
+c.name as cdn,
+parm.name as parm_name,
+parm.config_file as parm_config_file,
+parm.value as parm_value
+FROM profile prof
+JOIN cdn c ON prof.cdn = c.id
+LEFT JOIN profile_parameter as pp ON pp.profile = prof.id
+LEFT JOIN parameter as parm ON parm.id = pp.parameter
+WHERE prof.id=:id`
+	return query
```
ocket8888: I feel like the above looping could be cleaner if there was an Array in this query for the parameters. Did you consider trying that?

mhoppa: I have not. I can look into it though.

ocket8888: I'm not 100% sure given the use of `sqlx` that that'll work cleanly, but I do think it's worth looking into.

traffic_ops/traffic_ops_golang/profile/profile_export.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	// Get Profile Response
+	exportedProfileResp, err := getExportProfileResponse(profileID, inf.Tx)
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, err)
+		return
+	}
+
+	api.WriteRespRaw(w, r, exportedProfileResp)
```
ocket8888: Since this is expressly intended as a download, do you think you could set the `Content-Disposition` as appropriate? It's not what Perl did, but we don't document headers so it ought to be fine.

mhoppa: I would be fine with that

mitchell852:
@ocket8888 @mhoppa - did you make sure this still works in traffic portal?

mhoppa:
@mitchell852 will do that test right now! Great call out.

mhoppa:
@mitchell852 works as expected in the portal

PR #3925, opened 2019-09-17
Added SMTP configuration and APIInfo Email method

ocket8888:

- Traffic Ops

The configuration file is undocumented. It shouldn't be, but doing so is outside the scope of this PR.

Create a configuration file with a valid SMTP server configuration and try to start Traffic Ops. Config should parse fine.

- [x] Tests are unnecessary
- [x] I have explained why documentation is absent

### DIFF
--- a/traffic_ops/traffic_ops_golang/api/api.go
+++ b/traffic_ops/traffic_ops_golang/api/api.go
@@ -27,12 +27,14 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/smtp"
 	"regexp"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-rfc"
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/lib/go-util"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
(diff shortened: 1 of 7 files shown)


traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -369,6 +371,33 @@ func (inf *APIInfo) Close() {
...
+// The cfg parameter is used to set things like the "From" field, as well as for connection
+// and authentication with an external SMTP server.
+// SendMail returns (in order) an HTTP status code, a user-friendly error, and an error fit for
+// logging to system error logs. If either the user or system error is non-nil, the operation failed,
+// and the HTTP status code indicates the type of failure.
+func SendMail(to rfc.EmailAddress, msg []byte, cfg *config.Config) (int, error, error) {
+	if !cfg.SMTP.Enabled {
+		return http.StatusServiceUnavailable, errors.New("SMTP is not enabled!"), nil
+	}
+	auth := smtp.PlainAuth("", cfg.SMTP.User, cfg.SMTP.Password, strings.Split(cfg.SMTP.Address, ":")[0])
```
mhoppa: do you need to check that cfg.ConfigTO.EmailFrom is defined or not? Unsure if that causes an error in the SendMail func or not

traffic_ops/traffic_ops_golang/config/config.go
```diff
@@ -100,6 +103,22 @@ type ConfigTrafficOpsGolang struct {
 	CRConfigEmulateOldPath bool `json:"crconfig_emulate_old_path"`
 }
 
+// ConfigTO contains information to identify Traffic Ops in a network sense.
+type ConfigTO struct {
+	BaseURL               rfc.URL          `json:"base_url"`
```
mhoppa: Should we be using consistent types for our URL representations? in Config we use *url.URL for URL. Unsure of the difference

ocket8888: `rfc.URL` is just a wrapper around `url.URL` that provides methods for encoding and decoding as JSON. You can, in fact, cast it to `url.URL`.

mhoppa: ahh duh - sorry about that :) need to look closer :)

traffic_ops/traffic_ops_golang/api/api.go
ocket8888: No, that field is mandatory in the configuration. If it doesn't exist, Traffic Ops won't start.

mhoppa: can you point me to where it is marked as required and forces Traffic Ops not to start?

ocket8888: I think it's probably most clearly illustrated on [Line #175 in `traffic_ops/traffic_ops_golang/config/config.go`](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/config/config.go#L175). If the config fails to load startup is blocked. ConfigTO and its sub-fields are mandatory, so their absence would cause a failure to load.

mhoppa: I dont think that makes sure that the fields are defined it just does an unmarshal.

If we want to make it required to be defined doesn't it need to go in https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/config/config.go#L249

ocket8888: Yes, you're right. I thought if the JSON parser didn't find it and the passed pointer wasn't a pointer to a pointer (so it can't set the underlying pointer to `nil`) then it would throw an error. But that's not true at all.

ocket8888: So I fixed this by making ConfigTO and its children required parts of the config, which ought to be fine because Perl always required it anyway.

PR #3927, opened 2019-09-17
Fix using http2 between TO and TM when setting proxy

### LINKED ISSUE #3902: TP: internal server error on dashboard page
## I'm submitting a ...

- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [X] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

On login or navigation to Dashboard page. Internal server error is displayed. 
end point api/1.4/cdns/capacity returns 500  Internal server error 

## Expected / new behavior:

No error should be displayed.

mhoppa:

- [x] This PR fixes #3902

If TM has an https proxy set that is used by TO to connect it will fail as golang uses http2 be default in that case. This PR overrides that behavior based on the net godocs https://golang.org/src/net/http/doc.go.

There is not documentation changes as it is a bug fix and there is no tests as we dont have TM set up for our API tests 

- Traffic Ops

Setup TM proxy with HTTPs ensure endpoint cdns/capacity work

- master (5b018e2b)

- [x] This PR does not include tests OR I have explained why tests are unnecessary


mitchell852:
is there not a good way to do this globally? i'm just worried that other PRs will sneak thru without this type of change applied and generate a lot of bugs. i.e. #2305

mhoppa:
Agreed with you @mitchell852. One question is what services besides TM/TR does the TO talk to? If we put getting http clients for each of them into a common method we can enforce it a little better as https://github.com/apache/trafficcontrol/pull/2305/files#diff-6e647a5b8d470f399423ff712efb0050R40 does.

As we talked we can set it globally per https://golang.org/src/net/http/doc.go via `GODEBUG=http2client=0  # disable HTTP/2 client support` but I think we have agreed that feels a little cumbersome.

ocket8888:
It talks to TV and TS and a mail server (sort of and sometimes)

mhoppa:
retest this please

PR #3931, opened 2019-09-20
Gives admins the ability to mark delivery service requests as complete

### LINKED ISSUE #3871: Allow admins to 'complete' any delivery service request
## I'm submitting a ...

- [ ] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Currently, only the user that "fulfills" (persists the DS changes to the DB) a delivery service request (DSR) can mark it as 'complete'. This was a design decision to prevent other users from marking a DSR as complete prematurely (ie. before a snapshot/queue was performed). 

However, in practice, you end up with DSR's that can't be marked 'complete' if the fulfilling user is not available. Since an open DSR blocks subsequent changes to a DS, this can be problematic.
 
## Expected / new behavior:
Consider giving admin users the ability to 'complete' any/all open DSRs.

Turn ds requests on in traffic_portal_properties.json and then attempt to make a change to a DS in Traffic Portal. That change will be captured in the form of a DSR. Fulfill the DSR and then let another user attempt to 'complete' the DSR. They will fail.

mitchell852:

Allows users with the admin role to mark fulfilled delivery services requests as "complete".  Before this change, only the user that fulfilled the delivery service request could mark it as "complete". This can cause problems if the fulfilling user is not available to mark it as complete thus leaving the delivery service request in an "open" state. Because a delivery service can only have one open ds request at a time, this would block any further changes to a delivery service. Now anyone with the admin role can "complete" ds requests and unblock the delivery service from future updates.

- [x] This PR fixes #3871 

- Traffic Portal

- enable ds requests in traffic_portal_properties.json
- launch TP
- login as user 1 (with portal role or higher) and update a delivery service which will create a delivery service request and set the request status to 'submit request for review and deployment'
- login as user 2 (with operations role or higher) and fulfill the delivery service request
- login as user 3 (with admin role) and mark the delivery service request as "complete"

(before this change, only user 2 (the fulfiller of the ds request) could mark the ds request as complete)

There are no UI tests to cover ds requests currently as ds requests are disabled by default and no mechanism to turn them on for testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added fields to traffic_portal_properties.json to configure SSO through OAuth.
 - Added field to cdn.conf to configure whitelisted URLs for Json Key Set URL returned from OAuth provider.
 - Improved [profile comparison view in Traffic Portal](https://github.com/apache/trafficcontrol/blob/master/blueprints/profile-param-compare-manage.md).
+- Issue #3871 - provides users with a specified role the ability to mark any delivery service request as complete.
 - Fixed Traffic Ops Golang POST servers/id/deliveryservice continuing erroneously after a database error.
 - Fixed Traffic Ops Golang POST servers/id/deliveryservice double-logging errors.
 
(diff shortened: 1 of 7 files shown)


mitchell852:
> So it looks like this protection for who can and can't fulfill a request is entirely front-end - why isn't that enforced by the API?

I agree 100%. Here's some history. Delivery service requests were intended to be a short-lived feature so it's less than perfect imo. Full self-service was to take its place at some point.

These are the valid statuses of a DSR: draft, submitted, pending, complete, rejected. 

The creator of a DSR needs to put the DSR in draft or submitted (DSRs in draft cannot be fulfilled, DSRs in submitted status are ready for review/fulfillment), therefore, the permission level for updating DSR status is "portal" but that allows portal users to mark DSRs as "complete or rejected" which doesn't make a lot of sense. That part is really the job of the reviewer.

Anyhow, long story, short the design of `PUT deliveryservice_requests/{id}/status` was probably not very good hence the UI is doing half the job.

ocket8888:
Wild

PR #3932, opened 2019-09-20
Rewrote /user/login/token to Go

### LINKED ISSUE #3840: Rewrite /user/login/token to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/user/login/token` is handled by Perl

## Expected / new behavior:
`/user/login/token` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `POST`

N/A

ocket8888:
- [x] This PR fixes #3840 
Rewrites `/user/login/token` to Go, and adds client code and a test. Also some minor documentation updates.

- Documentation
- Traffic Control Client (Go)
- Traffic Ops

Run the associated API/client integration test

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/client/session.go
+++ b/traffic_ops/client/session.go
@@ -101,6 +101,20 @@ func loginCreds(toUser string, toPasswd string) ([]byte, error) {
 	return js, nil
 }
 
+// loginToken gathers token login credentials for Traffic Ops.
+func loginToken(token string) ([]byte, error) {
+	form := tc.UserToken {
+		Token: token,
+	}
+
+	j, e := json.Marshal(form)
+	if e != nil {
+		e := fmt.Errorf("Error creating token login json: %v", e)
+		return nil, e
+	}
+	return j, nil
+}
+
 // Deprecated: Login is deprecated, use LoginWithAgent instead. The `Login` function with its present signature will be removed in the next version and replaced with `Login(toURL string, toUser string, toPasswd string, insecure bool, userAgent string)`. The `LoginWithAgent` function will be removed the version after that.
 func Login(toURL string, toUser string, toPasswd string, insecure bool) (*Session, error) {
 	s, _, err := LoginWithAgent(toURL, toUser, toPasswd, insecure, "traffic-ops-client", false, DefaultTimeout)
@@ -140,6 +154,29 @@ func (to *Session) login() (net.Addr, error) {
 	}
 
 	return remoteAddr, nil
+}
+
+func (to *Session) loginWithToken(token []byte) (net.Addr, error) {
+	path := apiBase + "/user/login/token"
+	resp, remoteAddr, err := to.rawRequest(http.MethodPost, path, token)
+	resp, remoteAddr, err = to.ErrUnlessOK(resp, remoteAddr, err, path)
+	if err != nil {
+		return remoteAddr, fmt.Errorf("requesting: %v", err)
+	}
+	defer resp.Body.Close()
+
+	var alerts tc.Alerts
+	if err := json.NewDecoder(resp.Body).Decode(&alerts); err != nil {
+		return remoteAddr, fmt.Errorf("decoding response JSON: %v", err)
+	}
+
+	for _, alert := range alerts.Alerts {
+		if alert.Level == tc.SuccessLevel.String() && alert.Text == "Successfully logged in." {
+			return remoteAddr, nil
+		}
+	}
+
+	return remoteAddr, fmt.Errorf("Login failed, alerts string: %+v", alerts)
 }
 
 // logout of Traffic Ops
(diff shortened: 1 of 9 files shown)


traffic_ops/client/session.go
```diff
@@ -142,6 +156,29 @@ func (to *Session) login() (net.Addr, error) {
...
+		return remoteAddr, fmt.Errorf("requesting: %v", err)
+	}
+	defer resp.Body.Close()
+
+	var alerts tc.Alerts
+	if err := json.NewDecoder(resp.Body).Decode(&alerts); err != nil {
+		return remoteAddr, fmt.Errorf("decoding response JSON: %v", err)
+	}
+
+	for _, alert := range alerts.Alerts {
```
mhoppa: would checking 2xx be better then string match on the alerts?

traffic_ops/client/session.go
```diff
@@ -207,6 +244,36 @@ func LoginWithAgent(toURL string, toUser string, toPasswd string, insecure bool,
...
+
+	jar, err := cookiejar.New(&options)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	client := http.Client {
+		Timeout: requestTimeout,
+		Transport: &http.Transport {
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: insecure},
```
mhoppa: do we use insecure on all our api tests?

traffic_ops/testing/api/v14/loginfail_test.go
```diff
@@ -68,6 +69,29 @@ func LoginWithEmptyCredentialsTest(t *testing.T) {
 	}
 }
 
+func LoginWithTokenTest(t *testing.T) {
```
mhoppa: 👏

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -115,6 +115,47 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
...
+		respBts, err := json.Marshal(tc.CreateAlerts(tc.SuccessLevel, "Successfully logged in."))
+		if err != nil {
+			sysErr := fmt.Errorf("Marshaling response: %v", err)
+			errCode := http.StatusInternalServerError
+			api.HandleErr(w, r, nil, errCode, nil, sysErr)
+			return
+		}
+
+		w.Header().Set(tc.ContentType, tc.ApplicationJson)
+		w.Write(append(respBts, '\n'))
```
mhoppa: ?

traffic_ops/client/session.go
ocket8888: no, that's based on the value of the parameter given by the caller. So each program that uses the client provides the choice of whether or not to skip cert verification.

traffic_ops/traffic_ops_golang/login/login.go
ocket8888: You asking about the appended `'\n'` ? That's for POSIX compliance. A lot of clients tend to treat the API as a service for obtaining JSON files - hence the historic `.json` suffix on endpoints. But without a trailing newline, they aren't files.

traffic_ops/client/session.go
ocket8888: Possibly. This is a copy of the logic from `loginWithAgent`, and I think a change to that logic should be global so that logging in always checks for success the same way.

traffic_ops/traffic_ops_golang/login/login.go
mhoppa: makes sense but its not a file download? its a json response ? Do we do this across our other json responses?

ocket8888: No, but I don't think it hurts to add it - without a guarantee, of course. Either way it's valid JSON, and the logic for appending a newline is a single call to a builtin with a rune constant, so it shouldn't be hard to maintain.

rob05c: +1 on `\n`.

The trailing newline is still valid JSON, and makes it also a valid POSIX file. @ocket8888 has convinced me we ought to add it to all our endpoints, there's just no reason not to.

traffic_ops/client/session.go
```diff
@@ -207,6 +244,36 @@ func LoginWithAgent(toURL string, toUser string, toPasswd string, insecure bool,
 	return to, remoteAddr, nil
 }
 
+func LoginWithToken(toURL string, token string, insecure bool, userAgent string, useCache bool, requestTimeout time.Duration) (*Session, net.Addr, error) {
+	options := cookiejar.Options {
+		PublicSuffixList: publicsuffix.List,
+	}
+
+	jar, err := cookiejar.New(&options)
```
mhoppa: The build is failing because the jar variable is unused

ocket8888: ? It's not though... I use it on line 237

ocket8888: Ah, seems I forgot something. Should work now.

PR #3933, opened 2019-09-24
TP: adds a CSV link to all tables to enable export of table data to CSV

### LINKED ISSUE #3895: Add ability to export server data from Traffic Portal
## I'm submitting a ...
- [ ] bug report
- [ ] other

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Currently the only way to export server data is to copy/paste it out of the portal or to use the API.  Copy/pasting does not give you the ability to control which fields you are exporting, and using the API requires some basic knowledge on scripting/writing code.  Neither of these options are great, we should provide the ability to let a user export server data.  
## Expected / new behavior:

User can export server data to csv.  Traffic Portal would allow them to choose servers from a specific CDN, Cache group, etc as well as letting the user choose which fields they want to export.  This would be useful for external groups that use the Traffic Ops data for other purposes (e.g. network monitoring)

mitchell852:

A 'CSV' link has been added to all tables in TP that provides the user with the ability to download the table data (with the search filter applied) in csv format. 

- [x] This PR fixes #3895 

- Traffic Portal

Navigate to any table in TP and click the 'CSV' link at the top of the table. Verify that the contents of the table are downloaded in csv format.


traffic_portal/app/src/common/service/application/ApplicationService.js
```diff
@@ -38,6 +38,10 @@ var ApplicationService = function($rootScope, $anchorScroll, $http, messageModel
 
         // jquery DataTables default overrides
         $.extend(true, $.fn.dataTable.defaults, {
+            "dom": "lfBrtip",
+            "buttons": [
+                { "extend": "csv", "className": "btn-link", "titleAttr": "Export to CSV" }
+            ],
```
ocket8888: Is there any way this can be changed to a download icon like <kbd>⭳</kbd> or something? Failing that, "Export as CSV" or simply "Export" would be more clear than just "CSV", IMO.

traffic_portal/app/src/common/modules/table/profilesParamsCompare/table.profilesParamsCompare.tpl.html
```diff
@@ -48,8 +48,8 @@
                 <td data-search="^{{::pp.name}}$">{{::pp.name}}</td>
                 <td data-search="^{{::pp.configFile}}$">{{::pp.configFile}}</td>
                 <td data-search="^{{::pp.value}}$">{{::pp.value}}</td>
-                <td class="center" data-order="{{pp.selected1}}"><input type="checkbox" ng-change="onChange(1)" ng-class="{'dirty': isCheckboxDirty(pp, 1)}" ng-model="pp.selected1"></td>
-                <td class="center" data-order="{{pp.selected2}}"><input type="checkbox" ng-change="onChange(2)" ng-class="{'dirty': isCheckboxDirty(pp, 2)}" ng-model="pp.selected2"></td>
+                <td class="center" data-order="{{pp.selected1}}"><input type="checkbox" ng-change="onChange(1)" ng-class="{'dirty': isCheckboxDirty(pp, 1)}" ng-model="pp.selected1"><span hidden>{{pp.selected1}}</span></td>
+                <td class="center" data-order="{{pp.selected2}}"><input type="checkbox" ng-change="onChange(2)" ng-class="{'dirty': isCheckboxDirty(pp, 2)}" ng-model="pp.selected2"><span hidden>{{pp.selected2}}</span></td>
```
ocket8888: what's with the hidden span here?

mitchell852: it's to hide the span :P

mitchell852: it's because when the table is exported to csv, there is no value in this table cell. the export is not smart enough to figure out the value of the checkbox. instead, if i include this hidden field it is hidden from the UI but gets captured in the export.

ocket8888: > _"the export is not smart enough to figure out the value of the checkbox"_

That sounds like a dirty hack to me, but I think you'd be disinclined to write it better yourself?

traffic_portal/app/src/common/service/application/ApplicationService.js
mitchell852: i did try an icon and it looked kinda weird just hanging out on the top of the table. But yes, "Export as CSV" would be better. clarity is good.

traffic_portal/app/src/common/modules/table/profilesParamsCompare/table.profilesParamsCompare.tpl.html
mitchell852: it does feel hacky....but it is such a simple hack (so little code) i couldn't resist. plus, monkey patching the library felt more problematic than it's worth.

traffic_portal/app/src/common/service/application/ApplicationService.js
mitchell852: done

traffic_portal/app/src/common/modules/table/profilesParamsCompare/table.profilesParamsCompare.tpl.html
ocket8888: Oh I didn't mean patching the library, I meant using the download stuff from the utils service to make your own button.

PR #3939, opened 2019-09-27
fixed bug in CertificateRegistry

ajschmidt:
This PR adds exception handling and logging where the TR code interfaces with the tomcat-native library to create or modify the SSL Host data structures used by openssl. The exception handling prevents TR from saving host configurations when they do not get successfully written to the native structures in openssl. Therefore TR will continue to try and update them each time it polls for SSL cert changes. This PR also fixes a bug that may have been unregistering the native SSL host configurations with JMX causing them to become eligible for untimely garbage collection. 
This PR does not change any functionality of TR and does not affect the documentation. One unit test has been added to demonstrate that failed updates will cause TR to attempt the update again the next time it retrieves SSL certs from TO. 

- CDN in a Box
- Traffic Router

1. First start Traffic Router and wait for it to finish initializing.
2. Go into Traffic Portal and add a new certificate for an existing HTTPS delivery service. 
3. Wait for 5 minutes. There should be a log message in the TR logs indicating that new certificate has been loaded for the delivery service.
4. Use 'curl' to connect to the delivery service that has the updated certificate and then verify that the new certificate is being used by observing the 'curl' output. The expiration date in the curl output should match the expiration date set on the certificate. 
 
- master (79897553d46a)
- 3.0.0
- 3.0.1 (RC1)

### DIFF
--- a/traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
+++ b/traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
@@ -56,10 +56,10 @@ public class RouterNioEndpoint extends NioEndpoint {
 		}
 	}
 
-    synchronized private List<String> replaceSSLHosts(final Map<String, HandshakeData> sslHostsData) {
-        final Set<String> aliases = sslHostsData.keySet();
-        String lastHostName = "";
-        final List<String> failedUpdates = new ArrayList<>();
+	synchronized private List<String> replaceSSLHosts(final Map<String, HandshakeData> sslHostsData) {
+		final Set<String> aliases = sslHostsData.keySet();
+		String lastHostName = "";
+		final List<String> failedUpdates = new ArrayList<>();
 
 		for (final String alias : aliases) {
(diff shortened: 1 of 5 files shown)


traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
```diff
@@ -53,10 +56,10 @@ protected void initialiseSsl() throws Exception {
 		}
 	}
 
-	@SuppressWarnings({"PMD.NPathComplexity", "PMD.UseStringBufferForStringAppends"})
-	synchronized public void replaceSSLHosts(final Map<String, HandshakeData> sslHostsData) {
-		final Set<String> aliases = sslHostsData.keySet();
-		String lastHostName = "";
+    synchronized private List<String> replaceSSLHosts(final Map<String, HandshakeData> sslHostsData) {
```
rawlinp: Do you mind reformatting the additions here? This file appears to be mostly tab-indented rather than space-indented.

ajschmidt: Done!

PR #3940, opened 2019-09-30
Fix typo in GetProfileDS query used to generate cache.config

rob05c:
Fix typo from merge conflict

Fix typo from merge conflict

Endpoint already has tests, they caught this.

- Traffic Ops

curl ats config endpoints.

no versions


mhoppa:
Great find!! Are there API tests in place that would have caught this issue?

rob05c:
@mhoppa I thought it did cause the API tests to fail, but I was mistaken. I added tests: https://github.com/apache/trafficcontrol/pull/3941

They only test the 200, which catches this issue. They don't extensively test that the config is correct; but, we're moving to cache-side generation and these endpoints won't be used anymore, so it isn't worth adding lots of logic tests. The right place for more of those is the lib/go-atscfg unit tests.

PR #3947, opened 2019-10-01
Rewrite get federation deliveryservices to Go

### LINKED ISSUE #3793: Rewrite /federations/{{id}}/deliveryservices to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/federations/{{id}}/deliveryservices` is handled by Perl

## Expected / new behavior:
`/federations/{{id}}/deliveryservices` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`
- [x] `POST` (#2862)

N/A

mhoppa:

- [x] This PR fixes #3793

- Documentation
- Traffic Ops

Build TO with change locally and use the GET federations/{id}/deliveryservices endpoint. It should behave the exact way it did before. Another way to verify it is to run the API tests.

- [x] This PR includes tests


docs/source/api/federations_id_deliveryservices.rst
```diff
@@ -21,7 +21,7 @@
 
 ``GET``
 =======
-Retrieves :term:`Delivery Service`\ s assigned to a federation.
+Retrieves :term:`Delivery Service`\ s assigned to a :term:`Federation`.
```
ocket8888: Not necessary, but I'd appreciate it if while you're doing this if you could get rid of the `\ s`; `:term:` can handle plurals now.

traffic_ops/client/federation.go
```diff
@@ -78,3 +79,12 @@ func (to *Session) CreateFederationDeliveryServices(federationID int, deliverySe
 	inf, err := makeReq(to, http.MethodPost, apiBase+`/federations/`+strconv.Itoa(federationID)+`/deliveryservices`, jsonReq, &resp)
 	return inf, err
 }
+
+func (to *Session) GetFederationDeliveryServices(federationID int) ([]tc.FederationDeliveryServiceNullable, ReqInf, error) {
```
ocket8888: Client functions should have GoDocs IMO. Not as big a deal on other things, but this is directly exposed as a part of an interface to our system.

traffic_ops/testing/api/v14/federations_test.go
```diff
@@ -93,4 +93,16 @@ func PostTestFederationsDeliveryServices(t *testing.T) {
 	if _, err = TOSession.CreateFederationDeliveryServices(fedID, []int{ds.ID}, true); err != nil {
 		t.Errorf("creating federations delivery services: %v\n", err)
 	}
+
+	// Test get created Federation Delivery Service
+	fedDSes, _, err := TOSession.GetFederationDeliveryServices(fedID)
+	if err != nil {
+		t.Errorf("cannot GET Federation DeliveryServices: %v\n", err)
```
ocket8888: Should be a `t.Fatalf` - the caller can't safely assume that `fedDSes` in this case isn't `nil`.

I know that the way you wrote that this isn't exactly possible, but in general it's true. I guess I don't really have a problem with that, provided the GoDoc notes that.

traffic_ops/traffic_ops_golang/federations/ds.go
```diff
@@ -104,3 +105,37 @@ func getFedNameByID(tx *sql.Tx, id int) (string, bool, error) {
...
+	return map[string]dbhelpers.WhereColumnInfo{
+		"id":   dbhelpers.WhereColumnInfo{"fds.federation", api.IsInt},
+		"dsID": dbhelpers.WhereColumnInfo{"fds.deliveryservice", api.IsInt},
+	}
+}
+func (v *TOFedDSes) GetType() string {
+	return "federation_deliveryservice"
+}
+
+func (v *TOFedDSes) Read() ([]interface{}, error, error, int) { return api.GenericRead(v) }
```
ocket8888: I don't believe the Perl ever considered tenancy, so  won't demand that you do in the rewrite, but I wanted to point out that you could, if you so chose.

mhoppa: Do you think it is better as part of the port from perl to golang? Or as part of a different effort to add tenancy?

ocket8888: Either way, as you please.

mhoppa: Ill defer to another PR

PR #3949, opened 2019-10-01
Fixes broken regions and phys location UI tests

mitchell852:

other UI tests (regions, phys locations) depend on a division being available. deleting the newly created division causes the other tests to fail. This PR does not require additional tests or documentation.

- Traffic Portal

Run the [UI tests](https://github.com/apache/trafficcontrol/tree/master/traffic_portal/test/end_to_end)

### DIFF
--- a/traffic_portal/test/end_to_end/divisions/divisions-spec.js
+++ b/traffic_portal/test/end_to_end/divisions/divisions-spec.js
@@ -67,11 +67,4 @@ describe('Traffic Portal Divisions Test Suite', function() {
 		expect(pageData.name.getText() === myNewDiv.name + ' updated');
 	});
 
-	it('should delete the new division', function() {
-		console.log('Deleting the new division: ' + myNewDiv.name + ' updated');
-		pageData.deleteButton.click();
-		pageData.confirmWithNameInput.sendKeys(myNewDiv.name + ' updated');
-		pageData.deletePermanentlyButton.click();
-	});
-
 });


traffic_portal/test/end_to_end/divisions/divisions-spec.js
```diff
@@ -67,11 +67,4 @@ describe('Traffic Portal Divisions Test Suite', function() {
 		expect(pageData.name.getText() === myNewDiv.name + ' updated');
 	});
 
-	it('should delete the new division', function() {
-		console.log('Deleting the new division: ' + myNewDiv.name + ' updated');
-		pageData.deleteButton.click();
-		pageData.confirmWithNameInput.sendKeys(myNewDiv.name + ' updated');
```
ocket8888: Are you sure the problem isn't the `+ ' updated'` here?

mitchell852: no, the delete works fine. the problem is that for these tests to work,  you can't delete certain things. 

i.e. a division is needed to create a region which is needed to create a phys location. therefore, divisions-spec and regions-spec can't have a 'delete' test.

mitchell852: i will create another PR that does all the resource cleanup (deleting) at the end of all the tests.

ocket8888: ah, okay

PR #3952, opened 2019-10-02
TP: fixes table headers in cdn snapshot view

### LINKED ISSUE #3863: TP: table header in CDN Snapshot diff page is screwed up
## I'm submitting a ...

- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [X] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

Screen Shot 2019-08-12 at 2 33 26 PM
Appears correct only on General COnfig tab. All other tabs have same issue. 

## Expected / new behavior:

Screen Shot 2019-08-12 at 2 35 00 PM

login to TP
click on camera icon on the top right of TP page. 
enter the CDN and submit

## Anything else:

chrome browser v 75.0.3770.100

mitchell852:
##  What does this PR (Pull Request) do?

On the CDN snapshot view page, this change adjusts column headers when tab is selected and data table is visible to ensure the table header widths are correct.

- [x] This PR fixes #3863 

- Traffic Portal

1. Navigate to https://localhost:8442/#!/cdns/1/config/changes
2. click on the tabs and ensure the column headers of each table are the correct width

Verifying the widths of column headers depends on the user's window size so can't think of a way to test this programatically.
Documentation is already correct.
No need for a change log entry

- master (aea858f)
- 3.0.2
- 2.2.2

### DIFF
--- a/traffic_portal/app/src/modules/private/cdns/config/ConfigController.js
+++ b/traffic_portal/app/src/modules/private/cdns/config/ConfigController.js
@@ -174,6 +174,7 @@ let ConfigController = function (cdn, currentSnapshot, newSnapshot, $scope, $sta
 			"language": {
 				"emptyTable": "No pending changes"
 			},
+			"buttons": [],
 			"columnDefs": [
 				{ 'orderable': false, 'targets': [2, 3] }
 			]
--- a/traffic_portal/app/src/modules/private/cdns/config/ConfigController.js
+++ b/traffic_portal/app/src/modules/private/cdns/config/ConfigController.js
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-let ConfigController = function (cdn, currentSnapshot, newSnapshot, $scope, $state, $uibModal, locationUtils, cdnService, propertiesModel) {
+let ConfigController = function (cdn, currentSnapshot, newSnapshot, $scope, $state, $timeout, $uibModal, locationUtils, cdnService, propertiesModel) {
 
 	const oldConfig = currentSnapshot.config,
 		newConfig = newSnapshot.config;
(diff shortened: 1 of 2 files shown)


traffic_portal/app/src/modules/private/cdns/config/ConfigController.js
```diff
@@ -155,6 +155,14 @@ let ConfigController = function (cdn, currentSnapshot, newSnapshot, $scope, $sta
 		});
 	};
 
+	$scope.tabSelected = function() {
+		// issue 3863 - adjust column headers when tab is selected and data table is visible. hacky...sorry...
+		$timeout(function () {
```
ocket8888: using `$timeout` triggers an `$apply` and a `$digest` cycle. Since you're not changing angular view data and are directly manipulating the DOM with JQuery libraries, there's no need to do that. So it'd be more performant to use [`window.setTimeout`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout) instead - should have exactly the same syntax.

ocket8888:
I don't generally like using scripts to fix styling; styling should fix styling. What's wrong with:

```css
div.dataTables_scrollHeadInner {
    width: 100% !important;
}
```

?

mitchell852:
> I don't generally like using scripts to fix styling; styling should fix styling. What's wrong with:
> 
> ```css
> div.dataTables_scrollHeadInner {
>     width: 100% !important;
> }
> ```
> 
> ?

I can give it a shot. I was under the impression this bug was due to the datatables not being visible because they are hidden inside of tabs.

mitchell852:
i definitely like that solution better but the column headers are not quite right

image

ocket8888:
hm... I guess it's setting the widths explicitly, but just doing it wrong and never updating it. jQuery libraries..... Anyway, the little hiccup the page does when using a script to fix it is annoying, but if the CSS can't work properly I guess we've got no choice but to live with it until we eventually switch to Angular 7.

PR #3954, opened 2019-10-03
Added deprecation notices for /user/current/update

ocket8888:

Adds deprecation notices for the `/user/current/update` to all responses from that endpoint as well as to the documentation.

- Documentation
- Traffic Ops

Make requests to the `/user/current/update` endpoint and notice the new warning-level alert indicating its deprecation.
Also, build and read the documentation.

### DIFF
--- a/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
+++ b/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
@@ -105,5 +105,4 @@ ADD infrastructure/cdn-in-a-box/enroller/server_template.json \
 
 ADD infrastructure/cdn-in-a-box/traffic_ops_data /traffic_ops_data
 
-# RUN chown -R trafops:trafops /opt/traffic_ops
 CMD /run.sh
--- a/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
+++ b/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
@@ -105,5 +105,5 @@ ADD infrastructure/cdn-in-a-box/enroller/server_template.json \
 
 ADD infrastructure/cdn-in-a-box/traffic_ops_data /traffic_ops_data
 
-RUN chown -R trafops:trafops /opt/traffic_ops
+# RUN chown -R trafops:trafops /opt/traffic_ops
 CMD /run.sh
--- a/traffic_ops/app/lib/API/User.pm
+++ b/traffic_ops/app/lib/API/User.pm
@@ -599,6 +599,123 @@ sub current {
 	}
 }
 
+# Designated handler for the deprecated path to updating current users
+sub user_current_update {
+	my $self = shift;
+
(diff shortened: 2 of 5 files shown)


traffic_ops/app/lib/API/User.pm
```diff
@@ -599,6 +599,123 @@ sub current {
 	}
 }
 
+# Designated handler for the deprecated path to updating current users
+sub user_current_update {
```
mhoppa: Why did a new handler needed to be added for deprecation notice? Shouldn't we just be adding a deprecation message to the response on the current handler?

traffic_ops/app/lib/MojoPlugins/Response.pm
```diff
@@ -119,6 +119,27 @@ sub register {
 		}
 	);
 
+	$app->renderer->add_helper(
+		with_deprecation => sub {
```
mhoppa: Good work on this!

traffic_ops/app/lib/API/User.pm
ocket8888: The current handler is also the handler for `PUT /user/current`, which isn't deprecated. Since the rendering happens before the handler returns, I couldn't just wrap it, and if there's a way to access what path was requested from within a Mojolicious handler then I don't know it. So this seemed to be my only option.

mhoppa: ugh! i hope that we dont have to do that with a lot of the other endpoints we want to depreciate. Cool let me pull and test this then!

ocket8888: generally I don't think so.

ocket8888:
Well, technically, it is

I'll try to figure out why it's doing that.

ocket8888:
This took **hours** to figure out. Turns out, <kbd>[</kbd> is **VERY** different than <kbd>(</kbd> in Perl.

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
```diff
@@ -105,5 +105,5 @@ ADD infrastructure/cdn-in-a-box/enroller/server_template.json \
 
 ADD infrastructure/cdn-in-a-box/traffic_ops_data /traffic_ops_data
 
-RUN chown -R trafops:trafops /opt/traffic_ops
+# RUN chown -R trafops:trafops /opt/traffic_ops
```
mhoppa: ?

ocket8888: I'm honestly tempted to just remove that line it doesn't serve any apparent purpose, and takes _forever_ because it's recursing over probably tens of thousands of files. I had to disable it to get the builds to run fast enough to debug that earlier problem in one night..

I guess I shouldn't in this PR though.

mhoppa: I would be fine leaving it in this PR if it doesn’t cause issues which sounds like it doesn’t

PR #3959, opened 2019-10-04
Add qstring to path in TO access logs and log actual protocol used

rawlinp:
Capture the entire request qstring in the log which will include things like resource IDs that are important to log. Without this, those IDs would only be logged if they were path params not qstring params. Also, don't hardcode HTTP/1.1 as the protocol used when other HTTP versions can actually be used.

- Traffic Ops

Make some TO API requests with nonempty qstrings to TO while tailing the TO logs and verify that the qstrings are now showing up in the logs (e.g. `/api/1.4/ping?foo=bar`).

- [x] Enhances log output, no new tests necessary
- [x] Enhances log output, no docs necessary
- [x] Enhances log output, no changelog necessary

### DIFF
--- a/traffic_ops/traffic_ops_golang/routing/routing.go
+++ b/traffic_ops/traffic_ops_golang/routing/routing.go
@@ -193,10 +193,10 @@ func Handler(
 	reqID := getReqID()
 
 	reqIDStr := strconv.FormatUint(reqID, 10)
-	log.Infoln(r.Method + " " + r.URL.Path + " handling (reqid " + reqIDStr + ")")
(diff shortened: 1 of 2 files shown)


mhoppa:
I’m reviewing this / testing it

rawlinp:
I went back and forth on that myself and decided I like not including the `if` statement to handle that case better. It didn't seem like we really needed to differentiate between actual requests passing just `?` vs actual requests including no qstring whatsoever.

mhoppa:
:shipit:

PR #3961, opened 2019-10-07
added ability to select TLS version in TR

### LINKED ISSUE #3764: Specify TLS version support in Traffic Router server.xml
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [x] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When trying to specify one or multiple TLS versions for Traffic Router to allow it is ignored. Currently all TLS versions are used regardless of the configuration in server.xml.

## Expected / new behavior:

The expected behavior would be to allow one to specify which TLS versions should be allowed as referenced in the Tomcat documentation here: https://tomcat.apache.org/tomcat-8.5-doc/ssl-howto.html, https://tomcat.apache.org/tomcat-8.5-doc/config/http.html#SSL_Support and https://support.comodo.com/index.php?/Knowledgebase/Article/View/659/17/how-to----disable-weak-ciphers-in-tomcat-7--8 - Thanks @elsloo 

None of which work. I believe the problem lies here: https://github.com/apache/trafficcontrol/blob/2e6a6797d55ca28310d0c5eb1be5ec592dcc595f/traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java#L45 and here: https://github.com/apache/trafficcontrol/blob/2e6a6797d55ca28310d0c5eb1be5ec592dcc595f/traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java#L68 To me it looks like the SSL that Tomcat establishes is destroyed and then recreated with the "all" TLS protocols.

It is possible to test which TLS protocols are availbe by using the command:

```
docker run -it drwetter/testssl.sh:latest -p <delivery_service_url>
```

mattjackson220:

- [x] This PR fixes #3764 

This PR allows the selection of TLS version in Traffic Router.  It allows the protocols field in the Connector setup to be read in and used instead of the default "all" value.  No tests or documentation have been updated as this is a minor bug fix and adds functionality back in that was previously supported

- Traffic Router

In server.xml, add a protocols field to the 443 Connectors like this:
protocols="TLS1.2"

In CDN in a Box docker, run and wait until TR has been successfully started.  Log into any of the containers (i used TO) and verify that only the specified version of TLS is available.  Here is how i did the verification, though there are other ways:
Download this zip locally: https://github.com/drwetter/testssl.sh/archive/3.0.zip
Copy it into the TO container: docker cp ~/Downloads/testssl.sh-3.0.zip cdn-in-a-box_trafficops_1:/testsshl.zip
Log into the TO container: docker exec -it cdn-in-a-box_trafficops_1 /bin/bash
Install unzip: yum install unzip
Unzip the file: unzip /testsshl.zip 
Run the script on the test delivery service: testssl.sh-3.0/testssl.sh -p https://video.demo1.mycdn.ciab.test
Verify the results look like this:

 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      not offered
 TLS 1.1    not offered
 TLS 1.2    offered (OK)
 NPN/SPDY   not offered
 ALPN/HTTP2 not offered

Run the script against the API endpoint: testssl.sh-3.0/testssl.sh -p https://trafficrouter.infra.ciab.test:3443/crs/stats

Verify that this endpoint still allows all TLS versions with an output like this:

 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      offered (deprecated)
 TLS 1.2    offered (OK)
 NPN/SPDY   not offered
 ALPN/HTTP2 not offered

### DIFF
--- a/traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
+++ b/traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
@@ -32,6 +32,7 @@ import java.util.Set;
 
 public class RouterNioEndpoint extends NioEndpoint {
 	private static final Logger LOGGER = Logger.getLogger(RouterNioEndpoint.class);
+	private String protocols;
 
 	// Grabs the aliases from our custom certificate registry, creates a sslHostConfig for them
 	// and adds the newly created config to the list of sslHostConfigs.  We also remove the default config
@@ -65,7 +66,7 @@ public class RouterNioEndpoint extends NioEndpoint {
 			cert.setCertificateKeyAlias(alias);
 			sslHostConfig.addCertificate(cert);
 			sslHostConfig.setCertificateKeyAlias(alias);
-			sslHostConfig.setProtocols("all");
+			sslHostConfig.setProtocols(protocols != null ? protocols : "all");
 			sslHostConfig.setConfigType(getSslConfigType());
 			sslHostConfig.setCertificateVerification("none");
 			LOGGER.info("sslHostConfig: "+sslHostConfig.getHostName() + " " + sslHostConfig.getTruststoreAlgorithm());
@@ -118,6 +119,14 @@ public class RouterNioEndpoint extends NioEndpoint {
 				SSL.getLastErrorNumber();
 			}
 		}
+	}
+
+	public String getProtocols() {
+		return protocols;
+	}
+
+	public void setProtocols(final String protocols) {
+		this.protocols = protocols;
 	}
 
 }


ocket8888:
ok to test

traffic_router/connector/src/main/java/com/comcast/cdn/traffic_control/traffic_router/protocol/RouterNioEndpoint.java
```diff
@@ -120,5 +121,13 @@ protected void doRun(){
...
 	}
 
+	public String getProtocols() {
+		return protocols;
+	}
+
+	public void setProtocols(final String protocols) {
+		this.protocols = protocols;
+	}
+
```
ajschmidt: I think this should be using the 'sslProtocol' attribute that is already part of the specification. Was there a problem with that?  I don't think its good to invent custom attributes.

mattjackson220: i used 'protocols' instead of 'sslProtocol' because thats how it was setup in TR version 2.2 and below.  its a valid Tomcat Connector attribute, but their docs admit it has some overlap with the 'sslProtocol'.

ajschmidt: I see you are correct that 'protocols' is a legitimate attribute. I think it would be nice to honor all of the attributes that are set in the `Connector` element while you are at it instead of selecting specific ones. That is why I've suggested this alternate implementation. We could circle back to this in the future.

hbeatty:
I should have also mentioned that I backported Matt's changes to 3.1 and that is what we will be running.

rawlinp:
@hbeatty there is currently an open vote on the 3.1 release candidate. If you've tested this out and are satisfied, feel free to merge it and -1 the current release candidate for us to cut another one with this in it.

PR #3962, opened 2019-10-07
Add atscfg remap.config tests

rob05c:

Adds atscfg remap.config tests. With this, `go test -cover` reports 100% for `lib/go-atscfg/remadotconfig.go`.

Also removes some unused logic in remapdotconfig.go. Easily verifiable the function is unused, and the nil check is duplicated by the previous line.

Is tests.

- Traffic Ops ORT

Manually verify unused code in remapdotconfig.go is unused. Run unit tests on `lib/go-atscfg`.

Not a bug fix.

### DIFF
--- a/lib/go-atscfg/atscfg_test.go
+++ b/lib/go-atscfg/atscfg_test.go
@@ -22,6 +22,8 @@ package atscfg
 import (
 	"strings"
 	"testing"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
 )
 
 func TestGenericHeaderComment(t *testing.T) {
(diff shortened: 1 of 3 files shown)


rob05c:
>I do think use of table driven tests 

Huh, TIL. I do that a lot, didn't know it had a name.

But, if we wanted to reduce the "duplicate" lines, I think some kind of `MakeDefaultServerInfo` would go farther, and then each test would modify what it cared about.

I was on the fence about that. It'd reduce the lines of code, but I also kind of feel like it's good for each test to be completely independent, and saying "I'm testing exactly this data," and there's no shared func that would change all the tests if it were changed. 

I went with the latter, but I'm personally not sure which is a "better" test 🤷‍♂ .

PR #3963, opened 2019-10-07
TP: by default only show parameter diff when comparing 2 profiles

mitchell852:

This PR changes the default behavior when comparing the parameters of 2 profiles:

current behavior (#3919 ): ALL parameters (shared and not shared) are displayed.

new behavior: As requested by users, ONLY the parameters that belong to one profile and NOT the other (the diff) are displayed by DEFAULT and the user is given the option to "Show Shared Params" if so desired.

- Traffic Portal

- Launch TP

1. 
- Navigate to https://tp.domain.com/#!/profiles
- Select "Compare Profiles" under the More button
- Select 2 profiles to compare
- Ensure none of the "shared" parameters are displayed
- add or remove parameters from each profile; click "Update"
- ensure the comparison is updated correctly after confirmation.

2.
- Navigate to https://tp.domain.com/#!/profiles
- Select "Compare Profiles" under the More button
- Select 2 profiles to compare
- Select 'Show Shared Params'
- Ensure each profile has the proper parameters "checked". (you can double check by checking the parameters for the profile at https://tp.domain.com/#!/profiles/{profile_id}/parameters)
- add or remove parameters from each profile; click "Update"
- ensure the proper parameters are assigned to each profile after confirmation.

- [x] The tests associated with #3919 cover this new feature however a minor change was required
- [x] The CHANGELOG.md entry for #3919 is still applicable


ocket8888:
retest this please

ocket8888:
retest this please

ocket8888:
retest this please

ocket8888:
retest this please

traffic_portal/app/src/modules/private/profiles/compare/diff/index.js
```diff
@@ -0,0 +1,49 @@
...
+						resolve: {
+							profile1: function($stateParams, profileService) {
+								return profileService.getProfile($stateParams.profile1Id, { includeParams: true });
+							},
+							profile2: function($stateParams, profileService) {
+								return profileService.getProfile($stateParams.profile2Id, { includeParams: true });
+							},
+							profilesParams: function(profile1, profile2) {
+								// funky way to do a unique (by id) union of param objects
+								return _.chain(profile1.params).union(profile2.params).unique('id').value();
```
ocket8888: I don't know what this could possibly be doing. An underscore-wrapped array of hashes has no property `unique`, and I can't find anything like what you're doing in [the underscorejs documentation](https://underscorejs.org/) so I don't think it's wise to rely on this. Here's what I would do instead:
```javascript
const ids = new Set();
const uniq = new Array();
for (const param of profile1.params.concat(profile2.params)) {
    if (!ids.has(param.id)) {
        uniq.push(param);
    }
}
return uniq;
```

traffic_portal/app/src/common/service/utils/LocationUtils.js
```diff
@@ -17,13 +17,35 @@
...
+                resolve: {
+                    params: function () {
+                        return params;
+                    }
+                }
+            });
+            modalInstance.result.then(function() {
+                $location.url(path);
+            }, function () {
+                // do nothing
```
ocket8888: pls no do nothing
you could just log the error.

traffic_portal/app/src/modules/private/profiles/compare/diff/index.js
ocket8888: Alternatively:
```
const ids = new Set();
return profile1.params.concat(profile2.params).filter((x)=>{if (!ids.has(x.id)){ids.add(x.id);return true;} return false;});
```
but I think that's a little too messy for a one-liner.

PR #3964, opened 2019-10-07
Add GET/POST/DELETE for server_server_capabilities

mhoppa:
## What does this PR (Pull Request) do?

- [x] This PR is not related to any Issue 
Add server server_capabilities support to the API.

Adds the ability to

* GET
* POST
* DELETE

server_capabilities associations with servers.

Includes tests, API handlers, migrations

## Which Traffic Control components are affected by this PR?

- Documentation
- Traffic Control Golang Client 
- Traffic Ops

## What is the best way to verify this PR?

Apply migrations to db and seed server_capabilities in new table server_capability.

Build new traffic ops API and then associate/get/disassociates those newly seeded server_capability to servers via the new api path /server_server_capabilities

## The following criteria are ALL met by this PR

- [x] This PR includes tests
- [x] This PR includes documentation 
- [x] This PR does not include an update to CHANGELOG.md 
- [x] This PR includes any and all required license headers
- [x] This PR ensures that database migration sequence is correct
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #3966, opened 2019-10-08
Add server capabilities API

rawlinp:

Add a new API for creating, reading, and deleting (not updating) server capabilities:
```
POST /api/1.4/server_capabilities
GET /api/1.4/server_capabilities[?name=<name>]
DELETE /api/1.4/server_capabilities?name=<name>

expected POST body:
{"name": "bar"}

expected POST response:
{
  "alerts": [
    {
      "text": "server capability was created.",
      "level": "success"
    }
  ],
  "response": {
    "name": "bar",
    "lastUpdated": "2019-10-07 22:10:00+00"
  }
}

expected GET response (can use ?name=... qparam to filter by name):
{
  "response": [
    {
      "name": "bar",
      "lastUpdated": "2019-10-07 20:38:24+00"
    }
  ]
}

expected DELETE response:
{
  "alerts": [
    {
      "text": "server capability was deleted.",
      "level": "success"
    }
  ]
}
```

- Traffic Control Client (Go)
- Traffic Monitor
- Traffic Ops

Documentation will be done in a follow-up PR.

Read the added TO API tests to verify they're sufficient, then run the TO API tests.

### DIFF
--- a/traffic_ops/client/servercapability.go
+++ b/traffic_ops/client/servercapability.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
 )
@@ -67,8 +68,8 @@ func (to *Session) GetServerCapabilities() ([]tc.ServerCapability, ReqInf, error
 
 // GetServerCapability returns the given server capability by name.
 func (to *Session) GetServerCapability(name string) (*tc.ServerCapability, ReqInf, error) {
-	url := fmt.Sprintf("%s?name=%s", APIServerCapabilities, name)
-	resp, remoteAddr, err := to.request(http.MethodGet, url, nil)
+	reqUrl := fmt.Sprintf("%s?name=%s", APIServerCapabilities, url.QueryEscape(name))
+	resp, remoteAddr, err := to.request(http.MethodGet, reqUrl, nil)
 	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
 	if err != nil {
 		return nil, reqInf, err
@@ -88,8 +89,8 @@ func (to *Session) GetServerCapability(name string) (*tc.ServerCapability, ReqIn
 
 // DeleteServerCapability deletes the given server capability by name.
 func (to *Session) DeleteServerCapability(name string) (tc.Alerts, ReqInf, error) {
-	route := fmt.Sprintf("%s?name=%s", APIServerCapabilities, name)
-	resp, remoteAddr, err := to.request(http.MethodDelete, route, nil)
+	reqUrl := fmt.Sprintf("%s?name=%s", APIServerCapabilities, url.QueryEscape(name))
+	resp, remoteAddr, err := to.request(http.MethodDelete, reqUrl, nil)
 	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
 	if err != nil {
 		return tc.Alerts{}, reqInf, err
--- a/traffic_ops/client/servercapability.go
+++ b/traffic_ops/client/servercapability.go
@@ -66,7 +66,7 @@ func (to *Session) GetServerCapabilities() ([]tc.ServerCapability, ReqInf, error
 }
 
 // GetServerCapability returns the given server capability by name
-func (to *Session) GetServerCapability(name string) ([]tc.ServerCapability, ReqInf, error) {
+func (to *Session) GetServerCapability(name string) (*tc.ServerCapability, ReqInf, error) {
 	url := fmt.Sprintf("%s?name=%s", APIServerCapabilities, name)
 	resp, remoteAddr, err := to.request(http.MethodGet, url, nil)
 	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
@@ -80,7 +80,10 @@ func (to *Session) GetServerCapability(name string) ([]tc.ServerCapability, ReqI
 		return nil, reqInf, err
 	}
 
-	return data.Response, reqInf, nil
+	if len(data.Response) == 1 {
+		return &data.Response[0], reqInf, nil
+	}
+	return nil, reqInf, fmt.Errorf("expected one server capability in response, instead got: %+v", data.Response)
 }
 
 // DeleteServerCapability deletes the given server capability by name
--- a/traffic_ops/client/servercapability.go
+++ b/traffic_ops/client/servercapability.go
@@ -28,7 +28,7 @@ const (
 	APIServerCapabilities = apiBase + "/server_capabilities"
 )
 
-// CreateServerCapability creates a server capability and returns the response
+// CreateServerCapability creates a server capability and returns the response.
 func (to *Session) CreateServerCapability(sc tc.ServerCapability) (*tc.ServerCapabilityDetailResponse, ReqInf, error) {
 	var remoteAddr net.Addr
 	reqBody, err := json.Marshal(sc)
@@ -48,7 +48,7 @@ func (to *Session) CreateServerCapability(sc tc.ServerCapability) (*tc.ServerCap
 	return &scResp, reqInf, nil
 }
 
-// GetServerCapabilities returns all the server capabilities
+// GetServerCapabilities returns all the server capabilities.
 func (to *Session) GetServerCapabilities() ([]tc.ServerCapability, ReqInf, error) {
 	resp, remoteAddr, err := to.request(http.MethodGet, APIServerCapabilities, nil)
 	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
@@ -65,7 +65,7 @@ func (to *Session) GetServerCapabilities() ([]tc.ServerCapability, ReqInf, error
 	return data.Response, reqInf, nil
 }
 
-// GetServerCapability returns the given server capability by name
+// GetServerCapability returns the given server capability by name.
 func (to *Session) GetServerCapability(name string) (*tc.ServerCapability, ReqInf, error) {
 	url := fmt.Sprintf("%s?name=%s", APIServerCapabilities, name)
 	resp, remoteAddr, err := to.request(http.MethodGet, url, nil)
@@ -86,7 +86,7 @@ func (to *Session) GetServerCapability(name string) (*tc.ServerCapability, ReqIn
 	return nil, reqInf, fmt.Errorf("expected one server capability in response, instead got: %+v", data.Response)
 }
 
-// DeleteServerCapability deletes the given server capability by name
+// DeleteServerCapability deletes the given server capability by name.
 func (to *Session) DeleteServerCapability(name string) (tc.Alerts, ReqInf, error) {
 	route := fmt.Sprintf("%s?name=%s", APIServerCapabilities, name)
 	resp, remoteAddr, err := to.request(http.MethodDelete, route, nil)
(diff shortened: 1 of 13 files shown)


traffic_ops/traffic_ops_golang/servercapability/servercapability.go
```diff
@@ -0,0 +1,100 @@
...
+:name
+) RETURNING last_updated`
+	return query
+}
+
+func deleteQuery() string {
+	query := `DELETE FROM server_capability
+WHERE name=:name`
+	return query
+}
```
awilliams: I see that this follows an idiom used [elsewhere in the project](https://github.com/apache/trafficcontrol/blob/6ea4a3f53c104b2c3dfa9dc404d3c0f70d00c56a/traffic_ops/traffic_ops_golang/division/divisions.go#L108-L125), ie defining functions which are then used within methods.

For consistency would it be preferable to either define a `query` variable that is then returned, or directly return a string literal in all cases?

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
```diff
@@ -0,0 +1,100 @@
...
+		"name": {"sc.name", nil},
+	}
+}
+func (v *TOServerCapability) DeleteQuery() string { return deleteQuery() }
+
+func (v TOServerCapability) GetKeyFieldsInfo() []api.KeyFieldInfo {
+	return []api.KeyFieldInfo{{"name", api.GetStringKey}}
+}
+
+//Implementation of the Identifier, Validator interface functions
```
awilliams: Consider adding a space after the last `/`, since no space is typically used for machine-parsable comments.

traffic_ops/client/servercapability.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var data tc.ServerCapabilitiesResponse
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		return nil, reqInf, err
+	}
+
+	return data.Response, reqInf, nil
+}
+
+// GetServerCapability returns the given server capability by name
```
awilliams: Consider using `capabilities` to match the method signature. Also, is it ever expected to receive multiple capabilities? May be worth answering that question in the comment as well.

ocket8888:
Was this new endpoint brought to the mailing list's attention? Not that I think anyone will have a problem with it, just seems like something important enough to bother with.

traffic_ops/client/servercapability.go
rawlinp: Our API generally returns lists of resources even when it's only ever expected to return a single resource (as in this case). I suppose I could just have the client method return the single expected resource here, although most of our other client methods just return lists.

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
rawlinp: Done

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
rawlinp: I suppose the local `query` variable isn't really necessary, but I'm just following idioms used elsewhere like you said.

rawlinp:
I think Rob is working on a blueprint.

mitchell852:
Since there are no docs, can you add to your PR description what that routes are and the expected request payloads for each?

rawlinp:
Updated the description with routes and expected payloads.

mitchell852:
> Updated the description with routes and expected payloads.

GET /server_capabilities[?name=]
POST /server_capabilities
DELETE /server_capabilities?name=

right?

rawlinp:
right, updated to include that

rawlinp:
Opened placeholder draft PR for adding documentation: https://github.com/apache/trafficcontrol/pull/3971

mitchell852:
without reading thru the code, does DELETE validate that the "server capability" is:

a. not being used by any server and
b. is not required by any ds

with some sort of alert error like "server capability failed. server capability used by X servers/delivery services"

sorry, i should just read the code :)

rob05c:
>without reading thru the code, does DELETE validate that the "server capability" is:

>a. not being used by any server and
>b. is not required by any ds

Server-Capabilities and DS-Required-Capabilities don't exist in this PR, that'll have to be added in a subsequent PR.

But, I'd vote we don't add custom Go logic to handle that - rather, if we make them proper Foreign Keys that don't CASCADE DELETE, the DB will automatically prevent it. Then, all we have to do is parse the Postgres error to return a useful message to the user (we already have a func for that `api.ParseDBError`

mhoppa:
> > without reading thru the code, does DELETE validate that the "server capability" is:
> 
> > a. not being used by any server and
> > b. is not required by any ds
> 
> Server-Capabilities and DS-Required-Capabilities don't exist in this PR, that'll have to be added in a subsequent PR.
> 
> But, I'd vote we don't add custom Go logic to handle that - rather, if we make them proper Foreign Keys that don't CASCADE DELETE, the DB will automatically prevent it. Then, all we have to do is parse the Postgres error to return a useful message to the user (we already have a func for that `api.ParseDBError`

FWIW Did add that constraint to my PR per @rawlinp feedback https://github.com/apache/trafficcontrol/pull/3964/files#diff-7d3547a278b38f437ab01966666c94a8R26

mitchell852:
i don't suppose you happened to add entries to seeds.sql for creating a "capability" (i'm talking about a user capability here) like the following:

server-capabilities-read
server-capabilities-write

seems a bit silly since we don't use "user capabilities" yet but i feel like it will make our lives easier if every time we add new api endpoints we keep the "user capabilities" current.

rawlinp:
@mitchell852 I didn't add seeds.sql entries yet. @mhoppa probably needs those for his API endpoints as well. Should POST/DELETE only be admin? Right now I have it set to Operations.

mitchell852:
> Should POST/DELETE only be admin? Right now I have it set to Operations.

probably a question for @rob05c

rob05c:
> Should POST/DELETE only be admin? Right now I have it set to Operations.

I think Ops is right. Admin vs Ops is somewhat nebulous, but I think in general, Operations can do pretty much everything except see password and security stuff. So I think Operations is right for these endpoints.

mitchell852:
Are there any restrictions on server capability name? I.e. no spaces, alphanumeric only, etc...

traffic_ops/client/servercapability.go
```diff
@@ -0,0 +1,103 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		return nil, reqInf, err
+	}
+
+	return data.Response, reqInf, nil
+}
+
+// GetServerCapability returns the given server capability by name.
+func (to *Session) GetServerCapability(name string) (*tc.ServerCapability, ReqInf, error) {
+	url := fmt.Sprintf("%s?name=%s", APIServerCapabilities, name)
```
mhoppa: Non blocking as we dont do it anywhere but for here and below, on query parameters, we should use the net/url package to build them out so they are correctly url encoded.

rob05c:
>Are there any restrictions on server capability name? I.e. no spaces, alphanumeric only, etc...
The design reqs don't, I was originally thinking we'd let people put anything, but now that I think more about it, I think we should limit it to URI path chars, I.e. digit/alpha/_/-. That lets us put it in a path or query easier later, and I can't imagine why someone would _need_ any other chars.

Thoughts? If no one objects, I'll put that in the blueprint (but not hold up this PR for it, we can easily add that in another)

mitchell852:
Just a thought. What about a GET response like:

```
{
  "response": [
    {
      "name": "bar",
      "lastUpdated": "2019-10-07 20:38:24+00",
      "servers: [],
      "deliveryServices: []
    }
  ]
}
```

We do something similar with GET /parameters where we have a []profiles property. It gives me the ability in the UI to show a "server count" and a "delivery service count" which seems like interesting info to the user.

rawlinp:
+1 on restricting the name to alphanumeric/-/_.

Could the "server count" and "ds count" information be answered by just including links to pages which are backed by the server_server_capabilities API? E.g. click into the RAM capability, then click the "show servers" link to see all the servers with this capability or "show DSes" to see all the DSes with this capability.

traffic_ops/client/servercapability.go
rawlinp: Good call. I'll add that into this one.

mitchell852:
> Could the "server count" and "ds count" information be answered by just including links to pages which are backed by the server_server_capabilities API?

yep. i think that's fine.

lbathina:
1. ```{
    "alerts": [
        {
            "text": "no server capability with that id found",
            "level": "error"
        }
    ]
}``` message here needs to be corrected. we are deleting server_capabilities only by name.
2. when spaces are at the start and end of name, they are trimmed and accepted instead of throwing error. 
3. api parameters are not validated. for instance, requesting service_capabilities endpoint with param id doesn't throw any error.

rob05c:
> 1. message here needs to be corrected. we are deleting server_capabilities only by name.

Yep, looks like an artifact of `api.GenericDelete`. Good catch!

Looks like either `api.GenericDelete` will have to be changed to something like "no x with that key found," or otherwise extended to know the names of the key(s).

> 2. when spaces are at the start and end of name, they are trimmed and accepted instead of throwing error.

That should be fine. It should be documented, but the extra safety seems like a good thing to me.

> 3. api parameters are not validated. for instance, requesting service_capabilities endpoint with param id doesn't throw any error.

The API shouldn't do this. Unknown parameters should be ignored. This follows the Robustness Principle. See 
https://tools.ietf.org/html/rfc1122#section-1.2.2
https://en.wikipedia.org/wiki/Robustness_principle

This makes our API easier to use, and less likely to break users, and more forward-compatible. For example, someone could be using a client with a newer version, or newer code within the same version, but knowingly double-checking their filtering params (atstccfg does this); or posting some kind of debug information that helps themselves; or simply making a simple mistake that doesn't need to break things.

I know some other APIs do this. It is a gross violation of the Robustness Principle, and I strongly disapprove. It makes servers brittle and fragile, and complicates integration and forward-compatibility.

mitchell852:
> Trimming whitespace should be fine. It should be documented, but the extra safety seems like a good thing to me.

agree

mitchell852:
> Unknown parameters should be ignored.

agree

rawlinp:
> 1. ```
>     "alerts": [
>         {
>             "text": "no server capability with that id found",
>             "level": "error"
>         }
>     ]
>    ```
> 
> }``` message here needs to be corrected. we are deleting server_capabilities only by name.

Yeah, I noticed that as well and just kind of figured "name" is really the "id" of the resource, so I just left it as it was. I'll change the `GenericDelete` function to use "key" instead of "id".

> 2. when spaces are at the start and end of name, they are trimmed and accepted instead of throwing error.

I'm not intentionally trimming the name, but about to submit a fix that validates the name is alphanumeric/dash/underscore.

> 3. api parameters are not validated. for instance, requesting service_capabilities endpoint with param id doesn't throw any error.

I also agree unknown API parameters should be ignored.

> 4. on PUT which is not defined - may be we should do method not allowed?

This seems like a problem with our entire API rather than just this PR. We should fix that holistically and separately IMO.

> 5. When a post with no body is made. the following error is provided.
> 
> ```
>     "alerts": [
>         {
>             "text": "EOF",
>             "level": "error"
>         }
>     ]
> }```
>  instead of a meaningful message indicating the missing info.
> ```

I suspect that is also a larger problem with the majority of our API, since we share the common code for unmarshalling JSON and returning errors across multiple endpoints. We should fix that separately.

mitchell852:
> 5. When a post with no body is made. the following error is provided.
> 
> ```
>     "alerts": [
>         {
>             "text": "EOF",
>             "level": "error"
>         }
>     ]
> }```
>  instead of a meaningful message indicating the missing info.
> ```

what to create a separate issue for that @lbathina

PR #3967, opened 2019-10-08
Rewrite DELETE federation delivery service to golang

### LINKED ISSUE #3794: Rewrite /federations/{{id}}/deliveryservices/{{DSID}} to Go
## I'm submitting a ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/federations/{{id}}/deliveryservices/{{DSID}}` is handled by Perl

## Expected / new behavior:
`/federations/{{id}}/deliveryservices/{{DSID}}` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `DELETE`

N/A

mhoppa:

- [x] This PR fixes #3794 

- Documentation
- Traffic Control Go Client 
- Traffic Ops

Pull down and build traffic ops. Either run API tests or call to the API and delete a delivery service from a federation.

- [x] This PR includes tests
- [x] This PR includes does not include an update to CHANGELOG.md


traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -445,7 +445,7 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `federations/?(\.json)?$`, federations.Get, auth.PrivLevelFederation, Authenticated, nil},
 		{1.1, http.MethodPost, `federations/{id}/deliveryservices?(\.json)?$`, federations.PostDSes, auth.PrivLevelAdmin, Authenticated, nil},
 		{1.1, http.MethodGet, `federations/{id}/deliveryservices?(\.json)?$`, api.ReadHandler(&federations.TOFedDSes{}), auth.PrivLevelReadOnly, Authenticated, nil},
-		{1.1, http.MethodDelete, `federations/{id}/deliveryservices?(\.json)?$`, api.DeleteHandler(&federations.TOFedDSes{}), auth.PrivLevelAdmin, Authenticated, nil},
+		{1.1, http.MethodDelete, `federations/{id}/deliveryservices/{dsID}/?(\.json)?$`, api.DeleteHandler(&federations.TOFedDSes{}), auth.PrivLevelAdmin, Authenticated, nil},
```
ocket8888: Why remove the existing handler?

mhoppa: ?

ocket8888: This removes the handler for `DELETE /api/1.x/federations/(ID}/deliveryservices`

mhoppa: yeah it does from my initial commit but not master. my initial commit I was replacing the perl path of _federations/{id}/deliveryservices/{dsID}_  with _/api/1.x/federations/(ID}/deliveryservices_ but that was going to break the API so i backed it out and went back to the correct path

ocket8888: oh... How did I even comment on that?

traffic_ops/traffic_ops_golang/federations/ds.go
```diff
@@ -123,11 +131,93 @@ func (v *TOFedDSes) ParamColumns() map[string]dbhelpers.WhereColumnInfo {
...
+		return api.ParseDBError(err)
+	}
+
+	return nil, nil, http.StatusOK
+}
+
+func checkFedDSDeletion(tx *sql.Tx, fedID, dsID int) (int, error, error) {
+
+	q := `SELECT ARRAY(SELECT deliveryservice FROM federation_deliveryservice WHERE federation=$1)`
+	dsIDs := []int64{} // pg.Array does not support int slice needs to be int64
```
ocket8888: nit: `pq.Array`, not `pg.Array`

traffic_ops/traffic_ops_golang/federations/ds.go
```diff
@@ -123,11 +131,93 @@ func (v *TOFedDSes) ParamColumns() map[string]dbhelpers.WhereColumnInfo {
...
+	// Check that we can delete it
+	if respCode, usrErr, sysErr := checkFedDSDeletion(v.APIInfo().Tx.Tx, *v.fedID, dsID); usrErr != nil || sysErr != nil {
+		if usrErr != nil {
+			return usrErr, sysErr, respCode
+		}
+		return usrErr, sysErr, respCode
+	}
+
+	// Actually delete the DS from the Federation
+	if err := deleteFedDS(v.APIInfo().Tx.Tx, *v.fedID, dsID); err != nil {
```
ocket8888: Should this check tenancy? I'm sure Perl didn't, so it's probably not strictly necessary for a rewrite, but something to consider.

mhoppa: given that I did not add it on the GET rewrite (and the POST that was rewritten and merged before that)  I say that we do not add it here as we should add it to all three at the same time instead of partial support.

Going forward we should come to an agreeance if rewrites should include tenancy checks or not.

PR #3982, opened 2019-10-11
Add Traffic Monitor gbps calculated stat.

### LINKED ISSUE #3981: Add Traffic Monitor gbps calculated stat, for thresholds.
Right now, for bandwidth thresholds, the Monitor has `availableBandwidthInKbps` and `kbps`. For multi-gigabit servers, the Parameters end up being like `60000000`. Hard to keep track of all those zeros. 

It'd be super-easy to add another calculated stat to the Monitor, `gbps`, so you could use a parameter `health.threshold.gbps 4`, which is much easier for Ops to visually verify is correct.

## I'm submitting a ...
- [ ] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [x] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

Bandwidth thresholds are `kbps` , which has tons of zeros, hard for Ops to manage and verify correct.

## Expected / new behavior:

`gbps` Calculated Stat exists in the Monitor, so operators can use `health.threshold.gbps` as a Threshold.

Setting `health.threshold.gbps` Parameter in TO works as a TM threshold.

## Anything else:
Not really.

vincentalfred:

- [x] This PR fixes #3981 

- Traffic Monitor


traffic_monitor/cache/cache.go
```diff
@@ -154,6 +154,9 @@ func ComputedStats() map[string]StatComputeFunc {
 		"kbps": func(info ResultInfo, serverInfo tc.TrafficServer, serverProfile tc.TMProfile, combinedState tc.IsAvailable) interface{} {
 			return info.Vitals.KbpsOut
 		},
+		"gbps": func(info ResultInfo, serverInfo tc.TrafficServer, serverProfile tc.TMProfile, combinedState tc.IsAvailable) interface{} {
+			return info.Vitals.KbpsOut / 1000000
```
mhoppa: are we worried about losing precision here? ex ~ 1500000 kbps would be represented as 1 gbps

rob05c: I agree, gbps is too big, I think we need a float64 here. I think the Monitor will properly cast any numeric primitive to a float, but it'll need tested to be sure.

rob05c:
Thanks! This is great, this small change is a big help to CDN operators!

Would you mind adding a `CHANGELOG.md` entry? Since this is a new feature, it should have an entry for the next Release (`## [Unreleased]` header in the file). The `CHANGELOG.md` file is at the root of the project.

I think this is ok without docs, we don't currently document the calculated stats in the Monitor.

But, we've been trying to add tests for every new feature. Would you mind adding a small unit test for this?
The ComputedStat funcs take a lot of data, but your func doesn't need them most of the variables, so for the test, you can pass empty objects for everything but `info.Vitals.KbpsOut`, and then just verify that various Kbps inputs turn into the expected Gbps.

Have you tested already, or run ATC before? If not, there are several ways you can do it. The easiest is probably to run a CDN-in-a-Box, instructions are here: https://traffic-control-cdn.readthedocs.io/en/latest/admin/quick_howto/ciab.html. A CDN is a complex thing though, if you hit issues, we're more than happy to help. We're on Slack, and also have an email list: https://traffic-control-cdn.readthedocs.io/en/latest/faq.html#how-do-i-get-help-with-traffic-control

traffic_monitor/cache/cache.go
vincentalfred: I'm gonna cast it as float64 just to be safe.

vincentalfred:
I'm relatively new to Go and open source. I'll update the changelog and try to create a unit test for it. Thanks for the guidance!

ocket8888:
With this change, what happens if there are conflicting maxKBPS and maxGBPS Parameters?

rob05c:
You mean as threshold? You can apply any number of thresholds to the Monitor. Whatever is lowest will make it Unavailable first.

ocket8888:
:ok:

vincentalfred:
I write something like this for the test, but I don't know why it doesn't work.
```
func TestComputeStatGbps(t *testing.T) {
	computedStats := ComputedStats()
	got := computedStats["gbps"](info: ResultInfo{Vitals: Vitals{KbpsOut: 1500000}})
	want := 1.5
	if got != want {
		t.Errorf("ComputedStats[\"gbps\"] return %v instead of %v", got, want)
	}

	got = computedStats["gbps"](info: ResultInfo{Vitals: Vitals{KbpsOut: 1400000}})
	want = 1.4
	if got != want {
		t.Errorf("ComputedStats[\"gbps\"] return %v instead of %v", got, want)
	}
}
```
This is the error message I got:
```
# command-line-arguments
cache_test.go:55:35: missing ',' in argument list
cache_test.go:61:34: missing ',' in argument list
FAIL	command-line-arguments [setup failed]
FAIL
```
Sorry to trouble you. Thank you for being so patient.

mhoppa:
@vincentalfred you need to pass in all the arguments for the computedStats func so it would look like 
```
func TestComputeStatGbps(t *testing.T) {
	serverInfo := tc.TrafficServer{}
	serverProfile := tc.TMProfile{}
	combinedState := tc.IsAvailable{}
	computedStats := ComputedStats()
	got := computedStats["gbps"](ResultInfo{Vitals: Vitals{KbpsOut: 1500000}}, serverInfo, serverProfile, combinedState)
	want := 1.5
	if got != want {
		t.Errorf("ComputedStats[\"gbps\"] return %v instead of %v", got, want)
	}

	got = computedStats["gbps"](ResultInfo{Vitals: Vitals{KbpsOut: 1400000}}, serverInfo, serverProfile, combinedState)
	want = 1.4
	if got != want {
		t.Errorf("ComputedStats[\"gbps\"] return %v instead of %v", got, want)
	}
}
```

And then it should work

mhoppa:
looks great! @vincentalfred.  @ocket8888 @rawlinp can you guys look to approve/merge??

PR #3985, opened 2019-10-11
Remove commented out PATCH /jobs code, api_capability, and docs

rawlinp:

The PATCH method was not included and should not be referenced by
documentation or exist as an api_capability. In addition, commented-out
code in the codebase is unnecessary cruft that makes the codebase harder
to search and understand. If a PATCH method needs to be supported in the
future, the commented-out code can be retrieved from the git history
instead.

- Documentation
- Traffic Ops

1. Verify traffic_ops_golang still builds and the unit tests pass
2. Make sure the documentation still builds
3. Run `db/admin seed` to re-run the `seeds.sql` file on TODB, make sure it succeeds


mitchell852:
> commented-out
> code in the codebase is unnecessary cruft that makes the codebase harder
> to search and understand.

agree. git history can find it if needed.

PR #3988, opened 2019-10-11
WIP Remove Traffic Ops Golang legacy tc.ApiErrorType

### LINKED ISSUE #3349: Remove Traffic Ops Golang legacy tc.ApiErrorType
The `tc.ApiErrorType` is a legacy artifact that needs removed. TO Golang has standardized around the `userErr, sysErr, errCode` idiom, which is more powerful and flexible, and easier to work with (no type casting). The few remaining functions in TOGo using `tc.ApiErrorType` need changed to use the `userErr, sysErr, errCode` idiom, and the symbol removed.

zrhoffman:

- [x] This PR fixes #3349 

- Traffic Ops

ApiErrorType is not mentioned in the documentation, so its removal does not need to affect documentation.

Not a bug fix.

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
@@ -380,14 +380,14 @@ func (ds *TODeliveryService) Read() ([]interface{}, error, error, int) {
 	}
 
 	returnable := []interface{}{}
-	dses, errs, _ := readGetDeliveryServices(ds.APIInfo().Params, ds.APIInfo().Tx, ds.APIInfo().User)
-	if len(errs) > 0 {
-		for _, err := range errs {
-			if err.Error() == `id cannot parse to integer` { // TODO create const for string
-				return nil, errors.New("Resource not found."), nil, http.StatusNotFound //matches perl response
-			}
-		}
-		return nil, nil, errors.New("reading dses: " + util.JoinErrsStr(errs)), http.StatusInternalServerError
+	dses, userErr, sysErr, errCode := readGetDeliveryServices(ds.APIInfo().Params, ds.APIInfo().Tx, ds.APIInfo().User)
+
+	if sysErr != nil {
+		sysErr = errors.New("reading dses: " + sysErr.Error())
+		errCode = http.StatusInternalServerError
+	}
+	if userErr != nil || sysErr != nil {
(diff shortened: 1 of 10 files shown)


traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -821,14 +821,19 @@ func readGetDeliveryServices(params map[string]string, tx *sqlx.Tx, user *auth.C
 
 	where, orderBy, pagination, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(params, queryParamsToSQLCols)
 	if len(errs) > 0 {
-		return nil, errs, tc.DataConflictError
+		for _, err := range errs {
+			if err.Error() == `id cannot parse to integer` { // TODO create const for string
```
mhoppa: why the addition here?  Also why does an id not being able to be parsed to an int mean not found?

FWIW In our generic read, that is used across many handlers, we just join the errors and mark it is a bad request. Should we follow that here?

zrhoffman: @mhoppa I saw "matches perl response" and didn't immediately want to break BC. I agree that marking it as bad request makes more sense.

More info:

`TODeliveryService.Read()` has returned `tc.DataMissingError` for that case [since deliveryservices's original Go implementation](https://github.com/apache/trafficcontrol/blob/75a949dedb/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservicesv13.go#L280-L284) in 75a949dedb. Because `readGetDeliveryServices()`'s return values no longer include `[]error`, moving the logic from `TODeliveryService.Read()` to `readGetDeliveryServices()` is one way to preserve that response.

The original Perl module [did not distinguish](https://github.com/apache/trafficcontrol/blob/55ff2567d1/traffic_ops/app/lib/API/DeliveryService.pm#L53-L55) between not finding a delivery service and the id param failing int validation. So, even though that behavior was [originally documented](https://github.com/apache/trafficcontrol/blob/55ff2567d1/traffic_ops/doc/api_response_examples.txt#L165-L171), now that we have type safety, `http.StatusBadRequest` makes more sense.

zrhoffman: Addressed in ff7334ac25.

PR #3991, opened 2019-10-14
Fixed pip not working in CiaB

ocket8888:

This PR fixes an issue where `pip` was crashing when building CDN-in-a-Box images. The problem is that CentOS does not provide a way to install the modern `pip` needed by some dependencies, but `pip install --upgrade pip` generally breaks the `pip` install. The solution is to skip running the setuptools-provided entrypoint and directly call the package using `python3 -m pip`.

- CDN in a Box

Destroy your Docker cache, then confirm that the `mid` and `edge` services can be built. Note that at the time of this writing, the ATS CI server is experiencing problems, so testing may be difficult.

- master


ocket8888:
retest this please

PR #3996, opened 2019-10-15
Rewrote /user/current to Go

### LINKED ISSUE #3838: Rewrite /user/current to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/user/current` is handled by Perl

## Expected / new behavior:
`/user/current` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [x] `GET` (#2405)
- [ ] `PUT`

N/A

ocket8888:
- [x] This PR fixes #3838 

This rewrites the previously unimplemented PUT handler of `/user/current` into Go. It adds support in both the Go and Python clients, with some associated testing as well as documentation updates.

~This differs from the way the endpoint was handled in Perl - in Perl, every field in the request was optional (under the top-level "user" object, obviously) and only those included would be updated - as though it were a PATCH request. This instead demands a full representation of a user as is proper for PUT. Additionally, the Perl returned only a success message, this handler returns a representation of the user as it exists following the update.~

Pursuant to a [mailing list discussion](https://mail-archives.apache.org/mod_mbox/trafficcontrol-dev/201910.mbox/%3CCAMAKGdA5TL6c%3DXJXpB%2BbVj4%3D5gqjUhokqCBFZk20BoUCxYHSxg%40mail.gmail.com%3E) changing this endpoint's behavior - albeit for the better - was shot down, and I re-rewrote it to implement Perl-style PATCH-like behavior.

- Documentation
- Traffic Control Client (Go)
- Traffic Ops

Build a TO RPM, start Traffic Ops, make a `PUT` request to `/user/current`, run the Go client/API integration tests, build and read the documentation.

- [x] This PR includes tests

### DIFF
--- a/docs/source/api/user_current.rst
+++ b/docs/source/api/user_current.rst
@@ -21,7 +21,9 @@
 
 ``GET``
 =======
-Retrieves the profile for the authenticated user.
+.. caution:: As a username is needed to log in, any administrator or application must necessarily know the current username at any given time. Thus it's generally better to use the ``username`` query parameter of a ``GET`` request to :ref:`to-api-users` instead.
+
+Retrieves the details of the authenticated user.
 
 :Auth. Required: Yes
 :Roles Required: None
@@ -40,20 +42,20 @@ Response Structure
 :country:          The name of the country wherein the user resides
 :email:            The user's email address
 :fullName:         The user's full name, e.g. "John Quincy Adams"
-:gid:              A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user - now it is always ``null``
+:gid:              A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user
 :id:               An integral, unique identifier for this user
-:lastUpdated:      The date and time at which the user was last modified, in ISO format
+:lastUpdated:      The date and time at which the user was last modified, in an ISO-like format
 :newUser:          A meta field with no apparent purpose that is usually ``null`` unless explicitly set during creation or modification of a user via some API endpoint
 :phoneNumber:      The user's phone number
 :postalCode:       The postal code of the area in which the user resides
 :publicSshKey:     The user's public key used for the SSH protocol
 :registrationSent: If the user was created using the :ref:`to-api-users-register` endpoint, this will be the date and time at which the registration email was sent - otherwise it will be ``null``
-:role:             The integral, unique identifier of the highest-privilege role assigned to this user
-:rolename:         The name of the highest-privilege role assigned to this user
+:role:             The integral, unique identifier of the highest-privilege :term:`Role` assigned to this user
+:rolename:         The name of the highest-privilege :term:`Role` assigned to this user
 :stateOrProvince:  The name of the state or province where this user resides
-:tenant:           The name of the tenant to which this user belongs
-:tenantId:         The integral, unique identifier of the tenant to which this user belongs
-:uid:              A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX user ID of the user - now it is always ``null``
+:tenant:           The name of the :term:`Tenant` to which this user belongs
+:tenantId:         The integral, unique identifier of the :term:`Tenant` to which this user belongs
+:uid:              A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX user ID of the user
 :username:         The user's username
 
 .. code-block:: http
@@ -98,8 +100,7 @@ Response Structure
 
 ``PUT``
 =======
-.. deprecated:: 1.4
-	Use the ``PUT`` method of the :ref:`to-api-users` instead.
+.. warning:: Assuming the current user's integral, unique identifier is known, it's generally better to use the ``PUT`` method of the :ref:`to-api-users` instead.
 
 .. warning:: Users that login via LDAP pass-back cannot be modified
 
@@ -107,34 +108,39 @@ Updates the date for the authenticated user.
 
 :Auth. Required: Yes
 :Roles Required: None
-:Response Type:  ``undefined``
+:Response Type:  Object
+
+	.. versionchanged:: ATCv4
+		Starting in ATC version 4, all API versions respond to this endpoint with the updated user information. Prior to that, no response object was returned at all.
+
 
 Request Structure
 -----------------
-:addressLine1:       An optional field which should contain the user's address - including street name and number
-:addressLine2:       An optional field which should contain an additional address field for e.g. apartment number
-:city:               An optional field which should contain the name of the city wherein the user resides
-:company:            An optional field which should contain the name of the company for which the user works
-:confirmLocalPasswd: The 'confirm' field in a new user's password specification - must match ``localPasswd``
-:country:            An optional field which should contain the name of the country wherein the user resides
-:email:              The user's email address
+:user: The entire request must be inside a top-level "user" key for legacy reasons
 
-	.. versionchanged:: 1.4
-		Prior to version 1.4, the email was validated using the `Email::Valid Perl package <https://metacpan.org/pod/Email::Valid>`_ but is now validated (circuitously) by `GitHub user asaskevich's regular expression <https://github.com/asaskevich/govalidator/blob/9a090521c4893a35ca9a228628abf8ba93f63108/patterns.go#L7>`_ . Note that neither method can actually distinguish a valid, deliverable, email address but merely ensure the email is in a commonly-found format.
+	:addressLine1:       The user's address - including street name and number
+	:addressLine2:       An additional address field for e.g. apartment number
+	:city:               The name of the city wherein the user resides
+	:company:            The name of the company for which the user works
+	:confirmLocalPasswd: The 'confirm' field in a new user's password specification - must match ``localPasswd``. This is optional; when ``localPasswd`` is not present in the request, this need not be present
+	:country:            The name of the country wherein the user resides
+	:email:              The user's email address\ [#notnull]_
 
-:fullName:        The user's full name, e.g. "John Quincy Adams"
-:localPasswd:     The user's password
-:newUser:         An optional meta field with no apparent purpose - don't use this
-:phoneNumber:     An optional field which should contain the user's phone number
-:postalCode:      An optional field which should contain the user's postal code
-:publicSshKey:    An optional field which should contain the user's public encryption key used for the SSH protocol
-:role:            The number that corresponds to the highest permission role which will be permitted to the user
-:stateOrProvince: An optional field which should contain the name of the state or province in which the user resides
-:tenantId:        The integral, unique identifier of the tenant to which the new user shall belong
+		.. versionchanged:: ATCv4
+			Prior to version ATCv4, the email was validated using the `Email::Valid Perl package <https://metacpan.org/pod/Email::Valid>`_ but is now validated (circuitously) by `GitHub user asaskevich's regular expression <https://github.com/asaskevich/govalidator/blob/9a090521c4893a35ca9a228628abf8ba93f63108/patterns.go#L7>`_ . Note that neither method can actually distinguish a valid, deliverable, email address but merely ensure the email is in a commonly-found format.
 
-	.. note:: This field is optional if and only if tenancy is not enabled in Traffic Control
-
-:username: The user's new username
+	:fullName:        The user's full name, e.g. "John Quincy Adams"
+	:gid:             A legacy field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user - please don't use this
+	:id:              The user's integral, unique, identifier - this cannot be changed\ [#notnull]_
+	:localPasswd:     Optionally, the user's password. This should never be given if it will not be changed. An empty string or ``null`` can be used to explicitly specify no change.
+	:phoneNumber:     The user's phone number
+	:postalCode:      The user's postal code
+	:publicSshKey:    The user's public encryption key used for the SSH protocol
+	:role:            The integral, unique identifier of the highest permission :term:`Role` which will be permitted to the user - this cannot be altered from the user's current :term:`Role`\ [#notnull]_
+	:stateOrProvince: The state or province in which the user resides
+	:tenantId:        The integral, unique identifier of the :term:`Tenant` to which the new user shall belong\ [#tenancy]_\ [#notnull]_
+	:uid:             A legacy field only kept for legacy compatibility reasons that used to contain the UNIX user ID of the user - please don't use this
+	:username:        The user's new username\ [#notnull]_
 
 .. code-block:: http
 	:caption: Request Example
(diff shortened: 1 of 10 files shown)


ocket8888:
retest this please

docs/source/api/user_current.rst
```diff
@@ -147,49 +150,99 @@ Request Structure
...
-		"company": "not a real company",
-		"country": "not a real country",
-		"email": "not@real.email",
-		"fullName": "Not a real fullName",
-		"phoneNumber": "not a real phone number",
-		"postalCode": "not a real postal code",
-		"publicSshKey": "not a real ssh key",
-		"stateOrProvince": "not a real state or province",
-		"tenantId": 1,
+		"addressLine1": null,
```
mhoppa: do we really want users to give null in the request? I really dislike that of a user of any API. As a user do I need to define the key and pass in null? Is that the same behavior as omitting the key? How do various programming languages clients handle setting `null` fields.

But maybe it is a common approach in TO which in that case fine leave it but I strongly side with removing null in requests/responses going forward

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -207,6 +207,7 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodPost, `users/?(\.json)?$`, api.CreateHandler(&user.TOUser{}), auth.PrivLevelOperations, Authenticated, nil},
 
 		{1.1, http.MethodGet, `user/current/?(\.json)?$`, user.Current, auth.PrivLevelReadOnly, Authenticated, nil},
+		{1.1, http.MethodPut, `user/current(/|\.json)?$`, user.ReplaceCurrent, auth.PrivLevelReadOnly, Authenticated, nil},
```
mhoppa: is readonly the right priv

traffic_ops/traffic_ops_golang/user/current.go
```diff
@@ -80,3 +139,166 @@ WHERE u.id=$1
...
+		user.PostalCode,
+		user.PublicSSHKey,
+		user.StateOrProvince,
+		user.TenantID,
+		user.UID,
+		user.Username,
+		inf.User.ID,
+	)
+
+	err = row.Scan(&user.AddressLine1,
```
mhoppa: think about using sqlx structscan? and sqlx for the query above instead of using $1,$2,$3,$4. With a big parametrized query like this I worry about maintenance/debugging

docs/source/api/user_current.rst
ocket8888: In order for `PUT` to remain idempotent, all fields ought to be defined in every request. We could treat missing fields as implicitly `null`, but that introduces problems when you add fields to a request. Because old clients won't know about new fields they'll `PUT` without them and accidentally set something to `null`. The Perl acted as a `PATCH` by only updating the fields it found in the request, but as that was breaking with the consistency of our API  - and seemingly RFC non-compliant - I elected to change the behavior.

mhoppa: `but that introduces problems when you add fields to a request. Because old clients won't know about new fields they'll PUT without them and accidentally set something to null` API versioning alleviate this concern?

traffic_ops/traffic_ops_golang/routing/routes.go
ocket8888: Yeah, Perl didn't check any permissions role for this endpoint. I actually think that's why it exists, it allows read-only users to edit their own information without needing to add a check for that in the `PUT /users` handler.

docs/source/api/user_current.rst
ocket8888: Yes and no. In general, yes, but because the old Perl version of the endpoint didn't require fields to be defined and I didn't feel like rewriting it that way was a good idea, it seems safer to just return an error rather than use implicit `null`s, because clients coded against the Perl version of APIs 1.1 to 1.3 won't expect anything to have changed. Short of implementing the behavior unchanged failing fast I think failing fast is the best option.

Now that I'm saying all this, I think maybe I ought to bring that up to the mailing list, seems like a big change for me to make with minimal community input.

lib/go-tc/users.go
```diff
@@ -99,6 +109,179 @@ type UserCurrent struct {
...
+	ID                 json.RawMessage `json:"id"`
+	LocalPasswd        *string         `json:"localPasswd"`
+	PhoneNumber        json.RawMessage `json:"phoneNumber"`
+	PostalCode         json.RawMessage `json:"postalCode"`
+	PublicSSHKey       json.RawMessage `json:"publicSshKey"`
+	Role               json.RawMessage `json:"role"`
+	StateOrProvince    json.RawMessage `json:"stateOrProvince"`
+	TenantID           json.RawMessage `json:"tenantId"`
+	UID                json.RawMessage `json:"uid"`
+	Username           json.RawMessage         `json:"username"`
```
mhoppa: gofmt

traffic_ops/traffic_ops_golang/user/current.go
```diff
@@ -80,3 +190,213 @@ WHERE u.id=$1
...
+
+	user, exists, err := dbhelpers.GetUserByID(inf.User.ID, tx)
+	if err != nil {
+		sysErr = fmt.Errorf("Getting user by ID %d: %v", inf.User.ID, err)
+		errCode = http.StatusInternalServerError
+		api.HandleErr(w, r, tx, errCode, nil, sysErr)
+		return
+	} else if !exists {
+		sysErr = fmt.Errorf("Current user (#%d) doesn't exist... ??", inf.User.ID)
+		errCode = http.StatusInternalServerError
```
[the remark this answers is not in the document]
ocket8888: if you get that the currently logged in user doesn't exist, then that can only mean the user was deleted in between when the request handling began and when it got to this point. As users cannot be deleted through the API that indicates a corrupt or compromised database, so I figured an internal server error was appropriate.

lib/go-tc/users.go
```diff
@@ -101,6 +110,179 @@ type UserCurrent struct {
...
+		if u.ConfirmLocalPasswd == nil || *u.ConfirmLocalPasswd == "" {
+			errs = append(errs, errors.New("confirmLocalPasswd: required when changing password"))
+		} else if *u.LocalPasswd != *u.ConfirmLocalPasswd {
+			errs = append(errs, errors.New("localPasswd and confirmLocalPasswd do not match"))
+		}
+	} else if u.ConfirmLocalPasswd != nil && *u.ConfirmLocalPasswd != "" {
+		errs = append(errs, errors.New("localPasswd: required when changing password"))
+	}
+	user.ConfirmLocalPassword = u.ConfirmLocalPasswd
+	user.LocalPassword = u.LocalPasswd
```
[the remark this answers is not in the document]
ocket8888: nah, they can be `nil`. It's just that if one isn't then the other can't be, and when they both aren't they have to match.

traffic_ops/traffic_ops_golang/user/current.go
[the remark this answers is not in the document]
mhoppa: agreed on reasoning ^

ocket8888:
Eh. I don't think that difference is actually going to break anyone, and I did make sure identifiers can't be null. And to be honest I'm sick to death of working on this.

traffic_ops/traffic_ops_golang/user/current.go
```diff
@@ -80,3 +190,197 @@ WHERE u.id=$1
...
+		return
+	}
+
+	user, exists, err := dbhelpers.GetUserByID(inf.User.ID, tx)
+	if err != nil {
+		sysErr = fmt.Errorf("Getting user by ID %d: %v", inf.User.ID, err)
+		errCode = http.StatusInternalServerError
+		api.HandleErr(w, r, tx, errCode, nil, sysErr)
+		return
+	} else if !exists {
```
ChrisHines: `else` is redundant here since the `if` block aways returns.

traffic_ops/traffic_ops_golang/user/current.go
```diff
@@ -80,3 +190,197 @@ WHERE u.id=$1
...
+	// TODO: check for valid password via bad password list like Perl did? User creation doesn't...
+	if user.LocalPassword != nil && *user.LocalPassword != "" {
+		hashPass, err := auth.DerivePassword(*user.LocalPassword)
+		if err != nil {
+			sysErr = fmt.Errorf("Hashing new password: %v", err)
+			errCode = http.StatusInternalServerError
+			api.HandleErr(w, r, tx, errCode, nil, sysErr)
+			return
+		}
+		changePasswd = true
```
ChrisHines: Reading this raises a question in my mind: Is it true that a `nil` or empty `*user.LocalPassword` means "no password change" requested?

traffic_ops/traffic_ops_golang/user/current.go
```diff
@@ -80,3 +190,197 @@ WHERE u.id=$1
...
+		api.HandleErr(w, r, tx, errCode, nil, sysErr)
+		return
+	}
+
+	api.WriteRespAlertObj(w, r, tc.SuccessLevel, "User profile was successfully updated", user)
+}
+
+func updateUser(u *tc.User, tx *sql.Tx, cp bool) error {
+	var row *sql.Row
+	if cp {
```
ChrisHines: It is not clear why two different queries are needed here. Can you explain?

lib/go-tc/users.go
```diff
@@ -101,6 +110,179 @@ type UserCurrent struct {
...
+	}
+
+	if u.FullName != nil {
+		if err := json.Unmarshal(u.FullName, &user.FullName); err != nil {
+			errs = append(errs, fmt.Errorf("fullName: %v", err))
+		} else if user.FullName == nil || *user.FullName == "" {
+			// Perl enforced this
+			errs = append(errs, fmt.Errorf("fullName: cannot be set to 'null' or empty string"))
+		}
+	}
```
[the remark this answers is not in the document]
ocket8888: When a `json.RawMessage` compares like `rm == nil` it means the value was `undefined` - not `null` or anything else "nil-like". If not, it could be `null`, an empty string, an object, an array, NaN, Infinity, -Infinity, or a real number.

Perl treated this endpoint like a PATCH, so I have to distinguish between `null` and `undefined`.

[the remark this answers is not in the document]
ocket8888: Actually, I think NaN and the infinities are extensions of the Python JSON parser, and outside of the JSON spec itself. But that's a bit beside the point, I suppose.

lib/go-tc/users.go
```diff
@@ -101,6 +110,179 @@ type UserCurrent struct {
...
+	StateOrProvince    json.RawMessage `json:"stateOrProvince"`
+	TenantID           json.RawMessage `json:"tenantId"`
+	UID                json.RawMessage `json:"uid"`
+	Username           json.RawMessage `json:"username"`
+}
+
+// ValidateAndUnmarshal validates the request and returns a User into which the request's information
+// has been unmarshalled. This allows many fields to be "null", but explicitly checks that they are
+// present in the JSON payload.
+func (u *CurrentUserUpdateRequestUser) ValidateAndUnmarshal(user *User) error {
```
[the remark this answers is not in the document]
ocket8888: It does both concurrently, so I'm not sure the ordering is important.

traffic_ops/traffic_ops_golang/user/current.go
ocket8888: That's true. I don't really think it's a big deal either way, but I can change it if you want.

lib/go-tc/users.go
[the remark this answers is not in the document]
ChrisHines: Your top PR comment led me to believe that the new Go implementation was intended to more strictly adhere to the semantics of PUT and require a full user representation. Did that intent change or am I misunderstanding something?

lib/go-tc/users.go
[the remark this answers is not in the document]
ChrisHines: I don't think it is a big deal. Just something I noticed that each field is first unmarshaled and then validated. It also makes sense in general that parsing usually happens before validation in most code. It was just a suggestion.

traffic_ops/traffic_ops_golang/user/current.go
ocket8888: Yep. If there were no errors from ValidateAndUnmarshal, it means that either `localPasswd` and `confirmLocalPasswd` were both one of `null`, `undefined`, or an empty string, *or* they were both none of those and were strings that matched. So if `user.LocalPassword` is not `nil` and points to a non-empty string, then the user submitted strings of non-zero length in the `confirmLocalPasswd` and `localPasswd` fields that were identical. It could've been their current password, but in that case updating it is essentially a no-op.

traffic_ops/traffic_ops_golang/user/current.go
ocket8888: Because if you don't change your password, then `u.LocalPassword` and `u.ConfirmLocalPassword` are both either `nil` or pointers to empty strings. It would be unacceptable to update a user to have either of those values as a password. So I update those fields if and only if a password change was requested.

traffic_ops/traffic_ops_golang/user/current.go
ChrisHines: Which ever way you think the code is simpler and most clear. I personally think removing the `else` reduces the cognitive load of reading the code.

lib/go-tc/users.go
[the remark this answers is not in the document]
ocket8888: Oh, yeah, I didn't update the PR. Pursuant to a [mailing list discussion](https://mail-archives.apache.org/mod_mbox/trafficcontrol-dev/201910.mbox/%3CCAMAKGdA5TL6c%3DXJXpB%2BbVj4%3D5gqjUhokqCBFZk20BoUCxYHSxg%40mail.gmail.com%3E) changing this endpoint's behavior - albeit for the better - was shot down, and I re-rewrote it to implement Perl-style PATCH-like behavior.

traffic_ops/traffic_ops_golang/user/current.go
ChrisHines: I think I see, so there is some UI on the client side that only passes password data if the user opted in to changing it, and `GetUserByID` does not populate the password fields so we can't just write those back. Maybe a comment explaining these issues would help avoid mistakes by future maintainers.

ocket8888: You mean on the the `updateUser` subroutine or like a warning on `GetUserByID` stating that it doesn't return password info?

ChrisHines: I was thinking on the `updateUser` function. I have a different suggestion that occurred to me over the weekend, though.

Would it be acceptable to split the updating of the password fields and the rest of the user fields into two separate queries?

Since `updateUser` is given a `sql.Tx`, I assume that the caller has proper transaction management with rollback if we return an error so that two queries would have the same atomicity properties as the current code. It does have the potential to be slightly less performant, but I don't know how much that matters here.

On the other hand, I think we would gain a lot of code simplicity and readability. At the moment we have an if statement that chooses between two almost identical queries that are long enough to make it difficult to eyeball the difference between them. But if we split out the password update query we end up with an `updateUser` function structured along the lines of this:

```go
func updateUser(u *tc.User, tx *sql.Tx, cp bool) error {
    row := tx.QueryRow(replaceCurrentQuery,
        u.AddressLine1,
        ...
    )
    err := row.Scan(&u.AddressLine1,
        ...
    )
    if err != nil {
        return err
    }

    if cp {
        _, err := tx.Exec(replacePasswordQuery, u.LocalPassword, u.ConfirmLocalPassword)
        if err != nil {
            return err
        }
        u.LocalPassword = nil
        u.ConfirmLocalPassword = nil
    }
    return nil
}
```

(Note: I'm not sure about the correct place to nil out the password fields here, so adjust as necessary.)

With the above structure it becomes much easier to see what the `cp` parameter controls and how the password fields are managed. I think it is also less likely for a future maintainer to screw up the password handling.

Thoughts?

lib/go-tc/users.go
```diff
@@ -101,6 +110,179 @@ type UserCurrent struct {
...
+		if err := json.Unmarshal(u.UID, &user.UID); err != nil {
+			errs = append(errs, fmt.Errorf("uid: %v", err))
+		}
+	}
+
+	if u.Username != nil {
+		if err := json.Unmarshal(u.Username, &user.Username); err != nil {
+			errs = append(errs, fmt.Errorf("username: %v"))
+		} else if user.Username == nil || *user.Username == "" {
+			errs = append(errs, errors.New("username: cannot be null of empty string"))
```
awilliams: I believe there's a typo in: 
```
username: cannot be null of empty string
```

lib/go-tc/users.go
```diff
@@ -101,6 +110,179 @@ type UserCurrent struct {
...
+		}
+	}
+
+	var err error
+	if len(errs) > 0 {
+		err = util.JoinErrs(errs)
+	}
+	return err
```
awilliams: Could this be simplified as the following?

```go
return util.JoinErrs(errs)
```

mhoppa:
can you rebase?

traffic_ops/traffic_ops_golang/user/current.go
```diff
@@ -80,3 +144,184 @@ WHERE u.id=$1
...
+		errCode = http.StatusBadRequest
+		userErr = fmt.Errorf("Couldn't parse request: %v", err)
+		api.HandleErr(w, r, tx, errCode, userErr, nil)
+		return
+	}
+
+	changePasswd := false
+
+	// obfuscate passwords (UnmarshalAndValidate checks for equality with ConfirmLocalPassword)
+	// TODO: check for valid password via bad password list like Perl did? User creation doesn't...
```
rawlinp: I don't believe this is true. The shared handlers call the `Validate` method after decoding the json request body, and if the localPasswd is not nil, it checks to make sure the username and localPassword are a good login pair.

traffic_ops/traffic_ops_golang/user/current.go
```diff
@@ -80,3 +144,184 @@ WHERE u.id=$1
...
+		&u.Tenant,
+		&u.TenantID,
+		&u.UID,
+		&u.Username,
+	)
+	if err != nil {
+		return err
+	}
+
+	if cp {
```
rawlinp: `changePassword` was too much to type?

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -62,6 +62,35 @@ WHERE federation_deliveryservice.deliveryservice IN (
...
+       role.name AS role_name,
+       tm_user.state_or_province,
+       tenant.name AS tenant,
+       tm_user.tenant_id,
+       tm_user.token,
+       tm_user.uid,
+       tm_user.username
+FROM tm_user
+LEFT OUTER JOIN role ON role.id = tm_user.role
+LEFT OUTER JOIN tenant ON tenant.id = tm_user.tenant_id
```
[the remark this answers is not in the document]
ocket8888: Well if there's a constraint that's pure API I'd prefer not to write SQL assuming it's true. Bad data exists, and I don't wanna introduce bugs. But I can make the tenant join `INNER` for sure.

traffic_ops/traffic_ops_golang/user/current.go
ocket8888: Users doesn't use GenericCreate, the only validation I can see being done before the user is `INSERT`ed is checking that the password is non-nil. I also can't find a `tc.User.Validate` function, so if there is more validation being done my question is: where? I'd like to see it so I know how to write that enforcement.

rawlinp: `POST /users` uses `api.CreateHandler`, which eventually calls `TOUser.Validate`.

ocket8888: Ah thanks. I'm used to seeing `Validate` in `/lib/go-tc/`. Should be checking password validity now.

lib/go-tc/users.go
```diff
@@ -86,10 +94,11 @@ type commonUserFields struct {
...
 // User fields in v14 have been updated to be nullable
 type User struct {
-	Username         *string    `json:"username" db:"username"`
-	RegistrationSent *TimeNoMod `json:"registrationSent" db:"registration_sent"`
-	LocalPassword    *string    `json:"localPasswd,omitempty" db:"local_passwd"`
-	RoleName         *string    `json:"roleName,omitempty" db:"-"`
+	Username             *string    `json:"username" db:"username"`
+	RegistrationSent     *TimeNoMod `json:"registrationSent" db:"registration_sent"`
+	LocalPassword        *string    `json:"localPasswd,omitempty" db:"local_passwd"`
+	ConfirmLocalPassword *string    `json:"confirmLocalPasswd,omitempty" db:"confirm_local_passwd"`
```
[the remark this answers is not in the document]
ocket8888: So, Perl never checked that they matched but it *did* update it with a hash of the passed value when it was present and not an empty string. I've made updates so that it matches the Perl behavior by not checking that they match, not checking that both are present when either one is, and updating it whenever it's present and not empty. Is that acceptable?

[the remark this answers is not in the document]
rawlinp: While I think we'd be okay completely ignoring it in the rewritten version, I think matching the behavior Perl version is acceptable as well.

PR #4004, opened 2019-10-18
Make /deliveryservice_stats accessible to read-only users

### LINKED ISSUE #3759: deliveryservice_stats should be accessible to readonly users
## I'm submitting a ...
- [ ] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Currently, the `/deliveryservice_stats` endpoint requires permissions of at least the "operator" role. The endpoint is only capable of reading back data, and should thusly be available to all users at the "readonly" level or above (presuming tenancy is satisfied).

(*Currently* the endpoint actually only considers role if the delivery service is not accessible to the tenant - behavior will be changed pending merge of #3758 to no longer override role when tenancy is correct)

## Expected / new behavior:
Simply change the check to look for at least "readonly" privileges.

ocket8888:
- [x] This PR fixes #3759

This changes the required Role for a `GET` request to `/deliveryservice_stats` to "Read-Only", down from "Operations". Though the Perl handler explicitly checked for the "Operations" (or "Admin") Role, it would override any and all Role checking when the Delivery Service was assigned to the requesting user. The result is that existing users of possibly as low as the "Read-Only" Role expect access to this endpoint. Since the plan is to transition from assigning users to Delivery Services to using Tenancy for such management, the rewrite uses Tenancy and ignores user-to-Delivery Service assignments. Furthermore, because Tenancy and Roles are totally orthogonal checks, the access role must be relaxed to allow users to continue using the endpoint as they are used to.

This PR also includes documentation updates, and fixes for problems that exist in the documentation not related to this PR, and modifies CDN-in-a-Box's configuration to enable InfluxDB connections by default.

- CDN in a Box
- Documentation
- Traffic Ops

Handling logic is not being changed, so no additional tests are required

- Build and install Traffic Ops from this revision (probably using CDN-in-a-Box, up to you)
- Run Traffic Ops
- Log in (via API or a connected Traffic Portal) as a user with Read-Only permissions (user may need to be created if using CiaB)
- Make a `GET` request to `/api/1.2/deliveryservice_stats`
- Observe that user permissions are not denied.

Handling logic is not being changed, so no additional tests are required


docs/source/api/deliveryservice_stats.rst
```diff
@@ -26,7 +26,11 @@
 Retrieves time-aggregated statistics on a specific :term:`Delivery Service`.
 
 :Auth. Required: Yes
-:Roles Required: None
+:Roles Required: None\ [#tenancy]_
```
mitchell852: shouldn't this say "Read-Only and above" or however we're indicating "this role or anything greater"

ocket8888: "None" is standard for "read-only" or above, because that's how the old docs did it. It can be changed - and that would be more correct - but that should be done globally.

infrastructure/cdn-in-a-box/traffic_ops/config.sh
```diff
@@ -129,7 +129,8 @@ cat <<-EOF >/opt/traffic_ops/app/conf/cdn.conf
         "password" : "",
         "address" : ""
     },
-    "influxdb_conf_location": "/opt/traffic_ops/app/conf/production/influx.conf"
+    "InfluxEnabled": true,
```
mhoppa: good catch - I do wish the config parameter followed the `snake_case` style but not a change for this PR!

mhoppa: lets clean it up in another PR  and clean it up across the board - this PR looks good

ocket8888:
retest this please

ocket8888:
retest this please

PR #4006, opened 2019-10-19
implement deliveryservice_required_capabilities api endpoint

jheitz200:

Add deliveryservice_required_capabilities support to the API.
- Add the ability to GET, POST, DELETE an association between a delivery service and server capability.
- Includes tests, API handlers, migrations

- Documentation - being added by another developer
- Traffic Control Client - Go
- Traffic Ops

Apply migrations to db and seed server_capabilities

Build new traffic ops API and then associate/get/disassociates the newly seeded server_capability to deliveryservice via the new api path /deliveryservice_server_capabilities


traffic_ops/testing/api/v14/cookie.txt
```diff
@@ -0,0 +1,5 @@
+# Netscape HTTP Cookie File
+# https://curl.haxx.se/docs/http-cookies.html
+# This file was generated by libcurl! Edit at your own risk.
+
+#HttpOnly_localhost	FALSE	/	FALSE	1571451749	mojolicious	eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1NzE0NTE3NDksImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--acef429741dafd8da4c4e20ab26e772aa9ea4412
```
ocket8888: I don't think you meant to commit this file

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservice_server_capabilities.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// Update implements the api.CRUDer interface
+func (sc *ServerCapability) Update() (error, error, int) {
+	return nil, nil, http.StatusNotImplemented
+}
+
+// Read implements the api.CRUDer interface
+func (sc *ServerCapability) Read() ([]interface{}, error, error, int) {
+	return api.GenericRead(sc)
```
ocket8888: This isn't checking Tenancy on the "objects"' Delivery Service(s). Is this API not meant to respect Tenancy? Because it ought to - at the very least if an endpoint gives away that a Delivery Service exists it's circumventing the protection provided by Tenancy checks on the `/deliveryservices` endpoint.

jheitz200:
@ocket8888 another developer will be adding the documentation. It should be available Monday

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservice_server_capabilities.go
jheitz200: @ocket8888 I will look into this, this is my first PR against the go-api. I havent heard of this yet.

jheitz200: @ocket8888 thanks for the comment, I wasnt familiar with this. I will update the code to check tenancy.

mitchell852:
@jheitz200 - because there are no docs with this PR (as they will be coming with a subsequent PR) it makes it hard to understand the api design (without reading the code). Do you think you could beef up the PR description a bit like @rawlinp did with #3966 ?

traffic_ops/app/db/seeds.sql
```diff
@@ -541,6 +541,10 @@ insert into api_capability (http_method, route, capability) values ('GET', 'deli
...
 insert into api_capability (http_method, route, capability) values ('POST', 'deliveryservices/*/servers', 'delivery-service-servers-write') ON CONFLICT (http_method, route, capability) DO NOTHING;
 insert into api_capability (http_method, route, capability) values ('DELETE', 'deliveryservice_server/*/*', 'delivery-service-servers-write') ON CONFLICT (http_method, route, capability) DO NOTHING;
+-- delivery service server capabilities
+insert into api_capability (http_method, route, capability) values ('GET', 'deliveryservice_server_capabilities', 'delivery-services-read') ON CONFLICT (http_method, route, capability) DO NOTHING;
+insert into api_capability (http_method, route, capability) values ('POST', 'deliveryservice_server_capabilities', 'delivery-services-write') ON CONFLICT (http_method, route, capability) DO NOTHING;
+insert into api_capability (http_method, route, capability) values ('DELETE', 'deliveryservice_server_capabilities', 'delivery-service-server-capabilities-write') ON CONFLICT (http_method, route, capability) DO NOTHING;
```
mhoppa: Should the delete capability be `delivery-services-write`?

traffic_ops/client/deliveryservice_server_capabilities.go
```diff
@@ -0,0 +1,79 @@
...
+		return tc.Alerts{}, reqInf, err
+	}
+	reqInf, err = post(to, v14DeliveryServiceServerCapabilities, reqBody, &alerts)
+	return alerts, reqInf, err
+}
+
+// DeleteDeliveryServiceServerCapability unassigns a Server Capability from a Delivery Service
+func (to *Session) DeleteDeliveryServiceServerCapability(deliveryserviceID int, serverCapability string) (tc.Alerts, ReqInf, error) {
+	var alerts tc.Alerts
+	endpoint := fmt.Sprintf("%v?deliveryServiceID=%v&serverCapability=%v", v14DeliveryServiceServerCapabilities, deliveryserviceID, serverCapability)
```
mhoppa: Since you used url.Values here lets do that here as well for the query parameters

traffic_ops/testing/api/v14/deliveryservice_server_capabilities_test.go
```diff
@@ -0,0 +1,211 @@
...
+			capability  tc.DeliveryServiceServerCapability
+			expected    int
+			err         string
+		}{
+			{
+				description: "get all deliveryservice server capabilities",
+				expected:    len(testData.DeliveryServiceServerCapabilities),
+			},
+			{
+				description: fmt.Sprintf("get all deliveryservice server capabilities by delivery service id"),
```
mhoppa: nit but no need for `fmt.Sprintf` for any of the descriptions

traffic_ops/testing/api/v14/deliveryservice_server_capabilities_test.go
```diff
@@ -0,0 +1,211 @@
...
+	"strings"
+	"testing"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/lib/go-util"
+)
+
+func TestGetDeliveryServiceServerCapabilities(t *testing.T) {
+	WithObjs(t, []TCObj{CDNs, Types, Tenants, Users, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, ServerCapabilities, DeliveryServices, DeliveryServiceServerCapabilities}, func() {
+		testCases := []struct {
```
mhoppa: to match other api tests should we move this to a separate test func `GetTestDeliveryServiceServerCapabilities`

traffic_ops/testing/api/v14/tc-fixtures.json
```diff
@@ -497,7 +494,6 @@
             "cdnName": "cdn1",
             "cacheurl": "cacheUrl1",
             "ccrDnsTtl": 3600,
-            "cdnName": "cdn1",
```
mhoppa: reason for removing cdnName?

traffic_ops/testing/api/v14/todb.go
```diff
@@ -275,6 +281,20 @@ INSERT INTO to_extension (name, version, info_url, isactive, script_file, server
 	return nil
 }
 
+// SetupServerCapabilities sets up seed server capabilities
```
mhoppa: The seeding of server_capabilities is not needed

traffic_ops/app/db/seeds.sql
jheitz200: Thanks for pointing this out, I didnt understand these roles and forgot to change that one

traffic_ops/testing/api/v14/tc-fixtures.json
jheitz200: @mhoppa it appears to be a duplicate
```
>  "cdnName": "cdn1",
     "cacheurl": "cacheUrl1",
      "ccrDnsTtl": 3600,
>    "cdnName": "cdn1",```

File: CHANGELOG.md [hunk 9]
```diff
@@ -54,6 +55,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added an optional SMTP server configuration to the TO configuration file, api now has unused abilitiy to send emails
 - In Traffic Portal, added the ability to create, view and delete server capabilities.
 - Traffic Monitor now has "gbps" calculated stat, allowing operators to monitor bandwidth in Gbps.
+- Added an API 1.4 endpoint, /api/1.4/deliveryservice_server_capabilities, to create, read, and delete associations between a delivery service and a server capability.
```
mitchell852: deliveryservice_server_capability, right? not plural?

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservice_server_capabilities.go
```diff
@@ -0,0 +1,196 @@
...
+	deliveryServiceQueryParam  = "deliveryServiceID"
+	serverCapabilityQueryParam = "serverCapability"
+	xmlIDQueryParam            = "xmlID"
+)
+
+// ServerCapability is a type alias to define functions on
+type ServerCapability struct {
+	api.APIInfoImpl `json:"-"`
+	tc.DeliveryServiceServerCapability
+}
```
ChrisHines: What does "type alias" mean in this context? Go has a specific meaning for a type alias in the spec (https://golang.org/ref/spec#Type_declarations), but this code is not declaring that kind of type alias. Consider different language in the comment here to avoid confusion.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservice_server_capabilities_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/lib/go-util"
+
```
ChrisHines: Extra blank line.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservice_server_capabilities_test.go
```diff
@@ -0,0 +1,188 @@
...
+		},
+		tc.DeliveryServiceServerCapability{},
+	}
+
+	rows := sqlmock.NewRows([]string{"server_capability", "deliveryservice_id", "last_updated"}).AddRow(
+		util.StrPtr("mem"),
+		util.IntPtr(1),
+		time.Now(),
+	)
+	mock.ExpectQuery("INSERT*").WillReturnRows(rows)
```
ChrisHines: Is `ExpectQuery` the right choice for an INSERT? In my experience an INSERT should be run with a `DB.Exec` not `DB.Query` and would not return any rows. I expected to see something like `mock.ExpectExec(regex).WillReturnResult(result)`. See this sqlmock example: https://godoc.org/github.com/DATA-DOG/go-sqlmock#example-ExpectedExec

Consider passing a more specific regex here. (BTW, the current regex will accept both "INSER" and "INSERTTTTTT" anywhere in the query string.) I suggest requiring "INSERT" to be at the beginning of the query and checking that the insert will go to the right table with a regex similar to "^INSERT INTO "+tableName+".*".

traffic_ops/testing/api/v14/todb.go
```diff
@@ -103,6 +103,12 @@ func SetupTestData(*sql.DB) error {
 		os.Exit(1)
 	}
 
+	err = SetupServerCapabilities(db)
+	if err != nil {
+		fmt.Printf("\nError setting up server capabilities %s - %s, %v\n", Config.TrafficOps.URL, Config.TrafficOps.Users.Admin, err)
+		os.Exit(1)
```
ChrisHines: I know this PR is just following the established pattern here, but this function can never return a non-nil error since it calls os.Exit whenever it gets an error from the functions it calls. Perhaps something to note with a TODO comment to follow up on later.

traffic_ops/testing/api/v14/todb.go
jheitz200: I think this was a merge issue, they must have been removed from master after I submitted my PR

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservice_server_capabilities_test.go
ocket8888: FWIW a bunch of our `INSERT` queries will return rows via the `RETURNING` clause. Dunno if that's what's supposed to happen here or not. Those typically use [`database/sql.Tx.QueryRow`](https://godoc.org/database/sql#Tx.QueryRow).

PR #4010, opened 2019-10-22
Port legacy perl TO extension checks from perl to go

guzzijason:

- Traffic Ops Extensions (server checks) 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added an optional SMTP server configuration to the TO configuration file, api now has unused abilitiy to send emails
 - In Traffic Portal, added the ability to create, view and delete server capabilities.
 - Traffic Monitor now has "gbps" calculated stat, allowing operators to monitor bandwidth in Gbps.
+- Port Traffic Ops Extensions checks for ILO, 10G, 10G6, MTU, DSCP, CHR, CDU from perl to go (legacy versions retained)
 
 ### Changed
 - Traffic Router:  TR will now allow steering DSs and steering target DSs to have RGB enabled. (fixes #3910)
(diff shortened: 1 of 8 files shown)


ocket8888:
What's the motivation for this? Check extensions just run as scripts with no actual dependency on the Perl TO codebase, right?

ocket8888:
ok to test

guzzijason:
@ocket8888 I just emailed the dev/users lists with some more info regarding motivation.

The only dependency was an update to the go client library (https://github.com/apache/trafficcontrol/pull/3470) which has already been merged.

smalenfant:
Can we add those to the Traffic Ops build? Would be nice to be available and compiled when installing traffic_ops RPM.

guzzijason:
Do not merge - there were changes in #ccff78a that broke the code in this PR, that I was not anticipating. I need to sort that out now.

guzzijason:
I'm satisfied that the code in this PR is now building and running satisfactorily with the go client library version that are in currently `master`.

ocket8888:
Can you rebase to fix those conflicts?

I also think you probably have a better chance of getting this reviewed if each extension is a separate PR, but that's just my opinion.

smalenfant:
@guzzijason Can we rebase this on master again please? Some issues with the Traffic Ops client versioning (my guess)

guzzijason:
@smalenfant yes, I've recently updated the ones we're using internally here. I'll work on updating this PR.

smalenfant:
@guzzijason In light of TC 6 to remove the API 1.x API, could we resolve the conflict on this one?

smalenfant:
How do we build those components?  I tried to "wget" them but I can't figure how to compile them. I would like to install them on a different server than my Traffic Ops as well.

smalenfant:
It actually looks like they would be packaged in the Traffic Ops RPM. But the build failed.

```
github.com/apache/trafficcontrol/traffic_ops/app/bin/checks imports
	github.com/apache/trafficcontrol/traffic_ops/client: no matching versions for query "latest"
+ echo 'Could not vendor go module dependencies'
```

ocket8888:
The unversioned `client` import path was deprecated in 4.0 or 4.1 and disabled/removed in 5.0, I believe. Needs to add a version prefix - should probably just use `v4-` since it's packaged with TO.

PR #4011, opened 2019-10-22
Implement GET /api/1.1/roles handler

### LINKED ISSUE #3820: Rewrite /roles to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/roles` is handled by Perl

## Expected / new behavior:
`/roles` should be handled by Go.

### Minimum API Version
1.1 - **This is the real issue here.** The current Go implementation has a minimum version of 1.3

### Request Methods
- [ ] `GET`
(Go added other request methods in API v1.3 - this was the only one in Perl)

N/A

rawlinp:
Modify the 1.3 GET handler to also handle requests at 1.1 (the only
difference is the addition of the "capabilities" field in 1.3).

- [x] This PR fixes #3820

- Traffic Ops

1. Run the TO API tests
2. `GET /api/1.1/roles` and `GET /api/1.2/roles`, verify that the `capabilities` field **does not exist** in the result.
3. `GET /api/1.3/roles` and `GET /api/1.4/roles`, verify that the `capabilities` field **does exist** in the result.

- [x] This PR includes tests
- [x] Documentation already exists for /api/1.1/roles

### DIFF
--- a/traffic_ops/traffic_ops_golang/role/roles.go
+++ b/traffic_ops/traffic_ops_golang/role/roles.go
@@ -159,16 +159,32 @@ func (role *TORole) deleteRoleCapabilityAssociations(tx *sqlx.Tx) (error, error,
 }
 
 func (role *TORole) Read() ([]interface{}, error, error, int) {
+	version := role.APIInfo().Version
+	if version == nil {
+		return nil, nil, errors.New("TORole.Read called with nil API version"), http.StatusInternalServerError
+	}
+	if version.Major != 1 || version.Minor < 1 {
+		return nil, nil, fmt.Errorf("TORole.Read called with invalid API version: %d.%d", version.Major, version.Minor), http.StatusInternalServerError
+	}
 	vals, userErr, sysErr, errCode := api.GenericRead(role)
 	if userErr != nil || sysErr != nil {
 		return nil, userErr, sysErr, errCode
 	}
+	returnable := []interface{}{}
 	for _, val := range vals {
 		rl := val.(*TORole)
-		caps := ([]string)(*rl.PQCapabilities)
-		rl.Capabilities = &caps
+		switch {
+		case version.Minor >= 3:
+			caps := ([]string)(*rl.PQCapabilities)
+			rl.Capabilities = &caps
+			returnable = append(returnable, rl)
+		case version.Minor >= 1:
+			returnable = append(returnable, rl.RoleV11)
+		default:
+			return nil, nil, fmt.Errorf("TORole.Read called with invalid API version: %d.%d", version.Major, version.Minor), http.StatusInternalServerError
+		}
 	}
-	return vals, nil, nil, http.StatusOK
+	return returnable, nil, nil, http.StatusOK
 }
 
 func (role *TORole) Update() (error, error, int) {
--- a/traffic_ops/traffic_ops_golang/role/roles.go
+++ b/traffic_ops/traffic_ops_golang/role/roles.go
@@ -160,12 +160,6 @@ func (role *TORole) deleteRoleCapabilityAssociations(tx *sqlx.Tx) (error, error,
 
 func (role *TORole) Read() ([]interface{}, error, error, int) {
 	version := role.APIInfo().Version
-	if version == nil {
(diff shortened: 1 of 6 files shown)


traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -159,16 +159,32 @@ func (role *TORole) deleteRoleCapabilityAssociations(tx *sqlx.Tx) (error, error,
 }
 
 func (role *TORole) Read() ([]interface{}, error, error, int) {
+	version := role.APIInfo().Version
+	if version == nil {
+		return nil, nil, errors.New("TORole.Read called with nil API version"), http.StatusInternalServerError
+	}
+	if version.Major != 1 || version.Minor < 1 {
+		return nil, nil, fmt.Errorf("TORole.Read called with invalid API version: %d.%d", version.Major, version.Minor), http.StatusInternalServerError
+	}
```
ocket8888: I don't think you need to do this, because that check ought to be caught by [some middleware](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/routing/wrappers.go#L281-L287).

rawlinp: Yeah, I was on the fence about whether or not to include this check. This condition would only be true if there was a bug introduced into the routing middleware, so really it's just an extra safety check in case that ever happens. I can remove it.

rawlinp:
Fixed. Those are interesting unit test failures. I assumed anonymously embedding a new `RoleV11` struct into `Role` would be non-breaking, but I guess that actually breaks any previous usage of struct literals for `Role`. That's unfortunate :(.

rob05c:
>Fixed. Those are interesting unit test failures. I assumed anonymously embedding a new RoleV11 struct into Role would be non-breaking, but I guess that actually breaks any previous usage of struct literals for Role. That's unfortunate :(.

Yeah, it's an unfortunate quirk that Go doesn't allow direct access in embedded literals.

I try to encourage
```
foo := Foo{}
foo.Bar = 42
```

Instead of  `foo := Foo{Bar: 42}` For that reason.

In fact, we could prevent the latter by making our versions have the latest always be empty. For example, if the latest version is 1.5, we currently
```
type DeliveryService {
  DeliveryServiceV14
  New15Field int
}
```

I've wondered if instead we should:

```
type DeliveryService struct {
  DeliveryServiceV15
}

type DeliveryServiceV15 {
  DeliveryServiceV14
  New15Field int
}
```

That would force people to either use the assignment syntax above, which keeps working with future versions, or else use the incredibly-awkward deeply-nested literals (which hopefully nobody would do).

rawlinp:
I didn't write the PUT/POST/1.3+GET handlers. IMO we shouldn't even be able to create/update `roles` today since they're built-in, but I guess someone decided we needed this API for "role and capabilities" which was kind of started but never finished.

rawlinp:
Also, it seems intentional that `PUT` doesn't actually update the priv_level, since that would go away with roles and capabilities. The `PUT` response issue is just a byproduct of how our API is implemented. The struct that the request body is unmarshalled into is the same struct that is returned in the response.

ocket8888:
Forgot this was only a change for `GET` for a second there. But just so I'm clear, you're saying the issue is in `api.GenericUpdate`/some part of the "CRUDer" interface implementations on the Role object? Because the update handler doesn't itself seem to be aware that changing privilege level is impossible:
```go
if *role.PrivLevel > role.ReqInfo.User.PrivLevel {
	return errors.New("can not create a role with a higher priv level than your own"), nil, http.StatusForbidden
}
```

rawlinp:
Ok, now that's pretty funny. I guess this API wasn't reviewed/tested all that well before it was merged. But yes, that bad `PUT` response problem is a general problem with our API. For instance, you can `PUT` with ignored fields in the request body (e.g. most name fields, since we use IDs for almost everything), and those ignored fields will be returned in the response body.

ocket8888:
Hm. I thought that was fixed a while back, though.

rawlinp:
> Hm. I thought that was fixed a while back, though.

A holistic solution was never merged AFAIK.

PR #4012, opened 2019-10-22
Added IPv6 checking to Traffic Monitor

### LINKED ISSUE #2735: Traffic Monitor Polling over ipv6
Currently, Traffic Monitor only polls caches over IPv4, this can lead to issues if a DS is ipv6 enabled and the ipv6 interface is down on a cache.  Traffic Monitor should be modified such that it can poll on the necessary interfaces and report status based on the interfaces. Ideally this would be a config that can be set to tell TM to monitor IPv4, Ipv6 or both and make health determinations based on those configs.

mattjackson220:

This PR is an updated version of PR 1627.  It upgrades Traffic Monitor to poll over IPv6 as well as IPv4.
The config has two new parameters: cache_polling_protocol and peer_polling_protocol
the available values are IPV4Only, IPV6Only or Both. They default to IPV4Only so with no changes to the config there will be no change in behavior.

cache_polling_protocol controls the interface used to poll the caches, setting it to both will cause it to alternate which interface it uses on each poll.

peer_polling_protocol was included for the case that the Traffic Monitor is deployed in an ipv6 only environment.

- [x] This PR fixes #2735 

- Documentation
- Traffic Monitor

Upgrade TM in a system with servers that have both IPv4 and IPv6 addresses.  Verify that the new IPv4 and IPv6 columns are added to the TM UI and are showing the correct boolean for availability for IPv4 (since IPv4 only is the default) and that it is polling successfully.  Update the TM config file to include 
"cache_polling_protocol": "both",
"peer_polling_protocol": "both"

Restart TM with the new config.  Verify that it is successfully polling availability information for both IPv4 and IPv6.  Verify that servers with IPv4 are still successfully polled.  Verify that any issues are being reported correctly for both IPv4 and IPv6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Issue #3871 - provides users with a specified role the ability to mark any delivery service request as complete.
 - Fixed Traffic Ops Golang POST servers/id/deliveryservice continuing erroneously after a database error.
 - Fixed Traffic Ops Golang POST servers/id/deliveryservice double-logging errors.
+- Modified Traffic Monitor to poll over IPv6 as well as IPv4.
 
 ### Deprecated/Removed
 - The TO API `cachegroup_fallbacks` endpoint is now deprecated
--- a/docs/source/admin/traffic_monitor.rst
+++ b/docs/source/admin/traffic_monitor.rst
@@ -50,8 +50,15 @@ Configuring Traffic Monitor
 
 Configuration Overview
 ----------------------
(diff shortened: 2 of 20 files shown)


rob05c:
This is great! We've needed v6 polling for a while now.

I just have a few questions. I see `PeerPollingProtocol` in the config, but it doesn't appear to be used?

Running the Monitor, I'm also seeing a lot of `Peer is reachable` in the event log. It seems like it's logging every poll (when it should only log if the state changes). I could be misreading it, though. I'm not immediately seeing what's causing it in the code. Any ideas?

Also, just to verify, `both` polls both, and marks unavailable if either is Unavailable? I.e. if a v4 poll fails, then a v6 poll succeeds, it will still be Unavailable?

mattjackson220:
Updated with the PeerPollingProtocol
I'm digging through to see why there are so many "Peer is reachable" lines.  I would expect there to be double from before at TM startup since it will poll over v4 and v6.

Yes, both will poll over both v4 and v6 and mark it Unavailable if either is Unavailable.  It also will poll caches that are just v4 (or soon just v6) and mark those according to the result from the only address given (the previous PR would mark those as Unavailable since one was "Unavailable" just because it didnt have that address)

mattjackson220:
Closing because PR #4142 should contain all of these updates plus further updates

PR #4013, opened 2019-10-22
Add TO-Go /api/1.1/servers/status GET handler

### LINKED ISSUE #3826: Rewrite /servers/status to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/servers/status` is handled by Perl

## Expected / new behavior:
`/servers/status` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`

N/A

rawlinp:
Rewrite the /api/1.1/servers/status GET handler to Go.

- [x] This PR fixes #3826 

- Traffic Ops

1. Run to the TO API tests, make sure they pass.
2. Build the docs, make sure its successful, and view the output to make sure the new table looks good.


traffic_ops/traffic_ops_golang/server/status_count.go
```diff
@@ -0,0 +1,70 @@
...
+	api.WriteResp(w, r, statusCounts)
+}
+
+func getServersStatusCounts(tx *sql.Tx, typeName string) (map[string]int, error) {
+	q := `
+SELECT status.name, count(server.id)
+FROM server
+JOIN status ON server.status = status.id
+JOIN type ON server.type = type.id
+WHERE type.name ~ $1
```
ocket8888: Why use a regex on that?

rawlinp: to replicate the Perl version

ocket8888: I never knew that about Perl. You wanna document that? The `type` query parameter docs don't indicate that it supports POSIX regex.

rawlinp: Eh, I don't really want to document that. IMO it should really be `=`, but `~` makes it also count things like `EDGE_FOO` and `EDGE_BAR` if you pass `?type=EDGE`.

rob05c: Hm, regexes can be constructed to both be extremely CPU and Memory intensive. This will be a security vulnerability, if it's exposed with Self Service.

Since it's never been documented, what about changing it to `=`, or if you think people are using it for prefixes, maybe `LIKE '%' + $1` or `LIKE '%' + $1 + '%'`?

ocket8888: There's no input restriction on Type names, they can legally include backslashes. `strings.Replace` isn't powerful enough to guard against arbitrary bad Type names, I think.

rob05c: Escaping `%` and `_` was more to prevent accidents than malice. LIKE isn't overly expensive, it's not really a security risk for someone to do their own `%` if they really want to.

Though we could prevent that with an additional `strings.Replace(i, "\", "\\")` if we wanted.

ocket8888: what if I did `\%`? -> `\\%` -> `\\\\%` 

but if it's not a problem it's not a problem. I think `LIKE` is a better choice than parsing arbitrary regular expressions.

Also, if parsing user-input regexes can be considered a security risk then Content Invalidation Jobs are a huge hole.

rob05c: >what if I did \%? -> \\% -> \\\\%

The \ replace needs to come before the % replace, so `\%` becomes `\\\%`, which will be parsed by Postgres as a literal \ and a literal %.

> Also, if parsing user-input regexes can be considered a security risk then Content Invalidation Jobs are a huge hole.

Indeed. :(

rawlinp: If we're going to depart from the original Perl behavior, then I'll just make it `=`.

ocket8888: that sounds fine to me

PR #4015, opened 2019-10-23
Rewrite /federations to Go - POST/PUT/DELETE

### LINKED ISSUE #3790: Rewrite /federations to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/federations` is handled by Perl

## Expected / new behavior:
`/federations` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `DELETE`
- [x] `GET` (#2862)
- [ ] `POST`
- [ ] `PUT`

N/A

ocket8888:
- [x] This PR fixes #3790 

Rewrites `/federations` to Go. This also adds API version 1.4 specific logic that
- Adds the messages normally returned as the "response" object as alerts (doesn't replace `response`, much as I'd like to)
- Allows POST request bodies that don't use the redundant top-level "federations" key (still backward-compatible; just gives clients the choice)

- Documentation
- Traffic Control Client (Go)
- Traffic Ops

Run the unit tests and API/Go client integration tests. 

### Test Manually (CDN-in-a-Box)
1. Build and start CDN-in-a-Box
2. In Traffic Portal, go to <kbd>CDNs</kbd>&#8594;<kbd>CDN-in-a-Box</kbd>&#8594;<kbd>More</kbd>&#8594;<kbd>Manage Federations</kbd>
3. Click on the "+" button
4. Fill out the form - make sure to assign the user to `admin` (since I assume that's the user as whom you've logged in) and hit "Submit" to create a new Federation
5. Make a `POST` request to `/federations`, then a `PUT` then a `DELETE` (afaik there's no mechanism for this in Traffic Portal, so use `curl` or the `toaccess` scripts or something).

### Test Manually (Locally-run servers)
1. Build and start Traffic Ops
2. Assuming you already have a CDN, a Delivery Service, and a user as whom to sign in, make a new Federation using the `POST` method of `/cdns/{{your CDN name here}}/federations`.
3. Assign your federation to a Delivery Service (within your user's Tenant - I don't know if this will check) using the `POST` method of `/federations/{{ID of the federation you just created - probably 1}}/deliveryservices`
4. Assign the Federation to your user using the `POST` method of `/federations/{{same ID again}}/users`
5. Make a `POST` request to `/federations`, then a `PUT` then a `DELETE`

- [x] This PR includes tests

### DIFF
--- a/docs/source/api/federations.rst
+++ b/docs/source/api/federations.rst
@@ -21,7 +21,7 @@
 
 ``GET``
 =======
-Retrieves a list of federation mappings (aka federation resolvers) for the current user.
+Retrieves a list of :term:`Federation` mappings (i.e. :term:`Federation` Resolvers) for the current user.
 
 :Auth. Required: Yes
 :Roles Required: "admin", "Federation", "operations", "Portal", or "Steering"
@@ -34,12 +34,12 @@ No parameters available.
 Response Structure
 ------------------
 :deliveryService: The ``xml_id`` that uniquely identifies the :term:`Delivery Service` that uses the federation mappings in ``mappings``
-:mappings:        An array of objects that represent the mapping of a federation's Canonical Name (CNAME) to one or more resolvers
+:mappings:        An array of objects that represent the mapping of a :term:`Federation`'s :abbr:`CNAME (Canonical Name)` to one or more Resolvers
 
-	:cname:    The actual CNAME used by the federation
-	:resolve4: An array of IPv4 addresses capable of resolving the federation's CNAME
-	:resolve6: An array of IPv6 addresses capable of resolving the federation's CNAME
-	:ttl:      The Time To Live (TTL) of the CNAME in hours
+	:cname:    The actual CNAME used by the :term:`Federation`
+	:resolve4: An array of IPv4 addresses (or subnets in :abbr:`CIDR (Classless Inter-Domain Routing)` notation) capable of resolving the :term:`Federation`'s CNAME
+	:resolve6: An array of IPv6 addresses (or subnets in :abbr:`CIDR (Classless Inter-Domain Routing)` notation) capable of resolving the :term:`Federation`'s CNAME
+	:ttl:      The :abbr:`TTL (Time To Live)` of the CNAME in hours
 
 .. code-block:: http
 	:caption: Response Example
@@ -77,9 +77,9 @@ Response Structure
 
 ``POST``
 ========
-Allows a user to create federation resolvers for :term:`Delivery Service`\ s, providing the :term:`Delivery Service` is within a CDN that has some associated federation.
+Allows a user to create :term:`Federation` Resolvers for :term:`Delivery Services`, providing the :term:`Delivery Service` is within a CDN that has some associated :term:`Federation`.
 
-.. warning:: Confusingly, this endpoint does **not** create a new federation; to do that, the :ref:`to-api-cdns-name-federations` endpoint must be used. Furthermore, the federation must properly be assigned to a :term:`Delivery Service` using the :ref:`to-api-federations-id-deliveryservices` and assigned to the user creating resolvers using :ref:`to-api-federations-id-users`.
+.. warning:: Confusingly, this method of this endpoint does **not** create a new :term:`Federation`; to do that, the :ref:`to-api-cdns-name-federations` endpoint must be used. Furthermore, the :term:`Federation` must properly be assigned to a :term:`Delivery Service` using the :ref:`to-api-federations-id-deliveryservices` and assigned to the user creating Resolvers using :ref:`to-api-federations-id-users`.
 
 .. seealso:: The :ref:`to-api-federations-id-federation_resolvers` endpoint duplicates this functionality.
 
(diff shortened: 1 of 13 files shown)


traffic_ops/traffic_ops_golang/federations/federations.go
```diff
@@ -167,3 +206,296 @@ ORDER BY
...
+// adds federation resolvers of a specific type to the given federation
+func addFederationResolver(res []string, t tc.FederationResolverType, fedID uint, tx *sql.Tx) ([]string, error) {
+	inserted := []string{}
+	for _, r := range res {
+		var ip string
+		var id uint
+		if err := tx.QueryRow(insertResolverQuery, r, t).Scan(&ip, &id); err != nil && err != sql.ErrNoRows {
+			return nil, err
+		}
+		if ip != "" && id > 0 {
```
mhoppa: what would it mean if the ip is equal to "" and the id less than zero? Would that ever happen do we need to check for it?

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -86,7 +86,7 @@ func WriteRespRaw(w http.ResponseWriter, r *http.Request, v interface{}) {
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(bts)
+	w.Write(append(bts, '\n'))
```
mhoppa: I believe this is causing a unit test to fail in this package ->

```
--- FAIL: TestReadHandler (0.00s)
    /Users/mhoppa509/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api/shared_handlers_test.go:187: Expected body {"response":[{"ID":1}]} got {"response":[{"ID":1}]}
```

lib/go-tc/federation.go
```diff
@@ -67,36 +80,85 @@ type FederationMapping struct {
...
-	TTL      *int     `json:"ttl"`
-	CName    *string  `json:"cname"`
+type ResolverMapping struct {
 	Resolve4 []string `json:"resolve4,omitempty"`
 	Resolve6 []string `json:"resolve6,omitempty"`
 }
 
+func (r *ResolverMapping) Validate(tx *sql.Tx) error {
+	errs := []error{}
+	for _, res := range r.Resolve4 {
```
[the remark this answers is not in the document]
ocket8888: They do but it doesn't cover CIDRs

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -42,6 +42,75 @@ const BaseOrderBy = "\nORDER BY"
 const BaseLimit = "\nLIMIT"
 const BaseOffset = "\nOFFSET"
 
+const UserIDHasAccessToDeliveryServiceXMLIDQuery = `
```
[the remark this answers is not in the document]
ocket8888: You know it might not. I think that's a remnant from the old way I wanted to do it, but I couldn't figure out the query for it (that one doesn't work). I'll remove it.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -346,3 +430,17 @@ func GetCacheGroupNameFromID(tx *sql.Tx, id int64) (tc.CacheGroupName, bool, err
 	}
 	return tc.CacheGroupName(name), true, nil
 }
+
+// GetFederationIDForUserIDByXMLID retrieves the ID of the Federation assigned to the user defined by
+// userID on the Delivery Service identified by xmlid. If no such federation exists, the boolean
+// returned will be 'false', while the error indicates unexpected errors that occurred when querying.
+func GetFederationIDForUserIDByXMLID(tx *sql.Tx, userID int, xmlid string) (uint, bool, error) {
+	var id uint
```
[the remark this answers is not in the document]
ocket8888: because IDs can't be negative, and that's about it.

traffic_ops/traffic_ops_golang/federations/federations.go
ocket8888: Yep, those are the zero values of those two variables. The `ON CONFLICT DO NOTHING` means that rows that already exist aren't inserted, and thus return nothing. For whatever reason that's not a scan error and just leaves the variables unchanged. AFAIK there's no db restriction on the ip string length, but an ID can't be zero because the series generators explicitly start at one. So an ip could be blank (which means something very odd happened but ostensibly nothing changed) or the id could be zero, either of which means that nothing changed.

traffic_ops/traffic_ops_golang/api/api.go
ocket8888: whoops :P

traffic_ops/traffic_ops_golang/federations/federations.go
```diff
@@ -167,3 +206,296 @@ ORDER BY
...
+		api.HandleErr(w, r, tx, errCode, userErr, nil)
+		return
+	}
+
+	ipList := fmt.Sprintf("[ %s ]", strings.Join(ips, ", "))
+	msg := fmt.Sprintf("%s successfully deleted all federation resolvers: %s", inf.User.UserName, ipList)
+	changelogMsg := fmt.Sprintf("USER: %s, ID: %d, ACTION: %s", inf.User.UserName, inf.User.ID, msg)
+	api.CreateChangeLogRawTx(api.ApiChange, changelogMsg, inf.User, tx)
+
+	if inf.Version.Minor <= 3 {
```
rawlinp: ditto the previous comment

traffic_ops/traffic_ops_golang/federations/federations.go
```diff
@@ -167,3 +206,296 @@ ORDER BY
...
+	}
+
+	userErr, sysErr, errCode = addFederationResolverMappingsForCurrentUser(inf.User, tx, mappings)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+
+	createdMsg := fmt.Sprintf("%s successfully created federation resolvers.", inf.User.UserName)
+	if inf.Version.Minor <= 3 {
```
rawlinp: ditto previous comment

traffic_ops/traffic_ops_golang/federations/federations.go
```diff
@@ -167,3 +206,296 @@ ORDER BY
...
+func getMappingsFromRequestBody(v api.Version, body io.ReadCloser) (tc.DeliveryServiceFederationResolverMappingRequest, error, error) {
+	defer body.Close()
+	var mappings tc.DeliveryServiceFederationResolverMappingRequest
+
+	b, err := ioutil.ReadAll(body)
+	if err != nil {
+		return mappings, errors.New("Couldn't read request"), fmt.Errorf("Reading request body: %v", err)
+	}
+
+	if v.Minor <= 3 {
```
rawlinp: ditto previous comment about major version

docs/source/api/federations.rst
```diff
@@ -150,82 +173,122 @@ Request Structure
...
+	Access-Control-Allow-Origin: *
+	Set-Cookie: mojolicious=...; Path=/; HttpOnly
+	Whole-Content-Sha512: dQ5AvQULhc254zQwgUpBl1/CHbLr/clKtkbs0Ju9f1BM4xIfbbO3puFNN9zaEaZ1iz0lBvHFp/PgfUqisD3QHA==
+	X-Server-Name: traffic_ops_golang/
+	Date: Wed, 23 Oct 2019 23:22:03 GMT
+	Content-Length: 258
+	Content-Type: application/json
+
+	{ "alerts": {
+		"alerts": [
```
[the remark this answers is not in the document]
ocket8888: it shouldn't - I'll check.

traffic_ops/app/conf/cdn.conf
```diff
@@ -1,7 +1,7 @@
 {
     "hypnotoad" : {
         "listen" : [
-            "https://[::]:60443?cert=/etc/pki/tls/certs/localhost.crt&key=/etc/pki/tls/private/localhost.key&verify=0x00&ciphers=AES128-GCM-SHA256:HIGH:!RC4:!MD5:!aNULL:!EDH:!ED"
+            "https://[::]:60443?cert=/home/bwilli415/src/localhost.crt&key=/home/bwilli415/src/localhost.key&verify=0x00&ciphers=AES128-GCM-SHA256:HIGH:!RC4:!MD5:!aNULL:!EDH:!ED"
```
rawlinp: I think you may have accidentally committed these files

ocket8888: whoops. Must've popped that for running tests - should look for others like that

ocket8888: Okay, I think I fixed everything (and removed it from the change history too)

PR #4026, opened 2019-10-24
Feature/add deliveryservice required capability

jheitz200:
## What does this PR (Pull Request) do?

- [x] This PR is not related to any Issue 

Add deliveryservice_required_capabilities support to the API.
- Add the ability to GET, POST, DELETE an association between a delivery service and server capability.
- Includes tests, API handlers, migrations

## Which Traffic Control components are affected by this PR?

- Documentation - being added by another developer (#4018)
- Traffic Control Client - Go
- Traffic Ops

## What is the best way to verify this PR?

Apply migrations to db and seed server_capabilities

Build new traffic ops API and then associate/get/disassociates the newly seeded server_capability to deliveryservice via the new api path /deliveryservice_server_capabilities

## The following criteria are ALL met by this PR

- [X] This PR includes tests OR I have explained why tests are unnecessary
- [X] This PR includes documentation OR I have explained why documentation is unnecessary
- [X] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [X] This PR includes any and all required license headers
- [X] This PR ensures that database migration sequence is correct OR this PR does not include a database migration
- [X] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #4027, opened 2019-10-25
Fix endpoint name in docs

### LINKED ISSUE #4022: deliveryservice_server_capabilities Doc requires to be fixed with endpoint info
## I'm submitting a ...
- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [X] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
[The documentation](https://traffic-control-cdn.readthedocs.io/en/latest/api/deliveryservice_server_capabilities.html) refers to `deliveryservices_required_capabilities` as deliveryservice_server_capabilities

## Expected / new behavior:

deliveryservice_server_capabilities should be `deliveryservices_required_capabilities`

## Anything else:

Response Type for Delete is  undefined. Should it be object?

ChrisHines:
- [x] This PR fixes #4022

- Documentation

Build the docs and check that the API docs for `deliveryservice_server_capabilities` are renamed to `deliveryservices_required_capabilities`.

- master (426881af)

## Additional Information
Related endpoint implementation in #4026.

### DIFF
--- a/docs/source/api/deliveryservices_required_capabilities.rst
+++ b/docs/source/api/deliveryservices_required_capabilities.rst
@@ -13,11 +13,11 @@
 .. limitations under the License.
 ..
 
-.. _to-api-deliveryservice-server-capabilities:
+.. _to-api-deliveryservice-required-capabilities:
 
-***************************************
-``deliveryservice_server_capabilities``
-***************************************
+******************************************
+``deliveryservices_required_capabilities``
+******************************************
 
 .. versionadded:: 1.4
 
@@ -59,7 +59,7 @@ Request Structure
 .. code-block:: http
 	:caption: Request Example
 
-	GET /api/1.4/deliveryservice_server_capabilities HTTP/1.1
+	GET /api/1.4/deliveryservices_required_capabilities HTTP/1.1
 	Host: trafficops.infra.ciab.test
 	User-Agent: curl/7.47.0
 	Accept: */*
@@ -120,7 +120,7 @@ Request Structure
 .. code-block:: http
 	:caption: Request Example
 
-	POST /api/1.4/deliveryservice_server_capabilities HTTP/1.1
+	POST /api/1.4/deliveryservices_required_capabilities HTTP/1.1
 	Host: trafficops.infra.ciab.test
 	User-Agent: curl/7.47.0
 	Accept: */*
@@ -184,7 +184,7 @@ Request Structure
 .. code-block:: http
 	:caption: Request Example
 
-	POST /api/1.4/deliveryservice_server_capabilities HTTP/1.1
+	POST /api/1.4/deliveryservices_required_capabilities HTTP/1.1
 	Host: trafficops.infra.ciab.test
 	User-Agent: curl/7.47.0
 	Accept: */*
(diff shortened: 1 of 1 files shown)


docs/source/api/deliveryservices_required_capabilities.rst
```diff
@@ -13,11 +13,11 @@
 .. limitations under the License.
 ..
 
-.. _to-api-deliveryservice-server-capabilities:
+.. _to-api-deliveryservice-required-capabilities:
```
mhoppa: should this be deliveryservices-required-capabilites? ( missing s on end of deliveryservice)

ChrisHines: I guess so. What does this line get used for in the docs anyway?

mhoppa: I believe it is used to reference from other pages

ocket8888: yes, that's a reference label. You can, in practice put them anywhere to act as labels ("anchors" in HTML) but truly "label-able" things let you do
```rst
.. _section:

*************
SECTION TITLE
*************
Some paragraph text

.. _some-thing:

lorem ipsum dolor etc.

.. _figure:
.. figure:: jay-peg.png
    :caption: Figure caption

.. _table:
.. table:: My table

    A | B
    -+-
    1| 0

:ref:`section` (renders as "SECTION TITLE")
:ref:`figure` (renders as "Figure caption")
:ref:`table` (renders as "My table")
:ref:`some-thing` (this is an error - no title)
:ref:`Some thing <some-thing>` (Renders as a link to some-thing with the text "Some thing")
```

ChrisHines:
> I think maybe you also wanna change the success alert in the POST handler. It currently says "deliveryservice server_capability was updated."

If I'm reading ["How to Read this Documentaion"](https://traffic-control-cdn.readthedocs.io/en/latest/api/index.html#response-structure) right the "alerts" should be removed from the response structure example.

ocket8888:
Nah, that's just about the field lists that describe the structure:

> _"The field definitions listed in the Response Structure subsection of an endpoint method are the elements of this object"_

Examples should be complete requests and responses to that request. You just don't need to lay out "this endpoint returns an alerts array with one alert that has these properties..." because that's repetitive, and unnecessary. Honestly, it doesn't even really matter if the alerts shown aren't accurate, but the specific wording in this case could make it seem like that response is for a different endpoint.

ChrisHines:
Correct me if this is wrong. As I read the code for #4026, it uses [`api.CreateHandler`](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/api/shared_handlers.go#L310) to implement the POST request, which will format the alert message as `"deliveryservice.RequiredCapability was created."`

The alert in the DELETE request needed a similar fix, plus updating of new-line placement to be consistent with the other examples.

PR #4029, opened 2019-10-25
Rewrite cache_stats to Golang

### LINKED ISSUE #3847: Rewrite /cache_stats to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/cache_stats` is handled by Perl

## Expected / new behavior:
`/cache_stats` should be handled by Go.

### Minimum API Version 
1.2

### Request Methods
- [ ] `GET`

N/A

mhoppa:

- [x] This PR fixes #3847

This PR rewrites the `cache_stats` endpoint from Perl to Go. It does inherit/leverage a lot of work done apart of #3758.

As part of the rewrite there is some differences between perl implementation and golang implementation including:

* Checking of valid metricType in query parameter
* Omitting of `query` key in response as it was undocumented and not providing data back
* The interval query parameter is now restricted to strings matching ^\d+[mhdw]$ (inherited from #3758)
* The startDate and endDate query parameters now accept their inputs strictly in RFC3339 format (with optional sub-second precision), as a number of nanoseconds since the Unix Epoch, or in the special tc.TimeLayout format. (inherited from #3758)
* Timestamps in data series as output can now be in one of two formats, as specified by the client. By default, the output is in RFC3339 format (without sub-second precision), but optionally they can be presented as the number of nanoseconds since the Unix Epoch. (inherited from #3758)

- Documentation
- Traffic Ops

Since there is not currently the ability to write api tests for endpoints that rely on influxdb the best way to verify the PR is to pull down PR, build Traffic Ops and ensure the cache_stats endpoint has the same behavior as before.


traffic_ops/traffic_ops_golang/trafficstats/cache.go
```diff
@@ -0,0 +1,207 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
+	influx "github.com/influxdata/influxdb/client/v2"
```
ocket8888: Just want to be sure because idk when you started this rewrite, but this is the "vendored" client and not the one that actually exists on the Internet, right?

mhoppa: yep!

PR #4030, opened 2019-10-25
Rewrite PUT /api/1.1/servers/:id/status to Go

### LINKED ISSUE #3829: Rewrite /servers/{{id}}/status to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/servers/{{id}}/status` is handled by Perl

## Expected / new behavior:
`/servers/{{id}}/status` should be handled by Go.

### Minimum API Versions
1.1

### Request Methods
- [ ] `PUT`

N/A

rawlinp:
Rewrite `PUT /api/1.1/servers/:id/status Also, fix bug where updates weren't being queued on children via secondary_parent_cachegroup_id.

- [x] This PR fixes #3829 

- Traffic Control Client (Go)
- Traffic Ops
- testing

1. `go test ./traffic_ops/traffic_ops_golang/...`
2. `go test ./lib/go-util/...`
3. Run the TO API tests

- [x] API docs already exist

### DIFF
--- a/traffic_ops/testing/api/v14/serverupdatestatus_test.go
+++ b/traffic_ops/testing/api/v14/serverupdatestatus_test.go
@@ -104,7 +104,7 @@ func UpdateTestServerStatus(t *testing.T) {
 	// update status of MID server to OFFLINE via ID
 	status, _, err := TOSession.GetStatusByName("OFFLINE")
 	if err != nil {
-		t.Errorf("cannot GET status by name: %v", err)
+		t.Fatalf("cannot GET status by name: %v", err)
 	}
 	_, _, err = TOSession.UpdateServerStatus(mid1cdn1.ID, tc.ServerPutStatus{
 		Status:        util.JSONNameOrIDStr{ID: util.IntPtr(status[0].ID)},
--- a/traffic_ops/testing/api/v14/serverupdatestatus_test.go
+++ b/traffic_ops/testing/api/v14/serverupdatestatus_test.go
@@ -1,0 +1,149 @@
+package v14
+
+/*
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


traffic_ops/testing/api/v14/serverupdatestatus_test.go
```diff
@@ -0,0 +1,149 @@
...
+		t.Errorf("expected: server %s with updated status to have no updates pending, actual: updates pending", mid1cdn1.HostName)
+	}
+	if edge1cdn2.UpdPending {
+		t.Errorf("expected: server %s in different CDN than server with updated status to have no updates pending, actual: updates pending", edge2cdn1.HostName)
+	}
+
+	// update status of MID server to OFFLINE via ID
+	status, _, err := TOSession.GetStatusByName("OFFLINE")
+	if err != nil {
+		t.Errorf("cannot GET status by name: %v", err)
```
ocket8888: error should be fatal, as when an error occurs `status` could be `nil` (in fact the way the method is written I think it has to be)

traffic_ops/traffic_ops_golang/server/put_status.go
```diff
@@ -0,0 +1,137 @@
...
+		return
+	}
+	offlineReason := ""
+	if reqObj.OfflineReason != nil {
+		offlineReason = *reqObj.OfflineReason
+	}
+	msg := "Updated status [ " + *status.Name + " ] for " + serverInfo.HostName + "." + serverInfo.DomainName + " [ " + offlineReason + " ]"
+
+	// queue updates on child servers if server is ^EDGE or ^MID
+	if strings.HasPrefix(serverInfo.Type, tc.CacheTypeEdge.String()) || strings.HasPrefix(serverInfo.Type, tc.CacheTypeMid.String()) {
```
ocket8888: Is this how Perl checked, with a strict prefix? I thought it was more like `/.*(MID|EDGE).*/` but I'll take your word for it.

rawlinp: Yes, but you don't have to take my word for it: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/API/Server.pm#L880

PR #4031, opened 2019-10-28
Fix query parameter name

ChrisHines:
- [x] This PR updates documentation to match implementation in #4026

- Documentation

Build the docs and check that the query parameters are correct on the API page for the deliveryservices_required_capabilities endpoint.


ChrisHines:
@mhoppa PTAL

PR #4041, opened 2019-10-29
Block server servercapability delete if associated ds requires it

mhoppa:

This PR adds in extra validation in disassociating a server capability from a server. If that server is assigned to a DS that requires that server capability we now block the DELETE.

- Documentation
- Traffic Ops

Besides running API tests it can be manually tested by:
* Create DS
* Create server capability "ram" 
* Assign server capability to DS
* Create edge server 
* Assign server capability "ram"  to server
* Assign server to DS
* Attempt to DELETE server capability from server (SHOULD NOW BE BLOCKED)
* Remove  server capability "ram" from DS
* Attempt to DELETE server capability from server (SHOULD NOW WORK)

- [x] This PR includes tests 
- [x] This PR does not included an update to CHANGELOG.md

### DIFF
--- a/traffic_ops/testing/api/v14/cachegroupsdeliveryservices_test.go
+++ b/traffic_ops/testing/api/v14/cachegroupsdeliveryservices_test.go
@@ -19,11 +19,8 @@ import (
 	"testing"
 )
 
-func TestDeliveryServicesCachegroups(t *testing.T) {
-	WithObjs(t, []TCObj{CDNs, Types, Tenants, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, DeliveryServices}, func() {
-		CreateTestCachegroupsDeliveryServices(t)
-		DeleteTestCachegroupsDeliveryServices(t)
-	})
+func TestCacheGroupsDeliveryServices(t *testing.T) {
+	WithObjs(t, []TCObj{CDNs, Types, Tenants, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, DeliveryServices, CacheGroupsDeliveryServices}, func() {})
 }
 
 const TestEdgeServerCacheGroupName = "cachegroup1" // TODO this is the name hard-coded in the create servers test; change to be dynamic
--- a/traffic_ops/traffic_ops_golang/server/servers_server_capability.go
+++ b/traffic_ops/traffic_ops_golang/server/servers_server_capability.go
@@ -120,12 +120,10 @@ func (ssc *TOServerServerCapability) Read() ([]interface{}, error, error, int) {
 }
 
 func (ssc *TOServerServerCapability) Delete() (error, error, int) {
-	tx := ssc.APIInfo().Tx
-
 	// Ensure that the user is not removing a server capability from the server
 	// that is required by the delivery services the server is assigned to (if applicable)
 	dsIDs := []int64{}
-	if err := tx.QueryRow(checkDSReqCapQuery(), ssc.ServerID, ssc.ServerCapability).Scan(pq.Array(&dsIDs)); err != nil {
+	if err := ssc.APIInfo().Tx.QueryRow(checkDSReqCapQuery(), ssc.ServerID, ssc.ServerCapability).Scan(pq.Array(&dsIDs)); err != nil {
 		return nil, fmt.Errorf("checking removing server server capability would still suffice delivery service requried capabilites: %v", err), http.StatusInternalServerError
 	}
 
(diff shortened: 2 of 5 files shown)


traffic_ops/traffic_ops_golang/server/servers_server_capability.go
```diff
@@ -117,6 +120,22 @@ func (ssc *TOServerServerCapability) Read() ([]interface{}, error, error, int) {
...
+	if err := ssc.APIInfo().Tx.QueryRow(checkDSReqCapQuery(), ssc.ServerID, ssc.ServerCapability).Scan(pq.Array(&dsIDs)); err != nil {
+		return nil, fmt.Errorf("checking removing server server capability would still suffice delivery service requried capabilites: %v", err), http.StatusInternalServerError
+	}
+
+	if len(dsIDs) > 0 {
+		dsIdsStr, err := json.Marshal(dsIDs)
+		if err != nil {
+			return nil, fmt.Errorf("formatting response message on bad request to disassociate server capability from server: %v", err), http.StatusInternalServerError
+		}
+		return fmt.Errorf("cannot remove the capability %v from the server %v as the server is assigned to the delivery services %v that require it", *ssc.ServerCapability, *ssc.ServerID, string(dsIdsStr)), nil, http.StatusBadRequest
```
ocket8888: This check isn't respecting tenancy; if a user doesn't have tenant permissions over a DS they're not supposed to know it exists. In that case, you should probably give back an Internal Server Error with no user error, I think, to maintain obscurity.

traffic_ops/testing/api/v14/cachegroupsdeliveryservices_test.go
```diff
@@ -19,11 +19,8 @@ import (
 	"testing"
 )
 
-func TestDeliveryServicesCachegroups(t *testing.T) {
-	WithObjs(t, []TCObj{CDNs, Types, Tenants, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, DeliveryServices}, func() {
-		CreateTestCachegroupsDeliveryServices(t)
-		DeleteTestCachegroupsDeliveryServices(t)
-	})
+func TestCacheGroupsDeliveryServices(t *testing.T) {
+	WithObjs(t, []TCObj{CDNs, Types, Tenants, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, DeliveryServices, CacheGroupsDeliveryServices}, func() {})
```
ocket8888: You appear to have disabled these unrelated tests; why?

mhoppa: no they are just included in CacheGroupsDeliveryService now

traffic_ops/traffic_ops_golang/server/servers_server_capability.go
mhoppa: based on the result of https://github.com/apache/trafficcontrol/issues/4043 I am unsure if we would want to change this to a 500 or leave it as is

rob05c: This is a user error. 5xx codes are for server errors, it's a violation of HTTP to send them for a client error.

Like I commented in #4043, I think this should be a `403 Forbidden` since we can't really hide DS existence. But if we feel the need to try to obfuscate, a 400 or 404 would be acceptable; but 5xx isn't.

(404 is kind of a little lie; but we can kind of rationalize it as "it doesn't exist, as far as you're concerned." Which IMO is a much smaller lie/violation than a 5xx for a request error. But I'd vote for 400 over 404.)

ocket8888: So `api.GenericDelete` probably handles returning 404's for keys that aren't found, huh? I only suggested a 500 because otherwise it seemed like you'd get a 404 **only** when you don't have permission, which isn't an obfuscation at all. Definitely should be a 404 IMO as long as legitimate 404s are possible.

mhoppa: My first thought it to leave this as a 400 regardless of that decision ( which 40x code to use).

I am just checking that you can not remove a server capability from a server that is required from an associated ds.  Server server capabilities are not under tenancy so returning a 403 seems a little weird to me.

ocket8888: Yeah, 400 is fine.

mhoppa: I dont think this code needs to change then - as at the point of getting the affected DSes ids I  am going to return 400 regardless if they have tenancy or not.

One thing I could do is omit the ids from the alert message (either all or the ones they dont have permissions to)

ocket8888: I think the ids oughtta be excluded.

mhoppa: @ocket8888  I omit the ds IDs the user is not authorized to see from the error message. Can you please rereview the PR?

mitchell852:
what about the opposite scenario? blocking ds required capability addition IF assigned servers don't have that capability. i suppose that doesn't belong in this PR but just thought i'd bring it up.

PR #4063, opened 2019-11-01
Add atstccfg meta config

rob05c:

Adds meta config to atstccfg config generator.

Includes tests.
Includes changelog.

- Traffic Ops
- Traffic Ops ORT

Run unit tests.

Run `atstccfg` with request to the meta config endpoint, verify response is identical to Go and old Perl endpoints.

Not a bug fix.

### DIFF
--- a/lib/go-atscfg/meta.go
+++ b/lib/go-atscfg/meta.go
@@ -1,0 +1,175 @@
+package atscfg
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+)
+
+type ConfigProfileParams struct {
+	FileNameOnDisk string
+	Location       string
+	URL            string
+	APIURI         string
+}
+
+// APIVersion is the Traffic Ops API version for config fiels.
+// This is used to generate the meta config, which has API paths.
+// Note the version in the meta config is not used by the atstccfg generator, which isn't actually an API.
+// TODO change the config system to not use old API paths, and remove this.
+const APIVersion = "1.4"
+
+func MakeMetaConfig(
+	serverHostName tc.CacheName,
+	server *ServerInfo,
+	tmURL string, // global tm.url Parameter
+	tmReverseProxyURL string, // global tm.rev_proxy.url Parameter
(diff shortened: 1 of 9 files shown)


lib/go-atscfg/meta.go
```diff
@@ -0,0 +1,195 @@
...
+		} else {
+			scopeID := ""
+			if scope == tc.ATSConfigMetaDataConfigFileScopeCDNs {
+				scopeID = string(server.CDN)
+			} else if scope == tc.ATSConfigMetaDataConfigFileScopeProfiles {
+				scopeID = server.ProfileName
+			} else { // ATSConfigMetaDataConfigFileScopeServers
+				scopeID = server.HostName
+			}
+			atsCfg.APIURI = "/api/1.2/" + string(scope) + "/" + scopeID + "/configfiles/ats/" + cfgFile
```
alficles: Hardcoded 1.2?

lib/go-atscfg/meta.go
```diff
@@ -0,0 +1,195 @@
...
+	default:
+		return getScope(cfgFile, scopeParams)
+	}
+}
+
+const DefaultScope = tc.ATSConfigMetaDataConfigFileScopeServers
+
+// getScope returns the ATSConfigMetaDataConfigFileScope for the given config file, and potentially the given server. If the config is not a Server scope, i.e. was part of an endpoint which does not include a server name or id, the server may be nil.
+func getScope(cfgFile string, scopeParams map[string]string) tc.ATSConfigMetaDataConfigFileScope {
+	switch {
```
alficles: Is the fallthrough here better than multiple-case? Like: `cfgFile == "ip_allow.config", cfgFile == "parent.config", ...`

traffic_ops/traffic_ops_golang/ats/atsserver/meta.go
```diff
@@ -85,136 +70,14 @@ func GetConfigMetaData(w http.ResponseWriter, r *http.Request) {
...
-		scope = string(tc.ATSConfigMetaDataConfigFileScopeServers)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("GetConfigMetaData getting server uri-signed dses: "+err.Error()))
+		return
 	}
-	return tc.ATSConfigMetaDataConfigFileScope(scope), nil
+
+	txt := atscfg.MakeMetaConfig(serverName, server, tmParams.URL, tmParams.ReverseProxyURL, locationParams, uriSignedDSes, scopeParams)
+	w.Header().Set(tc.ContentType, tc.ApplicationJson)
+	w.Write([]byte(txt))
+	api.WriteRespRaw(w, r, txt)
```
rawlinp: I know `atstccfg` isn't using this API, but I'm pretty sure this line should be removed since we're calling `w.Write` just above it.

lib/go-atscfg/meta.go
rob05c: The config generator doesn't actually care what this number is. Unfortunately, the only place it's really defined is in the client, and we don't want this library to depend on that.
I changed it to a const.

lib/go-atscfg/meta.go
rob05c: Changed.

traffic_ops/traffic_ops_golang/ats/atsserver/meta.go
rob05c: Yeah, ORT/atstccfg doesn't use this anymore; but still a bug. Fixed.

rob05c:
>Also, I don't have a line to attach this comment to, but in routes.go I noticed the ATS meta config API route is duplicated (L436 and L469) if you'd like to remove one of those extra routes.

Fixed. Good catch! Guessing that was a non-conflict somewhere in all the merging around atstccfg PRs.

rob05c:
>I'm getting a TO API test failure on this:
Fixed. My fault for not running the API tests again after fixing Chris' version comment. And for making the API test brittle. I also fixed it to be less brittle.

rob05c:
Retest this please.

PR #4072, opened 2019-11-03
Prevent Assigning Servers to Delivery Services Without Required Capabilities

lactose:
This PR adds in extra validation in associating a server to a delivery service. If the user attempts to assign a server to a DS that requires a server capability the server does not have, an error will be returned. In addition, the endpoint /deliveryservices/{id}/servers/eligible will now only return those servers which meet the required capabilities (unless the DS does not have any).

- Documentation
- Traffic Ops

API Tests,

### Regression
*Case 1*:
- Create a delivery service
- Create a server, do not assign it the capability
- Assign server to DS (without required capabilities this should still pass)

*Case 2*:

- Create a delivery service
- Create a delivery service required capability for this delivery service
- Create a server, do not assign it the capability
- Check the /deliveryservice/{id}/servers/eligible endpoint. The server should be returned in the results.

### Feature
*Case 1*:
1. Create a delivery service
2. Create a delivery service required capability for this delivery service
3. Create a server, do not assign it the capability
4. Attempt to assign the server to the DS, and vice versa. Should receive an error.
5. Create a server server capability that matches the required capability
6. Attempt to assign the server to the DS, and vice versa. Should pass.

*Case 2*:
1. Create a delivery service
2. Create a delivery service required capability for this delivery service
3. Create a server, do not assign it the capability
4. Check the /deliveryservice/{id}/servers/eligible endpoint. The server should not be returned in the results.


traffic_ops/testing/api/v14/deliveryserviceservers_test.go
```diff
@@ -25,26 +28,86 @@ func TestDeliveryServiceServers(t *testing.T) {
...
+
+			_, _, err = TOSession.CreateDeliveryServicesRequiredCapability(ctc.capability)
+			if err != nil {
+				t.Errorf("*POST delivery service required capability: %v\n", err)
+			}
+
+			ctc.ssc.ServerID = &server.ID
+			_, _, err = TOSession.CreateServerServerCapability(ctc.ssc)
+			if err != nil {
+				t.Errorf("could not POST the server capability %v to server %v: %v\n", *ctc.ssc.ServerCapability, *ctc.ssc.Server, err)
```
mhoppa: nit but should this error and the above error be t.Fatalf

traffic_ops/testing/api/v14/deliveryserviceservers_test.go
```diff
@@ -85,3 +148,23 @@ func DeleteTestDeliveryServiceServers(t *testing.T) {
...
 	}
 }
+
+func getServersAndDSes(t *testing.T) ([]tc.DeliveryService, []tc.Server) {
+	dses, _, err := TOSession.GetDeliveryServices()
+	if err != nil {
+		t.Errorf("cannot GET DeliveryServices: %v\n", err)
+	}
+	if len(dses) < 1 {
+		t.Errorf("GET DeliveryServices returned no dses, must have at least 1 to test ds-servers")
```
mhoppa: fatal

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -274,6 +274,46 @@ func GetProfileIDFromName(name string, tx *sql.Tx) (int, bool, error) {
 	return id, true, nil
 }
 
+// GetServerCapabilitiesFromName returns the server's capabilities.
+func GetServerCapabilitiesFromName(name string, tx *sql.Tx) ([]string, error) {
+	caps := []string{}
```
mhoppa: this is just a clean up comment but in this function and the function below the code can be cleaned up by using the array type instead of iterating through rows.

```SELECT ARRAY(....)``` and then scanning that into pq.Array. There is a good amount examples of the pattern around the code base.

File: traffic_ops/traffic_ops_golang/deliveryservice/eligible.go [hunk 4]
```diff
@@ -169,11 +173,22 @@ AND (t.name LIKE 'EDGE%' OR t.name LIKE 'ORG%')
 			&s.Type,
 			&s.TypeID,
 			&s.UpdPending,
+			pq.Array(&s.ServerCapabilities),
```
mhoppa: like here :)

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
```diff
@@ -425,6 +446,30 @@ func GetCreateHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	for _, name := range serverNames {
+		sCaps, err = dbhelpers.GetServerCapabilitiesFromName(name, tx)
+		if err != nil {
+			return err
+		}
+		for _, dsc := range dsCaps {
+			if !util.ContainsStr(sCaps, dsc) {
+				return errors.New(fmt.Sprintf("Caching server cannot be assigned to this delivery service without having the required delivery service capabilities: [%v]", dsCaps))
```
mhoppa: nit should we include server host name in the error message?

traffic_ops/traffic_ops_golang/server/servers_assignment.go
```diff
@@ -81,6 +89,31 @@ func AssignDeliveryServicesToServerHandler(w http.ResponseWriter, r *http.Reques
 	api.WriteRespAlertObj(w, r, tc.SuccessLevel, "successfully assigned dses to server", tc.AssignedDsResponse{server, assignedDSes, replace})
 }
 
+// ValidateDSCapabilities checks that the server meets the requirements of each delivery service to be assigned.
+func ValidateDSCapabilities(dsIDs []int, serverName string, tx *sql.Tx) error {
+	fmt.Println("))) Validate was hit")
```
mhoppa: remove

traffic_ops/traffic_ops_golang/server/servers_assignment.go
```diff
@@ -81,6 +89,31 @@ func AssignDeliveryServicesToServerHandler(w http.ResponseWriter, r *http.Reques
...
+	}
+
+	for _, id := range dsIDs {
+		dsCaps, err = dbhelpers.GetDSRequiredCapabilitiesFromID(id, tx)
+		if err != nil {
+			return err
+		}
+		for _, dsc := range dsCaps {
+			if !util.ContainsStr(sCaps, dsc) {
+				return errors.New(fmt.Sprintf("Caching server cannot assign this delivery service without having the required delivery service capabilities: [%v]", dsCaps))
```
mhoppa: again include identifying ids to help the user

lib/go-tc/deliveryservice_servers.go
```diff
@@ -67,47 +67,49 @@ type DSServersAttrResponse struct {
...
+	Rack                        *string              `json:"rack" db:"rack"`
+	RouterHostName              *string              `json:"routerHostName" db:"router_host_name"`
+	RouterPortName              *string              `json:"routerPortName" db:"router_port_name"`
+	Status                      *string              `json:"status" db:"status"`
+	StatusID                    *int                 `json:"statusId" db:"status_id"`
+	TCPPort                     *int                 `json:"tcpPort" db:"tcp_port"`
+	Type                        string               `json:"type" db:"server_type"`
+	TypeID                      *int                 `json:"typeId" db:"server_type_id"`
+	UpdPending                  *bool                `json:"updPending" db:"upd_pending"`
+	ServerCapabilities          []string             `json:"serverCapabilities" db:"server_capabilities"`
```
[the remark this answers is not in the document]
mhoppa: actually lets omit these fields from the json output since we are not including server capabilities on servers/delivery_services

traffic_ops/traffic_ops_golang/server/servers_assignment.go
```diff
@@ -81,6 +89,31 @@ func AssignDeliveryServicesToServerHandler(w http.ResponseWriter, r *http.Reques
 	api.WriteRespAlertObj(w, r, tc.SuccessLevel, "successfully assigned dses to server", tc.AssignedDsResponse{server, assignedDSes, replace})
 }
 
+// ValidateDSCapabilities checks that the server meets the requirements of each delivery service to be assigned.
+func ValidateDSCapabilities(dsIDs []int, serverName string, tx *sql.Tx) error {
```
mhoppa: this need to differentiate between storage errors from `GetServerCapabilitiesFromName` and not meeting the delivery service req capabilities.

This is because the return back to the user differs. The first should be an internal server error and the latter needs to be bad request.

Right now getting internal server error on both.

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
```diff
@@ -425,6 +446,30 @@ func GetCreateHandler(w http.ResponseWriter, r *http.Request) {
 	api.WriteResp(w, r, tc.DeliveryServiceServers{payload.ServerNames, payload.XmlId})
 }
 
+// ValidateServerCapabilities checks that the delivery service's requirements are met by each server to be assigned.
+func ValidateServerCapabilities(dsID int, serverNames []string, tx *sql.Tx) error {
```
mhoppa: this need to differentiate between storage errors from `GetDSRequiredCapabilitiesFromID ` and not meeting the delivery service req capabilities.

This is because the return back to the user differs. The first should be an internal server error and the latter needs to be bad request.

Right now getting internal server error on both.

mitchell852:
Just so I'm clear. What are the API endpoints affected by this PR? I think:

GET /deliveryservices/{:id}/servers/eligible
POST /deliveryserviceserver

docs/source/api/deliveryservices_id_servers_eligible.rst
```diff
@@ -23,7 +23,11 @@
 
 ``GET``
 =======
-Retrieves properties of Edge-Tier servers eligible for assignment to a particular :term:`Delivery Service`.
+Retrieves properties of Edge-Tier servers eligible for assignment to a particular :term:`Delivery Service`. Eligibility is determined based on the following properties:
+
+1. Server Type: Edge, Origin
+2. CDN: Server, :term:`Delivery Service` must belong to the same CDN
+3. Server Capabilities: If the :term:`Delivery Service` has  :term:`Delivery Service required capabilities`, a server must have all defined capabilities
```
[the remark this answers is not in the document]
ocket8888: Unless you want to make it a term by adding it to the glossary! :smile:

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
```diff
@@ -447,7 +447,7 @@ func GetCreateHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // ValidateServerCapabilities checks that the delivery service's requirements are met by each server to be assigned.
-func ValidateServerCapabilities(dsID int, serverNames []string, tx *sql.Tx) (error, error) {
+func ValidateServerCapabilities(dsID int, serverNames []string, tx *sql.Tx) (error, int) {
```
mhoppa: So I think on this and the other Validation method you want to have the return of `(err, err, int)`

The function you are using in the caller `api.HandleErr` if you notice is taking in two errors and a status code. The reason being is if the error is a user error it returns it to the user in the response if it is a system error it just logs it as the user really doesn't need to know about it.

In this case on the db errors you would return a sysErr and on missing capabilities you would return a usrErr.

lactose:
> Just so I'm clear. What are the API endpoints affected by this PR? I think:
> 
> GET /deliveryservices/{:id}/servers/eligible
> POST /deliveryserviceserver

all of the ones mentioned here:
I believe new logic will need to be added to each endpoint:

> GET /deliveryservices/{:id}/servers/eligible (only return servers that have the capabilities required of the delivery service)
> POST /deliveryserviceserver (check that the servers being assigned to the delivery service have the needed capabilities)
> POST /servers/
> 
> {id}
> /deliveryservices (check that the delivery services being assigned to the server do not require a capability the server does not have)

PR #4074, opened 2019-11-04
Set Content-Length header in Traffic Router

### LINKED ISSUE #3965: Traffic Router inconsistent behavior on HTTP HEAD requests
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [x] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

The TR response to a HEAD request should resemble the same as a GET response. However, the behavior I'm seeing seems problematic. Following examples of each (from curl output) with timing info included:

**Proper GET response:**
```
> GET / HTTP/1.1
> Host: [REDACTED]
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 302 Found
< Location: [REDACTED]
< Content-Length: 0
< Date: Tue, 08 Oct 2019 14:58:45 GMT
<
* Connection #0 to host [REDACTED] left intact

real	0m0.705s
user	0m0.008s
sys	0m0.015s
```

**Improper HEAD response:**
```
> HEAD / HTTP/1.1
> Host: [REDACTED]
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 302 Found
< Location: [REDACTED]
< Transfer-Encoding: chunked
< Date: Tue, 08 Oct 2019 15:23:12 GMT
<
* transfer closed with outstanding read data remaining
* stopped the pause stream!
* Closing connection 0

real	0m10.855s
user	0m0.007s
sys	0m0.017s
```
Note that the HEAD response is including `Transfer-Encoding: chunked`, which is causing the client to wait for body data which never arrives. After ~10 seconds, Traffic Router times out the connection and resets the TCP connection (verified TR sends FIN at this point with a packet capture).

## Expected / new behavior:

TR should be responding to a HEAD request with `Content-Length: 0`, and **NOT** `Transfer-Encoding: chunked`.

Traffic Router version: 3.0.0-9930.f225e63b.el7

Problem can easily be replicated by sending a HEAD request with curl:
```
time curl -vs -o /dev/null -X HEAD http://ccr.invalid.cdn.example.com/
```

zrhoffman:

- This PR fixes #3965 by making the Traffic Router API and routing responses include a `Content-Length` header with every response
- Exposes LocationController endpoints to HEAD requests
- HEAD requests for routing responses return the same `Content-Type` and `Content-Length` as GET requests

- Documentation
- Traffic Router

Run the following external tests:
* org.apache.traffic_control.traffic_router.core.external.BufferedResponseTest
* org.apache.traffic_control.traffic_router.core.external.LocationsTest

- master (c65ee2eb43)
- 6.0.0

- [x] This PR includes tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -212,7 +212,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Updated `t3c-apply` to reduce mutable state in `TrafficOpsReq` struct.
 - Updated Golang dependencies
 - [#6506](https://github.com/apache/trafficcontrol/pull/6506) - Updated `jackson-databind` and `jackson-annotations` Traffic Router dependencies to version 2.13.1
-- Traffic Router now includes a `Content-Length` header in each response.
 
 ### Deprecated
 - Deprecated the endpoints and docs associated with `/api_capability` and `/capabilities`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - [#7158](https://github.com/apache/trafficcontrol/issues/7158) *Traffic Vault* Fix the `reencrypt` utility to uniquely reencrypt each version of the SSL Certificates.
 - [#7137](https://github.com/apache/trafficcontrol/pull/7137) *Cache Config* parent.config simulate topology for non topo delivery services.
 - Adds an extra T3C check for validity of an ssl cert (crash fix).
+- Traffic Router now always includes a `Content-Length` header in the response.
 
 ## [7.0.0] - 2022-07-19
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -211,6 +211,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Updated `t3c-apply` to reduce mutable state in `TrafficOpsReq` struct.
 - Updated Golang dependencies
 - [#6506](https://github.com/apache/trafficcontrol/pull/6506) - Updated `jackson-databind` and `jackson-annotations` Traffic Router dependencies to version 2.13.1
+- Traffic Router now includes a `Content-Length` header in each response.
 
 ### Deprecated
 - Deprecated the endpoints and docs associated with `/api_capability` and `/capabilities`.
--- a/docs/source/development/traffic_router/traffic_router_api.rst
+++ b/docs/source/development/traffic_router/traffic_router_api.rst
@@ -79,6 +79,7 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
+	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 1214
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -162,6 +163,7 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
+	Connection: close
 	Content-Disposition: inline;filename=f.txt
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 131
@@ -205,6 +207,7 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
+	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 35
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -235,6 +238,7 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
+	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 278
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -291,6 +295,7 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
+	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 253
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -418,6 +423,7 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
+	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 828
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -568,6 +574,7 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
+	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 137
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -609,6 +616,7 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
+	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 163
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
--- a/docs/source/development/traffic_router/traffic_router_api.rst
+++ b/docs/source/development/traffic_router/traffic_router_api.rst
@@ -79,7 +79,6 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
-	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 1214
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -163,7 +162,6 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
-	Connection: close
 	Content-Disposition: inline;filename=f.txt
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 131
@@ -207,7 +205,6 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
-	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 35
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -238,7 +235,6 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
-	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 278
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -295,7 +291,6 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
-	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 253
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -423,7 +418,6 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
-	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 828
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -574,7 +568,6 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
-	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 137
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
@@ -616,7 +609,6 @@ Response Structure
 	:caption: Response Example
 
 	HTTP/1.1 200 OK
-	Connection: close
 	Content-Type: application/json;charset=UTF-8
 	Content-Length: 163
 	Date: Mon, 04 Nov 2019 19:48:04 GMT
--- a/docs/source/development/traffic_router/traffic_router_api.rst
+++ b/docs/source/development/traffic_router/traffic_router_api.rst
@@ -70,24 +70,24 @@ Request Structure
 
 	GET /crs/stats HTTP/1.1
 	Host: trafficrouter.infra.ciab.test
-	User-Agent: curl/7.47.0
+	User-Agent: curl/7.52.1
 	Accept: */*
 
 Response Structure
 ------------------
 .. code-block:: http
-	:caption: Response Example
+	:caption: Response Example (JSON expanded)
 
-	HTTP/1.1 200 OK
+	HTTP/1.1 200
 	Content-Type: application/json;charset=UTF-8
-	Transfer-Encoding: chunked
-	Date: Tue, 15 Jan 2019 21:02:09 GMT
+	Content-Length: 1213
+	Date: Sun, 03 Nov 2019 05:57:17 GMT
 
 	{ "app": {
-		"buildTimestamp": "2019-01-10",
+		"buildTimestamp": "2019-11-03",
 		"name": "traffic_router",
 		"deploy-dir": "/opt/traffic_router",
-		"git-revision": "437e9df81",
+		"git-revision": "3ebc920b7",
 		"version": "3.0.0"
 	},
 	"stats": {
@@ -95,7 +95,7 @@ Response Structure
 		"httpMap": {
 			"video.demo1.mycdn.ciab.test": {
 				"czCount": 0,
-				"geoCount": 0,
+				"geoCount": 1,
 				"deepCzCount": 0,
 				"missCount": 0,
 				"dsrCount": 0,
@@ -107,29 +107,29 @@ Response Structure
 			}
 		},
 		"totalDnsCount": 0,
-		"totalHttpCount": 1,
-		"totalDsMissCount": 0,
-		"appStartTime": 1547584831677,
-		"averageDnsTime": 0,
-		"averageHttpTime": 1547584863270,
+		"totalHttpCount": 2,
+		"totalDsMissCount": 1,
+		"appStartTime": 1572759663457,
+		"averageHttpTime": 786379849939,
 		"updateTracker": {
-			"lastHttpsCertificatesCheck": 1547586068932,
-			"lastGeolocationDatabaseUpdaterUpdate": 1547584858917,
-			"lastCacheStateCheck": 1547586128932,
-			"lastCacheStateChange": 1547584867102,
-			"lastNetworkUpdaterUpdate": 1547584857484,
-			"lastHttpsCertificatesUpdate": 1547586071079,
-			"lastSteeringWatcherUpdate": 1547584923514,
-			"lastConfigCheck": 1547586127344,
-			"lastConfigChange": 1547584863406,
-			"lastNetworkUpdaterCheck": 1547584857465,
-			"lastGeolocationDatabaseUpdaterCheck": 1547584858906,
-			"lastFederationsWatcherUpdate": 1547584863433,
-			"lastHttpsCertificatesFetchSuccess": 1547586069070,
-			"lastSteeringWatcherCheck": 1547586124630,
-			"lastFederationsWatcherCheck": 1547586124584,
-			"lastHttpsCertificatesFetchAttempt": 1547586068932
-		}
+			"lastHttpsCertificatesCheck": 1572760605162,
+			"lastGeolocationDatabaseUpdaterUpdate": 1572759695064,
+			"lastCacheStateCheck": 1572760637365,
+			"lastCacheStateChange": 1572759703763,
+			"lastNetworkUpdaterUpdate": 1572759694009,
+			"lastHttpsCertificatesUpdate": 1572760607280,
+			"lastSteeringWatcherUpdate": 1572759760062,
+			"lastConfigCheck": 1572760633918,
+			"lastConfigChange": 1572759699962,
+			"lastNetworkUpdaterCheck": 1572759693998,
+			"lastGeolocationDatabaseUpdaterCheck": 1572759695056,
+			"lastFederationsWatcherUpdate": 1572759699998,
+			"lastHttpsCertificatesFetchSuccess": 1572760605275,
+			"lastSteeringWatcherCheck": 1572760601104,
+			"lastFederationsWatcherCheck": 1572760601085,
+			"lastHttpsCertificatesFetchAttempt": 1572760605162
+		},
+		"averageDnsTime": 0
 	}}
 
 .. _tr-api-crs-stats-ip-ip:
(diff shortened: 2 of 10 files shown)


docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -41,32 +41,32 @@ Request Structure
...
 	Accept: */*
 
 Response Structure
 ------------------
 .. code-block:: http
-	:caption: Response Example
+	:caption: Response Example (JSON expanded)
 
-	HTTP/1.1 200 OK
+	HTTP/1.1 200
```
ocket8888: Though the HTTP spec doesn't actually care whether or not the reason-phrase exists, our lexer does. They're not accepting pull requests, trust me I've tried.
So basically, this needs to be `HTTP/1.1 200 OK` not just `HTTP/1.1 200`.

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -122,19 +122,19 @@ Request Structure
...
```
ocket8888: Same as above RE: reason-phrase

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -160,20 +160,20 @@ Request Structure
...
 Response Structure
 ------------------
 :locations: An array of the names of :term:`Cache Groups` to which this Traffic Router is capable of routing client traffic
 
 .. code-block:: http
-	:caption: Response Example
+	:caption: Response Example (JSON expanded)
 
-	HTTP/1.1 200 OK
+	HTTP/1.1 200
```
ocket8888: Same as above RE: reason-phrase

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -190,27 +190,27 @@ Request Structure
...
 	Accept: */*
 
 Response Structure
 ------------------
 .. code-block:: http
-	:caption: Response Example
+	:caption: Response Example (JSON expanded)
 
-	HTTP/1.1 200 OK
+	HTTP/1.1 200
```
ocket8888: Same as above RE: reason-phrase

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -244,26 +244,26 @@ Request Structure
...
```
ocket8888: Same as above RE: reason-phrase

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -364,18 +364,18 @@ Request Structure
...
```
ocket8888: Same as above RE: reason-phrase

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -501,23 +508,22 @@ Request Structure
...
```
ocket8888: Same as above RE: reason-phrase

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -539,25 +545,24 @@ Request Structure
...
```
ocket8888: Same as above RE: reason-phrase

docs/source/development/traffic_router/traffic_router_api.rst
```diff
@@ -41,32 +41,32 @@ Request Structure
...
 	Host: trafficrouter.infra.ciab.test
-	User-Agent: curl/7.47.0
+	User-Agent: curl/7.52.1
 	Accept: */*
 
 Response Structure
 ------------------
 .. code-block:: http
-	:caption: Response Example
+	:caption: Response Example (JSON expanded)
```
ocket8888: You don't need to specify that the JSON has been expanded. Users are instructed to ignore the value of any `Content-Length` header.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: Reason-phrase re-added in 23d18b71a9.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: Reason-phrase re-added in 23d18b71a9.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: Reason-phrase re-added in 23d18b71a9.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: Reason-phrase re-added in 23d18b71.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: The reason-phrase was missing because I was sending HTTP requests and, for the API, [sendReasonPhrase is only set for HTTPS](https://github.com/apache/trafficcontrol/commit/9838c0e0d6#diff-383ce4b8e55d3e7e25a66bbe5b5f317eR40-R45). Is this PR a good place to add it for HTTP, or does that warrant its own PR?

Reason-phrase re-added in 23d18b71.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: Reason-phrase re-added in 23d18b71.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: Reason-phrase re-added in 23d18b71.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: Reason-phrase re-added in 23d18b71.

docs/source/development/traffic_router/traffic_router_api.rst
zrhoffman: Resolved in f682425f.

mitchell852:
> Docs look fine and I glanced over the Java and everything looks alright, but somebody more familiar with TR should take a look.

maybe @rawlinp or @ajschmidt can look at it.

ocket8888:
Can you rebase to fix conflicts? Idk if @rawlinp ever got around to that testing, but it can't be merged in any case while it has conflicts.

zrhoffman:
@ocket8888 Okay, I have rebased it. Merge conflicts are resolved.

rawlinp:
add to whitelist

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/http/BufferedResponseFilter.java
```diff
@@ -0,0 +1,46 @@
...
+
+		chain.doFilter(request, responseWrapper);
+
+		// Close the connection without waiting for the 10-second connect timeout,
+		// in case the client does not close the connection. Even though this is
+		// the only case for which we are interested in sending Connection: close,
+		// sending it sometimes means we must always send it. From RFC 2616:
+		// > HTTP/1.1 applications that do not support persistent connections MUST
+		// > include the "close" connection option in every message.
+		response.addHeader(HttpHeaders.CONNECTION, "close");
```
rawlinp, line 42: This worries me, because we might be fixing the performance of _invalid_ requests (e.g. `curl -X HEAD`), at the expense of _valid_ requests. Keeping client connections alive where possible can be a good thing.

rawlinp:
`mvn clean verify` appears to pass for me on master, but in this PR I get an error:
```
Results :

Tests in error:
  SteeringTest.z_itemsMigrateFromSmallerToLargerBucket:327 » HttpHostConnect Con...
```

zrhoffman:
Rebased to fix merge conflicts from #6390

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/http/BufferedResponseFilter.java
ocket8888: Does TR support HTTP/2? The spec for HTTP/2 says that clients must treat any message containing a connection-specific header as malformed.

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
```diff
@@ -0,0 +1,183 @@
...
+	@Test
+	public void itSetsContentLengthHeaderFor404() throws IOException {
+		final String encodedUrl = URLEncoder.encode("http://trafficrouter01.somedeliveryservice.somecdn.domain.foo/stuff", "utf-8");
+		final HttpGet httpGet = new HttpGet("http://localhost:3333/crs/deliveryservices?url=" + encodedUrl);
+		CloseableHttpResponse response = null;
+
+		try {
+			response = httpClient.execute(httpGet);
+			assertThat(response.getStatusLine().getStatusCode(), equalTo(404));
+			assertThat(response.getFirstHeader("Transfer-Encoding"), nullValue());
```
ocket8888, line 71: `Assert.assertThat` is deprecated, deprecation message says to use `MatcherAssert.assertThat` instead.

ocket8888, line 71: should use `org.springframework.http.HttpHeaders.TRANSFER_ENCODING` instead of string literal.

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
```diff
@@ -0,0 +1,183 @@
...
+			assertThat(response.getFirstHeader("Content-Length"), notNullValue());
```
ocket8888, line 72: should use `org.springframework.http.HttpHeaders.CONTENT_LENGTH` instead of string literal.

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
```diff
@@ -0,0 +1,183 @@
...
+			final List<Integer> contentLengths = new ArrayList<>();
+
+			for (final HttpRequestBase request : requests) {
+				CloseableHttpResponse response = null;
+
+				try {
+					response = httpClient.execute(request);
+					final Header contentLengthHeader = response.getFirstHeader("Content-Length");
+
+					assertThat(response.getFirstHeader("Transfer-Encoding"), nullValue());
```
ocket8888, line 99: same as the above 2

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
```diff
@@ -0,0 +1,183 @@
...
+			for (final HttpRequestBase request : requests) {
+				CloseableHttpResponse response = null;
+				request.setConfig(config);
+				request.addHeader("Host", testHostName);
+
+				try {
+					response = httpClient.execute(request);
+					final Header contentLengthHeader = response.getFirstHeader("Content-Length");
+
+					assertThat(response.getFirstHeader("Transfer-Encoding"), nullValue());
```
ocket8888, line 171: Same as above

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
```diff
@@ -0,0 +1,183 @@
...
+					assertThat(contentLengthHeader, notNullValue());
+					contentLengths.add(Integer.parseInt(contentLengthHeader.getValue()));
+				} finally {
+					if (response != null) response.close();
+				}
+			}
+
+			assertThat(contentLengths.size(), equalTo(2));
+			assertThat("Expected HEAD and GET requests for " + testHostName + path + "to have the same Content-Length", contentLengths.get(0), equalTo(contentLengths.get(1)));
```
ocket8888, line 180: I think you're missing a space before "to", causing it to get tacked on to the request path which can make it a bit hard to read.

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
```diff
@@ -0,0 +1,183 @@
...
+					assertThat(response.getFirstHeader("Transfer-Encoding"), nullValue());
+					assertThat(contentLengthHeader, notNullValue());
+					contentLengths.add(Integer.parseInt(contentLengthHeader.getValue()));
+				} finally {
+					if (response != null) response.close();
+				}
+			}
+
+			assertThat(contentLengths.size(), equalTo(2));
+			assertThat("Expected HEAD and GET requests for " + path + "to have the same Content-Length", contentLengths.get(0), equalTo(contentLengths.get(1)));
```
ocket8888, line 108: same as above

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
ocket8888, line 180: if TR's response isn't as expected, this could throw an `IndexOutOfBoundsException`. Idk much about Java testing, but you have a `try` block to catch something above, so is that alright? It also doesn't seem like an `IOException` so I'm not sure why the compiler isn't mad

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/LocationsTest.java
```diff
@@ -139,4 +144,38 @@ public void itGetsCachesForALocation() throws Exception {
...
+			fail(e.getMessage());
+		} finally {
+			if (response != null) response.close();
+		}
+
+		for (final String path : paths) {
+			final HttpHead httpHead = new HttpHead(path);
+			try {
+				response = closeableHttpClient.execute(httpHead);
+				assertThat(response.getStatusLine().getStatusCode(), equalTo(200));
```
ocket8888, line 175: Shouldn't you also check that it sets the Content-Length header?

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/http/BufferedResponseFilter.java
ocket8888: `close` is the default behavior for HTTP/1.0, so really you just want to add this header if the request is HTTP/1.1. You can check that with the request's `getProtocol` method, I think.

zrhoffman: > Does TR support HTTP/2?

No, and `curl -X HEAD` doesn't hang for HTTP/2 anyway.

As an alternative to `Connection: close`, we could just reduce `Connector.connectionTimeout` from 10 seconds to something shorter, like 3 or 5 seconds:

https://github.com/apache/trafficcontrol/blob/b24649869407d902301ce6997c629e367285b44f/traffic_router/core/src/main/conf/server.xml#L38

https://github.com/apache/trafficcontrol/blob/b24649869407d902301ce6997c629e367285b44f/traffic_router/core/src/main/conf/server.xml#L41

https://github.com/apache/trafficcontrol/blob/b24649869407d902301ce6997c629e367285b44f/traffic_router/core/src/main/conf/server.xml#L43

https://github.com/apache/trafficcontrol/blob/b24649869407d902301ce6997c629e367285b44f/traffic_router/core/src/main/conf/server.xml#L47

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
zrhoffman: Using `org.springframework.http.HttpHeaders.TRANSFER_ENCODING` in faafed127b

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
zrhoffman: Using `org.springframework.http.HttpHeaders.CONTENT_LENGTH` in faafed127b

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
zrhoffman: Using `org.springframework.http.HttpHeaders` in faafed127b

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
zrhoffman: Using `org.springframework.http.HttpHeaders.CONTENT_LENGTH` in faafed127b

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
zrhoffman: Space added in 21042581e2

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
zrhoffman: Space added in 21042581e2

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
zrhoffman: Using `MatcherAssert.assertThat` in 831774af3c

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/LocationsTest.java
zrhoffman: Checking that it sets the Content-Length header in 81bc50d2d0

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java
zrhoffman: > if TR's response isn't as expected, this could throw an `IndexOutOfBoundsException`.

No, because the line above it is
https://github.com/apache/trafficcontrol/blob/81bc50d2d0c32d7fe87c56fb33a23732f46c430a/traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/external/BufferedResponseTest.java#L170

> Idk much about Java testing, but you have a `try` block to catch something above, so is that alright?

Is what alright?

> It also doesn't seem like an `IOException` so I'm not sure why the compiler isn't mad

[`CloseableHttpResponse.execute()`](https://github.com/apache/httpcomponents-client/blob/rel/v4.5.13/httpclient/src/main/java/org/apache/http/impl/client/CloseableHttpClient.java#L106-L107) throws `IOException`:

```java
    public CloseableHttpResponse execute(
            final HttpUriRequest request) throws IOException, ClientProtocolException {
```

ocket8888: Based on what you're saying, I would guess that `assertThat` throws (some kind of `IOException`?) so that it won't actually proceed to the next line if the size assertion fails, is that right?

> Is what alright?

Is it alright for it to not catch the `IndexOutOfBoundsException`.

zrhoffman: > it won't actually proceed to the next line if the size assertion fails, is that right?

right

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/http/BufferedResponseFilter.java
ocket8888: Ideally @rawlinp would chime in to say whether or not that's acceptable

rawlinp: `Connection: close` still kind of worries me, especially because there may be clients out there relying on high connection rates to TR, whether that is really the _right_ approach or not, reusing connections could actually be helping TR performance. Maybe I don't fully understand the need for it here? Are `HEAD` requests broken without it?

Changing `connectionTimeout` in server.xml seems like it would be more fine-tuned via the CDN operator's ansible playbooks (or equivalent) if the CDN had issues with slow clients. I think Tomcat ships with a 20s default, and 10s seems reasonable for a TR default.

zrhoffman: > Looks like cloudfront.com is trying to close the connection but mistyped it as `Cneonction`:

As @rob05c pointed out to me, `Cneonction` is just [a hack](https://stackoverflow.com/questions/4798461/cneonction-and-nncoection-http-headers) to get around Connection header differences between the origin and caches.

> Maybe I don't fully understand the need for it here? Are `HEAD` requests broken without it?

Nope we don't need it, it would only be to appease invalid requests like `curl -X HEAD`. But `curl -X HEAD` doesn't cause a timeout anyway in HTTP/2, so there isn't a long-term reason to do it, especially with the benefits from reusing the connection like you mentioned.

Even if we did want to change any connection behavior, that belongs in a separate PR, #4074 should just be to resolve disparities between `HEAD` and `GET`.

zrhoffman:
Rebased to fix a changelog merge conflict

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/http/BufferedResponseFilter.java
zrhoffman: Forgot to mention: Reverted closing the connection in 97d38f2234

zrhoffman:
The *TR Ultimate Test Harness* workflow fails on the master branch (and on #4074), fixed by #7175.

zrhoffman:
Rebased onto master so tests will pass now that #7175 is merged

CHANGELOG.md
```diff
@@ -211,6 +212,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Updated `t3c-apply` to reduce mutable state in `TrafficOpsReq` struct.
 - Updated Golang dependencies
 - [#6506](https://github.com/apache/trafficcontrol/pull/6506) - Updated `jackson-databind` and `jackson-annotations` Traffic Router dependencies to version 2.13.1
+- Traffic Router now includes a `Content-Length` header in each response.
```
ocket8888, line 215: I don't believe this was added in 6.1. Unless you're doing this preemptively for a backport or something?

zrhoffman: Oops, didn't remove the 6.1.0 changelog entry in 88e2490439, which added a changelog entry to the unreleased section. Removed the 6.1.0 one in 7774d21768.

ocket8888 (approved):

PR #4079, opened 2019-11-04
Rewrote /federation_resolvers - POST

### LINKED ISSUE #3796: Rewrite /federation_resolvers to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/federation_resolvers` is handled by Perl

## Expected / new behavior:
`/federation_resolvers` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`
- [ ] `POST`

N/A

ocket8888:
- [x] This PR fixes #3796

Rewrite `/federation_resolvers` to Go. Also adds Traffic Ops and Traffic Control client code and associated tests (for the TO client, not the Python).

API version 1.3 makes no (intentional) changes from the Perl handler's behavior, but the 1.4 GET handler adds the common `last_updated` field to the response.

Finally, although this rewrite does not include `/federations/{{ID}}`, it does include one client method (for deleting Federation Resolvers) that uses that path.

- Documentation
- Traffic Ops Client (Golang)
- Traffic Ops

Run API and unit tests. Otherwise, just do GET/POST/DELETE requests to `/federation_resolvers` as documented.

- [x] This PR includes tests


traffic_ops/client/federation_resolver.go
```diff
@@ -0,0 +1,116 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+import "encoding/json"
+import "fmt"
+import "net/http"
+import "net/url"
+import "strconv"
```
awilliams: All files in the `client` package use the following import style. Consider matching for consistency.

```go
import (
  "a"
  "b"
  ...
)
```

ocket8888: That's not true. invalidationjobs and traffic_stats use the same style - as well as several others outside of the `client` package. and we don't have a style guide - only `go fmt` is enforced (which doesn't care).

docs/source/api/federation_resolvers.rst
```diff
@@ -19,6 +19,87 @@
...
+	| Name       | Required | Description                                                                                         |
+	+============+==========+=====================================================================================================+
+	| id         | no       | Return only the Federation Resolver identified by this integral, unique identifier                  |
+	+------------+----------+-----------------------------------------------------------------------------------------------------+
+	| ipAddress  | no       | Return only the Federation Resolver(s) that has/have this IP Address                                |
+	+------------+----------+-----------------------------------------------------------------------------------------------------+
+	| type       | no       | Return only the Federation Resolvers of this :term:`Type`                                           |
+	+------------+----------+-----------------------------------------------------------------------------------------------------+
+	| orderby    | no       | Choose the ordering of the results - must be the name of one of the fields of the objects in the    |
```
mhoppa: I think as we add new documentation we should fix `Choose the ordering of the results - must be the name of one of the fields of the objects in the ``response`` array` as that is not correct

mhoppa: We cant do it by lastUpdated

docs/source/api/federation_resolvers.rst
```diff
@@ -30,65 +111,57 @@ Creates a new federation resolver.
...
 -----------------
 :ipAddress: The IP address of the resolver - may be IPv4 or IPv6
-:typeId:    The integral, unique identifier of the type of resolver being created - will *represent* one of:
+:typeId:    The integral, unique identifier of the :term:`Type` of resolver being created
 
-	RESOLVE4
-		Resolver is for IPv4 addresses and ``ipAddress`` is IPv4
-	RESOLVE6
-		Resolver is for IPv6 addresses and ``ipAddress`` is IPv6
+	.. caution:: This field should only ever be an identifier for one of the :term:`Types` "RESOLVE4" or "RESOLVE6", but there is **no protection for this built into Traffic Ops** and therefore **any valid** :term:`Type` **identifier will be silently accepted by Traffic Ops** and so care should be taken to ensure that these :term:`Types` are properly identified. If any :term:`Type` besides "RESOLVE4" or "RESOLVE6" is identified, the resulting resolver *will* **not** *work*.
```
mhoppa: should we fix that in the rewrite? I couldnt imagine someone relying on this endpoint accepting invalid typeIds

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -0,0 +1,138 @@
...
+          ) AS type
+`
+
+const readQuery = `
+SELECT federation_resolver.id,
+       federation_resolver.ip_address,
+       federation_resolver.last_updated,
+       type.name AS type
+FROM federation_resolver
+LEFT OUTER JOIN type ON type.id = federation_resolver.type
```
mhoppa: reason for LEFT OUTER here compared to just LEFT JOIN?

docs/source/api/federation_resolvers.rst
ocket8888: My plan was to add a footnote clarifying that the `lastUpdated` field is unavailable for that. Do you think a different wording right in the table would be better?

mhoppa: that would be fine

docs/source/api/federation_resolvers.rst
ocket8888: :man_shrugging: I sort of wanted to, but it's been said that we should:

> _"... lean towards resisting the temptation during the perl -> go rewrites to change how things work as much as possible; even if it's wrong.  That way if something goes wrong unexpectedly, it's easier to determine which was the cause."_

Restricting the allowed values of `typeId` would be a breaking change so...

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
ocket8888: They're the same; `OUTER` is optional, just more explicit.

ocket8888:
Perl let you do the same thing. It's probably as a direct consequence of never validating the "Type". 

Honestly, that should never come into the equation, the payload should just be an IP Address of arbitrary version. Any special handling of it based on the protocol version can be figured out entirely on the back-end, and that's what I'd push for if this was being designed today.

docs/source/api/federation_resolvers.rst
ocket8888: I'm working on a branch to do this globally, since the problem exists globally.

mhoppa: awesome :) above and beyond 🚀

PR #4082, opened 2019-11-05
Fix federation user validation

### LINKED ISSUE #4068: Admin can assign User to a Federation with wrong PARAMETER
## I'm submitting a ...
- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Login as ADMIN
POST Request: `https://{{TO_BASE_URL}}/api/{{api_version}}/federations/{{federationsID}}/users. `
With wrong json payload parameter like "userIdssss"
```json
{
        "userIdssss": [2],
        "replaces": true
}
```
It will return 
```http
HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8

{
    "alerts": [
        {
            "text": "0 user(s) were Assign to the test.quest. federation",
            "level": "success"
        }
    ],
    "response": {
        "userIds": null,
        "replace": null
    }
}
```
## Expected / new behavior:
Should return 400 Bad Request

mhoppa:

This PR makes sure the on federation user post the userIDs is defined.

- [x] This PR fixes #4068 

- Traffic Ops

Make a POST request without a userIDs key to federation users endpoint and you should get back a validation error.

### DIFF
--- a/lib/go-tc/federation.go
+++ b/lib/go-tc/federation.go
@@ -26,6 +26,8 @@ import (
 	"net"
 
 	"github.com/apache/trafficcontrol/lib/go-util"
+
+	validation "github.com/go-ozzo/ozzo-validation"
 )
 
 // CDNFederationResponse represents a Traffic Ops API response to a request for one or more of a CDN's
@@ -184,7 +186,9 @@ type FederationUserPost struct {
 
 // Validate validates FederationUserPost
 func (f *FederationUserPost) Validate(tx *sql.Tx) error {
-	return nil
+	return validation.ValidateStruct(f,
+		validation.Field(&f.IDs, validation.NotNil),
+	)
 }
 
 type DeliveryServiceFederationResolverMapping struct {


lib/go-tc/federation.go
```diff
@@ -184,7 +186,9 @@ type FederationUserPost struct {
 
 // Validate validates FederationUserPost
 func (f *FederationUserPost) Validate(tx *sql.Tx) error {
-	return nil
+	return validation.ValidateStruct(f,
+		validation.Field(&f.IDs, validation.NotNil),
```
ocket8888: Does this also check for "not empty"?

ocket8888: You know what, that doesn't matter. The best anyone could hope for passing `userIds` as `null` or `undefined` is that it would do nothing. Specific status codes in response to certain requests aren't reliably documented (intentionally) so this isn't even a breaking change.

mhoppa: Perl did check if userIDs is defined and an array https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/API/FederationUser.pm#L61

ocket8888: Nice

PR #4084, opened 2019-11-05
TP: delivery service widget - changes search to ds xmlId rather than display name

### LINKED ISSUE #3753: unable to filter delivery services by xml-id on TP dashboard
on the TP dashboard is a delivery services widget that allows you to filter delivery services by "display name". it would be helpful from a usability perspective to allow a filter by xml id as well.

mitchell852:

xmlId is more widely used to identify a ds, therefore, switching the search to xmlId rather than display name.

This widget is not currently under test nor documented. Too small of a change to warrant a changelog.md entry.

- [x] This PR fixes #3753 

- Traffic Portal


mitchell852:
> Of course, after so long using XMLIDs operators may find the change jarring.

right. this PR is just to make consistent use of xmlId over displayname but i can see the argument of going the other way however, that's a bigger change. taking the shortest route to achieve consistency.

PR #4096, opened 2019-11-07
TR: Send reason-phrase for HTTP API responses

zrhoffman:

Reason-phrases are sent for routing over HTTP, routing over HTTPS, and the API over HTTPS, but not for the API over HTTP. This PR adds reason-phrases for the API over HTTP.

As @ocket8888 [noted](https://github.com/apache/trafficcontrol/pull/4074#discussion_r342164277) in #4074, Pygments needs reason-phrases for HTTP responses (of any HTTP version).

This PR does not include tests because
* The only component that depends on reason-phrases as far as I know is the documentation.
* HTTP/2 responses have no reason-phrases.
* If documentation ever stops failing to build due to no reason-phrase in a documented HTTP response, there is no reason to send one anymore, so building the documentation is a de facto test.

- Traffic Router

In the CDN-in-a-box directory,

```shell
docker-compose exec trafficrouter curl -I http://trafficrouter:3333/crs/stats
```

- master (1c2ebee4)
- 3.1.0

- [x] This PR does not need documentation because expected behavior does not change
- [x] This PR does not include an update to CHANGELOG.md, such an update is not necessary

### DIFF
--- a/traffic_router/core/src/main/conf/server.xml
+++ b/traffic_router/core/src/main/conf/server.xml
@@ -38,7 +38,8 @@
 				   connectionTimeout="10000" mbeanPath="traffic-router:name=languidState" readyAttribute="Ready" portAttribute="Port"
 					sendReasonPhrase="true"/>
 		<Connector port="3333" protocol="com.comcast.cdn.traffic_control.traffic_router.protocol.LanguidNioProtocol" maxThreads="10000"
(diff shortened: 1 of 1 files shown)


zrhoffman:
We're using Tomcat 9, which removes reason-phrases altogether, to #4096 is no longer needed.

PR #4097, opened 2019-11-07
Emulated vault

nir-sopher:
# What does this PR (Pull Request) do?

The PR introduces the `emulated_vault` module which supplies a HTTP server mimicking RIAK behavior for usage as Traffic Vault.
It may be used in order to replace RIAK-based Traffic Vault, as it is much more simple to install.
The server may use different type of persistent storage (e.g. file-system), using the proper adapter. 
The resiliency of the stored keys is derived from the resiliency of the underlying storage.

# Which Traffic Control components are affected by this PR?
None by default.
It introduces a module that may replace the RIAK-based Traffic Vault.

# What is the best way to verify this PR?
Install the emulated vault as described below, and add it to a running Traffic Ops instance instead of the standard RIAK-based Traffic Vault.
Run url-signing and SSL based scenarios.
Note: you would probably want to work on a fresh TC deployment, or somehow copy the RIAK vault to the emulated one...

# The following criteria are ALL met by this PR
- [x] This PR contains tests or I have explained why tests are unnecessary
- [x] This PR contains documentation or I have explained why tests are unnecessary 
- [x] This PR DOES NOT FIX A SERIOUS SECURITY VULNERABILITY (see the Apache Software Foundation's security guidelines for details)

Tests are not included in the PR as it is still under experimental. 
Would be a must if/when the code goes to production.
 
# Additional Information - Installation

Basic requirements: Centos ver >= 7; Python >= 2.7

In order to install the module on a server please:
1. Copy the module files to the server's root
2. Add the certificate and key to your favorite path
3. Adjust /opt/emulated_vault/conf/cfg.json - pointing at your certificate and key
4. "systemctl enable" the service

Logs may be found under /opt/emulated_vault/var/log

# Additional Information - Developer's Notes

If you just want to play around with the module, you may of course run the server script on its own.
Before doing that, you would probably need to adjust the opt/emulated_vault/conf/cfg.json:
1. Changing the db-path to one you have access to
2. Disable ssl (just to make it easier)

Additionally, the vault-cmd script is also available to work against the DB with command line. 
It is mostly useful when developing a new adapter.


experimental/emulated_vault/README
```diff
@@ -0,0 +1,23 @@
+The emulated_vault module supplies a HTTP server mimicking RIAK behavior for the usage as traffic-control vault.
+The server may use different type of persistent storage (e.g. file-system), using the proper adaptor.
+Basic requirements: Centos ver >= 7; Python
```
ocket8888: What versions of Python are supported (3.5+ only)?

experimental/emulated_vault/README
```diff
@@ -0,0 +1,23 @@
...
+
+In order to deploy the code on a server please:
+1. Copy the module files to the server's root
+2. Add the certificate and key to you favorite path
+3. Adjust opt/emulated_vault/conf/cfg.json - pointing at your certificate and key
+4. chkconfig the service
```
ocket8888: `chkconfig` is retired in CentOS version >= 7 as is required. I think you want to say something about a systemd command here?

experimental/emulated_vault/README
```diff
@@ -0,0 +1,23 @@
...
+
+Logs may be found under /opt/emulated_vault/var/log
+
+Developers notes:
+------------------------
```
ocket8888: Is this a Markdown file? if so, this isn't how you do a heading in markdown, those are done like:
```markdown
# Developer's Notes
```
if this is meant to be a reStructuredText file, then the length of the underline must match the length of the title itself - and the level of hyphens is improper, should probably be <kbd>=</kbd>.

Either way, you wanna have have a top-level title for this README.

ocket8888:
Please fill out the Pull Request template to the best of your ability, and avoid deleting it entirely.

experimental/emulated_vault/README
nir-sopher: Tested on python 2.7. Will update

experimental/emulated_vault/README
nir-sopher: systemctl enable ;)
Thanks for catching it

experimental/emulated_vault/README
nir-sopher: Ack
Thanks

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+            self.send_response(200)
+            self.send_header('Content-Type', 'application/json')
+            self.end_headers()
+            self.wfile.write(json.dumps(toReturn))
+            return
+
+        logger.info("GET %s", self.path)
+        success, value = adaptor.getParameter(parsed_path.path[1:])
+        if not success:
+            self.send_response(503)
```
ocket8888: is this the appropriate response code for not finding a parameter? Should this not be a more generic `500 Internal Server Error` response since we don't know why the value could not be gotten?

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+
+    do_PUT = do_POST
+
+    def do_DELETE(self):
+        try:
+            self._do_DELETE()
+        except:
+            logger.exception("do_DELETE exception")
+
+    def _do_GET(self):
```
ocket8888: Maybe you wanna separate out the path-specific handlers into their own methods instead of separating out all of the real logic of GET handling into a single, massive function?

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+
+import json
+import logging
+import sys
+import os
+import inspect
+from BaseHTTPServer import HTTPServer
+from BaseHTTPServer import BaseHTTPRequestHandler
+import ssl
+import urlparse, json
```
ocket8888: urlparse no longer exists, I think you want [`urllib.parse`](https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse) instead.

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+            self.send_header('Content-type', 'text/html')
+            self.end_headers()
+            self.wfile.write(value)
+            return
+
+        if parsed_path.path == "/search/query/sslkeys":
+            logger.info("Search SSL: path=%s query=%s", parsed_path.path, parsed_path.query)
+            filters = {}
+            keyFilters = {}
+            cdnFind = re.search(".*q=cdn:([^&]*).*", parsed_path.query)
```
ocket8888: instead of using regular expressions to search through the query string, consider using [`urllib.parse.parse_qs`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.parse_qs) to parse out a list of values for a given query parameter

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+                        return data['deliveryservice']==dsFind.group(1)
+                    except:
+                        return False
+                filters['ds'] = dsFilter
+
+            keyFind = re.search(".*q=_yz_rk:([^&]*).*", parsed_path.query)
+            if keyFind:
+                def keyFilter(key):
+                    return fnmatch.fnmatch(os.path.basename(key), keyFind.group(1))
+                keyFilters['key-match'] = keyFilter
```
ocket8888: nit: could be a lambda

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+
+            success, parameters = adaptor.searchParameters("riak/ssl/", keyFilters=keyFilters, filters=filters)
+            if not success:
+                self.send_response(503)
```
ocket8888: is this the right response code for a failure of parameter search?

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+        self.send_response(200)
+        self.send_header('Content-Type', 'application/json')
+        self.end_headers()
+        self.wfile.write(value)
+
+    def _do_POST(self):
+        logger.info("POST %s", self.path)
+        parsed_path = urlparse.urlparse(self.path)
+        content_len = int(self.headers.getheader('content-length'))
+        post_body = self.rfile.read(content_len)
```
ocket8888: this seems to intentionally not support chunked transfer encoding - is that intentional?

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+        data = json.loads(post_body)
```
ocket8888: unhandled `json.JSONDecodeError`

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+        #mimic vault beahvior
+        if parsed_path.path.startswith('/riak/ssl/'):
+            for item in data['certificate']:
```
ocket8888: probably want to safely access with `dict.get`.

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+                data['certificate.%s'%item]=data['certificate'][item]
```
ocket8888: Seems like a good use-case for format strings, dict comprehension, and key/value dict iteration:
```python3
data = {f"certificate.{key}": value for key, value in data.get("certificate", {})} 
```

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+        success = adaptor.setParameter(parsed_path.path[1:], json.dumps(data))
```
ocket8888: unhandled `json.JSONEncodeError` here

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+        self.end_headers()
+        return
+
+
+    def _do_DELETE(self):
+        logger.info("DELETE %s", self.path)
+        parsed_path = urlparse.urlparse(self.path)
+
+        success = adaptor.deleteParameter(parsed_path.path[1:])
+        if not success:
```
ocket8888: nit: could just check for success of the method directly without storing to an interim symbol

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+        logger.error("Failed initialization")
+        sys.exit(1)        
+
+    server = HTTPServer((listenIP, listenPort), RequestHandler)
+    if use_ssl:
+        key = cfg["data"]["ssl-key-path"]
+        cert = cfg["data"]["ssl-cert-path"]
+        server.socket = ssl.wrap_socket (server.socket, keyfile=key, certfile=cert, server_side=True)
+    msg = 'Starting server at http%s://%s:%d'%("s" if use_ssl else "", listenIP, listenPort)
+    print >> sys.stderr, msg
```
ocket8888: SyntaxError here, I think you probably want something like:
```python3
print(f"Starting server at http{'s' if use_ssl else ''}://{listenIP}:{listenPort}", file=sys.stderr)
```

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+        sys.exit(1)
+
+    if not adaptor.init_ping():
+        logger.error("Failed initialization")
+        sys.exit(1)        
+
+    server = HTTPServer((listenIP, listenPort), RequestHandler)
+    if use_ssl:
+        key = cfg["data"]["ssl-key-path"]
+        cert = cfg["data"]["ssl-cert-path"]
```
ocket8888: two unhandled `KeyError`s here

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+        server.socket = ssl.wrap_socket (server.socket, keyfile=key, certfile=cert, server_side=True)
```
ocket8888: two unhandled `OSError`s (probably `FileNotFoundError` or `PermissionDeniedError`) and an `ssl.SSLError` here.

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+    # add the handlers to the logger
+    logger.addHandler(fhm)
+    logger.addHandler(fhd)
+
+    basePath = cfg["data"]["db-path"]
+    pingRelPath = cfg["data"]["ping-rel-path"]
+    listenIP = cfg["data"]["listen-ip"]
+    listenPort = cfg["data"]["listen-port"]
+    adaptorType = cfg["data"]["adaptor-type"]
+    use_ssl = cfg["data"]["use-ssl"]
```
ocket8888: six unhandled `KeyError`s here

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+    confDir = os.path.join(progAbsPath, "conf")
+    confFile = os.path.join(confDir, "cfg.json")
+    with open(confFile) as fd:
+        cfg = json.load(fd)
+
+    logDir = os.path.join(progAbsPath, "var/log")
+    debugLogFile = os.path.join(logDir, "traffic-ops-vault-debug.log")
+    mainLogFile = os.path.join(logDir, "traffic-ops-vault.log")
+    if not os.path.exists(logDir):
+        os.makedirs(logDir)
```
ocket8888: unhandled `OSError` here

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+
+if __name__ == '__main__':
+    progPath = inspect.stack()[-1][1]
+    progAbsPath = os.path.abspath( progPath )
+    progAbsPath = os.path.dirname(os.path.normpath(progAbsPath))
+
+    confDir = os.path.join(progAbsPath, "conf")
+    confFile = os.path.join(confDir, "cfg.json")
+    with open(confFile) as fd:
+        cfg = json.load(fd)
```
ocket8888: unhandled `OSError` and `json.JSONEncodeError` here

experimental/emulated_vault/opt/emulated-vault/conf/cfg.json
```diff
@@ -0,0 +1,13 @@
...
+        "db-path": "/opt/emulated-vault/db/",
+        "adaptor-type": "fs",
+        "listen-ip": "0.0.0.0",
+        "listen-port": 8088,
+        "use-ssl": true,
+        "ssl-key-path": "path/to/cert/key.pem",
+        "ssl-cert-path": "path/to/cert/cert.pem"
+    }, 
+    "format-version": 100
+}
```
ocket8888: Instead of using JSON, you could maybe make this a bit easier on yourself by using [the Python standard library's configuraiton file format](https://docs.python.org/3/library/configparser.html)

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
```diff
@@ -0,0 +1,166 @@
...
+            return False, ""
+        if value is None:
+            self.logger.error("Could not find variable %s", parameterPath)
+            return True, None
+        self.logger.debug("Parameter get response for %s ready", parameterPath)
+        #self.logger.debug("Parameter get response: %s", value)
+        return True, value
+
+    def searchParameters(self, parameterPathPrefix, keyFilters = {}, filters = {}):
+        parameterPathPrefix = parameterPathPrefix.lstrip("/")
```
ocket8888: same as above RE `os.sep`

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+
+    server = HTTPServer((listenIP, listenPort), RequestHandler)
+    if use_ssl:
+        key = cfg["data"]["ssl-key-path"]
+        cert = cfg["data"]["ssl-cert-path"]
+        server.socket = ssl.wrap_socket (server.socket, keyfile=key, certfile=cert, server_side=True)
+    msg = 'Starting server at http%s://%s:%d'%("s" if use_ssl else "", listenIP, listenPort)
+    print >> sys.stderr, msg
+    logger.info(msg)
+    server.serve_forever()
```
ocket8888: This file is also missing docstrings

experimental/emulated_vault/README
rob05c: You can do a header in Markdown like this. It's uncommon, but it's valid, in both the original spec (https://daringfireball.net/projects/markdown/syntax#header) and GHM (https://github.github.com/gfm/#setext-heading).

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
```diff
@@ -0,0 +1,166 @@
+#
```
ocket8888: I think this whole package ought to be `adapter` as that's the proper spelling.

experimental/emulated_vault/README
```diff
@@ -0,0 +1,32 @@
+# Emulated Vault - Background
+
+The emulated_vault module supplies a HTTP server mimicking RIAK behavior for usage as traffic-control vault.
+It may be used in order to replace RIAK traffic_vault, as it is much more simple to install.
+The server may use different type of persistent storage (e.g. file-system), using the proper adaptor. 
+The resiliency of the stored keys is derived from the resiliency of the underlying storage.
+
+# Installation
+
+Basic requirements: Centos ver >= 7; Python >= 2.7
```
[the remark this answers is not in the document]
nir-sopher: @ocket8888 
Thank you for the effort you put into this review. Much appreciated.
My apologies as fixes would not be immediate. I would need to allocate time for resolving all comments, and especially for the Python 3 transition.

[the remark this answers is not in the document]
ocket8888: much of a transition from py2 to py3 can be done with the built-in Python 3 tool `2to3` automatically. You may want to look into it.

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
```diff
@@ -0,0 +1,166 @@
...
+        filtered = {}
+        for key, val in items.iteritems():
+            skip = False
+            for filterName, filter in filters.iteritems():
+                if not filter(key, val):
+                    self.logger.debug("Parameter %s dropped, not matching filter %s", key, filterName)
+                    skip = True
+                    break
+            if skip:
+                continue
```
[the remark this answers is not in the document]
nir-sopher: I find this syntax more elegant but less clear.
Prefer to keep it as is

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
```diff
@@ -0,0 +1,166 @@
...
+            self.logger.error("Failed to delete variable %s", parameterPath)
+            return False
+        return True
+
+    def _get_param(self, paramName):
+        self.logger.debug("Get parameter %s", paramName)
+        try:
+            with open(paramName) as fd:
+                value = fd.read()
+        except:
```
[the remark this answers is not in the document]
nir-sopher: doesn't logger.exception do just that?

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
```diff
@@ -0,0 +1,166 @@
...
+
+        for fileName in fileNames:
+            filteredOut = False
+            for filterName, filter in keyFilters.iteritems():
+                if not filter(fileName):
+                    self.logger.debug("Parameter %s dropped, not matching filter %s", fileName, filterName)
+                    filteredOut = True
+                    break
+            if filteredOut:
+                continue
```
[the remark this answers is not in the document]
nir-sopher: Again. I prefer KISS

experimental/emulated_vault/opt/emulated-vault/vault-cmd
```diff
@@ -0,0 +1,162 @@
...
+        sys.exit(0)
+
+    if options.getParameter:
+        if not options.parameterPath:
+            parser.error("Missing parameter-path.")
+        success, value = adaptor.getParameter(options.parameterPath)
+        if not success:
+            sys.exit(1)
+        if value is None:
+            sys.exit(1)
```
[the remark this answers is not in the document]
nir-sopher: I find being more explicit clearer to read

experimental/emulated_vault/opt/emulated-vault/vault-cmd
```diff
@@ -0,0 +1,162 @@
...
+        success, value = adaptor.searchParameters(options.parameterPath)
+        if not success:
+            sys.exit(1)
+        print json.dumps(value)
+        sys.exit(0)
+
+    if options.setParameter:
+        if not options.parameterPath:
+            parser.error("Missing parameter-path.")
+        if len(args) != 1:
```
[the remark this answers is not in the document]
nir-sopher: moved to"vault-cmd --set-parameter test quest" following an earlier comment

experimental/emulated_vault/opt/emulated-vault/vault-server
nir-sopher: I wrote it a while ago - need to check what is RIAK behavior. Might have been copied from it

experimental/emulated_vault/opt/emulated-vault/vault-server
nir-sopher: Probably copied RIAK behavior

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
```diff
@@ -0,0 +1,166 @@
...
+    def searchParameters(self, parameterPathPrefix, keyFilters = {}, filters = {}):
+        parameterPathPrefix = parameterPathPrefix.lstrip("/")
+        _parametersPathPrefix = os.path.join(self.basePath, parameterPathPrefix)
+        success, items = self._get_params(_parametersPathPrefix, prefixToRemove=self.basePath, keyFilters=keyFilters)
+        if not success:
+            self.logger.error("Failed to bring variable by prefix %s", parameterPathPrefix)
+            return False, ""
+
+        filtered = {}
+        for key, val in items.iteritems():
```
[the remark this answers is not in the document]
nir-sopher: moving to "items" to support both python 2 and 3

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
```diff
@@ -0,0 +1,166 @@
...
+        prefixToRemoveLen = len(prefixToRemove)
+        parameters = {}
+
+        fileNames = []
+        for (dirpath, dirnames, filenames) in os.walk(paramNamePrefix):
+            fileNames += [os.path.join(dirpath, file) for file in filenames]
+
+        for fileName in fileNames:
+            filteredOut = False
+            for filterName, filter in keyFilters.iteritems():
```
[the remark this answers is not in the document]
nir-sopher: moving to "items" to support both python 2 and 3

experimental/emulated_vault/opt/emulated-vault/vault-cmd
```diff
@@ -0,0 +1,162 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+#
+
+import json
+import logging
+import optparse
```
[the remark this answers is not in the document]
nir-sopher: as it still works with python 3, I prefer to leave it as is and support both versions for now

experimental/emulated_vault/opt/emulated-vault/vault-server
nir-sopher: I prefer not to mess with the logic. It was written a while ago and we do not have an automated test for it. Lets fix if it is taken to production.

experimental/emulated_vault/opt/emulated-vault/vault-server
nir-sopher: Unless the contribution guideline instruct otherwise, I prefer this still

experimental/emulated_vault/README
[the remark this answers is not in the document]
nir-sopher: aligned the code to work with py2 and py3

experimental/emulated_vault/opt/emulated-vault/conf/cfg.ini
```diff
@@ -0,0 +1,13 @@
+[general]
```
ocket8888: Love to see this format being adopted :+1: 
but you're missing a license header in this file

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,346 @@
...
+    from http.server import HTTPServer
+    from http.server import BaseHTTPRequestHandler
+    from configparser import ConfigParser
+    import urllib.parse as urlparse
+else:    
+    #python 2
+    from BaseHTTPServer import HTTPServer
+    from BaseHTTPServer import BaseHTTPRequestHandler
+    from ConfigParser import ConfigParser
+    import urlparse
```
ocket8888: This all looks good, but if you wanted to use `print` instead of `sys.stderr.write` you could also add `from __future__ import print_function`. I think it should be fine either way, just wanted to be sure you were aware.

experimental/emulated_vault/opt/emulated-vault/vault-server
ocket8888: Traffic Ops just checks for 200 responses (2XX? not sure), so it shouldn't care. But [the RIAK docs](https://docs.riak.com/riak/kv/2.2.3/developing/api/http/fetch-object.1.html#response)  says that it will return 503 in the event of an internal timeout whereas a key not being found is a `404 Not Found`. This probably will work with ATC today, and it may not ever make a behavioral difference, but even if only for the purposes of logging info it might be better to send the more appropriate `404 Not Found`.

experimental/emulated_vault/opt/emulated-vault/vault-server
ocket8888: I don't think the linter cares, will check in a bit.

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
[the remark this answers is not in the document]
ocket8888: Not exactly. By "swallow errors" I don't mean that it hides them from ever being seen, just that it can capture more than you would want. A bare `except` can capture things like a `SystemExit`, `SystemError` that indicate an issue with the interpreter, or `KeyboardInterrupt` which indicates that a running user (or control system like `systemd`) wants it to stop.

`except`ing `Exception` is similarly frowned-upon in the general Python community, but would at least avoid those problems, because all of those classes inherit from `BaseException` not `Exception`. If this was me I'd probably write something like:

```python3
try:
    with open(paramName) as fd:
        value = fd.read()
except OSError:
    self.logger.exception("%s parameter not found.", paramName)
    return False, None
```
and then maybe something under that to handle `KeyboardInterrupt` without printing a stack trace in the logs:
```
except KeyboardInterrupt:
    logging.error("Exiting...")
    sys.exit(1)
```
or maybe not.

experimental/emulated_vault/opt/emulated-vault/adaptor/fs.py
```diff
@@ -0,0 +1,166 @@
...
+        _parameterPath = os.path.join(self.basePath, parameterPath)
+        success = self._set_param(_parameterPath, value)
+        if not success:
+            self.logger.error("Failed to set variable %s", parameterPath)
+            return False
+        return True
+
+    def deleteParameter(self, parameterPath):
+        parameterPath = parameterPath.lstrip("/")
+        _parameterPath = os.path.join(self.basePath, parameterPath)
```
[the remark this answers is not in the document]
ocket8888: or you could use `os.sep.split`

experimental/emulated_vault/opt/emulated-vault/vault-cmd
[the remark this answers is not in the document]
ocket8888: I disagree that that's more explicit, but if the linter doesn't care I don't.

experimental/emulated_vault/opt/emulated-vault/vault-server
nir-sopher: How do I change it?

experimental/emulated_vault/opt/emulated-vault/vault-server
nir-sopher: The exception is caught and logged in the wrapping function. As I would not try to recover from the failure indicated by the exception, I do not see a real reason to handle it here just to raise it again.

experimental/emulated_vault/opt/emulated-vault/vault-server
nir-sopher: The exception is caught and logged in the wrapping function. As I would not try to recover from the failure indicated by the exception, I do not see a real reason to handle it here just to raise it again.

experimental/emulated_vault/opt/emulated-vault/vault-server
nir-sopher: Ack. Thanks

experimental/emulated_vault/opt/emulated-vault/vault-cmd
```diff
@@ -0,0 +1,162 @@
...
+    logger.addHandler(fhd)
+    logger.addHandler(verbose)
+    
+    parser = optparse.OptionParser()
+    operationModeGroup = optparse.OptionGroup(parser, "Operation mode options",
+                                              "The vault script may work in one of the following modes.")
+    parser.add_option_group(operationModeGroup)
+
+    operationModeGroup.add_option("--ping",
+                                  action="store_true", dest="ping", default=False, help="Ping")
```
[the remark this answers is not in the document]
nir-sopher: Marking as resolved as I'm staying with opt-parse.
Will reopen if we decide to change

experimental/emulated_vault/opt/emulated-vault/vault-cmd
```diff
@@ -0,0 +1,162 @@
...
+    operationModeGroup.add_option("--set-parameter",
+                                  action="store_true", dest="setParameter", default=False, 
+                                  help="Set parameter. Parameter relative path should be provided.")
+
+    operationModeGroup.add_option("--delete-parameter",
+                                  action="store_true", dest="deleteParameter", default=False, 
+                                  help="Delete parameter. Parameter relative path should be provided.")
+
+    operationModeGroup.add_option("--search-parameters",
+                                  action="store_true", dest="searchParameters", default=False,
```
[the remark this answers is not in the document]
nir-sopher: Marking as resolved as I'm staying with opt-parse.
Will reopen if we decide to change

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,249 @@
...
+import urlparse, json
+import re, fnmatch
+
+import adaptor.fs
+
+class RequestHandler(BaseHTTPRequestHandler):
+
+    def do_GET(self):
+        try:
+            self._do_GET()
```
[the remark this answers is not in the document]
nir-sopher: Just wrapping with "try/except". No strong preference either way

experimental/emulated_vault/opt/emulated-vault/vault-server
ocket8888: Well, you're explicitly checking for `content-length`. if you don't find it and `transfer-encoding` is set to `chunked`, then you've gotta read in the chunks. I'd be shocked if the HTTP server library didn't have something to do that for you (may just be as easy as `self.body()` instead of `self.rfile.read()`).

nir-sopher:
@ocket8888 
As far as I can see, there are 2 type of remarks left to be handled.
1) Coding style
2) Server behavior (response codes, encoding).
For #1 I replied.
For #2 I prefer to open an issue and make a separate PR

ocket8888:
That's certainly fine with me, especially since this is experimental. I'll try to be a bit lax about the contribution guidelines as well, with that fact that it is experimental in mind.

experimental/emulated_vault/opt/emulated-vault/vault-debug
```diff
@@ -0,0 +1,186 @@
...
+            sys.exit(1)
+        sys.exit(0)
+
+    logger.error("Operation is not set")
+    parser.print_help()
+    parser.error("No operation mode specified")
+    sys.exit(1)
+
+
+
```
ocket8888: current score of this file is -4.72/10. There's no established baseline for what experimental code has to score - and I'm not sure there should be - but you can increase it to a 5.92 by making two small changes:

- use tabs for indentation
- The naming conventions want all global scope variables to be UPPER_CASE, so currently it has a problem with every declaration under your `if __name__ == '__main__':` statement. There's two ways to fix that, one is to make them all UPPER_CASE, but that's probably not what you want. The other is to wrap it in a function like:

```python
def main():
	'''The main logic'''
	# other code here...

if __name__ == '__main__':
	main()
```
Personally I like to do this and make `main` return an integer return code so that it can be called from an interactive prompt without exiting the session. Then under `if __name__ ==  'main':` I just do `sys.exit(main())`. That's not a requirement or anything, just my hot tip of the day.

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,347 @@
+#! /usr/bin/python
```
ocket8888: The current score of this file is -6.34/10, but it can be raised as high as 4.42 using the same methods as above.

As a side note, there's quite a few false positives like `print_function is unused` and `do_GET does not conform to method naming pattern`. Wonder if there's anything we can do about that in the configuration...

experimental/emulated_vault/opt/emulated-vault/adapter/fs.py
```diff
@@ -0,0 +1,313 @@
+#
```
ocket8888: The current score of this file is -7.37/10, but that can be raised as high as 6.24 just by using tabs for indentation.

By the way, for future work into making more adapters I just wanna suggest writing out an [abstract base class](https://docs.python.org/3/library/abc.html) for people to implement. Seems like a good use of it, and it'd be helpful for type hints if you ever move to that.

nir-sopher: I actually started to work on it this morning but did not wanted to move files in the middle of the review.
Would you prefer me to push it now?

ocket8888: You mean abstract base classes? I don't mind if you want to push updates, I can start my review over. If you'd rather see this merged first and then work on top of that, that's fine too.

nir-sopher: Not exactly abstract, but base with some un-implemented functions.
See pushed update

experimental/emulated_vault/opt/emulated-vault/adapter/fs.py
```diff
@@ -16,298 +16,206 @@
...
+		return True
+
+	def _remove_parameter_by_storage_path(self, parameterStoragePath):
+		"""
+		Deleting the the provided file.
+		:param parameterStoragePath: the file name 
+		:type parameterStoragePath: str
+		:return: 'True' for successful deletion
+		:rtype: Boolean		
+	   """
```
ocket8888: Not sure how this happened, but these last two methods' closing triple-quotes are offset by one tab and three spaces instead of two tabs like the rest.

experimental/emulated_vault/opt/emulated-vault/adapter/fs.py
```diff
@@ -16,298 +16,206 @@
...
+
+	def _write_parameter_by_storage_path(self, parameterStoragePath, value):
+		"""
+		Writing the value to the provided file name.
+		:param parameterStoragePath: the file name 
+		:type parameterStoragePath: str
+		:param value: value to be writen
+		:type value: str
+		:return: 'True' for successful writing
+		:rtype: Boolean
```
ocket8888: You clearly want to support Python 2, so feel free to ignore this (although note it will no longer receive even critical security updates after next month) but if you abandon Python 2 support and use type hinting you can skip all of these `:type` and `:rtype:` tags:

```python3
def _write_parameter_by_storage_path(self, parameterStoragePath: str, value: str) -> bool:
	"""
	Writing the value to the provided file name.
	:param parameterStoragePath: the file name
	:param value: value to be written
	:return: Whether or not the write was successful
	"""
	...
```

experimental/emulated_vault/opt/emulated-vault/vault-debug
```diff
@@ -29,158 +29,158 @@ import sys
...
-        sys.exit(0)
-
-    logger.error("Operation is not set")
-    parser.print_help()
-    parser.error("No operation mode specified")
-    sys.exit(1)
-
-
-
+	sys.exit(main())
```
ocket8888: missing newline at EOF

experimental/emulated_vault/opt/emulated-vault/adapter/base.py
```diff
@@ -0,0 +1,292 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+#
+#
+
+import os
+
+class Base(object):
```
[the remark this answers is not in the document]
nir-sopher: Wow. Did not know python has that!
It would however require some restructuring of the code:
The adapter base class would become a "db" class that gets a "storage adapter"
Storage adapter would be abstract with fs as implementation.
Makes sense?

experimental/emulated_vault/opt/emulated-vault/adapter/fs.py
nir-sopher: Ack

experimental/emulated_vault/opt/emulated-vault/adapter/base.py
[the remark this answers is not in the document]
ocket8888: Yeah that sounds fine

experimental/emulated_vault/opt/emulated-vault/storage/adapter_base.py
```diff
@@ -0,0 +1,162 @@
...
+	def remove_parameter_by_storage_path(self, parameterStoragePath):# -> bool:
+		"""
+		Method to be implemented at derived classes
+		Deleting the the provided file.
+		:param parameterStoragePath: the file name 
+		:type parameterUrlPath: str
+		:return: 'True' for successful deletion
+		:rtype: bool		
+		"""
+		raise NotImplemented()#...
```
ocket8888: missing newline at EOF

nir-sopher: Forgot to commit this fix :(
Sorry

nir-sopher:
@ocket8888 Any additional AIs on my side?
Thanks,
Nir

experimental/emulated_vault/opt/emulated-vault/vault-server
```diff
@@ -0,0 +1,355 @@
...
+
+		success, parameters = db.searchParameters("/riak/ssl/", keyFilters=keyFilters, filters=filters)
+		if not success:
+			self.send_response(503)
+			self.send_header('Content-Type', 'application/json')
+			self.end_headers()
+			self.wfile.write('{"Failure"}'.encode())
+			return
+
+		docs = [json.loads(val) for i in parameters.values()]
```
ocket8888: `val` looks undefined here.

experimental/emulated_vault/opt/emulated-vault/storage/adapter_base.py
```diff
@@ -0,0 +1,163 @@
...
+	def init_cfg(self, fullConfig):# -> bool:
+		"""
+		Method to be implemented at derived classes
+		Initialize the class basic parameters. Part of Adapter required API.
+		:param fullConfig: configuration to operate upon.
+		:type fullConfig: configparser.ConfigParser class
+		:return: 'True' for successful initialization
+		:rtype: bool
+		"""
+		raise NotImplemented()#...
```
ocket8888: `NotImplemented` is not callable, it's a language constant. You don't `raise NotImplemented`, you `return NotImplemented`, but that's not appropriate here. I think what you want to do is `raise NotImplementedError()`.

experimental/emulated_vault/opt/emulated-vault/storage/adapter_base.py
```diff
@@ -0,0 +1,163 @@
...
+		"""
+		Method to be implemented at derived classes
+		Deleting the the provided file.
+		:param parameterStoragePath: the file name 
+		:type parameterUrlPath: str
+		:return: 'True' for successful deletion
+		:rtype: bool		
+		"""
+		raise NotImplemented()#...
+
```
ocket8888: Missing newline at EOF

PR #4099, opened 2019-11-08
Correctly handle assigning org servers to ds that have req caps

mhoppa:

Currently in the Delivery Service Required Capabilities if a delivery service has a required capability then all assigned servers must have it. This PR changes that so origin servers can be assigned to DSes even if they dont have the required capabilities.

***also fixes bug where origin to DS assignment could not happen in TP***

- Documentation
- Traffic Ops
- Traffic Portal

Assign required Capabilities to DSes and then make sure you can still assign origin servers to MSO DSes 

- [x] This PR includes tests

### Demo of TP Origin to Server Assignment
Nov-07-2019 20-44-28

### DIFF
--- a/docs/source/api/deliveryservices_id_servers_eligible.rst
+++ b/docs/source/api/deliveryservices_id_servers_eligible.rst
@@ -27,7 +27,7 @@ Retrieves properties of Edge-Tier servers eligible for assignment to a particula
 
 1. Server Type: Edge, Origin
 2. CDN: Server, :term:`Delivery Service` must belong to the same CDN
-3. Server Capabilities: If the :term:`Delivery Service` has  :term:`Delivery Service required capabilities`, a server must have all defined capabilities
+3. Server Capabilities: If the :term:`Delivery Service` has  :term:`Delivery Service required capabilities`, an edge server must have all defined capabilities
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"\ [1]_
--- a/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices_required_capabilities_test.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices_required_capabilities_test.go
@@ -78,7 +78,7 @@ func TestCreateDeliveryServicesRequiredCapability(t *testing.T) {
 	mockTenantID(t, mock, 1)
 
 	arrayRows := sqlmock.NewRows([]string{"array"})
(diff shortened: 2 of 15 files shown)


docs/source/api/deliveryservices_id_servers_eligible.rst
```diff
@@ -27,7 +27,7 @@ Retrieves properties of Edge-Tier servers eligible for assignment to a particula
 
 1. Server Type: Edge, Origin
 2. CDN: Server, :term:`Delivery Service` must belong to the same CDN
-3. Server Capabilities: If the :term:`Delivery Service` has  :term:`Delivery Service required capabilities`, a server must have all defined capabilities
+3. Server Capabilities: If the :term:`Delivery Service` has  :term:`Delivery Service required capabilities`, an edge server must have all defined capabilities
```
lactose: should this be origin server?

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices_required_capabilities_test.go
```diff
@@ -78,7 +78,7 @@ func TestCreateDeliveryServicesRequiredCapability(t *testing.T) {
 	mockTenantID(t, mock, 1)
 
 	arrayRows := sqlmock.NewRows([]string{"array"})
-	mock.ExpectQuery("SELECT server FROM deliveryservice_server").WillReturnRows(arrayRows)
+	mock.ExpectQuery("SELECT ds.server FROM deliveryservice_server").WillReturnRows(arrayRows)
```
lactose: no real comment here, just curious why this needed to be done? should it be documented?

docs/source/api/deliveryservices_id_servers_eligible.rst
ocket8888: No, because mids can't be assigned to Delivery Services directly. So only Edges need to be called out here.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices_required_capabilities_test.go
mhoppa: I changed the query and the mock library does a regex on the query to mock the return

docs/source/api/deliveryservices_id_servers_eligible.rst
mitchell852: technically i think mids can be assigned directly to ds's but it will have no effect.

PR #4105, opened 2019-11-09
Rewrite /federation_resolvers/{{ID}} to Go

### LINKED ISSUE #3797: Rewrite /federation_resolvers/{{id}} to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/federation_resolvers/{{id}}` is handled by Perl

## Expected / new behavior:
`/federation_resolvers/{{id}}` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `DELETE`

N/A

ocket8888:
- [x] This PR fixes #3797 

Rewrites the `/federation_resolvers/{{ID}}` endpoint to Go. Also adds a `DELETE` handler to `/federation_resolvers`. Includes deprecation notices for `/federation_resolvers/{{ID}}`.

(Draft pending deprecation approval on the mailing list)

- Documentation
- Traffic Ops Client (Go)
- Traffic Ops

Run the existing `/federations_resolvers` API and unit tests.


traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -136,3 +151,99 @@ func Read(w http.ResponseWriter, r *http.Request) {
...
+				tc.Alert{
+					Level: tc.WarnLevel.String(),
+					Text: "This endpoint is deprecated, please use '/federation_resolvers' instead",
+				},
+			},
+		},
+		&respObj,
+	}
+
+	if inf.Version.Major < 2 && inf.Version.Minor < 4 {
```
ChrisHines: Do we care about `Major == 0` here? Would this line be more clear if it checked for `Major == 1` instead?

ChrisHines:
I almost forgot... tests please.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -136,3 +151,99 @@ func Read(w http.ResponseWriter, r *http.Request) {
...
+
+	api.WriteRespAlertObj(w, r, tc.SuccessLevel, alert.Text, respObj)
+}
+
+func delete(inf *api.APIInfo) (error, error, int, tc.Alert, tc.FederationResolver) {
+	var userErr error
+	var sysErr error
+	var statusCode = http.StatusOK
+	var alert tc.Alert
+	var result tc.FederationResolver
```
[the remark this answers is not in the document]
ocket8888: named return values have been shot down at a project level

ocket8888:
There already is a test in the client integration test suite. It was added in an earlier PR because the testing fixtures needed to be created and deleted even though I had only rewritten creation to Go at the time.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -136,3 +151,99 @@ func Read(w http.ResponseWriter, r *http.Request) {
...
+			userErr = fmt.Errorf("No federation resolver by ID %d", inf.IntParams["id"])
+			statusCode = http.StatusNotFound
+		} else {
+			userErr, sysErr, statusCode = api.ParseDBError(err)
+		}
+		return userErr, sysErr, statusCode, alert, result
+	}
+
+	changeLogMsg := fmt.Sprintf("FEDERATION_RESOLVER: %s, ID: %d, ACTION: Deleted", *result.IPAddress, *result.ID)
+	api.CreateChangeLogRawTx(api.ApiChange, changeLogMsg, inf.User, inf.Tx.Tx)
```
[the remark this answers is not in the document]
ocket8888: A while ago someone went through the TO Go codebase and made all of our changelogs conform to a specific format - which is what I have there. I think it would've been much better to edit the now-unused `api.CreateChangeLogBuildMsg` but such is life.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -136,3 +151,99 @@ func Read(w http.ResponseWriter, r *http.Request) {
...
+
+	userErr, sysErr, statusCode, alert, respObj := delete(inf)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, tx, statusCode, userErr, sysErr)
+		return
+	}
+
+	var resp = struct {
+		tc.Alerts
+		Response *tc.FederationResolver `json:"response,omitempty"`
```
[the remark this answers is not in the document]
ocket8888: `go vet` isn't enforced afaik, but even though that change wasn't in this PR it was something I did previously and really ought to be fixed.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -136,3 +151,99 @@ func Read(w http.ResponseWriter, r *http.Request) {
...
+	userErr, sysErr, statusCode, alert, respObj := delete(inf)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, tx, statusCode, userErr, sysErr)
+		return
+	}
+
+	api.WriteRespAlertObj(w, r, tc.SuccessLevel, alert.Text, respObj)
+}
+
+func delete(inf *api.APIInfo) (error, error, int, tc.Alert, tc.FederationResolver) {
```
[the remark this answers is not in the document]
ocket8888: You mean change it so everything is before the errors or just the alerts and resolver? I'd rather not change the return order of "User Error, System Error, Response Code" because it's semi-standard in our code base.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
ocket8888: I guess that'd be fine. I just get paranoid thinking about edge cases, but the endpoint is hardcoded to only reach the handler at a version >= 1.1

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -49,6 +52,18 @@ FROM federation_resolver
...
+DELETE FROM federation_resolver
+WHERE federation_resolver.id = $1
+RETURNING federation_resolver.id,
+          federation_resolver.ip_address,
+          (
+          	SELECT type.name
+          	FROM type
+          	WHERE type.id = federation_resolver.type
+          ) AS type
+`
```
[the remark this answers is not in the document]
ocket8888: Idk, I think it's standard to put constants and global variables at the top then subroutines below that, isn't it? I couldn't find a source for Go best-practices on the matter. I could add comments to them although they're un-exported if you think that'd help?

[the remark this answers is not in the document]
ChrisHines: I don't believe there is a conventional placement for constants and global variables in Go. Go is somewhat unique in that the order of declaration of top level identifiers mostly doesn't matter. The compiler determines the order to initialize variables and constants based on lexical dependencies (see: https://golang.org/ref/spec#Package_initialization). So if one depends on the other it can still appear first in a .go file and the compiler does the right thing.

This property of Go frees us from needing to declare things before using them at package scope and allows us to place them where it is most convenient for readability.

In my opinion, code is more readable when values are easy to find when reading the code that uses them. I also think code is more readable when identifiers are declared in the narrowest possible scope. Minimizing the portion of the code that has access to a value helps a maintainer know the possible impact of changing it.

In this case, if a future maintainer jumps into this code and sees the query string they will wonder where is it used and if they start with the function that runs the query they will wonder what the SQL looks like. It would be nice to have the two next to each other.

I am sure opinions vary on this subject, and that's OK, but those are the reasons behind my suggestion here. Feel free to take it or leave it as you see fit.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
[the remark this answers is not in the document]
ChrisHines: I was suggesting to change it to follow the TO convention of the user error, sys error, and response code being the last three returned values and putting the other two return values at the beginning.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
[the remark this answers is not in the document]
ChrisHines: Are you saying that `api.CreateChangeLogBuildMsg` produces the wrong format? If so, should it be updated or deleted? Also, if it were updated then all code using it would also be fixed, including this code here if it were using it. It seems to me that using separate calls to `fmt.Sprintf` all over the code is a poor way to ensure the formatting these are consistent.

That said, it is just a code quality suggestion, not necessarily a correctness issue, so do as you think best.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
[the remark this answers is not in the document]
ocket8888: I'd rather leave it. The layout of "global constants/variables first" matches the output format of GoDoc so I think that will make it easier for people to read, and also suggests that this is the manner that many find comfortable for reading it.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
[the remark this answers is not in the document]
ocket8888: ah, well I've done that.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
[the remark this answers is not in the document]
ocket8888: > Are you saying that `api.CreateChangeLogBuildMsg` produces the wrong format? If so, should it be updated or deleted?

Yes and yes. Preferably updated.

> It seems to me that using separate calls to fmt.Sprintf all over the code is a poor way to ensure the formatting these are consistent.

It is and that ought to be changed. I'd just rather do it in a separate PR when/if I find the time.

docs/source/api/federation_resolvers.rst
```diff
@@ -139,6 +139,11 @@ Response Structure
 :ipAddress: The IP address or :abbr:`CIDR (Classless Inter-Domain Routing)`-notation subnet of the resolver - may be IPv4 or IPv6
 :type:      The :term:`Type` of the resolver
 
+	.. versionadded:: 1.4
```
mhoppa: should this change as 1.4 was already cut.

Also out of pure curiosity is there a reason why this was added?

docs/source/api/federation_resolvers.rst
```diff
@@ -163,5 +168,72 @@ Response Structure
...
 		"ipAddress": "::1/1",
+		"type": "RESOLVE6",
+		"typeId": 37
+	}}
+
+``DELETE``
+==========
+Deletes a federation resolver.
+
+.. versionadded:: 1.4
```
mhoppa: same here

docs/source/api/federation_resolvers.rst
```diff
@@ -163,5 +168,72 @@ Response Structure
...
+	Content-Length: 161
+
+	{ "alerts": [
+		{
+			"text": "Federation resolver deleted [ IP = 1.2.6.4/22 ] with id: 4",
+			"level": "success"
+		}
+	],
+	"response": {
+		"id": 4,
```
mhoppa: should this be a list? I know the id is required on the path so it will only be one atm but I think it is a better design moving forward so if we wanted to allow deletion of multiple resolvers at the same time the response structure doesnt need to change

docs/source/api/federation_resolvers_id.rst
```diff
@@ -18,14 +18,19 @@
 *******************************
 ``federation_resolvers/{{ID}}``
 *******************************
+.. deprecated:: 1.4
```
mhoppa: same here

docs/source/api/federation_resolvers_id.rst
```diff
@@ -18,14 +18,19 @@
...
 ``DELETE``
 ==========
 Deletes a federation resolver.
 
 :Auth. Required: Yes
 :Roles Required: "admin"
-:Response Type:  ``undefined``
+:Response Type:  Object
+
+	.. versionchanged:: 1.4
```
mhoppa: and here

lib/go-tc/federation_resolver.go
```diff
@@ -27,8 +27,8 @@ import "github.com/go-ozzo/ozzo-validation"
 
 // FederationResolver represents a resolver record for a CDN Federation.
 type FederationResolver struct {
-	ID          *uint      `json:"id", db:"id"`
-	IPAddress   *string    `json:"ipAddress" db:"ip_address`
+	ID          *uint      `json:"id" db:"id"`
```
mhoppa: huh good catch

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -136,3 +155,101 @@ func Read(w http.ResponseWriter, r *http.Request) {
...
+	var resp = struct {
+		tc.Alerts
+		Response *tc.FederationResolver `json:"response,omitempty"`
+	}{
+		tc.Alerts{
+			Alerts: []tc.Alert{
+				alert,
+				tc.Alert{
+					Level: tc.WarnLevel.String(),
+					Text:  "This endpoint is deprecated, please use '/federation_resolvers' instead",
```
mhoppa: nit but should this be `use '/federation_resolvers?id={id}' instead `

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -136,3 +155,101 @@ func Read(w http.ResponseWriter, r *http.Request) {
...
+		&respObj,
+	}
+
+	if inf.Version.Major == 1 && inf.Version.Minor < 4 {
+		resp.Response = nil
+	}
+
+	respBts, err := json.Marshal(resp)
+	if err != nil {
+		sysErr = fmt.Errorf("marhsaling response: %v", err)
```
mhoppa: `marshaling`, also isnt there a helper that does this ?

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -532,6 +532,8 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `federation_resolvers(/|\.json)?$`, federation_resolvers.Read, auth.PrivLevelReadOnly, Authenticated, nil, 556608759, perlBypass},
 		{1.1, http.MethodPost, `federations/{id}/federation_resolvers(/|\.json)?$`, federations.AssignFederationResolversToFederationHandler, auth.PrivLevelAdmin, Authenticated, nil, 556608760, perlBypass},
 		{1.1, http.MethodGet, `federations/{id}/federation_resolvers(/|\.json)?$`, federations.GetFederationFederationResolversHandler, auth.PrivLevelReadOnly, Authenticated, nil, 556608761, perlBypass},
+		{1.1, http.MethodDelete, `federation_resolvers(/|\.json)?$`, federation_resolvers.Delete, auth.PrivLevelAdmin, Authenticated, nil, 9001, perlBypass},
```
mhoppa: should this be a new endpoint 1.5?

docs/source/api/federation_resolvers.rst
ocket8888: To make it consistent with `GET` response bodies. Now it returns _extra_ information, but not _different_ information.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
ocket8888: Well it's not *wrong*, but I can certainly add that.

docs/source/api/federation_resolvers.rst
ocket8888: Is a 1.5 upgrade added yet, or do I have to do that in this PR too? I know there's a PR that would bump the version, dunno if it was merged yet.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
ocket8888: I think there is, now, there wasn't at the time.

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
mhoppa: not wrong at all and completely unblocking comment so take it or leave it :)

docs/source/api/federation_resolvers.rst
ocket8888: Yes, it ought to be

docs/source/api/federation_resolvers.rst
ocket8888: Actually, in this case, this message is correct. This is me adding to the docs for an already-rewritten endpoint. `/federation_resolvers` does this on master; i.e. in 1.4

mhoppa: ahhh good catch!

traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
ocket8888: Oh, actually there isn't. That helper takes an object as the response and puts it in a struct like:
```go
response := struct {
	Response interface{} `json:"response"`
}
```
Note that there's no `omitempty`. I needed it to not return a `response` object at all for API versions < 1.5, to be Perl-compatible.

docs/source/api/federation_resolvers.rst
ocket8888: Wait I thought this was the `GET` handler docs. It's currently pretty standard for creation/update/deletion handlers to return single objects, e.g. [`/asns`](https://traffic-control-cdn.readthedocs.io/en/latest/api/asns.html#post). In fact I think this'd be the only endpoint to break that convention if I changed it.

mhoppa: hmm how many of those are id paths? unsure how much value of moving from `DELETE /federation_resolvers/{id}` to `DELETE /federation_resolvers` would provide then

ocket8888: Well none of the creation endpoints use path parameters (for the things the create), but for updates/deletions generally both - where supported - will return just a single object (e.g. [`/jobs`](https://traffic-control-cdn.readthedocs.io/en/latest/api/jobs.html#delete)). Those won't change for API 1.0 and probably shouldn't for 2.0.

mhoppa:
Coming back to this finally, what did we decide on this? Are we okay deprecating DELETE /federation_resolvers/{{ID}?

ocket8888:
Well, it's [on the list](https://lists.apache.org/list.html?dev@trafficcontrol.apache.org:2019-12) so I'd say "yes". The more `docs/source/api/*.rst` files I get to delete the better, IMO.

mhoppa:
okay 🤷‍♂

mhoppa:
```
--- FAIL: TestFederationFederationResolvers (6.60s)
    federation_resolvers_test.go:148: Received expected success creating federation resolver: Federation Resolver created [ IP = 1.2.3.4 ] with id: 13
    federation_resolvers_test.go:148: Received expected success creating federation resolver: Federation Resolver created [ IP = 0.0.0.0/12 ] with id: 14
    federation_resolvers_test.go:148: Received expected success creating federation resolver: Federation Resolver created [ IP = dead::babe ] with id: 15
    federation_resolvers_test.go:148: Received expected success creating federation resolver: Federation Resolver created [ IP = ::f1d0:f00d/123 ] with id: 16
    --- PASS: TestFederationFederationResolvers/Successfully_assign_one_federation_resolver_to_a_federation (0.02s)
    --- PASS: TestFederationFederationResolvers/Successfully_assign_multiple_federation_resolver_to_a_federation (0.02s)
    --- PASS: TestFederationFederationResolvers/Successfully_replace_all_federation_resolver_for_a_federation (0.02s)
    --- PASS: TestFederationFederationResolvers/Fail_to_assign_federation_resolver_to_a_federation_when_federation_does_not_exist (0.01s)
    --- PASS: TestFederationFederationResolvers/successfully_get_federation_federation_resolvers_for_a_federation_with_some (0.01s)
    --- PASS: TestFederationFederationResolvers/successfully_get_federation_federation_resolvers_for_a_federation_without_any (0.03s)
    federation_resolvers_test.go:195: failed to delete Federation Resolver {ID:0xc0004ee1d8 IPAddress:0xc000245190 LastUpdated:0001-01-01 00:00:00 +0000 UTC Type:0xc0002451a0 TypeID:<nil>}: 404 Not Found[404] - Error requesting Traffic Ops https://localhost:6443/api/1.4/federation_resolvers?id=13 {"alerts":[{"level":"error","text":"Resource not found."}]}
                alerts: {Alerts:[]}
```

mhoppa:
API tests are failing for me

ocket8888:
This stumped me for longer than I care to admit. The problem is that the test client is request the 1.4 API while the route is only defined in the 1.5 API.

So here's the conundrum: this package is `github.com/apache/trafficcontrol/traffic_ops/testing/api/v14` and is meant to test the 1.4 API. I want to have tests for this endpoint; do I need to create a 1.5 test suite? Or alternatively, do our API/integration tests need to be versioned like this? After all, we don't have a `github.com/apache/trafficcontrol/traffic_ops/testing/api/v13` package. I'm not totally clear on the procedure here, or where I can look for guidance.

mhoppa:
great catch! I would personally vote for removing the version on the testing package package so would be `github.com/apache/trafficcontrol/traffic_ops/testing/api` although unsure what impact that will have when we have version 2.0/3.0 supported as those test suites will look different I would think

ocket8888:
That test is now passing. I moved all tests from `v14` to `v1` and changed the client's `apiBase` to version 1.5.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -532,6 +532,8 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `federation_resolvers(/|\.json)?$`, federation_resolvers.Read, auth.PrivLevelReadOnly, Authenticated, nil, 556608759, perlBypass},
 		{1.1, http.MethodPost, `federations/{id}/federation_resolvers(/|\.json)?$`, federations.AssignFederationResolversToFederationHandler, auth.PrivLevelAdmin, Authenticated, nil, 556608760, perlBypass},
 		{1.1, http.MethodGet, `federations/{id}/federation_resolvers(/|\.json)?$`, federations.GetFederationFederationResolversHandler, auth.PrivLevelReadOnly, Authenticated, nil, 556608761, perlBypass},
+		{1.5, http.MethodDelete, `federation_resolvers(/|\.json)?$`, federation_resolvers.Delete, auth.PrivLevelAdmin, Authenticated, nil, 9001, perlBypass},
```
rawlinp: If this route didn't exist in Perl it should be `noPerlBypass`

mhoppa: great catch

ocket8888: oh yeah, that makes sense.

PR #4095, opened 2019-11-07
Clean up Experimental code

mhoppa:
## What does this PR (Pull Request) do?

PR goes through and cleans up experimental code that was underneath traffic_ops/experimental.

It removes code that is dates/superseded and moves the rest to underneath the top level experimental directory.

### Directories being removed
traffic_ops/experimental/ats_config
traffic_ops/experimental/server
traffic_ops/experimental/goto
traffic_ops/experimental/auth

### Directories being moved 
traffic_ops/experimental/traffic_ops_auth
traffic_ops/experimental/url-rewriter-nginx
traffic_ops/experimental/postgrest
traffic_ops/experimental/webfront

## What is the best way to verify this PR?

Make sure unit tests, rpm build, docker build for ciab and api tests pass

## The following criteria are ALL met by this PR

- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR ensures that database migration sequence is correct OR this PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #4106, opened 2019-11-11
Add top level golang linter and golang unit test

mhoppa:

Adds top level dockerfile/docker compose file/Makefile to run golangci-lint and go test on the repo as a whole.

Had to fix up some of the repo for golangci-lint to actually run as some of the code was in a broken state. Also ran go fmt on the repo as a whole.

`make lint` - to run linter
`make unit` - to run unit tests

## Additional Information

# DEPENDENT ON https://github.com/apache/trafficcontrol/pull/4095 

Leverages the golang linter -> https://github.com/golangci/golangci-lint


.golangci.yml
```diff
@@ -0,0 +1,43 @@
...
+  enable:
+    - govet
+    - errcheck
+    - unused
+    - structcheck
+    - varcheck
+    - deadcode
+    - typecheck
+    - golint
+    - gofmt
```
ChrisHines: I'm not sure this is the best list of linters. In particular I would strongly encourage the use of staticcheck and take away unused (it was subsumed into staticcheck a while ago)? Also I suspect staticcheck does a better job at the checks that varcheck, deadcode, typecheck are trying to do.

ChrisHines: Here is the list of what staticcheck checks for: https://staticcheck.io/docs/checks

mhoppa: Sounds great to me! Thanks for the input. Let me make that change and test it locally

Makefile
```diff
@@ -0,0 +1,7 @@
+lint:
```
ocket8888: missing license header

targets that aren't produced by their recipes ought to be labelled as such with [the special `.PHONY` target](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html)

Makefile
```diff
@@ -0,0 +1,7 @@
...
+	@echo ? Running golangci-lint
+	@docker-compose -f tools/golang/docker-compose.yml up lint
```
ocket8888: I'm not sure what the `?` in the echo call is supposed to do (except print a <kbd>?</kbd>?) but it's maybe irrelevant because it looks like all you're doing is printing what's about to happen. Which is exactly what `make` would do by default if you got rid of your `echo`s and didn't preface your steps with `@`.

experimental/traffic_router_golang/crconfigdsservers/crconfigdsservers.go
```diff
@@ -28,8 +32,8 @@ type DSServer struct {
...
-type DSServer struct {
-	DirectMatches                      map[string]tc.DeliveryServiceName
-	DotStartSlashDotFooSlashDotDotStar map[string]tc.DeliveryServiceName
-	RegexMatch                         map[*regexp.Regexp]tc.DeliveryServiceName
-}
+// type DSServer struct {
+// 	DirectMatches                      map[string]tc.DeliveryServiceName
+// 	DotStartSlashDotFooSlashDotDotStar map[string]tc.DeliveryServiceName
+// 	RegexMatch                         map[*regexp.Regexp]tc.DeliveryServiceName
+// }
```
ocket8888: Can this be deleted instead?

experimental/traffic_router_golang/crstatespoller/crstatespoller.go
```diff
@@ -65,7 +65,7 @@ func Start(fetcher fetch.Fetcher, interval time.Duration, crc crconfig.Ths) (crs
 	thsCrs := crstates.NewThs()
 	availableServers := availableservers.New()
 	prevBts := []byte{}
-	prevCrs := (*tc.CRStates)(nil)
+	// prevCrs := (*tc.CRStates)(nil)
```
ocket8888: can this line be deleted?

experimental/traffic_router_golang/crstatespoller/crstatespoller.go
```diff
@@ -88,7 +88,7 @@ func Start(fetcher fetch.Fetcher, interval time.Duration, crc crconfig.Ths) (crs
 
 		thsCrs.Set(crs)
 		prevBts = newBts
-		prevCrs = crs
+		// prevCrs = crs
```
ocket8888: can this line be deleted?

test/router/dnssec/dnssec_suite_test.go
```diff
@@ -19,25 +19,24 @@ package dnssec_test
...
 
 var d *dnssec.DnssecClient
 var nameserver string
 var deliveryService string
 
 func init() {
-	flag.StringVar(&nameserver,"ns","changeit","ns is used to direct dns queries to a traffic router")
-	flag.StringVar(&deliveryService,"ds","changeit","ds is used to target some dns DS and DNS queries made by traffic router")
+	flag.StringVar(&nameserver, "ns", "changeit", "ns is used to direct dns queries to a traffic router")
+	flag.StringVar(&deliveryService, "ds", "changeit", "ds is used to target some dns DS and DNS queries made by traffic router")
```
ocket8888: > default option-argument is "changeit"

yikes.

Nothing you need to change just... yikes.

tools/golang/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN go get -v ./...
+
+FROM base AS lint
+
+RUN apk --no-cache add curl
+
+RUN curl -L https://github.com/golangci/golangci-lint/releases/download/v1.21.0/golangci-lint-1.21.0-linux-amd64.tar.gz \
+	| tar -zxC /tmp \
+	&& mv /tmp/golangci-lint-1.21.0-linux-amd64/golangci-lint /usr/local/bin/golangci-lint \
+	&& rm -rf /tmp/golangci-lint-1.21.0-linux-amd64
```
ocket8888: surely this can be done with `go get`?

tools/golang/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+CMD ["golangci-lint", "run", "./..."]
+
+FROM base AS unit
+
+CMD ["go", "test", "./..."]
```
ocket8888: missing newline at EOF

traffic_ops/testing/api/v14/withobjs_test.go
```diff
@@ -0,0 +1,109 @@
+/*
```
ocket8888: I think you meant to move `withobjs.go` to this new filename, but I don't see the old one deleted?

mhoppa:
@ocket8888 no we do not need docker to run the linter. It can be simply downloaded and be run at the root via ‘’’golangci-lint run ./...’’’ 

It will automatically pick up the same config. 

Docker was more there for future ci integration.

traffic_ops/testing/api/v14/withobjs_test.go
mhoppa: Oops!

tools/golang/Dockerfile
mhoppa: Probably do you see a problem with this approach

ocket8888: it just adds a layer and a tool

Makefile
mhoppa: Yeah the ? was more of just breaking out the stdout from the target and the start. 

```COSML-1547398:trafficcontrol mhoppa509$ make lint
? Running golangci-lint
Starting golang_lint_1 ... done
Attaching to golang_lint_1
```

vs 

```
COSML-1547398:trafficcontrol mhoppa509$ make lint
docker-compose -f tools/golang/docker-compose.yml up lint
Starting golang_lint_1 ... 
```
And yes if you omit the @ it would print out the steps but its verbatim instead of something as `Running golangci-lint` but I can change it if you are really against it.

TBH I am not very picky on Makefile format.

tools/golang/Dockerfile
mhoppa: the tool needs to be added regardless and I wanted it as part of the lint stage not the base stage as unit tests does not need it. I think a layer is still needed

ChrisHines: `go get` could probably be made to work, but there could be challenges keeping it stable. I believe we would prefer to only use tagged releases of golangci-lint. Despite the fact that golangci-lint has adopted Go modules and appears to do a good job tagging releases, using `go get` to retrieve them isn't always straightforward. Go 1.13 has made that easier, but this project is still using Go 1.11, which doesn't have the best support for this use case.

Makefile
ocket8888: I personally think the default behavior of printing exactly what it's doing is best, but I see what you're getting at. If that's necessary or even just beneficial for CI reasons, I'd suggest just pre-pending the `?` to the command that's actually being run rather than abbreviating it. Something like:

```make
lint:
	@echo -n '? '
	docker-compose -f tools/golang/docker-compose.yml up lint
```
which will output
```console
$ make
? docker-compose -f tools/golang/docker-compose.yml up lint
Starting golang_lint_1 ... done
Attaching to golang_lint_1
...
```
It's _possible_ that'll still mess with tools that invoke `make` and parse the output, but probably not likely. And also probably not even a big deal if it does either.

tools/golang/Dockerfile
ocket8888: well hey if we need it later anyway that's good enough for me

.golangci.yml
```diff
@@ -0,0 +1,38 @@
...
+
+linters:
+  disable-all: true
+  enable:
+    - govet
+    - errcheck
+    - structcheck
+    - golint
+    - gofmt
+    - staticcheck
```
ocket8888: Is there a way to configure these linters? What do they all check?

traffic_ops/testing/compare/compare.go
```diff
@@ -266,11 +266,11 @@ func handlePlainTextResponse(responses *[]result, route string) {
...
-		                result0Str,
-		                (*responses)[0].TO,
-		                result1Str,
-		                (*responses)[1].TO,
-		                checkOrderDiffs(scrubbedResult0, scrubbedResult1))
+			result0Str,
+			(*responses)[0].TO,
+			result1Str,
+			(*responses)[1].TO,
+			checkOrderDiffs(scrubbedResult0, scrubbedResult1))
```
ocket8888: alignment is bad?

mhoppa: according to gofmt :)

.golangci.yml
mhoppa: we can configure them in this file under the key `linters-settings` but I thought leaving defaults would be a first good start.

What they check can be seen at https://github.com/golangci/golangci-lint#supported-linters or list a small description and link to the linter itself

traffic_ops/testing/compare/compare.go
ocket8888: I really hate that, but oh well I guess.

ChrisHines: Consider the following gofmt friendly layout:

```
writeAllResults(
    route,
    ...
    checkOrderDiffs(...),
)
```

ocket8888: Yeah I've used that before. Still prefer the original.

Makefile
```diff
@@ -0,0 +1,26 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+.PHONY: lint unit
+
+lint:
+	@echo -n '? '
+	docker-compose -f tools/golang/docker-compose.yml up lint
```
ocket8888: If this is meant to be used in CI systems, you might want to provide at least the option of turning off all those muddying ANSI characters that get printed for things like e.g. coloring output. That's easily achieved by adding `--no-ansi` to the `docker-compose` commands, but the way Makefiles normally do stuff like that is by setting an opts variable that can be overridden on the command line. So something like:
```make
DC_OPTS ?= "--no-ansi"

.PHONY: lint

lint:
	@echo -n '? '
	docker-compose $(DC_OPTS) -f tools/golang/docker-compose.yml up lint

```
unless you think unprintable characters won't pose an issue? I think you're more familiar with Jenkins than I am.

mhoppa: I dont think it would pose an issue but I dont see any issue or negative side affects with adding that option in! so added

Issue #3397, opened 2019-03-11
Traffic Router doesn't refresh Cookie


smalenfant:
We upgraded Traffic Ops 2.2 to 3.0 and Traffic Router (not version specific) could not get at least 2 targets (steering.json and federation.json).

Traffic Router would need to handle cookie expiration (or change) and re-attempt logging in.

Workaround is to restart Traffic Router at this time. You have to be looking at the logs for those 401 or you could run into conditions where Traffic Router won't get updated steering/federations updates.

PR #4119, opened 2019-11-15
TO: Consistently set Expires and Max-Age cookie headers

rawlinp:

Include the Expires and Max-Age cookie headers in the Set-Cookie
response header for:
- new cookie generation via all login/logout endpoints
- cookie "refreshing" via authenticated endpoints

Per https://tools.ietf.org/html/rfc6265 the Max-Age header should have
precedence when both headers are present, so including both cookie
headers should not be a problem.

Possibly related to #3397 since TR doesn't know what to do with auth cookies if they don't have an `Expires` or `Max-Age` header.

This is technically a regression from the Perl from what I can tell, because TO-Perl includes the `Expires` cookie header in login as well as non-login responses (TO-Go was only returning the `Expires` cookie header for login responses).

- Traffic Ops

Run the TO API tests and the traffic_ops_golang unit tests.
Verify TP is still able to login/use the cookie through a few different pages in a few different browsers.

- since 3.x

- [x] Bugfix, no docs necessary


rawlinp:
Oh, good call. I forgot stuff like `Set-Cookie` headers are included in the examples. Maybe I should update _all_ the docs examples. I should probably also add a changelog since this regressed in the previous release.

docs/source/admin/quick_howto/federations.rst
```diff
@@ -59,7 +59,7 @@ Configure Federations
 		Access-Control-Allow-Credentials: true
 		Content-Type: application/json
 		Access-Control-Allow-Methods: POST,GET,OPTIONS,PUT,DELETE
-		Set-Cookie: mojolicious=eyJleHBpcmVzIjoxNDQ5MTA1MTI2LCJhdXRoX2RhdGEiOiJmZWRlcmF0aW9uX3VzZXIxIn0---06b4f870d809d82a91433e92eae8320875c3e8b0; expires=Thu, 03 Dec 2015 01:12:06 GMT; path=/; HttpOnly
+		Set-Cookie: mojolicious=...; Path=/; Expires=Mon, 18 Nov 2019 17:40:54 GMT; Max-Age=3600; HttpOnly
```
ocket8888: Good catch :+1:

PR #4125, opened 2019-11-18
Rewrite users/register to Go

### LINKED ISSUE #3835: Rewrite /users/register to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/users/register` is handled by Perl

## Expected / new behavior:
`/users/register` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `POST`

N/A

ocket8888:
- [x] This PR fixes #3835

Rewrites `/users/register` to Go. As opposed to the Perl, in this PR:

- `tenantID` is required, as tenancy is no longer optional.
- `tenantID` can no longer be a tenant to which the requesting user has no access
- `role` can no longer be a Role with a higher privilege level than the requesting user
- when new users are created, instead of using their login token as a username (?!?!?) a 32-character random string is generated and then appended to `registration_`.

- Documentation
- Traffic Ops Client (Go)
- Traffic Ops

Start up Traffic Ops from this branch in an environment with access to a mail server (which is properly configured in the TO configuration file) and attempt to register users using `/users/register`.

There are no unit or integration tests because this requires access to a mail server, which isn't built into the testing. There is, however, a test to ensure the email template renders given adequate input.

- [x] I have explained why tests are impossible/impractical

### DIFF
--- a/lib/go-tc/users.go
+++ b/lib/go-tc/users.go
@@ -155,7 +155,7 @@ type UserRegistrationRequest struct {
 
 // Validate implements the
 // github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api.ParseValidator interface.
-func (urr UserRegistrationRequest) Validate(tx *sql.Tx) error {
+func (urr *UserRegistrationRequest) Validate(tx *sql.Tx) error {
 	var errs = []error{}
 	if urr.Role == 0 {
 		errs = append(errs, errors.New("role: required and cannot be zero."))
--- a/lib/go-tc/users.go
+++ b/lib/go-tc/users.go
@@ -19,7 +19,10 @@ package tc
  * under the License.
  */
 
+import "database/sql"
+
 import "github.com/apache/trafficcontrol/lib/go-rfc"
+import "github.com/apache/trafficcontrol/lib/go-util"
 
 // UserCredentials contains Traffic Ops login credentials
 type UserCredentials struct {
@@ -138,4 +141,28 @@ type UserDeliveryServiceDeleteResponse struct {
 
 type UserPasswordResetRequest struct {
 	Email rfc.EmailAddress `json:"email"`
+}
+
+// UserRegistrationRequest is the request submitted by operators when they want to register a new
+// user.
+type UserRegistrationRequest struct {
+	Email rfc.EmailAddress `json:"email"`
+	// Role - despite being named "Role" - is actually merely the *ID* of a Role to give the new user.
+	Role uint `json:"role"`
+	TenantID uint `json:"tenantId"`
+}
+
+// Validate implements the
+// github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api.ParseValidator interface.
+func (urr *UserRegistrationRequest) Validate (tx *sql.Tx) (error) {
+	var errs = []error{}
+	if urr.Role == 0 {
+		errs = append(errs, errors.New("role: required and cannot be zero."))
+	}
+
+	if urr.TenantID == 0 {
+		errs = append(errs, errors.New("role: required and cannot be zero."))
+	}
+
+	return util.JoinErrs(errs)
 }
--- a/lib/go-tc/users.go
+++ b/lib/go-tc/users.go
@@ -93,7 +93,7 @@ type User struct {
 	Username         *string    `json:"username" db:"username"`
 	RegistrationSent *TimeNoMod `json:"registrationSent" db:"registration_sent"`
 	LocalPassword    *string    `json:"localPasswd,omitempty" db:"local_passwd"`
-	RoleName         *string    `json:"roleName,omitempty" db:"-"`
+	RoleName         *string    `json:"roleName,omitempty" db:"rolename"`
 	commonUserFields
 }
 
@@ -163,6 +163,12 @@ func (urr *UserRegistrationRequest) Validate(tx *sql.Tx) error {
 
 	if urr.TenantID == 0 {
 		errs = append(errs, errors.New("tenantId: required and cannot be zero."))
+	}
+
+	// This can only happen if an email isn't present in the request; the JSON parse handles actually
+	// invalid email addresses.
+	if Email.Address.Address == "" {
+		errs = append(errs, errors.New("email: required"))
 	}
 
 	return util.JoinErrs(errs)
--- a/lib/go-tc/users.go
+++ b/lib/go-tc/users.go
@@ -20,6 +20,7 @@ package tc
  */
 
 import "database/sql"
+import "errors"
 
 import "github.com/apache/trafficcontrol/lib/go-rfc"
 import "github.com/apache/trafficcontrol/lib/go-util"
@@ -154,7 +155,7 @@ type UserRegistrationRequest struct {
 
 // Validate implements the
 // github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api.ParseValidator interface.
-func (urr *UserRegistrationRequest) Validate (tx *sql.Tx) (error) {
+func (urr UserRegistrationRequest) Validate (tx *sql.Tx) (error) {
 	var errs = []error{}
 	if urr.Role == 0 {
 		errs = append(errs, errors.New("role: required and cannot be zero."))
--- a/lib/go-tc/users.go
+++ b/lib/go-tc/users.go
@@ -149,13 +149,13 @@ type UserPasswordResetRequest struct {
 type UserRegistrationRequest struct {
 	Email rfc.EmailAddress `json:"email"`
 	// Role - despite being named "Role" - is actually merely the *ID* of a Role to give the new user.
-	Role uint `json:"role"`
+	Role     uint `json:"role"`
 	TenantID uint `json:"tenantId"`
 }
 
 // Validate implements the
 // github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api.ParseValidator interface.
-func (urr UserRegistrationRequest) Validate (tx *sql.Tx) (error) {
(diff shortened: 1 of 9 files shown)


lib/go-tc/users.go
```diff
@@ -139,3 +143,27 @@ type UserDeliveryServiceDeleteResponse struct {
...
+// Validate implements the
+// github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api.ParseValidator interface.
+func (urr UserRegistrationRequest) Validate (tx *sql.Tx) (error) {
+	var errs = []error{}
+	if urr.Role == 0 {
+		errs = append(errs, errors.New("role: required and cannot be zero."))
+	}
+
+	if urr.TenantID == 0 {
+		errs = append(errs, errors.New("role: required and cannot be zero."))
```
awilliams: Error string should reference `tenantId`

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -62,6 +63,33 @@ WHERE federation_deliveryservice.deliveryservice IN (
 )
 `
 
+const getUserByEmailQuery = `
+SELECT u.id,
+       u.username as username,
```
awilliams: What's the purpose of this alias?

traffic_ops/traffic_ops_golang/login/register.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	var tmpl bytes.Buffer
+	if err := registrationEmailTemplate.Execute(&tmpl, &f); err != nil {
+		return nil, err
+	}
+	return tmpl.Bytes(), nil
+}
+
+func RegisterUser (w http.ResponseWriter, r *http.Request) {
```
awilliams: Consider documenting this function.

traffic_ops/traffic_ops_golang/login/register.go
```diff
@@ -0,0 +1,280 @@
...
+	defer r.Body.Close()
+
+	var req tc.UserRegistrationRequest
+	if userErr = api.Parse(r.Body, tx, req); userErr != nil {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, userErr, nil)
+		return
+	}
+
+	if ok, err := inf.IsResourceAuthorizedToCurrentUser(int(req.TenantID)); err != nil {
+		sysErr = fmt.Errorf("Checking tenancy permissions of current user (%+v) on tenant #%d", inf.User, req.TenantID)
```
awilliams: Is there any concern that the `%+v` of the User object will expose any unintended data?

lib/go-tc/users.go
```diff
@@ -139,3 +143,27 @@ type UserDeliveryServiceDeleteResponse struct {
 type UserPasswordResetRequest struct {
 	Email rfc.EmailAddress `json:"email"`
 }
+
+// UserRegistrationRequest is the request submitted by operators when they want to register a new
+// user.
+type UserRegistrationRequest struct {
+	Email rfc.EmailAddress `json:"email"`
+	// Role - despite being named "Role" - is actually merely the *ID* of a Role to give the new user.
+	Role uint `json:"role"`
```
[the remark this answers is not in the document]
ocket8888: Maybe, maybe not. "RoleID" is certainly more appropriate, but "Role" is in keeping with the de facto standard of this_case in the database, ThisCase in struct fields, and thisCase in JSON.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
ocket8888: There isn't one.

traffic_ops/traffic_ops_golang/login/register.go
ocket8888: I don't think so, it's a system log, so you'd need access to the TO server to read it. It also doesn't include any sensitive information, the `auth.CurrentUser` struct just contains id/username and permissions information.

mitchell852:
i've always thought it was annoying that when you "register" a new user it autogenerates a username for them like 20A34788-A877-11E7...

the idea is that when the new user gets the register email and follows the link, they will update the username to something more sensible. because this username is random, it makes it hard to sort the users on https://tp.domain.net/#!/users and see which ones are "open/abandoned registrations".

what about changing that username autogenerate to something like change_me_20A34788-A877-11E7...

this would have 2 benefits:

1. you can sort/filter all the change_me's to get a sense of how many there are
2. it's clear to the user that they should update that field.

ocket8888:
That's a good point. It now generates a username like `register_{{a 32-character random string}}`

ocket8888:
Can you show me the full HTTP stack; request and response?

mhoppa:
I can let me get it back up

mhoppa:
```POST https://127.0.0.1:8443/api/1.4/users/register```

**Body** 

```
{
	"role":1,
	"tenantId":1,
	"email":"test@gmail.com"
}
```

**Response**
```
{
    "alerts": [
        {
            "text": "decoding: json: cannot unmarshal object into Go value of type api.ParseValidator",
            "level": "error"
        }
    ]
}
```

**Resp Headers** 
```
Content-Encoding: gzip
Content-Type:application/json
Set-Cookie: mojolicious=... Path=/; Expires=Fri, 22 Nov 2019 01:46:10 GMT; Max-Age=3600; HttpOnly
Whole-Content-Sha512: /E/Bj+NptZRI6Yp8wbGG49hxZ/3mVfbOzEPh+AbnLhlqRemI+nufZQCbKF4xmaOEee0tR8Tp+LeQpsu+Y23QqA==
Date: Fri, 22 Nov 2019 00:46:10 GMT
Content-Length: 126
```

Log message from TO

```
DEBUG: api.go:149: 2019-11-21T17:46:10.813635-07:00: decoding: json: cannot unmarshal object into Go value of type api.ParseValidator
```

**Comment by ocket8888** [comment]
I think I fixed that.

**Comment by ocket8888** [comment]
it was a problem with my clause ordering in a query. I know I tried it out manually, must've changed it after that and forgot to verify. I didn't bother setting up a mail server for it since that's very complicated and I can't remember how I did it that one time - so when I saw the Internal Server error I just assumed the email couldn't be sent so it "worked" :P

**Comment by ocket8888** [comment]
I think I fixed everything for real this time. There was also a bug where it'd let you not specify an email.

**Comment by mhoppa** [comment]
testing now 👍

File: lib/go-tc/users.go [hunk 6]
```diff
@@ -139,3 +143,33 @@ type UserDeliveryServiceDeleteResponse struct {
...
+		errs = append(errs, errors.New("role: required and cannot be zero."))
+	}
+
+	if urr.TenantID == 0 {
+		errs = append(errs, errors.New("tenantId: required and cannot be zero."))
+	}
+
+	// This can only happen if an email isn't present in the request; the JSON parse handles actually
+	// invalid email addresses.
+	if Email.Address.Address == "" {
```
mhoppa: should be `urr.Email.Address.Address`

```
# github.com/apache/trafficcontrol/lib/go-tc
../../lib/go-tc/users.go:170:5: undefined: Email
```

traffic_ops/traffic_ops_golang/login/register.go
```diff
@@ -0,0 +1,284 @@
...
+
+	msg, err := createRegistrationMsg(req.Email, t, tx, inf.Config.ConfigPortal)
+	if err != nil {
+		sysErr = fmt.Errorf("Failed to create email message: %v", err)
+		errCode = http.StatusInternalServerError
+		api.HandleErr(w, r, tx, errCode, nil, sysErr)
+		return
+	}
+
+	log.Debugf("Sending password reset email to %s", req.Email)
```
mhoppa: sending password register email?

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -229,6 +229,7 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.4, http.MethodPost, `user/login/oauth/?$`, login.OauthLoginHandler(d.DB, d.Config), 0, NoAuth, nil, 1415886009, noPerlBypass},
 		{1.1, http.MethodPost, `user/login/token(/|\.json)?$`, login.TokenLoginHandler(d.DB, d.Config), 0, NoAuth, nil, 402408841, perlBypass},
 		{1.1, http.MethodPost, `user/reset_password(/|\.json)?$`, login.ResetPassword(d.DB, d.Config), 0, NoAuth, nil, 2092914630, perlBypass},
+		{1.1, http.MethodPost, `users/register(/|\.json)?$`, login.RegisterUser, auth.PrivLevelOperations, Authenticated, nil, 1337, perlBypass},
```
mhoppa: should we allow bypass given ```Prior to ATC version 4.0, this endpoint allowed for the registration of users with arbitrary
		:term:`Roles`. It now restricts the allowed values to identifiers for :term:`Roles` with at
		most the same permissions level as the requesting user.```

lib/go-tc/users.go
ocket8888: :man_facepalming: jesus

traffic_ops/traffic_ops_golang/routing/routes.go
ocket8888: Well the default is that it won't bypass, and it's up to administrators to eat that risk if they want to bypass it. I raised a concern about that (and other concerns) with the security team and it was decided that these didn't merit "vulnerability" status, so :man_shrugging:

mhoppa: okie dookie 👍

PR #4126, opened 2019-11-19
TO: Rewrite /osversions in Go

### LINKED ISSUE #3800: Rewrite /osversions to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/osversions` is handled by Perl

## Expected / new behavior:
`/osversions` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`

N/A

awilliams:

- [x] This PR fixes #3800 by rewriting the `/osversions` endpoint in Go

- CDN in a Box
- Documentation
- Traffic Control Client: Go
- Traffic Ops

- Run the `TestGetOSVersions` test case in the `traffic_ops/testing/api/v14/` package
- Run the `traffic_ops/traffic_ops_golang/iso` unit tests

\-

## Additional Information

- This PR introduces a change to the `osversions` configuration file. Previously, a Perl formatted `osversions.cfg` file was used by the Perl application. This changes the file to be JSON formatted and named `osversions.json`. This was discussed on the mailing list under the [_Proposal for rewriting /osversions handler_](https://lists.apache.org/thread.html/3e9b1710f62b11ca056f399cfd194f15be9c4bbfad77d3f2cd5c105b@%3Cdev.trafficcontrol.apache.org%3E) thread.
- A script to convert from the `.cfg` to `.json` format has been provided at `traffic_ops/app/bin/osversions-convert.pl` (thanks @ocket8888).


awilliams:
One question I had was whether to leave the old `osversions.cfg` as part of the RPM, in addition to the new `osversions.json`?

ocket8888:
For the time being, yes. Rawlin's working on an API blacklist that would let operators fall back on Perl in case of problems with an upgrade, so it'll need to support that for probably a major revision after the rewrite is merged.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -204,6 +205,9 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodPost, `user/login/token(/|\.json)?$`, login.TokenLoginHandler(d.DB, d.Config), 0, NoAuth, nil},
 		{1.1, http.MethodPost, `user/reset_password(/|\.json)?$`, login.ResetPassword(d.DB, d.Config), 0, NoAuth, nil},
 
+		//ISO
+		{1.1, http.MethodGet, `osversions/?$`, iso.GetOSVersions, auth.PrivLevelReadOnly, Authenticated, nil},
```
ocket8888: this should also support a trailing `.json` since Perl did.

PR #4127, opened 2019-11-19
Rewrite GET stats_summary from Perl to Golang

### LINKED ISSUE #3832: Rewrite /stats_summary to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/stats_summary` is handled by Perl

## Expected / new behavior:
`/stats_summary` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`

N/A

mhoppa:

- [x] This PR fixes #3832

- Documentation
- Traffic Control Golang Client 
- Traffic Ops
- Traffic Stats

Run API tests (once added to PR) or pull down and manually test to ensure stats summary matches parity with perl implementation


traffic_ops/traffic_ops_golang/trafficstats/stats_summary.go
```diff
@@ -0,0 +1,127 @@
...
+func getLastSummaryDate(w http.ResponseWriter, r *http.Request, inf *api.APIInfo) {
+	queryParamsToSQLCols := map[string]dbhelpers.WhereColumnInfo{
+		"statName": dbhelpers.WhereColumnInfo{"stat_name", nil},
+	}
+	where, _, _, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(inf.Params, queryParamsToSQLCols)
+	if len(errs) > 0 {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, util.JoinErrs(errs))
+		return
+	}
+	query := selectQuery() + where + " ORDER BY summary_time DESC"
```
ocket8888: I can cause an internal server error in this endpoint by passing the `orderby` query parameter. It isn't listed as supported for this type of request to this endpoint, but typically unexpected/undocumented parameters have no effect and when they cause errors like this it can make it hard for the client to figure out what they did wrong. Plus that's technically a client error, because the poor form of their request caused the failure.

I think maybe `dbhelpers.BuildWhereAndOrderByAndPagination` might be overkill for this when you just wanna add an optional `WHERE` and a static `ORDER BY`. The way I see it, you have three options:

- Don't use a `dbhelpers` function and just build the query manually. That's probably what I would do.
- Write a new `dbhelpers` function for "`WHERE`-only" queries as a wrapper around `dbhelpers.BuildWhereAndOrderByAndPagination`. This'll be helpful if you think this problem can/will/should surface in other endpoints
- Manipulate `inf.Params` prior to calling `dbhelpers.BuildWhereAndOrderByAndPagination` such that it doesn't contain pagination or ordering query parameters. I wouldn't block the PR on this, but for my money it's a bad idea if only because it means that adding supported semantics to that function will break this handler in non-obvious ways that we may not catch right away.

traffic_ops/client/stats_summary.go
```diff
@@ -15,111 +15,69 @@ package client
...
 
-func (to *Session) GetSummaryStats(cdn string, deliveryService string, statName string) ([]tc.StatsSummary, ReqInf, error) {
-	var queryParams []string
-	if len(cdn) > 0 {
-		queryParams = append(queryParams, fmt.Sprintf("cdnName=%s", cdn))
-	}
-	if len(deliveryService) > 0 {
-		queryParams = append(queryParams, fmt.Sprintf("deliveryServiceName=%s", deliveryService))
+// GetSummaryStats gets a list of summary stats with the ability to filter on cdn,deliveryService and/or stat
+func (to *Session) GetSummaryStats(cdn, deliveryService, statName *string) (tc.StatsSummaryResponse, ReqInf, error) {
```
ocket8888: I don't know whether we're allowed to change the signature of client functions without a deprecation period. Removing that one above that was deprecated in 3.x should be fine, but I'm not sure about this one. Not that I don't like the improvements.

My knee-jerk reaction is to ask the mailing list about this.

traffic_ops/client/stats_summary.go
```diff
@@ -15,111 +15,69 @@ package client
...
-	var data tc.StatsSummaryResponse
-	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
-		return nil, reqInf, err
+	reqInf, err := get(to, url, &resp)
+	if err != nil {
+		return resp, reqInf, err
 	}
-
-	return data.Response, reqInf, nil
+	return resp, reqInf, nil
```
ocket8888: Why bother with this `if` block? You could just `return resp, reqInf, err`. If `err` is `nil` then it'll be `nil`; no need to check.

traffic_ops/client/stats_summary.go
```diff
@@ -15,111 +15,69 @@ package client
...
 }
 
-// SummaryStatsLastUpdated ...
-// Deprecated: use GetSummaryStatsLastUpdated
-func (to *Session) SummaryStatsLastUpdated(statName string) (string, error) {
-	s, _, err := to.GetSummaryStatsLastUpdated(statName)
-	return s, err
-}
+// GetSummaryStatsLastUpdated time of the last summary for a given stat
+func (to *Session) GetSummaryStatsLastUpdated(statName *string) (tc.StatsSummaryLastUpdatedResponse, ReqInf, error) {
```
ocket8888: same as above RE: changing the method signature

traffic_ops/client/stats_summary.go
```diff
@@ -15,111 +15,69 @@ package client
...
+		param.Add("statName", *statName)
 	}
+	url := fmt.Sprintf("%s?%s", basePathSummaryStats, param.Encode())
 
-	resp, remoteAddr, err := to.request("GET", queryURL, nil)
-	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
+	reqInf, err := get(to, url, &resp)
 	if err != nil {
-		return "", reqInf, err
+		return resp, reqInf, err
```
ocket8888: Same as above RE: seemingly unnecessary `if`

traffic_ops/testing/api/v14/stats_summary_test.go
```diff
@@ -0,0 +1,186 @@
...
+			ds:                     util.StrPtr("all"),
+			expectedStatsSummaries: testStatsSummaries,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			tsr, _, err := TOSession.GetSummaryStats(tc.cdn, tc.ds, tc.stat)
+			if err != nil {
+				t.Fatalf("received unexpectant error %v on GET to stats_summary", err)
```
ocket8888: *unexpected

traffic_ops/testing/api/v14/stats_summary_test.go
```diff
@@ -0,0 +1,186 @@
...
+			for _, ess := range tc.expectedStatsSummaries {
+				found := false
+				for _, ss := range tsr.Response {
+					if ess == ss {
+						found = true
+						break
+					}
+				}
+				if !found {
+					t.Fatalf("expected to find stat %v in stats summary response", ess.StatName)
```
ocket8888: I think if you make this a non-fatal error you can get a more exhaustive list of what's wrong without risking anything due to bad state/data

traffic_ops/testing/api/v14/stats_summary_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+var (
+	testStatsSummaries []tc.StatsSummary
+	latestTime         time.Time
+)
+
+const dateFormat = "2006-01-02"
+
+func TestStatsSummary(t *testing.T) {
+	WithObjs(t, []TCObj{}, func() {
```
ocket8888: I don't think you need `WithObjs` if you're not using any `TCObj`s

docs/source/api/stats_summary.rst
```diff
@@ -0,0 +1,175 @@
...
+
+Summary Stats
+"""""""""""""
+
+:cdnName:             CDN name summary stat was taken for
+:deliveryServiceName: :term:`Delivery Service` name summary stat was taken for
+:statName:            Stat name summary stat represents
+:statValue:           Summary stat value
+:summaryTime:         Timestamp of summary, in :rfc:`3339` format
+:statDate:            Date stat was taken
```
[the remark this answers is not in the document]
mhoppa: nope :/ let me put the format

traffic_ops/traffic_ops_golang/trafficstats/stats_summary.go
mhoppa: Wait how could you cause an internal server error? that isnt returned as part of the WHERE return value so there will not be overlap and I am just ignoring it with `_`

ocket8888: uhhh....
NO U

mitchell852:
I tried testing this out with ciab but only got an empty response for the call to `GET /api/1.4/stats_summary`. Since @ocket8888 has done the review, I'll leave it to him to test it as maybe he knows a better way than me.

mhoppa:
@mitchell852 yeah the stats_summary is done daily from TS so you would either have to POST to /stats_summary/create or have TS up for a day.

mhoppa:
@ocket8888  fixed

docs/source/api/stats_summary.rst
```diff
@@ -0,0 +1,178 @@
...
+	Content-Length: 150
+
+	{ "response": [
+		{
+            "cdnName": "CDN-in-a-Box",
+            "deliveryServiceName": "all",
+            "statName": "daily_maxgbps",
+            "statValue": 5,
+            "summaryTime": "2019-11-19 00:04:06+00",
+            "statDate": "2019-11-19
```
ocket8888: You're missing a closing quote here, which is breaking syntax highlighting - because lexical analysis is failing.

But also this code block is mixing indentation with alignment (i.e. spaces and tabs). It's fine to indent the actual contained code with spaces, but for consistency the content of a `.. code-block` (or any other directive) should be tab-indented. Because of that it's also best for the code therein contained to be indented using tabs, but that's up to you.

PR #4136, opened 2019-11-20
Block server type update when server is assigned DSes

### LINKED ISSUE #4113: TO: Can Update a Server's Type after its been assigned to Delivery Service
## I'm submitting a ...

- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

Delivery Service Required Capability is created before any Servers are assigned to a Delivery Service. 
Making a  `POST` Request to `https://{{TO_BASE_URL}}/api/{{api_version}}/deliveryserviceserver` with a server of  type other than `EDGE` is allowed in the API.
Updating that same server: `PUT` `https://{{TO_BASE_URL}}/api/{{api_version}}/servers/{id}/` to Type `EDGE` is allowed. Even though that server does not have the required capability associated to it.
## Expected / new behavior:

1. Should not be allowed to link a server of ANY type to a delivery service, ONLY Edge and Origin
2. Should not be allowed to update a server's type when linked to a delivery service.

mhoppa:

- [x] This PR fixes #4113 

- Traffic Ops

Run API tests or manually attempt to change a servers type that is assigned to DSes and that is not assigned to DSes. The latter should pass but the first should pass.

### DIFF
--- a/traffic_ops/testing/api/v14/servers_test.go
+++ b/traffic_ops/testing/api/v14/servers_test.go
@@ -101,7 +101,7 @@ func UpdateTestServers(t *testing.T) {
 		t.Fatalf("cannot GET Server Types: %v", err)
 	}
 	if len(serverTypes) < 2 {
-		t.Fatal("GET Server Types returned no types, must have at least 2 to test invalid type server update")
+		t.Fatal("GET Server Types returned less then 2 types, must have at least 2 to test invalid type server update")
 	}
 	for _, t := range serverTypes {
 		if t.ID != remoteServer.TypeID {
--- a/traffic_ops/testing/api/v14/servers_test.go
+++ b/traffic_ops/testing/api/v14/servers_test.go
@@ -23,7 +23,7 @@ import (
 )
 
 func TestServers(t *testing.T) {
-	WithObjs(t, []TCObj{CDNs, Types, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers}, func() {
+	WithObjs(t, []TCObj{CDNs, Types, Tenants, Users, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, DeliveryServices, Servers}, func() {
 		UpdateTestServers(t)
 		GetTestServers(t)
 	})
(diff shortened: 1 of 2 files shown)


traffic_ops/testing/api/v14/servers_test.go
```diff
@@ -87,6 +87,41 @@ func UpdateTestServers(t *testing.T) {
...
+	if len(dses) < 1 {
+		t.Fatal("GET DeliveryServices returned no dses, must have at least 1 to test invalid type server update")
+	}
+
+	serverTypes, _, err := TOSession.GetTypes("server")
+	if err != nil {
+		t.Fatalf("cannot GET Server Types: %v", err)
+	}
+	if len(serverTypes) < 2 {
+		t.Fatal("GET Server Types returned no types, must have at least 2 to test invalid type server update")
```
ocket8888: It's possible that it returned 1 type, not no types

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -36,7 +39,7 @@ import (
 
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/tenant"
-	"github.com/go-ozzo/ozzo-validation"
+	validation "github.com/go-ozzo/ozzo-validation"
```
ocket8888: why?

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -322,6 +325,27 @@ func (sv *TOServer) Update() (error, error, int) {
...
+	}
+
+	// If type is changing ensure it isnt assigned to any DSes
+	if typeID != *sv.TypeID {
+		dsIDs := []int64{}
+		if err := sv.APIInfo().Tx.QueryRowx("SELECT ARRAY(SELECT deliveryservice FROM deliveryservice_server WHERE server = $1)", sv.ID).Scan(pq.Array(&dsIDs)); err != nil && err != sql.ErrNoRows {
+			return nil, fmt.Errorf("getting server assigned delivery services: %v", err), http.StatusInternalServerError
+		}
+		if len(dsIDs) != 0 {
+			return errors.New("server type can not be updated when it is currently assigned to delivery services"), nil, http.StatusBadRequest
```
ocket8888: Personally I like [`409 Conflict`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) for things like this. Do with that what you will.

traffic_ops/testing/api/v14/servers_test.go
mhoppa: good catch

traffic_ops/traffic_ops_golang/server/servers.go
mhoppa: just matched how its being called in code but 🤷‍♂ ill change it back

ocket8888: I've seen that in few places, but the package is named `validation` already, so idk why. Maybe it used to be necessary when the package name didn't match the terminating directory name?

rob05c: +1 on removing the alias. Yeah, we have a bunch of places that do it, I'm guessing whoever added the library first didn't realize Go follows the package name, not the directory; and everyone else just copied it. If it's an idiom, IMO it's an antipattern and we should remove them.

PR #4142, opened 2019-11-21
Allow IPv6 Only Caches, including monitoring and routing

### LINKED ISSUE #2735: Traffic Monitor Polling over ipv6

mattjackson220:

- [x] This PR fixes #2735

It is also an update to PR #4012 so that one can be closed

- Documentation
- Traffic Monitor
- Traffic Ops
- Traffic Portal
- Traffic Router

- Update CDN in a Box Docker to include IPv6 or test in an environment that has IPv6.  
- Verify that the new IPv4 and IPv6 columns are added to the TM UI and are showing the correct boolean for availability for IPv4 (since IPv4 only is the default) and that it is polling successfully. 
- Update the TM config file to include
"cache_polling_protocol": "both",
"peer_polling_protocol": "both"

- Restart TM with the new config. Verify that it is successfully polling availability information for both IPv4 and IPv6. Verify that servers with IPv4 are still successfully polled. Verify that any issues are being reported correctly for both IPv4 and IPv6.
- Verify that TP allows servers to have an IPv6 only address, but does require either IPv4, IPv6, or both to be defined.
- Create a new server object or update an old server object to have only an IPv6 address.  Verify that this its availability was updated in TM.
- Run a curl command with -4 or -6 to verify that the IPv4 and IPv6 routing both work.  
- In TP, uncheck the IsServiceAddress box for IPv4 to disable this address.  Perform a snapshot.
- Verify that IPv4 is marked as unavailable in TM, but that the Cache as a whole is marked as available still.
- Verify that a curl with -6 still works but a curl with -4 returns a 503 status.
- Repeat these steps but uncheck the IsServiceAddress for IPv6 and verify that -4 is functional and -6 returns a 503.
- Update IPv4 address to "127.0.0.0" (or any valid IPv4 address that will get an error when TM requests status) and verify that it gets an error in TM and has the IPv4 marked unavailable as expected.
- Update IPv6 address to "::1" (or any valid IPv6 address that will get an error when TM requests status) and verify that it gets an error in TM and has the IPv6 marked unavailable as expected.


ocket8888:
I also have some misgivings about the way this is implemented in the TO API. I think the most common case for cache server services is to serve regardless of whether they were contacted over IPv4 or IPv6, so it seems odd to me that I can't say "both" in the current configuration language. It also sort of feels like using two fields to convey one concept. Instead of "isIP6Service" "isIP4Service" what about "serviceIP" set to one of "ipv4", "ipv6" or "both"?

docs/source/admin/traffic_monitor.rst
```diff
@@ -50,8 +50,15 @@ Configuring Traffic Monitor
...
 
+Polling protocol can be set for peers and caches and has 3 options:
+
+ipv4only (the default): TM will communicate with the peers or caches only over IPv4
+ipv6only: TM will communicate with the peers or caches only over IPv6 (use case for peers is if the other TMs are only available over IPv6)
+both: TM will alternate its communication between IPv4 and IPv6 (note: this does not affect the polling frequency so if polling frequency is 1 second IPv4 will be polled every 2 seconds)
```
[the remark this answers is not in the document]
mattjackson220: done

docs/source/admin/traffic_monitor.rst
```diff
@@ -60,7 +67,8 @@ The :term:`cache servers` are polled at the URL specified in the ``health.pollin
 
 This :term:`parameter` must have the config file ``rascal.properties``.
 
-The value is a template with the text ``${hostname}`` being replaced with the :term:`cache server`'s Network IP (IPv4), and ``${interface_name}`` being replaced with the :term:`cache server`'s network Interface Name. For example, ``http://${hostname}/_astats?application=&inf.name=${interface_name}``.
+The value is a template with the text ``${hostname}`` being replaced with the :term:`cache server`'s Network IP (IPv4, IPv6, or alternating between IPv4 and IPv6 depending on the cache polling protocol described above), and ``${interface_name}`` being replaced with the :term:`cache server`'s network Interface Name. For example, ``http://${hostname}/_astats?application=&inf.name=${interface_name}``.
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_monitor/cache/data.go
```diff
@@ -24,19 +24,33 @@ import (
...
 // AvailableStatusReported is the status string returned by caches set to "reported" in Traffic Ops.
 // TODO put somewhere more generic
 const AvailableStatusReported = "REPORTED"
 
+type AvailableTuple struct {
+	IPv4 bool
+	IPv6 bool
+}
+
+func (a *AvailableTuple) SetAvailability(usingIPv4 bool, isAvailable bool) {
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_monitor/config/config.go
```diff
@@ -46,35 +49,77 @@ const (
...
+	case IPv6Only.String():
+		return IPv6Only
+	case Both.String():
+		return Both
+	default:
+		return InvalidPollingProtocol
+	}
+}
+
+func (t *PollingProtocol) UnmarshalJSON(b []byte) error {
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_monitor/manager/monitorconfig.go
```diff
@@ -146,11 +148,12 @@ func StartMonitorConfigManager(
 		staticAppData,
 		toSession,
 		toData,
+		url6Regex,
 	)
 	return monitorConfig
 }
 
-const DefaultHealthConnectionTimeout = time.Second * 2
+const DefaultHealthConnectionTimeout = time.Second * 20
```
rob05c: What's the thinking behind this? That's a pretty big jump

traffic_monitor/manager/manager.go
```diff
@@ -45,6 +46,8 @@ import (
 func Start(opsConfigFile string, cfg config.Config, appData config.StaticAppData, trafficMonitorConfigFileName string) error {
 	toSession := towrap.ITrafficOpsSession(towrap.NewTrafficOpsSessionThreadsafe(nil, cfg.CRConfigHistoryCount, cfg))
 
+	url6Regex := regexp.MustCompile(`/\d+`)
```
rob05c: Regexes are typically much slower than simple string manipulation, and TM is highly performance-sensitive. This isn't actually in the request path; but it does get called for every server, every time a new config is received.

Would you object to changing the usage to a string function? E.g.
 
```
// ipv6CIDRStrToAddr takes an IPv6 CIDR string, e.g. `2001:DB8::1/32` returns `2001:DB8::1`.
// It does not verify cidr is a valid CIDR or IPv6. It only removes the first slash and everything after it, for performance.
func ipv6CIDRStrToAddr(cidr string) string {
	i := strings.Index(cidr, `/`)
	if i == -1 {
		return cidr
	}
	return cidr[:i]
}
```
https://play.golang.org/p/1tLFJ7P1C91

ocket8888: If you're mostly just interested in getting the IP address part of a CIDR-notation network then you could also just get the first return value of [`net.ParseCIDR`](https://godoc.org/net#ParseCIDR)

rob05c: `net.ParseCIDR` is expensive too, though, and doing a lot of work that we don't need here.

rob05c: I wanted to be sure, so I wrote a benchmark:

```
package main

import (
	"math/rand"
	"net"
	"regexp"
	"strconv"
	"strings"
	"testing"
)

func BenchmarkRemoveCIDRFunc(b *testing.B) {
	cidrStrs := randCIDRStrs(b.N)
	ipStrs := make([]string, len(cidrStrs), len(cidrStrs))

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		ipStrs[i] = ipv6CIDRStrToAddr(cidrStrs[i])
	}
}

func BenchmarkRemoveCIDRRegex(b *testing.B) {
	re := regexp.MustCompile(`/\d+`)
	cidrStrs := randCIDRStrs(b.N)
	ipStrs := make([]string, len(cidrStrs), len(cidrStrs))

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		ipStrs[i] = re.ReplaceAllString(cidrStrs[i], "")
	}
}

func BenchmarkRemoveCIDRIP(b *testing.B) {
	cidrStrs := randCIDRStrs(b.N)
	ipStrs := make([]string, len(cidrStrs), len(cidrStrs))

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		ip, _, err := net.ParseCIDR(cidrStrs[i])
		if err != nil {
			b.Fatal("not an IP: '" + cidrStrs[i] + "'")
		}
		ipStrs[i] = ip.String()
	}
}

func randCIDRStrs(n int) []string {
	ips := []net.IP{}
	for i := 0; i < n; i++ {
		ip := make([]byte, 16, 16)
		rand.Read(ip) // math/rand.Read is documented to always return a nil error.
		ips = append(ips, net.IP(ip))
	}

	ipStrs := []string{}
	for _, ip := range ips {
		ipStrs = append(ipStrs, ip.String()+`/`+strconv.Itoa(rand.Intn(129)))
	}

	return ipStrs
}

// ipv6CIDRStrToAddr takes an IPv6 CIDR string, e.g. `2001:DB8::1/32` returns `2001:DB8::1`.
// It does not verify cidr is a valid CIDR or IPv6. It only removes the first slash and everything after it, for performance.
func ipv6CIDRStrToAddr(cidr string) string {
	i := strings.Index(cidr, `/`)
	if i == -1 {
		return cidr
	}
	return cidr[:i]
}
```

```
$ go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/apache/trafficcontrol/traffic_monitor
BenchmarkRemoveCIDRFunc-16      63100605                23.7 ns/op
BenchmarkRemoveCIDRRegex-16      4373604               290 ns/op
BenchmarkRemoveCIDRIP-16         2069672               528 ns/op
PASS
ok      github.com/apache/trafficcontrol/traffic_monitor        37.686s
```

ocket8888: Sure but if the result of stripping is used for TCP dials then you're paying that cost anyway. Idk how it's used later on, just wanted to throw that out there.

rob05c: It isn't. But that's not a bad idea, might be worth looking into. I don't know how the cost of parsing the IP compares to the request cost; but we do enough requests, and use enough CPU, it might be significant.

traffic_monitor/health/cache.go
```diff
@@ -171,45 +172,100 @@ func EvalCache(result cache.ResultInfo, resultStats *threadsafe.ResultStatValHis
...
+			return tuple.IPv6
+		case config.Both:
+			// only report availability based on defined IP addresses
+			if serverInfo.IP == "" {
+				return tuple.IPv6
+			} else if serverInfo.IP6 == "" {
+				return tuple.IPv4
+			}
+			// if both IP addresses are defined then report availability based on both
+			return tuple.IPv4 || tuple.IPv6
```
rob05c: This returns that the cache is healthy if _either_ IPv4 _or_ IPv6 polling reported available? Is that intentional?
If so, I think we should document that. I didn't see anything in the docs indicating that "both" is really "either."

It'd be nice if it were a config option, for "both" and "either." But I don't want to ask you to do more work, this works as-is, as long as we document the behavior. We can always add "either" later.

traffic_monitor/poller/cache.go
```diff
@@ -162,14 +181,30 @@ func poller(
 			pollID := atomic.AddUint64(&pollNum, 1)
 			pollFinishedChan := make(chan uint64)
 			log.Debugf("poll %v %v start\n", pollID, time.Now())
-			bts, reqEnd, reqTime, err := pollFunc(pollCtx, url, host, pollID)
-			rdr := io.Reader(nil)
-			if bts != nil {
-				rdr = bytes.NewReader(bts) // TODO change handler to take bytes? Benchmark?
+			if usingIPv4 {
+				bts, reqEnd, reqTime, err := pollFunc(pollCtx, url, host, pollID)
```
rob05c: It looks like this and the next block are identical, except for `url`. Could the duplication be reduced by changing this to 
```
pollURL := url
if !usingIPv4 {
  pollURL = url6
}
bts, reqEnd, reqTime, err := pollFunc(pollCtx, pollURL, host, pollID)
```
?

traffic_ops/traffic_ops_golang/crconfig/servers.go
```diff
@@ -111,7 +111,7 @@ func getAllServers(cdn string, tx *sql.Tx) (map[string]ServerUnion, error) {
 
 	// TODO select deliveryservices as array?
 	q := `
-select s.host_name, cg.name as cachegroup, concat(s.host_name, '.', s.domain_name) as fqdn, s.xmpp_id as hashid, s.https_port, s.interface_name, s.ip_address, s.ip6_address, s.tcp_port, p.name as profile_name, cast(p.routing_disabled as int), st.name as status, t.name as type
+select s.host_name, cg.name as cachegroup, concat(s.host_name, '.', s.domain_name) as fqdn, s.xmpp_id as hashid, s.https_port, s.interface_name, case when s.ip_address_is_service = true then s.ip_address else '' end, case when s.ip6_address_is_service = true then s.ip6_address else '' end, s.tcp_port, p.name as profile_name, cast(p.routing_disabled as int), st.name as status, t.name as type
```
rob05c: How strongly do you feel about this, vs an `if` statement in Go?

The `case` here is a conditional. I know this feels like a nitpick, but in my experience, embedding code in the database or queries leads to pain. I think if we keep code as code, and data as data, and SQL as a declarative query, it just makes everything easier to understand and work with, and less to misunderstand or go wrong.

Here, the `case` is technically SQL, but it's an imperative `if` statement, not declarative like SQL is intended to be. It's not a query, it's logic.

Would you object to changing this to something like:
```
	qry := `
SELECT
  s.host_name,
  cg.name AS cachegroup,
  CONCAT(s.host_name, '.', s.domain_name) AS fqdn,
  s.xmpp_id AS hashid,
  s.https_port,
  s.interface_name,
  s.ip_address_is_service,
  s.ip6_address_is_service,
  s.ip_address,
  s.ip6_address,
  s.tcp_port,
  p.name AS profile_name,
  CAST(p.routing_disabled AS int),
  st.name AS status,
  t.name AS type
FROM
  server s
  JOIN cachegroup cg ON cg.id = s.cachegroup
  JOIN type t on t.id = s.type
  JOIN profile p ON p.id = s.profile
  JOIN status st ON st.id = s.status
WHERE
  cdn_id = (SELECT id FROM cdn WHERE name = $1)
  AND (st.name = 'REPORTED' or st.name = 'ONLINE' or st.name = 'ADMIN_DOWN')
`
	rows, err := tx.Query(qry, cdn)
	if err != nil {
		return nil, errors.New("Error querying servers: " + err.Error())
	}
	defer rows.Close()

	for rows.Next() {
		port := sql.NullInt64{}
		ip6 := sql.NullString{}
		hashId := sql.NullString{}
		httpsPort := sql.NullInt64{}
		ipIsService := false
		ipIsService := false
		host := ""
		status := ""
		s := ServerUnion{}
		if err := rows.Scan(&host, &s.CacheGroup, &s.Fqdn, &hashId, &httpsPort, &s.InterfaceName, &ipIsService, &ip6IsService, &s.Ip, &ip6, &port, &s.Profile, &s.RoutingDisabled, &status, &s.ServerType); err != nil {
			return nil, errors.New("Error scanning server: " + err.Error())
		}
		if !ipIsService {
			s.Ip = util.StrPtr("")
		}
		if !ip6IsService {
			s.Ip6 = util.StrPtr("")
		}
```

Really, the very need for imperative logic here is a result of denormalization. But we certainly don't have time to properly normalize our schema right now.

I know my position is a little bit nuanced. If you feel strongly in the other direction, I won't push for it. But if it's all the same to you, IMO the separation of data vs code makes things easier and safer to work with.

traffic_monitor/manager/monitorconfig.go
mattjackson220: in testing it seemed necessary but must have been an environment issue because now 2 seems fine. reverted

traffic_monitor/manager/manager.go
mattjackson220: done

traffic_monitor/poller/cache.go
mattjackson220: done

traffic_ops/traffic_ops_golang/crconfig/servers.go
mattjackson220: i agree with that. done

traffic_monitor/health/cache.go
mattjackson220: yea i have it doing either because then TR can find all of the healthy caches even if only one is available. for example if IPv4 has an error but IPv6 is still reporting healthy, then TR can still direct traffic to the IPv6 address.  I like the idea of having it as a config option though so people can choose for themselves

docs/source/api/v1/servers.rst
```diff
@@ -90,8 +90,16 @@ Response Structure
 :interfaceMtu:   The :abbr:`MTU (Maximum Transmission Unit)` configured on ``interfaceName``
 :interfaceName:  The name of the primary network interface used by the server
 :ip6Address:     The IPv6 address and subnet mask of ``interfaceName``
+:ip6IsService:   A boolean value which if ``true`` indicates that the IPv6 address will be used for routing content.
+
+	.. versionadded:: 2.0
```
mhoppa: all the versionadded should be added to the v2.0 docs and removed completely from the v1 docs. Also when moving them no need to include versionadded anymore

mattjackson220: done

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -16,10 +16,7 @@
 package com.comcast.cdn.traffic_control.traffic_router.core.router;
 
 import java.io.IOException;
-import java.net.InetAddress;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.net.UnknownHostException;
+import java.net.*;
```
rawlinp: This is a bit of a nitpick, but I think we should stay consistent with our imports in explicitly importing every class individually that is actually used.

rawlinp:
Also as a note, I was poking around the TP servers page, and updating the "ip is service" boxes actually updates the columns in the DB, but TP `GET`s the server using 1.5 which doesn't include the "ip is service" fields. So TP always shows the checkboxes checked when you refresh currently.

When requesting at 2.0, the new fields are there as expected, so I'm guessing this really just depends on TP being updated to 2.0 (which I know is a PR in progress elsewhere) and isn't necessarily a bug.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -350,6 +375,27 @@ func (s *TOServer) Update() (error, error, int) {
 		}
 	}
 
+	current := TOServer{}
+	err := s.ReqInfo.Tx.QueryRowx(selectV20UpdatesQuery() + ` WHERE sv.id=` + strconv.Itoa(*s.ID)).StructScan(&current)
```
rawlinp: Thinking about sql injections here, should we make this query like this instead?
```
... +" WHERE sv.id = $1", *s.ID).StructScan...
```
Obviously an integer ID can never lead to sql injection, but it's probably good practice to use parameterized queries whenever possible.

mattjackson220: done

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
mattjackson220: done

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/cache/Cache.java
```diff
@@ -228,9 +234,53 @@ public InetAddress getIp6() {
...
+		} else if (ipv6Available && !ipv4Available) {
+			this.ipAvailableVersions = IpVersions.IPV6ONLY;
+		} else {
+			this.ipAvailableVersions = IpVersions.BOTH;
+		}
+
+		for (final InetRecord record : ipAddresses) {
+			if (record.getAddress() == ip4 && !ipv4Available) {
+				newlyUnavailable.add(record);
+				LOGGER.info("disabling IPv4 address " + record.getAddress());
```
rawlinp: `INFO` seems like the wrong logging level for this kind of information. It might make more sense if it was just logging the state changes (e.g. disabled -> enabled and vice versa), but logging states for every server for every state poll from TM seems like too much. Should we make this `DEBUG` instead?

traffic_ops/app/db/migrations/20200220000000_add_server_ip_toggles.sql
```diff
@@ -0,0 +1,31 @@
...
+ALTER TABLE server ADD CONSTRAINT need_netmask_if_ip CHECK (ip_address IS NULL OR ip_netmask IS NOT NULL);
+
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+ALTER TABLE server DROP COLUMN ip_address_is_service;
+ALTER TABLE server DROP COLUMN ip6_address_is_service;
+ALTER TABLE server ALTER COLUMN ip_address SET NOT NULL;
+ALTER TABLE server ALTER COLUMN ip_netmask SET NOT NULL;
+ALTER TABLE server ALTER COLUMN ip_gateway SET NOT NULL;
+ALTER TABLE server DROP CONSTRAINT need_at_least_one_ip;
```
rawlinp: This is missing a `DROP CONSTRAINT` for the other two constraints

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/cache/Cache.java
mattjackson220: this should only log the state changes. the top loop logs if a previously enabled ip address is now disabled, and the bottom does the opposite. this might still be too much for an INFO though. im good with changing it to DEBUG

mattjackson220: done

traffic_ops/app/db/migrations/20200220000000_add_server_ip_toggles.sql
mattjackson220: done

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
```diff
@@ -133,20 +133,21 @@
                     <span ng-show="hasError(serverForm.interfaceName)" class="form-control-feedback"><i class="fa fa-times"></i></span>
                 </div>
             </div>
-            <div class="form-group" ng-class="{'has-error': hasError(serverForm.ipAddress), 'has-feedback': hasError(serverForm.ipAddress)}">
+            <div class="form-group" ng-class="{'has-error': hasError(serverForm.ipAddress) && hasError(serverForm.ip6Address), 'has-feedback': hasError(serverForm.ipAddress) && hasError(serverForm.ip6Address)}">
                 <label class="control-label col-md-2 col-sm-2 col-xs-12">Network IP *</label>
```
mitchell852: how about adding a tooltip to explain what this new checkbox is for? Kinda like this:

```
                <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12">Network IP *<div class="helptooltip">
                    <div class="helptext">Add a description here about what this IP is for and what 'is service address' means</div>
                </div>
                </label>
```

also do that for IPv6?

mitchell852: you can find descriptions of each field here: https://traffic-control-cdn.readthedocs.io/en/latest/api/v2/servers.html#post

mitchell852:
is it a problem that i can set ip6Address to null and ip6IsService = true via the api? i'm guessing i can do the same with ipAddress and ipIsService as well. Seems like the api should maybe do some data massaging imo. i.e. if ip6Address is null, then set ip6IsService to false no matter what.

mattjackson220:
> is it a problem that i can set ip6Address to null and ip6IsService = true via the api? i'm guessing i can do the same with ipAddress and ipIsService as well. Seems like the api should maybe do some data massaging imo. i.e. if ip6Address is null, then set ip6IsService to false no matter what.

@ocket8888  and @mhoppa  what are your thoughts as TO working group leaders? I'm debating if this instance should return a 400 and a message like "empty addresses cannot be set as service addresses" or if it should just set the isServiceAddress field to false if there isnt an address provided.

mhoppa:
> > is it a problem that i can set ip6Address to null and ip6IsService = true via the api? i'm guessing i can do the same with ipAddress and ipIsService as well. Seems like the api should maybe do some data massaging imo. i.e. if ip6Address is null, then set ip6IsService to false no matter what.
> 
> @ocket8888 and @mhoppa what are your thoughts as TO working group leaders? I'm debating if this instance should return a 400 and a message like "empty addresses cannot be set as service addresses" or if it should just set the isServiceAddress field to false if there isnt an address provided.

I personally prefer APIs that dont change stuff for me automatically but instead tell me I am doing it wrong. that being said I would vote for 400.

ocket8888:
I agree, 400 is better

mitchell852:
@mattjackson220 - just want to make sure this is fine and TR will handle it correctly. Here's what the cdn snapshot looks like when setting a caches ip address as not a service address:

```
host-name: {
	...
	ip: "",
	ip6: "2001:558:fe36:66::2/64",
	,,,
},
```

traffic_portal/app/src/common/api/ServerService.js
```diff
@@ -42,6 +42,12 @@ var ServerService = function($http, locationUtils, messageModel, ENV) {
     };
 
     this.createServer = function(server) {
+        if (server.ip6Address === undefined || server.ip6Address === '') {
```
mitchell852: `if (!server.ip6Address || server.ip6Address === '')` will check if ip6Address is undefined, null or blank

mitchell852: actually, i think you can just do `if (!server.ip6Address)`

falsy values are null, undefined, empty strings, false, 0

so if it's any of those things, it will evaluate to true

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
```diff
@@ -134,8 +134,11 @@
                 </div>
             </div>
             <div class="form-group" ng-class="{'has-error': hasError(serverForm.ipAddress) && hasError(serverForm.ip6Address), 'has-feedback': hasError(serverForm.ipAddress) && hasError(serverForm.ip6Address)}">
-                <label class="control-label col-md-2 col-sm-2 col-xs-12">Network IP *</label>
-                <div class="col-md-10 col-sm-10 col-xs-12">
+				<label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12">Network IP *<div class="helptooltip">
```
mitchell852: looks like some formatting problems

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
```diff
@@ -184,8 +187,11 @@
                 </div>
             </div>
             <div class="form-group" ng-class="{'has-error': hasError(serverForm.ip6Address) && hasError(serverForm.ipAddress), 'has-feedback': hasError(serverForm.ip6Address) && hasError(serverForm.ipAddress)}">
-                <label class="control-label col-md-2 col-sm-2 col-xs-12">IPv6 Address</label>
-                <div class="col-md-10 col-sm-10 col-xs-12">
+				<label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12">IPv6 Address *<div class="helptooltip">
+						<div class="helptext">An IPv6 address and subnet mask of Interface Name.  If the Is Service Address box is checked, then this address will be included in the CrConfig and traffic will be routed to this address. At least one Network IP or IPv6 Address must be provided.</div>
```
mitchell852: maybe you should replace Crconfig with 'Traffic Router config' as i think that would make more sense to a normalish user.

traffic_portal/app/src/common/modules/form/server/FormServerController.js
```diff
@@ -80,6 +80,9 @@ var FormServerController = function(server, $scope, $location, $state, $uibModal
 
     $scope.server = server;
 
+    $scope.server.ip6IsService = $scope.server.ip6IsService === undefined ? true : $scope.server.ip6IsService;
```
mitchell852: this is a bit confusing. not sure what's going on there.

traffic_portal/app/src/common/modules/form/server/FormServerController.js
```diff
@@ -158,6 +161,40 @@ var FormServerController = function(server, $scope, $location, $state, $uibModal
...
+        var elementToChange = elementId === 'ipGateway' ? 'ip6Gateway' : 'ipGateway';
+
+        if (document.getElementById(elementId).value === '') {
+            document.getElementById(elementToChange).required = true;
+        } else {
+            document.getElementById(elementToChange).required = false;
+        }
+    };
+
+    $scope.isIPv4 = function() {
```
mitchell852: maybe you should rename these to hasIPv4 because this test is not really checking if it "IS" IPv4

traffic_portal/app/src/common/modules/form/server/FormServerController.js
mattjackson220: was just trying to default it to true if it is undefined, otherwise leave it as is

traffic_portal/app/src/common/api/ServerService.js
mattjackson220: done

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
mattjackson220: done

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
mattjackson220: done

traffic_portal/app/src/common/modules/form/server/FormServerController.js
mattjackson220: done

traffic_portal/app/src/common/api/ServerService.js
ocket8888: You could just do `if (server.ip6Address)`:
```javascript
if ("") {console.log("true")} else {console.log("false")}
// prints "false"
```

PR #4143, opened 2019-11-22
Deprecate /hwinfo

### LINKED ISSUE #3798: Rewrite /hwinfo to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/hwinfo` is handled by Perl

## Expected / new behavior:
`/hwinfo` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`

N/A

## Anything Else
A route named `/hwinfo-wip` seems to exist. Progress to completion is unknown.

ocket8888:
Adds deprecation notices to the client method and `alerts` response field of the API endpoint. It also fixes the endpoint to be compatible with its structure from the Perl handler.

- [x] This PR fixes #3798

Draft status pending community consensus on deprecation.

- Traffic Control Client (Go)
- Traffic Ops

Run the unchanged API tests and unit tests.

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/api/shared_handlers_test.go
+++ b/traffic_ops/traffic_ops_golang/api/shared_handlers_test.go
@@ -142,7 +142,7 @@ func TestCreateHandler(t *testing.T) {
 	createFunc(w, r)
 
 	//verifies the body is in the expected format
-	body := `{"alerts":[{"text":"tester was created.","level":"success"}],"response":{"ID":1}}`
+	body := "{\"alerts\":[{\"text\":\"tester was created.\",\"level\":\"success\"}],\"response\":{\"ID\":1}}\n"
 	if w.Body.String() != body {
 		t.Error("Expected body", body, "got", w.Body.String())
 	}
@@ -228,7 +228,7 @@ func TestUpdateHandler(t *testing.T) {
 	updateFunc(w, r)
 
 	//verifies the body is in the expected format
-	body := `{"alerts":[{"text":"tester was updated.","level":"success"}],"response":{"ID":1}}`
+	body := "{\"alerts\":[{\"text\":\"tester was updated.\",\"level\":\"success\"}],\"response\":{\"ID\":1}}\n"
 	if w.Body.String() != body {
 		t.Error("Expected body", body, "got", w.Body.String())
 	}
@@ -273,7 +273,7 @@ func TestDeleteHandler(t *testing.T) {
 	deleteFunc(w, r)
 
 	//verifies the body is in the expected format
-	body := `{"alerts":[{"text":"tester was deleted.","level":"success"}]}`
+	body := "{\"alerts\":[{\"text\":\"tester was deleted.\",\"level\":\"success\"}]}\n"
 	if w.Body.String() != body {
 		t.Error("Expected body", body, "got", w.Body.String())
 	}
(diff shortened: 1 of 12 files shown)


traffic_ops/traffic_ops_golang/api/shared_handlers_test.go
```diff
@@ -142,7 +142,7 @@ func TestCreateHandler(t *testing.T) {
 	createFunc(w, r)
 
 	//verifies the body is in the expected format
-	body := `{"alerts":[{"text":"tester was created.","level":"success"}],"response":{"ID":1}}`
+	body := "{\"alerts\":[{\"text\":\"tester was created.\",\"level\":\"success\"}],\"response\":{\"ID\":1}}\n"
```
rawlinp: This might be easier to read if you keep the backticks and instead just add `+"\n"` to the end. Does that work?

Same for the two lines below as well.

traffic_ops/traffic_ops_golang/login/login_test.go
```diff
@@ -48,7 +48,7 @@ func TestLoginWithEmptyCredentials(t *testing.T) {
 		}
 		LoginHandler(nil, config.Config{})(w, r)
 
-		expected := `{"alerts":[{"text":"username and password are required","level":"error"}]}`
+		expected := "{\"alerts\":[{\"text\":\"username and password are required\",\"level\":\"error\"}]}\n"
```
rawlinp: ditto earlier comment

traffic_ops/traffic_ops_golang/routing/wrappers_test.go
```diff
@@ -222,7 +222,7 @@ func TestWrapAuth(t *testing.T) {
 
 	f(w, r)
 
-	expectedError := `{"alerts":[{"text":"Unauthorized, please log in.","level":"error"}]}`
+	expectedError := "{\"alerts\":[{\"text\":\"Unauthorized, please log in.\",\"level\":\"error\"}]}\n"
```
rawlinp: ditto earlier comment

traffic_ops/traffic_ops_golang/hwinfo/hwinfo.go
```diff
@@ -20,74 +20,116 @@ package hwinfo
...
 
+const selectHWInfoQuery = `
+SELECT
+	s.host_name as serverhostname,
+	h.id,
+	h.serverid,
+	h.description,
+	h.val,
+	h.last_updated
+FROM hwInfo h
```
rawlinp: the table is `hwinfo`, but I guess the table names are case-insensitive?

rob05c: They are in queries, unless quoted. But better to use the real name anyway, safer and less confusing.

traffic_ops/traffic_ops_golang/api/shared_handlers_test.go
ocket8888: sure

traffic_ops/traffic_ops_golang/hwinfo/hwinfo.go
ocket8888: Ah, that was probably tab-completion

PR #4145, opened 2019-11-22
Add TO API Routing Blacklist via cdn.conf

rawlinp:
Add a `"routing_blacklist"` field to cdn.conf's "traffic_ops_golang"
section, consisting of two array fields:
- `"perl_routes"`: routes that will be handled by TO-Perl instead of Go
- `"disabled_routes"`: routes that will be disabled by immediately
    returning a 503

These fields are arrays of route IDs (integers). As part of this change,
routes in `traffic_ops_golang` get a unique ID now, as well as a boolean
field indicating whether or not the route can be bypassed to TO-Perl.

Add a `--api-routes` flag to `traffic_ops_golang` that prints out all the
route information (ID, version, path, etc) and exits.

- Documentation
- Traffic Ops

1. Run the `traffic_ops_golang` unit tests
2. Run the TO API tests
3. Configure `cdn.conf` to disable a certain route, make a request to that route, and verify that a 503 response is returned.
4. Configure `cdn.conf` to have Perl handle a certain route, make a request to that route, and verify that the route was handled by TO-Perl and not Go.


docs/source/admin/traffic_ops.rst
```diff
@@ -409,6 +415,12 @@ This file deals with the configuration parameters of running Traffic Ops itself.
...
+		:disabled_routes: A list of API route IDs to disable. Requests matching these routes will receive a 503 response. To find the route ID for a given path you would like to disable, run ``./traffic_ops_golang --api-routes`` to view all the route information, including route IDs and paths.
```
ocket8888: Since you made the `--api-routes` an `.. option::` you can link to it directly e.g.
```rst
... using the :option:`--api-routes` option...
```

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -109,433 +121,451 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
...
-		{1.1, http.MethodGet, `cdns/name/{name}/dnsseckeys/delete/?(\.json)?$`, cdn.DeleteDNSSECKeys, auth.PrivLevelAdmin, Authenticated, nil},
-		{1.4, http.MethodGet, `cdns/name/{name}/dnsseckeys/?(\.json)?$`, cdn.GetDNSSECKeys, auth.PrivLevelAdmin, Authenticated, nil},
-		{1.1, http.MethodGet, `cdns/name/{name}/dnsseckeys/?(\.json)?$`, cdn.GetDNSSECKeysV11, auth.PrivLevelAdmin, Authenticated, nil},
+		{1.1, http.MethodPost, `cdns/{id}/queue_update$`, cdn.Queue, auth.PrivLevelOperations, Authenticated, nil, 33, noPerlBypass},
+		{1.1, http.MethodPost, `cdns/dnsseckeys/generate(\.json)?$`, cdn.CreateDNSSECKeys, auth.PrivLevelAdmin, Authenticated, nil, 34, noPerlBypass},
+		{1.1, http.MethodGet, `cdns/name/{name}/dnsseckeys/delete/?(\.json)?$`, cdn.DeleteDNSSECKeys, auth.PrivLevelAdmin, Authenticated, nil, 35, noPerlBypass},
+		{1.4, http.MethodGet, `cdns/name/{name}/dnsseckeys/?(\.json)?$`, cdn.GetDNSSECKeys, auth.PrivLevelAdmin, Authenticated, nil, 36, noPerlBypass},
+		{1.1, http.MethodGet, `cdns/name/{name}/dnsseckeys/?(\.json)?$`, cdn.GetDNSSECKeysV11, auth.PrivLevelAdmin, Authenticated, nil, 37, noPerlBypass},
```
ocket8888: I think this last one is a new rewrite in 4.x, should it be Perl-bypass-able?

traffic_ops/traffic_ops_golang/config/config.go
```diff
@@ -108,6 +109,12 @@ type ConfigTrafficOpsGolang struct {
 	CRConfigEmulateOldPath bool `json:"crconfig_emulate_old_path"`
 }
 
+// RoutingBlacklist contains the list of route IDs that will be handled by TO-Perl and a list of route IDs that are disabled.
+type RoutingBlacklist struct {
+	PerlRoutes     []int `json:"perl_routes"`
+	DisabledRoutes []int `json:"disabled_routes"`
+}
```
alficles: Hrm, is there an alternative to integers here? It looks like we're exposing internal structure data in a config file. I see we added a way for the route numbers to be retrieved by an end-user, but it definitely makes this feature tough to use. Somebody reading the file will have to have the route numbers memorized. (JSON's lack of comments isn't doing us any favours, either.)

It also sounds like a code maintenance problem. New routes will conflict in a semantically meaningful way. This significantly increases the chance that merges and cherry picks will create semantic conflicts in the absence of code conflicts. That's going to be a pain to manage.

Ideally, the numbers would be assigned automatically on registration and only used internally, with the external config using some consistent route identifier.

ocket8888: This was discussed on the mailing list, and the general thinking was that asking people to craft regular expressions to match regular expressions was too much, and that we should use these IDs until the route handling has been reworked to not rely on regular expressions.

rawlinp: Please see the corresponding dev mailing list thread. This had already been hashed out a bit, and your concerns are not unique.

traffic_ops/traffic_ops_golang/routing/routes.go
rawlinp: I made a route Perl-able if it has not been "vetted" by any production environment yet. Since my organization has "vetted" certain routes that have been rewritten since 3.x, I didn't include those routes as Perl-able. If we think we should make all rewritten routes since 3.x Perl-able, we could do that, but I don't really think that's necessary.

traffic_ops/traffic_ops_golang/config/config.go
rob05c: Changing these "route IDs" is also going to break any Production blacklist config, so they need to be immutable. That seems to be suggested on the mailing list, but I don't see it anywhere here.

Would you mind adding something to that effect, both in the TO Dev Docs, and as a comment in the file here? 

Otherwise, someone might see the mostly-ordered numbers here, and think they should be ordered, and change them when adding a new route.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -89,17 +90,28 @@ import (
...
 
 // Authenticated ...
-var Authenticated = true
+const Authenticated = true
 
 // NoAuth ...
-var NoAuth = false
+const NoAuth = false
+
+const perlBypass = true
```
rob05c: Would you mind adding comments explaining what these variables are doing?

Reading the routes below, it wasn't obvious to me what "perl bypass" meant, I had to go look it up on the mailing list and make some inferences.

rob05c:
>Hrm, is there an alternative to integers here? It looks like we're exposing internal structure data in a config file.

Hm, I see Chris's point here, numbers seem a bit more difficult for people to use. But I also see your point on the list about regexes being a pain.

Question: is there a reason the route ID needs to be a number? What if every Route had a "Route ID" that was an arbitrary string, e.g. `get-cachegroups`?

```
		{1.1, `get-cachegroups`, http.MethodGet, `cachegroups/{id}$`, api.ReadHandler(&cachegroup.TOCacheGroup{}), auth.PrivLevelReadOnly, Authenticated, nil, noPerlBypass},

Would it be difficult to change the existing code to that?

If not, is there an advantage to numbers over that?

**Comment by rawlinp** [comment]
> Question: is there a reason the route ID needs to be a number? What if every Route had a "Route ID" that was an arbitrary string, e.g. get-cachegroups?

I considered that as well, but what would you name the 10 different variations of the `federations` endpoints? Plus, imagine how tedious it would be to come up with a meaningful name for every single endpoint we have today. It just seems like wasted work when we know we want the end goal to be using non-regex-based paths, and using those in the config rather than some type of ID.

File: traffic_ops/traffic_ops_golang/config/config.go (hunk 3 above)
Comment (alficles) [reply traffic_ops/traffic_ops_golang/config/config.go]: Fair enough. I read through the initial proposal, liked what I saw and saw it had considerable support. I didn't follow the rest of the thread to see the integer-version. You're 100% right that this is better than the regex version. I'm concerned that our definition of "short-lived" is probably measured in years, though. Rob is also right about the danger if route numbers ever change.

Sorry about coming into the conversation late. I was just perusing PRs and this one caught my eye. Maybe if we had the numbers spaced out by group so that collisions were less likely during development.

Comment (rawlinp) [reply traffic_ops/traffic_ops_golang/config/config.go]: I was trying really hard to not have to hand-choose every route ID. What if I generated a random set of IDs and basically gave every route a random ID?

**Comment by alficles** [comment]
> It just seems like wasted work

Well, the whole endeavor is kinda a short term patch. But I think that "short term" means "at least a year" here. So it's worth at least a little to consider the ops costs folks will incur over the year. Also, I'm a bit concerned about the opportunity for merge conflicts. Any time you create a registry of ids, you're going to get conflicts, especially at the rate we're developing in this area.

> What if every Route had a "Route ID" that was an arbitrary string

I really like this suggestion. It also solves the merge conflict issue.

> what would you name the 10 different variations of the `federations` endpoints

The name is mostly arbitrary. I'd do something fairly regular, though, like:

get-federations
post-federations
delete-federations
put-federations
get-federations-deliveryservices
post-federations-deliveryservices
delete-federations-deliveryservices
... and so on.

A route-name like that would also be handy for recording which handler was invoked for logging purposes if we wanted later.

File: traffic_ops/traffic_ops_golang/routing/routes.go (hunk 2 above)
Comment (ocket8888) [reply traffic_ops/traffic_ops_golang/routing/routes.go]: I'm fine with that. This was the only endpoint that I found that stood out like this btw, though there may be others. As long as they've been working for some large company in production for a while that's good enough for me.

File: traffic_ops/traffic_ops_golang/config/config.go (hunk 3 above)
Comment (alficles) [reply traffic_ops/traffic_ops_golang/config/config.go]: Hrm. Yeah. There are a lot of routes. I can see how that would be annoying. :/

If we did the string suggestion instead of an integer, you might be able to get away with using a macro to strip out magic regex chars and use that.

But random integers would also reduce the chance of operator error as well. Off-by-one is much more likely than putting the wrong number. You could group ids as well. So that all the routes in a specific category had the same first few digits. (Or same high order bytes.)

**Comment by rawlinp** [comment]
We should keep in mind, not only is at least `perl_routes` meant to be short-lived, both `perl_routes` and `disabled_routes` are meant to be used on rare occasions, so I don't really think it will lead to SRE headaches. That, and giving unique names to each route doesn't really solve the operability problem, because an SRE would still have to look up the route table to match up a particular route with its given name, and copy the given name into the config. That's just as painful as copying numeric IDs into the config. Plus, you could still add the routes with their IDs as a comment in cdn.conf (by using an ignored json field), if you really find it necessary to look at cdn.conf and know exactly what routes are disabled.

File: traffic_ops/traffic_ops_golang/config/config.go (hunk 3 above)
Comment (jhg03a) [reply traffic_ops/traffic_ops_golang/config/config.go]: I'm also not a fan of magic integers.  Why can't we just use a simple glob for the route instead of a full powered regex that's more complicated to read and write?  I added a sample of what pops into my head with https://github.com/apache/trafficcontrol/issues/2872#issuecomment-558273654

**Comment by alficles** [comment]
> an SRE would still have to look up the route table to match up a particular route with its given name

I'm less concerned about the "adding a route" case. Numbers work fine for that, imo. (But your random number suggestion works better in that case, too.)

I'm more concerned about the "trying to figure out why TO isn't behaving the way I expect it to" debugging case. For that, the engineer doing the debugging would need to look up each route id to see which thing it refers to. And while they might have to check the docs to see exactly which of the dozen or so federation routes it refers to, you won't need to check all of them, because you can rule out all the routes that obviously don't apply by name.

With few exceptions, config files are for humans and it's ideal to let them be readable for humans. You changed the regex precisely because of this. It was terrible for humans. This is a step up, but I think we can do slightly better.

> both perl_routes and disabled_routes are meant to be used on rare occasions

Writing features that aren't meant to be used is a path that often leads to long-term suffering. perl_routes should definitely go away (although I'm not sure quite how "soon".) disabled_routes will probably live forever, though.

**Comment by ocket8888** [comment]
One thing I was thinking about is maybe we could rework this into some kind of Go plugin at some point. Once request paths are actually strings, we could configure a plugin to accept a blacklist of endpoints to disable and then we can remove all this ID stuff from `cdn.conf` and have a dedicated plugin for it that uses actual request paths. Seems like a good candidate for it, but we'd also need to make a small improvement to the plugin hooks to allow it to reload its config when TO itself does.

**Comment by rob05c** [comment]
To be clear, I'm -0 on numeric IDs, +0 on string IDs. I don't agree numbers are easier for Dev or Ops, but I don't feel strongly enough to fight for it.

**Comment by rob05c** [comment]
>we could rework this into some kind of Go plugin

That's a great idea. I agree 100%, this is an excellent candidate for a Plugin. But I wouldn't hold up this PR for it, it can be converted later.

>we'd also need to make a small improvement to the plugin hooks to allow it to reload its config when TO itself does.

That's the intention. There are very few Plugin Hooks at the moment, the idea was that we'd add them as we need them. They're easy to add, and I'm happy to help anyone adding one.

**Comment by alficles** [comment]
And since we're being clear... I'm -2 on consecutive IDs (gotta make up numbers here :) ), -1 on random IDs, and +1 on string ids, especially if we can leverage those strings into the future for logging and plugin purposes. But the world won't end if we decide that naming all the routes is more work than we're willing to allocate at the moment and just go with random IDs.

**Comment by ocket8888** [comment]
I guess I'd need to know more about the "random ID" implementation to comment on it. Seems like it'd just make things more confusing and make it even harder to add new routes than sequential IDs.

**Comment by rawlinp** [comment]
All these suggestions are great, but we've successfully migrated the bike-shedding off of the mailing list into this PR. I can re-generate the route IDs to make them random instead of sequential, in order to reduce potential dev headaches due to merge conflicts, but I will leave the rest of the comments that require completely redoing this PR and/or the TO routing code to be done in the future, once our routes are simple paths and not regexes. If anyone feels strongly enough to go through and uniquely name all of our existing routes, replace the IDs, and open a PR before I cut 4.x, I would be open to that. But we are looking to cut 4.x within a week.

File: traffic_ops/traffic_ops_golang/config/config.go (hunk 3 above)
Comment (ocket8888) [reply traffic_ops/traffic_ops_golang/config/config.go]: Matching glob patterns to the set of possible matches for a regular expression sounds difficult IMO. it's fine for things like `/foo/{id}` but you can't use it against things that match like `/foo/?(\.json)?$` without picking the parts of a regular expression apart and deciding which capture groups can be matched in a glob and which can't.

**Comment by alficles** [comment]
I get the concern and will defer to the implementer. But it's worth noting that if we don't do it now, it's going to be much, much harder to do in the future because we'll have backward compatibility concerns.

**Comment by rawlinp** [comment]
I don't think it would be that much harder in terms of backwards-compatibility. We would just have to keep numeric IDs _and_ name-based IDs (or non-regex paths) on the route objects for a release, allowing both integer-based and name-based route IDs in the config (but separate fields). Then we'd tell everyone they have to switch to name-based route IDs before the next release, and rip out the route IDs in the next release.

I know this PR seems a little big at ~800 LOC changed, but ~600 of those is just updating the existing Route objects to add IDs and a boolean field. The code responsible for actually bypassing to Perl and disabling routes was only ~200 LOC changed and are fairly straightforward IMO. So adding support for name-based IDs on top of integer-based IDs would basically be copy/paste of the ID-based code, except checking for string matches instead of integer matches. It would take longer to come up with a unique name for all of our existing endpoints than to add support for disabling/Perling routes on those name-based IDs.

File: traffic_ops/traffic_ops_golang/config/config.go [hunk 5]
```diff
@@ -109,10 +109,12 @@ type ConfigTrafficOpsGolang struct {
 	CRConfigEmulateOldPath bool `json:"crconfig_emulate_old_path"`
 }
 
-// RoutingBlacklist contains the list of route IDs that will be handled by TO-Perl and a list of route IDs that are disabled.
+// RoutingBlacklist contains the list of route IDs that will be handled by TO-Perl, a list of route IDs that are disabled,
+// and whether or not to ignore unknown routes.
 type RoutingBlacklist struct {
-	PerlRoutes     []int `json:"perl_routes"`
-	DisabledRoutes []int `json:"disabled_routes"`
+	IgnoreUnknownRoutes bool  `json:"ignore_unknown_routes"`
```
ocket8888: This will default to "False" then, is that what we want? It means existing configs will use that behavior, but idk if that's necessarily a bad thing because they also wouldn't have "PerlRoutes" or "DisabledRoutes".

traffic_ops/traffic_ops_golang/routing/wrappers.go
```diff
@@ -285,3 +285,11 @@ func NotImplementedHandler() http.Handler {
 		w.Write([]byte(`{"alerts":[{"level":"error","text":"The requested api version is not implemented by this server. If you are using a newer client with an older server, you will need to use an older client version or upgrade your server."}]}`))
 	})
 }
+
+func DisabledRouteHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set(tc.ContentType, tc.ApplicationJson)
+		w.WriteHeader(http.StatusServiceUnavailable)
+		w.Write([]byte(`{"alerts":[{"level":"error","text":"The requested route is currently disabled."}]}`))
```
ocket8888: pls end with newline

but also, this handler isn't setting headers like e.g. `X-Server-Name` and CORS headers. I don't really know about CORS being important at endpoints that specifically aren't allowed to work, but it could be nice to still see what server served your request.

traffic_ops/traffic_ops_golang/traffic_ops_golang.go
```diff
@@ -68,6 +69,29 @@ func main() {
...
+			perlRoutes := routing.GetRouteIDMap(cfg.PerlRoutes)
+			disabledRoutes := routing.GetRouteIDMap(cfg.DisabledRoutes)
+			for _, r := range routes {
+				_, isBypassedToPerl := perlRoutes[r.ID]
+				_, isDisabled := disabledRoutes[r.ID]
+				fmt.Printf("id=%d\tmethod=%s\tversion=%.1f\tpath=%s\tcan_bypass_to_perl=%t\tis_bypassed_to_perl=%t\tis_disabled=%t\n", r.ID, r.Method, r.Version, r.Path, r.CanBypassToPerl, isBypassedToPerl, isDisabled)
+			}
+		} else {
+			for _, r := range routes {
+				fmt.Printf("id=%d\tmethod=%s\tversion=%.1f\tpath=%s\tcan_bypass_to_perl=%t\n", r.ID, r.Method, r.Version, r.Path, r.CanBypassToPerl)
```
ocket8888: Why not just `fmt.Println(r)` since you made the `String` method?

traffic_ops/traffic_ops_golang/config/config.go
rawlinp: I believe this is what we want. If you intentionally put unknown routes in the config, e.g. in anticipation of disabling routes that don't yet exist, you should also make sure they're explicitly ignored. By default, I'd think you'd want to know if you made a typo.

traffic_ops/traffic_ops_golang/traffic_ops_golang.go
rawlinp: the `String` method just uses a single space as a separator, whereas this uses a tab. I think the tab-delimited output looks more readable in the output. I also wanted it to basically match the `Printf` just above this.

traffic_ops/traffic_ops_golang/routing/wrappers.go
rawlinp: For this handler, I basically just copied the same `NotImplementedHandler` which does the exact same thing except with a different status code and text. I didn't feel like wrapping this in the `wrapHeaders` func to set all the extra headers, gzip, sha, etc. was necessary right now, especially when the `NotImplementedHandler` isn't used that way either. If it becomes necessary for clients in the future, we should add it then IMO.

If you so desire, I'll add a newline though.

ocket8888: I desire greatly

traffic_ops/traffic_ops_golang/traffic_ops_golang.go
ocket8888: No tabs are definitely the way to go for tabular data like this. Just missed that the `String` method used spaces.

traffic_ops/traffic_ops_golang/routing/wrappers.go
rawlinp: done

ocket8888:
@alficles if you're good with this the way it is now, can you submit an approving review? I don't wanna merge something out from under an active reviewer.

rawlinp:
Rebased to resolve merge conflict due to the newly-rewritten `current_stats` route.

rawlinp:
Rebased to resolve merge conflict due to `hwinfo` route changes.

traffic_ops/traffic_ops_golang/config/config.go
jhg03a: Randomized, immutable integers were a consensus and are implemented offline.  Can we resolve this?

PR #4152, opened 2019-12-03
Remove some duplicate routes, tighten up some route regexes in routes.go

rawlinp:

Remove some duplicate routes, add a unit test to check for duplicate route problems in the future, and tighten up some rewritten route regexes to match their corresponding Perl routes.

- Traffic Ops

Fixes some API routes, no docs necessary.

Run the TO unit and API tests, verify they pass. Manually query some of the fixed routes to confirm they only accept literal `.config` instead of `<insert any character here>config` and `deliveryservices` instead of either `deliveryservice` or `deliveryservices` (since the original `?` in those routes made the `'s'` optional).

- master


traffic_ops/traffic_ops_golang/routing/routing_test.go
```diff
@@ -174,11 +174,18 @@ func TestCompileRoutes(t *testing.T) {
 
 func TestRoutes(t *testing.T) {
 	fake := ServerData{Config: config.NewFakeConfig()}
-	_, _, _, err := Routes(fake)
+	routes, _, _, err := Routes(fake)
 	if err != nil {
 		t.Fatalf("expected: no error getting Routes, actual: %v", err)
 	}
 	// TODO: verify that all returned Routes are unique
```
ocket8888: You wanna get rid of this TODO?

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -511,17 +508,17 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
...
 
 		// Federation Resolvers
 		{1.1, http.MethodPost, `federation_resolvers(/|\.json)?$`, federation_resolvers.Create, auth.PrivLevelAdmin, Authenticated, nil, 1134373661, perlBypass},
 		{1.1, http.MethodGet, `federation_resolvers(/|\.json)?$`, federation_resolvers.Read, auth.PrivLevelReadOnly, Authenticated, nil, 556608759, perlBypass},
 
 		// Federations Users
-		{1.1, http.MethodPost, `federations/{id}/users?(\.json)?$`, federations.PostUsers, auth.PrivLevelAdmin, Authenticated, nil, 1779334930, perlBypass},
-		{1.1, http.MethodGet, `federations/{id}/users?(\.json)?$`, api.ReadHandler(&federations.TOUsers{}), auth.PrivLevelReadOnly, Authenticated, nil, 394075015, perlBypass},
+		{1.1, http.MethodPost, `federations/{id}/users(\.json)?$`, federations.PostUsers, auth.PrivLevelAdmin, Authenticated, nil, 1779334930, perlBypass},
+		{1.1, http.MethodGet, `federations/{id}/users(\.json)?$`, api.ReadHandler(&federations.TOUsers{}), auth.PrivLevelReadOnly, Authenticated, nil, 394075015, perlBypass},
```
ocket8888: I think the ones with the odd floating <kbd>?</kbd> were meant to be `{etc.}/?(\.json)?$` since the Perl API allowed trailing `/` and trailing `.json`

traffic_ops/traffic_ops_golang/routing/routing_test.go
rawlinp: yes, I do :)

traffic_ops/traffic_ops_golang/routing/routes.go
rawlinp: yeah, probably. I considered that, but we should probably allow an optional trailing `/` like everything else. I'll fix that.

Issue #3067, opened 2018-11-29
Add cachegroup should not default to deep routing enabled


dneuman64:
When you add a cachegroup, Traffic Portal has the box selected for deep coverage zone file enabled by default.  Since deep coverage zone is not a default feature, this box should not be checked by default.

![image](https://user-images.githubusercontent.com/5092346/49248548-36efbf00-f3d7-11e8-86df-a76cb69c5432.png)

rawlinp:
It shouldn't hurt anything to have DCZF checked by default. The default behavior is to allow all localization methods for a given cachegroup. Unfortunately, due to "Deep Cachegroups" not being "real" cachegroups (i.e. they're defined only in the DCZF as opposed to in Traffic Ops), you can't really disable deep routing to a deep cachegroup. But by definition a deep cachegroup should always be enabled for deep-routing.

So maybe what we really need to do here is just remove the checkbox for DCZF from TP altogether since it doesn't really apply.

mitchell852:
^^ how does that sound @dneuman64 ?

dneuman64:
Sorry, this was missed a looonnnggg time ago.  I think what Rawlin is proposing is fine.  I just wanted to avoid unnecessary confusion for the user.

mitchell852:
> So maybe what we really need to do here is just remove the checkbox for DCZF from TP altogether since it doesn't really apply.

@rawlinp - Just so I understand correctly. If i remove the `deep coverage zone file` checkbox from TP, what value should be sent on cachegroup create?

DEEP_CZ: true or false?

sounds like true as all cache groups are capable of being deep cz routable....assuming you stick them in the DCZF....

rawlinp:
Under the hood it's an array not a boolean, but I'd say just always include it in the array in the request sent by TP, and remove the deep_cz checkbox altogether.

PR #4154, opened 2019-12-04
Added Cache Groups overview section

ocket8888:
Adds a Cache Groups subsection to the Overview section of the documentation. Also cleans up some less-than-excellent formatting and removes a slew of false deprecation notices as well as linking the new subsubsections of the CG overview all over the docs.

- Documentation

Build and read the docs, ensure no warnings, and that all information is comprehensible, complete, and renders well in the HTML output.

No test beacuase is docs

- [x] Tests are unnecessary


docs/source/overview/cache_groups.rst
```diff
@@ -31,4 +49,273 @@ All servers have to have a :term:`Physical Location`, which defines their geogra
...
+	+-----------------------+----------------------+----------------------------------------+
+
+.. _cache-group-localization-methods:
+
+Localization Methods
+--------------------
+The :dfn:`Localization Methods` of a Cache Group define the methods by which Traffic Router is allowed to route clients to :term:`cache servers` within this Cache Group. This is a collection of the allowed methods, and the values in the collection are restricted to the following.
+
+- "Coverage Zone File" (alias ``CZ`` in source code, database entries, and :ref:`to-api` requests/responses) allows Traffic Router to direct clients to this Cache Group if they were assigned a geographic location by looking up their IP address in the :term:`Coverage Zone File`.
+- "Deep Coverage Zone File" (alias ``DEEP_CZ`` in source code, database entries, and :ref:`to-api` requests/responses) allows Traffic Router to direct clients to this Cache Group if they were assigned a geographic location by looking up their IP addresses in the :term:`Deep Coverage Zone File`.
```
mitchell852: This is going to be removed from TP per #3067. want to make the necessary changes to these docs?

docs/source/overview/cache_groups.rst
```diff
@@ -31,4 +49,273 @@ All servers have to have a :term:`Physical Location`, which defines their geogra
...
+""""""""
+This :term:`Type` of Cache Group contains :term:`Edge-tier cache servers`.
+
+MID_LOC
+"""""""
+This :term:`Type` of Cache Group contains :term:`Mid-tier cache servers`
+
+ORG_LOC
+"""""""
+This :term:`Type` of Cache Group contains :term:`origins`. The primary purpose of these is to group :term:`origins` for the purposes of "multi-site-origins", and it's suggested that if that doesn't meet your use-case that these be mostly avoided. In general, it's not strictly necessary to create :term:`origin` *servers* in ATC at all, unless you have to support "multi-site-origins".
```
mitchell852: i don't suppose there is any documentation regarding msos that you can link to

ocket8888: There's a guide on setting them up, that's what most things link to. It doesn't feel totally appropriate when you just wanna know what it is, but it's all we got.

docs/source/overview/cache_groups.rst
ocket8888: I'm hesitant to do that. Because I don't want the docs for something to get into a release where it's wrong - if #3067 merges before this one I'll be happy to update it, but if this merges first #3067 should do it. That way the docs are always exactly correct for the features that exist.

mitchell852: makes sense

docs/source/overview/cache_groups.rst
mitchell852: i think a link to the setup guide makes sense probably.

ocket8888: right, that's in the seealso box right under that paragraph

mitchell852: oh ok. i missed that.

docs/source/overview/cache_groups.rst
mitchell852: deep czf checkbox was removed from the cachegroup form with #4172

ocket8888: But it's still present in API responses? Then this really oughtta be left in. But if it truly doesn't affect anything I need to add that as a note or something. In fact I think right now I assert that it does in fact affect things.

mitchell852: yeah, you probably needs some kind of note in there because it's not as simple as setting cachegroup.localizationMethods = ['DEEP_CZ'] to enable deep routing on a cache group.

docs/source/overview/cache_groups.rst
```diff
@@ -31,4 +49,273 @@ All servers have to have a :term:`Physical Location`, which defines their geogra
...
+
+Localization Methods
+--------------------
+The :dfn:`Localization Methods` of a Cache Group define the methods by which Traffic Router is allowed to route clients to :term:`cache servers` within this Cache Group. This is a collection of the allowed methods, and the values in the collection are restricted to the following.
+
+- "Coverage Zone File" (alias ``CZ`` in source code, database entries, and :ref:`to-api` requests/responses) allows Traffic Router to direct clients to this Cache Group if they were assigned a geographic location by looking up their IP address in the :term:`Coverage Zone File`.
+- "Deep Coverage Zone File" (alias ``DEEP_CZ`` in source code, database entries, and :ref:`to-api` requests/responses) allows Traffic Router to direct clients to this Cache Group if they were assigned a geographic location by looking up their IP addresses in the :term:`Deep Coverage Zone File`.
+
+	.. warning:: In order to make use of "deep caching" for a :term:`Delivery Service`, this localization method must be enabled on the Cache Groups that contain servers assigned to said :term:`Delivery Service` **and** that :term:`Delivery Service` must have :ref:`ds-deep-caching` enabled.
```
rawlinp: I don't think this is really true. For example, let's say cache1 is in cachegroup1. Cachegroup1 only has CZ and GEO enabled. Let's say cache1 is in a "deep" cachegroup per the DCZF. Cache1 will still be routed to as part of the "deep" cachegroup, even though cache1's "real" cachegroup -- cachegroup1 -- doesn't have DEEP_CZ enabled. This is because "deep" cachegroups do not map to "real" cachegroups in TO -- they are a concept only within the DCZF.

Basically, I think this warning, along with the "deep coverage zone file" bullet point, can be removed entirely. It can be enabled for a TO cachegroup, but it doesn't actually change/do anything for that cachegroup.

ocket8888: How do you feel about what I just pushed up?

rawlinp: better

PR #4160, opened 2019-12-04
Targeted cleanup

ChrisHines:
This PR fixes a couple issues flagged by staticcheck, but mostly it improves readability and idiomaticness of a limited portion of the code I've been reading while working on a Perl to Go rewrite.

- Traffic Ops

Look at the diffs and ensure CI builds pass.

### DIFF
--- a/traffic_ops/client/server_update_status.go
+++ b/traffic_ops/client/server_update_status.go
@@ -32,7 +32,7 @@ func (to *Session) UpdateServerStatus(serverID int, req tc.ServerPutStatus) (*tc
 		return nil, reqInf, err
 	}
 
-	path := fmt.Sprintf(apiBase+"/servers/%d/status", serverID)
+	path := fmt.Sprintf("%s/servers/%d/status", apiBase, serverID)
 	resp, remoteAddr, err := to.request(http.MethodPut, path, reqBody)
 	reqInf.RemoteAddr = remoteAddr
 	if err != nil {
--- a/traffic_ops/client/server_update_status.go
+++ b/traffic_ops/client/server_update_status.go
@@ -18,33 +18,32 @@ package client
 import (
 	"encoding/json"
 	"fmt"
-	"net"
 	"net/http"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
 )
 
-const (
-	APIUpdateServerStatus = apiBase + "/servers/%d/status"
-)
-
 // UpdateServerStatus updates a server's status and returns the response.
 func (to *Session) UpdateServerStatus(serverID int, req tc.ServerPutStatus) (*tc.Alerts, ReqInf, error) {
-	var remoteAddr net.Addr
+	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss}
+
 	reqBody, err := json.Marshal(req)
-	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
 	if err != nil {
 		return nil, reqInf, err
 	}
-	path := fmt.Sprintf(APIUpdateServerStatus, serverID)
+
+	path := fmt.Sprintf(apiBase+"/servers/%d/status", serverID)
 	resp, remoteAddr, err := to.request(http.MethodPut, path, reqBody)
+	reqInf.RemoteAddr = remoteAddr
 	if err != nil {
 		return nil, reqInf, err
 	}
 	defer resp.Body.Close()
+
 	alerts := tc.Alerts{}
 	if err = json.NewDecoder(resp.Body).Decode(&alerts); err != nil {
 		return nil, reqInf, err
 	}
+
 	return &alerts, reqInf, nil
 }
--- a/traffic_ops/client/session.go
+++ b/traffic_ops/client/session.go
@@ -13,6 +13,7 @@
    limitations under the License.
 */
 
+// Package client provides Go bindings to the Traffic Ops RPC API.
 package client
 
 import (
--- a/traffic_ops/traffic_ops_golang/server/servers.go
+++ b/traffic_ops/traffic_ops_golang/server/servers.go
@@ -1,3 +1,5 @@
+// Package server provides tools for manipulating the server database table and
+// corresponding http handlers.
 package server
 
 /*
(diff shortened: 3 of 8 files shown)


traffic_ops/client/server_update_status.go
```diff
@@ -18,33 +18,32 @@ package client
...
+	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss}
+
 	reqBody, err := json.Marshal(req)
-	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
 	if err != nil {
 		return nil, reqInf, err
 	}
-	path := fmt.Sprintf(APIUpdateServerStatus, serverID)
+
+	path := fmt.Sprintf(apiBase+"/servers/%d/status", serverID)
```
ocket8888: bit of a nitpick but if you're already doing a `fmt.Sprintf` you could just fully utilize it like
```go
path := fmt.Sprintf("%s/servers/%d/status", apiBase, serverID)
```
 instead of concatenating a string first. I'm sure the compiler will sort it out, but it's just weird to see two different string building methods in the same expression.

traffic_ops/client/session.go
```diff
@@ -13,6 +13,7 @@
    limitations under the License.
 */
 
+// Package client provides Go bindings to the Traffic Ops RPC API.
```
ocket8888: "RPC"?

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -27,104 +29,103 @@ import (
...
+
+	validation "github.com/go-ozzo/ozzo-validation"
 	"github.com/go-ozzo/ozzo-validation/is"
 	"github.com/jmoiron/sqlx"
+	"github.com/lib/pq"
 )
 
-//we need a type alias to define functions on
+// A TOServer combines data about a server with metadata from an API request and
+// provides methods that implement several interfaces from the api package.
```
ocket8888: Shouldn't this just start with `TOServer`?

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -27,104 +29,103 @@ import (
...
 }
 
-func (v *TOServer) SetLastUpdated(t tc.TimeNoMod) { v.LastUpdated = &t }
-func (v *TOServer) InsertQuery() string           { return insertQuery() }
-func (v *TOServer) UpdateQuery() string           { return updateQuery() }
-func (v *TOServer) DeleteQuery() string           { return deleteQuery() }
+func (s *TOServer) SetLastUpdated(t tc.TimeNoMod) { s.LastUpdated = &t }
+func (s *TOServer) InsertQuery() string           { return insertQuery() }
+func (s *TOServer) UpdateQuery() string           { return updateQuery() }
+func (s *TOServer) DeleteQuery() string           { return deleteQuery() }
```
ocket8888: I actually don't think these last three need reference bindings at all, they could just be:
```go
func (*TOServer) InsertQuery() string           { return insertQuery() }
func (*TOServer) UpdateQuery() string           { return updateQuery() }
func (*TOServer) DeleteQuery() string           { return deleteQuery() }
```

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -27,104 +29,103 @@ import (
...
 
-func (server TOServer) GetKeyFieldsInfo() []api.KeyFieldInfo {
+func (s TOServer) GetKeyFieldsInfo() []api.KeyFieldInfo {
```
ocket8888: same as above RE: reference binding

traffic_ops/traffic_ops_golang/server/servers_update_status_test.go
```diff
@@ -48,8 +47,7 @@ func TestGetServerUpdateStatus(t *testing.T) {
 	mock.ExpectQuery("SELECT").WillReturnRows(serverStatusRow)
 	mock.ExpectCommit()
 
-	dbCtx, _ := context.WithTimeout(context.TODO(), time.Duration(10)*time.Second)
-	tx, err := db.BeginTx(dbCtx, nil)
+	tx, err := db.BeginTx(context.Background(), nil)
```
ocket8888: you've eliminated the timeout here, is that going to be alright?

traffic_ops/client/session.go
ChrisHines: Remote Procedure Call (as opposed to a native Go API for example).

traffic_ops/traffic_ops_golang/server/servers.go
ChrisHines: Good point.

traffic_ops/traffic_ops_golang/server/servers_update_status_test.go
ChrisHines: It's a test. `go test` has a built in timeout and the user can set it from the command line with the `-timeout` flag.

`go vet` flagged this code with the message:

>the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak

It seemed simpler to me to just remove the timeout since the test is not testing the timeout behavior in any way.

ocket8888: As long as we know the timeout isn't being used in the tests that's fine by me

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -27,104 +29,103 @@ import (
...
 	"github.com/apache/trafficcontrol/lib/go-util"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
-
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/tenant"
-	"github.com/go-ozzo/ozzo-validation"
+
+	validation "github.com/go-ozzo/ozzo-validation"
```
ocket8888: Is this necessary? The package name is already `validation`.

ChrisHines: This was done automatically by `gopls`. I believe the tool does that because the package name differs significantly from the the last part of the import path. I guess the tool is trying to show the actual package name explicitly in the code since the package doesn't follow the Go convention of the package name matching the last segment of the import path.

ocket8888: I don't think a language server ought to be making stylistic changes without user input or configuration in the tools that use it, but I suppose this isn't hurting anything. Just to be clear though, we don't enforce `gopls` suggestions/edits.

PR #4171, opened 2019-12-05
Change Traffic Ops middleware wrapper symbols to be public

rob05c:
Changes Traffic Ops middleware wrapper symbols to be public.

The primary goal of this is to make them accessible to plugins,
which will almost always want to add the access control, checksum,
server, and gzip headers.

To that end, this also includes an example plugin using middleware.

This doesn't change any code, it only moves and capitalizes existing symbols.

No new tests, no interface change and not a bug fix.

- Traffic Ops

Run unit and API tests, verify they still pass.
Make some requests, verify things work as expected, still return middleware headers appropriately.

Not a bug fix.

## Additional Information

### DIFF
--- a/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go
+++ b/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go
@@ -1,4 +1,4 @@
-package routing
+package middleware
 
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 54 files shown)


traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go
```diff
@@ -39,23 +39,39 @@ import (
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/tocookie"
 )
 
+const DefaultRequestTimeout = time.Second * time.Duration(60)
+
+// Middleware ...
```
rawlinp: While you're moving things around, do you mind adding godoc comments to `Middleware`, `GetDefault`, and `Use`?

rob05c: Done. The middleware package now has full GoDoc.

I also moved the RFC and utility symbols that were in there to lib/go-util and lib/go-rfc where they belong. Some of the RFC funcs were using RFC things in lib/go-tc, so I also had to move those to lib/go-rfc. It was a string I didn't intend to pull, but it needed done anyway, and it's done now.

If it's easier to review, I made a new changeset, you should be able to just review https://github.com/apache/trafficcontrol/pull/4171/commits/44bce47c04df52f3a7721f62ee34c8ffd0f19998 - I didn't modify the original changeset.

rob05c:
> which longer -> which no longer

Fixed

PR #4173, opened 2019-12-06
Rewrite stats summary post to golang

### LINKED ISSUE #3833: Rewrite /stats_summary/create to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/stats_summary/create` is handled by Perl

## Expected / new behavior:
`/stats_summary/create` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [x] `POST`

N/A

mhoppa:

- [x] This PR does not have a related issue 

Rewrite /stats_summary/create into a new route of /stats_summary 

- Documentation
- Traffic Control Client 
- Traffic Ops
- Traffic Stats


traffic_ops/client/stats_summary.go
```diff
@@ -72,6 +72,6 @@ func (to *Session) CreateSummaryStats(statsSummary tc.StatsSummary) (tc.Alerts,
 	if err != nil {
 		return tc.Alerts{}, reqInf, err
 	}
-	reqInf, err = post(to, fmt.Sprintf("%s/%s", basePathSummaryStats, "create"), reqBody, &alerts)
+	reqInf, err = post(to, basePathSummaryStats, reqBody, &alerts)
```
ocket8888: If you're going to change this client method then it needs to exactly match the behavior of the old endpoint. This one doesn't, because `summaryTime` and `statDate` aren't optional anymore.

docs/source/api/stats_summary.rst
```diff
@@ -173,4 +176,74 @@ Last Updated Summary Stat
 
 	{ "response": {
 		"summaryTime": "2019-11-19 00:04:06+00"
-	}}
\ No newline at end of file
+	}}
+
+``POST``
+.. versionadded:: 1.4.1
```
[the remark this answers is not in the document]
mhoppa: 🤦‍♂

docs/source/api/stats_summary.rst
```diff
@@ -173,4 +176,74 @@ Last Updated Summary Stat
...
+
+:Auth. Required: Yes
+:Roles Required: None
+:Response Type: Object
+
+Request Structure
+-----------------
+:cdnName:             The CDN name for which the summary stat was taken for
+
+	.. note:: If the ``cdn`` is equal to ``all`` it represents summary_stats across all delivery services within the given CDN
```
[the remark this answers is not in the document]
mhoppa: good catch yes

traffic_ops/client/stats_summary.go
mhoppa: statDate is optional and summaryTime has never been optional in my testing I get a 500 when testing it on the old stats_summary/create if I dont pass it in

mhoppa: nvm talked to @ocket8888 offline and he pointed at the differences that I need to match

mhoppa: Okay looking at this a bit I dont think this needs to match the behavior of the old endpoint as we decided on the mailing list we do not need deprecation periods for TO client methods. 

The problem is to match the old behavior you are seeing I would have to change the underlying struct (or make a new one) to take in strings instead of time.Time as you pointed out strings like "today", "yesterday" work. 

I think leaving this as is is fine imo. I guess I could make another client method with a different struct input and post to the old route. Thoughts?

ocket8888: > _"...we decided on the mailing list we do not need deprecation periods for TO client methods."_

That's good enough for me.

docs/source/api/stats_summary.rst
```diff
@@ -173,4 +176,76 @@ Last Updated Summary Stat
...
```
ocket8888: This still says that `"cdn": "all"` means that the stats are for "... all delivery services within the given CDN"

mhoppa: 🤦‍♂

mhoppa:
@ocket8888 once you get a chance can you re review?? :)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Traffic Ops Golang Endpoints
   - /api/1.1/cachegroupparameters/{{cachegroupID}}/{{parameterID}} `(DELETE)`
   - /api/1.5/to_extensions `(POST)`
+  - /api/1.5/stats_summary `(POST)`
```
ocket8888: This should probably also include a deprecation notice for `stats_summary/create` which this route is replacing.

mhoppa:
https://github.com/apache/trafficcontrol/pull/4186 already posted waiting for this to merge let me remove the issue from the pull request template so it doesnt close it

PR #4181, opened 2019-12-09
Add Go implementation of POST servers/{id}/queue_update

### LINKED ISSUE #3828: Rewrite /servers/{{id}}/queue_update to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/servers/{{id}}/queue_update` is handled by Perl

## Expected / new behavior:
`/servers/{{id}}/queue_update` should be handled by Go.

### Minimum API Versions
1.1

### Request Methods
- [ ] `POST`

N/A

ChrisHines:
- [x] This PR fixes #3828

- Traffic Ops

Run the added API test against both the Perl and Go backends.

- master (25181a650cc236028a9f7f9154b5a8919abfa3d0)

### DIFF
--- a/traffic_ops/traffic_ops_golang/server/queue_update.go
+++ b/traffic_ops/traffic_ops_golang/server/queue_update.go
@@ -1,0 +1,93 @@
+package server
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import (
+	"database/sql"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/lib/go-util"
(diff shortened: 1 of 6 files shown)


rawlinp:
add to whitelist

ChrisHines:
> add to whitelist

Is that a request to change something? If so, where is the whitelist and what needs to change?

rawlinp:
That comment was to add your user to the TC asf-ci whitelist so that the PR builder CI will run against your PRs.

traffic_ops/traffic_ops_golang/server/queue_update.go
```diff
@@ -0,0 +1,93 @@
...
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusNotFound, nil, nil)
+		return
+	}
+
+	api.WriteResp(w, r, tc.ServerQueueUpdate{
+		ServerID: util.JSONIntStr(serverID),
+		Action:   reqObj.Action,
+	})
+
+	api.CreateChangeLogBuildMsg(
```
mhoppa: we should catch an error here and return internal server error to the user (prior to writing the response) via the api.HandleErr(....) func

traffic_ops/traffic_ops_golang/server/queue_update.go
```diff
@@ -0,0 +1,93 @@
...
+
+	serverID := int64(inf.IntParams["id"])
+	queue := reqObj.Action == "queue"
+	ok, err := queueUpdate(inf.Tx.Tx, serverID, queue)
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, fmt.Errorf("queueing updates: %v", err))
+		return
+	}
+	if !ok {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusNotFound, nil, nil)
```
mhoppa: nit but I wonder if we include the server id in the user id -> server id {id} not found

traffic_ops/traffic_ops_golang/server/queue_update.go
```diff
@@ -0,0 +1,93 @@
...
+		api.Updated,
+		inf.User,
+		inf.Tx.Tx,
+		"server",
+		fmt.Sprint(serverID),
+		map[string]interface{}{"id": serverID},
+	)
+}
+
+func queueUpdate(tx *sql.Tx, serverID int64, queue bool) (bool, error) {
```
awilliams: Consider a comment describing what the `bool` return value represents.

Issue #3533, opened 2019-04-29
TP: Can add an empty cache group fallback


mitchell852:
When creating or updating a cachegroup, you can add fallback cachegroups, however, if you select "Add failover cachegroup", then you end up with an "empty" cg fallback.

![image](https://user-images.githubusercontent.com/251272/56916096-5b7ac080-6a75-11e9-8ea8-12c2235d10aa.png)

![image](https://user-images.githubusercontent.com/251272/56916118-6a617300-6a75-11e9-87a1-461aada575f5.png)

mitchell852:
I think @mattjackson220 is going to look at this

lbathina:
issue is fixed however, the ```Add Failover cache group...``` should be the default option and blank shouldn't be displayed.

mitchell852:
> issue is fixed however, the `Add Failover cache group...` should be the default option and blank shouldn't be displayed.

@lbathina - i fixed that in this PR #4178

PR #4190, opened 2019-12-09
TP: limits delivery service cloning to source ds type = target ds type

### LINKED ISSUE #3403: Cloning a Steering DS from an non-steering DS copies origin URL
Steps followed:

* Open a DS of type HTTP_LIVE_NATNL
* Click ‘More’ button in upper right corner, then select ‘Clone Delivery Service’
* Select ‘Steering’, then ‘Submit’
* Under ‘Content Routing Type’ select ‘CLIENT_STEERING’
* Update the ‘Key(XML ID)’ and ‘Display Name’
* Click ‘Clone’ in the bottom right corner

The result appears to copy the origin URL from the original DS that was cloned, even though a Client Steering DS has no origin URL. This can be seen from the view of all Delivery Services in TP.

I marked this as minor because I'm pretty sure this doesn't cause a break in functionality

mitchell852:

- [x] This PR fixes #3403 

because different ds "categories" (anymap, dns, http or steering) have different attributes, cloning a ds needs to be limited where source ds and target ds have types of the same category.

for example, you should not be able to clone a DNS* ds to an STEERING ds because that would add attributes to the steering ds that are not applicable (i.e. origin fqdn).

- Traffic Portal

- navigate to a ds
- click More > Clone Delivery Service

you no longer have the option to choose the category of the target ds. Instead, the type of the source ds will be used to create the clone, however, that type can be changed but must be within a "category". Here are the categories and associated types:

DNS
- DNS
- DNS_LIVE
- DNS_LIVE_NATNL

HTTP
- HTTP
- HTTP_LIVE
- HTTP_LIVE_NATNL
- HTTP_NO_CACHE

ANY_MAP

STEERING
- STEERING
- CLIENT_STEERING

With this change you can no longer clone ds's across categories. You must stay within the category of the source ds.

### DIFF
--- a/traffic_portal/app/src/common/modules/form/cacheGroup/FormCacheGroupController.js
+++ b/traffic_portal/app/src/common/modules/form/cacheGroup/FormCacheGroupController.js
@@ -95,7 +95,7 @@ var FormCacheGroupController = function(cacheGroup, types, cacheGroups, $scope, 
         }
     };
 
-    $scope.fallbackSelected = '';
+    $scope.fallbackSelected = null;
 
     $scope.draggedFallback = '';
 
@@ -123,7 +123,7 @@ var FormCacheGroupController = function(cacheGroup, types, cacheGroups, $scope, 
                 break;
             }
         }
-        $scope.fallbackSelected = '';
+        $scope.fallbackSelected = null;
     };
 
     $scope.updateForNewType = function() {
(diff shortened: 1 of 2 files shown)


traffic_portal/app/src/common/modules/form/cacheGroup/FormCacheGroupController.js
```diff
@@ -95,7 +95,7 @@ var FormCacheGroupController = function(cacheGroup, types, cacheGroups, $scope,
         }
     };
 
-    $scope.fallbackSelected = '';
+    $scope.fallbackSelected = null;
```
rawlinp: do these two changes fix a different bug?

mitchell852: yeah, i snuck them in there :)

https://github.com/apache/trafficcontrol/issues/3533#issuecomment-562677120

PR #4191, opened 2019-12-09
Add Go implementation of GET, POST federations/{id}/federation_resolvers

### LINKED ISSUE #3795: Rewrite /federations/{{id}}/federation_resolvers to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/federations/{{id}}/federation_resolvers` is handled by Perl

## Expected / new behavior:
`/federations/{{id}}/federation_resolvers` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`
- [ ] `POST`

N/A

lactose:
This is a rewrite of the perl endpoints GET,POST for /federations/{id}/federation_resolvers

- [x] This PR fixes #3795

- Traffic Ops

Run the added API test against both the Perl and Go backends.


traffic_ops/traffic_ops_golang/federations/federation_resolvers.go
```diff
@@ -0,0 +1,92 @@
...
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
+		return
+	}
+	defer inf.Close()
+
+	fedID := inf.IntParams["id"]
+	frs, _, err := dbhelpers.GetFederationResolversByFederationID(inf.Tx.Tx, fedID)
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, fmt.Errorf("database exception: %v", err), nil)
```
ocket8888: You shouldn't disclose the details of a database error to the client, system errors like that should go in the last parameter of `api.HandleErr` and then you can set the "user error" to just `nil`.

traffic_ops/traffic_ops_golang/federations/federation_resolvers.go
```diff
@@ -0,0 +1,92 @@
...
+	inf, userErr, sysErr, errCode := api.NewInfo(r, []string{"id"}, []string{"id"})
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
+		return
+	}
+	defer inf.Close()
+
+	var reqObj tc.AssignFederationResolversRequest
+	if err := json.NewDecoder(r.Body).Decode(&reqObj); err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, fmt.Errorf("malformed JSON: %v", err), nil)
```
ocket8888: This is a client error; `500 Internal Server Error` is inappropriate here. This should just be using `http.StatusBadRequest`.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -290,6 +290,57 @@ func GetDSNameAndCDNFromID(tx *sql.Tx, id int) (tc.DeliveryServiceName, tc.CDNNa
 	return dsName, cdnName, true, nil
 }
 
+// GetFederationResolversByFederationID fetches all of the federation resolvers currently assigned to a federation.
```
ocket8888: True, but it returns three things and this only describes one of them.

ocket8888: It also sort of seems like since this returns a slice there's no real point to the second return value, which is `false` if and only if an error occurred. Typically that's used to indicate whether or not the requested object exists, but in this case if there are no rows you'd just return an empty slice.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -290,6 +290,57 @@ func GetDSNameAndCDNFromID(tx *sql.Tx, id int) (tc.DeliveryServiceName, tc.CDNNa
...
+			return nil, false, fmt.Errorf(
+				"error scanning federation_resolvers rows for federation ID [%d]: %s", fedID, err.Error(),
+			)
+		}
+		resolvers = append(resolvers, fr)
+	}
+	return resolvers, true, nil
+}
+
+// GetFederationNameFromID returns the federation's name.
```
ocket8888: as above there are other return values

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -290,6 +290,57 @@ func GetDSNameAndCDNFromID(tx *sql.Tx, id int) (tc.DeliveryServiceName, tc.CDNNa
...
+func GetFederationNameFromID(id int, tx *sql.Tx) (string, bool, error) {
+	var name string
+	if err := tx.QueryRow(`SELECT cname from federation where id = $1`, id).Scan(&name); err != nil {
+		return name, false, fmt.Errorf(
+			"error querying federation name from id [%d]: %s", id, err.Error(),
```
ocket8888: You should check here for the error's equality with `database/sql.ErrNoRows` and in that event return `"", false, nil` to indicate that no actual error occurred, just that the requested object didn't exist. That allows the caller to act more intelligently without having to parse the error or check this equality itself - which is the ultimate point of that boolean return value.

traffic_ops/traffic_ops_golang/federations/federation_resolvers.go
```diff
@@ -0,0 +1,92 @@
...
+		return
+	}
+
+	fedID := inf.IntParams["id"]
+	name, _, err := dbhelpers.GetFederationNameFromID(fedID, inf.Tx.Tx)
```
ocket8888: This second return value shouldn't be ignored, assuming my earlier advice is taken regarding its use.

traffic_ops/traffic_ops_golang/federations/federation_resolvers.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, fmt.Errorf("database exception: %v", err), nil)
```
ocket8888: This is a server error; `400 Bad Request` is inappropriate here. This should just be using `http.StatusInternalServerError`.

Also, as above, don't expose database error details to the client, the "user error" in this case should be `nil` and the database error ultimately passed as the "system error". The client _could_ be in error if they requested assignment to a federation that doesn't exist, which is why we use that boolean return value to easily distinguish between that case and when an internal error occurs.

traffic_ops/traffic_ops_golang/federations/federation_resolvers.go
```diff
@@ -0,0 +1,92 @@
...
+		return
+	}
+
+	if reqObj.Replace {
+		if _, err := inf.Tx.Tx.Exec(deleteFederationFederationResolversQuery, fedID); err != nil {
+			api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, fmt.Errorf("database exception: %v", err), nil)
```
ocket8888: Same as above RE: wrong response code and exposing raw DB errors

traffic_ops/traffic_ops_golang/federations/federation_resolvers.go
```diff
@@ -0,0 +1,92 @@
...
+			return
+		}
+	}
+
+	for _, id := range reqObj.FedResolverIDs {
+		if _, err := inf.Tx.Tx.Exec(associateFederationWithResolverQuery, fedID, id); err != nil {
+			api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, fmt.Errorf("database exception: %v", err), nil)
```
ocket8888: Some as above RE: wrong response code and exposing raw DB errors

traffic_ops/traffic_ops_golang/federations/federation_resolvers.go
```diff
@@ -0,0 +1,92 @@
...
+			return
+		}
+	}
+
+	api.WriteRespAlertObj(
+		w, r, tc.SuccessLevel,
+		fmt.Sprintf("%d resolver(s) were assigned to the %s federation", len(reqObj.FedResolverIDs), name),
+		tc.AssignFederationFederationResolversResponse{Response: reqObj},
+	)
```
ocket8888: A changelog message should also be added - e.g. https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/cachegroup/dspost.go#L109

traffic_ops/traffic_ops_golang/federations/federations.go
```diff
@@ -58,6 +58,11 @@ VALUES ($1, $2)
 ON CONFLICT DO NOTHING
 `
 
+const deleteFederationFederationResolversQuery = `
+DELETE FROM federation_federation_resolver ffr
+WHERE ffr.federation = $1
+`
```
ocket8888: I think this ought to go in the same file where it's actually used

.golangci.yml
```diff
@@ -1,38 +0,0 @@
...
-
-linters:
-  disable-all: true
-  enable:
-    - govet
-    - errcheck
-    - structcheck
-    - golint
-    - gofmt
-    - staticcheck
```
ocket8888: I don't think you meant to delete this file

traffic_ops/testing/api/v14/cdnfederations_test.go
```diff
@@ -107,6 +116,111 @@ func GetTestCDNFederations(t *testing.T) {
...
 }
 
+func AssignTestFederationFederationResolvers(t *testing.T) {
+	// Setup
+	if len(fedIDs) < 1 {
+		t.Fatal("no federations to test")
+	}
+
+	frCnt := len(testData.FederationResolvers)
+	if frCnt < 1 {
```
ocket8888: You're still referencing `frIDs[1]` so this should make sure there are at least _two_ Federation Resolvers in the test data.

traffic_ops/traffic_ops_golang/federations/federation_resolvers.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	fedID := inf.IntParams["id"]
+	name, ok, err := dbhelpers.GetFederationNameFromID(fedID, inf.Tx.Tx)
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, fmt.Errorf("database exception: %v", err))
+		return
+	}
+	if !ok {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, fmt.Errorf("'%d': no such Federation", fedID), nil)
```
ocket8888: This is fine, but I think `http.StatusNotFound` is better. Just a suggestion for you while I test.

mitchell852:
@lactose - i'm seeing a couple of failing API tests:

```
=== RUN   TestCDNFederations/successfully_get_federation_federation_resolvers_for_a_federation_with_some
    --- FAIL: TestCDNFederations/successfully_get_federation_federation_resolvers_for_a_federation_with_some (0.02s)
        cdnfederations_test.go:222: expected federation of ID 19 to have associated federation resolvers, but had 0
```

```
=== RUN   TestFederationFederationResolvers/Fail_to_assign_federation_resolver_to_a_federation_when_federation_does_not_exist
    --- FAIL: TestFederationFederationResolvers/Fail_to_assign_federation_resolver_to_a_federation_when_federation_does_not_exist (0.01s)
        cdnfederations_test.go:186: error: expected error result 404 Not Found[404] - Error requesting Traffic Ops https://localhost:8443/api/1.4/federations/-1/federation_resolvers {"alerts":[{"text":"'-1': no such Federation","level":"error"}]}
            , want: Resource not found
```

lactose:
@mitchell852 tests are now working

mitchell852:
@lactose - i'm seeing a difference between the Perl's behavior and the Go's behavior:

Perl:

GET api/1.4/federations/1/federation_resolvers where no fed resolvers are found returns:

`{"response":[]}`

Go:

GET api/1.4/federations/1/federation_resolvers where no fed resolvers are found returns:

`{"response":null}`

lactose:
@mitchell852 pushed an update

PR #4196, opened 2019-12-10
TP: Provides the ability to view all users for a given role

mitchell852:

1. allows the result of GET /api/users to be filtered by role
2. Provides a view in TP to view users of a specific role

- Documentation
- Traffic Ops
- Traffic Portal

1. Run the TO API tests and ensure they all pass
2. Launch TP and navigate to `User Admin > Roles > Role` and then select `View Users` and ensure that only users of the selected role are displayed.

### DIFF
--- a/docs/source/api/users.rst
+++ b/docs/source/api/users.rst
@@ -31,8 +31,6 @@ Request Structure
 -----------------
 .. table:: Request Query Parameters
 
-.. versionadded:: ATCv4.1 The ``role`` query parameter was added in version 4.1 of :abbr:`ATC (Apache Traffic Control)`.
-
 	+-----------+----------+------------------------------------------------------------------------------------------+
 	| Name      | Required | Description                                                                              |
 	+===========+==========+==========================================================================================+
@@ -59,8 +57,9 @@ Request Structure
 	|           |          | parameter has no effect. ``limit`` must be defined to make use of ``page``.              |
 	+-----------+----------+------------------------------------------------------------------------------------------+
 
-.. versionadded:: 1.4
-	The ``id`` and ``username`` query parameters were added in the 1.4 API.
+.. versionadded:: ATCv4.1 The ``role`` query parameter was added in version 4.1 of :abbr:`ATC (Apache Traffic Control)`.
+
+.. versionadded:: ATCv4.0 The ``id`` and ``username`` query parameters were added in version 4.0 of :abbr:`ATC (Apache Traffic Control)`.
 
 .. code-block:: http
 	:caption: Request Example
--- a/docs/source/api/users.rst
+++ b/docs/source/api/users.rst
@@ -31,6 +31,8 @@ Request Structure
 -----------------
 .. table:: Request Query Parameters
 
+.. versionadded:: ATCv4.1 The ``role`` query parameter was added in version 4.1 of :abbr:`ATC (Apache Traffic Control)`.
+
 	+-----------+----------+------------------------------------------------------------------------------------------+
 	| Name      | Required | Description                                                                              |
 	+===========+==========+==========================================================================================+
--- a/docs/source/api/users.rst
+++ b/docs/source/api/users.rst
@@ -38,6 +38,8 @@ Request Structure
 	+-----------+----------+------------------------------------------------------------------------------------------+
 	| tenant    | no       | Return only users belonging to the :term:`Tenant` identified by tenant name              |
 	+-----------+----------+------------------------------------------------------------------------------------------+
+	| role      | no       | Return only users belonging to the :term:`Role` identified by role name                  |
+	+-----------+----------+------------------------------------------------------------------------------------------+
 	| username  | no       | Return only the user with this username                                                  |
 	+-----------+----------+------------------------------------------------------------------------------------------+
 	| orderby   | no       | Choose the ordering of the results - must be the name of one of the fields of the        |
--- a/traffic_ops/client/user.go
+++ b/traffic_ops/client/user.go
@@ -40,6 +40,7 @@ func (to *Session) GetUsers() ([]tc.User, ReqInf, error) {
 	route := apiBase + "/users"
 	resp, remoteAddr, err := to.request(http.MethodGet, route, nil)
 	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
+
 	if err != nil {
 		return nil, reqInf, err
 	}
@@ -55,20 +56,10 @@ func (to *Session) GetUsers() ([]tc.User, ReqInf, error) {
 
 // GetUsersByRole returns all users accessible from current user for a given role
 func (to *Session) GetUsersByRole(roleName string) ([]tc.User, ReqInf, error) {
-	route := apiBase + "/users?role=" + url.QueryEscape(roleName)
-	resp, remoteAddr, err := to.request(http.MethodGet, route, nil)
-	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
-	if err != nil {
-		return nil, reqInf, err
-	}
-	defer resp.Body.Close()
-
-	var data tc.UsersResponse
-	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
-		return nil, reqInf, err
-	}
-
-	return data.Response, reqInf, nil
+	data := tc.UsersResponse{}
+	route := fmt.Sprintf("%s/users?role=%s", apiBase, url.QueryEscape(roleName))
+	inf, err := get(to, route, &data)
+	return data.Response, inf, err
 }
 
 func (to *Session) GetUserByID(id int) ([]tc.User, ReqInf, error) {
--- a/traffic_ops/client/user.go
+++ b/traffic_ops/client/user.go
@@ -37,21 +37,10 @@ func (to *Session) Users() ([]tc.User, error) {
 
 // GetUsers returns all users accessible from current user
 func (to *Session) GetUsers() ([]tc.User, ReqInf, error) {
-	route := apiBase + "/users"
-	resp, remoteAddr, err := to.request(http.MethodGet, route, nil)
-	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
-
-	if err != nil {
-		return nil, reqInf, err
-	}
-	defer resp.Body.Close()
-
-	var data tc.UsersResponse
-	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
-		return nil, reqInf, err
-	}
-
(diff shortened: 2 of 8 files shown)


traffic_portal/app/src/common/modules/table/roleUsers/table.roleUsers.tpl.html
```diff
@@ -47,7 +47,7 @@
                 <td data-search="^{{::u.username}}$">{{::u.username}}</td>
                 <td data-search="^{{::u.email}}$">{{::u.email}}</td>
                 <td data-search="^{{::u.tenant}}$">{{::u.tenant}}</td>
-                <td data-search="^{{::u.rolename}}$">{{::u.rolename}}</td>
+                <td data-search="^{{::u.roleName}}$">{{::u.roleName}}</td>
```
rawlinp: Just as a reminder, this update shouldn't be required now that the roleName -> rolename bug has been fixed, right?

mitchell852: good catch. thanks.

traffic_ops/client/user.go
```diff
@@ -36,8 +37,26 @@ func (to *Session) Users() ([]tc.User, error) {
...
+		return nil, reqInf, err
+	}
+	defer resp.Body.Close()
+
+	var data tc.UsersResponse
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		return nil, reqInf, err
+	}
+
+	return data.Response, reqInf, nil
```
ocket8888: Why not just return the error, instead of using an `if` to explicitly return it if it's not `nil` and return `nil` when it's `nil`?

traffic_ops/testing/api/v1/user_test.go
```diff
@@ -292,6 +294,26 @@ func GetTestUsers(t *testing.T) {
 	}
 }
 
+func GetTestAdminUsers(t *testing.T) {
+	adminUsers, _, err := TOSession.GetUsersByRole("admin")
+	if err != nil {
+		t.Errorf("cannot GET users by role: %v", err)
+	}
+	if *adminUsers[0].RoleName != "admin" {
```
ocket8888: This can cause a segfault, need to check if the length of the returned slice is > 0 to ensure safe indexing, then need to make sure that the 0th item's `RoleName` property isn't `nil` to ensure safe dereferencing.

traffic_ops/testing/api/v1/user_test.go
```diff
@@ -292,6 +294,26 @@ func GetTestUsers(t *testing.T) {
...
+		t.Errorf("users get: RoleName expected %v actual %v", "admin", *adminUsers[0].RoleName)
+	}
+}
+
+func GetTestOperationsUsers(t *testing.T) {
+	opsUsers, _, err := TOSession.GetUsersByRole("operations")
+	if err != nil {
+		t.Errorf("cannot GET users by role: %v", err)
+	}
+	if *opsUsers[0].RoleName != "operations" {
```
ocket8888: Same as above RE: segfault possibility.

traffic_portal/app/src/common/modules/form/role/form.role.tpl.html
```diff
@@ -25,7 +25,7 @@
         </ol>
         <div class="pull-right" role="group" ng-show="!settings.isNew">
             <button class="btn btn-primary" title="View Capabilities" ng-show="enforceCapabilities" ng-click="viewCapabilities()">View Capabilities</button>
-            <!--<button class="btn btn-primary" title="View Users" ng-click="viewUsers()">View Users</button>-->
+            <button class="btn btn-primary" title="View Users" ng-click="viewUsers()">View Users</button>
```
ocket8888: This button (and actually the above 'viewCapabilities' button too) is a `button[type="submit"]`, should set `type` to `button`.

traffic_portal/app/src/modules/private/roles/users/index.js
```diff
@@ -30,8 +30,8 @@ module.exports = angular.module('trafficPortal.private.roles.users', [])
 							roles: function($stateParams, roleService) {
 								return roleService.getRoles({ id: $stateParams.roleId });
 							},
-							roleUsers: function($stateParams, userService) {
-								return userService.getUsers({ roleId: $stateParams.roleId });
+							roleUsers: function(roles, userService) {
+								return userService.getUsers({ role: roles[0].name });
```
ocket8888: I know this should never happen in the context of a UI that a user must sign in to use, but it still scares me to see potentially uncaught `TypeError`s being calculated in every digest cycle.

mitchell852: i don't follow

traffic_ops/client/user.go
mitchell852: i don't follow. this seems right to me. if there's an error, return nil for the slice of users and return the error. oh, are you saying on this line:

`return data.Response, reqInf, nil`

change it to 

`return data.Response, reqInf, err`

mitchell852: oh, i see what's your saying. lots of extra code.

traffic_portal/app/src/modules/private/roles/users/index.js
ocket8888:
```javascript
const roles = [];
console.log(roles[0].name); //TypeError: cannot read property 'name' of undefined
```

traffic_ops/client/user.go
ocket8888: yes exactly

traffic_portal/app/src/modules/private/roles/users/index.js
mitchell852: roles should never be an empty array as it pulls a valid role id from the url. i.e. tp.example.net/#!/roles/4/users

however, yes, you could change that role id to 999 but....don't do that. :)

mitchell852: i suppose extra code could be added to catch the error but seems like it would just be adding more code for a scenario that is super rare and has no real benefit.

ocket8888: Well in a proper Angular test bed we'd be mocking up services like the API so it doesn't depend on an actual, running instance of TO. But since we know that this code is always running in a world where it's connected to a TO where it's authenticated and users must have roles, we do know that the list of all roles must not be empty.

So I guess if we have no immediate plans to remove the tests' dependency on TO (and I don't think we do) then we can let this go.

traffic_ops/client/user.go
mitchell852: well `GetUsers` i didn't write so hesitant to rewrite it. although it is under test so i guess i could.

ocket8888: well you don't need to rewrite `GetUsers` if you don't want to. Git diffs are weird, I put this in the part it said you changed but it looks like what you actually did was remove `.json` and add the `GetUsersByRole` method, so really it just ought to be changed in `GetUsersByRole`. Though I'd consider the same change to `GetUsers` an improvement and encourage you to do that too, if you want.

traffic_ops/testing/api/v1/user_test.go
mitchell852: done

traffic_ops/testing/api/v1/user_test.go
mitchell852: done

traffic_ops/testing/api/v1/user_test.go
```diff
@@ -296,8 +296,8 @@ func GetTestUsers(t *testing.T) {
 
 func GetTestAdminUsers(t *testing.T) {
 	adminUsers, _, err := TOSession.GetUsersByRole("admin")
-	if err != nil {
-		t.Errorf("cannot GET users by role: %v", err)
+	if err != nil || len(adminUsers) == 0 || adminUsers[0].RoleName == nil {
```
ocket8888: If this test errors because `adminUsers` is `nil`/empty or its first member's `RoleName` property is `nil` then it'll print `cannot GET admin users by role: <nil>` which isn't super helpful. But what's worse is that in that case the test continues because it's only an error, not a fatal one. So it would still segfault.

traffic_ops/testing/api/v1/user_test.go
```diff
@@ -306,8 +306,8 @@ func GetTestAdminUsers(t *testing.T) {
 
 func GetTestOperationsUsers(t *testing.T) {
 	opsUsers, _, err := TOSession.GetUsersByRole("operations")
-	if err != nil {
-		t.Errorf("cannot GET users by role: %v", err)
+	if err != nil || len(opsUsers) == 0 || opsUsers[0].RoleName == nil {
```
ocket8888: Same as above RE: segfault

traffic_ops/testing/api/v1/user_test.go
mitchell852: gotcha. still learning a lot about these tests

docs/source/api/users.rst
```diff
@@ -38,6 +38,8 @@ Request Structure
 	+-----------+----------+------------------------------------------------------------------------------------------+
 	| tenant    | no       | Return only users belonging to the :term:`Tenant` identified by tenant name              |
 	+-----------+----------+------------------------------------------------------------------------------------------+
+	| role      | no       | Return only users belonging to the :term:`Role` identified by role name                  |
+	+-----------+----------+------------------------------------------------------------------------------------------+
```
ocket8888: There should be a `versionadded` directive either right above or right below this table to mention that this was unavailable in 4.0. I think our next version is ATC 4.1 - barring any urgent bugfixes.

docs/source/api/users.rst
```diff
@@ -31,13 +31,17 @@ Request Structure
 -----------------
 .. table:: Request Query Parameters
 
+.. versionadded:: ATCv4.1 The ``role`` query parameter was added in version 4.1 of :abbr:`ATC (Apache Traffic Control)`.
```
ocket8888: > `WARNING: Content block expected for the "table" directive; none found.`

You added it in between the declaration of a table and its definition. I'm actually shocked that the table still renders fine (I guess it's a table caption with no body and then a floating table with no related caption that just happen to be sequential?)

At any rate, the notice doesn't appear. I see this section already has a `.. versionadded::` for the query parameters added in API v1.4. You should try putting this directly above or below that (separated by a blank line, of course).

shamrickus:
I'm getting the following failures when running the v1 tests
```
user_test.go:305: users get: RoleName expected admin actual disallowed
user_test.go:318: users get: RoleName expected operations actual disallowed
```

ocket8888:
Did you rebuild your traffic ops? I was seeing that error until I did a rebuild.

shamrickus:
I thought so, but after rebuilding again that fixed the issue.

PR #4198, opened 2019-12-11
EDNS0 Client Subnet for Delivery Services

### LINKED ISSUE #2929: eDNS0 client subnet extension support by DS
eDNS0 client subnet extensions are currently only supported at the CDN level, we should support eDNS0 client subnet extensions at the Delivery Service level.

mattjackson220:

This PR adds the ability to enable EDNS0 client subnet on the delivery service level.  Currently, this can be enabled on the CDN level, and this PR still supports that functionality but allows for enabling it per delivery service as well.

- [x] This PR fixes #2929

- Documentation
- Traffic Ops
- Traffic Portal
- Traffic Router

- Verify that the Delivery Services page in Traffic Portal has been updated in the Routing section to include a new boolean value for EcsEnabled.
- Verify that updating this value gets updated in the TO database.
- Verify that the API delivery services endpoints work as expected.
- In a public facing environment, update a DNS Delivery Service to be Ecs Enabled. Perform a snapshot.
- Run the following commands replacing yourDeliverySevice with your delivery service edge location (ex: edge.example.com), yourSubnet and yourSubnet2 with a valid subnets from 2 far apart locations (can be found from /opt/traffic_router/db/czmap.json):
        dig +trace @8.8.8.8 yourDeliverySevice +dnssec ALL
        dig +trace @8.8.8.8 yourDeliverySevice +dnssec +subnet=yourSubnet ALL
        dig +trace @8.8.8.8 yourDeliverySevice +dnssec +subnet=yourSubnet2 ALL
- In Traffic Router (could be in any of the active Traffic Routers) view the access log and find the entry for your requests.  
- Verify that the first request has "chi=anIPAddress rhi=-" and that ans= list of IPs of the edge cache/s closest to YOU.
- Verify that the second request has "chi=anIPAddressInYourSubnet rhi=IPAddressFromFirstRequest" and that ans = list of IPs of the edge cache/s closest to the subnet location.
- Verify that the third request has "chi=anIPAddressInYourSubnet2 rhi=IPAddressFromFirstRequest" and that ans = list of IPs of the edge cache/s closest to the subnet2 location.  

- Follow steps again for a different DNS Delivery Service but leave Ecs Disabled and verify that all requests get "chi=anIPAddress rhi=-" and that ans= list of IPs of the edge cache/s closest to YOU.

- [x] This PR includes tests


rawlinp:
add to whitelist

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -630,6 +630,25 @@ <h3>Current Value</h3>
                             </aside>
                         </div>
                     </div>
+
+					<div class="form-group" ng-class="{'has-error': hasError(routingConfig.ecsEnabled), 'has-feedback': hasError(routingConfig.ecsEnabled)}">
```
mitchell852: looks like you have some formatting issues going on in these tpl files.

mattjackson220: fixed

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -255,7 +255,8 @@ func createV14(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, reqDS t
 		&ds.TRRequestHeaders,
 		&ds.TRResponseHeaders,
 		&ds.TypeID,
-		&ds.XMLID)
+		&ds.XMLID,
```
rawlinp: Also for the 1.4 -> 1.5 bump, there are some more changes that will need to be made to this file:
1. rename `createV14` to `createV15` and change the DSv14 struct usages in the function to DSv15.
2. create a _new_ `createV14` function almost exactly like the existing `createV13` (the basic idea is the V14 func takes a V14 struct as input, converts it to a V15 struct, passes that to the `createV15` method, that method does all the real work with the v15 struct, returns it to `createV14`, then `createV14` extracts the DSv14 from that result and returns it).
3. similar to step 2, except for the `updateV14` -> `updateV15` funcs. The only difference is that the `updateV*` funcs need to read existing fields from the DB before passing the struct up the chain to the next handler. So the new `updateV14` handler would have to read the existing `ecs_enabled` value from the DB into the V15 struct before passing it to `updateV15`.
4. Copy the `CreateV14` func to `CreateV15` and change the v14 stuff to v15.
5. Same as step 4 for `UpdateV14` func.
6. Modify the `switch` statement in the `Read` function to add a new `version.Minor >= 5` case which appends `ds` and modify the `>= 4` case to append a `ds.DeliveryServiceNullableV14`.
7. In `routes.go`, add new DS Routes for 1.5 POST and 1.5 PUT, calling `deliveryservice.CreateV15` and `deliveryservice.UpdateV15`, respectively. Order matters here, so the 1.5 handlers need to be above the 1.4 handlers, and so on.

I think that pretty much covers all the steps for bumping from v14 to v15. The basic idea is that all the functions chain together, "upgrading" the request as they go, until they arrive at the latest version which actually handles the request.

docs/source/api/deliveryservices.rst
```diff
@@ -90,6 +90,10 @@ Response Structure
 :dnsBypassIp6:      A :ref:`ds-dns-bypass-ipv6`
 :dnsBypassTtl:      The :ref:`ds-dns-bypass-ttl`
 :dscp:              A :ref:`ds-dscp` to be used within the :term:`Delivery Service`
+:ecsEnabled:        A boolean that defines the :ref:`ds-ecs` setting on this :term:`Delivery Service`
+
+	.. versionadded:: 1.4
```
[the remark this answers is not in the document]
mattjackson220: done

lib/go-tc/deliveryservices.go
```diff
@@ -164,6 +164,7 @@ type DeliveryServiceNullable struct {
 	DeliveryServiceNullableV13
 	ConsistentHashRegex       *string  `json:"consistentHashRegex"`
 	ConsistentHashQueryParams []string `json:"consistentHashQueryParams"`
+	EcsEnabled                bool     `json:"ecsEnabled" db:"ecs_enabled"`
```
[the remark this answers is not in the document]
mattjackson220: done

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/dns/NameServer.java
```diff
@@ -162,6 +162,34 @@ private void addAnswers(final Message request, final Message response, final Ine
 		}
 	}
 
+	@SuppressWarnings({"PMD.UseStringBufferForStringAppends"})
+	private boolean isDeliveryServiceEcsEnabled(final Name name) { //, int qtype, InetAddress clientAddress) {
```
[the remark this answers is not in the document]
mattjackson220: removed

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
mattjackson220: done. thanks for the steps!

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Deprecated/Removed
 - The TO API `servers/totals` endpoint is now deprecated
+- Added a boolean to delivery service in Traffic Portal and Traffic Ops to enable EDNS0 client subnet at the delivery service level and include it in the cr-config.
```
rawlinp: these should go under the `Added` section

rawlinp:
Also, I'm not sure if you want to do this as a follow-up PR or in this one, but in order for TP to work with the new DS field, it needs to be pointed to 1.5 now -- which I'm pretty sure is this: 
```
traffic_portal/app/src/scripts/config.js:.constant('ENV', { api: { root:'/api/1.4/' } })
```
With TP making 1.5 requests, I think TO-Perl also needs updated to have a 1.5 section so that any remaining Perl routes will be found when requesting 1.5, like there are for 1.4:
```
TrafficOpsRoutes.pm: # 1.4 Routes
TrafficOpsRoutes.pm:    $version = "1.4";
```

CHANGELOG.md
mattjackson220: done. sorry about that

mattjackson220:
> Also, I'm not sure if you want to do this as a follow-up PR or in this one, but in order for TP to work with the new DS field, it needs to be pointed to 1.5 now -- which I'm pretty sure is this:
> 
> ```
> traffic_portal/app/src/scripts/config.js:.constant('ENV', { api: { root:'/api/1.4/' } })
> ```
> 
> With TP making 1.5 requests, I think TO-Perl also needs updated to have a 1.5 section so that any remaining Perl routes will be found when requesting 1.5, like there are for 1.4:
> 
> ```
> TrafficOpsRoutes.pm: # 1.4 Routes
> TrafficOpsRoutes.pm:    $version = "1.4";
> ```

I'll open a new PR for this so it isn't tied up with this one

PR #4203, opened 2019-12-11
TO: rewrite /isos to Go

### LINKED ISSUE #3801: Rewrite /isos to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/isos` is handled by Perl

## Expected / new behavior:
`/isos` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `POST`

N/A

awilliams:
- [x] This PR fixes #3801 such that the `/isos` handler has been rewritten in Go

- Documentation (TODO)
- Traffic Control Client (TODO?)
- Traffic Ops

The majority of the code added is in the `./traffic_ops/traffic_ops_golang/iso` package, which includes a number of Go tests. To execute:

```shell
go test ./traffic_ops/traffic_ops_golang/iso
```

## Additional Information

- The Go version of the endpoint attempts to produce the same behavior as the Perl version, with the following exceptions: 
    
    - The Go endpoint does not support non-streaming downloads. In Perl, non-streaming downloads worked by saving the ISO image to the server's disk and returning to the requesting client a URL that allowed them to download that file at a later time. If a non-streaming request is made to the Go endpoint, it returns an error in the standard JSON response structure.
    This was discussed on the dev mailing list under the [_Proposal: Only allow streaming downloads of ISOs_](https://lists.apache.org/thread.html/60fd81126f54c1fbc831dc68bce7c87f00ecc5fccb6ca6457dc04772%40%3Cdev.trafficcontrol.apache.org%3E) subject. It was decided that this is a breaking API change, and therefore the Go API should be a version 2.0 endpoint.

    - In a number of error cases, the Perl version returns HTML. The Go version returns JSON alerts in these cases, for example when executing the `mkisofs` command fails.

    - The Perl version, when streaming the response, would fail silently if the subprocess failed ([code](https://github.com/apache/trafficcontrol/blob/19aa398b2550ee5f308b6678cba0647931efb381/traffic_ops/app/lib/API/Iso.pm#L264)). The Go version returns an error to the user in this case.

    - When creating the `mgmt_network.cfg` file, the Perl version does not quote the `DEVICE` value ([code](https://github.com/apache/trafficcontrol/blob/19aa398b2550ee5f308b6678cba0647931efb381/traffic_ops/app/lib/API/Iso.pm#L180)), while the Go version does.

- This PR adds a new Go library (https://github.com/GehirnInc/crypt). Relevant discussion on the mailing list was under message with subject: [_Proposal to add GehirnInc/crypt library_](https://lists.apache.org/thread.html/6da4dae2cefb4413a67c82132c7ba5b871e607793410384f96c6f863%40%3Cdev.trafficcontrol.apache.org%3E)

- The `mkisofs` package could be added to the TrafficOps Go Docker container as part of ciab.

- It's likely that custom `generate` scripts will use the `implantisomd5` program provided on Centos by the `isomd5sum` package. The `mkisofs` package is already a [requirement](https://github.com/apache/trafficcontrol/blob/e3c48b6fde2a154acb5e557dba36dc615ec38b8e/traffic_ops/build/traffic_ops.spec#L36) of the TrafficOps RPM, so we should considering making `isomd5sum` a requirement as well.

- There's a cronjob ([./traffic_ops/etc/cron.d/trafops_clean_isos](https://github.com/apache/trafficcontrol/blob/6d9b495ec72b9942956ce7cd47975634c4eae7f8/traffic_ops/etc/cron.d/trafops_clean_isos)) that removes stale ISO files. This isn't necessary for the Go version of this endpoint. Once the Perl endpoint becomes obsolete, it can likely be deleted. The Go version does no store copies of the ISO images, therefore there's no cleanup necessary.

### Questions

- Should support for this endpoint be added to the `client` package?
- Should the endpoint limit concurrency? Since the endpoint creates/updates files as part of the kickstart configuration, it's possible for multiple requests to interfere with each other. It might also be possible to copy the base kickstart files to a temp directory first, and then operate on that unique directory per request. I'm not sure if the Perl version has this issue -- depends on if requests can happen concurrently. 

### TODO

- [ ] Documentation
- [ ] Changelog
- [ ] Change route to v2.0 and start discussion around this (this didn't seem to work in testing, maybe because of a bug in the routing code).
- [x] Use `osversions.json` to limit values that a user can request
- [ ] Remove CIDR formatting from IP certain addresses, per [Perl version](https://github.com/apache/trafficcontrol/blob/19aa398b2550ee5f308b6678cba0647931efb381/traffic_ops/app/lib/API/Iso.pm#L181)


mitchell852:
If behavior is going to change (and for the better), sounds like this needs to be held until api 2.0 where the `?stream` query param is completely ignored?

rawlinp:
> Should support for this endpoint be added to the client package?

Hmm...maybe that's not really necessary unless you believe we need it for API integration tests.

>Should the endpoint limit concurrency? Since the endpoint creates/updates files as part of the kickstart configuration, it's possible for multiple requests to interfere with each other. It might also be possible to copy the base kickstart files to a temp directory first, and then operate on that unique directory per request. I'm not sure if the Perl version has this issue -- depends on if requests can happen concurrently.

Also a good question. I'm pretty sure Perl handles requests concurrently; Mojolicious handles each request in one of a pre-configured number of "worker" processes. How large are the files in question? Can they be copied to a temp dir quickly enough to keep the API reasonably quick?

mhoppa:
> > Should support for this endpoint be added to the client package?
> 
> Hmm...maybe that's not really necessary unless you believe we need it for API integration tests.
> 
> > Should the endpoint limit concurrency? Since the endpoint creates/updates files as part of the kickstart configuration, it's possible for multiple requests to interfere with each other. It might also be possible to copy the base kickstart files to a temp directory first, and then operate on that unique directory per request. I'm not sure if the Perl version has this issue -- depends on if requests can happen concurrently.
> 
> Also a good question. I'm pretty sure Perl handles requests concurrently; Mojolicious handles each request in one of a pre-configured number of "worker" processes. How large are the files in question? Can they be copied to a temp dir quickly enough to keep the API reasonably quick?

I looked into copying files into a temp directory that we are acting upon to solve this but one problem is the logic looks for a custom executable and it doesnt pass in the directory of the kickstart files we are modifying so it relies on that executable to know exactly where those files should live. With that and the fact that Perl handles requests concurrently making this not a new problem I say we leave it is implemented to avoid not breaking the custom executables that are out there.

mhoppa:
This is replaced by https://github.com/apache/trafficcontrol/pull/4371 and can be closed. The only differences being https://github.com/apache/trafficcontrol/pull/4379/commits/862d368b500f928c769cf7d312a8e275cbb25eb6

ocket8888:
Hang on, I'm going to try to get the original author to close it - since we can't commit his work to the project without his express consent (and because draft PRs can't be merged a draft may not meet that standard)

awilliams:
Closing in favor of #4371. 

Consider this comment consent as requested by @ocket8888 in  https://github.com/apache/trafficcontrol/pull/4203#issuecomment-582651113.

PR #4208, opened 2019-12-12
Block DSRC assignment when DS type is not HTTP or DNS

### LINKED ISSUE #4101: Delivery Service Required Capability assignment should be restricted to sensible DS types
## I'm submitting a ...
- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [X] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Delivery Service Capabilities can be assigned to Delivery Services of arbitrary type - although that causes an incorrect error.

## Expected / new behavior
As STEERING, CLIENT_STEERING Delivery Services cannot be assigned servers, and ANY_MAP Delivery Services are not routed by ATC, it makes little sense to require capabilities on them. In particular, this could be confusing on STEERING and CLIENT_STEERING Delivery Services because the capability restriction isn't actually enforced, since it may or may not exist on targets.

Instead, it should only be possible to assign required capabilities to Delivery Services of the types:
- DNS
- DNS_LIVE
- DNS_LIVE_NATNL
- HTTP
- HTTP_LIVE
- HTTP_LIVE_NATNL
- HTTP_NO_CACHE

1. Choose or create a CLIENT_STEERING Delivery Service
1. Choose or create a capability
1. Assign the capability to the Delivery Service as a Required Capability
1. Observe that the error says that there are servers assigned without this capability, which is nonsensical. This error should not be displayed as there is no way to link server to CLIENT_STEERING DS other than the targets which we have removed. so basically there are no servers associated yet it gives error as if it has something associated.

mhoppa:

- [x] This PR fixes #4101 

- Traffic Ops
- Traffic Portal

### Manual 

Attempt to assign DSRC to ANY_MAP or STEERING DS and it should fail now

### Automated 

Run unit tests


traffic_ops/testing/api/v14/deliveryservices_required_capabilities_test.go
```diff
@@ -290,14 +297,14 @@ func DeleteTestDeliveryServicesRequiredCapabilities(t *testing.T) {
 	}
 }
 
-func helperGetDeliveryServiceID(t *testing.T, capability tc.DeliveryServicesRequiredCapability) *int {
+func helperGetDeliveryServiceID(t *testing.T, xmlID *string) *int {
 	t.Helper()
-	ds, _, err := TOSession.GetDeliveryServiceByXMLID(*capability.XMLID)
+	ds, _, err := TOSession.GetDeliveryServiceByXMLID(*xmlID)
```
ocket8888: The way `xmlID` is getting passed in here it could be `nil` if defined that way in the test data. That's not valid (I'm pretty sure) for the database, but this is getting called prior to creation so you can't make the assumption that it conforms properly to the data model. Even if you could, though, I'd still always advise that pointers constructed from external data be checked against null before being dereferenced.

tl;dr pls check if `xmlID` is `nil` dereferencing, either at the top of this helper or before calling it

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -1472,15 +1472,15 @@ func getDSTenantIDByName(tx *sql.Tx, ds tc.DeliveryServiceName) (*int, bool, err
 }
 
 // GetDeliveryServiceType returns the type of the deliveryservice.
-func GetDeliveryServiceType(dsID int, tx *sql.Tx) (tc.DSType, error) {
+func GetDeliveryServiceType(dsID int, tx *sql.Tx) (tc.DSType, bool, error) {
 	var dsType tc.DSType
 	if err := tx.QueryRow(`SELECT t.name FROM deliveryservice as ds JOIN type t ON ds.type = t.id WHERE ds.id=$1`, dsID).Scan(&dsType); err != nil {
 		if err == sql.ErrNoRows {
-			return tc.DSTypeInvalid, errors.New("a deliveryservice with id '" + strconv.Itoa(dsID) + "' was not found")
+			return tc.DSTypeInvalid, false, nil
```
ocket8888: thanks for doing this, I really like this model.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -223,9 +223,12 @@ func getServers(params map[string]string, tx *sqlx.Tx, user *auth.CurrentUser) (
...
 `
 		// depending on ds type, also need to add mids
-		dsType, err := deliveryservice.GetDeliveryServiceType(dsID, tx.Tx)
+		dsType, exists, err := deliveryservice.GetDeliveryServiceType(dsID, tx.Tx)
 		if err != nil {
-			return nil, err, nil, http.StatusBadRequest
+			return nil, nil, err, http.StatusInternalServerError
+		}
+		if !exists {
+			return nil, errors.New("a deliveryservice with id '" + strconv.Itoa(dsID) + "' was not found"), nil, http.StatusBadRequest
```
ocket8888: nit: `fmt.Errorf` is more clean for formatting errors than concatenation of string values in `errors.New`.

mitchell852:
I'm getting this error when running the API tests:

```
=== RUN   TestDeliveryServiceServersWithRequiredCapabilities/successful_server_->_DS_assignment
    --- FAIL: TestDeliveryServiceServersWithRequiredCapabilities/successful_server_->_DS_assignment (0.06s)
        deliveryserviceservers_test.go:87: *POST delivery service required capability: 400 Bad Request[400] - Error requesting Traffic Ops https://localhost:8443/api/1.4/deliveryservices_required_capabilities {"alerts":[{"text":"Invalid DS type. Only DNS and HTTP DSes can have required capabilities","level":"error"}]}
```

PR #4210, opened 2019-12-13
Fix TO to remove proxy route for Go route

rob05c:
Removes a proxyHandler route, for an endpoint which was rewritten in
Go and has a Go route.

See https://github.com/apache/trafficcontrol/pull/2305/files

No tests, just removing a proxy, existing route has been manually tested.
No docs, no interface change.

- Traffic Ops

Request cdns/health route, verify matches Perl, and has `x-server-name` Golang header.

Not a bug fix.

### DIFF
--- a/traffic_ops/traffic_ops_golang/routing/routes.go
+++ b/traffic_ops/traffic_ops_golang/routing/routes.go
@@ -170,8 +170,10 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `cdns/capacity$`, cdn.GetCapacity, auth.PrivLevelReadOnly, Authenticated, nil, 697185281, perlBypass},
 		{1.1, http.MethodGet, `cdns/configs/?(\.json)?$`, cdn.GetConfigs, auth.PrivLevelReadOnly, Authenticated, nil, 1768437852, noPerlBypass},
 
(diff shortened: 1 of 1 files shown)


traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -171,7 +171,6 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `cdns/configs/?(\.json)?$`, cdn.GetConfigs, auth.PrivLevelReadOnly, Authenticated, nil, 1768437852, noPerlBypass},
 
 		{1.1, http.MethodGet, `cdns/domains/?(\.json)?$`, cdn.DomainsHandler, auth.PrivLevelReadOnly, Authenticated, nil, 296902560, noPerlBypass},
-		{1.1, http.MethodGet, `cdns/health$`, handlerToFunc(proxyHandler), 0, NoAuth, []middleware.Middleware{}, 1443074329, noPerlBypass},
```
rawlinp: Oy, the route ordering strikes again. I think this route was here because it otherwise gets eaten by the less specific `GET /cdns/{id}` route. So the new `GET /cdns/health` routes need to be moved up here as well.

rob05c: Ah, I'll fix. Give me a bit, currently using my machine to diff configs, planning to test this when that's done.

rob05c: Fixed

Issue #4246, opened 2019-12-30
Service 'trafficops-perl' failed to build: [...] dial tcp: lookup geolite.maxmind.com: No address associated with hostname


zrhoffman:
## I'm submitting a ...

- [x] bug report
- [ ] new feature / enhancement request
- [ ] improvement request (usability, performance, tech debt, etc.)
- [ ] other 

## Traffic Control components affected ...

- [x] CDN in a Box
- [x] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

The `trafficops-perl` Docker image for CDN-in-a-Box fails to build because the geolite.maxmind.com subdomain no longer resolves.

## Expected / new behavior:

`trafficops-perl` should build successfully.

## Minimal reproduction of the problem with instructions:

```shell script
docker rmi cdn-in-a-box_perl;
docker-compose build trafficops-perl;
```

Failed build log follows:
```shell script
Building trafficops-perl
Step 1/16 : FROM centos:7
 ---> 5e35e350aded
Step 2/16 : LABEL   name="Traffic Ops Perl"         maintainer="dev@trafficcontrol.apache.org"
 ---> Using cache
 ---> cfa2c7d6e139
Step 3/16 : EXPOSE 443
 ---> Using cache
 ---> f20b3afea2b4
Step 4/16 : ENV MOJO_MODE production
 ---> Using cache
 ---> 07c71701c21e
[...]

PR #4247, opened 2019-12-31
TO image: Copy files from a previous build stage instead of trafficops-perl

zrhoffman:

This PR copies the `traffic_ops.rpm` files from from a previous build stage instead of from the `trafficops-perl` image. This mean that you no longer need to update `trafficops-perl` if the only part of the RPM that changed is `traffic_ops_golang`, which reduces build time by ~16~ 12 minutes (`trafficops-perl` build time on my Linux machine).

For me, `trafficops-go` now takes a little under a minute to build.

AFAIK, no documentation needs to change as a result of this PR.

- CDN-in-a Box

* Ensure the `trafficops-go` image builds
* Ensure the `trafficops-perl` image builds
* Ensure CDN-in-a-Box starts successfully

- [x] I have explained why documentation in this PR is unnecessary
- [x] No need to update CHANGELOG.md

## Additional Information

~Blocked by #4249~ (unblocked in [b2d282f~...2cdb540](https://github.com/apache/trafficcontrol/compare/b2d282f~...2cdb540))

### DIFF
--- a/infrastructure/cdn-in-a-box/docker-compose.yml
+++ b/infrastructure/cdn-in-a-box/docker-compose.yml
@@ -58,8 +58,10 @@ services:
   # defined below
   trafficops:
     build:
-      context: .
-      dockerfile: traffic_ops/Dockerfile-go
+      context: ../..
+      dockerfile: infrastructure/cdn-in-a-box/traffic_ops/Dockerfile-go
+      args:
+        TRAFFIC_OPS_RPM: infrastructure/cdn-in-a-box/traffic_ops/traffic_ops.rpm
     depends_on:
       - db
       - enroller
--- a/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile-go.dockerignore
+++ b/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile-go.dockerignore
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
 # To use this dockerignore:
 # - Set these environment variables: COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1
 # - Use a docker version of 19.03.0 or higher (Docker Desktop >=2.1.0.1)
--- a/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile-go.dockerignore
+++ b/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile-go.dockerignore
@@ -1,0 +1,10 @@
+# To use this dockerignore:
+# - Set these environment variables: COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1
+# - Use a docker version of 19.03.0 or higher (Docker Desktop >=2.1.0.1)
+# For more information, see:
+# - https://github.com/moby/buildkit/pull/901
+# - https://github.com/docker/compose/pull/6865
+# - https://docs.docker.com/develop/develop-images/build_enhancements/
+*
+!infrastructure/cdn-in-a-box/enroller/
+!infrastructure/cdn-in-a-box/traffic_ops/
--- a/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile.dockerignore
+++ b/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile.dockerignore
@@ -1,0 +1,12 @@
+# To use this dockerignore:
+# - Set these environment variables: COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1
+# - Use a docker version of 19.03.0 or higher (Docker Desktop >=2.1.0.1)
+# For more information, see:
+# - https://github.com/moby/buildkit/pull/901
+# - https://github.com/docker/compose/pull/6865
+# - https://docs.docker.com/develop/develop-images/build_enhancements/
+*
+!infrastructure/cdn-in-a-box/
+!traffic_ops/app/cpanfile
+!traffic_ops/install/bin/install_goose.sh
+!traffic_router/core/src/test/resources/geo/GeoLite2-City.mmdb.gz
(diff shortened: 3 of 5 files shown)


ocket8888:
The reason the builds were coupled was to avoid a mismatch between the TO release versions of the Perl and Go. Arguably the Perl doesn't change much anymore, but it does change e.g. to include deprecation warnings.

It does seem, though, that such a mismatch could only occur with these changes by the direct intent of the person building the images, but another thing to keep in mind is that this is also a problem that solves itself once we can finally kill Perl.

With all that said I'm not opposed to including a change like this, but I do have one question: what do you think about going the route of our caching servers and stemming both services from a common base image that does little more than install the RPM? Then your build layer just becomes an image that is the backbone of both `trafficops` and `trafficops-perl`, which has the same effect but the output images of `docker-compose build` are forcibly always consistent in their TO versions. Though by specifying either `trafficops` or `trafficops-perl` you could still build only the one you care about, thus retaining the build speed improvement.

zrhoffman:
If we are going to have the Perl and Go images use a minimal base image, we could just use `trafficops-go` itself as the base image for `trafficops-perl`. There are no Yum packages installed in the `trafficops-go` image that `trafficops-perl` does not also use. The only bloat that I see from this approach would be that `trafficops-perl`'s base image initially would include the `traffic_ops_golang` binary (and the 7.1MB `traffic_ops.rpm` file, depending on how we want to install its remaining dependencies in the Perl image).

Because we would have the RPM at the start of the Perl Dockerfile, we would lose the "cached layers" benefit of #3733, though that would now only affect rebuilding the Perl image.

zrhoffman:
Let me know your thoughts on [fdd02b5~...5ac6509](https://github.com/apache/trafficcontrol/compare/fdd02b5~...5ac6509).

`trafficops-go` gains around 16MB, `trafficops-perl` loses around 112MB.

zrhoffman:
(Still needs the GeoLite2 line commented out due to #4246)

ocket8888:
I don't think you need to get rid of the pre-installation of Perl dependencies; that was being done by directly including the cartonfile, not using the one installed by the RPM.

zrhoffman:
What is the benefit of pre-installing Perl dependencies in this case? `trafficops-go` builds relatively quickly (on my machine, under a minute), and the bulk of the time taken building `trafficops-perl` is in processing the cpanfile. If that layer is the part we want to cache, its placement in the Dockerfile won't significantly impact build time.

ocket8888:
By "Perl dependencies" I mean Carton. It takes something like 97% of the image's build time and - to my knowledge - has never changed since the inception of the CDN-in-a-Box project.

zrhoffman:
Right. But what is an instance where a CiaB user would benefit from doing that before installing RPM dependencies? The base image still needs to be updated to change the RPM, so the whole image would still need to be rebuilt if someone wanted to do that.

ocket8888:
If you install the TO RPM before installing carton dependencies, then every time your source code changes it will pull all of the dependencies. If you install the carton dependencies first, then they can be cached until they change regardless of any changes to the actual TO codebase.

zrhoffman:
Okay. In order for this to be effective, `trafficops-perl` cannot use `trafficops-go` as a base image, since we want to copy in the RPM after installing the carton dependencies.

zrhoffman:
Reverted [53c8bc5~...5ac6509](https://github.com/apache/trafficcontrol/compare/53c8bc5~...5ac6509) in [b2d282f~...2cdb540](https://github.com/apache/trafficcontrol/compare/b2d282f~...2cdb540), please review.

infrastructure/cdn-in-a-box/docker-compose.yml
```diff
@@ -58,8 +58,10 @@ services:
   # defined below
   trafficops:
     build:
-      context: .
-      dockerfile: traffic_ops/Dockerfile-go
+      context: ../..
```
ocket8888: Why did the context need to be expanded?

zrhoffman: The `trafficops` and `trafficops-perl` containers both receive a build arg named `TRAFFIC_OPS_RPM`. IMO, both of these build args should be the same value. That means that `trafficops` needs `infrastructure/cdn-in-a-box` in its build context in order to successfully be able to run its `COPY` command.

Within the constraints discussed, in order to only specify the build arg once, it would be possible for `trafficops-go` and `trafficops-perl` to both be child images of a `centos:7` image that does nothing else but receive the `TRAFFIC_OPS_RPM` build arg and export it as an environment variable. However, in this case, the `trafficops` container would still require a build context that includes `infrastructure/cdn-in-a-box`.

ocket8888: I think that the consistency is certainly a good point, and it only impacts build time by about 2 minutes give or take, but you would know better than me: does this affect the size of objects stored on the filesystem or in memory in the long-term? It doesn't end up in a cache layer, but I don't know enough about the inner workings of Docker to say definitively that it isn't stored somewhere, and I was wondering if you knew.

zrhoffman: We can add a `.dockerignore`, list everything, and except the files we use. That should speed it back up.

ocket8888: That  sounds like a great idea

zrhoffman: Added in 7e96aaa. For me, `trafficops-perl` built ~3.5 minutes faster and `trafficops-go` built 7 seconds faster. Note that use of these dockerignores requires Docker Buildkit. Without it, a limitation of `docker-compose` is only one dockerignore per build context. That means we would need to use a single dockerignore for all `docker-compose` projects in the repo that use the repo's base path as a build context, which IMO is not worth maintaining.

Somewhere down the road, it would be nice to update the Makefile to copy in files that CDN-in-a-Box Dockerfiles add to somewhere within the CiaB directory itself so we can use the CiaB directory as a build context for all of the images.

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile-go.dockerignore
```diff
@@ -0,0 +1,10 @@
+# To use this dockerignore:
```
ocket8888: missing license header

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile.dockerignore
```diff
@@ -0,0 +1,12 @@
+# To use this dockerignore:
```
ocket8888: same as above RE license header

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile-go.dockerignore
zrhoffman: Added in e6dd453

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile.dockerignore
zrhoffman: Added in e6dd453

zrhoffman:
> [W]hat do you think about going the route of our caching servers and stemming both services from a common base image that does little more than install the RPM? Then your build layer just becomes an image that is the backbone of both `trafficops` and `trafficops-perl`, which has the same effect but the output images of `docker-compose build` are forcibly always consistent in their TO versions, even if the execution environment changes between runs. Though by specifying either `trafficops` or `trafficops-perl` you could still build only the one you care about, thus retaining the build speed improvement.

Oh, I see what you mean. Opened a PR to stem both services from a common base image: #4784

PR #4256, opened 2020-01-04
Fix atscfg meta to omit config files for nonexistent delivery services

rob05c:
This fixes atscfg to omit files with location parameters for config
files, hdr_rw_, regex_remap_, url_sig_, and uri_signing_ for DSes
which don't exist (typically after deleting a DS but forgetting to
delete the corresponding location Parameters).

This is important to make atstccfg work because without it, ORT will
request url_sig_ files for DSes which don't exist, and the API
endpoint to get the keys will fail. Because atstccfg uses the API,
there's no reasonable way to make it work, except to fix the meta
config to omit those nonexistent files.

In addition to nonexistent files, this also omits files for DSes
not assigned to the server. There won't even be remap lines for
those DSes anyway, so it's better and safer to omit the ancillary
files altogether.

Tagged major, because it causes bad data to break ORT, where it previously didn't and doesn't need to.

Includes tests.
Includes changelog.
No documentation, bug fix.

- Traffic Ops
- Traffic Ops ORT

Run tests. Request meta config `/api/1.4/servers/{id}/configfiles/ats` on a TO with location parameters for a nonexistent DS (the odds are good your Production environment has some), and verify those files are no longer in the returned JSON.

- 4.0.x RC0

## Additional Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Traffic Ops now supports a "sortOrder" query parameter on some endpoints to return API responses in descending order
 - Traffic Ops now uses a consistent format for audit logs across all Go endpoints
 - Added cache-side config generator, atstccfg, installed with ORT. Includes all configs. Includes a plugin system.
+- Fixed ATS config generation to omit regex remap, header rewrite, URL Sig, and URI Signing files for delivery services not assigned to that server.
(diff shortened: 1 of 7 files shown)


rob05c:
Ah, looks like the test isn't assigning the DS it's testing to that server. Thanks for finding that! I'll fix.

rob05c:
Ok, I fixed the API test. API tests all pass for me (after fixing Perl number-string issues), unit tests pass.

PR #4260, opened 2020-01-06
Codecov

ocket8888:
Adds code-coverage output to Go unit tests, also outputs junit XML

None

Build and run the unit test service from `tools/golang/docker-compose.yml`, make sure it outputs what you want.

### DIFF
--- a/tools/golang/Dockerfile
+++ b/tools/golang/Dockerfile
@@ -27,7 +27,7 @@ ADD . /go/src/$DIR
 
 WORKDIR /go/src/github.com/apache/trafficcontrol
 
-RUN go get -v ./...
+RUN go list -f {{.Deps}} ./... | tr -d '[]' | tr -s ' ' '\n' | sed -re 's/vendor\///g' | grep golang.org/x/ | grep -v chacha | sort | uniq | tr '_' '.' | xargs go get -v
 
 FROM base AS lint
 
--- a/tools/golang/Dockerfile
+++ b/tools/golang/Dockerfile
@@ -45,7 +45,6 @@ FROM base AS unit
 VOLUME ["/junit"]
 
 RUN go get github.com/wadey/gocovmerge &&\
-    go get -u github.com/jstemmer/go-junit-report &&\
-    go get -v ./...
+    go get -u github.com/jstemmer/go-junit-report
 
 CMD /go/src/github.com/apache/trafficcontrol/tools/golang/unit-tests.sh
--- a/tools/golang/Dockerfile
+++ b/tools/golang/Dockerfile
@@ -42,4 +42,10 @@ CMD ["golangci-lint", "run", "./..."]
 
 FROM base AS unit
 
-CMD sh -c 'go test $(go list ./... | grep -v  -e experimental -e "test/router" -e "api/v1")'
+VOLUME ["/junit"]
+
+RUN go get github.com/wadey/gocovmerge &&\
+    go get -u github.com/jstemmer/go-junit-report &&\
+    go get -v ./...
+
+CMD /go/src/github.com/apache/trafficcontrol/tools/golang/unit-tests.sh
--- a/tools/golang/docker-compose.yml
+++ b/tools/golang/docker-compose.yml
@@ -32,5 +32,4 @@ services:
       dockerfile: tools/golang/Dockerfile
       target: unit
     volumes:
-      - ../..:/go/src/github.com/apache/trafficcontrol
       - ./junit:/junit
--- a/tools/golang/docker-compose.yml
+++ b/tools/golang/docker-compose.yml
@@ -32,4 +32,5 @@ services:
       dockerfile: tools/golang/Dockerfile
       target: unit
     volumes:
+      - ../..:/go/src/github.com/apache/trafficcontrol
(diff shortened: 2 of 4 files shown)


tools/golang/docker-compose.yml
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 ---
-version: '2.4'
+version: '2.3'
```
mhoppa: why the downgrade?

ocket8888: Because the version of docker engine available from "dockerrepo" on CentOS7 isn't compatible with `docker-compose.yml` version 2.4 - and the contents are perfectly compatible with 2.3 (apparently).

tools/golang/Dockerfile
```diff
@@ -42,4 +42,10 @@ CMD ["golangci-lint", "run", "./..."]
 
 FROM base AS unit
 
-CMD sh -c 'go test $(go list ./... | grep -v  -e experimental -e "test/router" -e "api/v14")'
+VOLUME ["/junit"]
+
+RUN go get github.com/wadey/gocovmerge &&\
+    go get -u github.com/jstemmer/go-junit-report &&\
+    go get -v ./...
```
mhoppa: do you need to rerun`go get ./...` here since it was run in the base layer?

ocket8888: Was it? I didn't realize. Thought the linter ran that. I can take this out.

mhoppa: yep! the base layer runs it

ocket8888:
Thanks, still want a review from @lbathina to make sure it fits her original use case.

lbathina:
Just saw. Will review it later tonight or tomorrow morning.

lbathina:
looks good to me.

lbathina:
can we have the coverage post somewhere for review?

ocket8888:
I can paste it into the #dev channel on the ATC slack, or put it somewhere on the wiki.

mhoppa:
would like to get this merged so we can replace the duplicated unit tests in the repo. @ocket8888 whats blocking this besides the merge conflict? A committer?

ocket8888:
There were still some issues running it with Docker, so I've fixed those now. Posted the results of running it against master at the time of this writing in the #general channel of the ATC slack. Should've gone in #dev, but I didn't think of that until looking at the comments here again. Whoops.

mhoppa:
:shipit:

tools/golang/docker-compose.yml
```diff
@@ -32,4 +32,4 @@ services:
       dockerfile: tools/golang/Dockerfile
       target: unit
     volumes:
-      - ../..:/go/src/github.com/apache/trafficcontrol
```
mhoppa: wait why the removal of this volume?

ocket8888: Couple reasons

1. Because the Dockerfile is already `ADD`ing the whole context to that path.
2. Using that volume meant that some of the temporary files being created were being output into the source directory owned by `root:root` with permissions `-rw-------` and in general if we did things like `go fmt` it'd modify the actual source when ran. Seemed dangerous.

mhoppa: well it makes me rebuild the container everytime I want to run it locally which is super annoying as it takes forever to build since we do `go get`

ocket8888: Well it shouldn't be `go get`-ing *everything* anymore - that was actually causing problems because it was fetching vendored dependencies (and some of ours are so old they don't exist anymore). So now it should only be fetching things from `golang.org/x/`.

mhoppa: still takes forever and will probably make me not use it locally :/

ocket8888: Why would you use Docker to use it locally?

I clocked the `go get` step at 20 seconds running locally, which I wouldn't consider 'forever' but also note that it would still need to do that step if the volume still existed. The only thing that could be faster is that `ADD` line for the repository that doesn't need to exist if it's in a volume, and I didn't put that in there. Not sure why it was in there, tbh.

So deleting this line didn't make anything slower - although that doesn't mean it couldn't help it be faster.

mhoppa: I like/prefer using docker locally for my work as it matches what will be seen in ci and will be consistent but thats personally preference.

We dont change our dependencies much so my thought of putting it together was I would have to just build the container once in awhile as dependencies change and be able to quickly verify all tests in the repo are passing with passing in the current code as a volume. 

my review isnt really voting anywho 🤷‍♂ so you can leave it as is

ocket8888: Ohh, I see. Well lemme ask you this: why do we need the `ADD` at all if there's going to be a volume? Do we? Can I just get rid of it?

mhoppa: Ahh I get what your saying now :) yeah we should be able to remove the ADD since its a volume.

ocket8888: No actually we can't unless we move the `go get`s to runtime. Because the volume isn't available during the build.

PR #4268, opened 2020-01-08
Fix Traffic Ops Perl endpoints to always be numbers

rob05c:
Fixes the serverchecks and to_extensions endpoints to always return
numbers, as they should.

Apparently, certain versions of Perl or its libraries sometimes turn
the numbers into strings. This forces it to always be a number.

These two endpoints are specifically used in the API tests. This
makes the API tests pass correctly.

We should probably make all numbers in all endpoints do this; but
because Perl is going away, that may not be worth it. But these two
at least let devs verify all API tests pass, instead of potentially
having to ignore failures and hope their code changes didn't cause
other failures in the same tests.

No new tests, tests already exist, and the entire purpose of this PR is to fix them.
No docs, no interface change.

- Traffic Ops

Run API tests. Request serverchecks and to_extensions endpoints, verify response is still correct.

- everything ever. But it's only a bug if the user has certain versions of Perl installed, and it's unclear what causes those versions to be installed or when or which versions. Probably.

## Additional Information

### DIFF
--- a/traffic_ops/app/lib/API/ServerCheck.pm
+++ b/traffic_ops/app/lib/API/ServerCheck.pm
@@ -74,7 +74,7 @@ sub read {
 	my @data;
 	while ( my $server = $rs->next ) {
 		my $v;
-		$v->{id}		= $server->id;
+		$v->{id}		= $server->id + 0;
 		$v->{hostName}		= $server->host_name;
 		$v->{profile}		= $server->profile->name;
 		$v->{adminState}	= $server->status->name;
@@ -84,7 +84,11 @@ sub read {
 		$v->{revalPending}	= \$server->reval_pending;
 		foreach my $col (qw/aa ab ac ad ae af ag ah ai aj ak al am an ao ap aq ar at au av aw ax ay az ba bb bc bd be bf/) {
 			if ( defined( $mapping{$col} ) && defined( $server->servercheck ) ) {
-				$v->{checks}->{ $mapping{$col} } = $server->servercheck->$col();
+				my $server_check_val = $server->servercheck->$col();
+				if ( defined($server_check_val) ) {
+					$server_check_val = $server_check_val + 0; # coerce to a number, if it's not null
+				}
+				$v->{checks}->{ $mapping{$col} } = $server_check_val;
 			}
 		}
 		push( @data, $v );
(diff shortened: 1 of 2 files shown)


traffic_ops/app/lib/API/ServerCheck.pm
```diff
@@ -84,7 +84,7 @@ sub read {
 		$v->{revalPending}	= \$server->reval_pending;
 		foreach my $col (qw/aa ab ac ad ae af ag ah ai aj ak al am an ao ap aq ar at au av aw ax ay az ba bb bc bd be bf/) {
 			if ( defined( $mapping{$col} ) && defined( $server->servercheck ) ) {
-				$v->{checks}->{ $mapping{$col} } = $server->servercheck->$col();
+				$v->{checks}->{ $mapping{$col} } = $server->servercheck->$col() + 0;
```
rawlinp: When I test this endpoint locally I now get the following warning hundreds of time (one for each server I'm guessing?):
```
Use of uninitialized value in addition (+) at lib/API/ServerCheck.pm line 87
```
If the column's value is null, this code converts it into 0, which might be a bad unintended consequence. I tried this out, and it seems to work (keeps the null values null but makes sure the integer values are actually integers not strings):
```
                my $foo = $server->servercheck->$col();
		if ( defined($foo) ) {
                    $foo = $foo + 0;
		}
		$v->{checks}->{ $mapping{$col} } = $foo;
```
Don't you just love Perl? Can you try this out on your end and see if that works?

rob05c: Sure, I can do that, to preserve the existing behavior.

It makes me a little worried though, the documentation says nothing about those values possibly being null, and I had no idea they could be null in the db. In fact, I checked it, and the docs don't even say the type, it just has a number in the example. We could change the docs to say it's possibly null. But then that could cause issues if we decide to make it always a number when we convert it to Go. But, I have no idea who is using this endpoint for what, and if we'd break anyone making nulls in the db become 0's.

So, I don't know what the right answer is, for docs or code. But IMO we should try to remember to revisit it when we rewrite it in Go.

ocket8888: TBH, I avoided typing information in a lot of the Perl endpoints because I didn't understand why sometimes things were numbers and sometimes they were booleans or strings or strings containing numbers or booleans. Just some fun history.

I'd vote that if things can optionally be null the docs should be updated to match the behavior, but I'm fine with silently changing strings containing numbers into just numbers. Since with Perl you generally had to try both anyway.

rob05c: Changed.

rawlinp: I'm not sure what to set either, but I run TO-Perl using morbo on macOS (the `traffic_ops/app/bin/start.pl` script). Something there might be setting something, but your guess is as good as mine.

PR #4282, opened 2020-01-09
Fix role name not being set on users get

mhoppa:

This PR fixes a problem in 4.0.x where the users endpoint does not populate the rolename key as it has duplicate db keys colliding between here and https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/user/user.go#L207

Note this is from https://github.com/apache/trafficcontrol/pull/3996. I just did not want to pull in a whole rewrite

- Traffic Ops

Bring up TO and do a get on TO API users GET path

- 4.0.x (RC1)


mhoppa:
@ocket8888 do you see any issues just pulling this out of the PR #3996 for 4.0.x release. I did not see anything that stuck out to me but since you are the author wanted your opinion

mitchell852:
I put the backport label on this to get on the @rawlinp radar but it's not technically a backport...

ocket8888:
No, as long as it works that should be fine.

PR #4286, opened 2020-01-10
Fixed wrong import order and Py3-incompatibility

### LINKED ISSUE #4285: Kickstart script broken
## I'm submitting a ...
- bug report

- tools (`kickstart_create_network_line.py`)

## Current behavior:
There's a bug in the script:
```
  File "./kickstart_create_network_line.py", line 25
    from __future__ import print_function
SyntaxError: from __future__ imports must occur at the beginning of the file
```

## Expected / new behavior:
There shouldn't be a bug in the script

Run the script

ocket8888:
- [x] This PR fixes #4285 

Changes the import order to be correct (`__future__` first) and fixes an incompatibility with Python 3 (tested CPython versions 3.8.1 and 2.7.17).

 - Tools (`kickstart_create_network_line.py`)
No documentation because this tool is undocumented - and should (hopefully) be on its way out.

Try to run the script with a valid network.cfg file - if you don't have one it'll crash. If you don't have one that sets `BOND_DEVICE` and `BOND_OPTIONS` it'll crash. Probably others.

- master
- 4.0.0 (RC1)

- [x] Tests are unnecessary

### DIFF
--- a/misc/kickstart_create_network_line.py
+++ b/misc/kickstart_create_network_line.py
@@ -19,10 +19,10 @@
     network links in /sys/class/net/*, then emits a ks.cfg network line.
     '''
 
+from __future__ import print_function
 import os
 import re
 
-from __future__ import print_function
 
 global TO_LOG
 # This "logs" to stdout which is captured during kickstart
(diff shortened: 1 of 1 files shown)


misc/kickstart_create_network_line.py
```diff
@@ -170,8 +170,7 @@ def useable_interfaces(net_devs, nc, iface_speed):
         #  if not it is anyway.
         # Thus we are doing a bond of some sort.
         # This gives us the fastest interfaces first:
-        speeds = net_devs.keys()
-        speeds.sort(reverse=True)
+        speeds = [k for k in sorted(net_devs.keys(), reverse=True)]
```
rawlinp: you just couldn't resist

ocket8888: That was the Python3 compatibility error! (`dict_keys` has no member `sort`)

rawlinp: Oh, I see. I thought you were just making it more pythonic!

ocket8888: No, although that was my pleasure.

PR #4289, opened 2020-01-13
Fix atstccfg caching

rob05c:
Fixes atstccfg caching, which was breaking when the new cache file
was smaller than the old, because it wasn't truncating the file.

Also changes deliveryservice_server to always fetch all Servers and
Delivery Services and cache them, because it's faster for an ORT
run to reuse and cache them all.

Testing a complete `badass` on a largeish CDN, old Perl ORT against Perl TO takes ~4m, atstccfg without fixed caching takes ~5m, with this fix takes ~2.5m. So, this probably isn't a huge deal. But good to fix.

Includes unit tests.
No docs, no interface change.

- Traffic Ops ORT

Run unit tests. Run ORT, verify it's still correct and faster than without this PR.

It's in master, but it's not really a "bug" in that it still produces the right output, just not as fast as it could.

## Additional Information

### DIFF
--- a/traffic_ops/ort/atstccfg/toreq/caching.go
+++ b/traffic_ops/ort/atstccfg/toreq/caching.go
@@ -77,7 +77,7 @@ func WriteCacheJSON(tempDir string, cacheFileName string, obj interface{}) {
 
 	objPath := filepath.Join(tempDir, cacheFileName)
 	// Use os.OpenFile, not os.Create, in order to set perms to 0600 - cookies allow login, therefore the file MUST only allow access by the current user, for security reasons
-	objFile, err := os.OpenFile(objPath, os.O_RDWR|os.O_CREATE, 0600)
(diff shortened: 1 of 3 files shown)


traffic_ops/ort/atstccfg/toreq/caching.go
```diff
@@ -84,7 +84,12 @@ func WriteCacheJSON(tempDir string, cacheFileName string, obj interface{}) {
 	}
 	defer objFile.Close()
 
-	if _, err := objFile.Write(objBts); err != nil {
+	if err := objFile.Truncate(0); err != nil { // must truncate, or Write will leave whatever existed after the beginning, if the file already existed and was longer than our new write.
```
rawlinp: Could we just add the `os.O_TRUNC` flag to `os.OpenFile` above? That would be more straightforward than manually truncating the file after opening and writing specifically to offset 0.

rob05c: Changed.

PR #4291, opened 2020-01-14
Added documentation for atstccfg

ocket8888:

Adds a documentation page for atstccfg under "Tools", adds a README markdown file to the atstccfg directory, and adds usage documentation to the atstccfg package's GoDoc. Also adds the `-v`/`--version` option which causes it to print out the version and then exit.

### Edit
Also added the `-h`/`--help` option which used to sort-of function using the `flag` package's bultin behavior. Basically, the option is now shown in usage output, and when used causes the exit code to be a success instead of a failure. Still prints to stderr.

- Documentation
- Traffic Ops ORT (`atstccfg`)

Build and read the documentation to make sure it renders right and is accurate. Build godoc documentation to be sure it renders right and is accurate. Check out the rendered markdown to make sure it renders right and is accurate.
- [x] Tests are unnecessary

### DIFF
--- a/docs/source/tools/atstccfg.rst
+++ b/docs/source/tools/atstccfg.rst
@@ -22,6 +22,8 @@ atstccfg
 ********
 :program:`atstccfg` is a tool for generating configuration files server-side on :abbr:`ATC (Apache Traffic Control)` cache servers. It stores its generated/cached files in ``/tmp/atstccfg_cache/`` for re-use.
 
+.. warning:: :program:`atstccfg` does not have a stable command-line interface, it can and will change without warning. Scripts should avoid calling it for the time being.
+
 The source code for :program:`atstccfg` may be found in :atc-file:`traffic_ops/ort/atstccfg`, and the Go module reference is :atc-godoc:`traffic_ops/ort/atstccfg`.
 
 Usage
--- a/docs/source/tools/atstccfg.rst
+++ b/docs/source/tools/atstccfg.rst
@@ -72,7 +72,7 @@ Options
 
 	If given, SSL certificate errors will be ignored when communicating with Traffic Ops.
 
-	.. caution:: For (hopefully) obvious reasons, the use of this option in production environments is discouraged.
+	.. caution:: The use of this option in production environments is discouraged.
 
 .. option:: -t TIMEOUT, --traffic-ops-timeout-milliseconds TIMEOUT
 
--- a/docs/source/tools/atstccfg.rst
+++ b/docs/source/tools/atstccfg.rst
@@ -22,13 +22,13 @@ atstccfg
 ********
 :program:`atstccfg` is a tool for generating configuration files server-side on :abbr:`ATC (Apache Traffic Control)` cache servers. It stores its generated/cached files in ``/tmp/atstccfg_cache/`` for re-use.
 
-.. warning:: :program:`atstccfg` does not have a stable command-line interface, it can and will change without warning. Scripts should avoid calling it for the time being.
+.. warning:: :program:`atstccfg` does not have a stable command-line interface, it can and will change without warning. Scripts should avoid calling it for the time being, as its only intended caller is :term:`ORT`.
 
 The source code for :program:`atstccfg` may be found in :atc-file:`traffic_ops/ort/atstccfg`, and the Go module reference is :atc-godoc:`traffic_ops/ort/atstccfg`.
 
 Usage
 =====
-``atstccfg [-u TO_URL] [-U TO_USER] [-P TO_PASSWORD] [-n] [-r N] [-e ERROR_LOCATION] [-w WARNING_LOCATION] [-i INFO_LOCATION] [-g] [-s] [-t TIMEOUT] [-a MAX_AGE] [-l]``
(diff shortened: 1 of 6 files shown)


rob05c:
The documentation should state that `atstccfg` does not have a stable interface. It's a utility for ORT, and we don't provide any API guarantee that its interface won't have breaking changes from one minor version of TC to the next (like other APIs such as ORT have). The docs should inform people of that.

traffic_ops/ort/atstccfg/README.md
```diff
@@ -0,0 +1,43 @@
...
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+# atstccfg
+atstccfg is a tool for generating configuration files server-side on ATC cache servers.
+
+## Usage
```
rob05c: Would it be good to inform users here that they can get all this information with `atstccfg -h`?

ocket8888: Oh, I thought that was just a by-product of printing usage information when given an unrecognized option. Why does that exit with a failure?

ocket8888: I added it as an option so it gets printed in its own output and exits with a success, plus is now documented.

ocket8888:
> _"The docs should inform people of that [the fact that its interface is unstable]"_

Done.

traffic_ops/ort/atstccfg/README.md
rob05c: >Why does that exit with a failure?

Because it didn't successfully generate a config file. Linux exit codes aren't _necessarily_ errors, sometimes they're useful just to distinguish operation. I don't feel that strongly about it though.

rob05c: >a by-product of printing usage information when given an unrecognized option.

In the code, it sort of is. But it works to tell people that. 
I think we should make it print the full config help when given no arguments, too. Just haven't gotten around to that yet.

docs/source/tools/atstccfg.rst
```diff
@@ -0,0 +1,110 @@
...
+
+.. option:: -r N, --num-retries N
+
+	The number of times to retry getting a file if it fails. Default: 5
+
+.. option:: -s, --traffic-ops-insecure
+
+	If given, SSL certificate errors will be ignored when communicating with Traffic Ops.
+
+	.. caution:: For (hopefully) obvious reasons, the use of this option in production environments is discouraged.
```
mhoppa: although agreed we probably should take out the first part, should just be `the use of this option in production environments is discouraged`

traffic_ops/ort/atstccfg/README.md
```diff
@@ -0,0 +1,47 @@
...
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+# atstccfg
+atstccfg is a tool for generating configuration files server-side on ATC cache servers.
+
+!!! Warning !!!
+    atstccfg does not have a stable command-line interface, it can and will change without warning. Scripts should avoid calling it for the time being.
```
mhoppa: should this be in bold and italicized to stick out?

ocket8888: Unfortunately there's no way to really add proper admonitions to Markdown. I can nest certain kinds of HTML tags, though, so I'll try to wrap this in a "bring-attention-to" tag (`<b>`) or a "strong importance" tag (`<strong>`), though italics are typically reserved for verbal stress or emphasis, to indicate that a set of text is set aside from a main body of text for some reason, or to mark the defining instance of a word.

ocket8888:
There aren't any required arguments, because any information that *must* be supplied may be given by either a command-line option-argument or through the execution environment.

mhoppa:
well that still means those values are required just by how you set it can differ. 🤷‍♂  not a blocking comment as the tool will output an error if they are not set.

rawlinp:
Other than calling `atstccfg` directly, I don't really see an easy way of testing/verifying ATS config generation that doesn't involve having a full ORT/ATS test environment. If anything, we should be striving to make testing ATS config gen as easy as possible, rather than harder. Especially given that we won't be able to use the TO API for config gen, using `atstccfg` directly should be made as easy as possible to make up for that difference in testing convenience.

rob05c:
>we should be striving to make testing ATS config gen as easy as possible

That's exactly my point: it _looks_ easy, but it's not _correct_. It's deceptive and insidious (like so much Perl). ORT and atstccfg are both complex, and have complex interactions. For example, ORT has diff checks, which may correctly call atstccfg, but then never place the file if it incorrectly thinks there wasn't a diff. Another good example is the recent caching bug: a single `atstccfg` run wouldn't catch it, and in fact any number of combinations of `atstccfg` file runs would have no issue. Only if a person (or Integration Test) ran a full ORT would the issue show up. Or ORT may pass unexpected parameters, or not call `atstccfg` at all, or manipulate its output, or any number of other things.

>using atstccfg directly should be made as easy as possible

I still maintain that no person should ever run `atstccfg` directly. I get that we don't have an Integration Test framework for ORT yet, and besides smaller scoped tests are so much easier they're sometimes worth the inaccuracy. But I still maintain that unless they're developing it, nobody should ever run it directly. It _looks_ easy, but it's frequently deceptively wrong.

PR #4293, opened 2020-01-14
tc package: Split enum into its own package

zrhoffman:

This PR splits `enum.go` within the `tc` library into its own package. Reasons for doing this:

* Looking at the line count of source files in the `tc` package (`wc -l *.go | sort -g`), `enum.go` is 727 lines, more lines than the next-longest file in the package (`deliveryservices.go`) by 122 lines. Excluding tests and vendor files, only 10 go sources (out of 584) in the whole project are longer than this.

* [The godoc for the `tc` package](https://godoc.org/github.com/apache/trafficcontrol/lib/go-tc) currently begins with this line:
    > Package enum contains enumerations and strongly typed names. 

Behavior should be unchanged by this PR, and the documentation contains no references to the `tc` library or `enum.go`, so changes to documentation are not necessary.

Because no code is added, existing tests are sufficient.

- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Experimental - Traffic Router Golang

* Run Traffic Ops Go unit tests
* Run Traffic Ops API tests
* Run Traffic Monitor unit tests
* Run atstccfg unit tests
* Run Traffic Router Golang's unit test

- [x] I have explained why tests in this PR are unnecessary

## Additional Information

The contents of 9010b50, the package rename itself, is a result of using Goland's "refactor" feature to move  `enum.go`.


rob05c:
Personally, I'd really prefer we kept this stuff in `tc`. It's used so much in so many places, it makes the code much terser to have e.g. `tc.DeliveryServiceName` everywhere than `enum.DeliveryServiceName`. The brevity was a big part of naming it `tc`.

I think it's also good to keep as much as we can in a library that indicates it's Traffic Control, so it's easy for people not intricately familiar with the code to immediately see it's ours and not a third-party. And we could do something like `tcenum.DeliveryServiceName`, but that's even more verbose.

zrhoffman:
I can see the benefit of a short package name like `tc`. I'll submit an issue for the GoDoc and open a PR for breaking that file into multiple files later on.

PR #4296, opened 2020-01-15
Fix ip_allow config generation for mids to include rascal servers

mhoppa:

Currently in ip_allow.config generation for mids it does not include rascal server IPs causing them to be blocked on attempting to pull astats.

The behavior in the perl implementation is that it would include all servers of type Rascal as well as EDGEs that are in either parent/secondary cachegroup of the mid. https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/API/Configs/ApacheTrafficServer.pm#L2210-L2228

### TO GO API Bug
In the go TO implementation it checks the parent/secondary cachegroup on both Rascal AND Edge servers which is a mismatch of logic.

Going from `is rascal OR (is edge AND edge.cg in (mid parent cg, mid secondary cg)` to `is rascal OR edge AND s.cg in (mid parent cg, mid secondary cg)`

### atstccfg Bug
In the atstccfg cfg it does not attempt to include rascal servers 

- Traffic Ops
- atstccfg

Generate the ip_allow.cfg for a mid and ensure the rascal servers are included via TO API and atstccfg


mhoppa:
Looks to have the same issue in atstccfg as well https://github.com/apache/trafficcontrol/blob/master/traffic_ops/ort/atstccfg/cfgfile/ipallowdotconfig.go#L112-L117 need to test and look into that

mhoppa:
TO GO and ATSTCCFG should both be fixed for mid ip_allow generation now

mhoppa:
Yeah I can do that!

mhoppa:
API tests are modified to be more robust incase the rascal server is included in a range instead on its own. CRED goes to @rob05c for the `IP4ToNum`, `IP4InRange` implementations :)

PR #4302, opened 2020-01-15
Fixed unavailable cache servers not being highlighted in red

### LINKED ISSUE #3504: Missing variable declaration in JS files
If a variable is not declared as a local variable, it becomes a global variable by default, which may be unintentional and could lead to unexpected behavior.

In JavaScript, if a variable is used in a function but not declared as a local variable, it becomes a global variable by default. This can have unintended consequences: unlike local variables, global variables can be read and modified by all functions. If different functions use the same global variable, they may end up overwriting each others values, leading to subtle and difficult to diagnose bugs.

Recommendation
Check whether the variable in question was meant to be local; if so, declare it by means of a var declaration. If the variable is really meant to be global, it is best to document this fact by inserting a global var declaration at the beginning of the source file.

[More Info About Issue](https://lgtm.com/projects/g/apache/trafficcontrol/alerts/?mode=tree&ruleFocus=3970086)

ocket8888:
- [x] This PR fixes #3504

fixed an issue where caches that were marked down because "availableBandwidthKbps" was too high/low weren't getting marked as red in the GUI

fixed an issue where caches that were deleted/no longer the TM's responsibility (don't appear in monitoring.json) weren't getting removed when they were in the first row of the table

fixed an issue where extra rows were continuously added for servers instead of just updating existing rows

stopped styling things in JS (for the cache-states table)

bumped up the difference in color between the striped rows

- Traffic Monitor

Build and run TM in a valid environment, check out the servers table. Things that have too high/low "availableBandwidthInKbps" should now be highlighted red. Also, the number of table rows should always exactly match the number of cache servers the TM is monitoring.

- master
- 3.x
- 4.0 (RC1)

- [x] Tests are unnecessary

### DIFF
--- a/traffic_monitor/static/index.html
+++ b/traffic_monitor/static/index.html
@@ -48,11 +48,18 @@ under the License.
 		}
 
 		tbody tr:nth-child(even) {
-			background: #dfe
+			background: #ced;
 		}
 		tbody tr:nth-child(odd) {
-			background: #fff
+			background: #fff;
 		}
+		#cache-states td:nth-child(n+4) {
+			text-align: right;
+		}
+		#cache-states td:first-child {
+			white-space: nowrap;
+		}
+
 
 		li.endpoint {
 			margin: 4px 0;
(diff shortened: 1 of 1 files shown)


ocket8888:
test this again, please

mitchell852:
retest this please

mattjackson220:
This looks great! Tested locally and TM UI appropriately turns red when availableBandwidthInKbps is too high or too low. Servers get added/deleted from table successfully.

PR #4303, opened 2020-01-15
Updated GetDeliveryServiceByXMLID function to return a DS nullable struct

ZugNZwang:

Replaces the struct returned by the `GetDeliveryServiceByXMLID` client function to `DeliveryServiceNullable` instead of `DeliveryService`.

- Traffic Control Go Client 

Run the traffic ops tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added deep coverage zone routing percentage to the Traffic Portal dashboard.
 - Added a `traffic_ops/app/bin/osversions-convert.pl` script to convert the `osversions.cfg` file from Perl to JSON as part of the `/osversions` endpoint rewrite.
 - Added [Experimental] - Emulated Vault suppling a HTTP server mimicking RIAK behavior for usage as traffic-control vault.
+- Added Traffic Ops Client function that returns a Delivery Service Nullable Response when requesting for a Delivery Service by XMLID 
 
 ### Changed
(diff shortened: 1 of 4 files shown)


rob05c:
Yeah, like @ocket8888 says, this will break existing things, but it also breaks the stability of the Go library, and it will fundamentally break Go Modules if and when we switch to them (which we'll probably have to at some point).

The way we've typically done this in the past is to add a new function with a similar name, e.g. `DeliveryServiceByXMLID` or `GetDeliveryServiceByXMLIDNullable`, and add a Deprecated comment to the old one.

ocket8888:
The [mailing list decided](https://www.mail-archive.com/search?l=dev@trafficcontrol.apache.org&q=subject:%22Re%5C%3A+Deprecation+period+for+TO+Client+Methods%5C%3F%22&o=newest&f=1) that TO client methods don't have a deprecation period - but this will need a line in CHANGELOG.md

ZugNZwang:
> but this will need a line in CHANGELOG.md
Would this go under `Added`?

ocket8888:
Uhh sure. If you're doing what rob said and adding a new function it goes under Added. Otherwise I would think under Changed.

rawlinp:
Yeah, in addition to the mailing list discussion I'd also add that going the deprecation route would be great if we actually cleaned up deprecated functions, stopped increasing the usage of already-deprecated functions, etc. But historically nobody has really gone back to clean things up, and I've definitely encountered new code that used already-deprecated client methods. At least if you make a breaking change to the TO Go client you're forced to fix all the existing usages in the repo (otherwise the build fails), and we don't end up w/ 3 different client methods that all do the same thing.

PR #4305, opened 2020-01-17
Prevents line breaks in delivery service remapText field (UI and API)

### LINKED ISSUE #4298: Line breaks in DS raw remap text will result in a malformed remap.config file
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
When creating or editing a delivery service of type HTTP*, DNS* or ANY_MAP, a line break in the `remapText` field will result in a malformed remap.config file being deployed to all of the edge caches assigned to the delivery service which can have a negative impact across multiple delivery services (if multiple delivery services are assigned to the cache).

When remapText = foo[line break]bar, the applicable line in remap.config will look like this:

```
map	http://edge.demo1.mycdn.ciab.test/     http://origin.infra.ciab.test/  foo
bar
map	http://edge.demo2.mycdn.ciab.test/     http://origin2.infra.ciab.test/
```

## Expected / new behavior:
On API DS create/update, any line breaks in the `remapText` field need to be removed.

1. In TP, create or edit an HTTP*, DNS* or ANY_MAP delivery service and enter a line break in the `Raw Remap Text` field. Note: a textarea has been provided for readability of the `Raw Remap Text` field, however, it has the downside of allowing line breaks.
2. Navigate to the assigned servers for DS. Pick one.
3. View config files for the server.
4. View the contents of remap.config file and notice the line that doesn't start with `map`

mitchell852:

Rejects delivery service remapText with line breaks as it will result in a malformed remap.config file.

- [x] This PR fixes #4298 

- Traffic Ops
- Traffic Portal

2. Run the API tests
3. Manually verify in TP by trying to enter a line break in the `Raw Remap Text` delivery service form field for a DNS, HTTP or ANY_MAP delivery service.
4. Attempt to create or update a delivery service using the API where remapText contains \n or \r. It should fail.

- master (6b11bd9)
- 3.1.0
- 4.0.0 (RC1)


traffic_ops/testing/api/v14/deliveryservices_test.go
```diff
@@ -189,6 +190,39 @@ func UpdateNullableTestDeliveryServices(t *testing.T) {
 	}
 }
 
+// UpdateDeliveryServiceWithInvalidRemapText ensures that a delivery service can't be updated with a remap text value with a line break in it
```
ocket8888: nit: GoDoc missing period

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -534,7 +534,8 @@ <h3>Current Value</h3>
                         </div>
                         </label>
                         <div class="col-md-10 col-sm-10 col-xs-12">
-                            <textarea id="remapText" name="remapText" class="form-control" ng-model="deliveryService.remapText" rows="3"></textarea>
+                            <textarea id="remapText" name="remapText" class="form-control" ng-model="deliveryService.remapText" ng-pattern="/^[^\n\r]*$/" rows="3"></textarea>
```
ocket8888: if this can't have multiple lines, why not just make it an `input[type="text"]`?

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -534,7 +534,8 @@ <h3>Current Value</h3>
                         </div>
                         </label>
                         <div class="col-md-10 col-sm-10 col-xs-12">
-                            <textarea id="remapText" name="remapText" class="form-control" ng-model="deliveryService.remapText" rows="3"></textarea>
+                            <textarea id="remapText" name="remapText" class="form-control" ng-model="deliveryService.remapText" ng-pattern="/^[^\n\r]*$/" rows="3"></textarea>
```
ocket8888: same as above RE `textarea` vs `input[type="text"]`

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.anyMap.tpl.html
```diff
@@ -262,7 +262,8 @@ <h3>Current Value</h3>
                         </div>
                         </label>
                         <div class="col-md-10 col-sm-10 col-xs-12">
-                            <textarea id="remapText" name="remapText" class="form-control" ng-model="deliveryService.remapText" rows="3" required></textarea>
+                            <textarea id="remapText" name="remapText" class="form-control" ng-model="deliveryService.remapText" ng-pattern="/^[^\n\r]*$/" rows="3" required></textarea>
```
ocket8888: and one more time

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
mitchell852: remap text gets long and i want the word wrapability of a TA.

ocket8888: Yeah, I suppose that makes sense. You _could_ make the number of rows dynamic so that it's always exactly big enough to contain its contents. But I won't make you do that to fix a bug.

mitchell852: textareas are stretchable so i suppose if the content goes over the 3 rows, they can stretch the TA

rawlinp: I think because it's easier to read long inputs with text-wrapping.

mitchell852:
> The regular expression you're using isn't multi-line; `^` and `$` match the beginning and end of a line, respectively. So I can still do this:
> 
> ```js
> >> temp0.name
> <- "remapText"
> >> temp0.tagName
> <- "TEXTAREA"
> >> temp0.attributes[4]
> <- ng-pattern="/^[^\\n\\r]*$/"
> >> temp0.value
> <- "
> this is a test
> "
> >> temp0.validity.patternMismatch
> <- false
> >> temp0.classList.contains("ng-invalid")
> <- false
> >> temp0.classList.contains("ng-valid")
> <- true
> ```

well in the UI i am unable to enter anything on 2 lines so that's what i was going for

ocket8888:
No, you still can.

Depiction of the problem

That'll get rejected by the API but the form shows no problems. According to the issue, the content of this field is:

> _"appended to the one line [in remap.config for most Delivery Services]"_

so this value would still exhibit the problem. So it ought to not be considered correct - especially since submission will result in an error.

mitchell852:
> so this value would still exhibit the problem. So it ought to not be considered correct - especially since submission will result in an error.

hmm. i can't replicate. this submits fine in both chrome and firefox.

image

it makes it past the api so it appears to be getting trimmed.

mitchell852:
retest this please

PR #4306, opened 2020-01-17
Fix omitempty issue for ATS config metadata generation

### LINKED ISSUE #4299: /servers/{{server}}/configfiles/ats Go response differs from Perl
# I'm submitting a ...
- bug report

- Traffic Ops
- Traffic Ops ORT

## Current behavior:
Traffic Ops Perl returns responses that look like:
```json
{ "info": {
	"profileId": 9,
	"toUrl": "https://trafficops.infra.ciab.test:443/",
	"serverIpv4": "172.19.0.2",
	"serverTcpPort": 80,
	"serverName": "edge",
	"cdnId": 2,
	"cdnName": "CDN-in-a-Box",
	"serverId": 12,
	"profileName": "ATS_EDGE_TIER_CACHE"
},
"configFiles": [
	{
		"fnameOnDisk": "astats.config",
		"location": "/etc/trafficserver",
		"apiUri": "/api/1.4/profiles/ATS_EDGE_TIER_CACHE/configfiles/ats/astats.config",
		"scope": "profiles"
	},
	{
		"fnameOnDisk": "myconfig",
		"location": "/some/path",
		"url": "https://google.com/",
		"scope": "profiles"
	}
]}
```
While Go returns responses that look like:
```json
{ "info": {
	"cdnId": 2,
	"cdnName": "CDN-in-a-Box",
	"serverId": 12,
	"serverIpv4": "172.19.0.2",
	"serverName": "edge",
	"serverTcpPort": 80,
	"profileId": 9,
	"profileName": "ATS_EDGE_TIER_CACHE",
	"toRevProxyUrl": "",
	"toUrl": "https://trafficops.infra.ciab.test:443/"
},
"configFiles": [
	{
		"fnameOnDisk": "hosting.config",
		"location": "/etc/trafficserver/",
		"apiUri": "/api/1.4/servers/edge/configfiles/ats/hosting.config",
		"url": "",
		"scope": "servers"
	},
	{
		"fnameOnDisk": "myconfig",
		"location": "/some/path",
		"apiUri": "",
		"url": "https://github.com/",
		"scope": "cdns"
	}
]}
```
Those are _mostly_ the same, but with a couple of key differences. In Perl, `toRevProxyUrl` is omitted if it's empty (and/or null?), but in Go that's no longer the case. That may or may not be a big deal.

The other one is a bigger deal. If you have a Parameter with the name URL, Traffic Ops won't generate that config file. Instead, it puts that Parameter's Value in as the `url` field of the `configFilesResponse`, _instead of including an `apiUri` field_. As [the docs](https://traffic-control-cdn.readthedocs.io/en/latest/api/servers_server_configfiles_ats.html#id3) say:

> _"Exactly one of these fields [`apiUri` and `url`] is guaranteed to exist for any given configuration file - although “apiUrl” is far more common."_

That's because ORT will fetch a configuration file's contents from one of those places if it exists, and then move on without checking the other (not sure which at the moment).

Additionally, the new "meta configfiles route" doesn't appear to actually generate entries for the special "URL" Parameters, which is potentially a big problem.

## Expected / new behavior:
The Go version should spit out exactly one of those fields for each "configFiles" array member. Ideally, `toRevProxy` would also be omitted if it's empty (and/or null?).

TO GO should also faithfully generate `configFiles` entries that populate the `url` field if a URL Parameter is in assigned to the server's Profile.

Request `/servers/{{server}}/configfiles/ats` both with and without URL Parameters existing, and against both Perl and Go implementations.

## Anything else:
The special "URL" Parameter is undocumented (but to be fair so were all of the other special name values until a couple of months ago) and should be documented, but I'm going to open a separate issue about that.

rawlinp:
If empty, the 'url' and 'apiUri' fields should be omitted from the marshalled JSON, which is how Perl handled it. This fixes ORT's ability to use "URL" parameters which ORT fetches places them on the filesystem like config files. Without this fix, ORT requests `https://mytrafficops.example.com/` since `apiUri == ""` instead of the given URL in the parameter.

- [x] This PR fixes #4299 

- Traffic Ops
- Traffic Ops ORT

No docs or changelog necessary since this just fixes a regression in the API.

Run the Go unit tests (`go test ./lib/... ./traffic_ops/traffic_ops_golang/... ./traffic_monitor/... ./traffic_stats/... ./grove/... ./traffic_ops/ort/atstccfg/...`) and make sure they all pass.

- master
- 4.0.0-RC1

- [x] This PR includes tests

### DIFF
--- a/lib/go-atscfg/meta_test.go
+++ b/lib/go-atscfg/meta_test.go
@@ -109,6 +109,11 @@ func TestMakeMetaConfig(t *testing.T) {
 			FileNameOnDisk: "custom.config",
 			Location:       "/my/location/",
 		},
+		"external.config": ConfigProfileParams{
+			FileNameOnDisk: "external.config",
+			Location:       "/my/location/",
+			URL:            "http://myurl/remap.config",
+		},
 	}
 	uriSignedDSes := []tc.DeliveryServiceName{"mydsname"}
 	dses := map[tc.DeliveryServiceName]struct{}{"mydsname": {}}
@@ -233,6 +238,14 @@ func TestMakeMetaConfig(t *testing.T) {
 				t.Errorf("expected scope '%v', actual '%v'", expected, cf.Scope)
 			}
 		},
+		"external.config": func(cf tc.ATSConfigMetaDataConfigFile) {
+			if expected := "/my/location/"; cf.Location != expected {
+				t.Errorf("expected location '%v', actual '%v'", expected, cf.Location)
+			}
+			if expected := string(tc.ATSConfigMetaDataConfigFileScopeCDNs); cf.Scope != expected {
+				t.Errorf("expected scope '%v', actual '%v'", expected, cf.Scope)
+			}
+		},
 	}
 
 	for _, cfgFile := range cfg.ConfigFiles {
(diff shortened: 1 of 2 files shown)


mhoppa:
retest this please

PR #4307, opened 2020-01-17
Add an optimistic quorum feature to peering within Traffic Monitor to prevent false negatives from propagating to downstream components when network isolation occurs

elsloo:
## What does this PR (Pull Request) do? This PR adds an optimistic quorum to peering within Traffic Monitor to prevent false negative states from propagating to downstream components if peering is lost due to losing network connectivity.

- Documentation
- Traffic Monitor

Run the unit test, read the documentation on the feature, and experiment with different settings and numbers of peers if manually integration testing.


traffic_monitor/conf/traffic_monitor.cfg
```diff
@@ -5,6 +5,7 @@
 	"http_timeout_ms": 2000,
 	"peer_polling_interval_ms": 5000,
 	"peer_optimistic": true,
+	"peer_optimisitc_quorum_min": 0,
```
zrhoffman: Any reason not to add this config setting to [the CDN-in-a-Box TM config](https://github.com/apache/trafficcontrol/blob/RELEASE-4.0.0-RC1/infrastructure/cdn-in-a-box/traffic_monitor/traffic_monitor.cfg)?

elsloo: For posterity we should, however, the feature will not work unless there are a minimum of three Traffic Monitors. I'll add it to the PR.

PR #4320, opened 2020-01-22
Deprecate api_capabilities/:id route

### LINKED ISSUE #3824: Rewrite /api_capabilities/{{id}} to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/api_capabilities/{{id}}` is handled by Perl

## Expected / new behavior:
`/api_capabilities/{{id}}` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`
- [ ] `PUT`
- [ ] `DELETE`

N/A

mhoppa:

- [x] This PR fixes #3824  

- Documentation
- Traffic Ops

Build perl TO API with change and make a GET/PUT/DELETE request and ensure a deprecation alert is returned

### DIFF
--- a/docs/source/api/api_capabilities_id.rst
+++ b/docs/source/api/api_capabilities_id.rst
@@ -22,6 +22,8 @@ Manages a specific API capability.
 
 ``GET``
 =======
+.. deprecated:: 1.1
+
 Get an API-capability mapping by id.
 
 :Auth. Required: Yes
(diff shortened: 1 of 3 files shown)


docs/source/api/api_capabilities_id.rst
```diff
@@ -22,6 +22,8 @@ Manages a specific API capability.
 
 ``GET``
 =======
+.. deprecated:: 1.1
```
ocket8888: I think this was actually deprecated in ATC version 4.0, not API version 1.1. Or was it in API version 1.5? That is our current API version, but the deprecation decision predates this notice, and was decided just prior to cutting a 4.x RC. This is the type of stuff that confuses me when a service has two versioning systems.

mhoppa: right :( lets do ``` .. deprecated:: ATCv4``` I think that makes the most sense. Let me change all my current deprecate PRs

PR #4327, opened 2020-01-23
Rewrite to_extension POST

### LINKED ISSUE #3843: Rewrite /to_extensions to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/to_extensions` is handled by Perl

## Expected / new behavior:
`/to_extensions` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [ ] `GET`
- [x] `POST`

N/A

mhoppa:

- [x] This PR partially addresses #3843 (only the POST portion)

- Documentation
- Traffic Control Client 
- Traffic Ops

Manually testing - build TO with change and make POST requests to the to_extensions endpoint
Automate testing - run API tests


mhoppa:
retest this please

docs/source/api/to_extensions.rst
```diff
@@ -103,6 +103,8 @@ Response Structure
 
 ``POST``
 ========
+.. note:: As of ATCv4 ``POST /to_extensions`` only supports CHECK_EXTENSION extensions. As prior TO would fail everytime with a CONFIG_EXTENSION or STATISTIC_EXTENSION.
```
ocket8888: I think maybe this is a good use-case for `.. versionchanged::`

docs/source/api/to_extensions.rst
```diff
@@ -130,13 +132,13 @@ Request Structure
 
 	.. note:: This field has meaning only for "Check Extensions"
 
-:type:    The :term:`Type` of extension - there are a set number of allowed values which are not recorded anywhere at the time of this writing
+:type:    The :term:`Type` of extension. It must be a CHECK_EXTENSION type with the naming convention of ``CHECK_EXTENSION_*``.
```
ocket8888: ... and I think it's best to put things like that as a block-quote under the actual things that they reference. So:
```rst
:type: sample text

	.. versionchanged:: ATCv4 lorem ipsum dolor...

:version: text for your header
```

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -701,6 +701,18 @@ func UsernameExists(uname string, tx *sql.Tx) (bool, error) {
 	return exists, err
 }
 
+// GetTypeIDByName reports whether or not a type exists with the given name and the id of that type.
```
ocket8888: nit: the ordering of this `func`'s returns is different in the GoDoc than in the actual code.

traffic_ops/traffic_ops_golang/toextension/toextension.go
```diff
@@ -0,0 +1,154 @@
...
+	// Validate request body
+	if err := api.Parse(r.Body, inf.Tx.Tx, &toExt); err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, err, nil)
+		return
+	}
+
+	// Get Type ID
+	typeID, exists, err := dbhelpers.GetTypeIDByName(*toExt.Type, inf.Tx.Tx)
+	if !exists {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, fmt.Errorf("expected type %v does not exist in type table", *toExt.Type))
```
ocket8888: This seems more like a client error TBH - is this a 500 for backward compatibility or something? Cuz we don't promise things like that won't change at random. If it were up to me I'd make the return code `409 Conflict`.

mhoppa: Perl did not check this all, it should be 4xx error youre right Ill change this

ocket8888: That sounds like Perl

docs/source/api/to_extensions.rst
```diff
@@ -103,6 +103,9 @@ Response Structure
 
 ``POST``
 ========
+.. versionchanged:: ATCv4
```
ocket8888: This only applies to the 1.5 handler, right? I think here and below it should name the API version in which it changed, because if you request an older version you will still get the old behavior, even in ATCv4.

PR #4332, opened 2020-01-24
Added deprecation notice to /divisions/name/{{name}}

### LINKED ISSUE #3789: Rewrite /divisions/name/{{name}} to Go
## I'm submitting a ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/divisions/name/{{name}}` is handled by Perl

## Expected / new behavior:
`/divisions/name/{{name}}` should be handled by Go.

### Minimum API Version
1.1

### Request Methods
- [x] `DELETE`
- [x] `GET` (#2341)

N/A

ocket8888:

- [x] This PR fixes #3789

Adds a deprecation notice to API responses for `/divisions/name/{{name}}` handlers, and adds one to the GoDoc for the client method that uses that route.

Not documenting deprecated route.

- Traffic Ops

Make requests to `/divisions/name/{{name}}` and observe that the deprecation notice always appears. It supports `GET` and `DELETE`, which both do pretty much what you'd expect.

- [x] Tests are unnecessary

### DIFF
--- a/traffic_ops/traffic_ops_golang/api/shared_handlers.go
+++ b/traffic_ops/traffic_ops_golang/api/shared_handlers.go
@@ -144,6 +144,65 @@ func ReadHandler(reader Reader) http.HandlerFunc {
 	}
 }
 
+// DeprecatedReadHandler creates a net/http.HandlerFunc for the passed Reader object, and adds a deprecation
+// notice, optionally with a passed alternative route suggestion.
+func DeprecatedReadHandler(reader Reader, alternative *string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var alerts tc.Alerts
+		if alternative != nil {
+			alerts = tc.CreateAlerts(tc.WarnLevel, fmt.Sprintf("This endpoint is deprecated, please use %s instead", *alternative))
+		} else {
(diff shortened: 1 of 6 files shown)


traffic_ops/traffic_ops_golang/api/shared_handlers.go
```diff
@@ -144,6 +144,65 @@ func ReadHandler(reader Reader) http.HandlerFunc {
...
+				log.Errorf(r.RemoteAddr + " " + sysErr.Error())
+			}
+			if userErr == nil {
+				userErr = errors.New(http.StatusText(errCode))
+			}
+			alerts.AddAlerts(tc.CreateErrorAlerts(userErr))
+			WriteAlerts(w, r, errCode, alerts)
+			log.Debugln(userErr.Error())
+			*r = *r.WithContext(context.WithValue(r.Context(), tc.StatusKey, errCode))
+			return
```
mhoppa: this is the same code as above lines 160-170 from what I can tell, can we collapse that into another utility func

ocket8888: It's also very similar to `handleSimpleErr` So I think the best thing would be to abstract it out of there as well.

PR #4343, opened 2020-01-27
traffic_ops_ort: fix for syncds having too many header rewrite false …

traeak:
…positives

What does this PR (Pull Request) do?
This PR fixes an issue when canned comments aren't consistently stripped for header_rewrites, logs_xml.config or *.cer files.  Also promotes File changed message from DEBUG to ERROR for report mode.

- Traffic Ops ORT perl

Set up a test box with traffic_ops_ort.  Badass an initial EDGE configuration with some DS's that contain EDGE header rewrite rules.

First test:  Queue CDN.  Run syncds, WARN mode.  Old version the header_rewrite header file will trigger false positives. New version should not.  Edit a DS's header rewrite rule, queue, run syncds.  Only the edited header rewrite rule should update.

Next test: Edit header_rewrite for DS and queue.  Run report mode with WARN mode.  Ensure the changed accompanying header rewrite and only that one shows up as an ERROR message.

- master (8e11e3db7f4e708e6ba42e7e766f9bb10387d750)
- 4.0.x (1523be3d2238243f0f2ca3ab9d3d93c69b9ae3c9)

### DIFF
--- a/traffic_ops/ort/traffic_ops_ort.pl
+++ b/traffic_ops/ort/traffic_ops_ort.pl
@@ -246,8 +246,6 @@ else {
 
 ( my $my_profile_name, $cfg_file_tracker, my $my_cdn_name ) = &get_cfg_file_list( $hostname_short, $traffic_ops_host, $script_mode );
 
-
-
 if ( $script_mode == $REVALIDATE ) {
 	( $log_level >> $INFO ) && print "\nINFO: ======== Revalidating, no package processing needed ========\n";
 }
@@ -410,6 +408,7 @@ sub process_cfg_file {
 	}
 
 	my @db_file_lines = @{ &scrape_unencode_text($result) };
+	@db_file_lines = @{ &scrape_canned_comments(\@db_file_lines) };
 
 	my $file = $config_dir . "/" . $cfg_file;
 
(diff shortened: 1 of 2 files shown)


traffic_ops/ort/traffic_ops_ort.pl
```diff
@@ -2652,6 +2658,23 @@ sub scrape_unencode_text {
...
 }
 
+sub scrape_canned_comments {
+	my $linesin = $_[0];
+
+	my @linesout;
+
+	foreach my $line (@$linesin) {
+		if ( $line =~ m/^\#/ ) {
+			if ( $line =~ m/DO NOT EDIT - Generated for / || $line =~ m/$header_comment/ || $line =~ m/TRAFFIC OPS NOTE\:/ || $line =~ m/^##OVERRID.*##/ ) {
```
rob05c: Question: why don't we skip all comments?

traeak: This function applies not just to ats config files but also to other type files that may not use '#' as a comment convention.  This is part of the hackery that is ORT.  Honestly I'm a bit scared about increasing scope of the changes.  You'll notice code that's left over is now checking for things that have already been stripped.  How much refactoring should be done?

rob05c: Ah, got it. Agreed.

PR #4344, opened 2020-01-27
Fix TO atscfg missing max origin conns

rob05c:
Fixes a misordered SQL scan, which caused the Server Status to be
missing, causing the config gen to think there were no online servers,
and thus not create the max_origin_connections entry.

Note this is only in the old TO endpoints, not the config generator. The `atstccfg` generator doesn't have this issue.

Includes TO API (Integration) tests.
No docs, no interface change.

- Traffic Ops

Run TO API Tests.
Set Max Origin Connections on a DS, request the header rewrite endpoint for that DS, verify it includes a max_origin_connections entry.

- master 
- 4.0.0 RC2

## Additional Information


rob05c:
@mitchell852 Ah, the Mid query had the same bug. I fixed that too, both bugs should be fixed now.

mitchell852:
@rob05c - reminder to backport this to 4.0.x branch

PR #4345, opened 2020-01-28
Substitute tab for leading 4 spaces in documentation

zrhoffman:

This PR searches for instances of 4 spaces that are either at the beginning of the line or prefixed only by tabs and replaces each instance with a tab.

Tests are not necessary because only documentation is changed.

- Documentation

Attempting to replace leading 4 spaces should not change any reStructuredText files:

```shell
find docs/source -type f -name '*.rst' -exec sed -i 's/^\($'\t'*\)    /\1$'\t'/g' {} ';'
```

Also, try to build the docs.

- master (d8f46315cd)
- 3.1.0
- 4.0.0 (RC2)


docs/source/admin/quick_howto/anonymous_blocking.rst
```diff
@@ -33,9 +33,9 @@ Configure Anonymous Blocking
 			"name": "Anonymous IP Blocking Policy",
 
 			"anonymousIp": { "blockAnonymousVPN": true,
-			                 "blockHostingProvider": true,
-			                 "blockPublicProxy": true,
-			                 "blockTorExitNode": true},
+				"blockHostingProvider": true,
+				"blockPublicProxy": true,
+				"blockTorExitNode": true},
```
ocket8888: This was actually alignment, not indentation. The intent is to line up all the text - and that's totally valid because the example block is properly indented so the contents are not interpreted by the RST compiler at all (they are lex'd by Pygments but this is lexically and syntactically correct JSON so that doesn't matter either).

It is more consistent to indent these rather than align them, but if you're gonna do that I'd ask that you also indent the top property (`"blockAnonymousVPN"`) so that everything still lines up, and also check the output to be sure it isn't being expanded wider than a typical browser window (because Sphinx auto-expands tabs in code examples to GNU-standard 8 spaces and I've already tried everything I can think of to fix that).

docs/source/api/osversions.rst
```diff
@@ -66,17 +66,17 @@ The **directory** of the file can be changed by creating a specific :term:`Param
...
 
 .. code-block:: json
-        :caption: Example osversions.json file
+	:caption: Example osversions.json file
 
-        {
-          "CentOS 7.2": "centos72"
-        }
+	{
+			"CentOS 7.2": "centos72"
```
ocket8888: Looks like this got indented one too many times.

zrhoffman: Adjusted in 0220c7acba.

docs/source/admin/quick_howto/anonymous_blocking.rst
zrhoffman: Reverted in 0220c7acba.

PR #4347, opened 2020-01-28
Fix internal server error in user current put

### LINKED ISSUE #4241: Internal Server Error when updating user's password using /user/current endpoint
## I'm submitting a ...

- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When making a `PUT` request to `/user/current` with the following request body:
```
{
    "user": {
        "localPasswd": "newPassword"
    }
}
```
The response returned is:
```
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```
## Expected / new behavior:

Should either allow to change password or block the request.
## Anything else:

Error log: 
`Updating user: resetting password: pq: syntax error at or near "WHERE"`

ocket8888:
- [x] This PR fixes #4241 

Fixes an error in a SQL query that was causing internal server errors whenever a user's password was updated using a PUT request to `/user/current`. Also updates docs which had an invalid value for `fullName` in the request example.

- Documentation
- Traffic Ops

Besides building and reading the documentation, make a PUT request to `/user/current` with a `localPasswd` key and verify that your password actually changes, and no internal server error occurs. I didn't write a test because I don't fully comprehend the implications of changing a shared user's password partway through testing.

- master
- 4.0 (RC2 - I *think*)


rawlinp:
@ocket8888 I'm wondering why this is considered minor -- does this mean without being fixed, nobody would be able to update their password?

ocket8888:
No, there are other methods to update a password. `/user/current/update` is basically the same endpoint, and anyone with a registered email can reset their password. Plus, if the `/user/current` PUT handler was heavily used enough it could just be perlBypassed as of right now.

I can change the severity label, though, I don't have a preference as to what it is.

rawlinp:
Oh, thanks for the clarification -- I forgot we had about 18 endpoints to do the same thing. Minor seems fair then.

mitchell852:
@ocket8888 - this looks like a good backport candidate for the 4.0.x branch

rawlinp:
This bug does not actually affect the 4.0.x branch, since it appears this rewrite was merged _after_ the branch was cut.

PR #4348, opened 2020-01-28
Deprecate GET jobs/{id}

mhoppa:

The route was already rewritten but since the functionality is duplicated on the /jobs endpoint this route can be deprecated 

- Documentation
- Traffic Ops

Build Traffic ops with change and make request to `GET /jobs/{id}`. Result response should be the same with a warning deprecation notice in the alerts section pointing the user to the `/jobs` endpoint.

### DIFF
--- a/docs/source/api/jobs_id.rst
+++ b/docs/source/api/jobs_id.rst
@@ -18,6 +18,7 @@
 ***************
 ``jobs/{{ID}}``
 ***************
+.. deprecated:: ATCv4
 .. caution:: In the vast majority of cases, it is preferred to use the ``id`` query parameter of the :ref:`to-api-jobs` endpoint instead.
 
 ``GET``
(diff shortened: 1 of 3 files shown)


mhoppa:
retest this please

docs/source/api/jobs_id.rst
```diff
@@ -18,6 +18,7 @@
 ***************
 ``jobs/{{ID}}``
 ***************
+.. deprecated:: ATCv4
 .. caution:: In the vast majority of cases, it is preferred to use the ``id`` query parameter of the :ref:`to-api-jobs` endpoint instead.
```
ocket8888: You can collapse these, they're not really different admonitions. Use other thing because this is deprecated.

mhoppa: done

PR #4355, opened 2020-01-30
Fix API tests for header rewrite

mhoppa:

Run API tests


mhoppa:
@rob05c unsure if that was the best way to fix it but API tests are passing now let me know if you think it should be handled in other ways

mhoppa:
Thanks @rob05c for the offline help for the better solution :) and FWIW they were not broken it was an intermittent bug dependent on what DS was received for the test

PR #4357, opened 2020-01-30
Deprecate /regions/name/{{name}}

### LINKED ISSUE #3819: Rewrite /regions/name/{{name}} to Go
## I'm submitting an ...

- Documentation (possibly)
- Traffic Ops

## Current behavior:
`/regions/name/{{name}}` is handled by Perl

## Expected / new behavior:
`/regions/name/{{name}}` should be handled by Go.

### Minimum API Versions
1.1

### Request Methods
- [x] `GET` (#2353)
- [ ] `DELETE`

N/A

zrhoffman:

- [x] This PR fixes #3819 

This PR deprecates the following endpoints:

* GET `regions/name/{{name}}`
* DELETE `regions/name/{{name}}`

and adds DELETE `regions?name={{name}}`.

This PR also sorts the changelog's list of deprecated/removed endpoints for easier reading.

- Traffic Control Client (Go)
- Documentation
- Traffic Ops

* Run the Traffic Ops API tests
* Build the documentation

- master (1dddf38530)
- 4.0.0 (RC3)

- [x] This PR includes tests
- [x] This PR PR does not include a database migration

### DIFF
--- a/docs/source/api/regions.rst
+++ b/docs/source/api/regions.rst
@@ -169,6 +169,8 @@ Response Structure
 ``DELETE``
 ==========
 
+.. versionadded:: 1.5
+
 Deletes a region. If no query parameter is specified, a ``400 Bad Request`` response is returned.
 
 :Auth. Required: Yes
--- a/docs/source/api/regions.rst
+++ b/docs/source/api/regions.rst
@@ -165,3 +165,61 @@ Response Structure
 		"lastUpdated": "2018-12-06 02:14:45+00",
 		"name": "Manchester"
 	}}
+
+``DELETE``
+==========
+
+Deletes a region. If no query parameter is specified, a ``400 Bad Request`` response is returned.
+
+:Auth. Required: Yes
+:Roles Required: "admin" or "operations"
+:Response Type:  Object
+
+Request Structure
+-----------------
+
+.. table:: Request Query Parameters
+
+	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
+	| Name      | Required | Description                                                                                                   |
+	+===========+==========+===============================================================================================================+
+	| id        | no       | Delete :term:`Region` by integral, unique identifier                                                          |
+	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
+	| name      | no       | Delete :term:`Region` by name                                                                                 |
+	+-----------+----------+---------------------------------------------------------------------------------------------------------------+
+
+.. code-block:: http
+	:caption: Request Example
+
+	DELETE /api/1.5/regions?name=Manchester HTTP/1.1
+	User-Agent: curl/7.29.0
+	Host: trafficops.infra.ciab.test
+	Accept: */*
+	Cookie: mojolicious=...
+
+Response Structure
+------------------
+
+.. code-block:: http
+	:caption: Response Example
+
+	HTTP/1.1 200 OK
+	Access-Control-Allow-Credentials: true
+	Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept, Set-Cookie, Cookie
+	Access-Control-Allow-Methods: POST,GET,OPTIONS,PUT,DELETE
+	Access-Control-Allow-Origin: *
+	Content-Type: application/json
+	Set-Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1ODEwODM3ODQsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--f6b59c9dde67bcd545bde3bbb4e56b33e23827e9; Path=/; Expires=Fri, 07 Feb 2020 13:56:24 GMT; Max-Age=3600; HttpOnly
+	Whole-Content-Sha512: yNqXKcoiohEbJrEkH8LD1zifh87dIusuvUqgQnYueyKqCXkfd5bQvQ0OhQ2AAdSZa/oe2SAqMjojGsUlxHtIQw==
+	X-Server-Name: traffic_ops_golang/
+	Date: Fri, 07 Feb 2020 12:56:24 GMT
+	Content-Length: 62
+
+	{
+		"alerts": [
+			{
+				"text": "region was deleted.",
+				"level": "success"
+			}
+		]
+	}
--- a/docs/source/api/regions.rst
+++ b/docs/source/api/regions.rst
@@ -211,7 +211,7 @@ Response Structure
 	Access-Control-Allow-Methods: POST,GET,OPTIONS,PUT,DELETE
 	Access-Control-Allow-Origin: *
 	Content-Type: application/json
-	Set-Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1ODEwODM3ODQsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--f6b59c9dde67bcd545bde3bbb4e56b33e23827e9; Path=/; Expires=Fri, 07 Feb 2020 13:56:24 GMT; Max-Age=3600; HttpOnly
+	Set-Cookie: mojolicious=...; Path=/; Expires=Fri, 07 Feb 2020 13:56:24 GMT; Max-Age=3600; HttpOnly
 	Whole-Content-Sha512: yNqXKcoiohEbJrEkH8LD1zifh87dIusuvUqgQnYueyKqCXkfd5bQvQ0OhQ2AAdSZa/oe2SAqMjojGsUlxHtIQw==
 	X-Server-Name: traffic_ops_golang/
 	Date: Fri, 07 Feb 2020 12:56:24 GMT
--- a/traffic_ops/app/lib/API/Region.pm
+++ b/traffic_ops/app/lib/API/Region.pm
@@ -220,7 +220,7 @@ sub create_for_division {
 	return $self->with_deprecation("create region " . $params->{name} . " failed.", "error", 400, $alt);
 }
 
-sub delete_by_id {
+sub delete {
 	my $self = shift;
 	my $id     = $self->param('id');
 
--- a/traffic_ops/app/lib/API/Region.pm
+++ b/traffic_ops/app/lib/API/Region.pm
@@ -247,7 +247,7 @@ sub delete_by_name {
 	my $alt		 = 'DELETE /regions?name={{name}}';
 
 	if ( !&is_oper($self) ) {
-		return $self->forbidden();
+		return $self->with_deprecation("Forbidden", "error", 403, $alt)
 	}
 
 	my $region = $self->db->resultset('Region')->find( { name => $name } );
--- a/traffic_ops/app/lib/API/Region.pm
+++ b/traffic_ops/app/lib/API/Region.pm
@@ -220,7 +220,7 @@ sub create_for_division {
 	return $self->with_deprecation("create region " . $params->{name} . " failed.", "error", 400, $alt);
 }
 
-sub delete {
+sub delete_by_id {
 	my $self = shift;
 	my $id     = $self->param('id');
 
@@ -244,6 +244,7 @@ sub delete {
 sub delete_by_name {
 	my $self = shift;
 	my $name     = $self->param('name');
+	my $alt		 = 'DELETE /regions?name={{name}}';
 
 	if ( !&is_oper($self) ) {
 		return $self->forbidden();
@@ -251,14 +252,14 @@ sub delete_by_name {
 
 	my $region = $self->db->resultset('Region')->find( { name => $name } );
 	if ( !defined($region) ) {
-		return $self->not_found();
+		return $self->with_deprecation("Resource not found.", "error", 404, $alt);
 	}
 
 	my $rs = $region->delete();
 	if ($rs) {
-		return $self->success_message("Region deleted.");
+		return $self->with_deprecation("Region deleted.", "success", 200, $alt);
 	} else {
-		return $self->alert( "Region delete failed." );
+		return $self->with_deprecation( "Region delete failed.", "error", 400, $alt);
 	}
 }
 
--- a/traffic_ops/app/lib/API/Region.pm
+++ b/traffic_ops/app/lib/API/Region.pm
@@ -69,7 +69,7 @@ sub index_by_name {
 			}
 		);
 	}
-	$self->success( \@data );
+	$self->deprecation(200, 'GET /regions?name={{name}}', \@data);
 }
 
 sub show {
--- a/traffic_ops/app/lib/TrafficOpsRoutes.pm
+++ b/traffic_ops/app/lib/TrafficOpsRoutes.pm
@@ -745,7 +745,7 @@ sub api_routes {
 	$r->put("/api/$version/regions/:id" => [ id => qr/\d+/ ] )->over( authenticated => 1, not_ldap => 1 )->to( 'Region#update', namespace => $namespace );
 	$r->post("/api/$version/regions")->over( authenticated => 1, not_ldap => 1 )->to( 'Region#create', namespace => $namespace );
 	$r->post("/api/$version/divisions/:division_name/regions")->over( authenticated => 1, not_ldap => 1 )->to( 'Region#create_for_division', namespace => $namespace );
-	$r->delete("/api/$version/regions/:id" => [ id => qr/\d+/ ] )->over( authenticated => 1, not_ldap => 1 )->to( 'Region#delete', namespace => $namespace );
+	$r->delete("/api/$version/regions/:id" => [ id => qr/\d+/ ] )->over( authenticated => 1, not_ldap => 1 )->to( 'Region#delete_by_id', namespace => $namespace );
 	$r->delete("/api/$version/regions/name/:name")->over( authenticated => 1, not_ldap => 1 )->to( 'Region#delete_by_name', namespace => $namespace );
 
 	# -- ROLES
--- a/traffic_ops/app/lib/TrafficOpsRoutes.pm
+++ b/traffic_ops/app/lib/TrafficOpsRoutes.pm
@@ -745,7 +745,7 @@ sub api_routes {
 	$r->put("/api/$version/regions/:id" => [ id => qr/\d+/ ] )->over( authenticated => 1, not_ldap => 1 )->to( 'Region#update', namespace => $namespace );
 	$r->post("/api/$version/regions")->over( authenticated => 1, not_ldap => 1 )->to( 'Region#create', namespace => $namespace );
 	$r->post("/api/$version/divisions/:division_name/regions")->over( authenticated => 1, not_ldap => 1 )->to( 'Region#create_for_division', namespace => $namespace );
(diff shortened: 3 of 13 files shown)


docs/source/api/regions_name_name.rst
```diff
@@ -0,0 +1,153 @@
+..
```
mhoppa: if an endpoint is being deprecated we shouldnt/dont need to add documentation around it

zrhoffman: Okay, rebased and documentation removed in [`f22b0e9c3d~`...`c954e8776b`](https://github.com/apache/trafficcontrol/compare/f22b0e9c3d~...c954e8776b).

ocket8888: Yeah, there's no point in doing that since we don't want people to use them. It's a waste of time, but providing the docs are well-formed and accurate I wouldn't hold up a PR because of adding docs, IMO. Reviewer's discretion, I suppose.

mhoppa:
@ocket8888 thoughts ^

ocket8888:
The way I used to do this was with a mentality of "everything gets rewritten to Go, and *then* refactored." So I'd have rewritten it with a Go DELETE handler for `/regions/name/{{name}}` **and** `/regions?name={{name}}` and a deprecation notice wrapper around the former.

I guess the compromise here being: add a 1.5 DELETE handler for `/regions?name={{name}}` and deprecate the original route (`/regions/name/{{name}}`). You can do that by rewriting to Go or not; since you're already making a handler it'd only entail making a wrapper. That's my suggestion, anyway.

I'm +1 on deprecating them, fwiw.

mhoppa:
okay I am fine either way as along as there is a path to delete regions by name 🤷‍♂

zrhoffman:
Got it, I will write a  DELETE handler for `/regions/name/{{name}}` but keep it deprecated and add a DELETE handler for `/regions?name={{name}}`.

Title changed to WIP for now.

zrhoffman:
Okay, now we have Go handlers for both `/regions/name/{{name}}` (proxyHandler, 6e63cde165), which is deprecated, and `/regions?name={{name}}` (d3353a9b46).

traffic_ops/client/region.go
```diff
@@ -159,3 +160,28 @@ func (to *Session) GetRegionByNamePath(name string) ([]tc.RegionName, ReqInf, er
...
+// DELETE a Region
+func (to *Session) DeleteRegionByName(id *int, name *string) (tc.Alerts, ReqInf, error) {
+	v := url.Values{}
+	if id != nil {
+		v.Add("id", strconv.Itoa(*id))
+	}
+	if name != nil {
+		v.Add("name", *name)
+	}
+	URI := API_v13_REGIONS
```
mhoppa: could we use ```apiBase + "/regions"``` instead?

docs/source/api/regions.rst
```diff
@@ -165,3 +165,61 @@ Response Structure
 		"lastUpdated": "2018-12-06 02:14:45+00",
 		"name": "Manchester"
 	}}
+
+``DELETE``
+==========
+
+Deletes a region. If no query parameter is specified, a ``400 Bad Request`` response is returned.
```
mhoppa: we should have this API marked as version added 1.5

traffic_ops/client/region.go
```diff
@@ -159,3 +160,28 @@ func (to *Session) GetRegionByNamePath(name string) ([]tc.RegionName, ReqInf, er
 	}
 	return resp.Response, reqInf, nil
 }
+
+// DELETE a Region
+func (to *Session) DeleteRegionByName(id *int, name *string) (tc.Alerts, ReqInf, error) {
```
mhoppa: godoc is off from name of func and nit but I wonder if this should be called DeleteRegion as you can delete it by id or name

traffic_ops/traffic_ops_golang/region/name.go
```diff
@@ -35,7 +35,13 @@ func GetName(w http.ResponseWriter, r *http.Request) {
...
 	}
 	defer inf.Close()
-	api.RespWriter(w, r, inf.Tx.Tx)(getName(inf.Tx.Tx, inf.Params["name"]))
+
+	regionNames, err := getName(inf.Tx.Tx, inf.Params["name"])
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("getting region: "+err.Error()))
+		return
+	}
+	api.WriteAlertsObj(w, r, http.StatusOK, tc.Alerts{Alerts: []tc.Alert{api.DeprecationWarning("GET /regions?name={{name}}")}}, regionNames)
```
mhoppa: this doesnt have deprecation notices on the errors

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -301,6 +301,8 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `regions/name/{name}/?(\.json)?$`, region.GetName, auth.PrivLevelReadOnly, Authenticated, nil, 503583197, noPerlBypass},
 		{1.1, http.MethodPut, `regions/{id}$`, api.UpdateHandler(&region.TORegion{}), auth.PrivLevelOperations, Authenticated, nil, 226308224, noPerlBypass},
 		{1.1, http.MethodPost, `regions/?$`, api.CreateHandler(&region.TORegion{}), auth.PrivLevelOperations, Authenticated, nil, 1288334488, noPerlBypass},
+		{1.1, http.MethodDelete, `regions/?$`, api.DeleteHandler(&region.TORegion{}), auth.PrivLevelOperations, Authenticated, nil, 2032626758, noPerlBypass},
```
mhoppa: this should be version 1.5

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -156,6 +166,37 @@ func GenericUpdate(val GenericUpdater) (error, error, int) {
...
+	where, _, _, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(val.APIInfo().Params, val.DeleteKeyOptions())
+	if len(errs) > 0 {
+		return util.JoinErrs(errs), nil, http.StatusBadRequest
+	}
+
+	query := val.DeleteQueryBase() + where
+	tx := val.APIInfo().Tx
+	result, err := tx.NamedExec(query, queryValues)
+	if err != nil {
+		return nil, errors.New("deleting " + val.GetType() + ": " + err.Error()), http.StatusInternalServerError
```
mhoppa: should this be ParseDBError? that parses some postgres errors into non 500 errors

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -156,6 +166,37 @@ func GenericUpdate(val GenericUpdater) (error, error, int) {
...
+	}
+
+	if rowsAffected, err := result.RowsAffected(); err != nil {
+		return nil, errors.New("deleting " + val.GetType() + ": getting rows affected: " + err.Error()), http.StatusInternalServerError
+	} else if rowsAffected < 1 {
+		return errors.New("no " + val.GetType() + " with that key found"), nil, http.StatusNotFound
+	} else if rowsAffected > 1 {
+		err := tx.Rollback()
```
mhoppa: this should be handled by the HandleError -> https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/api/api.go#L131

traffic_ops/traffic_ops_golang/api/shared_handlers.go
```diff
@@ -298,23 +298,36 @@ func DeleteHandler(deleter Deleter) http.HandlerFunc {
...
-				return
+		deleteKeyOptionExists := false
+		if d, ok := obj.(HasDeleteKeyOptions); ok {
+			options := d.DeleteKeyOptions()
+			for key, _ := range options {
+				if inf.Params[key] != "" {
+					deleteKeyOptionExists = true
+					break
+				}
 			}
```
mhoppa: should this error as they said it has delete key options but none was specified in the params?

zrhoffman: Having that error makes sense to me.

traffic_ops/app/lib/API/Region.pm
```diff
@@ -244,21 +244,22 @@ sub delete {
 sub delete_by_name {
 	my $self = shift;
 	my $name     = $self->param('name');
+	my $alt		 = 'DELETE /regions/{{ID}}';
 
 	if ( !&is_oper($self) ) {
 		return $self->forbidden();
```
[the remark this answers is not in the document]
zrhoffman: Add a deprecation notice in 2ada94e8f2.

traffic_ops/client/region.go
zrhoffman: Changed to use `apiBase` in 2ada94e8f2.

docs/source/api/regions.rst
zrhoffman: Marked in 2ada94e8f2.

traffic_ops/client/region.go
zrhoffman: Fixed GoDoc and renamed in 2ada94e8f2.

traffic_ops/traffic_ops_golang/region/name.go
zrhoffman: Add deprecation notices to the errors in 2ada94e8f2.

traffic_ops/traffic_ops_golang/routing/routes.go
zrhoffman: Changed to 1.5 in 2ada94e8f2.

traffic_ops/traffic_ops_golang/api/generic_crud.go
zrhoffman: Changed to ParseDBError in 2ada94e8f2.

traffic_ops/traffic_ops_golang/api/generic_crud.go
zrhoffman: Right you are, removed in 2ada94e8f2.

traffic_ops/traffic_ops_golang/api/shared_handlers.go
zrhoffman: Made specifying a delete key in 2ada94e8f2.

docs/source/api/regions.rst
```diff
@@ -165,3 +165,63 @@ Response Structure
...
+.. code-block:: http
+	:caption: Response Example
+
+	HTTP/1.1 200 OK
+	Access-Control-Allow-Credentials: true
+	Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept, Set-Cookie, Cookie
+	Access-Control-Allow-Methods: POST,GET,OPTIONS,PUT,DELETE
+	Access-Control-Allow-Origin: *
+	Content-Type: application/json
+	Set-Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1ODEwODM3ODQsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--f6b59c9dde67bcd545bde3bbb4e56b33e23827e9; Path=/; Expires=Fri, 07 Feb 2020 13:56:24 GMT; Max-Age=3600; HttpOnly
```
ocket8888: API examples shouldn't include actual `mojolicious` cookies, as that may or may not (though hopefully not) constitute a security risk. You can replace the value with `...` like you did above in the request example, or just omit the header entirely (I'd prefer the former, but the latter is allowed).

traffic_ops/app/lib/TrafficOpsRoutes.pm
```diff
@@ -745,7 +745,7 @@ sub api_routes {
 	$r->put("/api/$version/regions/:id" => [ id => qr/\d+/ ] )->over( authenticated => 1, not_ldap => 1 )->to( 'Region#update', namespace => $namespace );
 	$r->post("/api/$version/regions")->over( authenticated => 1, not_ldap => 1 )->to( 'Region#create', namespace => $namespace );
 	$r->post("/api/$version/divisions/:division_name/regions")->over( authenticated => 1, not_ldap => 1 )->to( 'Region#create_for_division', namespace => $namespace );
-	$r->delete("/api/$version/regions/:id" => [ id => qr/\d+/ ] )->over( authenticated => 1, not_ldap => 1 )->to( 'Region#delete', namespace => $namespace );
+	$r->delete("/api/$version/regions/:id" => [ id => qr/\d+/ ] )->over( authenticated => 1, not_ldap => 1 )->to( 'Region#delete_by_id', namespace => $namespace );
 	$r->delete("/api/$version/regions/name/:name")->over( authenticated => 1, not_ldap => 1 )->to( 'Region#delete_by_name', namespace => $namespace );
```
ocket8888: I think maybe renaming the sub in Region.pm like you did will break this handler. I haven't tested yet, but it seems likely.

traffic_ops/client/region.go
```diff
@@ -159,3 +160,28 @@ func (to *Session) GetRegionByNamePath(name string) ([]tc.RegionName, ReqInf, er
 	}
 	return resp.Response, reqInf, nil
 }
+
+// DeleteRegion lets you DELETE a Region
```
ocket8888: nit: should end with a period.

Should also mention that only one of the parameters is required.

traffic_ops/client/region.go
```diff
@@ -159,3 +160,28 @@ func (to *Session) GetRegionByNamePath(name string) ([]tc.RegionName, ReqInf, er
...
+
+	resp, remoteAddr, err := to.request(http.MethodDelete, URI, nil)
+	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
+	if err != nil {
+		return tc.Alerts{}, reqInf, err
+	}
+	defer resp.Body.Close()
+	var alerts tc.Alerts
+	err = json.NewDecoder(resp.Body).Decode(&alerts)
+	return alerts, reqInf, nil
```
ocket8888: You seem to be ignoring the parse error on the line above; should `return alerts, reqInf, err` instead of `return alerts, reqInf, nil`

traffic_ops/traffic_ops_golang/region/regions.go
```diff
@@ -44,8 +44,13 @@ func (v *TORegion) ParamColumns() map[string]dbhelpers.WhereColumnInfo {
 		"id":       dbhelpers.WhereColumnInfo{"r.id", api.IsInt},
 	}
 }
-func (v *TORegion) UpdateQuery() string { return updateQuery() }
-func (v *TORegion) DeleteQuery() string { return deleteQuery() }
+func (v *TORegion) UpdateQuery() string     { return updateQuery() }
+
+// DeleteQuery returns a query, including a WHERE clause
```
ocket8888: nit: should end in a period

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -301,6 +301,8 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{1.1, http.MethodGet, `regions/name/{name}/?(\.json)?$`, region.GetName, auth.PrivLevelReadOnly, Authenticated, nil, 503583197, noPerlBypass},
 		{1.1, http.MethodPut, `regions/{id}$`, api.UpdateHandler(&region.TORegion{}), auth.PrivLevelOperations, Authenticated, nil, 226308224, noPerlBypass},
 		{1.1, http.MethodPost, `regions/?$`, api.CreateHandler(&region.TORegion{}), auth.PrivLevelOperations, Authenticated, nil, 1288334488, noPerlBypass},
+		{1.5, http.MethodDelete, `regions/?$`, api.DeleteHandler(&region.TORegion{}), auth.PrivLevelOperations, Authenticated, nil, 2032626758, noPerlBypass},
+		{1.1, http.MethodDelete, `regions/name/{name}$`, handlerToFunc(proxyHandler), auth.PrivLevelReadOnly, Authenticated, nil, 1925881096, noPerlBypass},
```
ocket8888: I think normally routes that use the `proxyHandler` have `NoAuth` because the assumption is that the Perl server will handle authentication.

docs/source/api/regions.rst
zrhoffman: Ack, I missed one. Fixed in 820659e14.

traffic_ops/app/lib/TrafficOpsRoutes.pm
zrhoffman: I can't reproduce the behavior that motivated me to change it. Reverted in 2edc8f445.

traffic_ops/client/region.go
zrhoffman: Added a period and explanation in 6e1617fae.

traffic_ops/client/region.go
zrhoffman: Returning the error now in 820659e14.

traffic_ops/traffic_ops_golang/region/regions.go
zrhoffman: Period added in 6e1617fae.

traffic_ops/traffic_ops_golang/routing/routes.go
zrhoffman: That makes sense, changed in c947d23d9.

rawlinp: Why is the 1.1 route proxied to Perl? Reading the code, it appears the new handler can delete by either `id` or `name`, right?

mhoppa: well this is just an explicit proxy to the deprecated handler in the perl, either way that route is deprecated and the notices were added in the perl implementation I didnt see any harm in having this in here

ocket8888: Yeah I agree, seems like a fine way to ensure the deprecation notice. It saves him from needing to write a wrapper that inspects the request path.

rawlinp: Ok, I guess usually the routes are just left out so be handled by the "route not found handler" which right now is "check TO-Perl". _Explicit_ `proxyHandler` routes are usually only there to handle requests explicitly before they're unintentionally handled by a different route (i.e. GET /cdns/foo being unintentionally handled by GET /cdns because the request matches it). That doesn't seem to be the case here, right?

zrhoffman: It's not, and the new handler can delete by either `id` or `name`. However, if we wanted to handle it in Go we would need to add a way for that handler to receive arbitrary alerts. The only shared handler I see that can already do that is `DeprecatedReadHandler()`, so doing that here would mean making a completely new DELETE handler just for this case.

Of course, it would be easier to just make a `region`-specific handler. In this case, though, the new handler cannot be used for a deprecated route.

zrhoffman: The reason I made it an explicit `proxyHandler` route was to keep `routes.go` useful as a reference for which routes exist. It would be super easy to remove, though.

rawlinp: Ok, well `routes.go` might not even be a true list of all routes that exist currently because I'm not sure we've followed that pattern for other "deprecate but don't rewrite" routes. Not a big deal, just made me confused when I saw it.

PR #4379, opened 2020-02-05
Rewrite isos to Golang

### LINKED ISSUE #3801: Rewrite /isos to Go

mhoppa:

- [x]  This PR fixes #3801 such that the /isos handler has been rewritten in Go

- CDN in a Box
- Documentation
- Traffic Ops

Build TO binary with changes and then generate an iso.

## Additional Information

This PR is based off of #4203 

- The Go version of the endpoint attempts to produce the same behavior as the Perl version, with the following exceptions:
  - The Go endpoint does not support non-streaming downloads. In Perl, non-streaming downloads worked by saving the ISO image to the server's disk and returning to the requesting client a URL that allowed them to download that file at a later time. If a non-streaming request is made to the Go endpoint, it returns an error in the standard JSON response structure.
This was discussed on the dev mailing list under the Proposal: Only allow streaming downloads of ISOs subject. It was decided that this is a breaking API change, and therefore the Go API should be a version 2.0 endpoint.
  - In a number of error cases, the Perl version returns HTML. The Go version returns JSON alerts in these cases, for example when executing the mkisofs command fails.
  - The Perl version, when streaming the response, would fail silently if the subprocess failed (code). The Go version returns an error to the user in this case.
  - When creating the mgmt_network.cfg file, the Perl version does not quote the DEVICE value (code), while the Go version does.
- This PR adds a new Go library (https://github.com/GehirnInc/crypt). Relevant discussion on the mailing list was under message with subject: Proposal to add GehirnInc/crypt library
- The mkisofs package could be added to the TrafficOps Go Docker container as part of ciab.
- It's likely that custom generate scripts will use the implantisomd5 program provided on Centos by the isomd5sum package. The mkisofs package is already a requirement of the TrafficOps RPM, so we should considering making isomd5sum a requirement as well.
- There's a cronjob (./traffic_ops/etc/cron.d/trafops_clean_isos) that removes stale ISO files. This isn't necessary for the Go version of this endpoint. Once the Perl endpoint becomes obsolete, it can likely be deleted. The Go version does no store copies of the ISO images, therefore there's no cleanup necessary.


traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+		// At this point, it's not possible to modify the HTTP response.
+		log.Errorf("error creating changelog entry for ISO creation: %v", err)
+	}
+}
+
+// writeRespErrorAlerts writes to w an Alerts JSON response with
+// an error level. Note: api.WriteRespAlertObj could not be used here
+// because it accepts a single "alert", whereas this response
+// may contain multiple alerts.
+func writeRespErrorAlerts(w http.ResponseWriter, req *http.Request, errMsgs []string) {
```
ocket8888: I think this could be done using [`api.WriteAlerts`](https://godoc.org/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api#WriteAlertsObj).

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+	IP6Address    net.IP          `json:"ip6Address"`
+	IP6Gateway    net.IP          `json:"ip6Gateway"`
+	InterfaceName string          `json:"interfaceName"`
+	InterfaceMTU  util.JSONIntStr `json:"interfaceMtu"`
+	Disk          string          `json:"disk"`
+	MgmtIPAddress net.IP          `json:"mgmtIpAddress"`
+	MgmtIPNetmask net.IP          `json:"mgmtIpNetmask"`
+	MgmtIPGateway net.IP          `json:"mgmtIpGateway"`
+	MgmtInterface string          `json:"mgmtInterface"`
+	Stream        boolStr         `json:"stream"`
```
ocket8888: Does this struct still need this property?

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+		writeRespErrorAlerts(w, req, errMsgs)
+		return
+	}
+
+	// Ensure that the given OSVersionDir is defined in the osversions.json config
+	// file as a valid directory. This directory is later referenced for ISO creation
+	// and therefore must an allowed value.
+	if ok, err := ir.validateOSDir(tx); err != nil {
+		statusCode := http.StatusInternalServerError
+		userErr := errors.New("unable to read osversions configuration")
```
ocket8888: We're not supposed to say anything about the state of the server in responses

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
+	"github.com/jmoiron/sqlx"
+)
+
+// Various directories and filenames related to ISO generation.
+const (
+	cfgDefaultDir   = "/var/www/files"  // Default directory containing config file
+	cfgFilename     = "osversions.json" // The JSON config file containing mapping of OS names to directories
+	cfgFilenamePerl = "osversions.cfg"  // Config file name in the Perl version
```
ocket8888: I feel like these should just be implemented at the configuration parsing step.

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+const (
+	ksFilesParamName       = "kickstart.files.location"
+	ksFilesParamConfigFile = "mkisofs"
+)
+
+// Various HTTP-related values.
+const (
+	httpHeaderContentDisposition = "Content-Disposition"
+	httpHeaderContentType        = "Content-Type"
+	httpHeaderContentDownload    = "application/download"
```
ocket8888: The "Content-Type" header is already a defined constant at `github.com/apache/trafficcontrol/lib/go-rfc.ContentType` - I'd suggest that "Content-Disposition" be similarly defined.

Also, "application/download" isn't a MIME-Type recognized by [IANA's MIME registry](https://www.iana.org/assignments/media-types/media-types.xhtml). This is just binary data, so probably "application/octet-stream" is appropriate. Which I'd suggest be defined alongside `github.com/apache/trafficcontrol/lib/go-rfc.ApplicationJSON` (and/or `github.com/apache/trafficcontrol/lib/go-rfc.MIME_JSON`).

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+	inf, userErr, sysErr, errCode := api.NewInfo(req, nil, nil)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, req, inf.Tx.Tx, errCode, userErr, sysErr)
+		return
+	}
+	defer inf.Close()
+
+	// Decode request body into isoRequest instance.
+	var ir isoRequest
+	if err := json.NewDecoder(req.Body).Decode(&ir); err != nil {
```
ocket8888: Since `isoRequest` already has a validation method, why not make it return an `error` instead of a `string` and accept a `*database/sql.Tx` (which would be unused, I know) and then it could be an [`api.ParseValidator`](https://godoc.org/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api#ParseValidator) and then you could use [`api.Parse`](https://godoc.org/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api#Parse)

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+	isoFilename := fmt.Sprintf("%s-%s.iso", ir.fqdn(), ir.OSVersionDir)
+	// strings.ReplaceAll was added in Go 1.12
+	isoFilename = strings.Replace(isoFilename, "/", "_", -1)
+
+	w.Header().Set(httpHeaderContentDisposition, fmt.Sprintf("attachment; filename=%q", isoFilename))
+	w.Header().Set(httpHeaderContentType, httpHeaderContentDownload)
+
+	if err = genISOCmd.stream(w); err != nil {
+		statusCode := http.StatusInternalServerError
+		userErr := errors.New("unable to generate ISO")
```
ocket8888: Same as above RE exposing server state to clients

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+// an error level. Note: api.WriteRespAlertObj could not be used here
+// because it accepts a single "alert", whereas this response
+// may contain multiple alerts.
+func writeRespErrorAlerts(w http.ResponseWriter, req *http.Request, errMsgs []string) {
+	alerts := tc.CreateAlerts(tc.ErrorLevel, errMsgs...)
+
+	body, err := json.Marshal(alerts)
+	if err != nil {
+		statusCode := http.StatusInternalServerError
+		api.HandleErr(w, req, nil, statusCode, errors.New(http.StatusText(statusCode)), fmt.Errorf("error marshalling JSON: %v", err))
```
ocket8888: Hopefully my earlier suggestion to use `api.WriteAlerts` makes this unnecessary to point out but `errors.New(http.StatusText(statusCode))` is superfluous; that's the default behavior of `api.HandleErr`.

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+// 'isSet' will be true.
+type boolStr struct {
+	isSet bool // false if UnmarshalText is given an unrecognized value
+	v     bool
+}
+
+// UnmarshalText decodes strings representing boolean values.
+// It nevers returns an error to allow for all validation errors
+// to be grouped together.
+func (b *boolStr) UnmarshalText(text []byte) error {
```
ocket8888: Why doesn't this just return an error in the case that the text doesn't represent a string boolean?

traffic_ops/traffic_ops_golang/iso/iso.go
```diff
@@ -0,0 +1,368 @@
...
+		if len(i.IPGateway) == 0 {
+			addErr("ipGateway is required if DHCP is no")
+		}
+	}
+
+	// Use of stream is deprecated it will always stream
+	// regardless if this is set or not.
+	if v, ok := i.Stream.val(); ok && !v {
+		addErr("stream request parameter is deprecated")
+	}
```
ocket8888: No reason to check/allow/parse this, because the handler only services API 2.x

traffic_ops/traffic_ops_golang/iso/iso_test.go
```diff
@@ -0,0 +1,625 @@
...
+			if err = os.MkdirAll(filepath.Join(tmpDir, tc.input.OSVersionDir, ksCfgDir), 0777); err != nil {
+				t.Fatal(err)
+			}
+
+			// Create osversions.json file, which is read by validateOSDir method
+			osVersions, err := json.Marshal(libtc.OSVersionsResponse{"test": tc.input.OSVersionDir})
+			if err != nil {
+				t.Fatal(err)
+			}
+			if err = ioutil.WriteFile(filepath.Join(tmpDir, cfgFilename), osVersions, 0777); err != nil {
```
ocket8888: I really don't like seeing files made with `0777` permissions; I think this really only needs to be `0644` (or even `0600` might work but maybe a bit overkill).

docs/source/api/v1/isos.rst
```diff
@@ -61,7 +61,7 @@ Request Structure
 .. code-block:: http
 	:caption: Request Example
 
-	POST /api/1.3/isos HTTP/1.1
+	POST /api/2.0/isos HTTP/1.1
```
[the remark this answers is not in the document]
mhoppa: duh copy pasta

traffic_ops/traffic_ops_golang/iso/crypt_test.go
```diff
@@ -0,0 +1,97 @@
...
+		t.Run(fmt.Sprintf("%d", i), func(t *testing.T) {
+			// Ensure salt is correct length and contains valid characters.
+
+			got := rndSalt(i)
+			if gotL := len(got); gotL != i {
+				t.Fatalf("rndSalt(%d) = %q (length = %d); expected length = %d", i, got, gotL, i)
+			}
+
+			// Ensure proper characters
+		MATCH:
```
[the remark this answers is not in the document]
mhoppa: let me look into it

traffic_ops/traffic_ops_golang/iso/iso.go
mhoppa: it doesnt I guess our docs point out it isnt supported anymore

traffic_ops/traffic_ops_golang/iso/iso.go
mhoppa: my thought was to catch ppl trying to use it in a legacy manner but our other endpoints do that so yeah removing

traffic_ops/traffic_ops_golang/iso/iso.go
ocket8888: I thought these were read in with the configuration file, for some reason.

ocket8888: Oh, it was, but you just removed that since it's unnecessary for streaming downloads.

mhoppa: 👍

traffic_ops/traffic_ops_golang/iso/iso_test.go
```diff
@@ -0,0 +1,529 @@
...
```
ocket8888: Did `0644`/`0600` not work?

mhoppa: it does! forgot to test and push it up :(

PR #4382, opened 2020-02-06
Replace debugln with testing logger

### LINKED ISSUE #4017: Go client/API integration tests are a bit too verbose
## I'm submitting a ...
- improvement request (usability, performance, tech debt, etc.)

- Traffic Control Client (Go)
- Traffic Ops

## Current behavior:
Tests like `CreateTestType` are outputting verbose debug logging by default, using `github.com/apache/trafficcontrol/lib/go-log.Debugln`, which clutters output.

## Expected / new behavior:
Tests should use [`testing.T.Logf`](https://godoc.org/testing#T.Logf) to output verbose debug output, so that it can be turned on and off with the `-v` flag of `go test`.

shamrickus:

- [x] This PR fixes #4017 

- Traffic Ops

Run the TO tests and verify that information is not output by default, but it is when specifying -v with go test.

### DIFF
--- a/traffic_ops/testing/api/v1/capabilities_test.go
+++ b/traffic_ops/testing/api/v1/capabilities_test.go
@@ -21,7 +21,6 @@ package v1
 
 import "testing"
 
-import "github.com/apache/trafficcontrol/lib/go-log"
 import "github.com/apache/trafficcontrol/lib/go-tc"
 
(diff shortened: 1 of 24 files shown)


shamrickus:
The todb_test doesn't have access to testing.T, only testing.M. That having been said I'll look into the others

PR #4394, opened 2020-02-12
TP: adds configurability (show/hide columns) to the parameters and users tables

mitchell852:

Provides the ability to show/hide parameters and users table columns in TP. Hiding a column will also exclude it from the table search.

No documentation updates required as the current table structure is not documented.

- Traffic Portal

2. Navigate to https://tp.domain.com/#!/users and use the following drop down button to toggle columns on/off:

image

- ensure that the state is persistent on browser refresh
- ensure that hidden columns are not included in the search

3. Navigate to https://tp.domain.com/#!/parameters and use the following drop down button to toggle columns on/off:

image

- ensure that the state is persistent on browser refresh
- ensure that hidden columns are not included in the search

### DIFF
--- a/traffic_portal/app/src/common/modules/table/cacheGroupParameters/table.cacheGroupParameters.tpl.html
+++ b/traffic_portal/app/src/common/modules/table/cacheGroupParameters/table.cacheGroupParameters.tpl.html
@@ -33,7 +33,7 @@ under the License.
                     <span class="caret"></span>
                 </button>
                 <menu ng-click="$event.stopPropagation()" class="column-settings dropdown-menu-right dropdown-menu" uib-dropdown-menu>
-                    <li role="menuitem" ng-repeat="c in columns | orderBy:'name'">
+                    <li role="menuitem" ng-repeat="c in columns | orderBy:'name' | filter:columnFilterFn">
                         <div class="checkbox">
                             <label><input type="checkbox" ng-model="c.visible" ng-click="toggleVisibility(c.name)"> {{::c.name}}</label>
                         </div>
@@ -52,6 +52,7 @@ under the License.
                 <th>Config File</th>
                 <th>value</th>
                 <th>Secure</th>
+                <th></th>
             </tr>
             </thead>
             <tbody>
@@ -60,6 +61,9 @@ under the License.
                 <td data-search="^{{::p.configFile}}$">{{::p.configFile}}</td>
                 <td data-search="^{{::p.value}}$">{{::p.value}}</td>
                 <td data-search="^{{::p.secure}}$">{{::p.secure}}</td>
+                <td style="text-align: right;">
+                    <a class="link action-link" title="Unlink Parameter from Cache Group" ng-click="removeParameter(p.id)"><i class="fa fa-sm fa-chain-broken"></i></a>
+                </td>
             </tr>
             </tbody>
         </table>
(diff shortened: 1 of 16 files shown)


traffic_portal/app/src/common/modules/table/cacheGroupParameters/table.cacheGroupParameters.tpl.html
```diff
@@ -48,7 +60,6 @@
                 <td data-search="^{{::p.configFile}}$">{{::p.configFile}}</td>
                 <td data-search="^{{::p.value}}$">{{::p.value}}</td>
                 <td data-search="^{{::p.secure}}$">{{::p.secure}}</td>
-                <td><button type="button" class="btn btn-link" title="Unlink Parameter from Cache Group" ng-click="removeParameter(p.id)"><i class="fa fa-chain-broken"></i></button></td>
```
mhoppa: why the removal of this? I think it was the easiest way to remove associations without the need to right click or search through the big list

traffic_portal/app/src/common/modules/table/capabilityUsers/TableCapabilityUsersController.js
```diff
@@ -17,11 +17,14 @@
...
+var TableCapabilityUsersController = function(capability, capUsers, $controller, $scope, $state, locationUtils) {
 
-	$scope.capability = capability;
+	// extends the TableUsersController to inherit common methods
+	angular.extend(this, $controller('TableUsersController', { users: capUsers, $scope: $scope }));
 
-	$scope.capUsers = capUsers;
+	let capUsersTable;
+
+	$scope.capability = capability[0];
```
mhoppa: more of just for my understanding but why did this go to indexing of the capability when setting it in the scope?

traffic_portal/app/src/common/modules/table/cacheGroupParameters/table.cacheGroupParameters.tpl.html
mitchell852: this is kind of a "fake" column. it doesn't map to a property of a "parameter" so it makes things messy when defining columns in a dynamic and configurable way. i.e. 


```
			"columns": [
				{ "name": "Name", "visible": true, "searchable": true },
				{ "name": "Config File", "visible": true, "searchable": true },
				{ "name": "Value", "visible": true, "searchable": true },
				{ "name": "Secure", "visible": true, "searchable": true }
			],
```

so rather than deal with it, i just removed it. you can still remove a param from a cg using the "context menu" (right click)

mitchell852: would i have liked to leave it? yeah, but wasn't finding a good solution to keep it AND provide configurable show/hide capability of parameter columns....

traffic_portal/app/src/common/modules/table/capabilityUsers/TableCapabilityUsersController.js
mitchell852: the call to get a capability looked like this:

`GET /api/capabilities?name=foo` which returns an array of 1 so this was a bug that i fixed with this PR as you need to fetch it with the index.

traffic_portal/app/src/common/modules/table/cacheGroupParameters/table.cacheGroupParameters.tpl.html
mitchell852: actually, give me a sec. i have an idea.

traffic_portal/app/src/common/modules/table/capabilityUsers/TableCapabilityUsersController.js
mhoppa: ahh make sense 👍

traffic_portal/app/src/common/modules/table/cacheGroupParameters/table.cacheGroupParameters.tpl.html
mitchell852: updated

PR #4374, opened 2020-02-05
Create API v2 tests

shamrickus:
## What does this PR (Pull Request) do?

- [x] This PR is not related to any Issue 

## Which Traffic Control components are affected by this PR?

- Traffic Ops

## What is the best way to verify this PR?

Run the go api tests for v2, validate v1 tests still pass

## The following criteria are ALL met by this PR

- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR ensures that database migration sequence is correct OR this PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #4371, opened 2020-02-04
API 2.0 documentation

mhoppa:
## What does this PR (Pull Request) do?

Updates API documentation for version 2.0. Deletes deprecated routes. Also deletes versionadded/versionchanged as version 2.0 will be top of tree. 

## Which Traffic Control components are affected by this PR?

- Documentation

## What is the best way to verify this PR?

Build docs, as this is a pure documentation change there are no tests.

## The following criteria are ALL met by this PR

- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR ensures that database migration sequence is correct OR this PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

## Additional Information

***THIS SHOULD NOT BE MERGED OR MARKED READY UNTIL 2.0 is cut***

PR #4397, opened 2020-02-13
Fixed /api/2.0/roles always returning empty array

ocket8888:

The `/roles` endpoint when called in API version 2.0 is always returning an empty array because of how it handles the new v1.3 fields. This fixes that.

- Traffic Ops

Build and run, make a GET request to `/roles`, verify the resulting array of roles is not empty.

- master

- [x] Tests are being added in #4374 
- [x] Documentation is being added in #4371 

### DIFF
--- a/traffic_ops/traffic_ops_golang/role/roles.go
+++ b/traffic_ops/traffic_ops_golang/role/roles.go
@@ -165,15 +165,11 @@ func (role *TORole) Read() ([]interface{}, error, error, int) {
 		return nil, userErr, sysErr, errCode
 	}
 
-	if version.Major > 1 {
-		return vals, nil, nil, http.StatusOK
-	}
-
 	returnable := []interface{}{}
 	for _, val := range vals {
 		rl := val.(*TORole)
 		switch {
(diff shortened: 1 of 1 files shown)


ocket8888:
Not really sure about "medium", but at best it means that new users can not be created and old ones can't be updated - which is a bit of a big deal when users can't be deleted.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -164,6 +164,11 @@ func (role *TORole) Read() ([]interface{}, error, error, int) {
 	if userErr != nil || sysErr != nil {
 		return nil, userErr, sysErr, errCode
 	}
+
+	if version.Major > 1 {
+		return vals, nil, nil, http.StatusOK
```
mhoppa: this needs to fill in the capabilities else it will return nil for capabilities

ocket8888: It that case I'll just move this into the loop below.
Done.

PR #4410, opened 2020-02-18
Eliminating underscore

ocket8888:

Removes UnderscoreJS (`_.js`) from the Traffic Portal dependencies and codebase.

- Traffic Portal

Run the existing end-to-end TP tests, but also start a TP instance and pay close attention to any errors coming from the jQuery "data tables" and the charts on the dashboard. I think those are using lodash/defining their own variables named `_`, but the problem could just not be manifesting for me for some reason.

- [x] Existing tests should be sufficient
- [x] Documentation of managed dependencies is not necessary


mitchell852:
this is good (it's always good to remove a dependency imo) but it's highly intrusive and dangerous as the reality is that the TP end-to-end tests probably don't cover many of these "states" (not sure however). Bottom line: this is going to require a good deal of testing which is fine. Just wanted to point that out if someone (besides me) ends up reviewing/merging this.

PR #4416, opened 2020-02-20
Upgrade Traffic Router to use Traffic Ops API 2.0

ocket8888:

Upgrades Traffic Router to use Traffic Ops API v2.0 - it still uses `/internal`-based endpoints, though.

- Traffic Router

Run the existing unit and integration tests - mine were passing up to the point that it needed to connect to Traffic Monitor because I didn't know how to configure it. But that part shouldn't be affected.

- [x] Tests are unchanged because intended behavior is unchanged
- [x] Documentation unchanged because intended behavior is unchanged


rawlinp:
Changes look good, and TR tests pass. But you made the weasel angry:
```
Error                                 Unknown! traffic_router/core/src/test/resources/api/2.0/cdns/name/thecdn/sslkeys
Error                                 Unknown! traffic_router/core/src/test/resources/api/2.0/cdns/name/thecdn/sslkeys-missing-1
```

ocket8888:
ohh, I didn't think of that

PR #4418, opened 2020-02-20
Some tiny Python-client-related fixes

ocket8888:

* Addresses inaccuracies, bad formatting in the clients/README.md file
* Fixed the warning in `setup.py` reporting that "3.4 <= version 3.6" being supported (3.4 isn't even receiving security fixes anymore)
* Removed an unused import from the toaccess scripts
* Added `-v`/`--version` support to the toaccess scripts
* Fixed a bug when formatting the API version that caused all toaccess scripts to be unusable (now using fstrings)
* Incremented micro version to account for bugfix
* Removed admonition about dropping support for Python 2 - that's been done and the package will now refuse to build for Python 2.

- Documentation
- Traffic Control Client (Python) - but really mostly just toaccess

Install the client and try to use the toaccess scripts. I tried `toget -fkp ping` and got the expected output (my environment variables are set).

- master

- [x] Python client is untested

### DIFF
--- a/traffic_control/clients/python/trafficops/__version__.py
+++ b/traffic_control/clients/python/trafficops/__version__.py
@@ -27,4 +27,4 @@ this client will remain non-breaking for existing code using it until the next m
 released.
 """
 
-__version__ = '2.0.0'
+__version__ = '2.0.1'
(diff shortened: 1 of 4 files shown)


traffic_control/clients/python/trafficops/__version__.py
```diff
@@ -21,17 +21,10 @@
 client to grow in a version-controlled manner without being tied to the release cadence of Apache
 Traffic Control as a whole.
 
-Version 1.0 is supported for use with Apache Traffic Control version 3.0 (release pending at the
+Version 2.0 is supported for use with Apache Traffic Control version 4.0 (release pending at the
```
mhoppa: it wont work with ATC 4.0 but 4.1 right?

ocket8888: oh, right. API 2.0 isn't in ATC 4.0.

PR #4419, opened 2020-02-20
ORT Speed Improvements

rob05c:

This makes a number of changes to speed up ORT runs.
- changes `atstccfg` to not retry 404's from Traffic Ops. If TO returns a 404, it's not going to succeed on retries. With the exponential backoff retry, this saves about a minute on a TO missing Server Capabilities
- changes the default `atstccfg` cache format to Gob instead of JSON. This cuts the time of a total ORT run in about half, and for an individual cache file, reduces the parse time to about 1/5 the JSON time.
- changes Delivery Service Servers cache format to CSV, which takes about half the time of Gob, and is by far the largest file to load and parse.
- Changes the default cache time to 15m instead of 1m. The time was always intended to be a full ORT run, and 1m is far too short. This also exposes the flag in ORT itself, so users can change it if they want. It also changes ORT to delete the cache when it starts, to ensure the `atstccfg` cache is never reused by different ORT runs.
- Incidentally, this also abstracts the cache encoding, making it very easy for anyone to change or add new cache formats, if they find something faster, or prefer the easier-to-debug JSON. I left the JSON encoder in, and it's a 1-line compile-time change. I also tested CBOR, which is preferable to Gob as a standard, but it was over twice as slow as Gob, and `atstccfg` currently has 0 external dependencies, so I decided not to include it; but it's trivial to add.

Altogether, this cuts the new ORT total run time to about 1/3 what it was, and puts it close to the time it took before the `atstccfg` client-side config generation was added.

- Traffic Ops ORT

Run unit tests. Run ORT, verify it still creates correct config files.

Not a bug fix.

## Additional Information

### DIFF
--- a/traffic_ops/ort/atstccfg/toreq/caching.go
+++ b/traffic_ops/ort/atstccfg/toreq/caching.go
@@ -20,30 +20,53 @@ package toreq
  */
 
 import (
+	"bufio"
+	"encoding/gob"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/traffic_ops/ort/atstccfg/config"
 )
 
-// GetCachedJSON attempts to get the given object from tempDir/cacheFileName.
+// DefaultCacheFormat is the encoder, decoder, and file extension to use for caching.
+// This may be changed at compile-time. See CacheFormatJSON.
+// Note this is not used for all cache files. Notably, Delivery Service Servers use a custom CSV format, which is faster than gob.
+var DefaultCacheFormat = CacheFormatGob
+
+// GetCached attempts to get the given object from tempDir/cacheFileName.
 // If the cache file doesn't exist, is too old, or is malformed, it uses getter to get the object, and stores it in cacheFileName.
-// The object is placed in obj (which must be a pointer to the type of object to decode from JSON), and the error from getter is returned.
-func GetCachedJSON(cfg config.TCCfg, cacheFileName string, obj interface{}, getter func(obj interface{}) error) error {
-	err := GetJSONObjFromFile(cfg.TempDir, cacheFileName, cfg.CacheFileMaxAge, obj)
+// The object is placed in obj (which must be a pointer to the type of object to decode), and the error from getter is returned.
+// The cache format is defined by CacheEncoder and CacheDecoder, which may be changed at compile-time.
+func GetCached(cfg config.TCCfg, cacheFileName string, obj interface{}, getter func(obj interface{}) error) error {
+	return GetCachedWithFormat(cfg, cacheFileName, obj, getter, DefaultCacheFormat)
+}
+
+// GetCachedDSS is like GetCached, but optimized for Delivery Service Servers, which is massive.
+func GetCachedDSS(cfg config.TCCfg, cacheFileName string, obj *[]tc.DeliveryServiceServer, getter func(obj interface{}) error) error {
+	return GetCachedWithFormat(cfg, cacheFileName, obj, getter, CacheFormatDSS)
+}
+
+func GetCachedWithFormat(cfg config.TCCfg, cacheFileName string, obj interface{}, getter func(obj interface{}) error, cacheFormat CacheFormat) error {
+	cacheFileName += cacheFormat.Extension
+	start := time.Now()
+	err := ReadCache(cacheFormat.Decoder, cfg.TempDir, cacheFileName, cfg.CacheFileMaxAge, obj)
 	if err == nil {
+		log.Infof("ReadCache %v (hit) took %v\n", cacheFileName, time.Since(start).Round(time.Millisecond))
 		return nil
 	}
 
-	log.Infoln("GetCachedJSON failed to get object from '" + cfg.TempDir + "/" + cacheFileName + "', calling getter: " + err.Error())
+	log.Infoln("ReadCache failed to get object from '" + cfg.TempDir + "/" + cacheFileName + "', calling getter: " + err.Error())
 
 	currentRetry := 0
 	for {
@@ -51,7 +74,10 @@ func GetCachedJSON(cfg config.TCCfg, cacheFileName string, obj interface{}, gett
 		if err == nil {
 			break
 		}
-
+		if strings.Contains(strings.ToLower(err.Error()), "not found") {
+			// if the server returned a 404, retrying won't help
+			return errors.New("getting uncached: " + err.Error())
+		}
 		if currentRetry == cfg.NumRetries {
 			return errors.New("getting uncached: " + err.Error())
 		}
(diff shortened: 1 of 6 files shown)


traffic_ops/ort/atstccfg/toreq/caching.go
```diff
@@ -105,24 +125,42 @@ func GetJSONObjFromFile(tempDir string, cacheFileName string, cacheFileMaxAge ti
...
 		return fmt.Errorf("object file too old, max age %dms less than file age %dms", int(cacheFileMaxAge/time.Millisecond), int(objFileAge/time.Millisecond))
 	}
-
-	bts, err := ioutil.ReadAll(objFile)
-	if err != nil {
-		return errors.New("reading object from file '" + objPath + "':" + err.Error())
-	}
-
-	if err := json.Unmarshal(bts, obj); err != nil {
+	if err := decode(objFile, obj); err != nil {
```
alficles: Likewise with the decoder. Should a nil encoder have a default meaning?

traffic_ops/ort/atstccfg/toreq/caching.go
```diff
@@ -177,3 +215,65 @@ func GetCookiesFromFile(tempDir string, cacheFileMaxAge time.Duration) (string,
...
+	// Please don't change this to use encoding/csv unless you benchmark and prove it's at least as fast. DSS is massive, and performance is important.
+	obj, ok := iObj.(*[]tc.DeliveryServiceServer)
+	if !ok {
+		return fmt.Errorf("object is '%T' must be a *[]tc.DeliveryServiceServer\n", iObj)
+	}
+	for _, dss := range *obj {
+		if dss.DeliveryService == nil || dss.Server == nil {
+			continue // TODO warn?
+		}
+		if _, err := io.WriteString(w, strconv.Itoa(*dss.DeliveryService)+`,`+strconv.Itoa(*dss.Server)+"\n"); err != nil {
```
alficles: Would this be faster with multiple writes instead of string concatenation?

traffic_ops/ort/atstccfg/toreq/caching.go
```diff
@@ -177,3 +215,65 @@ func GetCookiesFromFile(tempDir string, cacheFileMaxAge time.Duration) (string,
...
+// If iObj is not a *[]tc.DeliveryServiceServer it immediately returns an error.
+func DSSDecode(r io.Reader, iObj interface{}) error {
+	// Please don't change this to use encoding/csv unless you benchmark and prove it's at least as fast. DSS is massive, and performance is important.
+	obj, ok := iObj.(*[]tc.DeliveryServiceServer)
+	if !ok {
+		return fmt.Errorf("object is '%T' must be a *[]tc.DeliveryServiceServer\n", iObj)
+	}
+	objFileReader := bufio.NewReader(r)
+	for {
+		dsStr, err := objFileReader.ReadString(',')
```
alficles: Why use strings here? Wouldn't []byte be faster?

alficles: There's a lot of string allocation that is totally unnecessary here.

traffic_ops/ort/atstccfg/toreq/caching.go
```diff
@@ -177,3 +215,65 @@ func GetCookiesFromFile(tempDir string, cacheFileMaxAge time.Duration) (string,
...
+		dsStr = dsStr[:len(dsStr)-1] // ReadString(',') includes the comma; remove it
+
+		ds, err := strconv.Atoi(dsStr)
+		if err != nil {
+			return errors.New("malformed object file: first field should be ds id, but is not an integer: '" + dsStr + "'")
+		}
+		svStr, err := objFileReader.ReadString('\n')
+		if err != nil {
+			if err == io.EOF {
+				return errors.New("malformed object file: final line had ds but not server")
```
alficles: This can also happen if the final line is missing the newline.

traffic_ops/ort/traffic_ops_ort.pl
```diff
@@ -1847,7 +1855,12 @@ sub get_cfg_file_list {
 	my $cdn_name;
 	my $uri = "/api/1.4/servers/$host_name/configfiles/ats";
 
+	# Telling atstccfg to use-cache=false will cause it to delete the cache directory
+	# Which is what we want: when ORT starts to run, delete the cache. We only want to use the atstccfg cache within the same ORT run, not across different runs.
+	my $real_use_cache = $use_cache;
+	$use_cache = 0;
```
alficles: This is a very sad way to pass an argument, even for perl. :(

traffic_ops/ort/traffic_ops_ort.pl
```diff
@@ -1847,7 +1855,12 @@ sub get_cfg_file_list {
 	my $cdn_name;
 	my $uri = "/api/1.4/servers/$host_name/configfiles/ats";
 
+	# Telling atstccfg to use-cache=false will cause it to delete the cache directory
+	# Which is what we want: when ORT starts to run, delete the cache. We only want to use the atstccfg cache within the same ORT run, not across different runs.
```
alficles: So, this only works because this is the first request of the run. If someone adds a call before this at some point, they could easily wind up introducing a very subtle race condition. It would be nice to see this a bit more robust? Maybe a global clear-cache flag that is checked and set by lwp_get itself that would run on first call regardless of what that was?

traffic_ops/ort/atstccfg/toreq/caching.go
rob05c: There's no `[]byte` to int conversion func in Go.

traffic_ops/ort/atstccfg/toreq/caching.go
```diff
@@ -84,17 +105,16 @@ func WriteCacheJSON(tempDir string, cacheFileName string, obj interface{}) {
 	}
 	defer objFile.Close()
 
-	if _, err := objFile.Write(objBts); err != nil {
-		log.Errorln("writing object cache file '" + objPath + "': " + err.Error())
+	if err := encode(objFile, obj); err != nil {
```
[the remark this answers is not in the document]
rob05c: If it's nil it should probably be the default, which is currently Gob.
But, I'd rather it logged/returned an error. I think requiring a valid value is safer and much easier to debug, than defaulting when the nil pointer could be an accident.
But yes, the panic is bad.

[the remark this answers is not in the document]
rob05c: Ok, I changed it to log/return errors instead of panicing.
Heaven forbid we have the type safety of nonnullable functions.

traffic_ops/ort/atstccfg/toreq/caching.go
rob05c: Changed to return an error, as above

traffic_ops/ort/atstccfg/toreq/caching.go
rob05c: Which will never happen unless someone manually modifies the cache file. But, I changed the message to be more informative.

traffic_ops/ort/traffic_ops_ort.pl
rob05c: It's following the idiom of other arguments 🤷

traffic_ops/ort/traffic_ops_ort.pl
rob05c: Not entirely sure I agree adding a global variable increases the overall robustness. But, I changed it.

traffic_ops/ort/traffic_ops_ort.pl
```diff
@@ -1508,12 +1512,22 @@ sub lwp_get {
...
 		$no_cache_arg = '--no-cache';
 	}
 
-	$response_content = `$atstccfg_cmd $no_cache_arg --traffic-ops-user='$TO_USER' --traffic-ops-password='$TO_PASS' --traffic-ops-url='$request' --log-location-error=stderr --log-location-warning=stderr --log-location-info=null 2>$atstccfg_log_path`;
+	my $cache_age_arg='';
+	if (length $cache_max_age_seconds > 0) {
+		$cache_age_arg = "--cache-file-max-age-seconds=$cache_max_age_seconds";
+	}
+
+	$response_content = `$atstccfg_cmd $no_cache_arg $cache_age_arg --traffic-ops-user='$TO_USER' --traffic-ops-password='$TO_PASS' --traffic-ops-url='$request' --log-location-error=stderr --log-location-warning=stderr --log-location-info=null 2>$atstccfg_log_path`;
```
alficles: This introduces an unnecessary space when you don't pass `--no-cache`.

rob05c: It doesn't hurt anything though, the arg parser can handle it. IMO safer and easier to understand than the code to avoid it.

PR #4423, opened 2020-02-21
Deprecate cachegroup/:id/unassigned_parameters

mhoppa:

- Documentation
- Traffic Ops


mhoppa:
All done 👍

traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
```diff
@@ -17,13 +17,15 @@
...
-var TableCacheGroupParamsUnassignedController = function(cg, parameters, $scope, $uibModalInstance) {
+var TableCacheGroupParamsUnassignedController = function(cg, allParams, assignedParams, $scope, $uibModalInstance) {
 
 	var selectedParams = [];
 
 	$scope.cg = cg;
 
-	$scope.unassignedParams = parameters;
+	$scope.unassignedParams = _.filter(allParams, function(param) {
+		return !_.find(assignedParams, function(assignedParam) { return assignedParam.id == param.id });
```
ocket8888: Instead of using `_.find`, consider using the Javascript built-in [`Array.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) method.

Also, though, you could avoid iterating the whole array of assigned params if it's a Set of ids first:
```javascript
// In TableCacheGroupParametersController.js
// ...
assignedParams: function() {
	return new Set(parameters.map(x=>x.id));
	// I think maybe webpack won't actually handle fat-arrow functions, so might need to do:
	// return new Set(parameters.map(function(x){return x.id;}));
}
//...
// now in TableCacheGroupParamsUnassignedController.js
$scope.unassignedParams = allParams.filter(x=>!assignedParams.has(x.id));
// or, if you can't use "fat-arrow" functions:
$scope.unassignedParams = allParams.filter(
	function(p) {
		return !assignedParams.has(p.id);
	}
);
```

traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
```diff
@@ -17,13 +17,15 @@
...
 
-var TableCacheGroupParamsUnassignedController = function(cg, parameters, $scope, $uibModalInstance) {
+var TableCacheGroupParamsUnassignedController = function(cg, allParams, assignedParams, $scope, $uibModalInstance) {
 
 	var selectedParams = [];
 
 	$scope.cg = cg;
 
-	$scope.unassignedParams = parameters;
+	$scope.unassignedParams = _.filter(allParams, function(param) {
```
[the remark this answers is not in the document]
mhoppa: for my knowledge why?

traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
mhoppa: sure ill look into it

traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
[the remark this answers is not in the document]
ocket8888: Because it saves me work :P

Everything we're using `_` for can be done natively in Javascript in every browser with above a 0.5% market share - except Internet Explorer and Opera Mini, both of which we don't support.

So I have a PR open to remove our dependency on it: #4410 .

[It's also every so slightly faster because it is built in](https://jsperf.com/array-prototype-filter-vs-filter).

traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
```diff
@@ -17,22 +17,26 @@
...
 	var addParam = function(paramId) {
-		if (_.indexOf(selectedParams, paramId) == -1) {
+		if (selectedParams.indexOf(paramId) == -1) {
 			selectedParams.push(paramId);
 		}
 	};
 
 	var removeParam = function(paramId) {
-		selectedParams = _.without(selectedParams, paramId);
+		selectedParams = selectedParams.without(paramId);
```
ocket8888: > ``` TypeError: Array.prototype.without is not a function ```

There isn't a 1:1 correspondence between `_` and ES6 collection types in terms of exact function names, just functionality. The equivalent in this case is just a negative filter:

```javascript
selectedParams = selectedParams.filter(x=>x!==paramId);
```

Or, without "fat-arrow" functions

```javascript
selectedParams = selectedParams.filter(
	function (param) {
		return param !== paramId;
	}
);
```

traffic_portal/app/src/common/modules/table/cacheGroupParameters/TableCacheGroupParamsUnassignedController.js
```diff
@@ -17,22 +17,26 @@
...
-	$scope.unassignedParams = parameters;
+	$scope.unassignedParams = allParams.filter(
+		function(p) {
+			return !assignedParams.has(p.id);
+		}
+	);
 
 	var addParam = function(paramId) {
-		if (_.indexOf(selectedParams, paramId) == -1) {
+		if (selectedParams.indexOf(paramId) == -1) {
```
ocket8888: comparison should use `===` not `==`

PR #4429, opened 2020-02-21
Add support for TO_API_VERSION env var in to-access.sh

rawlinp:
By default, make requests at 2.0 but allow overriding the version when
necessary to deal with older TO servers.

- CDN in a Box

With TO_URL, TO_USER, and TO_PASSWORD vars set appropriately:
`. to-access.sh; to-auth` -> verify that the login request was made to TO at API version 2.0.
`TO_API_VERSION=1.4; . to-access.sh; to-auth` -> verify that the login request was made to TO at API version 1.4.

Start up a Docker CIAB instance, make sure it all starts up properly as it depends on the `to-access.sh` script.

- [x] Starting up CIAB tests this script pretty well -- doesn't have its own specific tests
- [x] The to-access.sh script doesn't have docs
- [x] A changelog entry would be a little overkill for this undocumented script


zrhoffman:
Do you think `TO_API_VERSION` should be in `variables.env`?

rawlinp:
I don't think it would be required there. In general CIAB should just take the default version set in the script itself -- only users sourcing the script directly for non-CIAB purposes would care to set a specific version.

PR #4430, opened 2020-02-21
DNSSEC refresh breaks on DS if KSK and ZSK both are expired

mhoppa:

- Traffic Ops

Make a call to `cdns/dnsseckeys/refresh` with the given environment having a DNSSEC enabled CDN and HTTP/DNS/STEERING DS with both expired KSK and ZSK keys. Make sure the result is both being updated.

master

### DIFF
--- a/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
+++ b/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
@@ -251,6 +251,10 @@ func doDNSSECKeyRefresh(tx *sql.Tx, cfg *config.Config) {
 				if err != nil {
 					log.Errorln("refreshing DNSSEC Keys: regenerating expired ZSK keys for ds '" + string(ds.DSName) + "': " + err.Error())
 				} else {
+					if existingNewKeys, ok := keys[string(ds.DSName)]; ok {
+						existingNewKeys.ZSK = newKeys.ZSK
+						newKeys = existingNewKeys
+					}
 					keys[string(ds.DSName)] = newKeys
 					updatedAny = true
 				}


rawlinp:
What does this mean by "breaks"? It refreshes just the KSK but not also the ZSK?

mhoppa:
opposite refreshes ZSK but not KSK

traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
```diff
@@ -251,6 +251,10 @@ func doDNSSECKeyRefresh(tx *sql.Tx, cfg *config.Config) {
 				if err != nil {
 					log.Errorln("refreshing DNSSEC Keys: regenerating expired ZSK keys for ds '" + string(ds.DSName) + "': " + err.Error())
 				} else {
+					if existingNewKeys, ok := keys[string(ds.DSName)]; ok {
```
rawlinp: So, I'm not entirely convinced this is correct. The `regenExpiredKeys` function is supposed to return the right keyset to put in the `keys` map -- take a look at https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/cdn/genksk.go#L233-L239

We have to be extremely careful with DNSSEC keys, and the existing code _seems_ to match up w/ what the Perl did. Does this mean Perl had this bug? And maybe we just didn't catch it because generally the KSK and ZSK expire at different intervals?

mhoppa: yeah maybe I guess. And yeah the key set is returned but it’s overwritten. I mean we can leave it broken w/ the bug since ideally the next cron run will catch the ksk key as long as the expiration date is bigger then a cron time for the zsk key

PR #4432, opened 2020-02-22
Fix broken ORT caching resulting in malformed config files

rob05c:
Fixes ORT caching to CBOR. The current Gob encoding is fundamentally broken. See https://github.com/golang/go/issues/4609 . It's not isomorphic: it does not decode what in encodes. In practical terms, default values like 0 become null pointers, resulting in missing remap lines in the config.

This changes the default to CBOR, which is an IETF RFC standard. See https://tools.ietf.org/html/rfc7049 .

It also adds a unit test, to verify the default cache format is isomorphic. Gob fails the test, CBOR and JSON pass it.

- Traffic Ops ORT

Run unit tests.

Generate remap.config with assigned delivery services of type "HTTP" (which is represented by `0`, the default int value). Without this fix, remap lines will be missing. With fix, config file should have appropriate remap lines for HTTP DSes.

- master
- 4.0.x

## Additional Information

### DIFF
--- a/.dependency_license
+++ b/.dependency_license
@@ -113,6 +113,8 @@ traffic_portal/app/src/assets/css/colReorder.dataTables\..*, MIT
 traffic_portal/app/src/assets/js/colReorder.dataTables\..*, MIT
 traffic_portal/app/src/assets/js/dataTables.buttons\..*, MIT
 traffic_portal/app/src/assets/js/buttons.html5\..*, MIT
+traffic_ops/ort/atstccfg/vendor/github\.com/fxamacker/cbor/.*, MIT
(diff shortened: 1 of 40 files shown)


rob05c:
retest this please

rob05c:
Ok, I added updating the Go build version to 1.12 to this PR. I tested building, installing, and running the compiled binaries, on both TO and ORT.

I also created a PR using JSON instead of CBOR, which make ORT ~50% slower, but doesn't require updating Go to 1.12, see #4433 . I'll let the community decide which to merge.

rob05c:
retest this please

rob05c:
Ah. I was not aware of that :(

rob05c:
Closing in lieu of #4433

PR #4435, opened 2020-02-24
[Backport 4.0.x] Fix ort caching to JSON (#4433)

rawlinp:
See #4432 - this is an alternative to that PR, using JSON, which make an ORT run ~50% slower than CBOR, but doesn't require upgrading Go to 1.12. I'll let the community decide which to merge.

Fixes ORT caching to JSON. The current Gob encoding is fundamentally broken. See https://github.com/golang/go/issues/4609 . It's not isomorphic: it does not decode what in encodes. In practical terms, default values like 0 become null pointers, resulting in missing remap lines in the config.

This changes the default to JSON.

It also adds a unit test, to verify the default cache format is isomorphic. Gob fails the test, CBOR and JSON pass it.

- Traffic Ops ORT

Run unit tests.

Generate remap.config with assigned delivery services of type "HTTP" (which is represented by `0`, the default int value). Without this fix, remap lines will be missing. With fix, config file should have appropriate remap lines for HTTP DSes.

- master
- 4.0.x

## Additional Information

### DIFF
--- a/traffic_ops/ort/atstccfg/toreq/caching.go
+++ b/traffic_ops/ort/atstccfg/toreq/caching.go
@@ -21,7 +21,6 @@ package toreq
 
 import (
 	"bufio"
-	"encoding/gob"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


rawlinp:
Whoops, didn't see https://github.com/apache/trafficcontrol/pull/4434 -- closing as a duplicate backport.

PR #4448, opened 2020-02-25
Deprecate /deliveryservices/hostname/:name/sslkeys

mhoppa:

- Documentation
- Traffic Ops


traffic_ops/traffic_ops_golang/deliveryservice/keys.go
```diff
@@ -226,30 +231,38 @@ func GetSSLKeysByXMLID(w http.ResponseWriter, r *http.Request) {
 }
 
 func getSSLKeysByXMLIDHelper(xmlID string, inf *api.APIInfo, w http.ResponseWriter, r *http.Request) {
+	alerts := tc.CreateAlerts(tc.WarnLevel, hostnameKeyDepMsg)
```
ocket8888: doing this in these "*Helper" functions is adding the deprecation notice to `deliveryservices/xmlId/{{xmlId}}/sslkeys` as well, which is ironic.

mhoppa: 🤦‍♂ - good catch

PR #4400, opened 2020-02-13
TP: Fixes missing ecs enabled property on a number of ds tables

mitchell852:
## What does this PR (Pull Request) do?

The newly added ECS Enabled field was not added to a handful of ds tables which caused each table to fail to render.

- [x] This PR fixes #4404 

no documentation changes needed for this regression.
existing UI tests do not cover tables at the column level.
no change log entry required.

## Which Traffic Control components are affected by this PR?

- Traffic Portal

## What is the best way to verify this PR?

1. Run the [UI tests](https://github.com/apache/trafficcontrol/tree/master/traffic_portal/test/end_to_end)

2. Navigate to each table and ensure it renders properly and you can show/hide table columns.

https://tp.example.com/#!/cdns/1/delivery-services
https://tp.example.com/#!/profiles/458/delivery-services (must be a profile of type=DS_PROFILE)
https://tp.example.com/#!/servers/1036/delivery-services (must be a server of type=EDGE)
https://tp.example.com/#!/tenants/4/delivery-services
https://tp.example.com/#!/types/9/delivery-services (must be a type where useintable=deliveryservice)
https://tp.example.com/#!/users/597/delivery-services

## If this is a bug fix, what versions of Traffic Control are affected?
- master

## The following criteria are ALL met by this PR

- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR ensures that database migration sequence is correct OR this PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #4455, opened 2020-02-28
Add support for slice range request plugin in HTTP* & DNS* delivery services

mhoppa:

- This PR adds support for enabling the ATS [Slice Plugin](https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/slice.en.html) for the range request handler. It includes changes for TP/TO/ATSCONFIG

- Documentation
- Traffic Ops
- Traffic Ops ORT
- Traffic Portal

This PR includes changes for three components so to verify it tests needs to be run against TO/TP/ATSCONFIG.

### TO

* Create/update a HTTP/DNS DS with `Range Request Handler` set to 3 and the new field `Slice Plugin Block Size` to a value between `262144 - 33554432` it should succeed
*  Create/update a HTTP/DNS DS with `Range Request Handler` set to 3 and leave new field `Slice Plugin Block Size` as null it should fail
*  Create/update a HTTP/DNS DS with `Range Request Handler` not set to 3 and the new field `Slice Plugin Block Size` to a value between `262144 - 33554432` it should fail
* Create/update a HTTP/DNS DS with `Range Request Handler` set to 3 and the new field `Slice Plugin Block Size` to a value above `33554432` it should fail
* Create/update a HTTP/DNS DS with `Range Request Handler` set to 3 and the new field `Slice Plugin Block Size` to a value below `262144` it should fail

### TP

* Create/update a HTTP/DNS DS with `Range Request Handler` set to `Use the slice plugin` and the `Slice Plugin Block Size` should become un grayed to fill out. And any value outside of `262144 - 33554432` should have an error.
* Create/update a HTTP/DNS DS with `Range Request Handler` not set to `Use the slice plugin` and the `Slice Plugin Block Size` should become empty and grayed out

### ATSTCCFG 

1. Create a DNS/HTTP DS with `Range Request Handler` set to `3` and `Slice Plugin Block Size` set between `262144 - 33554432`.
2. Assign a edge to that DS
3. Generate the remap config file for the edge and it should include `@plugin=slice.so @pparam=blockbytes:[Block Size] @plugin=cache_range_requests.so` 

If in CIAB this can be done by running the following locally ->
```
./atstccfg --no-cache  -s -P twelve -U admin -u "https://127.0.0.1:6443/api/2.0/servers/mid/configfiles/ats/remap.config"

./atstccfg --no-cache  -s -P twelve -U admin -u "https://127.0.0.1:6443/api/2.0/servers/edge/configfiles/ats/remap.config"
```

## Additional Information

slice_plugin_demo


traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -423,6 +425,25 @@ <h3>Current Value</h3>
...
+                            </div>
+                        </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <input id="slicePluginBlockSize" name="slicePluginBlockSize" type="number" class="form-control" ng-model="deliveryService.slicePluginBlockSize" min="262144" max="33554432" ng-readonly="dsCurrent.rangeRequestHandling != 3">
+                            <small class="input-error" ng-show="hasPropertyError(cacheConfig.slicePluginBlockSize, 'min')">Too Small (must be greater than or equal to 262144)</small>
+                            <small class="input-error" ng-show="hasPropertyError(cacheConfig.slicePluginBlockSize, 'max')">Too Large (must be smaller than or equal to 33554432)</small>
+                            <small class="input-error" ng-show="dsCurrent.rangeRequestHandling == 3 && dsCurrent.slicePluginBlockSize == null && !(hasPropertyError(cacheConfig.slicePluginBlockSize, 'min') || hasPropertyError(cacheConfig.slicePluginBlockSize, 'max'))">Required</small>
```
mitchell852: unfortunately, the way this is done, an empty value when it IS required will not disable the "update" button on the form. i think you can add an ng-required="dsCurrent.rangeRequestHandling == 3" to the input tag and then just go with this:

<small class="input-error" ng-show="hasPropertyError(cacheConfig. slicePluginBlockSize, 'required')">Required</small>

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -415,6 +416,7 @@ <h3>Current Value</h3>
                                 <option ng-value="0">Don't cache Range Requests</option>
                                 <option ng-value="1">Use the background_fetch plugin</option>
                                 <option ng-value="2">Use the cache_range_requests plugin</option>
+                                <option ng-value="3">Use the slice plugin</option>
```
mitchell852: you missed your ng-change="rangeRequestSelected()" on the select.

mitchell852:
when adding a property to a ds, you are going to want to add it to all the UI tables as well. 

Start by adding the property here: https://github.com/apache/trafficcontrol/blob/master/traffic_portal/app/src/common/modules/table/deliveryServices/TableDeliveryServicesController.js#L215

leave it visible=false, searchable=false by default.

then you'll need to add it to all the delivery services tables (yeah, the duplication sucks):

table.deliveryServices.tpl.html
table.cdnDeliveryServices.tpl.html
table.profileDeliveryServices.tpl.html
table.serverDeliveryServices.tpl.html
table.tenantDeliveryServices.tpl.html
table.typeDeliveryServices.tpl.html
table.userDeliveryServices.tpl.html

Here's a PR that did the same thing: https://github.com/apache/trafficcontrol/pull/4400

mhoppa:
@mitchell852 done! FWIW CI is failing due to weasel errors that is being addressed in another PR

mitchell852:
conflict

mitchell852:
@mhoppa - this is only for HTTP* and DNS* delivery services, correct? want to put that in the title of this PR maybe?

ocket8888:
Why is the block size restricted to that, specific value range?

Do you think we could change the name to something that isn't so directly tied to ATS? Instead of "Slice Plugin Block Size" maybe just something like "Range Slice Block Size"? That way if other caching proxy software is capable of implementing it the option will still make sense, regardless of what's being run on the back-end.

mhoppa:
Block size in the plugin is restricted to that. I thought better to put that in then fail silently on the caches.

Isnt most of these settings already tied to ATS very much so? and FWIW I read "Slice Plugin Block Size" literally the same as "Range Slice Block Size"

ocket8888:
So I guess my new question is: why does the plugin limit to within that range? Which isn't really something I can expect you to answer, so fair enough.

A lot of attributes of Delivery Services (and many other things) are very ATS-specific, but I was hoping we could avoid adding to that problem.

The only difference is that "Slice Plugin" is clearly referring to some specific plugin configuration for ATS whereas "Range Slice" instead is more generally referring to the idea of slicing ranges, which may or may not be implemented as a plugin in whatever caching software you're using.

mhoppa:
> So I guess my new question is: why does the plugin limit to within that range? Which isn't really something I can expect you to answer, so fair enough.
hmm I am not sure - @traeak  @dsouza93 might be able to shed light on it 
> 
> A lot of attributes of Delivery Services (and many other things) are very ATS-specific, but I was hoping we could avoid adding to that problem.
> 
> The only difference is that "Slice Plugin" is clearly referring to some specific plugin configuration for ATS whereas "Range Slice" instead is more generally referring to the idea of slicing ranges, which may or may not be implemented as a plugin in whatever caching software you're using. In this context (today) they are - and should be - pretty much exactly the same. It's just Pendantic Semantics ™️ .

Ill buy this ^ let me change it. :)

traeak:
The choice of range limitation was fairly arbitrary.  On the ATS side it's most practical to use just under 1MB as by default the cache drives write in 1MB chunks (with some small binary header ~72 bytes).  Going to huge sizes above 10MB (WAG) or so doesn't make sense.  When TC is able to create a test configuration and send it off to an ATS instance for rule validation then that can be caught.

ocket8888:
I guess that makes some sense. Still seems weird to me place arbitrary limits on a plugin's functionality, but I guess that ship has sailed.

lib/go-atscfg/remapdotconfig_test.go
```diff
@@ -1204,6 +1204,111 @@ func TestMakeRemapDotConfigMidRangeRequestHandling(t *testing.T) {
...
+	txt := MakeRemapDotConfig(serverName, toToolName, toURL, atsMajorVersion, cacheURLConfigParams, dsProfilesCacheKeyConfigParams, serverPackageParamData, serverInfo, remapDSData)
+
+	txt = strings.TrimSpace(txt)
+
+	testComment(t, txt, string(serverName), toToolName, toURL)
+
+	txtLines := strings.Split(txt, "\n")
+
+	if len(txtLines) != 2 {
+		t.Errorf("expected one line for each remap plus a comment, actual: '%v' count %v", txt, len(txtLines))
```
ocket8888: This should probably be fatal, to avoid a segfault in the next statement.

lib/go-atscfg/remapdotconfig_test.go
```diff
@@ -4663,6 +4768,110 @@ func TestMakeRemapDotConfigEdgeRangeRequestBGFetch(t *testing.T) {
...
+	txt := MakeRemapDotConfig(serverName, toToolName, toURL, atsMajorVersion, cacheURLConfigParams, dsProfilesCacheKeyConfigParams, serverPackageParamData, serverInfo, remapDSData)
+
+	txt = strings.TrimSpace(txt)
+
+	testComment(t, txt, string(serverName), toToolName, toURL)
+
+	txtLines := strings.Split(txt, "\n")
+
+	if len(txtLines) != 2 {
+		t.Fatalf("expected 1 remaps from HTTP_TO_HTTPS DS, actual: '%v' count %v", txt, len(txtLines))
```
ocket8888: same as above

lib/go-tc/deliveryservices.go
```diff
@@ -390,6 +392,23 @@ func (ds *DeliveryServiceNullable) validateTypeFields(tx *sql.Tx) error {
...
 			validation.NewStringRule(validateOrgServerFQDN, "must start with http:// or https:// and be followed by a valid hostname with an optional port (no trailing slash)")),
+		"slicePluginBlockSize": validation.Validate(ds,
+			validation.By(func(dsi interface{}) error {
+				ds := dsi.(*DeliveryServiceNullable)
+				if ds.RangeRequestHandling != nil {
+					if *ds.RangeRequestHandling == 3 {
+						return validation.Validate(ds.SlicePluginBlockSize, validation.Required,
+							// Per Slice Plugin implementation
+							validation.Min(262144),   // 256KB
+							validation.Max(33554432), // 32MB
```
ocket8888: nit but these are actually 256KiB and 32MiB not 256KB and 32MB, respectively.

traffic_ops/app/db/migrations/20200227000000_add_ds_slice_block_size.sql
```diff
@@ -0,0 +1,22 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE deliveryservice ADD COLUMN slice_plugin_block_size int DEFAULT NULL;
```
ocket8888: IMO data model constraints should be implemented in the database (not that there's then anything wrong with checking at the API level) with a `CHECK` clause like:
```sql
ALTER TABLE deliveryservice
ADD COLUMN range_slice_block_size integer
CHECK (range_slice_block_size >= 262144 AND range_slice_block_size <= 33554432)
DEFAULT NULL;
```

lib/go-atscfg/remapdotconfig_test.go
mhoppa: this is fatal?

traffic_ops/app/db/migrations/20200227000000_add_ds_slice_block_size.sql
mhoppa: im indifferent - changed it to this ^5

lib/go-tc/deliveryservices.go
mhoppa: :)

lib/go-atscfg/remapdotconfig_test.go
mhoppa: 🖖

lib/go-atscfg/remapdotconfig_test.go
ocket8888: Sure looks that way. It's rare that the reviewer has copy/paste errors and the code they're reviewing doesn't.

lib/go-tc/deliveryservices.go
```diff
@@ -390,6 +392,22 @@ func (ds *DeliveryServiceNullable) validateTypeFields(tx *sql.Tx) error {
...
+							return errors.New("rangeSliceBlockSize must be set if the rangeRequestHandling is set to 3 (Use the Slice Plugin)")
+						}
+					} else {
+						if ds.RangeSliceBlockSize != nil {
+							return errors.New("rangeSliceBlockSize can only be set if the rangeRequestHandling is set to 3 (Use the Slice Plugin)")
+						}
+					}
+				}
+				return nil
+			})),
```
ocket8888: Since you removed value checking from the API validations, giving it a value for rangeSliceBlockSize that isn't in the acceptable range throws a database error, which gets reported to the user as an Internal Server Error. I think a Bad Request response with appropriate alerts is a better UX.

There's two ways to do this, I think. You can add a case to [`api.ParseDBError`](https://godoc.org/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api#ParseDBError) which _should_ cause it to Just Work :tm: . Or, you could just add back in the constraints here. I don't think having the constraints in both the API and the database is a bad thing, so I'm not really particular what approach you take.

mhoppa: ahhh great catch!

ocket8888:
So I was doing the `atstccfg` test runs, and I see the block slicing plugin/param in the edge's remap.config, but the mid doesn't have it. Should it? Because you don't mention that it does, but the first `./atstccfg` command you have there is asking for the mid's remap.config.

mhoppa:
The mid should have the following ‘map http://ats/ http://parent/ @plugin=cache_range_requests.so’

PR #4452, opened 2020-02-27
deprecate /servers/checks and move functionality to /servercheck

mattjackson220:
## What does this PR (Pull Request) do?

- [x] This PR is not related to any Issue 

This PR deprecates /servers/checks and moves functionality to /servercheck

## Which Traffic Control components are affected by this PR?

- Documentation
- Traffic Ops
- Traffic Portal

## What is the best way to verify this PR?

Verify that GET /api/1.5/servers/checks now has an appropriate deprecation alert
Verify that GET /api/2.0/servers/checks does not exist
Verify that GET /api/2.0/servercheck returns the same information that /api/1.5/servers/checks does but without the alert
Verify that Traffic Portal Cache Checks link still works (requires TP to use api 2.0)

## The following criteria are ALL met by this PR

- [x] This PR does not need testing updates
- [x] This PR includes documentation
- [x] This PR includes an update to CHANGELOG.md
- [x] This PR includes any and all required license headers
- [x] This PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #4477, opened 2020-03-10
Deprecate parameters/profile/:name

shamrickus:

- [x] This PR deprecates the TO api endpoint parameters/profile/:name

- Documentation
- Traffic Ops

Confirm doc and TO builds/tests. Ensure 1.1 api returns a deprecation warning, and 2.0 router does not exist.


shamrickus:
Not ready for review, waiting on some changes from #4452

traffic_ops/traffic_ops_golang/profileparameter/parameterprofilebyname.go
```diff
@@ -31,11 +31,16 @@ import (
 func GetProfileName(w http.ResponseWriter, r *http.Request) {
 	inf, userErr, sysErr, errCode := api.NewInfo(r, []string{"name"}, nil)
 	if userErr != nil || sysErr != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
+		api.HandleDeprecatedErr(w, r, nil, errCode, userErr, sysErr, nil)
```
mattjackson220: this should have an alternative at `profiles/name/{name}/parameters`

mattjackson220: so this function is actually used for this endpoint and the `profiles/name/{name}/parameters` endpoint which is the alternative for this after its deprecated. we will have to split that up so that the alternative doesnt also get the deprecation notice

traffic_ops/traffic_ops_golang/profileparameter/parameterprofilebyname.go
```diff
@@ -22,20 +22,59 @@ package profileparameter
...
+
+func GetProfileNameDeprecated(w http.ResponseWriter, r *http.Request) {
+	getProfileName(w, r, true)
+}
+
 func GetProfileName(w http.ResponseWriter, r *http.Request) {
+	getProfileName(w, r, false)
+}
+
+func deprecationString(deprecated bool) *string {
```
mattjackson220: just to clean up a bit and get rid of some of the ifs, what are your thoughts on something like this instead?
`func handleLocalErr(w http.ResponseWriter, r *http.Request, tx *sql.Tx, errCode int, userErr error, sysErr error, deprecated bool) {
	if deprecated {
		api.HandleDeprecatedErr(w, r, tx, errCode, userErr, sysErr, util.StrPtr(API_PROFILES_NAME_NAME_PARAMETERS))
	} else {
		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
	}
}`

mattjackson220: actually it might be good to put that in api.go and pull the alternative out as a parameter as well so that we can use it if we have other deprecated endpoints that are used in their respective alternatives as well

traffic_ops/traffic_ops_golang/profileparameter/parameterprofilebyname.go
```diff
@@ -22,20 +22,59 @@ package profileparameter
...
+		profiles, err := getParametersByProfileName(inf.Tx.Tx, name)
+		if err != nil {
+			if deprecation == nil {
+				api.HandleDeprecatedErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, err, nil, deprecation)
+			} else {
+				api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, err, nil)
+			}
+			return
+		}
+		api.WriteAlertsObj(w, r, http.StatusOK, api.CreateDeprecationAlerts(nil), profiles)
```
mattjackson220: should have alternative in the api.CreateDeprecationAlerts

traffic_ops/traffic_ops_golang/profileparameter/parameterprofilebyname.go
shamrickus: I like the idea, that's what I was originally trying to do with my change to HandleDeprecatedErr.

docs/source/api/v1/parameters_profile_name.rst
```diff
@@ -27,6 +27,9 @@ Gets details about a specific :term:`Profile`'s :term:`Parameters`
 :Roles Required: None
 :Response Type:  Array
 
+.. deprecated:: ATCv4
```
mattjackson220: for consistency do you mind moving this up directly underneath the **** above GET? looks like its up there in the other deprecated endpoints and since this whole endpoint is deprecated

docs/source/api/v1/parameters_profile_name.rst
```diff
@@ -69,9 +72,14 @@ Response Structure
...
 	X-Server-Name: traffic_ops_golang/
 	Date: Wed, 05 Dec 2018 21:08:56 GMT
-	Content-Length: 542
+	Content-Length: 628
 
-	{ "response": [
+	{ "alerts": [
+		{
+			"level": "warning",
+			"text": "This endpoint is deprecated, please use /profiles/name/:name/parameters instead"
```
mattjackson220: also for consistency can you change :name to be {name} here and in the actual response in go?

PR #4486, opened 2020-03-12
Added Alerts to several response structs

### LINKED ISSUE #3447: TO Client: Alerts are not returned even when they exist?
I am writing an api test for a PR I submitted: #3445

I have the following code:

```
func TryToDeleteDivision(t *testing.T) {
    division := testData.Divisions[0]
    resp, _, err := TOSession.DeleteDivisionByName(division.Name)

    fmt.Printf("resp: %v\n", resp)
    fmt.Printf("err: %v\n", err)

    if err == nil {
        t.Errorf("should not be able to delete a division prematurely")
        return
    }

    if len(resp.Alerts) == 0 {
        t.Errorf("error occurred while deleting division, but no alerts exist")
        return
    }

    switch resp.Alerts[0].Text {
    case "Resource not found.":
        t.Errorf("division with name %v does not exist", division.Name)
    case "This division is currently used by regions.":
        // expected return message
    default:
        t.Errorf("unexpected alert occurred: %v", resp)
    }
}
```

Below is the result. You can see that the error clearly returns the alert message I'm looking for, but when an error occurs the client doesn't return the alerts. At first glance this is confusing. Is there a particular reason this is done? I believe there is a go convention that says a returned error implies the first value is nil, but I don't think that makes sense in this case.

I can write this particular test differently, but maybe this makes sense to change?

```
resp: {[]}
err: 400 Bad Request[400] - Error requesting Traffic Ops https://localhost:6443/api/1.3/divisions/name/division1 {"alerts":[{"level":"error","text":"This division is currently used by regions."}]}
--- FAIL: TestDivisions (4.38s)
	divisions_test.go:47: error occurred while deleting division, but no alerts exist
FAIL
```

ZugNZwang:

Adds the `Alerts` struct to several response structs. 

- [X] This PR fixes #3447  

- Traffic Control Go Client 

Run Traffic Ops API Tests 
Run Go Client tests


ocket8888:
Why'd you add all those new structures? They aren't used anywhere as far as I can see.

ZugNZwang:
They will be used when testing the multiple federation routes.

ocket8888:
But you didn't add them to any tests?

ZugNZwang:
No, they'll be useful in replacing the similar structures that exist in the internal tests. We wouldn't need to redeclare them, because we could use the ones declared here.

PR #4512, opened 2020-03-17
Deprecate servers/hostname/hostName/details

mhoppa:

- Documentation
- Traffic Ops


traffic_control/clients/python/trafficops/tosession.py
```diff
@@ -1713,12 +1713,13 @@ def get_server_status_count(self):
 		:raises: Union[LoginError, OperationError]
 		"""
 
-	@api_request('get', 'servers/hostname/{name}/details', ('2.0',))
-	def get_server_details(self, name=None):
+	@api_request('get', 'servers//details', ('2.0',))
```
ocket8888: typo: double-<kbd>/</kbd>

docs/source/api/v2/servers_details.rst
```diff
@@ -0,0 +1,166 @@
...
+``servers/details``
+*******************
+Retrieves details of :ref:`tp-configure-servers`.
+
+
+``GET``
+=======
+:Auth. Required: Yes
+:Roles Required: None
+:Response Type:  Array
```
ocket8888: This should mention somewhere that the response includes additional, top-level properties besides `response` or `alerts`, because from the "Response Structure" subsection under "How to Read this Documentation", the "Array" response type is defined:

> _"The fields in the field list refer to the keys of the objects in the `response` array."_

So based on that, people might think the response here has the format:

```json
{
	"response": [{
		"limit": 1000,
		"hostName": "test",
		"response": ["some objects in here"]
	}]
}
```

if they don't look at the response example carefully.

PR #4518, opened 2020-03-18
Add service category feature to TO and TP

jehunte:

- [x] This PR is not associated with any existing issue.

This PR enables the user to associate a delivery service with a specific service category. The user can create, add, update, delete categories as necessary and the value will be used in combination with the xmlid to set the HTTP header "X-CDN-SVC" 

- Documentation
- Traffic Control Client Go
- Traffic Ops
- Traffic Ops ORT - atstccfg
- Traffic Portal

1. Validate database changes 
       - New table service category with columns (id, name, last_updated)
       - service_category column added to delivery service table
2. Verify in TP that you can:
       - create a new delivery service with a service category
       - Add, update, delete, get service categories
3. Run the ORT script and verify that the header is being set properly 
        X-CDN-SVC "{xmlid}|{service-category}"

## Additional Information

Screen Shot 2020-03-18 at 9 41 41 AM

Screen Shot 2020-03-18 at 1 19 57 PM

Screen Shot 2020-03-20 at 3 09 57 PM

Screen Shot 2020-03-20 at 3 10 07 PM

Screen Shot 2020-03-18 at 1 21 07 PM


mitchell852:
Something to think about. Should "service categories" be restricted to certain tenants? For example, imagine the following tenant tree:

```
- root
-- tenant 1
--- tenant 1.1
-- tenant 2
--- tenant 2.1
```

If I am a user in tenant 2.1 and i create a Foo service category, should users in root, tenant 1, 1.1, 2 be able to attach the Foo service category to their delivery services? Maybe that's not a problem as the tenant 2.1 user will still only see delivery services scoped to tenant 2.1 in that category. However, this could potentially result in information leakage across tenants. I.e. if i'm in the company A tenant and i create a company-a-service-category-1, everybody across all tenants can see/use that service category.

The more I think about this, service categories should be scoped to tenancy. most tenants don't want to be concerned with the service categories created by other tenants.

rob05c:
>If I am a user in tenant 2.1 and i create a Foo service category
I think the question for anything is, "Will delivery service customers who are not CDN operators be able to create this?" If so, it needs to be under tenancy, and invisible to other tenants.

mitchell852:
> I think the question for anything is, "Will delivery service customers who are not CDN operators be able to create this?" If so, it needs to be under tenancy, and invisible to other tenants.

Well i'm assuming only operators (role=operations or above) can create service categories but those service categories will show up as an option of EVERY delivery service across all tenants which seems problematic.

rob05c:
Do we expect Service Categories to be created by tenants? Or is it something only Operators will create?

mitchell852:
> Do we expect Service Categories to be created by tenants? Or is it something only Operators will create?

everybody is scoped to a tenant. operators and non operators. all users are tenant scoped.

rob05c:
Right. But for example, Delivery Service Types aren't "scoped" to tenants, because they're created by operators of the system, not customers of the system. I think it's a valid distinction, to decide if something applies across the entire CDN, or only to the Tenant who created it.

rob05c:
@mitchell852 I think I agree with you. It's possible a TC Operator will want to make a hard list of categories, e.g. "Linear," "VOD," "Images." But it's also possible the TC Operator will want to allow any DS owner to make any categories they want.

If Service Categories are scoped to Tenants, and under Roles&Capabilities, both are possible. 
For the former, the Operator just doesn't give any tenants the Create Service Category Capability, and they make the hard-coded categories on the Root tenant.
For the latter, they give everyone the Capability.

Keeping YAGNI in mind, IMO it's sufficiently likely someone will need the latter, to go ahead and support it from the start.

mitchell852:
> IMO it's sufficiently likely someone will need the latter, to go ahead and support it from the start.

Yeah, it's harder to add tenancy after the fact imo. It seems to me if i were to create a service category i might want to specify that it applies to all ds's (tenant=root) or only applies to ds's of the 2.1 tenant (tenant=2.1)

mitchell852:
also, those 2 sql scripts under additional info seem like something very specific to a certain companies use of edge header rewrites. probably not something needed as part of this PR. also, @jehunte can you make sure your screenshots are generic? i.e. no company-specific info.

jehunte:
@rob05c @mitchell852 These are great points. I will require a tenant to be associated with the service categories so tenants can only see the service categories they've created. I've also replaced the screenshots and removed the sql queries.

traffic_ops/app/db/migrations/20200315000000_add_deliveryservice_service_category.sql
```diff
@@ -0,0 +1,25 @@
...
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE deliveryservice ADD COLUMN service_category int REFERENCES service_category(id);
+CREATE TABLE IF NOT EXISTS service_category (
```
mhoppa: doesnt the table need to be created prior to adding the reference in the deliveryservice table?

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,132 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		return nil, reqInf, err
+	}
+
+	return data.Response, reqInf, nil
+}
+
+// GET a service category by the service category name
+func (to *Session) GetServiceCategoryByName(name string) ([]tc.ServiceCategory, ReqInf, error) {
+	url := fmt.Sprintf("%s?name=%s", API_serviceCategories, name)
```
mhoppa: this should use url formatting on the query parameter.

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+)
+
+const (
+	API_serviceCategories = apiBase + "/servicecategories"
+)
+
+// Create a service category
+func (to *Session) CreateServiceCategory(serviceCategory tc.ServiceCategory) (tc.Alerts, ReqInf, error) {
```
mhoppa: nit but the go docs for the functions should start with the function name

traffic_ops/testing/api/v1/todb_test.go
```diff
@@ -335,6 +335,7 @@ func Teardown(db *sql.DB) error {
 	DELETE FROM cachegroup;
 	DELETE FROM coordinate;
 	DELETE FROM type;
+	DELETE FROM service_category;
```
mhoppa: I dont think this needs to be in the v1 tests?

traffic_ops/testing/api/v2/withobjs_test.go
```diff
@@ -99,6 +100,7 @@ var withFuncs = map[TCObj]TCObjFuncs{
 	ServerChecks:                         {CreateTestServerChecks, DeleteTestServerChecks},
 	ServerServerCapabilities:             {CreateTestServerServerCapabilities, DeleteTestServerServerCapabilities},
 	Servers:                              {CreateTestServers, DeleteTestServers},
+	ServiceCategories:					  {CreateTestServiceCategory, DeleteTestServiceCategories},
```
mhoppa: gofmt

mhoppa: is there tests cases related to these? looks like it was added to the json fixtures but nothing else?

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -184,6 +183,29 @@ func CreateV15(w http.ResponseWriter, r *http.Request) {
 	api.WriteRespAlertObj(w, r, tc.SuccessLevel, "Deliveryservice creation was successful.", []tc.DeliveryServiceNullableV15{*res})
 }
 
+// 	TODO allow users to post names (type, cdn, etc) and get the IDs from the names. This isn't trivial to do in a single query, without dynamically building the entire insert query, and ideally inserting would be one query. But it'd be much more convenient for users. Alternatively, remove IDs from the database entirely and use real candidate keys.
+func CreateV16(w http.ResponseWriter, r *http.Request) {
```
mhoppa: This should be v20. V16 is non existent

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -904,6 +987,7 @@ func readGetDeliveryServices(params map[string]string, tx *sqlx.Tx, user *auth.C
 		"logsEnabled":      dbhelpers.WhereColumnInfo{"ds.logs_enabled", api.IsBool},
 		"tenant":           dbhelpers.WhereColumnInfo{"ds.tenant_id", api.IsInt},
 		"signingAlgorithm": dbhelpers.WhereColumnInfo{"ds.signing_algorithm", nil},
+		"serviceCategory":  dbhelpers.WhereColumnInfo{"ds.service_category", api.IsInt},
```
mhoppa: this was added as a query parameter 👍 the docs should be updated with it

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -477,8 +484,9 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 
 		////DeliveryServices
 		{api.Version{2, 0}, http.MethodGet, `deliveryservices/?$`, api.ReadHandler(&deliveryservice.TODeliveryService{}), auth.PrivLevelReadOnly, Authenticated, nil, 2238317294, noPerlBypass},
-		{api.Version{2, 0}, http.MethodPost, `deliveryservices/?$`, deliveryservice.CreateV15, auth.PrivLevelOperations, Authenticated, nil, 206431432, noPerlBypass},
-		{api.Version{2, 0}, http.MethodPut, `deliveryservices/{id}/?$`, deliveryservice.UpdateV15, auth.PrivLevelOperations, Authenticated, nil, 2766567527, noPerlBypass},
+		{api.Version{2, 0}, http.MethodGet, `deliveryservices/{id}/?$`, api.ReadHandler(&deliveryservice.TODeliveryService{}), auth.PrivLevelReadOnly, Authenticated, nil, 244348195, noPerlBypass},
```
mhoppa: the deliveryservices/id got deprecated in 2.0.

Also the .json should be removed from regex on the routes

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -0,0 +1,124 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import (
+"strconv"
```
mhoppa: gofmt

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -0,0 +1,124 @@
...
+		"name": validation.Validate(serviceCategory.Name, validation.NotNil, validation.Required),
+	}
+	return util.JoinErrs(tovalidate.ToErrors(errs))
+}
+
+func (dv *TOServiceCategory) Create() (error, error, int) { return api.GenericCreate(dv) }
+func (dv *TOServiceCategory) Read() ([]interface{}, error, error, int) {
+	params := dv.APIInfo().Params
+	// TODO move to router, and do for all endpoints
+	if strings.HasSuffix(params["name"], ".json") {
```
mhoppa: huh?

jehunte: Fixing this - Realized it wasn't necessary when making the tenant changes.

traffic_ops/client/serviceCategory.go
jehunte: All of the other client files' functions are set up this way. Would you still like me to change it or keep it consistent with everything else?

traffic_ops/client/serviceCategory.go
jehunte: Can you be more specific on what you mean by url formatting?

traffic_ops/traffic_ops_golang/routing/routes.go
jehunte: How is the update done in 2.0 without deliveryservices/id? It looks like that path still exists.

mhoppa: oh I meant just the `GET` was deprecated in 2.0 for deliveryservices/id - the other verbs are still supported

traffic_ops/client/serviceCategory.go
mhoppa: Would you mind changing it ?

traffic_ops/client/serviceCategory.go
mhoppa: https://golang.org/pkg/net/url/#QueryEscape

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -333,6 +334,12 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{api.Version{2, 0}, http.MethodPost, `server_server_capabilities/?$`, api.CreateHandler(&server.TOServerServerCapability{}), auth.PrivLevelOperations, Authenticated, nil, 2293166834, noPerlBypass},
 		{api.Version{2, 0}, http.MethodDelete, `server_server_capabilities/?$`, api.DeleteHandler(&server.TOServerServerCapability{}), auth.PrivLevelOperations, Authenticated, nil, 2058714058, noPerlBypass},
 
+		//Service Categories: CRUD
+		{api.Version{2, 0}, http.MethodGet, `servicecategories/?$`, api.ReadHandler(&servicecategory.TOServiceCategory{}), auth.PrivLevelReadOnly, Authenticated, nil, 1085181543, noPerlBypass},
```
mitchell852: Before I forget, when adding new api routes, you need to make some changes to the seeds.sql file to account for the day when rather than enforcing privLevel on a route, we enforce a role's "capabilities":

1. create 2 new capabilities (service-categories-read and service-categories-write) here: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/db/seeds.sql#L93

2. add both capabilities to the admin and operations role:

https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/db/seeds.sql#L209
https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/db/seeds.sql#L326

3. add service-categories-read to the read-only role:

https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/db/seeds.sql#L285

4. Actually define which routes are in each of the 2 capabilities

https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/db/seeds.sql#L406

when this is done,  you will basically have granted the admin/operations role with the service-categories-read and service-categories-write capability and the read-only role with the service-categories-read capability.

traffic_portal/app/src/common/api/ServiceCategoryService.js
```diff
@@ -0,0 +1,88 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+var ServiceCategoryService = function($http, ENV, locationUtils, messageModel) {
+
+    this.getServiceCategories = function(queryParams) {
+        return $http.get(ENV.api['root'] + 'servicecategories', {params: queryParams}).then(
```
mattjackson220, line 23: please change these to be `ENV.api['latest']`.  once PR 4633 gets merged that will be API v3 and root will still be v2.  You can also hardcode in /api/3.0 if you want with a TODO to fix it later if PR4633 isnt merged as soon as we would like this to be

mattjackson220:
one small comment then this looks good! might depend on PR4633 to get TP set up to use API v3

lib/go-atscfg/headerrewritedotconfig.go
```diff
@@ -32,6 +34,8 @@ import (
 const HeaderRewritePrefix = "hdr_rw_"
 const ContentTypeHeaderRewriteDotConfig = ContentTypeTextASCII
 
+const ServiceCategoryHeader = "X-CDN-SVC"
```
rob05c, line 37: Would you mind changing this to `CDN-SVC`? The IETF has asked that people stop using `X-` for nonstandard headers, because they often become widespread, and  then it's difficult to make a new standard without the `X-`. Where if people don't  use `X-` to begin with, it's easy to make the header a standard when it becomes widespread.

rob05c: Source: https://tools.ietf.org/html/rfc6648

traffic_portal/app/src/common/api/ServiceCategoryService.js
```diff
@@ -20,7 +20,7 @@
 var ServiceCategoryService = function($http, ENV, locationUtils, messageModel) {
 
     this.getServiceCategories = function(queryParams) {
-        return $http.get(ENV.api['root'] + 'servicecategories', {params: queryParams}).then(
+        return $http.get(ENV.api['latest'] + 'servicecategories', {params: queryParams}).then(
```
mitchell852: https://github.com/apache/trafficcontrol/blob/master/traffic_portal/app/src/scripts/config.js

latest is now gone. you'll want to set these back to root.

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+		return tc.Alerts{}, reqInf, err
+	}
+	resp, remoteAddr, err := to.request(http.MethodPost, API_SERVICE_CATEGORIES, reqBody)
+	if err != nil {
+		return tc.Alerts{}, reqInf, err
+	}
+	defer resp.Body.Close()
+	var alerts tc.Alerts
+	err = json.NewDecoder(resp.Body).Decode(&alerts)
+	return alerts, reqInf, nil
```
ocket8888, line 48: I know you probably copied this from a different method, but please don't just ignore the error on the above line and just return `nil` here - we should return the error.

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// UpdateServiceCategoryByID updates a service category by ID
```
ocket8888, line 51: Same as above RE: complete sentence

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	route := fmt.Sprintf("%s/%d", API_SERVICE_CATEGORIES, id)
+	resp, remoteAddr, err := to.request(http.MethodPut, route, reqBody)
+	if err != nil {
+		return tc.Alerts{}, reqInf, err
+	}
+	defer resp.Body.Close()
+	var alerts tc.Alerts
+	err = json.NewDecoder(resp.Body).Decode(&alerts)
+	return alerts, reqInf, nil
```
ocket8888, line 68: same as above RE: returning vs ignoring errors

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// GetServiceCategories returns a list of service categories
```
ocket8888, line 71: Same as above RE: complete sentence.

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+	resp, remoteAddr, err := to.request(http.MethodGet, API_SERVICE_CATEGORIES, nil)
+	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
+	if err != nil {
+		return nil, reqInf, err
+	}
+	defer resp.Body.Close()
+
+	var data tc.ServiceCategoriesResponse
+	err = json.NewDecoder(resp.Body).Decode(&data)
+	return data.Response, reqInf, nil
```
ocket8888, line 82: Same as above RE: returning vs ignoring errors.

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// GetServiceCategoryByID gets a service category by the service category id
```
ocket8888, line 85: Same as above RE: complete  sentence

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+
+	var data tc.ServiceCategoriesResponse
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		return nil, reqInf, err
+	}
+
+	return data.Response, reqInf, nil
+}
+
+// GetServiceCategoryByName gets a service category by the service category name
```
ocket8888, line 103: Same as above RE: complete sentence

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+
+	var data tc.ServiceCategoriesResponse
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		return nil, reqInf, err
+	}
+
+	return data.Response, reqInf, nil
+}
+
+// DeleteServiceCategoryByID deletes a service category by service category id
```
ocket8888, line 121: Same as above RE: complete sentence

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+	route := fmt.Sprintf("%s/%d", API_SERVICE_CATEGORIES, id)
+	resp, remoteAddr, err := to.request(http.MethodDelete, route, nil)
+	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
+	if err != nil {
+		return tc.Alerts{}, reqInf, err
+	}
+	defer resp.Body.Close()
+	var alerts tc.Alerts
+	err = json.NewDecoder(resp.Body).Decode(&alerts)
+	return alerts, reqInf, nil
```
ocket8888, line 132: Same as above RE: returning vs ignoring errors

traffic_ops/client/serviceCategory.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	defer resp.Body.Close()
+
+	var data tc.ServiceCategoriesResponse
+	if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+		return nil, reqInf, err
+	}
+
+	return data.Response, reqInf, nil
+}
```
ocket8888, line 119: Per a recent mailing list discussion, this should instead just be one function that takes in optional `url.Values` (i.e. a `*url.Values` so that it can be `nil`) so that callers can filter as they please without needing to maintain multiple functions.

traffic_ops/testing/api/v3/servicecategories_test.go
```diff
@@ -0,0 +1,137 @@
...
+			t.Errorf("cannot UPDATE Service Category by id: %v - %v", err, alert)
+		}
+
+		// Retrieve the Service Category to check service category got updated
+		resp, _, err = TOSession.GetServiceCategoryByID(remoteServiceCategory.ID)
+		if err != nil {
+			t.Errorf("cannot GET Service Category by service category: %v - %v", firstServiceCategory.Name, err)
+		}
+
+		respServiceCategory := resp[0]
```
ocket8888, line 77: possible segfault if `len(resp)` is 0.

traffic_ops/testing/api/v3/servicecategories_test.go
```diff
@@ -0,0 +1,137 @@
...
+func UpdateTestServiceCategories(t *testing.T) {
+	firstServiceCategory := testData.ServiceCategories[0]
+	// Retrieve the Service Category by service category so we can get the id for the Update
+	resp, _, err := TOSession.GetServiceCategoryByName(firstServiceCategory.Name)
+	if err != nil {
+		t.Errorf("cannot GET Service Category by service category: %v - %v", firstServiceCategory.Name, err)
+	}
+	if len(resp) > 0 {
+		remoteServiceCategory := resp[0]
+		expectedServiceCategory := "service-category-test"
```
ocket8888, line 63: This is a bit of a nitpick, but this test could give false positives if the category's name was already "service-category-test". Instead it's generally better to concatenate something to the original value, so that you know for sure it's different no matter what it was initially.

traffic_ops/testing/api/v3/servicecategories_test.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, sc := range testData.ServiceCategories {
+		resp, _, err := TOSession.GetServiceCategoryByName(sc.Name)
+		if err != nil {
+			t.Errorf("cannot GET Service Category by name: %v - %v", err, resp)
+		}
+	}
+}
+
+func UpdateTestServiceCategories(t *testing.T) {
+	firstServiceCategory := testData.ServiceCategories[0]
```
ocket8888, line 55: this will segfault if the testdata has no service categories. Which _shouldn't_ be true, but all the same it's best not to make assumptions about what's in the test data.

traffic_ops/testing/api/v3/servicecategories_test.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		t.Error("tenant4user cannot GET service categories")
+	}
+
+	// assert that tenant4user cannot read service categories outside of its tenant
+	for _, sc := range serviceCategoriesReadableByTenant4 {
+		if sc.Name == "serviceCategory1" {
+			t.Error("expected tenant4 to be unable to read service categories from tenant 3")
+		}
+	}
```
ocket8888, line 109: This whole function relies on a specific structure for the test data. If I change something to do some other test, this will break even though the endpoint itself may work fine.

Instead, any specifically desired data structure should be set up and torn down by the test.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -184,6 +183,29 @@ func CreateV15(w http.ResponseWriter, r *http.Request) {
...
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("decoding: "+err.Error()), nil)
+		return
+	}
+
+	res, status, userErr, sysErr := createV30(w, r, inf, ds)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, status, userErr, sysErr)
+		return
+	}
+	api.WriteRespAlertObj(w, r, tc.SuccessLevel, "Deliveryservice creation was successful.", []tc.DeliveryServiceNullableV30{*res})
```
ocket8888, line 206: Per API guidelines this should use a `201 Created` status code, ideally setting a `Location` HTTP Header with a URL that points to the location from whence the new object may be fetched.

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/apache/trafficcontrol/lib/go-util"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/tenant"
+
+	validation "github.com/go-ozzo/ozzo-validation"
+)
+
+//we need a type alias to define functions on
```
ocket8888, line 40: GoDoc should start with the name of the object it describes and be a complete sentence with proper punctuation.

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -0,0 +1,199 @@
...
+type TOServiceCategory struct {
+	api.APIInfoImpl `json:"-"`
+	tc.ServiceCategoryNullable
+}
```
ocket8888, line 44: I just wanna say, I think using the "CRUDer" machinery is in general a bigger pain than it's worth. Just my two cents on that.

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (serviceCategory TOServiceCategory) GetType() string {
+	return "serviceCategory"
+}
+
+func (serviceCategory TOServiceCategory) Validate() error {
+	errs := validation.Errors{
+		"name":     validation.Validate(serviceCategory.Name, validation.NotNil, validation.Required),
+		"tenantId": validation.Validate(serviceCategory.TenantID, validation.Min(1)),
```
ocket8888, line 90: This isn't everything that needs to be validated here. Specifically, that tenant has to exist.

btw, I've been mostly holding off on mentioning things like this until it's fixed in the database, but there are some other places where this kind of thing needs to be checked.

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	return util.JoinErrs(tovalidate.ToErrors(errs))
+}
+
+func (serviceCategory *TOServiceCategory) Create() (error, error, int) {
+	return api.GenericCreate(serviceCategory)
```
ocket8888, line 96: Per API guidelines, this should return a `201 Created` status code, ideally with a `Location` HTTP Header set to a URL from whence the newly created object may be requested.

Note that this is impossible with the CRUDer as-is, but you could edit the CRUDer. Of course, that'd change it globally for a ton of endpoints across all API versions.

... aaand that's why I don't recommend using the CRUDer.

docs/source/api/v3/deliveryservices.rst
```diff
@@ -234,7 +238,9 @@ Response Structure
 		],
 		"maxOriginConnections": 0,
 		"ecsEnabled": false,
-		"rangeSliceBlockSize": null
+		"rangeSliceBlockSize": null,
+                "serviceCategory": null,
+                "serviceCategoryName": null
```
ocket8888, line 243: Use tabs for indentation.

Code blocks may contain space indents for their contents, but syntactically the indentation to associate the content with the code block must be tabs. So it's easier to just use tabs throughout.

docs/source/api/v3/deliveryservices.rst
```diff
@@ -31,37 +31,39 @@ Request Structure
...
+	| type            | no       | Return only :term:`Delivery Services` of the :term:`Delivery Service` :ref:`ds-types` identified by this integral, unique identifier    |
+	+-----------------+----------+-----------------------------------------------------------------------------------------------------------------------------------------+
+	| accessibleTo    | no       | Return the :term:`Delivery Services` accessible from a :term:`Tenant` *or it's children* identified by this integral, unique identifier |
+	+-----------------+----------+-----------------------------------------------------------------------------------------------------------------------------------------+
+    | serviceCategory | no       | Show only the :term:`Delivery Services` belonging to the :term:`Service Category` identified by this integral, unique identifier        |
+    +-----------------+----------+-----------------------------------------------------------------------------------------------------------------------------------------+
```
ocket8888, line 52: This table is malformed because you used spaces instead of tabs for indentation.

docs/source/api/v3/deliveryservices.rst
```diff
@@ -345,6 +352,7 @@ Request Structure
 		"rangeRequestHandling": 0,
 		"regionalGeoBlocking": false,
 		"routingName": "test",
+        	"serviceCategory": null,
```
ocket8888, line 355: same as above RE: indentation

docs/source/api/v3/deliveryservices.rst
```diff
@@ -507,6 +516,7 @@ Response Structure
 			"regionalGeoBlocking": false,
 			"remapText": null,
 			"routingName": "test",
+            		"serviceCategory": null,
```
ocket8888, line 519: Same as above RE: indentation

docs/source/api/v3/servicecategories.rst
```diff
@@ -0,0 +1,168 @@
...
+
+:Auth. Required: Yes
+:Roles Required: None
+:Response Type:  Array
+
+Request Structure
+-----------------
+.. table:: Request Query Parameters
+
+    +-----------+---------------------------------------------------------------------------------------------------------------+
```
ocket8888, line 35: This table is malformed because you used spaces for indentation instead of tabs

docs/source/api/v3/servicecategories.rst
```diff
@@ -0,0 +1,168 @@
...
+.. distributed under the License is distributed on an "AS IS" BASIS,
+.. WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+.. _to-api-service-categories:
+
+**********************
+``service categories``
```
ocket8888, line 19: This should be the name of the endpoint.

Also, from our API guidelines:

> _"Request paths MUST use 'snake_case' to separate words whenever necessary..."_

So really the route name ought to be `service_categories`.

docs/source/api/v3/servicecategories.rst
```diff
@@ -0,0 +1,168 @@
...
+            "response": [
+                {
+                    "id": 5,
+                    "lastUpdated": "2020-03-04 15:46:20-07",
+                    "name": "SERVICE_CATEGORY_NAME",
+                    "tenantId": 1,
+                    "tenant": "TENANT_NAME"
+                }
+            ]
+        }
```
ocket8888, line 100: Same as above RE: indentation

docs/source/api/v3/servicecategories.rst
```diff
@@ -0,0 +1,168 @@
...
+	Host: trafficops.infra.ciab.test
+	User-Agent: curl/7.47.0
+	Accept: */*
+	Cookie: mojolicious=...
+	Content-Length: 48
+	Content-Type: application/json
+
+	{
+		"name": "SERVICE_CATEGORY_NAME",
+                "tenantId": 1,
```
ocket8888, line 128: Same as above RE: indentation.

docs/source/api/v3/servicecategories.rst
```diff
@@ -0,0 +1,168 @@
...
+                }
+            ],
+            "response": {
+                "id": 1,
+                "lastUpdated": "2020-03-11 14:12:20-06",
+                "name": "SERVICE_CATEGORY_NAME",
+                "tenantId": 1,
+                "tenant": null
+            }
+        }
```
ocket8888, line 168: Same as above RE: indentation.

docs/source/overview/delivery_services.rst
```diff
@@ -697,6 +697,12 @@ Servers
 -------
 Servers can be assigned to Delivery Services using the :ref:`tp-configure-servers` and :ref:`tp-services-delivery-service` Traffic Portal sections, or by directly using the :ref:`to-api-deliveryserviceserver` endpoint. Only :term:`Edge-tier cache servers` can be assigned to a Delivery Service, and once they are so assigned they will begin to serve content for the Delivery Service (after updates are queued and then applied). Any servers assigned to a Delivery Service must also belong to the same CDN_ as the Delivery Service itself. At least one server must be assigned to a Delivery Service in order for it to serve any content.
 
+.. _ds-service-category:
+
+Service Category
+----------------
+Defines the type of content being delivered by the Delivery Service. Some example values are: "Linear", "VOD", and "OTT"
```
ocket8888, line 704: Paragraphs should consist of complete sentences (missing punctuation)

Also, I think this explanation could be clearer. This doesn't "define" the type of content because it has no semantic value. This is essentially just like a "tag" for the Delivery Service that more qualitatively "describes" the type of its content.

(also also: pls drop that last example as it's very Comcast-specific and nobody should need to know or even wonder what that could be)

traffic_portal/app/src/common/modules/table/serviceCategories/table.serviceCategories.tpl.html
```diff
@@ -0,0 +1,49 @@
...
+-->
+
+<div class="x_panel">
+    <div class="x_title">
+        <ol class="breadcrumb pull-left">
+            <li class="active">Service Categories</li>
+        </ol>
+        <div class="pull-right">
+            <button name="createServiceCategoryButton" class="btn btn-primary" title="Create Service Category" ng-click="createServiceCategory()"><i class="fa fa-plus"></i></button>
+            <button class="btn btn-default" title="Refresh" ng-click="refresh()"><i class="fa fa-refresh"></i></button>
```
ocket8888, line 27: Same as above RE: button type

traffic_portal/app/src/common/modules/table/serviceCategoryDeliveryServices/table.serviceCategoryDeliveryServices.tpl.html
```diff
@@ -0,0 +1,173 @@
...
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<div class="x_panel">
+    <div class="x_title">
+        <ol class="breadcrumb pull-left">
+            <li><a ng-click="navigateToPath('/servicecategories')">Service Categories</a></li>
+            <li><a ng-click="navigateToPath('/servicecategories/' + serviceCategory.id)">{{serviceCategory.name}}</a></li>
```
ocket8888, line 24: Same as above RE: links without `href`

docs/source/api/v3/deliveryservices.rst
```diff
@@ -234,7 +238,9 @@ Response Structure
 		],
 		"maxOriginConnections": 0,
 		"ecsEnabled": false,
-		"rangeSliceBlockSize": null
+		"rangeSliceBlockSize": null,
+                "serviceCategory": null,
```
mitchell852, line 242: sorry, small formatting

lib/go-tc/service_category.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// ServiceCategory ID
+	//
+	ID int `json:"id" db:"id"`
+
+	// LastUpdated
+	//
+	LastUpdated TimeNoMod `json:"lastUpdated" db:"last_updated"`
+
+	// Division Name
```
mitchell852, line 49: division?

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -228,6 +228,22 @@ <h3>Current Value</h3>
...
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(generalConfig.serviceCategory), 'has-feedback': hasError(generalConfig.serviceCategory)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear, ott, and vod.</div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
+                                <option hidden disabled selected value="">Select...</option>
```
mitchell852, line 238: see last comment

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.Steering.tpl.html
```diff
@@ -208,6 +208,22 @@ <h3>Current Value</h3>
...
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(generalConfig.serviceCategory), 'has-feedback': hasError(generalConfig.serviceCategory)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear, ott, and vod.</div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
+                                <option hidden disabled selected value="">Select...</option>
```
mitchell852, line 218: see last comment

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.anyMap.tpl.html
```diff
@@ -167,6 +167,22 @@ <h3>Current Value</h3>
...
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(generalConfig.serviceCategory), 'has-feedback': hasError(generalConfig.serviceCategory)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear, ott, and vod.</div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
+                                <option hidden disabled selected value="">Select...</option>
```
mitchell852, line 177: see last comment

traffic_portal/app/src/common/api/ServiceCategoryService.js
```diff
@@ -0,0 +1,88 @@
...
+
+    // todo: change to use query param when it is supported
+    this.deleteServiceCategory = function(id) {
+        return $http.delete(ENV.api['root'] + 'servicecategories/' + id).then(
+            function(result) {
+                messageModel.setMessages(result.data.alerts, true);
+                return result;
+            },
+            function(err) {
+                messageModel.setMessages(err.data.alerts, true);
```
mitchell852, line 79: this needs to change to false so any delete error can show up. i know....confusing.

traffic_portal/app/src/common/api/ServiceCategoryService.js
```diff
@@ -0,0 +1,88 @@
...
+                return result;
+            },
+            function(err) {
+                messageModel.setMessages(err.data.alerts, false);
+                throw err;
+            }
+        );
+    };
+
+    // todo: change to use query param when it is supported
```
mitchell852, line 58: probably don't need the todos if you designed the route to use route params instead of query params.

traffic_portal/app/src/common/modules/form/serviceCategory/form.serviceCategory.tpl.html
```diff
@@ -0,0 +1,58 @@
...
+                    <input id="name" name="name" type="text" class="form-control" ng-model="serviceCategory.name" required autofocus>
+                    <small class="input-error" ng-show="hasPropertyError(serviceCategoryForm.name, 'required')">Required</small>
+                    <span ng-show="hasError(serviceCategoryForm.name)" class="form-control-feedback"><i class="fa fa-times"></i></span>
+                </div>
+            </div>
+            <div class="form-group" ng-class="{'has-error': hasError(serviceCategoryForm.tenantId), 'has-feedback': hasError(serviceCategoryForm.tenantId)}">
+                <label class="control-label col-md-2 col-sm-2 col-xs-12">Tenant</label>
+                <div class="col-md-10 col-sm-10 col-xs-12">
+                    <select name="tenantId" class="form-control" ng-model="serviceCategory.tenantId" ng-options="tenant.id as tenantLabel(tenant) for tenant in tenants" required>
+                        <option value="">Select...</option>
```
mitchell852, line 46: this is where you could actually use that other syntax because this field is required so once something is selected it's ok for the Select... to disappear forever.

`<option hidden disabled selected value="">Select...</option>`

traffic_portal/app/src/common/modules/navigation/navigation.tpl.html
```diff
@@ -48,6 +48,7 @@
                         <li class="side-menu-category-item" ng-if="hasCapability('params-read')" ng-class="{'current-page': isState('trafficPortal.private.parameters')}"><a href="/#!/parameters">Parameters</a></li>
                         <li class="side-menu-category-item" ng-if="hasCapability('types-read')" ng-class="{'current-page': isState('trafficPortal.private.types')}"><a href="/#!/types">Types</a></li>
                         <li class="side-menu-category-item" ng-if="hasCapability('statuses-read')" ng-class="{'current-page': isState('trafficPortal.private.statuses')}"><a href="/#!/statuses">Statuses</a></li>
+                        <li class="side-menu-category-item" ng-if="hasCapability('service-categories-read')" ng-class="{'current-page': isState('trafficPortal.private.serviceCategories')}"><a href="/#!/servicecategories">Service Categories</a></li>
```
mitchell852, line 51: I'm kinda thinking this belongs under the Services section as the 3rd item. Plus the `Configure` section is getting lengthy.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -228,6 +228,22 @@ <h3>Current Value</h3>
                             </aside>
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(generalConfig.serviceCategory), 'has-feedback': hasError(generalConfig.serviceCategory)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear, ott, and vod.</div>
```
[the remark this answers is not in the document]
mitchell852: yeah, just ditch the ott

traffic_portal/app/src/common/modules/form/serviceCategory/form.serviceCategory.tpl.html
```diff
@@ -0,0 +1,58 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<div class="x_panel">
+    <div class="x_title">
+        <ol class="breadcrumb pull-left">
+            <li><a ng-click="navigateToPath('/servicecategories')">Service Categories</a></li>
```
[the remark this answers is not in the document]
mitchell852: @ocket8888 is right here but this is everywhere and probably requires a separate effort to fix that.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -420,6 +421,12 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{api.Version{3, 0}, http.MethodGet, `servers/{host_name}/update_status$`, server.GetServerUpdateStatusHandler, auth.PrivLevelReadOnly, Authenticated, nil, 2384515993, noPerlBypass},
 		{api.Version{3, 0}, http.MethodPost, `servers/{id-or-name}/update$`, server.UpdateHandler, auth.PrivLevelOperations, Authenticated, nil, 143813233, noPerlBypass},
 
+		//Service Categories: CRUD
+		{api.Version{3, 0}, http.MethodGet, `servicecategories/?$`, api.ReadHandler(&servicecategory.TOServiceCategory{}), auth.PrivLevelReadOnly, Authenticated, nil, 1085181543, noPerlBypass},
```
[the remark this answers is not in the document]
ocket8888: > _"...legacy..."_

That's why. New routes should follow the guidelines. Which I swear will be in the official docs someday soon.

docs/source/api/v2/deliveryservices.rst
```diff
@@ -31,37 +31,37 @@ Request Structure
...
+	| limit           | no       | Choose the maximum number of results to return                                                                                          |
+	+-----------------+----------+-----------------------------------------------------------------------------------------------------------------------------------------+
+	| offset          | no       | The number of results to skip before beginning to return results. Must use in conjunction with limit                                    |
+	+-----------------+----------+-----------------------------------------------------------------------------------------------------------------------------------------+
+	| page            | no       | Return the n\ :sup:`th` page of results, where "n" is the value of this parameter, pages are ``limit`` long and the first page is 1.    |
+	|                 |          | If ``offset`` was defined, this query parameter has no effect. ``limit`` must be defined to make use of ``page``.                       |
+	+-----------------+----------+-----------------------------------------------------------------------------------------------------------------------------------------+
```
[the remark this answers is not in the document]
jehunte: This feature was originally in v2 but then pushed back to v3. I just removed the content and forgot to revert the structure. 
 I will revert this file - There are no changes.

traffic_ops/testing/api/v3/servicecategories_test.go
jehunte: Line 61 there is a check for this `if len(resp) > 0`

traffic_portal/app/src/common/modules/form/serviceCategory/form.serviceCategory.tpl.html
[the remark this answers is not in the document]
ocket8888: Is it? I didn't realize. We should have an issue for that.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -228,6 +228,22 @@ <h3>Current Value</h3>
...
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear, ott, and vod.</div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
+                                <option hidden disabled selected value="">Select...</option>
+                            </select>
+                            <small ng-show="deliveryService.serviceCategory"><a href="/#!/servicecategories/{{deliveryService.serviceCategory}}" target="_blank">View Details&nbsp;&nbsp;<i class="fa fs-xs fa-external-link"></i></a></small>
```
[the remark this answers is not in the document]
jehunte: We have it after every View Details button in all of the form delivery service html files

traffic_portal/app/src/modules/private/serviceCategories/index.js
```diff
@@ -21,7 +21,7 @@ module.exports = angular.module('trafficPortal.private.serviceCategories', [])
     .config(function($stateProvider, $urlRouterProvider) {
         $stateProvider
             .state('trafficPortal.private.serviceCategories', {
-                url: 'servicecategories',
+                url: 'service_categories',
```
mitchell852, line 24: sorry, should be `service-categories` if you want to be consistent with the other TP urls. also, i guess it's a agreed upon web convention:

https://www.ecreativeim.com/blog/index.php/2011/03/30/seo-basics-hyphen-or-underscore-for-seo-urls/#:~:text=The%20short%20answer%20is%20that,same%20as%20redsneakers%20to%20Google.&text=Again%2C%20SEO%20URLs%20should%20use%20hyphens%20to%20separate%20words.

traffic_portal/app/src/common/modules/navigation/navigation.tpl.html
mitchell852: you want to make this change?

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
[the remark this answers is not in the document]
mitchell852: yeah, i think it's fine. it's a global issue can be fixed globally w/ another PR.

traffic_portal/app/src/common/modules/form/serviceCategory/form.serviceCategory.tpl.html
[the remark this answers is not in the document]
mitchell852: > Is it? I didn't realize. We should have an issue for that.

if we don't. want to create?

traffic_portal/app/src/common/modules/form/serviceCategory/form.serviceCategory.tpl.html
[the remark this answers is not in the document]
mitchell852: fixed. i provided the same feedback.

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
jehunte: This would require a large architecture change and I believe is out of scope for this feature.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -313,6 +313,7 @@ func createV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, reqDS t
 		&ds.RegionalGeoBlocking,
 		&ds.RemapText,
 		&ds.RoutingName,
+		&ds.ServiceCategoryId,
```
mattjackson220, line 316: this line should be moved to the end of the list to match where its at in your insertQuery()

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -822,6 +825,7 @@ func updateV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, reqDS *
...
```
mattjackson220, line 828: this line should be moved to the end of the list right above `&ds.ID` to match the updateDSQuery()

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (serviceCategory *TOServiceCategory) Read(h http.Header, useIMS bool) ([]interface{}, error, error, int, *time.Time) {
+	return api.GenericRead(h, serviceCategory, useIMS)
```
mattjackson220, line 110: looks like we lost the tenancy check when we used GenericRead

traffic_ops/app/db/migrations/20200727000000_add_deliveryservice_service_category.sql
```diff
@@ -0,0 +1,26 @@
...
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+CREATE TABLE IF NOT EXISTS service_category (
+                                                id BIGSERIAL PRIMARY KEY,
+                                                name TEXT UNIQUE NOT NULL CHECK (name <> ''),
```
ocket8888, line 17: If name is unique and not null, do these really need an id? I know this has been out a while, and I won't block the PR for this, but it seems like the whole concept could be a lot simpler by having only one unique key.

docs/source/api/v3/servicecategories.rst
```diff
@@ -0,0 +1,168 @@
...
+.. table:: Request Query Parameters
+
+    +-----------+---------------------------------------------------------------------------------------------------------------+
+    | Name      | Description                                                                                                   |
+    +===========+===============================================================================================================+
+    | id        | Filter for :term:`Service Categories` having this integral, unique identifier                                 |
+    +-----------+---------------------------------------------------------------------------------------------------------------+
+    | name      | Filter for :term:`Service Categories` with this name                                                          |
+    +-----------+---------------------------------------------------------------------------------------------------------------+
+    | tenant    | Return only :term:`Service Categories` belonging to the tenant identified by this integral, unique identifier |                                                   |
```
ocket8888, line 42: Table is malformed; pretty sure it's this line.

docs/source/api/v3/servicecategories.rst
```diff
@@ -0,0 +1,168 @@
...
+    {
+        "alerts": [
+            {
+                "text": "serviceCategory was created.",
+                "level": "success"
+            }
+        ],
+        "response": {
+            "id": 1,
+            "lastUpdated": "2020-03-11 14:12:20-06",\
```
ocket8888, line 163: backslash here is invalid JSON, breaks syntax highlighting

docs/source/api/v3/servicecategories.rst
```diff
@@ -0,0 +1,168 @@
...
+.. _to-api-service-categories:
+
+**********************
+``service_categories``
+**********************
+
+
+``GET``
+=======
+Get all requested :term:`Service Categories`.
```
ocket8888, line 25: This term isn't in the glossary - I think ideally you should add it to the glossary. The same error occurs 11 more times in this file, and once in the Delivery Service API pages.

lib/go-tc/deliveryservices.go
```diff
@@ -167,6 +167,8 @@ type DeliveryServiceNullable struct {
 	FirstHeaderRewrite *string `json:"firstHeaderRewrite" db:"first_header_rewrite"`
 	InnerHeaderRewrite *string `json:"innerHeaderRewrite" db:"inner_header_rewrite"`
 	LastHeaderRewrite  *string `json:"lastHeaderRewrite" db:"last_header_rewrite"`
+	ServiceCategoryId *int `json:"serviceCategory" db:"service_category"`
+	ServiceCategoryName *string `json:"serviceCategoryName"`
```
ocket8888, line 171: Needs go fmt

lib/go-tc/service_category.go
```diff
@@ -0,0 +1,40 @@
...
+	Response []ServiceCategory `json:"response"`
+}
+
+// ServiceCategoryResponse is a single Service Category response for Update and Create to
+// depict what changed.
+type ServiceCategoryResponse struct {
+	Response ServiceCategory `json:"response"`
+}
+
+// ServiceCategory ...
```
ocket8888, line 33: Would you mind making a better GoDoc here?

lib/go-tc/service_category.go
```diff
@@ -0,0 +1,40 @@
...
+type ServiceCategory struct {
+	ID int `json:"id" db:"id"`
+	LastUpdated TimeNoMod `json:"lastUpdated" db:"last_updated"`
+	Name string `json:"name" db:"name"`
+	TenantID int `json:"tenantId" db:"tenant_id"`
+	TenantName string `json:"tenant" db:"tenant"`
```
ocket8888, line 39: needs go fmt

traffic_ops/app/db/seeds.sql
```diff
@@ -175,6 +175,9 @@ insert into capability (name, description) values ('server-capabilities-write',
 -- servers
 insert into capability (name, description) values ('servers-read', 'Ability to view servers') ON CONFLICT (name) DO NOTHING;
 insert into capability (name, description) values ('servers-write', 'Ability to edit servers') ON CONFLICT (name) DO NOTHING;
+-- service categories
+insert into capability (name, description) values ('service-categories-read', 'Ability to view service categories') ON CONFLICT (name) DO NOTHING;
+insert into capability (name, description) values ('service-categories-write', 'Ability to edit service categories') ON CONFLICT (name) DO NOTHING;
```
ocket8888, line 180: I think you meant to insert service categories here?

traffic_ops/app/db/seeds.sql
```diff
@@ -313,6 +318,7 @@ INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHER
 INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHERE name = 'read-only'), 'roles-read' WHERE EXISTS (SELECT id FROM role WHERE name = 'read-only') ON CONFLICT DO NOTHING;
 INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHERE name = 'read-only'), 'server-capabilities-read' WHERE EXISTS (SELECT id FROM role WHERE name = 'read-only') ON CONFLICT DO NOTHING;
 INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHERE name = 'read-only'), 'servers-read' WHERE EXISTS (SELECT id FROM role WHERE name = 'read-only') ON CONFLICT DO NOTHING;
+INSERT INTO role_capability (role_id, cap_name) SELECT (SELECT id FROM role WHERE name = 'read-only'), 'service-categories-read' WHERE EXISTS (SELECT id FROM role WHERE name = 'operations') ON CONFLICT DO NOTHING;
```
ocket8888, line 321: Is this supposed to be checking the existence of the read-only role?

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -244,6 +244,22 @@ <h3>Current Value</h3>
...
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(generalConfig.serviceCategory), 'has-feedback': hasError(generalConfig.serviceCategory)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear and vod.</div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
+                                <option value="">Select...</option>
```
ocket8888, line 254: Since this isn't a valid option, you should maybe make it an `option[selected][hidden][disabled][value=""]`

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -244,6 +244,22 @@ <h3>Current Value</h3>
...
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(generalConfig.serviceCategory), 'has-feedback': hasError(generalConfig.serviceCategory)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear and vod.</div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
+                                <option value="">Select...</option>
```
ocket8888, line 254: Same as above RE disabled/hidden/selected

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.Steering.tpl.html
```diff
@@ -208,6 +208,22 @@ <h3>Current Value</h3>
...
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(generalConfig.serviceCategory), 'has-feedback': hasError(generalConfig.serviceCategory)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear and vod.</div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
+                                <option value="">Select...</option>
```
ocket8888, line 218: Same as above RE disabled/hidden/selected

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.anyMap.tpl.html
```diff
@@ -183,6 +183,22 @@ <h3>Current Value</h3>
...
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(generalConfig.serviceCategory), 'has-feedback': hasError(generalConfig.serviceCategory)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="serviceCategory">Service Category<div class="helptooltip">
+                                <div class="helptext">The type of content being delivered. Some examples are linear and vod.</div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
+                                <option value="">Select...</option>
```
ocket8888, line 193: Same as above RE disabled/hidden/selected

traffic_portal/app/src/common/modules/form/serviceCategory/form.serviceCategory.tpl.html
```diff
@@ -0,0 +1,58 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<div class="x_panel">
+    <div class="x_title">
+        <ol class="breadcrumb pull-left">
+            <li><a ng-click="navigateToPath('/service-categories')">Service Categories</a></li>
```
ocket8888, line 23: shouldn't this just be an `a[href="#!/service-categories"]?

jehunte: All other forms use navigateToPath and I couldn't get href working. Left as is.

traffic_ops/app/db/seeds.sql
jehunte: This is adding service category capability so it should be the capability table

traffic_ops/app/db/migrations/20200727000000_add_deliveryservice_service_category.sql
jehunte: I was unable to find bandwidth for this change. If you feel strongly about it, I can get it prioritized but it may take a little longer.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -251,7 +251,7 @@ <h3>Current Value</h3>
                         </label>
                         <div class="col-md-10 col-sm-10 col-xs-12">
                             <select id="serviceCategory" name="serviceCategory" class="form-control" ng-model="deliveryService.serviceCategory" ng-options="serviceCategory.id as serviceCategory.name for serviceCategory in serviceCategories">
-                                <option value="">Select...</option>
+                                <option [selected][hidden][disabled][value=""]>Select...</option>
```
ocket8888, line 254: Oh, sorry, I gave it to you as a CSS selector. `option[selected][hidden][disabled][value=""]` is a selector that matches `<option selected hidden disabled value="">Anything here</option>` - the <kbd>[</kbd><kbd>]</kbd> just indicate that it's matching an attribute; it shouldn't actually appear as a part of the attribute names.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -834,8 +841,8 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 
 		////DeliveryServices
 		{api.Version{2, 0}, http.MethodGet, `deliveryservices/?$`, api.ReadHandler(&deliveryservice.TODeliveryService{}), auth.PrivLevelReadOnly, Authenticated, nil, 2238317294, noPerlBypass},
-		{api.Version{2, 0}, http.MethodPost, `deliveryservices/?$`, deliveryservice.CreateV15, auth.PrivLevelOperations, Authenticated, nil, 206431432, noPerlBypass},
-		{api.Version{2, 0}, http.MethodPut, `deliveryservices/{id}/?$`, deliveryservice.UpdateV15, auth.PrivLevelOperations, Authenticated, nil, 2766567527, noPerlBypass},
+		{api.Version{2, 0}, http.MethodPost, `deliveryservices/?$`, deliveryservice.CreateV15, auth.PrivLevelOperations, Authenticated, nil, 506431433, noPerlBypass},
+		{api.Version{2, 0}, http.MethodPut, `deliveryservices/{id}/?$`, deliveryservice.UpdateV15, auth.PrivLevelOperations, Authenticated, nil, 1766567528, noPerlBypass},
```
ocket8888, line 845: I don't think you need to delete these route IDs

traffic_ops/app/db/migrations/20200805000000_add_deliveryservice_service_category.sql
```diff
@@ -0,0 +1,25 @@
...
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+CREATE TABLE IF NOT EXISTS service_category (
```
mitchell852, line 15: this needs to be renamed to `2020080500000000_add_deliveryservice_service_category.sql`(add 2 zeros) to ensure this is the last migration in the list.

ocket8888: I didn't think this would matter because the alphabetic sort order was unchanged, but it looks like goose first sorts by string length, then alphabetically - for some reason:

```psql
traffic_ops=> SELECT * FROM goose_db_version;
 id |    version_id    | is_applied |           tstamp           
----+------------------+------------+----------------------------
  1 |                0 | t          | 2020-08-06 23:49:38.325314
  2 |   20181206000000 | t          | 2020-08-06 23:49:38.331942
  3 |   20190219000000 | t          | 2020-08-06 23:49:38.333987
  4 |   20190319000000 | t          | 2020-08-06 23:49:38.335588
  5 |   20190513000000 | t          | 2020-08-06 23:49:38.36628
  6 |   20191004000000 | t          | 2020-08-06 23:49:38.375986
  7 |   20191005000000 | t          | 2020-08-06 23:49:38.384214
  8 |   20191024000000 | t          | 2020-08-06 23:49:38.39299
  9 |   20191215000000 | t          | 2020-08-06 23:49:38.401664
 10 |   20200218000000 | t          | 2020-08-06 23:49:38.432109
 11 |   20200227000000 | t          | 2020-08-06 23:49:38.437236
 12 |   20200313000000 | t          | 2020-08-06 23:49:38.439051
 13 |   20200408000000 | t          | 2020-08-06 23:49:38.516281
 14 |   20200422101648 | t          | 2020-08-06 23:49:38.524859
 15 |   20200507000000 | t          | 2020-08-06 23:49:38.569508
 16 |   20200805000000 | t          | 2020-08-06 23:49:38.588792
 17 | 2020061622101648 | t          | 2020-08-06 23:49:38.598169
 18 | 2020062923101648 | t          | 2020-08-06 23:49:38.600003
 19 | 2020072700000000 | t          | 2020-08-06 23:49:38.909138
(19 rows)
```

ocket8888:
I can't reproduce those test failures

mitchell852:
> I can't reproduce those test failures

mitchell852:
whoops. hit the wrong button. :)

mitchell852:
> ERROR: api.go:201: 2020-08-10T12:24:53.097693-06:00: 127.0.0.1:56366 dsr scanning: sql: Scan error on column index 10, name "deliveryservice": json: cannot unmarshal number into Go struct field DeliveryServiceNullable.serviceCategory of type string

nm. i think this is because my data was bad. service cat used to be an id (number) stored on a ds but it was changed to the name (string)

Issue #2500, opened 2018-07-03
astats providing bad interface data with virtio KVM driver on ATS servers


jhowell-comcast:
When using the virtio driver on a KVM / QEMU VM running ATS, astats reports the interface speed (inf.speed) as zero. This causes Traffic Monitor to do odd things and report available bandwidth as a negative number. This doesn't allow cache groups to come online and Traffic Router attempts to do DS_BYPASS as it thinks the ATS servers are maxed out on bandwidth. 

Using the e1000 driver in KVM / QEMU VM provides expected behaviors.

rawlinp:
👍 as a super hack workaround, you might be able to configure the minimum available bandwidth parameter in the TM profile as a negative number to "trick" TM into thinking there is enough bandwidth available. I _believe_ that was something I had to do when running ATS on a VM using the virtio driver.

Virtio reports the NIC speed as zero, and TM doesn't know how to handle that. I'm not sure _how_ it should handle that (ignore bandwidth availability? give it a default NIC speed?).

jhowell-comcast:
I tried the super hack workaround, it did enable TM to online the cache groups, but TR still tried to do DS_BYPASS. 
 
This is totally from running ATS on a KVM / QEMU VM with virtio driver. The virtio driver doesn't report zero for the interface speed. It reports nothing. astats is assigning a zero value to the variable. Perhaps a default nic speed if it's virtio that won't trigger tm to shut it down. 

It is easily gotten around by using e1000 driver in the VM, but you lose some performance from that. 

This isn't likely something that will ever be bumped into in a production instance, but somewhat likely in lab.

jhg03a:
If we could have a "default_bw" file or ats profile parameter, that could be used in the case where normal bandwidth detection methods can't function.

jhg03a:
Another option is to add a new astats.config parameter for this as well.

jhg03a:
Upon further thought it seems like either it's going to be a fix in astats with a new default_link_speed parameter in astats.config, or we simply remove bandwidth calculation from the TM health protocol in the case where the link speed cannot be determined.

The reason virtio doesn't have a link speed is because it's not a concept with that network driver.  It's simply shoving bits at the hypervisor as fast as possible which means sometimes you may have exceptionally fast transfers while other times they may be congested.  The e1000 driver still has the same problems but it establishes an upper bound to emulate physical hardware with a 1G interface instead of taking advantage of a deeper integration with the hypervisor.

[...]

PR #4521, opened 2020-03-18
Multi-interface Servers blueprint

ocket8888:

Adds a blueprint for a new feature - "Multi-Interface Servers".

- Documentation

Read the blueprint.

- [x] Tests are unnecessary


blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,657 @@
...
+	"iloIpNetmask": "",
+	"iloPassword": "",
+	"iloUsername": "",
+	"interfaces": [
+		{
+			"ipAddresses": [
+				{
+					"address": "172.16.239.120/24",
+					"gateway": "172.16.239.1",
+					"serviceAddress": true
```
mitchell852: just wondering. is `serviceAddress` applicable to all interfaces? for the interface that is included in the TR snapshot, this has meaning but does it have meaning for the other interfaces?

ocket8888: From a different point in the blueprint:

> _"A server object should be accepted (based only on its `interfaces` property) if it has at least one interface, every interface has at least one IP address, exactly one interface has at least one IP address that is a service address, and no two IP addresses of the same family/version are simultaneously marked as 'service addresses'."_

So to directly answer your question, only one interface on a server object is allowed to have ipAddresses that are "service addresses". The meaning that field has for other, non-"active" interfaces is that its value being `true` indicates an erroneous configuration, because only the "active" interface has "service addresses", and in fact the "active" interface, as you put it, is defined by having said "service addresses".

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,666 @@
...
+```
+Another new table will need to be created to contain address information, which
+is described below.
+```text
+                 Table "public.ip_address"
+     Column     |  Type   | Collation | Nullable | Default
+----------------+---------+-----------+----------+---------
+ address        | inet    |           | not null |
+ gateway        | inet    |           | not null |
+ serviceaddress | boolean |           | not null |
```
mitchell852: camel case?

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,666 @@
...
+}
+```
+Note that only interfaces that have "monitor" set to `true` should appear at all
+in the payloads of this endpoint. Further note that neither gateways nor
+<abbr title="Maximum Transmission Unit">MTU</abbr> are reported.
+
+##### `/cdns/{{name}}/snapshot`
+This is a semantic change only - payloads remain structurally unchanged. The
+reported `ip`, `ip6` and `interfaceName` fields for elements of the
+`contentServers` array property should be an arbitrarily chosen IPv4 service
```
mitchell852: arbitrarily?

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,666 @@
...
+controls for dynamically adding to them should be added.
+
+A sample screenshot is shown below.
+
+![](img/multi-interface-ui-sample.png "Mockup of proposed interface controls")
+
+If desired, the markup used for that sample may be found in
+[Appendix A](#appendix-a)
+
+Additionally, the properties of servers no longer present on a server object
```
mitchell852: It might be beneficial however to show server.interface.ipAddress.address where serviceAddress=true however in the servers table for each server. I could see that being a property you might want to see when viewing a collection of servers. Something to think about.

blueprints/multi-interface-servers.md
ocket8888: snake_case is what we typically use, because PostgreSQL (SQL standard?) case-folds everything you don't mark as an explicit identifier with double-quotes (<kbd>"</kbd>) to lowercase. Which is what happened here. That's just an example name, though, the only thing that's meant to specify is table structure, the column and table names are all arbitrary (provided you know what they're for).

blueprints/multi-interface-servers.md
ocket8888: It's not. The intention was to convey that a service-address-containing interface may or may not have an ipv4 or ipv6 address, and the mechanism for choosing one isn't specified in that section. But it needs to be worded better.

blueprints/multi-interface-servers.md
ocket8888: but you could have one or two of those - can you handle sorting by collection properties in TP tables?

ocket8888:
> ... changes to TM UI...

None are necessary. The TM UI has no special handling regarding _why_ a server is unavailable, it only knows if it that _is_ the case.

> ... TM cacheStats ...

idk what that is - the API endpoint that fetches it? I hadn't considered the TM API, might wanna look into that.

> ... how TM determines health based on stats of multiple interfaces ...

This may not be desirable, but the current impact description says:

> _"The cache server should be marked unavailable - or 'down' - if and only if the interface proving the service addresses is considered unhealthy."_

So it is addressed, at least.

blueprints/multi-interface-servers.md
mitchell852: gotcha and sorry, i meant snake. got my snakes and camels screwed up. :)

blueprints/multi-interface-servers.md
mitchell852: > but you could have one or two of those - can you handle sorting by collection properties in TP tables?

maybe?

mitchell852:
Could you add a blurb under the TP impact regarding the need to update the ISO generation form to handle multiple server interfaces when selecting the "Copy Server Attributes From" drop down?

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,666 @@
...
+The answer to this question should absolutely never be "no," can't imagine a
+world wherein that would need to change for this - or any other - particular
+feature.
+
+## Upgrade Impact
+There will be at least one, but possibly two, database migrations. The changes
+to Traffic Monitor ought to be made in such a way that it can parse either the
+old `/monitoring` payloads or the new ones. Then Traffic Monitor should be
+upgraded before Traffic Ops, and no additional, special steps need be taken to
+upgrade.
```
dneuman64: What's with the deployment order dependency here? Are we anticipating breaking the old interfaces such that we have to deploy TM first?  I usually think that we want to deploy TO first so I want to be sure to understand why that would change.

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,666 @@
...
+}
+```
+##### `/servers/details`
+This endpoint is subject to the same changes described above for
+`/servers`/`/servers/{{ID}}` - though obviously the full output objects will
+differ.
+
+#### Client Impact
+The structures output by various clients will change, but no client code changes
+should be necessary.
```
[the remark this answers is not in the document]
ocket8888: ish. The actual code in `traffic_ops/client` won't need to change. The structs all live under `/lib` so they can be shared between components.

blueprints/multi-interface-servers.md
ocket8888: Well, the way this will work is this: if these changes get into the 2.0 API release, you can upgrade TO first because TM will still be using the 1.x API.

And actually doing anything else would be a violation of semantic versioning, because we're removing fields not just adding them. So if this doesn't make it into API 2.0 it'd need to go in 3.0.

dneuman64: OK, so if we bump the API to 3.0 with this new API then we don't have an issue, right?

ocket8888: Right

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,670 @@
...
+	cachegroupId: number;
+	cdnId:        number;
+	cdnName?:     string | null;
+	domainName:   string;
+	guid?:        string | null;
+	hostName:     string;
+	httpsPort?:   number | null;
+	id?:          number;
+
+	// ILO things aren't being moved in because they also require a
```
jhg03a: We might consider changing the nomenclature around iLo to OOB as that's an HP specific term for an out-of-band management interface.

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,670 @@
...
+
+	/**
+	 * New field containing additional interfaces.
+	 * It is illegal for a server to not have at least one interface.
+	 */
+	interfaces: Array<Interface> & {0: Interface};
+	lastUpdated?:  string;
+
+	// notice not here: mgmtIpAddress, mgmtIpGateway, mgmtIpNetmask
+	// these will be moved into the interfaces property
```
jhg03a: Where?  Should there be an purpose field on an interface?

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,670 @@
...
+A new table will need to be introduced to contain interface information:
+`interface`, which is described below.
+```text
+              Table "public.interface"
+ Column  |  Type   | Collation | Nullable | Default
+---------+---------+-----------+----------+---------
+ name    | text    |           | not null |
+ server  | bigint  |           | not null |
+ monitor | boolean |           | not null |
+ mtu     | integer |           |          | 1500
```
jhg03a: if we're going to only support two values and null, why use an integer type over an enum or domain (with appropriate check constraint)?

blueprints/multi-interface-servers.md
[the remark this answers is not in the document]
jhg03a: Anyone not using one of the native clients will have to change their client code.  Maybe I'm misreading this though.

blueprints/multi-interface-servers.md
jhg03a: This would mean you could also model out-of-band management the same way as mgmtIpAddress etc.

blueprints/multi-interface-servers.md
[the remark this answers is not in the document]
ocket8888: This section is only about the client code maintained as a part of ATC. Any other clients will need to be aware of API changes across versions and adjust their code accordingly, just like always.

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,666 @@
...
+added).
+
+### Traffic Monitor Impact
+Traffic Monitor will need to update its parsing of `/monitoring` payloads to
+account for the new structure, and will need to poll all listed interfaces for
+health (instead of just one per cache server as was done previously).
+
+The cache
+server should be marked unavailable - or "down" - if and only if the interface
+proving the service addresses is considered unhealthy.
```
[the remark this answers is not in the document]
ocket8888: So interfaces now have `maxBandwidth` specified, but "stat aggregation changes" remain unaddressed.

blueprints/multi-interface-servers.md
ocket8888: I agree, but that's out of scope for this blueprint.

blueprints/multi-interface-servers.md
ocket8888: That's probably better, but to be honest I'm still not sure what MTU actually is or if 1500 and 9000 are the only values that ever can or will be possible for network interface devices. I was sort of hoping someone would call me out on that assertion if it was wrong.

blueprints/multi-interface-servers.md
ocket8888: It would need to be associated with the username and password information currently on the server object, which is why the decision was made to leave it be, as no other interface would ever use those fields.

ocket8888: As far as "purpose" fields... ¯\\\_(ツ)\_/¯

could be useful, but I worry that they'd fill up with mostly junk like "interface eth0". Maybe if it was optional, and definitely with a character limit. Like 60 characters, or something.

blueprints/multi-interface-servers.md
jhg03a: MTU is Maximum Transmission Unit and has to do with the size of Layer 2 Ethernet frames.  1500 is more or less the default, but if you can get 9000 beginning to end you should. https://en.wikipedia.org/wiki/Jumbo_frame

Super jumbo frames technically exist, but I've not run across them except in router configs.

blueprints/multi-interface-servers.md
jhg03a: How do you represent a server with two public and two private interface (used for different private networks)?

ocket8888: There's no distinction between "public" and "private" interfaces, if that's what your asking

blueprints/multi-interface-servers.md
ocket8888: There doesn't appear to be a spec for "super jumbo frames", but maybe I should add the "baby giant frames" value (1600).

Also, what're they gonna do when "super jumbo" *is* standardized and they want something even bigger? "Giga Frames"? "Maximum Over-Frames"?

dneuman64: There are many different values that you _could__ set for MTU.  1500 and 9000 are the most common, there's a 4500ish one that is pretty common too.  I  DO NOT think this should be an enum.

blueprints/multi-interface-servers.md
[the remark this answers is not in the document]
dneuman64: +100.  Maybe this is a good push to use the client we provide and maintain as part of the project?

blueprints/multi-interface-servers.md
dneuman64: You distinguish those using the "monitored" field.  If an interface is monitored it is a "public" interface, if it is not monitored it is a "private" interface.

blueprints/multi-interface-servers.md
ocket8888: Well then should we be restricting it at all? Maybe just >= 1500? Because currently TP will only let you set 1500 or 9000, which is why I chose those restrictions in the first place, but if it could theoretically be set to anything then maybe TP should just set a reasonable default but not restrict allowable values there, or in the API or database?

blueprints/multi-interface-servers.md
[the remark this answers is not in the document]
jhg03a: Disagree, but I guess my only real objection here is wording.  We use the term client when we mean atc supported sdk.  We need to always maintain compatibility with curl so that we can better triage breakage in the API, or support simple use cases in languages not supported by ATC.

blueprints/multi-interface-servers.md
rawlinp: Per https://www.ietf.org/rfc/rfc2460.txt we should probably constrain it to _at least_ 1280 or higher, but TP should probably still show 1500 and 9000 as the available options until we know we need different sizes.

blueprints/multi-interface-servers.md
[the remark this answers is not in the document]
ocket8888: curl support can't be broken unless we stop doing HTTP, tho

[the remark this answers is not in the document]
jhg03a: I'm referring to the promise that the API itself follows the API versioning rules, and not just our native clients (sdk).  Additionally, acknowledgement that unsupported clients leveraging the API directly must change as this will break the API going forward.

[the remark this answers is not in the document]
jhg03a: I'd be happy with something along the lines of adding:
```
Any unsupported 3rd-party client code will need to be updated to
leverage this new functionality in connection with these changes to core API/datamodel 
concepts going forward.
```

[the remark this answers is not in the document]
ocket8888: But that's implicit in the versioning process? This doesn't contain a proposal to break the versioning promise. If it doesn't make it into 2.0 before release it'll need to go into 3.0 because this is a breaking API change. Clients always need to check what's changed between major releases. A blueprint is a statement of intent to do some work on the project, nothing in it should be directed at nor even directly relevant to third parties. They can just look at the changelog.

blueprints/multi-interface-servers.md
ocket8888: I went with >1280

mitchell852:
Is there any impact to:

https://tp.domain.net/#!/cache-checks and the backing api endpoint(s)
https://tp.domain.net/#!/cache-stats and the backing api endpoint(s)

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,829 @@
...
+# Additional Server Interfaces
+
+## Problem Description
+Multiple network interfaces may be used by cache servers to service requests.
+However, these cannot be expressed as parts of the current "server" object.
+Specifically, assuming each interface has its own IP address, then each would
+need to be treated as a separate cache server. This means that getting clients
+to use one when the other is unavailable requires manual intervention.
+
+## Proposed Change
```
limited: Having a single shared IP address is one potential solution. I think just as likely would be different IPs per interface. If at all possible, it would be great if this feature could support that as well

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,829 @@
...
+```typescript
+interface IPAddress {
+	/**
+	 * The actual IP (v4 or v6) address which is used by an interface.
+	 * If it is an IPv6 address
+	 */
+	address: string;
+
+	/**
+	 * The netmask used by the address.
```
limited: is netmask needed?

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,829 @@
...
+
+	/**
+	 * The IP (v4 or v6) address of the gateway used by this IP address.
+	 */
+	gateway: string;
```
limited: gateway should be optional, not all IPs have an associated gateway.

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,829 @@
...
+
+	/**
+	 * Tells whether or not this address of this interface is the server's
+	 * "service" address.
+	 * At least one address of EXACTLY ONE interface MUST have this set to
+	 * 'true' for a server.
+	 */
+	serviceAddress: boolean;
```
limited: Why restrict it to exactly one serviceAddress per cache?

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,829 @@
...
+	ipAddresses: Array<IPAddress> & {0: IPAddress};
+	/**
+	 * The maximum allowed bandwidth for this interface to be considered "healthy"
+	 * by Traffic Monitor.
+	 *
+	 * This has no effect if `monitor` is not true.
+	 * Values are in kb/s.
+	 * The value `0` means "no limit".
+	 */
+	maxBandwidth: bigint;
```
limited: Will we continue to discover this from the interface stats?

blueprints/multi-interface-servers.md
```diff
@@ -0,0 +1,829 @@
...
+Another new table will need to be created to contain address information, which
+is described below.
+```text
+               Table "traffic_ops.ip_address"
+     Column     |  Type   | Collation | Nullable | Default
+----------------+---------+-----------+----------+---------
+ address        | inet    |           | not null |
+ gateway        | inet    |           | not null |
+ serviceaddress | boolean |           | not null |
+ server         | bigint  |           | not null |
```
limited: How do we guarantee consistency between interface.server and ip_address.server/ip_address.interface?

blueprints/multi-interface-servers.md
ocket8888: I don't think so. That comment was left in from an earlier iteration on accident.

blueprints/multi-interface-servers.md
ocket8888: "Optional" means something very specific in TypeScript. It means that the property can be missing from an object, and I'd prefer allowing things to be `null`. Just want to be clear what you're suggesting.

blueprints/multi-interface-servers.md
ocket8888: Because Traffic Router needs to be able to put the cache server in a hash ring.

blueprints/multi-interface-servers.md
ocket8888: Discover this? This will be set on the object when it's created.

blueprints/multi-interface-servers.md
ocket8888: With a foreign key constraint. The primary key of an interface is `(name, server)`, so `(interface, server)` is a foreign key into interfaces.

blueprints/multi-interface-servers.md
ocket8888: A "single shared IP address" is implemented by having a service address that is load-balanced to the other interfaces on the cache server side, and is the primary motivation behind this feature.

blueprints/multi-interface-servers.md
rawlinp: I could see supporting multiple service addresses per cache in the future, but I think we want to limit the scope of this feature to one service address per cache for the time being. Supporting that would mean we'd just have to drop that constraint and implement that in Traffic Router and ATS config (if necessary), right? So should need little API/DB rework.

blueprints/multi-interface-servers.md
jhg03a: That number hasn't been present in the database before as it's derived by the astats plugin to ATS based on which interface name you ask for.

jhg03a: There is a minor bug #2500 where it's impossible for astats to figure it out, and a gap where link aggregate member statuses aren't taken into account.

jhg03a: Even though it could help, I'm not a fan of moving this to TODB because it's making more state and more opportunity for humans to do it wrong.

ocket8888: Oh, I see. This isn't what you guys think it is. This is the equivalent of the `health.threshold.availableBandwidthInKbps` Parameter today.

dneuman64: Yes, this is being set in the database so that TM knows what the max_bw is per interface.  We will still use the `health.threshold.kbps` parameter for the overall cache bandwidth threshold.  The idea here is that TM needs to mark a cache unhealth if the overall max bw is reached OR the max bw for an interface is reached.

blueprints/multi-interface-servers.md
dneuman64: This is optional in the scope of database optionality.  I don't know what Typescript has to do with it.  In other words, this field should be nullable.

ocket8888: Because that's a typescript block.

But the root of the matter was I just wasn't sure if he was suggesting that
```json
{
  "address": "0.0.0.0",
  "gateway": null,
  "serviceAddress": false
}
```
be acceptable or
```json
{
  "address": "0.0.0.0",
  "serviceAddress": false
}
```
be acceptable. And I prefer the former.

blueprints/multi-interface-servers.md
jhg03a: So you want to move the concept of the bandwidth limit partially out of profiles/parameters.  It feels a bit odd to split the implementation like that, but ok.  Instead of fixed caps, could it be either percentage or fixed reservation?  That would allow us to fix the loophole with bonded interfaces in various link state mixes down the road.

blueprints/multi-interface-servers.md
ocket8888: Gateways are now allowed to be `null`

blueprints/multi-interface-servers.md
ocket8888: What loophole?

Also, I think adding support for percentages would entail a different body of work and ought to be considered separately from implementing the multiple-interface servers as outlined in this blueprint.

ocket8888:
> Is there any impact to:
> https://tp.domain.net/#!/cache-checks and the backing api endpoint - `GET /servercheck`
> https://tp.domain.net/#!/cache-stats and the backing api endpoint - `GET /caches/stats`
> And these api endpoints as well:
> `GET /cdns/health`
> `GET /cdns/{name}/health`
> `GET /deliveryservices/{id}/health`

- `/servercheck` - No. That endpoint does not interact with the Monitor at all.
- `/caches/stats` - This _does_ pull some stats from the Monitor, but the specific ones for which it's looking should be totally unaffected, as they are in relation to the operation of ATS and not the machine's network interfaces.
- `/cdns/health`/`/cdns/{{name}}/health` - The output of this is generated by tabulating which cache servers are "healthy". The determination of what "healthy" means will change, but not the ability to or structure of doing so. That is, the output value may change for a particular CDN may change for that endpoint, but the endpoint itself need not change.
- `/deliveryservices/{{ID}}/health` - Same as the CDN-based route.

blueprints/multi-interface-servers.md
ocket8888: Looking into that, it seems like this really ought to just be solved by having the stats system on the Cache Server read the interface's bonded directory and calculating the totals from its slave interfaces. Which seems like a bug with that/those plugin(s) that should be remedied there regardless of how many interfaces exist on cache server objects in TO.

Issue #3754, opened 2019-07-24
GET api/{version}/users/:userid/deliveryservices returns incorrect results


mitchell852:
With the advent of tenancy, a few things have changed and one of the more notable things is that users now have access to ALL delivery services contained within the scope of their tenancy and not just the delivery services that are explicitly assigned to the user.

However `GET api/{version}/users/:userid/deliveryservices` returns delivery services that are explicitly assigned to the user in the deliveryservice_tmuser table. This table is no longer relevant and should be removed but that's another issue (#2910).

`GET api/{version}/users/:userid/deliveryservices` should take into account the specified user's tenant and return all delivery services based on that tenant.

This bug manifests itself in TP at https://tp.domain.com/#!/users/95/delivery-services. Once the api is fixed, TP will display the correct results.

ocket8888:
I don't think this a bug. The docs for that endpoint say:

> _"Retrieves all Delivery Services assigned to the user."_

which is what it does - although those are filtered by tenancy. Changing that behavior would be a breaking change, although personally I don't have a problem with that. Just that it may need to be restricted to the 1.4 API.

And for the record, I do think the behavior of the endpoint as-is are no longer useful, but using this endpoint to return Delivery Services may not be intuitive, since it has `user` and `userId` in the path. What if instead the `tenant` query parameter of a request to `/deliveryservice` was expanded to include sub-tenants? Or if that's undesirable - which I could see - maybe add a parameter `accessibleTo` to name a tenant and return all Delivery Services accessible to that tenant?

mitchell852:
Honestly, I think the docs simply need to be changed to:

> _"Retrieves all Delivery Services assigned to the user via tenancy."_

I really don't think this falls under the category of an api breaking change. It's just an overall change of behavior in the system.

It used to be that a user's delivery services were determined by the contents of the deliveryservice_tmuser table but not it's determined by the user's tenancy.

By using GET /users/:id/deliveryservices the api consumer is just looking for a list of delivery services accessible by the user. how that is determined (deliveryservice_tmuser or tenancy) should not be a concern of the api consumer.

mitchell852:
however, if we're not going to fix `GET /users/:id/deliveryservices` i think:

1. it should be deprecated and
2. it's probably a good idea to support something like `GET /deliveryservices? accessibleTo=tenant1

mitchell852:
closing this in favor of #4402

PR #4528, opened 2020-03-23
Deprecate users/{userId}/deliveryServices

### LINKED ISSUE #4402: Provide the ability to fetch all delivery services accessible to a certain tenant
## I'm submitting a ...

- [ ] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
There is no current way to pull all delivery services accessible to a given tenant. I.e. imagine the following tenant tree:

```
- root
-- tenant_1
--- tenant_1.1
```

Currently, you can fetch all delivery services assigned to a tenant as such:

GET /deliveryservices?tenant=tenant_1

but you cannot fetch all delivery services "accessible to" tenant_1 which means all delivery services assigned to tenant_1 AND tenant_1.1

## Expected / new behavior:
Provide a query param to GET /deliveryservices that allows the user to fetch all delivery services accessible to a certain tenant. i.e. GET /deliveryservices?accessibleTo=tenant_1 which will provide all the delivery services assigned to tenant_1 OR tenant_1.1

## Anything else:
Once this is done, the following API route can be deprecated in 2.x:

GET users/{id}/deliveryservices

shamrickus:

- [X] This PR fixes #4402, deprecating `users/{id}/deliveryservices` and adding the `accessibleTo` query parameter to `deliveryservices` 

- Traffic Control Client 
- Traffic Ops
- Traffic Portal

Ensure docs build, TO tests pass, TP tests pass.

For TO:
- Verify `1.1/users/:id/deliveryservices` returns a deprecation notice (This endpoint does not work see #3754)
- Verify `2.0/users/:id/deliveryservices` doesn't exist
- Verify `2.0/deliveryservices?accessibleTo={tenantId}` works and returns all delivery services that have this tenant, or any delivery services that are children of the given tenant.

For TP:
- Verify when clicking the `View Delivery Services` button on the user editting screen (`users/{userId}`), it takes the user to `tenant/{tenantId}/deliveryservices`
- Verify that `tenant/{tenantId}/deliveryservices` returns the delivery services referencing this tenant, **and** this tenants children (previously does not show children).


traffic_portal/app/src/common/modules/form/user/FormUserController.js
```diff
@@ -50,7 +50,7 @@ var FormUserController = function(user, $scope, $location, formUtils, stringUtil
     $scope.labelize = stringUtils.labelize;
 
     $scope.viewDeliveryServices = function() {
-        $location.path($location.path() + '/delivery-services');
+        $location.path('/tenants/' + userModel.user.tenantId + '/delivery-services');
```
mitchell852: fyi for whoever reviews this. i was the one that suggested when viewing a user's delivery services, it's probably easiest to simply send them to view the delivery services of the user's tenant instead.

this does mean, however, that we are abandoning a view (users/id/deliveryservices) that should probably be cleaned up.

mitchell852: Here is a PR where @ocket8888 removed a view. In this case he removed the ability to view cachegroups for a parameter - https://github.com/apache/trafficcontrol/pull/3636

traffic_ops/testing/api/v2/deliveryservices_test.go
```diff
@@ -296,6 +298,53 @@ func UpdateDeliveryServiceWithInvalidSliceRangeRequest(t *testing.T) {
...
+	tenant = &resp.Response
+
+	//No delivery services are associated with this new tenant
+	err = getByTenants(tenant.ID, 0)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	//First and only child tenant, no access to root
+	err = getByTenants(3, len(testData.DeliveryServices) - 1)
```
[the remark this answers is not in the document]
shamrickus: We will, the test failures are from that. I was only running the individual tests so the id was always the same, but we can't rely on execution order (and therefore id)

traffic_portal/app/src/common/modules/form/user/FormUserController.js
mitchell852: So what this does is it directs you to the logged in user's tenant delivery services page every time which is not what you want. You want this to direct you to the user that was clicked's delivery services page. 

For example, if i'm on the this page: https://tp.domain.com/#!/users/244, then clicking "view delivery services" should direct me to `$location.path('/tenants/' + user[244].tenantId + '/delivery-services');`

mitchell852: actually, i think you an just get rid of the `userModel` part.....or you can pass user into the viewDeliveryServices function if you want to be more explicit.

PR #4531, opened 2020-03-23
Astats plugin - Add null check around strtok_r token usage

ezelkow1:
According to clang analyzer its possible that the token might be null, which you cannot pass to strtok_r as the save pointer (last argument).  Adding a null check before calling those

- Traffic Server

Compile the astats plugin and run with an ATS installation and ensure that stats are output properly based on the config file used (correct path and IP blocking, etc)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
   - /api/2.0/servercheck `GET`
 
 ### Changed
+- Add null check in astats plugin before calling strtok to find ip mask values in the config file
 - Fix to traffic_ops_ort.pl to strip specific comment lines before checking if a file has changed.  Also promoted a changed file message from DEBUG to ERROR for report mode.
 - Fixed Traffic Portal regenerating CDN DNSSEC keys with the wrong effective date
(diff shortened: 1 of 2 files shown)


ocket8888:
This is being done in a lot of other places in that file, is there a reason this line in particular is a problem?

ezelkow1:
> 
> 
> This is being done in a lot of other places in that file, is there a reason this line in particular is a problem?

Its because of the usage of tok1. So all the other usages of strtok_r in there are taking a single pointer for the save value, and then just iterating over strtok_r until it returns null.  However in these 2 instances they are re-using the tok1 value from previous strtok_r calls.  Strtok_r can return a null value if it doesnt find the string its looking for

So in these 2 instances its possible that the first call to strtok_r will have returned a null value, and then you are using an explicit null pointer as the save value

mitchell852:
@ezelkow1 - can you fix the description of this PR? Looks like this PR has an impact on every TC component which I'm guessing it does not. I thought we had a bullet for Traffic Server. maybe i'm wrong.

ezelkow1:
> @ezelkow1 - can you fix the description of this PR? Looks like this PR has an impact on every TC component which I'm guessing it does not. I thought we had a bullet for Traffic Server. maybe i'm wrong.

Ahh sure, didnt realize I needed to edit that list. I added Traffic Server, it wasnt in there

PR #4534, opened 2020-03-24
Fix ORT atstccfg to allow using new features in the latest Traffic Ops

rob05c:
## What does this Pull Request (PR) do?
Adds a client of the latest Traffic Ops version, and refactors the
existing vendored client, so both can be used at once.

Includes an idiom of using the latest and falling back to vendored.

Also includes an example of the latest client getting Delivery
Services, and falling back to the vendored if the TO isn't the latest.
This isn't actually necessary right now, but it's likely to be very
shortly, and it serves as an example of the idiom for future dev.

Includes unit tests. ORT doesn't have an integration test framework yet, so the larger part of this that requires TO can't be tested yet, unfortunately.
No docs, no changelog, no interface change.

- Traffic Ops ORT

Run ORT/atstccfg against a TO of the previous release version with warning logs enabled, verify the message "Traffic Ops newer than ORT, falling back to previous API Delivery Services". Run against a TO of the latest version, verify no message.

There aren't any new features that ORT uses yet, so that can't be verified yet. But there probably will be shortly.

Not a bug fix.

## Additional Information


rawlinp:
I'm confused as to why `atstccfg` needs to vendor an old client -- why can't we always stay up to date with master? If we upgrade TO first like we usually do, won't the latest `atstccfg` always be able to use the latest Go client version?

rob05c:
>If we upgrade TO first like we usually do

TC has always supported running any component with any other component one version back, ORT breaking that was a regression, and IMO we should continue to support that.

TC components should be loosely-coupled. There are innumerable disadvantages to tightly-coupled components. 

In fact, at our organization, we didn't used to have a strict order, and IMO we should never have adopted that deployment procedure. Requiring a strict upgrade order makes TC brittle and dangerous. It's much safer, robust, and resilient to require all components to support one version back of all the other components they talk to, so they can be upgraded or rolled back in any order.

Sometimes there are operational reasons one component needs updated first, and mandating an order makes that impossible. 

In the event a component unexpectedly needs to be rolled back, supporting one version back also allows that to be done quickly and safely. Without it, it would mean every component after it the "deployment order" has to be rolled back in sequence. Making an already tricky situation much more dangerous.

It also tightly couples our development, and makes it more difficult to develop each individual component. It means one person or team can't develop on one product, and implement a feature one component at a time with a clean interface. They have to be aware of all the ramifications of all the other components and their deployment order.

rawlinp:
Ok, I agree that tight coupling is bad, but this seems like something that should be well-communicated with the entire project for everyone to get on board with if our goal is to follow this pattern for every component. This seems more like a backwards-compatibility issue than a coupling issue. API clients will always be coupled to the API, and if a client is coupled to two versions of the same API, doesn't that make it even more tightly coupled to that API? It seems like the trade-off being made is really getting better backwards compatibility with TO at the cost of tighter coupling and increased complexity of components.

Maybe we're ok with that, but maybe we're ok with always upgrading TO first? The order doesn't necessarily need to be TO -> TM -> TR -> ORT for example, it could just be "TO first". If we need to fully roll back TO after we've already upgraded most of the other components in the chain, wouldn't we be risking more serious data loss if we rolled back TO at that point rather than hotfixing it and rolling forwards?

rob05c:
>This seems more like a backwards-compatibility issue than a coupling issue.

I think it's both.

>if a client is coupled to two versions of the same API, doesn't that make it even more tightly coupled to that API?

I guess everyone's definition is different, but that's not what I mean when I say "tightly-coupled." One of the benefits of an API is that you're "coupled" to the interface rather than the app itself. It seems inevitable that you have to communicate with some interface.

>better backwards compatibility with TO at the cost of tighter coupling and increased complexity of components.

The point of TC components supporting a version back isn't backward compatibility per-se, but rather the ability to deploy and roll back in any order, which gives us safety around deployment. It allows deployments of production components to be isolated around a single component of the system, rather than the full system all-or-nothing.

Getting 100,000 lines of code to work as a single, perfect, cohesive unit is way more difficult and error-prone, than getting ten 10,000-line apps to each independently work.

>If we need to fully roll back TO after we've already upgraded most of the other components in the chain, wouldn't we be risking more serious data loss if we rolled back TO at that point rather than hotfixing it and rolling forwards?

It depends on the situation. It's possible. It's also possible that a hotfix will be more dangerous, or take longer, or disrupt other systems, where a quick rollback of say ORT, or TM, would be faster and safer, and fix a broken production system in minutes instead of hours or days.

Or suppose it's a bug in TR, which requests TM, which requests TO. What if there's a bug in TM? Now you have to roll back TR, then TM, then TO. And maybe it's possible to hotpach TM. But maybe it's not, maybe it's a large, complex feature or code refactor and the only quick solution is to roll back. Now TR, TM, and TO all have to be rolled back, in order, and re-deployed, in order. 

Where if TM and TR support the prior version, TM can safely and quickly be rolled back.

That's why our components need to be decoupled. Not from the TO API, but from each-other, and from specific binaries. You're right, there are scenarios where it doesn't matter. But there are also a great many failure scenarios where it does, and it means the difference in minutes versus hours of outage time.

Letting TC users upgrade in any order can also be safer. In theory, TC shouldn't have bugs. But in the past, it often has, and upgrading certain components first can be safer and less likely to bring the CDN down.

>this seems like something that should be well-communicated with the entire project for everyone to get on board with

We did, but I guess that was before you came on to the project. We didn't used to do extensive documentation for these things, we just kind of agreed and then did it.

rob05c:
@ezelkow1 fyi I added a script to automate upgrading the vendored client and changing the TOClientNew usages to regular TOClient on a new release. I don't know if you feel the need to re-review.

It doesn't touch the existing code, and it's never run on config gen. It's just a helper script, if we find it useful.

rawlinp:
> We did, but I guess that was before you came on to the project. We didn't used to do extensive documentation for these things, we just kind of agreed and then did it.

Links to the mailing list discussions where decisions have supposedly been made are greatly appreciated if they're going to be used as an excuse for not wanting to open up the discussion to a wider audience. Even if the decision was made 3 years ago, that's a really long time, and a lot of developers who were here 3 years ago are no longer active on the project. We've got a lot of new developers who have no prior knowledge of these discussions, and changing the "TO client paradigm" _today_ (after discussing it 3+ years ago) would be something that would be a great discussion to have _now_, not only as a refresher for devs who were on the project 3+ years ago but also for teaching _newer_ devs. Since decisions like these have ramifications for the entire project, it's important that we not only get it right but also get everyone on board.

alficles:
@rawlinp Rob said:

> We can certainly bring it up again. I don't agree with removing the support in each component for the previous version of the components it talks to, but I do agree that things shouldn't be set in stone. If you feel strongly that we should, you can bring it up on the list and we can discuss it as a community again.

That sounds to me like he's game for revisiting the conversation on the list. I know especially back then, a lot of decisions were made in github tickets, conference rooms, and sometimes private spaces. The team was still getting the hang of the "Open" part of open source. (And honestly, could probably still use some improvement in that space still.)

You've got a really good point about the team members being very different now, as well as the project. It really sounds like this is a conversation ripe for the list.

rawlinp:
Yeah, that was not in the unedited version of his comment that I read in my email, and it came off very differently, though I doubt it was intended. All I'm asking is that the community be made aware of paradigm-shifting ideas like this so that other people can be involved before paradigm-shifting PRs get drive-by merged and end up setting the paradigm little input from other project members. Is this not a paradigm shift?

rob05c:
@rawlinp Can you please clarify: would you like to have a discussion on the list before considering this for merging?
We have a need for this at Comcast, and it's blocking further work. But I certainly don't want to ask someone to merge if you object. Thanks.

dneuman64:
We should absolutely avoid a deployment order at all costs.  I agree that this is not the most elegant solution, but it is the best we could come up with to allow us to be backwards compatible given the breaking changes made with the 2.0 API.  Hopefully this is just temporary and we can avoid this in future major API version changes.  I don't believe that this is a pattern that we want to or should adopt unless absolutely necessary.
As for the decision to not be tightly coupled and not have a deployment order, I don't think any discussion is required; this is a founding tenant of our project.  It's not about tenure on the project and how decisions are made, it is a core tenant that frankly we could be better at following.

rawlinp:
Maybe I overreacted when I saw this in the PR description:
> it serves as an example of the idiom for future dev.

To me that came across as "all of our Go components should follow this example in the future". Those kinds of wide-spread decisions that affect all developers in the project need to be communicated so that everyone can get on board. However, based on Neuman's comment, it sounds like this is _not_ supposed to be a decision like that since it's supposed to be a temporary solution just to get ORT to bridge the gap with the 2.0 API. Is that correct @rob05c?

Thinking about this more, this seems like an issue that would be permanent -- rather than just bridging the gap from 1.4 to 2.0, we will likely need a solution to bridge the gap from 2.0 to 2.1, 2.1 to 2.2, and so on, indefinitely.

For example, let's say ORT is using an unvendored TO Go client that's currently at 2.0. Traffic Ops is currently running 2.0 in Prod. A dev adds a new field for deliveryservices and bumps the version on master to 2.1. The TO Go client on master is then updated to request 2.1. A new release is cut from master, so now the new build of ORT requests 2.1 as well. However, Prod TO is still only running 2.0, and it will 501 any request for 2.1 -- therefore, ORT will not function if it's upgraded before TO is.

Using both a vendored and un-vendored version of the Go client _kind of_ solves this problem, but the problem with that is that ORT is _not_ vendoring the corresponding Go structs. So ORT is making 2.0 requests but unmarshalling them into 2.1 structs. Unless ORT also "sanitizes" the 2.1 fields to their default values before using them, this approach will lead to some odd bugs where ORT assumes the wrong default value. If this is going to be the solution going forward, indefinitely, then I think we might have to put some more thought into this approach.

If the correct 2.1 defaults are set on the resulting 2.1 struct when a request is made to 2.0 but unmarshalled into a 2.1 struct, I think that would be fine, but sanitizing defaults seems like something that the client itself should handle. In fact, maybe the TO Go client should optionally downgrade the request one minor version if it gets back a 501. For instance `GET /api/2.1/foos` returned a 501, so optionally retry one time with `GET /api/2.0/foos`. If the retry was successful, set the proper 2.1 defaults and return the result. However, it wouldn't try to downgrade across major versions from e.g. 2.0 to 1.4, since we can't always assume that would be safe.

Would that work going forward from minor version to minor version, instead of vendoring the TO Go client indefinitely?

rob05c:
>it serves as an example of the idiom for future dev.
>To me that came across as "all of our Go components should follow this example in the future".

That wasn't the intention.

>Also includes an example of the latest client getting Delivery
>Services, and falling back to the vendored if the TO isn't the latest.
>This isn't actually necessary right now, but it's likely to be very
>shortly, and it serves as an example of the idiom for future dev.

When I said "it serves as an example of the idiom for future dev" I was referring to the phrase "example of the latest client getting Delivery Services, and falling back to the vendored if the TO isn't the latest," specifically these lines:

```

            dses, unsupported, err := cfg.TOClientNew.GetCDNDeliveryServices(server.CDNID)
            if err == nil && unsupported {
                log.Warnln("Traffic Ops newer than ORT, falling back to previous API Delivery Services!")
                dses, err = cfg.TOClient.GetCDNDeliveryServices(server.CDNID)
            }
```

The sentence "it serves as an example of the idiom for future dev" wasn't intended as a broad sweeping declaration, but in context, for this very particular code pattern in ORT, as long as it exists. Or in other words, "I wrote something to show people how to do it, even though it isn't actually necessary yet."

>the problem with that is that ORT is not vendoring the corresponding Go structs

You are correct: if a developer uses a new feature in the latest version of Traffic Ops in ORT with a `TOClient` call and not a `TOClientNew` call, it will be a runtime error, not a compile-time error.

I agree, that isn't desirable. This solution isn't perfect.

You are also correct that the cause of this is ORT/atstccfg not vendoring the client structs.

The problem is, if we vendor the client structs, they become different symbols than used by `lib/go-atscfg`. Further, it's not possible to write conversion functions (e.g. `VendoredDSToDS()`) because Go doesn't allow naming vendored symbols. We would have to either convert every object in-place, with a huge amount of duplicate and illegible code, or else not use libraries for anything that uses `lib/go-tc`.

Again, this solution isn't perfect. There are a number of alternatives, each with their own disadvantages. If there's a perfect solution, I haven't seen it.

>maybe the TO Go client should optionally downgrade the request one minor version if it gets back a 501. For instance GET /api/2.1/foos returned a 501, so optionally retry one time with GET /api/2.0/foos. If the retry was successful, set the proper 2.1 defaults and return the result.

There are several disadvantages with that. It hides the behavior from apps using the client, and makes it impossible for them to discern between the latest and a client-assumed default. In cases where the client (ORT/TM/etc) needs to discern and act differently than the presumed-default, it can lead to the app doing the wrong thing, without even realizing it. And because the TO client hides it, it becomes impossible to do the right thing. It's also considerably more code in the client; likely more than it would take in actual apps using the client (e.g. ORT only uses a few functions/endpoints). I'm guessing most apps will also need to do their own handling for previous versions, in which case it hasn't saved any code for apps using the TO Client.

I'm not saying that solution is bad, just that there are disadvantages, as with this solution.

I think both this PR and your suggestion both have potential run-time errors. But this solution seems pretty easy for ORT developers to maintain. It seems unlikely that someone working on ORT won't be aware that the new feature they're adding is new, or if they don't, that they won't test and immediately discover it doesn't work. And this seems relatively easy for ORT developers to work with - it's a few lines to add new features, the extra code goes away as soon as the next release is made, and it's equally easy to refresh the vendored client (even without the helper script I added).

For the time being, would this solution be ok, just to make ORT work? And maybe we can rethink the client later?

rob05c:
>maybe the TO Go client should optionally downgrade the request

I'd also like to add, as someone who is frequently forced to use the TO client, a frequent pain point is the opacity and hiding of information from the user, trying to do the "right thing," which ends up making it difficult or impossible for the client's user to do what needs done. This isn't a new pain point, and it isn't exclusive to version things. It's bad enough, I've _frequently_ seriously considered making my apps just use http.Client instead, because it would be easier. But the biggest example of this is the API version. You can't tell the client to request a different API version, or even find out the API version it's using. In theory, this is the "right" solution, but in practice, it often makes it extremely painful to work with.

If we go with something along the lines of your suggestion, I think it would make it a lot easier to use if instead of automatically requesting different versions, we let the user decide. For example, we could have two clients, client and "traffic_ops/clientprev" which actually share all the logic code except the request functions. Or maybe the same client with e.g. GetDeliveryServices() and GetDeliveryServicesV14(), where the latter either actually returns a DeliveryServiceV14, or else a DeliveryService with any fields newer than V14 nil (the advantage of that is the symbol doesn't change, doesn't break compiles, and doesn't require updating client apps when the real symbol changes; the disadvantage is that users would have to be aware of it). Along with some kind of GetSupportedVersion(). Something along those lines. Speaking from experience, giving users control like that would be a lot easier to use, and help with a lot of the pain points of the client.

rob05c:
It does. Sorry, I guess I wasn't clear, when I said it includes an example, I meant it actually does it, as an example:
https://github.com/apache/trafficcontrol/pull/4534/files#diff-e51ad5f29db169bda046e91786532d5fR82

mhoppa:
oh - I build the binary and deployed in my local environment and not seeing that behavior.

Getting this in the access logs of TO 
```{{ip}} - admin [27/Mar/2020:22:11:27 +0000] "GET /api/1.4/deliveryservices?cdn=2 HTTP/1.1" 200 2696 13 "atstccfg/0.2"```

**Comment by rob05c** [comment]
Hm, it's supposed to, maybe there's a bug. I tested locally too, but maybe I missed something. I'll try again, and specifically try to test this new feature. 

If you run with warning logs (ORT does by default), do you see this message:
`log.Warnln("Traffic Ops newer than ORT, falling back to previous API Delivery Services!")` ?

**Comment by rawlinp** [comment]
@rob05c You bring up a good point about the client hiding things from the user -- things it already does today like attempting to re-login once when a 401 is received using `request()`. My idea is similar to that, except it wouldn't have to be a hidden behavior -- it could be optionally enabled when creating a new client instance but disabled by default. We should probably make re-login attempts optional as well. Regardless, I don't think that needs to be done in this PR. Really the entire Go client needs an overhaul.

Unmarshalling 1.4 requests into a 1.5 struct without sanitizing to 1.5 defaults does still kind of worry me, but that depends on whether or not the fields actually have defaults to sanitize to. For DSes going from 1.4 to 2.0, they probably don't need sanitized.

Anyways, might this be the problem that @mhoppa found? https://github.com/apache/trafficcontrol/pull/4534/files#diff-f0dfb4074818384cd25f1aeb52e7f0f4R42 That looks like it would make the "new" client still use the "old" (vendored) client, right?

**Comment by mhoppa** [comment]
I do not see the log message:

```Traffic Ops newer than ORT, falling back to previous API Delivery Services!```

**Comment by rawlinp** [comment]
It seems like we would have to  ~vendor~ copy the old TO client to a new package name in order to differentiate between the vendored and non-vendored client, right? Otherwise the imports will be exactly the same in `toreq` and `toreqnew`, and I guess the vendored client always wins?

**Comment by rob05c** [comment]
`/toreq` vendors the client itself, and `/toreqnew` doesn't. So `/toreqnew` shouldn't be using the vendored client.

Each package uses its own set of libraries, including vendoring. And the vendored library is treated as different, so they aren't compatible. I.e. if you tried to `cfg.TOClient.C = cfg.TOClientNew.C` it wouldn't compile, because their `traffic_ops/client.Session` symbols are different types.

**Comment by rob05c** [comment]
@mhoppa Ok, I just ran it with a master TO, and my TO logs have:
```
172.18.0.1:55408 - rbutts201 [27/Mar/2020:23:12:43 +0000] "GET /api/1.4/deliveryservices_required_capabilities? HTTP/1.1" 200 40 10 "atstccfg/0.2"
172.18.0.1:55480 - rbutts201 [27/Mar/2020:23:12:43 +0000] "GET /api/2.0/deliveryservices?cdn=2 HTTP/1.1" 200 52384 66 "atstccfg/0.2"
172.18.0.1:55446 - rbutts201 [27/Mar/2020:23:12:44 +0000] "GET /api/1.4/deliveryservices_regexes.json? HTTP/1.1" 200 9558 974 "atstccfg/0.2"
172.18.0.1:55434 - rbutts201 [27/Mar/2020:23:12:45 +0000] "GET /api/1.4/deliveryserviceserver?limit=999999 HTTP/1.1" 200 2493286 1907 "atstccfg/0.2"
```

Which is exactly what I'd expect, requesting 2.0 for the `toreqnew` call, and 1.4 for all the other `toreq` funcs.

Just to be sure, I also added a debug print in the master not-vendored `traffic_ops/client`, and saw it get printed:
```
ERROR: deliveryservice.go:164: 2020-03-27T17:12:43.522066-06:00: DEBUG master GetDeliveryServicesByCDNID path '/api/2.0/deliveryservices?cdn=2' apiBase '/api/2.0'
```

Is it possible the client in your master `traffic_ops/client` dir is still 1.4 somehow?

Alternatively, is it possible you have a vendored directory somewhere? Go will check for one all the way up, e.g. if there's one in `/` or `/traffic_ops` or `/traffic_ops/ort` or `/traffic_ops/ort/atstccfg` it'll use that.

In particular, this PR moves the vendor dir from `/traffic_ops/ort` to `toreq` - which, git moves are actually dones as delete and create - did the delete not work, is `/traffic_ops/ort/vendor` still there?

rob05c:
>things it already does today like attempting to re-login once when a 401 is received using request(). 

Yep, that's one of the pain points I've hit. And I'm certainly not blaming anyone--pretty sure I wrote that. And it's mostly what you need as a user. Until you don't.

>My idea is similar to that, except it wouldn't have to be a hidden behavior -- it could be optionally enabled when creating a new client instance but disabled by default. 

+1. In fact, if you have to have the funcs anyway, e.g. `GetDSPrev() DSV14`, `GetDSLatest() DS`, `GetDS() { if !GetDSPrev && cl.FallbackToPrev { GetDSLatest }`, if it's not any more code, I'd vote exposing those. Give users the option of always using the prev, if they don't need new features and feel it's safer. Even more client control, for people who want/need it.

>We should probably make re-login attempts optional as well. 

+1

>Regardless, I don't think that needs to be done in this PR. Really the entire Go client needs an overhaul.

+1. Again, as someone who frequently uses the client, I've put a lot of thought into it. And to be honest, I'm still not certain what the ideal, least-painful client looks like. But I agree it's worth trying to make it better.

>Unmarshalling 1.4 requests into a 1.5 struct without sanitizing to 1.5 defaults does still kind of worry me, but that depends on whether or not the fields actually have defaults to sanitize to. For DSes going from 1.4 to 2.0, they probably don't need sanitized.

I'd vote we make all new fields default nil, and require that all TC apps handle that. That seems safest anyway. Doing otherwise would require hackery to make upgrading work at all anyway.

rawlinp:
> Each package uses its own set of libraries, including vendoring.

Oh I see, you moved the vendor dir down into the `toreqnew`. That makes sense. FWIW I just built and tried this locally too, and I see `atstccfg` requesting 2.0 for `/deliveryservices?cdn=` but 1.4 for everything else. So seems to be working as expected for me at least.

mhoppa:
btw the issues I was running into was using the incorrect URL for TO_URL it just needs to be the base url and I was using the old format causing issues. Thanks to @rob05c for helping me through it

PR #4535, opened 2020-03-24
Fix v2 tests in trafficops integration container

mhoppa:

- CDN in a Box

Build integration container and make sure it runs with both test suites passing

master


zrhoffman:
Yep looks good.

- v1 AND v2 API tests pass!
- Unit tests pass (not that there is any reason to think they would fail, but every PR should test this!)
- None of files that `go fmt` caught are related to this PR

PR #4536, opened 2020-03-24
API Guidelines blueprint

ocket8888:

Adds a blueprint for some API guidelines.

- Documentation

Read the blueprint. Maybe pass it through a Markdown parser/validator/renderer if you want.

- [x] Tests are unnecessary

### DIFF
--- a/blueprints/api.guidelines.md
+++ b/blueprints/api.guidelines.md
@@ -138,7 +138,7 @@ results that could possibly be returned given the non-pagination query
 parameters supplied by the client.
 
 ### HTTP Request Methods
-[RFC 2616 Section 9](https://tools.ietf.org/html/rfc2616#section-9) defines the
+[RFC 7231 Section 4](https://tools.ietf.org/html/rfc7231#section-4) defines the
 semantics of HTTP/1.1 request methods. Authors should conform to that set of
 standards whenever possible, but for convenience the methods recognized by
 Traffic Ops and their meanings in that context are herein defined.
--- a/blueprints/api.guidelines.md
+++ b/blueprints/api.guidelines.md
@@ -1,0 +1,535 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# API Guidelines
+
+## Problem Description
+The Traffic Ops API duplicates a lot of functionality, is at times
+self-inconsistent, and can be difficult to work with.
+
+## Proposed Changes
+Many manifestations of the above problems could be avoided entirely by adhering
+to a set of guidelines for how our API ought to behave. That would certainly
+make it more consistent, and would make the intended behavior much easier to
+discern. Herein lie a proposed set of API guidelines that have been discussed
+by the Traffic Ops working group over the past few months.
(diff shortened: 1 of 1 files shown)


blueprints/api.guidelines.md
```diff
@@ -0,0 +1,535 @@
...
+from being fulfilled via one or more `error`-level alerts.
+
+#### 500 Internal Server Error
+When the Traffic Ops server encounters some error - through no fault of the
+client or their request - that renders it incapable of servicing the client's
+request, it MUST return this status code if no other code is more appropriate.
+The response body in this case SHOULD indicate that an error occurred, but
+MUST NOT divulge details about what data was being processed, what (if any)
+other components are not functioning properly, or what process failed. Generally
+it is advisable that the resultant \code{alerts} array entry just say "Internal
```
mitchell852: do you need to state that the details of the error (if known) should be written to TO logs instead?

blueprints/api.guidelines.md
```diff
@@ -0,0 +1,535 @@
...
+Server Error" and nothing else.
+
+#### 501 Not Implemented}
```
mitchell852: curly

blueprints/api.guidelines.md
```diff
@@ -0,0 +1,535 @@
...
+to be passed in the query string to separate it from the object definition.
+- The following query parameters are reserved for special use by Traffic Ops
+endpoint handlers, and may not be used for any purpose other than their
+prescribed functions.
+
+	- `limit`
+	- `newerThan`
+	- `offset`
+	- `olderThan`
+	- `orderby`
```
mitchell852: should this be `orderBy`?

blueprints/api.guidelines.md
```diff
@@ -0,0 +1,535 @@
...
+DELETE destroys an object stored on the server. Typically the request will
+contain identifying information for the object(s) to be destroyed either in the
+request URI or in the request's body. Traffic Ops API endpoint authors MUST use
+this request method whenever an object identified by the request URI is being
+destroyed. When such deletion successfully occurs, the response body MUST
+contain a representation of the destroyed object.
+
+### HTTP Response Codes
+Proper use of HTTP response codes can significantly improve user interfaces
+built on top of the API. What follows is a (non-exhaustive) set of response
```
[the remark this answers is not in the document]
ocket8888: Nah, I don't want to limit the API's growth. A new response code could be standardized at any time, so even if these are sufficient out of the ones that exist today, they aren't representative of all possible response codes for all time.

blueprints/api.guidelines.md
```diff
@@ -0,0 +1,535 @@
...
+access to the resource they are requesting. It MUST also be used when they have
+some access to the resource, but not with the specific request method they used.
+This can pertain to restricted access on the basis of Role, User Permissions, as
+well as Tenancy.
+
+The response body MUST NOT disclose any information regarding why the user was
+denied access.
+
+#### 404 Not Found
+This MUST be the returned status code when the client requests a path that does
```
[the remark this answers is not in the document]
ocket8888: I suppose I can, though I think that's part of the definition of a "path" part of a URL.

blueprints/api.guidelines.md
ocket8888: nope. Our current API uses "orderby," unfortunately. Changing that would be a breaking change unless they both worked - and I don't think anyone wants to bother with that. I know I don't. So we'll save consistent casing for v3 (or even v4 maybe, at this point).

blueprints/api.guidelines.md
ocket8888: I would hope not. This is just about the API itself.

blueprints/api.guidelines.md
```diff
@@ -0,0 +1,549 @@
...
+Endpoints MUST use the following, reserved properties of `summary` for their
+described purposes (when use of `summary` is appropriate) rather than defining
+new `summary` or `response` properties to suit the same purpose:
+
+- `count` - Count contains an unsigned integer that defines the total number of
+results that could possibly be returned given the non-pagination query
+parameters supplied by the client.
+
+### HTTP Request Methods
+[RFC 2616 Section 9](https://tools.ietf.org/html/rfc2616#section-9) defines the
```
rob05c, line 141: RFC 2616 is obsolete, should be RFC 7231 Section 4: https://tools.ietf.org/html/rfc7231#section-4

PR #4537, opened 2020-03-25
Add blueprint for Flexible Topologies

rawlinp:
Adds a blueprint for Flexible Topologies.

- Documentation

Read it!

- [x] blueprint -- no tests necessary
- [x] blueprint -- no changelog necessary

### DIFF
--- a/blueprints/flexible-topologies.md
+++ b/blueprints/flexible-topologies.md
@@ -85,7 +85,7 @@ The following table describes the `node` sub-object:
 | cachegroup | string            | required    | the `name` of a cachegroup this node maps to in the Topology                                                                                                                                                |
 | parents    | array of integers | required    | zero-based indexes to other nodes in the Topology's `nodes` array, where the 1st element is for the *primary* parent relationship and the 2nd element is for the *secondary* parent relationship, and so on |
 
-API constraints:
+**API constraints:**
 - a Topology's `name` must consist of alphanumeric or hyphen characters
 - a Topology must have at least 1 `node`; otherwise, it is useless
 - there cannot be multiple `nodes` for the same cachegroup in a Topology
@@ -250,11 +250,29 @@ response JSON:
 
 ##### `/deliveryservices` endpoints
 
-All relevant Delivery Service APIs will have their JSON request and response objects modified to include a new `topology` field which references the name of the topology it's assigned to:
-```JSON
+All relevant Delivery Service APIs will have their JSON request and response objects modified to include the following new fields:
+- `topology` - the name of the topology it's assigned to
+- `firstHeaderRewrite` - the header_rewrite ATS config that will be applied to the *first tier* caches in the Delivery Service's Topology
+- `middleHeaderRewrite` - the header_rewrite ATS config that will be applied to all *middle tier* caches in this Delivery Service's Topology
+- `lastHeaderRewrite` - the header_rewrite ATS config that will be applied to all *last tier* caches in this Delivery Service's Topology
+
+**Note:** these three header_rewrite "buckets" were chosen because header_rewrite configs usually fall into one of three categories:
+- affects requests to/from the client ("first")
+- affects requests to/from the origin ("last")
+- don't affect clients or origins ("middle")
+
+**API Constraints:**
+- `firstHeaderRewrite`, `middleHeaderRewrite`, and `lastHeaderRewrite` can only be set if `topology` is not null
+- `edgeHeaderRewrite` and `midHeaderRewrite` cannot be set if `topology` is not null
+
+**Example JSON:**
+```JSONC
 {
-    ...
-    "topology": "foo"
+    // ...
+    "topology": "foo",
+    "firstHeaderRewrite": "foo",
+    "middleHeaderRewrite": "foo",
+    "lastHeaderRewrite": "foo"
 }
 ```
 
@@ -333,7 +351,11 @@ New structs will be added for the `/topologies` endpoints, mapping directly to t
 - `Topology`: for the top-level Topology object, which includes the `nodes` array
 - `TopologyNode`: for objects in a Topology's `nodes` array
 
-The `DeliveryService` struct will be updated with a new `Topology` field which is the name of the Topology the Delivery Service is assigned to.
+The `DeliveryService` struct will be updated with new fields, mapping directly to the new JSON fields outlined in the REST API Impact section:
+- `Topology`
+- `FirstHeaderRewrite`
+- `MiddleHeaderRewrite`
+- `LastHeaderRewrite`
 
 ##### Traffic Ops Database
 
@@ -346,11 +368,11 @@ A new `topology` table will be created:
 
 A new `topology_cachegroup` table will be created to model the association of cachegroups to topologies:
 
-| column     | type | modifiers                                       |
-| ---------- | ---- | ----------------------------------------------- |
-| id         | int  | not null, PK                                    |
-| topology   | text | not null, FK: references topology(name)         |
-| cachegroup | text | not null, FK: references cachegroup(name)       |
+| column     | type | modifiers                                 |
+| ---------- | ---- | ----------------------------------------- |
+| id         | int  | not null, PK                              |
+| topology   | text | not null, FK: references topology(name)   |
+| cachegroup | text | not null, FK: references cachegroup(name) |
 
 **Constraints**:
 - unique (topology, cachegroup) -- a cachegroup can only be in a Topology once.
@@ -368,11 +390,14 @@ A new `topology_cachegroup_parents` table will be created to model the parent re
 - unique (child, parent) -- within a Topology, a cachegroup cannot relate to another cachegroup more than once.
 - check (rank is either 1 or 2) -- a cachegroup can only have primary and secondary parents currently.
 
-The `deliveryservice` table will be updated to add a new column for the topology it is associated to:
+The `deliveryservice` table will be updated to add the following new columns:
 
-| column   | type | modifiers                     |
-| -------- | ---- | ----------------------------- |
-| topology | text | FK: references topology(name) |
+| column                | type | modifiers                     |
+| --------------------- | ---- | ----------------------------- |
+| topology              | text | FK: references topology(name) |
+| first_header_rewrite  | text |                               |
+| middle_header_rewrite | text |                               |
+| last_header_rewrite   | text |                               |
 
 ### ORT Impact
 
@@ -384,7 +409,9 @@ Since new Topologies can be more than 2 tiers (`EDGE` -> `MID`), `atstccfg` may 
 
 #### ATS config-related fields on Delivery Services
 
-Edge header rewrite rules and raw remap rules should still be applied only to the `EDGE` tier of Topologies. Mid header rewrite rules should probably only be applied at the *first* `MID` tier if there are multiple `MID` tiers in a Topology, because it might not be safe to assume that all mid header rewrite rules could be applied safely through multiple `MID` tiers. This does bring up the question of whether or not we also want to be able to apply header rewrites to any arbitrary "tier" of a Topology. That might not be a requirement for this MVP, but it may be something we'll need to add in the future when a use case arises.
+*Topology-based* delivery services will use the new `firstHeaderRewrite`, `middleHeaderRewrite`, and `lastHeaderRewrite` fields (described in the REST API Impact section above) for adding ATS header_rewrite configs. If a cache is both a "first" and a "last" in a given Topology (for instance, in a single-tier, edge-only Topology), both the `firstHeaderRewrite` and `lastHeaderRewrite` should be considered for adding. Otherwise, a given cache will be either a "first", "middle", or a "last".
+
+*Legacy* delivery services (not assigned to a Topology) will continue to use the `edgeHeaderRewrite` and `midHeaderRewrite` fields as they do today.
 
 ### Traffic Monitor Impact
 
@@ -447,7 +474,7 @@ Creating or updating Topologies should be restricted to users with the `operatio
 
 ### Upgrade Impact
 
-This feature will require a database migration to create new tables and add a new field to the `deliveryservice` table, but existing data does not need to be modified or migrated. Therefore, rolling back the database migration will not cause any data loss until Topologies are actually created and assigned. Additionally, since this feature will not remove any existing tables or columns, the new database schema should be backwards-compatible with the previous version of Traffic Ops. However, this blueprint cannot make any claim as to the backwards-compatibility of schema changes required by *other* features, which would affect the overall backwards-compatibility of the entire release.
+This feature will require a database migration to create new tables and add new fields to the `deliveryservice` table, but existing data does not need to be modified or migrated. Therefore, rolling back the database migration will not cause any data loss until Topologies are actually created and assigned. Additionally, since this feature will not remove any existing tables or columns, the new database schema should be backwards-compatible with the previous version of Traffic Ops. However, this blueprint cannot make any claim as to the backwards-compatibility of schema changes required by *other* features, which would affect the overall backwards-compatibility of the entire release.
 
 This feature will not require components to be upgraded in a specific order, and no special manual steps will be required before, during, or after the upgrade is done.
 
--- a/blueprints/flexible-topologies.md
+++ b/blueprints/flexible-topologies.md
@@ -95,6 +95,7 @@ API constraints:
 - a Topology cannot contain a cycle (through any combination of primary/secondary parent relationships)
 - a Topology cannot be deleted if one or more Delivery Services are still assigned to it
 - a Cachegroup cannot be deleted if it is currently being used in a Topology
+- a Topology cannot have `STEERING` or `CLIENT_STEERING` delivery services assigned to it (because those types are not assigned to caches -- their _targets_ are)
 
 The following new endpoints will be required:
 
--- a/blueprints/flexible-topologies.md
+++ b/blueprints/flexible-topologies.md
@@ -82,7 +82,7 @@ The following table describes the `node` sub-object:
 
 | field      | type              | optionality | description                                                                                                                                                                                                 |
 | ---------- | ----------------- | ----------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| cachegroup | string            | required    | the `short_name` of a cachegroup this node maps to in the Topology                                                                                                                                          |
+| cachegroup | string            | required    | the `name` of a cachegroup this node maps to in the Topology                                                                                                                                          |
 | parents    | array of integers | required    | zero-based indexes to other nodes in the Topology's `nodes` array, where the 1st element is for the *primary* parent relationship and the 2nd element is for the *secondary* parent relationship, and so on |
 
 API constraints:
@@ -266,7 +266,7 @@ The `GET /deliveryservices` endpoint should be updated to support `?topology=foo
 
 `DELETE /cachegroups` should be updated to return a useful error when trying to delete a Cachegroup that is currently used in a Topology.
 
-`PUT /cachegroups` should update all foreign key references if `short_name` is updated. This should be done automatically in the database via the FK references.
+`PUT /cachegroups` should update all foreign key references if `name` is updated. This should be done automatically in the database via the FK references.
 
 ##### The various `/snapshot` endpoints
 
@@ -307,7 +307,7 @@ A new `topology_cachegroup` table will be created to model the association of ca
 | ---------- | ---- | ----------------------------------------------- |
 | id         | int  | not null, PK                                    |
 | topology   | text | not null, FK: references topology(name)         |
-| cachegroup | text | not null, FK: references cachegroup(short_name) |
+| cachegroup | text | not null, FK: references cachegroup(name) |
 
 **Constraints**:
 - unique (topology, cachegroup) -- a cachegroup can only be in a Topology once.
--- a/blueprints/flexible-topologies.md
+++ b/blueprints/flexible-topologies.md
@@ -341,7 +341,7 @@ Since new Topologies can be more than 2 tiers (`EDGE` -> `MID`), `atstccfg` may 
 
 #### ATS config-related fields on Delivery Services
 
-Edge header rewrite rules and raw remap rules should still be applied only to the `EDGE` tier of Topologies. Mid header rewrite rules should probably only be applied at the *first* `MID` tier if there are multiple `MID` tiers in a Topology, because it might not be safe to assume that all mid header rewrite rules could be applied safely through multiple `MID` tiers.
+Edge header rewrite rules and raw remap rules should still be applied only to the `EDGE` tier of Topologies. Mid header rewrite rules should probably only be applied at the *first* `MID` tier if there are multiple `MID` tiers in a Topology, because it might not be safe to assume that all mid header rewrite rules could be applied safely through multiple `MID` tiers. This does bring up the question of whether or not we also want to be able to apply header rewrites to any arbitrary "tier" of a Topology. That might not be a requirement for this MVP, but it may be something we'll need to add in the future when a use case arises.
 
 ### Traffic Monitor Impact
 
@@ -404,6 +404,8 @@ Until legacy `deliveryservice_server` assignments are fully removed in favor of 
 Once all Delivery Services have been migrated to Topologies, CDN operators will no longer have to "clone Delivery Service assignments" from a nearby edge cache in the same cachegroup when adding a new edge cache into the CDN. Since Topologies are composed of cachegroups, simply adding a server into a particular cachegroup will give it all the Delivery Service assignments from all the Topologies the cachegroup is used in.
 
 Troubleshooting issues with Delivery Services on the CDN may become slightly more difficult due to the fact that you will no longer be able to assume that any given Delivery Service will follow the same global hierarchy. Unique Topologies will provide unique paths through the CDN from client to origin, so an operator now needs to look up what the path *should be* based on the Delivery Service's Topology in order to triage issues along the delivery path. This may necessitate the development of Topology-based troubleshooting tools and visualizations.
+
+As something to be aware of, content revalidations (via `regex_revalidate.config`) may take longer to propagate through a Topology the more tiers that it has, due to the nature of how ORT waits to revalidate until its cache's parents no longer have revalidations pending.
 
 #### Third Party Logging/Monitoring/Analytics
 
--- a/blueprints/flexible-topologies.md
+++ b/blueprints/flexible-topologies.md
@@ -333,7 +333,7 @@ The `deliveryservice` table will be updated to add a new column for the topology
 ### ORT Impact
 
 `atstccfg` will need to be updated to request the Topologies from Traffic Ops and use that data to determine the following for config generation:
-- what delivery services are assigned to a cache via Topologies -- in addition to legacy `deliveryservice_server` assignments used today
+- what delivery services are assigned to a cache via Topologies -- in addition to legacy `deliveryservice_server` assignments used today -- still taking Server Capabilities into account
 - if a delivery service is assigned to a cache via a Topology, the parent and secondary parent cachegroups for that delivery service are determined via the Topology it's assigned to. Otherwise, the parents are determined by the server's cachegroup as they are today.
 
 Since new Topologies can be more than 2 tiers (`EDGE` -> `MID`), `atstccfg` may need to break some assumptions about the current 2-tier hierarchy in order to work with an arbitrary number of "forward proxy tiers" -- e.g. `EDGE` -> `MID` -> `MID` -> `ORIGIN`. Basically, `MID` caches need to be able to forward requests to other `MID` caches -- they can no longer assume that their parents are always origins.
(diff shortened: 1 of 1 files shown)


blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+                "cachegroup": "secondary-parent-cachegroup",
+                "parents": []
+            }
+        ]
+    }
+}
+
+```
+
+##### `PUT /topologies?name=foo`
```
mitchell852: are we using query params for PUT/DELETE? I'm fine with it, just thought we were not.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+            "text": "topology was deleted successfully",
+            "level": "success"
+        }
+    ]
+}
+```
+
+##### `/deliveryservices` endpoints
+
+All relevant Delivery Service APIs will have their JSON request and response objects modified to include a new `topology` field which references the name of the topology it's assigned to:
```
mitchell852: Also, the following query param should be supported: GET /deliveryservices?topology=foo to retrieve all the ds's that use a given topology.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+            {
+                "cachegroup": "secondary-parent-cachegroup",
+                "parents": []
+            }
+        ]
+    }
+}
+```
+
+##### `DELETE /topologies?name=foo`
```
mitchell852: delete should probably fail if top used by a ds, correct?

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+
+### Traffic Portal Impact
+
+Traffic Portal will need new pages for creating and viewing Topologies, and the Delivery Service form will need to be updated to add a new Topology field for assigning a Delivery Service to a Topology. If a Delivery Service is assigned to a Topology, Traffic Portal should prohibit assigning `EDGE` servers to the Delivery Service (`ORIGIN` servers may still need to be assignable for MSO).
+
+Since Delivery Services will no longer be constrained to one global Topology as they are today, it would be extremely useful to be able to visualize a Delivery Service's Topology like a tree, where each node in the tree is a cachegroup, and the edges between nodes are the primary/secondary parent relationships between them. Clicking on a particular node would show all the servers in that cachegroup that could serve a request for the Delivery Service. This visualization will most likely be different from the Topology form for creating a Topology and does not necessarily need to be provided by Traffic Portal.
```
mitchell852: `Clicking on a particular node would show all the servers in that cachegroup that could serve a request for the Delivery Service`

This probably requires the following API support:

GET /servers?cg=x&cdn=y&ds=z so servers can be filtered by cg, then cdn (as cachegroups can have servers of multiple cdns), then ds (as a ds's required server capabilities are yet another filter. However, I suppose some of this filtering can be done on the UI side....but not sure about the server capability filter

GET /servers?cachegroup=mid-west&cdn=cdn1&

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+            {
+                "cachegroup": "secondary-parent-cachegroup",
+                "parents": []
+            }
+        ]
+    }
+]}
+```
+
+##### `POST /topologies`
```
mitchell852: who can PUT/POST/DELETE topologies? Operations+?

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+A new `topology_cachegroup` table will be created to model the association of cachegroups to topologies:
+
+| column     | type | modifiers                                       |
+| ---------- | ---- | ----------------------------------------------- |
+| id         | int  | not null, PK                                    |
+| topology   | text | not null, FK: references topology(name)         |
+| cachegroup | text | not null, FK: references cachegroup(short_name) |
+
+**Constraints**:
+- unique (topology, cachegroup) -- a cachegroup can only be in a Topology once.
```
mitchell852: i thought they could be in a topology more than once...

blueprints/flexible-topologies.md
ocket8888: A big push in API 2.0 was to cut down on duplicated endpoints which involved using query parameters instead of path parameters for a lot of things.

For my money, a path parameter would be fine, but then PUT with query parameters should not be supported, that way endpoints aren't duplicated. But more importantly, endpoints (like this one) should always support GET and `/foos/{{ID}}` shouldn't be handled exactly like `/foos?id={{ID}}`, which has historically never been the case. If no foo with ID=4 exists, then `GET /foos/4` should be a 404, while `/foos?id=4` should be an empty array. And actually, `/foos/{{ID}}` should never return an array while `/foos` should always return an array (for GET, at least).

Using query params for everything like this certainly cuts down on the amount of code you need to write, and especially since we currently have no real API standards I'm personally fine with that.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+| ---------- | ----------------- | ----------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| cachegroup | string            | required    | the `short_name` of a cachegroup this node maps to in the Topology                                                                                                                                          |
+| parents    | array of integers | required    | zero-based indexes to other nodes in the Topology's `nodes` array, where the 1st element is for the *primary* parent relationship and the 2nd element is for the *secondary* parent relationship, and so on |
+
+API constraints:
+- a Topology must have at least 1 `node`; otherwise, it is useless
+- there cannot be multiple `nodes` for the same cachegroup in a Topology
+- `parents` must have 0, 1 or 2 elements, cannot contain duplicates, cannot contain the index of its own `node`, and cannot contain the index of `nodes` whose cachegroup is of type `EDGE_LOC`
```
mitchell852: could parents not look like this:

`"parents": [1, 1]"`

which is essentially is the same as "no 2ndary parent" or i suppose its just easier to prevent duplicates.

blueprints/flexible-topologies.md
ocket8888: fwiw all of that filtering could be done client-side, but I'm not against adding query params either

blueprints/flexible-topologies.md
ocket8888: I'm definitely in favor of restricting duplicates.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Flexible Topologies
```
mitchell852: does this need to be renamed `Flexible Topologies Phase 1` with a section to briefly touch on what future phases will add?

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+
+If a Delivery Service is assigned to a Topology, any `deliveryservice_server` assignments it has to `EDGE` caches will be ignored, because it will be assigned to all caches in the Delivery Service's CDN (filtered by server capabilities) that belong to the Topology's cachegroups. Ideally, this feature will obsolete legacy `deliveryservice_server` assignments, since Topologies negate the need to assign Delivery Services to individual `EDGE` caches. Nonetheless, legacy `deliveryservice_server` assignments will be supported alongside Topology-based Delivery Services for some time until all Delivery Services have been migrated to Topologies.
+
+### Traffic Portal Impact
+
+Traffic Portal will need new pages for creating and viewing Topologies, and the Delivery Service form will need to be updated to add a new Topology field for assigning a Delivery Service to a Topology. If a Delivery Service is assigned to a Topology, Traffic Portal should prohibit assigning `EDGE` servers to the Delivery Service (`ORIGIN` servers may still need to be assignable for MSO).
```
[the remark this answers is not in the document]
mitchell852: > If a Delivery Service is assigned to a Topology, Traffic Portal should prohibit assigning `EDGE` servers to the Delivery Service

As should the API

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+
+## Proposed Change
+
+Traffic Control will provide the ability to define one or more Topologies, and a Topology can have any number of Delivery Services assigned to it. A Topology will be composed of Cachegroups along with their primary/secondary parent relationships to other Cachegroups as defined by the Topology.
+
+If a Delivery Service is assigned to a Topology, any `deliveryservice_server` assignments it has to `EDGE` caches will be ignored, because it will be assigned to all caches in the Delivery Service's CDN (filtered by server capabilities) that belong to the Topology's cachegroups. Ideally, this feature will obsolete legacy `deliveryservice_server` assignments, since Topologies negate the need to assign Delivery Services to individual `EDGE` caches. Nonetheless, legacy `deliveryservice_server` assignments will be supported alongside Topology-based Delivery Services for some time until all Delivery Services have been migrated to Topologies.
+
+### Traffic Portal Impact
```
mitchell852: Want to add a paragraph along these lines:

> Traffic Portal's cdn snapshot view will need to be modified to account for a new section top-level section (Topologies) of the CRconfig.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+
+### Performance Impact
+
+There should be no visible impact to performance from an application perspective -- this feature does not introduce anything particularly CPU, network, or storage-intensive. However, this feature will allow the tuning of Topologies in the CDN, which *will* affect end-to-end performance of the CDN in terms of things like latency, cache hit ratio, cache efficiency, etc. CDN architects will be able to make certain trade-offs in their Topologies until their desired end-to-end performance characteristics are met.
+
+Additionally, the more Delivery Services that are migrated from legacy `deliveryservice_server` assignments to Topologies, the smaller the size of the `CRConfig` will get. Currently, `deliveryservice_server` assignments are responsible for the most of the growth in `CRConfig` size due to the natural addition of both Delivery Services and Servers to a CDN over time, and migrating to Topology-based Delivery Services should noticeably reduce the size and growth of the `CRConfig` over time.
```
mitchell852: this is a nice bonus!

blueprints/flexible-topologies.md
mitchell852: scratch that. you mention it later.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+
+### Operations Impact
+
+One of the bigger day-to-day operational impacts this feature will have is in the assignment of Delivery Services to Topologies instead of to individual `EDGE` caches. In theory, if the number of unique Topologies is kept as low as possible, it should be easier to choose a specific Topology for a Delivery Service than to assign it to individual caches (unless the default approach is to just assign delivery services to *all* caches). This could be made easier by defining a kind of *default* Topology which is assigned to until it is determined that the default Topology does not meet the Delivery Service's requirements.
+
+Until legacy `deliveryservice_server` assignments are fully removed in favor of Topology-based assignments, there will be extra operational overhead due to having two different ways to assign Delivery Services, but this overhead should be temporary as CDN operators should migrate all Delivery Services to Topologies as soon as possible. That one-time manual migration should also be considered an operational impact and should be scheduled to take place sometime after the upgrade is completed.
```
mitchell852: this default topology thing is kind of interesting and maybe we can pursue in future phases of this feature? 
i.e.

cdn1, default topology=foo
cdn2, default topology=bar

this isn't something that can be seeded with TO as a topology consists of cgs and servers that are company specific but maybe giving operations or admin users the ability to select a default topology for each cdn would be very beneficial imo.

blueprints/flexible-topologies.md
zrhoffman: If we add topologies, what will we rename the "Topology" section to in Traffic Portal?

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+
+#### Client Impact
+
+New Go client methods will be added for the `/topologies` endpoints in order to write TO API tests for the new endpoints. The `/deliveryservices` client methods won't need modified as the new `Topology` field will simply be added to the `DeliveryService` struct. New client methods for the Python client will also be added for each of the new `/topologies` endpoints.
+
+#### Data Model Impact
+
+##### Go structs
+
+New `Topology` and `TopologyNode` structs will be added for the `/topologies` endpoints, mapping directly to the JSON request bodies in the REST API Impact section. The `DeliveryService` struct will be updated with a new `Topology` field which is the name of the Topology the Delivery Service is assigned to.
```
zrhoffman: How will `TopologyNode` be used? It is not mentioned anywhere else in the blueprint.

blueprints/flexible-topologies.md
mitchell852: nevermind, the design supports it but for now a cachegroup will only appear in a topology once max. got it.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+```
+
+##### The various `/snapshot` endpoints
+
+The various `/snapshot` endpoints will need to be updated to include new Topologies data along with their associations to Delivery Services in the `CRConfig.json` snapshot. The data should only include the `EDGE_LOC` cachegroups of the Topologies, because those are all Traffic Router needs.
+
+##### Various endpoints that are affected by cachegroup parentage or deliveryservice-server assignment
+
+API endpoints that do things such as the following may need to be updated to take Topology-based Delivery Service assignment and parentage into account:
+- assign a Delivery Service to a server (or vice versa)
```
mitchell852: I would suggest leaving all the ds/server and server/ds endpoints as is. We can circle back and mark them as deprecated at some point. Technically, they are still relevant as you can either a) assign a topology to a ds or b) assign servers to a ds.

In my opinion, there could be 2 ways to fetch servers "assigned to a ds"

1. GET /ds/{id}/servers (existing) which will return servers DIRECTLY assigned to a ds
2. GET /servers?topology=foo which will return servers INDIRECTLY assigned to a ds via a topology.

We can make TP smart enough to use the correct one based on if a ds has a topology or not and we can update documentation to be clear what each endpoint is for.

blueprints/flexible-topologies.md
ocket8888: I believe "TopologyNode" maps to the `node` property of a Topology

blueprints/flexible-topologies.md
ocket8888: Yes but a Delivery Service cannot simultaneously have an assigned Topology and assigned Servers. So the Delivery Service-to-server assignment endpoints will need to check if a Delivery Service has a Topology before allowing servers to be assigned.

mitchell852: so what's the topology equivalent of this:

GET servers/{id}/deliveryservices?

hmm. maybe 

GET /topologies?server=id would give you all the topologies that include the server...but how to fetch all the delivery services for those topologies...?

i guess my ultimate point is leave the ds/server server/ds endpoints as is and come up with other topology related ones to achieve the same result from a GET perspective.

blueprints/flexible-topologies.md
mitchell852: also the TP menu needs to change:

image

Since Topologies will have it's own submenu item, the parent menu name `Topology` probably needs a better name.

blueprints/flexible-topologies.md
mitchell852: also, the following query param should probably be added to cachegroups endpoint:

GET /cachegroups?topology=foo

blueprints/flexible-topologies.md
mitchell852: and then by calling something like GET /topologies/default?type=http_live it would return the default, if any, which could be selected in a new DS form in TP.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+
+Once all Delivery Services have been migrated to Topologies, CDN operators will no longer have to "clone Delivery Service assignments" from a nearby edge cache in the same cachegroup when adding a new edge cache into the CDN. Since Topologies are composed of cachegroups, simply adding a server into a particular cachegroup will give it all the Delivery Service assignments from all the Topologies the cachegroup is used in.
+
+Troubleshooting issues with Delivery Services on the CDN may become slightly more difficult due to the fact that you will no longer be able to assume that any given Delivery Service will follow the same global hierarchy. Unique Topologies will provide unique paths through the CDN from client to origin, so an operator now needs to look up what the path *should be* based on the Delivery Service's Topology in order to triage issues along the delivery path. This may necessitate the development of Topology-based troubleshooting tools and visualizations.
```
jhg03a: Yes.  Operators need better tools to know how parentage should work for a given DS and for a given DS from a given server up and down the connectivity path.  I would be hesitant to support even phase 1 without it due to the additional overhead incurred triaging what's going on.

jhg03a: It might also be worth a note that any external 3rd party logging, analytics, or monitoring tools presuming a 2 tier architecture will need an update and may become more complex to model successfully.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+API endpoints that do things such as the following may need to be updated to take Topology-based Delivery Service assignment and parentage into account:
+- assign a Delivery Service to a server (or vice versa)
+- return the servers that are assigned to a Delivery Service (or vice versa)
+- perform an operation on "child" cachegroups -- like queueing updates on "child" caches when changing the status of a "parent" cache
+
+#### Client Impact
+
+New Go client methods will be added for the `/topologies` endpoints in order to write TO API tests for the new endpoints. The `/deliveryservices` client methods won't need modified as the new `Topology` field will simply be added to the `DeliveryService` struct. New client methods for the Python client will also be added for each of the new `/topologies` endpoints.
+
+#### Data Model Impact
```
jhg03a: How does topologies affect our current hardcoded edge and mid header rewrite rules & raw remaps on delivery services?

blueprints/flexible-topologies.md
mitchell852: Good question. What would call a section that contains topologies, cache groups, coordinates, physical locations, divisions, regions, asns?

zrhoffman: Structure? Hierarchy? Architecture? Scope? Grouping?

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+}
+
+```
+
+##### `PUT /topologies?name=foo`
+
+request JSON:
+```JSON
+{
+    "name": "foo",
```
mitchell852: any rules on topology name? i.e. no spaces? a-zA-z0-9?

mitchell852:
Currently, assigning caches to a _steering_  delivery service has no effect as the target delivery services are the delivery services that really serve the content. I think the same should apply to assigning a topology to a _steering_  delivery service. In fact, the delivery service create/update apis should prevent any such assignment by returning a 400 Bad Request if trying to assign a topology to a _steering_  delivery service.

blueprints/flexible-topologies.md
mitchell852: > Yes but a Delivery Service cannot simultaneously have an assigned Topology and assigned Servers. So the Delivery Service-to-server assignment endpoints will need to check if a Delivery Service has a Topology before allowing servers to be assigned.

actually, i think it can have both....but if you have both, topology wins.

blueprints/flexible-topologies.md
mitchell852: i think supporting /servers?topology=x should probably be fine and let the UI filter out from there if needed.

mitchell852: > i think supporting /servers?topology=x should probably be fine and let the UI filter out from there if needed.

actually i don't think that will work. the server object does not have server capabilities attached to it so it's impossible to filter out servers that don't have the capabilities required of the DS. Maybe this endpoint needs to support the topology query param:

GET deliveryservices/{id}/servers/eligible?topology=foo

^^ deliveryservices/{id}/servers/eligible already removes servers not in a ds's cdn / that don't have the ds's required capabilities.

mitchell852:
Currently, you can queue updates at the cdn, cache group or server (cache) level. Would it make sense to be able to queue updates for a Topology? I.e. DS A uses Topology B. You make a change to DS A that requires a queue updates. Might make sense to queue at the topology level as opposed to the cdn level.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,395 @@
...
+**Constraints**:
+- unique (topology, cachegroup) -- a cachegroup can only be in a Topology once.
+
+A new `topology_cachegroup_parents` table will be created to model the parent relationships of cachegroups within a topology:
+
+| column | type | modifiers                                        |
+| ------ | ---- | ------------------------------------------------ |
+| child  | int  | not null, FK: references topology_cachegroup(id) |
+| parent | int  | not null, FK: references topology_cachegroup(id) |
+| rank   | int  | not null                                         |
```
zrhoffman: In #4564, I included `last_updated` columns for each of these tables.

rawlinp: Yeah, that's good. I'll include those in the blueprint in the next revision.

rawlinp: actually, I think I'll leave out the last_updated column and field, since they're basically assumed at this point and are kind of distracting in the blueprint

rawlinp:
> Currently, you can queue updates at the cdn, cache group or server (cache) level. Would it make sense to be able to queue updates for a Topology? I.e. DS A uses Topology B. You make a change to DS A that requires a queue updates. Might make sense to queue at the topology level as opposed to the cdn level.

I'm hesitant to add "yet another way to queue updates on caches". Topologies are multi-CDN, so would queuing updates on a Topology queue updates on all servers in the Topology across all CDNs? IMO it seems like "queue updates on CDN" is the default operator behavior when making changes to delivery services.

blueprints/flexible-topologies.md
```diff
@@ -0,0 +1,426 @@
...
+##### `PUT /topologies?name=foo`
+
+request JSON:
+```JSON
+{
+    "name": "foo",
+    "description": "a foo topology",
+    "nodes": [
+        {
+            "cachegroup": "child-cachegroup",
```
mitchell852, line 192: i think cachegroup type would be a good thing to know. i.e. MID_LOC, EDGE_LOC

rawlinp: So we discussed this offline a bit already, and it sounds like we're just going to use the results of `GET /cachegroups` rather than including that in the topology resource.

blueprints/flexible-topologies.md
jhg03a: So what about when we need to add a header rewrite to adjust the ATS behavior on a second mid tier?

jhg03a: Every tier can have wildly different header rewrites.  They can also vary by raw remap equally even though there isn't a field for it.  We had to invent the anymap override hack just to fabricate it as a concept.

jhg03a:
Something else to think about, when adding more tiers to a topology that's going to increase revalidation delays (or worse, likelyhood of never completing).  Currently there is a rule that parent tiers must revalidate before child tiers.

blueprints/flexible-topologies.md
rawlinp: The concept of a "tier" is not really the best way to think of things anymore because a Topology doesn't have to be symmetric. For example, maybe clients in CO would go through three "tiers" to reach the origin, but clients in WA only have to go through two. So any kind of mapping of ATS rules to specific "tiers" would have to be directed to specific _nodes_ in the Topology. We don't really know if that's a problem we have today, since we only have two tiers, but I think we could enhance that functionality if it becomes an issue.

rawlinp:
That is a good point, I can mention that in the operational impacts.

blueprints/flexible-topologies.md
```diff
@@ -405,6 +405,8 @@ Once all Delivery Services have been migrated to Topologies, CDN operators will
 
 Troubleshooting issues with Delivery Services on the CDN may become slightly more difficult due to the fact that you will no longer be able to assume that any given Delivery Service will follow the same global hierarchy. Unique Topologies will provide unique paths through the CDN from client to origin, so an operator now needs to look up what the path *should be* based on the Delivery Service's Topology in order to triage issues along the delivery path. This may necessitate the development of Topology-based troubleshooting tools and visualizations.
 
+As something to be aware of, content revalidations (via `regex_revalidate.config`) may take longer to propagate through a Topology the more tiers that it has, due to the nature of how ORT waits to revalidate until its cache's parents no longer have revalidations pending.
```
jhg03a, line 408: This is actually a big problem if you have revalidations coming in faster than they can chain through your topology since it's a global state on the server object.  I also don't know how any of that would work if you have multiple topologies for a given server with different parentages.

rawlinp: Since revalidations are basically per-CDN at this point, a given server would have to wait for all of its parents (via all of the Topologies it is part of). You do bring up a good point about revalidations coming in too fast -- if that becomes a problem we will probably have to come up with a new process where each revalidation has a timestamp associated with it so that revalidations can be done up to a certain point in time rather than needing to wait until everything is caught up. Although that would be out of the scope of this blueprint -- we can cross that bridge when it comes :)

jhg03a: If you're measuring revalidation time based on how long it takes to fully propagate to the client edge today it's per CDN * number of tiers * revalidation interval.  I raise it mostly because of the scalability concern that as the number of tiers increases, the total time will also increase.  Additionally, the probability that it will be pre-empted by the submission of another job also increases.  If we want to just kick that can down the road, we can.

blueprints/flexible-topologies.md
ocket8888: Yes, that's all exactly right. The only thing is the guidelines recommend PATCH as well.

blueprints/flexible-topologies.md
```diff
@@ -250,11 +250,29 @@ response JSON:
 
 ##### `/deliveryservices` endpoints
 
-All relevant Delivery Service APIs will have their JSON request and response objects modified to include a new `topology` field which references the name of the topology it's assigned to:
-```JSON
+All relevant Delivery Service APIs will have their JSON request and response objects modified to include the following new fields:
+- `topology` - the name of the topology it's assigned to
+- `firstHeaderRewrite` - the header_rewrite ATS config that will be applied to the *first tier* caches in the Delivery Service's Topology
+- `middleHeaderRewrite` - the header_rewrite ATS config that will be applied to all *middle tier* caches in this Delivery Service's Topology
+- `lastHeaderRewrite` - the header_rewrite ATS config that will be applied to all *last tier* caches in this Delivery Service's Topology
```
zrhoffman, line 257: IMO `midHeaderRewrite` sounds too close to the existing `midHeaderRewrite` rules that we have today.

It would be nice if `firstHeaderRewrite`/`lastHeaderRewrite` were named something that implies their place in the topology, like `leafHeaderRewrite`/`rootNodeHeaderRewrite`.

rawlinp: Yeah, I was considering `intermediateHeaderRewrite`/`innerHeaderRewrite` vs `middleHeaderRewrite` because it is close like you said. Honestly I would like to continue using `edgeHeaderRewrite` too since that still makes sense in the topology world, but I'm not sure we can take those fields as-is without some manual migration to the new fields. `root` can be confusing because origins are really the "root", but they don't have to be in the topology. Plus a topology isn't truly a tree, so there isn't really a root either.

jhg03a: Edge/Parents/LastParent ?

blueprints/flexible-topologies.md
```diff
@@ -250,11 +250,29 @@ response JSON:
...
+
+**Note:** these three header_rewrite "buckets" were chosen because header_rewrite configs usually fall into one of three categories:
+- affects requests to/from the client ("first")
+- affects requests to/from the origin ("last")
+- don't affect clients or origins ("middle")
+
+**API Constraints:**
+- `firstHeaderRewrite`, `middleHeaderRewrite`, and `lastHeaderRewrite` can only be set if `topology` is not null
+- `edgeHeaderRewrite` and `midHeaderRewrite` cannot be set if `topology` is not null
```
jhg03a, line 266: Instead of having parallel mutually exclusive fields serving the same purpose, couldn't we just drop the first and last fields?  Seems like that creates additional tradition steps when converting to topologies.  The concept of a middleHeaderRewrite doesn't currently exist, so it's fine to be topology only.

blueprints/flexible-topologies.md
```diff
@@ -250,11 +250,29 @@ response JSON:
 
 ##### `/deliveryservices` endpoints
 
-All relevant Delivery Service APIs will have their JSON request and response objects modified to include a new `topology` field which references the name of the topology it's assigned to:
-```JSON
+All relevant Delivery Service APIs will have their JSON request and response objects modified to include the following new fields:
```
jhg03a, line 253: The same treatment should be applied to rawRemap.  This also lets us remove the ##OVERRIDE anymap hack I think.

blueprints/flexible-topologies.md
rawlinp: This is on purpose -- we want it to be part of the process of migrating legacy DSes to Topologies. Part of that is making sure the existing header-rewrite configs are safe and make sense in the Topology you're assigning the DS to.

blueprints/flexible-topologies.md
zrhoffman: Feel free to resolve this convo since your response makes sense and @jhg03a additionally suggested Edge/Parents/LastParent elsewhere. I don't have permission to or I would.

PR #4538, opened 2020-03-25
Deprecate steering/:ds/targets/:target 

shamrickus:

- [x] This PR deprecates steering/:ds/targets/:target

- Documentation
- Traffic Ops

Ensure endpoint is gone in 2.0 and deprecated in 1.1


traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -957,7 +956,7 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{api.Version{1, 1}, http.MethodGet, `deliveryservices/{id}/health/?(\.json)?$`, deliveryservice.GetHealth, auth.PrivLevelReadOnly, Authenticated, nil, 2034590101, perlBypass},
 
 		{api.Version{1, 1}, http.MethodGet, `steering/{deliveryservice}/targets/?(\.json)?$`, api.ReadHandler(&steeringtargets.TOSteeringTargetV11{}), auth.PrivLevelReadOnly, Authenticated, nil, 1569607824, noPerlBypass},
-		{api.Version{1, 1}, http.MethodGet, `steering/{deliveryservice}/targets/{target}$`, api.ReadHandler(&steeringtargets.TOSteeringTargetV11{}), auth.PrivLevelReadOnly, Authenticated, nil, 105995848, noPerlBypass},
+		{api.Version{1, 1}, http.MethodGet, `steering/{deliveryservice}/targets/{target}$`, api.DeprecatedReadHandler(&steeringtargets.TOSteeringTargetV11{}, nil), auth.PrivLevelReadOnly, Authenticated, nil, 105995848, noPerlBypass},
```
mhoppa: is the replacement route the base path `steering/{deliveryservice}/targets` with a query parameter of ID ?

docs/source/api/v2/steering_id_targets_targetID.rst
```diff
@@ -1,230 +0,0 @@
...
-	X-Server-Name: traffic_ops_golang/
-	Date: Tue, 11 Dec 2018 14:42:54 GMT
-	Content-Length: 69
-
-	{ "alerts": [
-		{
-			"text": "steeringtarget was deleted.",
-			"level": "success"
-		}
-	]}
```
ocket8888: You only deprecated the GET method, so all of this documentation should still be in there.

traffic_ops/traffic_ops_golang/routing/routes.go
shamrickus: You are correct, I have updated it accordingly

PR #4543, opened 2020-03-25
'fixed' grafana scripted dashboard to use exact hostname instead of regular expression match

### LINKED ISSUE #4541: traffic_ops_server.js searches based on which= and can find data for multiple servers
## I'm submitting a ...
- [x] bug report
- [ ] other

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
The traffic_ops_server.js script pulls data to display using which=hostname.  However with a hostname "edge-01" and another "edge-010", using the script with "edge-01" finds data from both servers, and shows a very odd graph. If I fix the issue by using "edge-01.infra then the mem/cpu/netstat data is fine, but the count and bandwidth data are then not displayed.

## Expected / new behavior:
The script should discern each host, and not combine data from 2 hosts. 

See the output of the script with a `which` parameter that is contained in multiple cache server names.

ocket8888:

- [x] This PR fixes #4541

I _think_ this fixes the issue where it would select multiple servers when "which" is specified with something that is contained in the hostnames of multiple servers. I've tested it in CDN-in-a-Box and everything appeared to be working, but for some reason about half of the graphs are trying to source their data from "telegraf", which looks like some kind of hold-over from when this was a Comcast internal project.

- Traffic Stats

Open the scripted dashboard with a "which" value that is contained in multiple cache server hostnames. Verify that only the cache server with a hostname exactly matching `which` is selected. Unless there are multiple with the same hostname, which is totally allowed and can happen. Idk what the dashboard is supposed to do in that case.

### DIFF
--- a/traffic_stats/grafana/traffic_ops_server.js
+++ b/traffic_stats/grafana/traffic_ops_server.js
@@ -13,1368 +13,1359 @@
      limitations under the License.
 */
 
-/* global _ */
-
 /*
  * Scripted dashboard for traffic ops.
  *
  * Based on the grafana scripted.js script (which is ASF 2.0 licensed).
  */
 
-
-
-// accessible variables in this scope
-var window, document, ARGS, $, jQuery, moment, kbn;
+'use strict';
 
 // Setup some variables
 var dashboard;
 
-// All url parameters are available via the ARGS object
+// All URL parameters are available via the ARGS object
 var ARGS;
 
-// Intialize a skeleton with nothing but a rows array and service object
+// Intialize a skeleton with nothing but a rows array and service object,
+// and setting default time and refresh interval.
 dashboard = {
-  rows : [],
+  refresh: "30s",
+  rows: [],
+  // time can be overridden in the URL using from/to parameters, but this is
+  // handled automatically in grafana core during dashboard initialization
+  time: {
+    from: "now-24h",
+    to: "now"
+  }
 };
 
+let which = 'argName';
 
-// Set default time
-// time can be overriden in the url using from/to parameters, but this is
-// handled automatically in grafana core during dashboard initialization
-dashboard.time = {
-  from: "now-24h",
-  to: "now"
-};
-
-var which = 'argName';
-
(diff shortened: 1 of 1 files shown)


traffic_stats/grafana/traffic_ops_server.js
```diff
@@ -13,1368 +13,1359 @@
...
+    from: "now-24h",
+    to: "now"
+  }
 };
 
-var which = 'argName';
+let which = 'argName';
 
-if(!_.isUndefined(ARGS.which)) {
+if(ARGS.which !== undefined) {
```
zrhoffman, line 45: Nit: Since the whole file is reformatted now, `if(` should be `if (`.

ocket8888: Oh god it is isn't it?

zrhoffman: It looks like the change that blew up the diff was removing the curly braces around `dashboard.rows.push()`. After adding those back and reformatting, it's way more straightforward to see what changed.

ocket8888: Yes... but also idk why that nested block was there in the first place. I'll have to put it back in, I suppose.

zrhoffman: The brackets were added in Comcast/traffic_control#592, before which `traffic_ops_server.js` was just [the `else if ( type == "server" ) {}` clause](https://github.com/apache/trafficcontrol/blob/958e32a0f6/traffic_stats/grafana/traffic_ops_scripted.js#L376-L523) of a big if/else statement in `traffic_ops_scripted.js`, so there is no reason to keep them. But yeah, better to save removing the brackets from this and the other scripted dashboards for a separate PR.

PR #4547, opened 2020-03-26
Add default sort to phys_locations

### LINKED ISSUE #2794: order by name server location dropdown (physLocationService.getPhysLocations)
choosing site on server add page is a random order, with over 50 sites, it's painful to navigate.

in FormServerController.js
needs { orderby: 'name' }

shamrickus:

- [x] This PR fixes #2794

- Traffic Ops
- Traffic Portal

Verify TO api default orders phys_locations by name and that it can be overridden.
Verify in TP that when adding a server, sites are sorted by name

### DIFF
--- a/traffic_ops/testing/api/v2/phys_locations_test.go
+++ b/traffic_ops/testing/api/v2/phys_locations_test.go
@@ -16,6 +16,7 @@ package v2
 */
 
 import (
+	"sort"
 	"testing"
 
 	tc "github.com/apache/trafficcontrol/lib/go-tc"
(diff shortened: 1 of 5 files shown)


traffic_ops/testing/api/v2/phys_locations_test.go
```diff
@@ -76,6 +79,33 @@ func GetTestPhysLocations(t *testing.T) {
 			t.Errorf("cannot GET PhysLocation by name: %v - %v", err, resp)
 		}
 	}
+
+}
+
+func GetSortPhysLocationsTest(t *testing.T) {
```
mattjackson220: What are your thoughts about adding these tests to the v1 tests also since it will change the ordering in v1 and v2? not sure its super important but might be worth doing

PR #4556, opened 2020-03-27
PUT /snapshot?cdn=not-found returns 404

### LINKED ISSUE #4551: PUT request to snapshot endpoint returns Internal Server Error
## I'm submitting a ...

- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When making a `PUT` request to `/snapshot` with the query parameter `cdn` set to a cdn name that doesn't exist, `i.e. /snapshot?cdn=doesntexist` the response returned is:
```http
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```
## Expected / new behavior:

Should return:
```http
HTTP/1.1 404 Not Found
{
    "alerts": [
        {
            "text": "No CDN found with that name",
            "level": "error"
        }
    ]
}
```
## Anything else:

Error Log:
```ERROR: api.go:180: ... Error getting CDN info: Error querying CDN domain name: sql: no rows in result set```

### PR DESCRIPTION
**Pull request by mitchell852**

- [x] This PR fixes #4551 

- Traffic Ops

- Run the API tests. Ensure they all pass.

- master

### DIFF
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -31,6 +31,7 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-rfc"
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/monitoring"
 )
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: traffic_ops/traffic_ops_golang/crconfig/handler.go [hunk 1]
```diff
@@ -174,6 +174,12 @@ func snapshotHandler(w http.ResponseWriter, r *http.Request, deprecated bool) {
 			return
 		}
 		cdn = name
+	} else {
+		_, _, err := getCDNInfo(cdn, inf.Tx.Tx)
+		if err != nil {
```
mhoppa: The `getCDNInfo` func call doesnt differentiate between database query error (lets say the db was down) or the cdn not existing.  Those would call for different errors being returns to the user.

I would suggest using https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go#L485 as it returns if it exists or not in a separate var

PR #4557, opened 2020-03-27
Add ORT/atstccfg Deterministic Config Generation

rob05c:
This changes ORT config gen to be deterministic. Configs should always be identical, except for the timestamp in the comment header.

This should make future changes much safer and easier to identify, easier to identify bugs, and just easier and safer to diff configs to see changes in general. With deterministic gen, diffing should be as simple as a line diff, e.g. the POSIX `diff` command.

Includes tests. No docs, no changelog, no documented interface change (we don't document config order).

- Traffic Ops ORT

Run ORT and/or it's helper atstccfg multiple times, and verify the output files are identical except for the comment header timestamp.

Not a bug fix.

## Additional Information

### DIFF
--- a/lib/go-atscfg/logsdotxml.go
+++ b/lib/go-atscfg/logsdotxml.go
@@ -27,12 +27,18 @@ import (
 const LogsXMLFileName = "logs_xml.config"
 const ContentTypeLogsDotXML = `text/xml`
 
+const LineCommentLogsDotXML = `<!--`
+
 func MakeLogsXMLDotConfig(
 	profileName string,
 	paramData map[string]string, // GetProfileParamData(tx, profile.ID, LoggingYAMLFileName)
 	toToolName string, // tm.toolname global parameter (TODO: cache itself?)
 	toURL string, // tm.url global parameter (TODO: cache itself?)
 ) string {
+
+	// Note LineCommentLogsDotXML must be a single-line comment!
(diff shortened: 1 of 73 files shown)


lib/go-atscfg/logsdotxml.go
```diff
@@ -27,12 +27,18 @@ import (
...
 	toToolName string, // tm.toolname global parameter (TODO: cache itself?)
 	toURL string, // tm.url global parameter (TODO: cache itself?)
 ) string {
+
+	// Note LineCommentLogsDotXML must be a single-line comment!
+	// But this file doesn't have a single-line format, so we use <!-- for the header and promise it's on a single line
+	// Note! if this file is ever changed to have multi-line comments, LineCommentLogsDotXML will have to be changed to the empty string.
 	hdrComment := GenericHeaderComment(profileName, toToolName, toURL)
 	hdrComment = strings.Replace(hdrComment, `# `, ``, -1)
 	hdrComment = strings.Replace(hdrComment, "\n", ``, -1)
```
alficles, line 44: This should probably just replace `# ` with `<!--` and `\n` with `-->\n`? That would make it at least work in the future if someone returns a multiline comment. This works, though, doesn't change current behaviour, and won't break in the future. This is fine, though.

dneuman64:
Good enough for Chris, good enough for me 👍

PR #4575, opened 2020-03-31
Create Flexible Topologies tables

### LINKED ISSUE #4564: Add Flexible Topologies tables and modify deliveryservice table
## I'm submitting a ...

- [x] Traffic Ops DB

## Acceptance Criteria

- `topology` table adheres to this schema:

    | column      | type | modifiers    |
    | ----------- | ---- | ------------ |
    | name        | text | not null, PK |
    | description | text |              |
    | last_updated | timestamptz  | default `now()`                                  |

- `topology_cachegroup` table adheres to this schema:

    | column     | type | modifiers                                       |
    | ---------- | ---- | ----------------------------------------------- |
    | id         | int  | not null, PK                                    |
    | topology   | text | not null, FK: references topology(name)         |
    | cachegroup | text | not null, FK: references cachegroup(short_name) |
    | last_updated | timestamptz  | default `now()`                                  |

- topology_cachegroup_parents table adheres to this schema:

    | column       | type         | modifiers                                        |
    | ------------ | ------------ | ------------------------------------------------ |
    | child        | int          | not null, FK: references topology_cachegroup(id) |
    | parent       | int          | not null, FK: references topology_cachegroup(id) |
    | rank         | int          | not null                                         |
    | last_updated | timestamptz  | default `now()`                                  |

- After running migrations, `deliveryservice` table has a `topology` column that can be described as:

    | column   | type | modifiers                     |
    | -------- | ---- | ----------------------------- |
    | topology | text | FK: references topology(name) |
The line

See the Flexible Topologies PR for specifics: apache/trafficcontrol#4537

zrhoffman:

This PR adds the `topology`, `topology_cachegroup`, and `topology_cachegroup_parents` tables to `create_tables.sql` and adds a migration to create the `topology` column in the `deliveryservice` table.

- [x] This PR fixes #4564

* The [existing documentation](https://traffic-control-cdn.readthedocs.io/en/release-4.0.0/development/traffic_ops.html#cmdoption-admin-arg-command) is sufficient for `create_tables.sql` and migrations.
* We do not currently lint or test `create_tables.sql` or migrations.

- Traffic Ops database

- Confirm the database is initialized correctly in CDN-in-a-Box
- Confirm that the tables and columns mentioned in [the Flexible Topologies blueprint](https://github.com/rawlinp/trafficcontrol/blob/topologies-blueprint/blueprints/flexible-topologies.md#traffic-ops-database) are present, including foreign keys and constraints

Not a bug fix.

### DIFF
--- a/traffic_ops/app/db/create_tables.sql
+++ b/traffic_ops/app/db/create_tables.sql
@@ -1503,6 +1503,23 @@ CREATE TABLE topology_cachegroup (
 ALTER TABLE topology_cachegroup OWNER TO traffic_ops;
 
 --
+-- Name: topology_cachegroup_parents; Type: TABLE; Schema: public; Owner: traffic_ops
+--
+
+CREATE TABLE topology_cachegroup_parents (
+    child bigint NOT NULL,
+    parent bigint NOT NULL,
+    rank integer NOT NULL,
+    last_updated timestamp with time zone DEFAULT now() NOT NULL,
+    CONSTRAINT topology_cachegroup_parents_rank_check CHECK (((rank = 1) OR (rank = 2))),
+    CONSTRAINT unique_child_rank UNIQUE (child, rank),
+    CONSTRAINT unique_child_parent UNIQUE (child, parent)
+);
+
+
+ALTER TABLE topology_cachegroup_parents OWNER TO traffic_ops;
+
+--
 -- Name: type; Type: TABLE; Schema: public; Owner: traffic_ops
 --
 
@@ -2604,6 +2621,18 @@ CREATE INDEX origin_tenant_fkey ON origin USING btree (tenant);
 CREATE INDEX topology_cachegroup_cachegroup_fkey ON topology_cachegroup USING btree (cachegroup);
 
 --
+-- Name: topology_cachegroup_parents_child_fkey; Type: INDEX; Schema: public; Owner: traffic_ops
+--
+
+CREATE INDEX topology_cachegroup_parents_child_fkey ON topology_cachegroup_parents USING btree (child);
+
+--
+-- Name: topology_cachegroup_parents_parents_fkey; Type: INDEX; Schema: public; Owner: traffic_ops
+--
+
+CREATE INDEX topology_cachegroup_parents_parents_fkey ON topology_cachegroup_parents USING btree (parent);
+
+--
 -- Name: topology_cachegroup_topology_fkey; Type: INDEX; Schema: public; Owner: traffic_ops
 --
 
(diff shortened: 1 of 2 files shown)


traffic_ops/app/db/migrations/20200331031637_add_deliveryservice_topology.sql
```diff
@@ -0,0 +1,27 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+* License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE deliveryservice
+    ADD COLUMN topology text,
+    ADD CONSTRAINT deliveryservice_topology_fkey FOREIGN KEY (topology) REFERENCES topology (name) ON UPDATE CASCADE ON DELETE CASCADE;
```
rawlinp: also, it's generally not required to explicitly name constraints like this and define indexes for foreign keys. Postgres will create those indexes and name them automatically as long as you do something like this for instance:
```
ALTER TABLE deliveryservice ADD COLUMN topology text REFERENCES topology(name) ON UPDATE CASCADE ON DELETE RESTRICT;
```
And that reminds me, this should be `ON DELETE RESTRICT` because we don't want to allow a Topology to be deleted if a delivery service is still assigned, right?

traffic_ops/app/db/create_tables.sql
```diff
@@ -1474,6 +1474,50 @@ ALTER TABLE to_extension_id_seq OWNER TO traffic_ops;
...
+--
+
+CREATE TABLE topology (
+    name text PRIMARY KEY NOT NULL,
+    description text NOT NULL,
+    last_updated timestamp with time zone DEFAULT now() NOT NULL
+);
+
+
+ALTER TABLE topology OWNER TO traffic_ops;
```
rawlinp: goose migrations generally don't require explicitly setting the owner either

traffic_ops/app/db/migrations/20200331031637_add_deliveryservice_topology.sql
rob05c: I'm +1 on explicitly naming things like this.

We've had cases in the past where automatically named constraints, keys, etc ended up with different names on different installations, and then we later needed to make a migration to modify the symbol, and it was difficult or impossible, because the name was different on different systems.

rawlinp:
Also, I'm kind of torn about creating all of the database tables in a PR without the API implementation using them yet. It's possible we'll miss something in the schema that we'd realize only when implementing the API. Then we'd have to go back and fix the DB migration, which usually means adding another DB migration, but as long as it isn't released maybe we can get away with just fixing the existing migration without having to add a new one. Thoughts?

traffic_ops/app/db/migrations/20200331031637_add_deliveryservice_topology.sql
rawlinp: I'm not -1 on explicit names -- I would be fine either way, I'm just saying it's generally not required.

zrhoffman:
> Also, I'm kind of torn about creating all of the database tables in a PR without the API implementation using them yet. It's possible we'll miss something in the schema that we'd realize only when implementing the API. Then we'd have to go back and fix the DB migration, which usually means adding another DB migration, but as long as it isn't released maybe we can get away with just fixing the existing migration without having to add a new one. Thoughts?

Fair enough. My goal is to avoid a huge PR that would be a pain to review, but these commits (as migrations) would not be so much to add to the CRUD PR. I'll save DB stuff for that.

rawlinp:
I appreciate wanting to avoid a huge PR, but I think in this case grouping DB and API changes into the same PR is ok since they're so dependent upon each other.

PR #4584, opened 2020-04-01
updated docs for LetsEncrypt endpoint

### LINKED ISSUE #4478: deliveryservices/sslkeys/generate/letsencrypt doesn't mention deliveryservice key
## I'm submitting a ...

- [X] bug report
- [X] improvement request (usability, performance, tech debt, etc.)
- [ ] other 

- [ ] CDN in a Box
- [X] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When making a `POST` request to `deliveryservices/sslkeys/generate/letsencrypt` you must have either `key` or `deliveryservice`, representing a delivery service xmlid. 

If you have both in your request body then their values need to match. The documentation should also mention this.

The documentation does not mention the `deliveryservice` key, only the `key` key. The documentation should be updated or if the `deliveryservice` key should not be allowed then this should be updated accordingly.
## Anything else:

https://traffic-control-cdn.readthedocs.io/en/latest/api/v2/deliveryservices_sslkeys_generate_letsencrypt.html

mattjackson220:

- [x] This PR fixes #4478 

This PR updates the documentation for the Let's Encrypt endpoint to allow a 'deliveryservice' field in the POST as well as the 'key' but notes that if both are provided then they must match

- Documentation

Verify the docs build correctly and that the documentation for the deliveryservices/sslkeys/generate/letsencrypt endpoint is updated

This is a doc fix so no tests of CHANGELOG updates are necessary

### DIFF
--- a/docs/source/api/v2/deliveryservices_sslkeys_generate_letsencrypt.rst
+++ b/docs/source/api/v2/deliveryservices_sslkeys_generate_letsencrypt.rst
@@ -29,11 +29,8 @@ Generates an SSL certificate and private key using Let's Encrypt for a :term:`De
 
 Request Structure
 -----------------
-:key:             The :ref:`ds-xmlid` of the :term:`Delivery Service` for which keys will be generated
-:deliveryservice: The :ref:`ds-xmlid` of the :term:`Delivery Service` for which keys will be generated
-
-.. NOTE:: Either the ``key`` or the ``deliveryservice`` field must be provided. If both are provided, then they must match.
-
+:key:             The :ref:`ds-xmlid` of the :term:`Delivery Service` for which keys will be generated [#needOne]_
+:deliveryservice: The :ref:`ds-xmlid` of the :term:`Delivery Service` for which keys will be generated [#needOne]_
 :version:         An integer that defines the "version" of the key - which may be thought of as the sequential generation; that is, the higher the number the more recent the key
 :hostname:        The desired hostname of the :term:`Delivery Service`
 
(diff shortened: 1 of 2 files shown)


docs/source/api/v2/deliveryservices_sslkeys_generate_letsencrypt.rst
```diff
@@ -30,6 +30,10 @@ Generates an SSL certificate and private key using Let's Encrypt for a :term:`De
 Request Structure
 -----------------
 :key:             The :ref:`ds-xmlid` of the :term:`Delivery Service` for which keys will be generated
+:deliveryservice: The :ref:`ds-xmlid` of the :term:`Delivery Service` for which keys will be generated
+
+.. NOTE:: Either the ``key`` or the ``deliveryservice`` field must be provided. If both are provided, then they must match.
```
ocket8888, line 35: This should probably be a footnote instead, because it applies to multiple things and the way it is now it's breaking the list. So the generated output is something like:
```
<field list>
  <list item>...</list item>
  <list item>...</list item>
</field list>
<admonition>...</admonition>
<field list>
  <list item>...</list item>
  <list item>...</list item>
</field list>
```
so things aren't all part of the same list. But they oughtta be.

mattjackson220: done

PR #4596, opened 2020-04-03
Add support for server search server side

### LINKED ISSUE #4354: Improve performance (pagination and search) of TP's tables
## I'm submitting a ...

- [ ] bug report
- [ ] other 

This would be an improvement for TP and a potential enhancement to the TO API.

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Currently TP fetches ALL records (i.e. servers, delivery services, etc) from the TO API and does the pagination and search on the client-side in the browser. As the number of records grow, the amount of time required by jquery datatables to render the TP tables and the amount of browser memory required to do this grows. Eventually that memory may run out and TP tables may become unusable.

## Expected / new behavior:
2 possible solutions:
1. Implement a better table library
2. push the pagination and search to the server-side (the API) and leveraging some of the existing as well as some new features of the API.

Create thousands of records (i.e. servers) and navigate to https://tp.domain.net/#!/servers and watch how long it takes for the table to render.

mhoppa:

- [x] This PR fixes #4354 

This does not completely fix the issue but it adds server side search on servers.

- Documentation
- Traffic Control Client 
- Traffic Ops

There is API tests and unit tests to test the new search ability but to manually test it you need to set up the TO server with the PR and make requests to `GET 2.0/servers`.

The search parameters work as follows:

`search` - comma separated list of server attributes to search 
`searchValue` - string value to search with (its contains search)
`searchOperator` - optional operator to use on the search. Can either be OR or AND. The default is OR meaning any of the search attributes must contain the searchValue to return

## Additional Information

This is a stop gap solution to alleviate the load on TP as business grow their servers causing a tremendous bad user experience.

Going forward TO should adopt its own query language for searching that is not so rigid much like http://www.persvr.org/rql/ as an example


traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -242,6 +242,30 @@ func getServers(params map[string]string, tx *sqlx.Tx, user *auth.CurrentUser) (
 		"dsId":             dbhelpers.WhereColumnInfo{"dss.deliveryservice", nil},
 	}
 
+	searchQueryParamsToSQLCols := map[string]dbhelpers.WhereColumnInfo{
```
rob05c, line 245: We should make sure all the columns we're allowing search on have indexes, to prevent performance issues and accidental or intentional DDOS

mitchell852:
2 things i like about this feature:

1. "contains" search: this allows you to do a contains search - ?search=hostName&searchValue=foo - where this does not: ?hostName=foo

^^ maybe this could even be changed at some point to include regex?

2. OR search: this allows you to do an OR search - ?search=hostName,domainName&searchValue=foo&searchOperator=OR - where this does not: ?hostName=foo&domainName=foo

mitchell852:
closing

PR #4598, opened 2020-04-03
Allow negative limit to receive all records back

### LINKED ISSUE #3867: TO API ?limit=-1 is not supported
## I'm submitting a ...
- [ ] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Pagination was added to the TO api for some endpoints which allows the user to specify these query params:

- limit
- offset (starts at 1. imo should start at 0 but is what it is)
- page

however, a limit of -1 is not supported. if you want no limit, you can simply leave off the `limit` param but it makes it harder to do programatically. 

GET api/1.4/asns?offset=1&limit=-1 <-- returns an error

## Expected / new behavior:
GET api/1.4/asns?offset=1&limit=-1 <-- should NOT return an error and should return everything.

GET api/1.4/asns?offset=10&limit=-1 returns an error

mhoppa:

- [x] This PR fixes #3867 

- Traffic Ops

Make a request with a limit of -1 and make sure all records are returned.

### DIFF
--- a/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
+++ b/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
@@ -157,12 +157,16 @@ func BuildWhereAndOrderByAndPagination(parameters map[string]string, queryParams
 	if limit, exists := parameters["limit"]; exists {
 		// try to convert to int, if it fails the limit parameter is invalid, so return an error
 		limitInt, err := strconv.Atoi(limit)
-		if err != nil || limitInt < 1 {
-			errs = append(errs, errors.New("limit parameter must be a positive integer"))
+		if err != nil || limitInt < -1 {
+			errs = append(errs, errors.New("limit parameter must be bigger then -1"))
 			return "", "", "", queryValues, errs
 		}
 		log.Debugln("limit: ", limit)
-		paginationClause += " " + limit
+		if limitInt == -1 {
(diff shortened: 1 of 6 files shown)


traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -157,12 +157,16 @@ func BuildWhereAndOrderByAndPagination(parameters map[string]string, queryParams
 	if limit, exists := parameters["limit"]; exists {
 		// try to convert to int, if it fails the limit parameter is invalid, so return an error
 		limitInt, err := strconv.Atoi(limit)
-		if err != nil || limitInt < 1 {
-			errs = append(errs, errors.New("limit parameter must be a positive integer"))
+		if err != nil || limitInt < -1 {
+			errs = append(errs, errors.New("limit parameter must be bigger then -1"))
```
rawlinp, line 161: `bigger then` -> `bigger than` (applies to the rest of the PR as well)

PR #4605, opened 2020-04-06
Stats over http parse

ocket8888:
- [x] This PR relates to the [Additional Server Interfaces blueprint](https://github.com/apache/trafficcontrol/blob/master/blueprints/multi-interface-servers.md)

This PR implements parsing of the `stats_over_http`-format statistics in Traffic Monitor, which allows for multiple reported network interfaces.

This PR also needed to overhaul the Traffic Monitor plugin system a bit - which I hope was fine on the basis of no one I've talked to about it having ever heard of the undocumented beastie. The problem was that it wanted me to provide things like a raw `/proc/net/dev` line to read network interfaces out of later in the main routines. But the thing is that `stats_over_http` provides all of the information it parses out of that line as their own, separate numbers. Now, I could either overhaul the plugin system to be more generic - which seemed like something we'd want to do anyway once `astats` was truly deprecated - or I could take the pre-parsed stats and create a fake `/proc/net/dev` line to pass into the main routines only for them to spend more time re-parsing them again. Which seemed like maybe just not an option in such a performance-intensive component.

The plugin system is documented in a GoDoc comment, and I plan to open a PR soon to properly document it in our "dev" docs section.

- Documentation
- Traffic Monitor

Run the associated tests, start CDN-in-a-Box and verify caches are properly polled.

- [x] This PR includes tests


lib/go-tc/crstates.go
```diff
@@ -25,8 +25,8 @@ import (
 
 // CRStates includes availability data for caches and delivery services, as gathered and aggregated by this Traffic Monitor. It is designed to be served at an API endpoint primarily for Traffic Routers (Content Router) to consume.
 type CRStates struct {
-	Caches          map[CacheName]IsAvailable                       `json:"caches"`
-	DeliveryService map[DeliveryServiceName]CRStatesDeliveryService `json:"deliveryServices"`
+	Caches          map[string]IsAvailable             `json:"caches"`
```
rob05c, line 28: How strongly do you feel about all these `string` changes?

So, there were a couple reasons I started using strong typing. 

For one, it self-documents `map[string]string`. I got tired of `//map[CacheName]DeliveryServiceName` everywhere. Worse, if the comments aren't there, it's not always obvious. Even here, `Caches map[string]IsAvailable` could actually be a map[DeliveryServiceName]IsAvailable` for some reason. It's unlikely, but `tc.CacheName` makes it clear for sure.

The other big reason, was that I kept getting them wrong. I'd pass a Cache Group name to something that needed a Cache name. I know that sounds ridiculous and unlikely, but in all the complexity of things like parsing DS stats, I found myself doing it, multiple times. And worse, the failure often isn't obvious. A stat just doesn't get populated, or a lookup thinks a DS doesn't exist, and things _mostly_ work.

That might seem ridiculous, but I really was doing it, repeatedly. Maybe I'm just not as clever as some people, but this kind of thing really helped me at least, to avoid subtle and sometimes very-hard-to-catch errors at IMO a relatively small verbosity cost.

Would you object strenuously if we left these in, and kept using the strong types, to make the code a little more self-documenting and help catch subtle runtime bugs like this?

lib/go-tc/crstates.go
```diff
@@ -25,8 +25,8 @@ import (
...
+	DeliveryService map[string]CRStatesDeliveryService `json:"deliveryServices"`
```
alficles, line 29: So, I haven't done a full review, just a once-over of the code mostly looking at how it interfaces with ATS, but why did this change? The CacheName and DeliveryServiceName types generally improve readability and safety, since you'd have to cast to and from them in order to convert them. They help put arguments in context and improve the information in compile errors. I'm curious what the reasoning here.

alficles: /facepalm. I'll leave this up, but it looks like I commented at the precise moment that Rob did. He said basically the same thing, so feel free to respond only to his comments.

ocket8888: Yeah, that was exactly my plan

lib/go-tc/crstates.go
ocket8888: I do really wish that this wasn't done so widely. Really I just wanted to use strings in the code that I was writing, because I found all the casting back and forth obnoxious. But I got so frustrated with recursive compile errors (x is a string now, so you can't pass a cachename as the variable anymore, and then that cachename value from a map can't be assigned to the variable that gets passed into the function etc.) that I threw up my hands and did a search/replace on the whole thing. Which is why `tc.CacheGroupName` is unchanged everywhere - because it wasn't in the code I wrote.

That really does seem pretty ridiculous, but like I said I didn't have to touch anything with Cache Groups, so maybe it wouldn't if I had to deal with more complex things.

I can try to take as much of it out as possible, but I would still like the code I wrote to use strings. Just seems sensible to me. And it really does suck that I had to touch code in `/lib` and `/traffic_ops/traffic_ops_golang`.

ocket8888: Alright, I put it back in in most places.

PR #4621, opened 2020-04-08
Reduce Weasel exceptions

zrhoffman:

This PR:
*  Adds the Apache License Version 2.0 to the following files:
    - cron jobs
    - .dependency_license
    - .gitattributes
    - gitignores
    - .mailmap
    - traffic_portal/.jshintrc - It is JSON-C, so it can handle comments
    - traffic_router/core/.pmd
    - traffic_router/core/src/main/conf/COMODOHigh-AssuranceSecureServerCA.crt
* Removes CSVs, .classpath, .test, and specific TTF files from patterns that Weasel does not check

- CDN in a Box
- Traffic Control Client - Python, Java
- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Traffic Portal
- Traffic Router

- Build affected components
- Run weasel

Not a bug fix.

- [x] Only comments are added, no tests are necessary

### DIFF
--- a/vendor/gopkg.in/fsnotify.v1/.gitignore
+++ b/vendor/gopkg.in/fsnotify.v1/.gitignore
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
 # Setup a Global .gitignore for OS and editor generated files:
 # https://help.github.com/articles/ignoring-files
 # git config --global core.excludesfile ~/.gitignore_global
--- a/vendor/gopkg.in/fsnotify.v1/.gitignore
+++ b/vendor/gopkg.in/fsnotify.v1/.gitignore
@@ -1,20 +1,3 @@
-# Licensed to the Apache Software Foundation (ASF) under one
-# or more contributor license agreements.  See the NOTICE file
-# distributed with this work for additional information
-# regarding copyright ownership.  The ASF licenses this file
-# to you under the Apache License, Version 2.0 (the
-# "License"); you may not use this file except in compliance
-# with the License.  You may obtain a copy of the License at
-#
-#   http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing,
-# software distributed under the License is distributed on an
-# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-# KIND, either express or implied.  See the License for the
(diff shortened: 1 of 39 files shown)


ocket8888:
So I'm not sure about this (or it'd be in a review, not a comment) but I *think* that git (`.gitignore`, `.gitkeep`, `.gitattributes` etc.) files are exempt from needing the header.

zrhoffman:
They were. .gitattributes was exempt until 7c182b6309, gitignores were exempt until 74169894ab.

ocket8888:
No I mean exempt as in they aren't legally required to have license headers. I can't remember why I think that, though, so it might be totally wrong.

zrhoffman:
Not exempt. After adding a .gitignore on this branch and running Weasel:
```
Error                                 Unknown! needs-license/.gitignore
```

rob05c:
>No I mean exempt as in they aren't legally required to have license headers. I can't remember why I think that, though, so it might be totally wrong.

https://www.apache.org/legal/src-headers.html

> What files in an Apache release do not require a license header?
> 
> A file without any degree of creativity in either its literal elements or its structure is not protected by copyright law; therefore, such a file does not require a license header. If in doubt about the extent of the file's creativity, add the license header to the file.
> 
> Other files may make sense to have no license header. Three examples are:
> 
>     Short informational text files; for example README, INSTALL files. The expectation is that these files make it obvious which product they relate to.
>     Test data for which the addition of a source header would cause the tests to fail.
>     'Snippet' files that are combined as form a larger file where the larger file would have duplicate licensing headers.
> 
> PMCs should use their judgement, err on having a source header and contact legal-discuss@ if unsure. 

I believe files like .gitignore fall under "without any degree of creativity."

That said, our general policy has been to put license headers in all files that are possible, and since .gitignore and .gitattributes support comments, I'd vote we do, and "err on having a source header."

zrhoffman:
IMO, "without any degree of creativity" does not cover files with comments, which our repo's root gitignore has.

Licensing gitignores definitely isn't a *must*. That said, there seems to be [a number of repos](https://github.com/search?q=filename:.gitignore+"Apache+License"+org:apache&type=Code) within @apache alone that license their gitignores, [including apache/tomcat](https://github.com/apache/tomcat/blob/master/.gitignore).

ocket8888:
I guess I don't care if we add headers to files that don't actually need them. It is good to know that I was right about exemptions, though. Sort of sounds like we don't even need license headers in our TP templates/controllers/services either...

alficles:
Yeah, the bar for "no degree of creativity" is pretty high, I think. Adding the headers is easy and makes it clear. The first job of the license header is to let people know precisely what rights they do and do not have to use a particular project, file or bit of code. This all looks like progress forward to me.

vendor/gopkg.in/fsnotify.v1/.gitignore
```diff
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
rawlinp, line 1: This isn't technically our code since it's a vendored dependency -- are we allowed to add the license here?

rob05c: Nope. Good catch. We shouldn't be touching vendored stuff, other people's code has its own licensing documentation in LICENSE.

Moreover, touching vendored stuff runs the risk of creating a Derivative Work, which has its own legal and copyright implications and we may or may not have the rights to do. So we should avoid modifying files we don't own if at all possible.

zrhoffman: My mistake, I meant to exclude all vendored files from the gitignore commit, I must have missed one. Needless to say, Weasel doesn't mind if there's no license there. Removed in 2f75f3a3fa.

zrhoffman:
Since @alficles maintains weasel and approved, did anyone else have thoughts on this?

PR #4625, opened 2020-04-08
updated to store Lets Encrypt user account information

mattjackson220:

This PR updates the Let's Encrypt integration to store user account information in order to use it later and avoid LE's rate limits.

- Documentation
- Traffic Ops

Run this version of TO in an environment that is publicly accessible.  Ensure that an email address has been provided for Let's Encrypt and that it is pointed to the LE staging environment in the cdn.conf file like so:

    "lets_encrypt" : {
        "user_email" : "youremail",
        "send_expiration_email": false,
        "convert_self_signed": false,
        "renew_days_before_expiration": 30,
        "environment": "staging"
    }

Go to an https enabled delivery service and generate new certs using Let's Encrypt.  Verify that it works and that the account information is now stored in the lets_encrypt_account table.  Re-generate new certificates and verify that it 
1) successfully gets new certs 
2) does NOT store anything in the dnschallenges table (when using a user account for a domain for which you've already proven ownership this is not necessary)
3) a debug level alert was logged in TO saying "Found existing account with Let's Encrypt"

Then regenerate new certs over 50 times.  In staging the rate limit is 50 new accounts per 3 hours from a single IP address so if it goes over 50 successfully then it has successfully used your account to bypass the rate limit.

note 50 sounds really high but in their prod environment its 10 so that is why this is important

This PR does not require test or CHANGELOG updates as it is an upgrade / bug fix for existing code
- [x] This PR does not require an update to CHANGELOG.md 


docs/source/admin/traffic_router.rst
```diff
@@ -715,19 +715,20 @@ Let's Encrypt can be set up through :ref:`cdn.conf` by updating the following fi
...
-	+------------------------------+---------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+	+------------------------------+---------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+	| Name                         | Type    | Description                                                                                                                                                            |
+	+==============================+=========+========================================================================================================================================================================+
+	| user_email                   | string  | Required. Email to create account with Let's Encrypt or to receive expiration updates.                                                                                 |
```
zrhoffman, line 721: This is a good time to create a required/optional column.

docs/source/admin/traffic_ops.rst
```diff
@@ -341,7 +341,7 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 
 	.. versionadded:: 4.1
 
-	:user_email: An optional email address to create an account with Let's Encrypt or to receive expiration updates
+	:user_email: A requred email address to create an account with Let's Encrypt or to receive expiration updates. If this is not included then `rate limits <https://letsencrypt.org/docs/rate-limits>`_ may apply for the number of certificates.
```
rawlinp, line 344: `requred` -> `required`

traffic_ops/app/db/migrations/20200408000000_add_lets_encrypt_account_info.sql
```diff
@@ -0,0 +1,29 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+
+-- lets_encrypt_account
+CREATE TABLE IF NOT EXISTS lets_encrypt_account (
+  email text NOT NULL,
```
rawlinp, line 21: we should make `email` the PK here

traffic_ops/traffic_ops_golang/deliveryservice/letsencryptcert.go
```diff
@@ -330,7 +374,86 @@ func GetLetsEncryptCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEn
...
+		}
+		if rows == 0 {
+			return errors.New("zero rows affected when inserting Let's Encrypt account information")
+		}
+	}
+	return nil
+}
+
+type LEInfo struct {
+	Email      *string `db:"email"`
```
rawlinp, line 455: Do any of these fields actually have to be nullable? The DB fields are `NOT NULL`, and it looks like `PrivateKey` could set by dereferencing the `decodedKey` pointer on L432.

traffic_ops/traffic_ops_golang/deliveryservice/letsencryptcert.go
```diff
@@ -330,7 +374,87 @@ func GetLetsEncryptCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEn
...
+
 	api.CreateChangeLogRawTx(api.ApiChange, "DS: "+*req.DeliveryService+", ID: "+strconv.Itoa(dsID)+", ACTION: Added SSL keys with Lets Encrypt", currentUser, logTx)
 
 	return nil
 }
+
+func getStoredLetsEncryptInfo(tx *sql.Tx, email string) (*LEInfo, error) {
+	leInfo := LEInfo{}
+	selectQuery := `SELECT email, private_key, uri FROM lets_encrypt_account WHERE email = $1 LIMIT 1`
+	rows, err := tx.Query(selectQuery, email)
```
rawlinp, line 410: Since we only expect to get 0 or 1 row back, we can use `QueryRow` here and simplify this a little bit. Basically like this: https://github.com/apache/trafficcontrol/blob/f01019a7c72bae1212fc2f08f31177305a58741c/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go#L448

Also, the error message on L412 looks like a copy pasta

mattjackson220: done!

PR #4628, opened 2020-04-09
Add ORT Rewrite Blueprint

rob05c:
Adds a ORT Rewrite proposal Blueprint

No tests, it's a blueprint.
No docs, it's a blueprint.
No changelog, it's a blueprint.

None, it's a blueprint.

Read the blueprint, verify you like it.

Not a bug fix, it's a blueprint.

### DIFF
--- a/blueprints/ort-rewrite-unix-style.md
+++ b/blueprints/ort-rewrite-unix-style.md
@@ -1,0 +1,209 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+# ORT Rewrite in UNIX Philosophy
+
+## Problem Description
+ORT is:
+- Difficult to maintain. Writing Perl is difficult, and reading it is even more difficult.
+- Dangerous to modify. Perl is not compiled, and even validity checks (`perl –c`) fail to verify dynamic runtime errors. This makes it very easy to introduce a bug in seldom-executed areas.
+- Untested. Perl ORT has no unit or integration tests.
+- Opaque. Nobody really knows everything it does, or when, or why.
+
+## Proposed Change
+
+ORT will be rewritten into a series of standalone executables, in the "UNIX Philosophy"
+
+> 1. Make each program do one thing well. To do a new job, build afresh rather than complicate old programs by adding new "features".
+> 2. Expect the output of every program to become the input to another, as yet unknown, program. Don't clutter output with extraneous information. Avoid stringently columnar or binary input formats. Don't insist on interactive input.
+
+- Each executable should do exactly 1 thing, and if a new "thing" becomes necessary, a new executable will be created.
+- The input and output of executables should be text which is easily parseable, so the executables can easily be pipelined (passing the output of one to the input of another), as well as easily read by humans and manipulated by standard Linux/POSIX tools.
+
+This makes ORT:
+- Easier to maintain. Each binary does one thing, is much smaller, and is more obvious. Presumably they’re also written in a language easier to read and write, such as Go.
+- Safer to modify. If each component is smaller, it’s more obvious what it does. We also presume the apps will be written with good development practices (such as modularization), with a language which verifies more at compile-time, and with tests.
+- Clear and easy for operators to understand what each app does. We assume clean interfaces, and good documentation (ideally in the app itself, via help flags, printing usage when no arguments are received, and/or man pages).
+
+#### Implementation
+
+The implementation should adhere to the "UNIX Philosophy," POSIX, Linux Standard Base (LSB), and GNU as much as possible.
+
+ORT will continue to consist of a single OS package (e.g. RPM), which installs all executables.
+
+ORT will require the following executables:
+- **Aggregator**. This is the “primary application” which will emulate the existing ORT script, and be called by CRON or operators to deploy all configs, as ORT does today. Note this is similar to how git works, and several other common Linux CLI utilities. Will include a "Report Mode" which executes the pipeline of commands necessary to emulate the existing ORT Report Mode.
+  This app will have no logic itself, except to call the other executables.
+    - INPUT: configuration and specification to fetch and emplace config files.
+    - BEHAVIOR: fetches and places config files
+    - OUTPUT: success or failure message
+
+- **Traffic Ops Requestor**. This will fetch data needed from Traffic Ops, such as the Update Pending flag, packages, etc. This should never modify TO data, and should be guaranteed read-only. Any status modifications should go in the Traffic Ops Updater.
+    - INPUT: Traffic Ops URL and credentials, and data to fetch
+    - BEHAVIOR: Requests data from Traffic Ops
+    - OUTPUT: Traffic Ops data requested
+        - Format is probably multipart/mixed, but format may be different if a better format is determined. Ideal "UNIX Philosophy" format is line-delimited text, but the complexity may preclude that. The more complex and difficult to parse, the further from the "UNIX Philosophy." E.g. multipart/mixed is preferable to JSON.
+- **Config File Generator**. This will take TO data and produce config files.
+    - INPUT: Traffic Ops data, and config file(s) to generate or exclude (typically all, possibly “reval only” or other behaviors of ORT)
+    - BEHAVIOR: No side effects. Computationally: builds requested files.
+    - OUTPUT: Config files
+        - Format is probably multipart/mixed, but as above, may be different, multipart is preferable to JSON, etc.
+- **Config File Preprocessor**. Preprocesses generated config files, making post-generation modifications such as underscore directive replacements, and remap OVERRIDE replacements. TODO: determine if this should be rolled into the Config File Generator.
+    - INPUT: Config files, TO data/metadata.
+    - BEHAVIOR: No side effects; computationally: modifies files per rules and data.
+    - OUTPUT: modified config files.
+- **Server Config Readiness Verifier**. Verifies the operating system is ready and safe to apply the given config. Currently, this is just checking udev rules and verifying ATS block devices don’t have filessytems. But it may be more in the future. TODO: determine if necessary; should this even be ORT’s job? Will we ever need anything besides udev/50-ats.rules? 
+    - INPUT: config files
+    - BEHAVIOR: No side effects; reads configs and inspects server state.
+    - OUTPUT: whether server is safe and ready to apply config files.
+- **ATS Plugin Readiness Verifier**. Verifies ATS has the necessary plugins for the config files.
+    - INPUT: config files
+    - BEHAVIOR: No side effects; reads configs and inspects server state.
+    - OUTPUT: whether ATS has all necessary plugins, or which files require which missing plugins.
+- **Diff Tool**. This will take two config files (presumably an existing and new file) and return their differences. TODO: determine if this is necessary; if configs are deterministic, can POSIX diff be used?
+    - INPUT: Two config files
+    - BEHAVIOR: No side effects. Computationally: diffs given files.
+    - OUTPUT: file diff. Ideally in a standard format.
+- **Backup Tool**. This will take a file and copy it to a backup location. TODO: determine if necessary; are standard POSIX cp/mv/etc enough?
+    - INPUT: config file to backup, and backup location. May be text of new file or path of existing file.
+    - BEHAVIOR: Save or copy given file to given backup location. Should be atomic.
+    - OUTPUT: success or failure message.
+- **Restart Determiner**. Takes the config to be applied (only the changed files, after diffing), encapsulates the logic of what changes require a reload or restart, and returns whether a restart, reload, or neither is required.
+    - INPUT: config files to be applied.
+    - BEHAVIOR: no side effects; computationally: inspects files and determines action.
+    - OUTPUT: whether ATS needs reloaded or restarted.
+- **Service Reloader**. Takes the name of the service (possibly only ATS) to reload or restart, and reloads or restarts as necessary. TODO determine if necessary; is this any logic beyond calling service restart and/or traffic_ctl?
+    - INPUT: service, whether to reload or restart
+    - BEHAVIOR: reloads or restarts the service
+    - OUTPUT: success or failure
+- **Traffic Ops Updater**. This will set the server’s update status in Traffic Ops. This should be only the Update and Reval Pending flags; ORT should never modify server configuration data, only ever server configuration status data.
+    - INPUT: Traffic Ops URL and credentials, and status to set
+    - BEHAVIOR: Makes a POST request to TO setting the status
+    - OUTPUT: success or failure message
+
+#### Features Omitted
+
+The following features of the current ORT are specifically not being implemented in the redesign:
+
+- **Chkconfig**. Chkconfig is not used by CentOS 7+; specifically, SystemD does not use it. It is misleading that ORT sets it today.
+- **Ntpd**. ORT currently has custom logic to restart ntpd if an ntpd.conf is changed. This should be managed by whatever system is managing the server (Ansible/Puppet/Manual/etc). Network time should not be the responsibility of Traffic Control or its config applicator. 
+- **Interactive mode**. This mode is rarely possibly never used today. Further, by dividing ORT into UNIX-style apps for each function, an operator can easily see what results would be before running them.
+- **Revalidate Mode**. ORT is now fast enough to make a separate Revalidate unnecessary. It should always check and apply all files.
+- **Package Installation**. ORT will cease to perform this. OS (RPM) package installation will no longer be done by Traffic Control, but rather by whoever or whatever is managing the machine and operating system (Ansible, Puppet, human system administrators, etc).
+    - Whatever is managing the other hundreds of packages on the operating system should also manage ATS and its plugins. ORT's job is to manage Traffic Control configuration data, not the operating system.
+
+#### Additional Utilities.
+
+Shell scripts which are “one-liners” pipelining common operations should be provided with the OS Package. TODO: add a list of scripts under Implementation heading.
+
+Additionally, a .pl script which emulates the existing ORT behavior will be provided in the old location, to preserve backwards-compatibility. This script should be very small, and essentially translate old calls and flags to the new Aggregator.
+
+#### Notes
+
+I started to list requirements for all apps, such as unit tests, integration tests, modular design, argument/manpage for usage info, etc. But then I realized I was just listing good design principles. So I decided to omit that.
+
+### Traffic Portal Impact
+None.
+
+### Traffic Ops Impact
+None.
+
+#### REST API Impact
+None.
+
+#### Client Impact
+None.
+
+#### Data Model / Database Impact
+None.
+
+### ORT Impact
+Completely rewrites ORT. Backward-compatibility for safe upgrades will be preserved.
+
+The Interface will be similar, but compatibility is not a goal. Further, LSB-compliant options and parameters is a goal, and will require incompatibility.
+
+A `traffic_ops_ort.pl` script will be provided, whose interface _does_ preserve backward compatibility, and calls the new ORT "aggregator". This will preserve existing CRON or other tools users are using with ORT, and make an upgrade not break a production system.
+
+### Traffic Monitor Impact
+None.
+
+### Traffic Router Impact
+None.
+
+### Traffic Stats Impact
+None.
+
+### Traffic Vault Impact
+None.
+
+### Documentation Impact
+Rewrite will provide MAN pages for ORT. ORT arguments are not currently documented, and that will not be changed, to avoid duplicate documentation. TC "read-the-docs" may include a small comment pointing users to the MAN page.
+
+### Testing Impact
+
+ORT does not currently include an Integration Test framework. Creating one is orthogonal to this project, but this project is a good opportunity to do so at the same time.
+
+ORT Integration Tests need exactly the same thing as the Traffic Ops API tests: a running TO instance populated with data. It is suggested that an ORT Integration Test framework copy or abstract the TO API Tests.
+
+### Performance Impact
+Not significant. ORT in master as of this writing takes less than a minute to run, on a large CDN with at least 1000 Delivery Services and 1000 Servers. The Rewrite should still take less than a minute to run.
+
+It should be a goal for ORT to take less than 15 seconds to run, given a fast Traffic Ops. Because a faster ORT means TC changes propogate faster, which is ideal. But ORT is not in the Request Path, so performance is not critical to TC operation. Further, the previous release of ORT took 5-8 minutes on a large CDN.
+
+### Security Impact
+None.
+
+### Upgrade Impact
+None. The rewrite will provide a compatibility script, so existing tools and runs continue to work as before.
+
+### Operations Impact
+Operators should learn the new real ORT interface, to better operate Traffic Control. However, it's not strictly required, they can continue to use the compatibility script for the immediate future.
+
+### Developer Impact
+Developers on ORT will have to learn the new codebase and write in Go.
+
+The new code should not be logically larger or more complex than the existing ORT. The Go language is vertically verbose, but this project does not significantly increase the actual logic, or add new logical features.
+
+## Alternatives
+There are innumerable ways ORT could be rewritten. Some possibilities:
+- Transliterated in-place to another language.
+    - Disadvantages: Monolithic, more difficult to develop, more difficult to understand each component and "thing" done, more difficult to understand all things done.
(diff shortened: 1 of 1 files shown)


ocket8888:
Your blueprint doesn't mention a decided language - though offers Go as a suitable candidate.
For what it's worth, I'd be happy to help move ORT.py in this direction. It has the advantage of having transliterated most of ORT into separate modules - which could make dividing it into separate utilities simpler than starting from scratch.

On the other hand, if the plan is also to include the necessary parts of ATSTCCFG as part and parcel of the utilities (as opposed to the separate utility it is today) then Go is probably a better choice. Because the configuration file generation is the hardest part to get right.

rob05c:
>On the other hand, if the plan is also to include the necessary parts of ATSTCCFG as part and parcel of the utilities (as opposed to the separate utility it is today) 

The plan is for each app to be small and independent. They could be totally different languages if we want. The config generator is one app, and will probably be a modified version of `atstccfg`. But there's nothing preventing the other apps from being in other languages.

>I'd be happy to help move ORT.py

I'm not opposed to that. Personally, I'd prefer we were more open to more diverse languages. Especially for small apps like these, where it's not a big deal if a new primary developer doesn't like the language, they can just rewrite it, because it's very small.

I also don't object to Python specifically, especially for small things, especially if it includes type annotations and compiling in the build.

If you're the one writing these apps, I wouldn't have any objections. But I'd prefer to leave it up to whoever writes it. Personally, Go would be faster for me and I already have a good grasp of traffic_ops_ort.pl at this point. But I may not be the one writing it, either.

>parts of ATSTCCFG

I'd strongly encourage referring to the command in its proper case, atstccfg. In technical writing it's imperative to not confuse the reader. The application is `atstccfg`, executing `./atstccfg` will not work.

I'm not aware of a universally-accepted technical writing style guide, but the Microsoft Style Guide is one of the most popular. It says 
https://docs.microsoft.com/en-us/style-guide/capitalization

>Capitalize the first word of labels and terms that appear in UI and APIs unless they're always lowercase (for example, fdisk).
> In programming languages, follow the traditional capitalization of keywords and other special terms.

And

>Always capitalize the first word of a new sentence. Rewrite sentences that start with a word that's always lowercase.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+
+ORT will require the following executables:
+- **Aggregator**. This is the “primary application” which will emulate the existing ORT script, and be called by CRON or operators to deploy all configs, as ORT does today. Note this is similar to how git works, and several other common Linux CLI utilities.
+  This app will have no logic itself, except to call the other executables.
+    - INPUT: configuration and specification to fetch and emplace config files.
+    - BEHAVIOR: fetches and places config files
+    - OUTPUT: success or failure message
+
+- **Traffic Ops Requestor**. This will fetch data needed from Traffic Ops, such as the Update Pending flag, packages, etc. This should never modify TO data, and should be guaranteed read-only. Any status modifications should go in the Traffic Ops Updater.
+    - INPUT: Traffic Ops URL and credentials, and data to fetch
```
alficles, line 57: There are other important details about fetching data, like what sorts of timeouts, proxies, or caches should be used. Should these be inputs to the requestor or generic configuration you find elsewhere. Similarly, should credentials be passed each time or should they be available in a protected file the caller can access?

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+    - BEHAVIOR: Requests data from Traffic Ops
+    - OUTPUT: Traffic Ops data requested
+        - Format is probably multipart/mixed, but format may be different if a better format is determined. Ideal "UNIX Philosophy" format is line-delimited text, but the complexity may preclude that. The more complex and difficult to parse, the further from the "UNIX Philosophy." E.g. multipart/mixed is preferable to JSON.
```
alficles, line 60: I disagree. JSON or similar should generally be preferable to multipart/mixed because there are better and more plentiful tools for transforming it. And if the objects returned may be large, perhaps application/json-seq is a good job. (See [RFC 7464](https://tools.ietf.org/html/rfc7464).) Or even [JSONLines](http://jsonlines.org/).

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+- **Config File Generator**. This will take TO data and produce config files.
+    - INPUT: Traffic Ops data, and config file(s) to generate or exclude (typically all, possibly “reval only” or other behaviors of ORT)
+    - BEHAVIOR: No side effects. Computationally: builds requested files.
+    - OUTPUT: Config files
+        - Format is probably multipart/mixed, but as above, may be different, multipart is preferable to JSON, etc.
```
alficles, line 65: You say it outputs config files. Does it output them to disk or just stdout to be written by the caller? I ask because you mention multipart/mixed again. And something needs to do things like determining file locations, file permissions, file update strategy (truncate and write vs. new file and copy), backup strategy, and managing parallel processing for performance.

It feels like this component should probably be called like `cat to-data | ort-cfg /etc/trafficserver/records.config` and be expected to manage all except the parallel processing, which the caller would manage by limiting the number of to-data calls.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+- **Config File Preprocessor**. Preprocesses generated config files, making post-generation modifications such as underscore directive replacements, and remap OVERRIDE replacements. TODO: determine if this should be rolled into the Config File Generator.
+    - INPUT: Config files, TO data/metadata.
+    - BEHAVIOR: No side effects; computationally: modifies files per rules and data.
+    - OUTPUT: modified config files.
```
alficles, line 69: IMO, roll this into the generator. And maybe remove as much that depends on it as possible. :)

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+- **Server Config Readiness Verifier**. Verifies the operating system is ready and safe to apply the given config. Currently, this is just checking udev rules and verifying ATS block devices don’t have filessytems. But it may be more in the future. TODO: determine if necessary; should this even be ORT’s job? Will we ever need anything besides udev/50-ats.rules? 
+    - INPUT: config files
+    - BEHAVIOR: No side effects; reads configs and inspects server state.
+    - OUTPUT: whether server is safe and ready to apply config files.
```
alficles, line 73: Could this also do things like checking to be sure that the server is not currently ONLINE or is otherwise fit for service? Might it verify that the OS matches what TO expects to deploy into? I think this might do other things, but some of them might need a read-only TO credential.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+- **ATS Plugin Readiness Verifier**. Verifies ATS has the necessary plugins for the config files.
+    - INPUT: config files
+    - BEHAVIOR: No side effects; reads configs and inspects server state.
+    - OUTPUT: whether ATS has all necessary plugins, or which files require which missing plugins.
```
alficles, line 77: I assume this uses the return code to signal failure?

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+- **Diff Tool**. This will take two config files (presumably an existing and new file) and return their differences. TODO: determine if this is necessary; if configs are deterministic, can POSIX diff be used?
+    - INPUT: Two config files
+    - BEHAVIOR: No side effects. Computationally: diffs given files.
+    - OUTPUT: file diff. Ideally in a standard format.
```
alficles, line 81: Even if configs are deterministic, a diff output for humans will be very valuable. And POSIX diff on things like YAML is basically useless.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+- **Backup Tool**. This will take a file and copy it to a backup location. TODO: determine if necessary; are standard POSIX cp/mv/etc enough?
+    - INPUT: config file to backup, and backup location. May be text of new file or path of existing file.
+    - BEHAVIOR: Save or copy given file to given backup location. Should be atomic.
+    - OUTPUT: success or failure message.
```
alficles, line 85: Feels like standard tools are probably the right call here. Maybe a backup file name generator, though? It really kinda depends what the backup strategy is.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+- **Restart Determiner**. Takes the config to be applied (only the changed files, after diffing), encapsulates the logic of what changes require a reload or restart, and returns whether a restart, reload, or neither is required.
+    - INPUT: config files to be applied.
+    - BEHAVIOR: no side effects; computationally: inspects files and determines action.
+    - OUTPUT: whether ATS needs reloaded or restarted.
+- **Service Reloader**. Takes the name of the service (possibly only ATS) to reload or restart, and reloads or restarts as necessary. TODO determine if necessary; is this any logic beyond calling service restart and/or traffic_ctl?
+    - INPUT: service, whether to reload or restart
+    - BEHAVIOR: reloads or restarts the service
+    - OUTPUT: success or failure
```
alficles, line 93: This tool could be clever about the operating system it's working on. Not sure it's totally necessary for a first release, though.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+- **Interactive mode**. This mode is rarely possibly never used today. Further, by dividing ORT into UNIX-style apps for each function, an operator can easily see what results would be before running them.
+- **Revalidate Mode**. ORT is now fast enough to make a separate Revalidate unnecessary. It should always check and apply all files.
+- **Report Mode**. The Aggregator itself does not need a Report Mode. Operators wanting to see what would be changed can run the Traffic Ops Requestor, Config File Generator, Diff Tool, Restart Determiner, and Package Install Determiner.
+- **Package Installation**. ORT will cease to perform this. OS (RPM) package installation will no longer be done by Traffic Control, but rather by whoever or whatever is managing the machine and operating system (Ansible, Puppet, human system administrators, etc).
+    - Whatever is managing the other hundreds of packages on the operating system should also manage ATS and its plugins. ORT's job is to manage Traffic Control configuration data, not the operating system.
```
alficles, line 109: I mentioned this on the list, so I won't go in detail here, but I think this should be considered for inclusion.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Additional Utilities.
+
+Shell scripts which are “one-liners” pipelining common operations should be provided with the OS Package. TODO: add a list of scripts under Implementation heading.
+
+Additionally, a .pl script which emulates the existing ORT behavior will be provided in the old location, to preserve backwards-compatibility. This script should be very small, and essentially translate old calls and flags to the new Aggregator.
```
alficles, line 115: This is a critically important stepping stone in the process, but it may prove to be a pain to maintain in the future. Let's make sure we're clear that this tool is deprecated via a formal deprecation notice with plenty of fanfare so folks are aware. We don't want this shim to live forever.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+None.
+
+### Traffic Stats Impact
+None.
+
+### Traffic Vault Impact
+None.
+
+### Documentation Impact
+Rewrite will provide MAN pages for ORT. ORT arguments are not currently documented, and that will not be changed, to avoid duplicate documentation. TC "read-the-docs" may include a small comment pointing users to the MAN page.
```
alficles, line 156: Consider having man pages generated from Markdown so that they can easily be read from both Github directly, the file system of a developer working on them, and via man after install.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+It should be a goal for ORT to take less than 15 seconds to run, given a fast Traffic Ops. Because a faster ORT means TC changes propogate faster, which is ideal. But ORT is not in the Request Path, so performance is not critical to TC operation. Further, the previous release of ORT took 5-8 minutes on a large CDN.
+
+### Security Impact
+None.
+
+### Upgrade Impact
+None. The rewrite will provide a compatibility script, so existing tools and runs continue to work as before.
+
+### Operations Impact
+Operators should learn the new real ORT interface, to better operate Traffic Control. However, it's not strictly required, they can continue to use the compatibility script for the immediate future.
```
alficles, line 176: We should be clear about when they'll need to upgrade to the new systems. It's important that operators have a timeline for their transition.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Developer Impact
+Developers on ORT will have to learn the new codebase and write in Go.
+
+The new code should not be logically larger or more complex than the existing ORT. The Go language is vertically verbose, but this project does not significnatlly increase the actual logic, or add new logical features.
```
alficles, line 181: s/significnatlly/significantly/

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,210 @@
...
+    - More power, can receive requests from Traffic Ops, can run irrespective of CRON or operator deployment.
+- Removed, changed to a server orchestration tool like Ansible or Puppet.
+    - Disadvantages: radical change, makes TC dependent on an orchestration tool and platform, tool will require nearly as much code to do the same thing in its own instruction language, requires operators to learn a particular and very large and complex third-party tool, requires even small CDN Operators to deploy a large orchestration tool they don't otherwise need to learn or deploy, requires CDN operators with a different orchestration tool to deploy and learn multiple large redundant systems.
+    - Advantages: Simpler, an orchestration config language is (hopefully) simpler than a Turing-Complete app. Easier for Operators to learn, if they're already using the orchestrator everywhere else.
+
+## Dependencies
+None anticipated. Small Go library dependencies may be needed, which will be considered as they arise. Any dependencies should be compile-time.
```
alficles, line 199: Yeah, it will definitely have compile-time dependencies unless you plan to write an HTTP client in whatever language you use. At a minimum, it will depend on the standard library. All that should be fine, though. It's hard to guess which dependencies you'll need from the get-go.

blueprints/ort-rewrite-unix-style.md
rob05c: The Blueprint doesn't go in that level of detail. I'm +1 on both of those.

I think it's implied "all behavior not specifically stated as being removed will be preserved"

I see the Blueprint as the larger design idea, for community approval. Small features like that, I'd vote we make as regular Github Issues.

blueprints/ort-rewrite-unix-style.md
rob05c: IMO JSON on the CLI strongly violates the UNIX Philosophy. Unix apps should return columnar text, for ease of parsing, and to interoperate with other tools.

multipart is very not-ideal either; but we need a way to delimit files. You can almost do two newlines, but you at least need the metadata of the file name, and other metadata is really useful too. Which, if you were making a custom format, leads you to something very similar to multipart.

JSON is a tree. It's nice that it's a simple language, but trees are drastically more complex than lines.

I wish there were something closer to the standard UNIX line-delimiting, but I don't think there is. Any tree format is way more complex to parse, search, hold in memory, etc. It basically makes operating on with the standard UNIX toolset (awk/grep/sed/etc) impossible, without `jq`. Where multipart is at least possible to use the standard tools on, if slightly more complex from the two delimiter levels.

You're right, JSON is more popular than multipart. But multipart is really just a delimiter and headers, it's super-easy to parse. And most languages do have libraries for it.

It took me maybe an hour to do it in Perl, language I'm very much not an expert in:
https://github.com/apache/trafficcontrol/blob/5693451bb898f62d0e14afa60a9af9824d3c8e06/traffic_ops/ort/traffic_ops_ort.pl#L1600
And e.g. Go has libraries for it: https://golang.org/pkg/mime/multipart/

I could not write a JSON parser in an hour. And no, I won't need to. But, it shouldn't even be a possibility. _Whatever_ has to be parsed should be as easy, human readable, and line-delimited as possible.

blueprints/ort-rewrite-unix-style.md
rob05c: The config generator outputs to stdout. The Aggregator that's acting as a single "ORT run" will then write them to disk (after comparing, diffing, etc).

We could have another tool for that, to take config files as input and write them to files. So your example would be `cat to-data | ort-cfg | ort-write-files`. That didn't seem necessary to me, but I'm not strongly opposed.

blueprints/ort-rewrite-unix-style.md
rob05c: It could. I'm not opposed, but I'd vote we put details like that in Github Issues, and keep this design proposal higher-level.

>I think this might do other things, but some of them might need a read-only TO credential.

If so, the TO Data Fetcher should get that data.

blueprints/ort-rewrite-unix-style.md
rob05c: Yes, and also empty output probably means success, and non-empty output probably means an error. I trust the implementer to do the thing is most in the Unix Philosophy. And to document it in the man page.

blueprints/ort-rewrite-unix-style.md
rob05c: Not opposed to a smarter diff. But: 

>POSIX diff on things like YAML is basically useless.

Why? YAML is line-based. And even if a config weren't, the operator can and should format their code on lines, so the diff does make sense. I can't imagine a config needing to be minified

blueprints/ort-rewrite-unix-style.md
rob05c: >Maybe a backup file name 

I assume the backup filename is input to this app, and the Aggregator can pick that. The Aggregator could even read a TO Parameter (the answer to all problems) if we wanted. Though that feature doesn't exist today.

blueprints/ort-rewrite-unix-style.md
rob05c: Agree, with both points.

blueprints/ort-rewrite-unix-style.md
rob05c: +1
In fact, the ideal would be if the same file is both the README.md in the app's directory for Github, and used to generate the man. But IMO man and not duplicating text should take priority.

blueprints/ort-rewrite-unix-style.md
rob05c: Fixed

blueprints/ort-rewrite-unix-style.md
jrushford: I was thinking that the "Server Config Readiness Verifier" shouldn't do too much beyond verifying as Chris mentioned that the server was not ONLINE and leave checking UDEV rules and OS things to other tools such as Ansible but the  "Server Config Readiness Verifier" should definitely insure that the proper ATS RPM's are installed.

blueprints/ort-rewrite-unix-style.md
ocket8888: Sphinx has an output mode for manual pages. So the official docs for the utility could also be its man pages.

hbeatty:
@rob05c 

Hello Rob,

Removing the ability to run a "report" mode is problematic for us. We use the current report mode to verify the config on select servers before applying. I understand that each component will have its own "report" mode but, having an overall "report" mode makes it simpler for our operations team. Is there a way to get an overall "report?"

Thanks,
Hank

rob05c:
@hbeatty Sure. You're not the first person who's expressed an interest in keeping Report Mode.

It should be easy enough to include in the "Aggregator" as a dedicated command for the existing functionality.

rob05c:
Ok, I added Report Mode to the list of things the Aggregator will do, and removed it from the "features removed" section.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,209 @@
...
+ORT will continue to consist of a single OS package (e.g. RPM), which installs all executables.
+
+ORT will require the following executables:
+- **Aggregator**. This is the “primary application” which will emulate the existing ORT script, and be called by CRON or operators to deploy all configs, as ORT does today. Note this is similar to how git works, and several other common Linux CLI utilities. Will include a "Report Mode" which executes the pipeline of commands necessary to emulate the existing ORT Report Mode.
+  This app will have no logic itself, except to call the other executables.
+    - INPUT: configuration and specification to fetch and emplace config files.
+    - BEHAVIOR: fetches and places config files
+    - OUTPUT: success or failure message
+
+- **Traffic Ops Requestor**. This will fetch data needed from Traffic Ops, such as the Update Pending flag, packages, etc. This should never modify TO data, and should be guaranteed read-only. Any status modifications should go in the Traffic Ops Updater.
```
ocket8888, line 56: is `to_access` sufficient for this? I don't wanna get to bogged down in an app spec right now, just wondering if there are things it doesn't do that this "ORT component" needs to be able to do.

blueprints/ort-rewrite-unix-style.md
ocket8888: JSON is just plain text, as far as anything reading it from stdin knows. Maybe you just want sequential requests like `toget /servers /deliveryservices` and that outputs "servers payload -> deliveryservices payload" on separate lines?

blueprints/ort-rewrite-unix-style.md
ocket8888: I think it'd be neat if this just accepted a configuration file name and output its contents. That'd make it really simple, easy to test, and would let you add extensions for e.g. nginx configuration files without needing to change how the program interfaces with others. No need for JSON or multipart.

blueprints/ort-rewrite-unix-style.md
ocket8888: agree - down with `## OVERRIDE ##`.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,209 @@
...
+    - INPUT: Traffic Ops data, and config file(s) to generate or exclude (typically all, possibly “reval only” or other behaviors of ORT)
+    - BEHAVIOR: No side effects. Computationally: builds requested files.
+    - OUTPUT: Config files
+        - Format is probably multipart/mixed, but as above, may be different, multipart is preferable to JSON, etc.
+- **Config File Preprocessor**. Preprocesses generated config files, making post-generation modifications such as underscore directive replacements, and remap OVERRIDE replacements. TODO: determine if this should be rolled into the Config File Generator.
+    - INPUT: Config files, TO data/metadata.
+    - BEHAVIOR: No side effects; computationally: modifies files per rules and data.
+    - OUTPUT: modified config files.
+- **Server Config Readiness Verifier**. Verifies the operating system is ready and safe to apply the given config. Currently, this is just checking udev rules and verifying ATS block devices don’t have filessytems. But it may be more in the future. TODO: determine if necessary; should this even be ORT’s job? Will we ever need anything besides udev/50-ats.rules?
```
ocket8888, line 70: I agree that this is weird, and I really hope it's not necessary.

blueprints/ort-rewrite-unix-style.md
ocket8888: YAML diffs being useless feels like it's based on the assumption that config outputs are _not_ deterministic. Which I hope won't be the case.

Currently the only reason POSIX diff doesn't work is because (besides lingering non-determinism) the files all come with that header that says "generated by x on y date". I hope we can reduce that to just `generated by x` now because the file modification time should tell you the "when", and that'll mean the output is truly only dependent on TO state (which is arguably an input so "stateless" may apply). So all non-zero POSIX diffs would be real.

I guess what I'm saying is: whether or not POSIX diff can be used instead comes down to the implementation of the configuration file generator. If we can declare in this blueprint that it be deterministic and change the header format a bit, then we can say that POSIX diff can be used.

blueprints/ort-rewrite-unix-style.md
ocket8888: Please god no more obscure parameters.

I agree that standard tools are the right call.

blueprints/ort-rewrite-unix-style.md
```diff
@@ -0,0 +1,209 @@
...
+    - OUTPUT: file diff. Ideally in a standard format.
+- **Backup Tool**. This will take a file and copy it to a backup location. TODO: determine if necessary; are standard POSIX cp/mv/etc enough?
+    - INPUT: config file to backup, and backup location. May be text of new file or path of existing file.
+    - BEHAVIOR: Save or copy given file to given backup location. Should be atomic.
+    - OUTPUT: success or failure message.
+- **Restart Determiner**. Takes the config to be applied (only the changed files, after diffing), encapsulates the logic of what changes require a reload or restart, and returns whether a restart, reload, or neither is required.
+    - INPUT: config files to be applied.
+    - BEHAVIOR: no side effects; computationally: inspects files and determines action.
+    - OUTPUT: whether ATS needs reloaded or restarted.
+- **Service Reloader**. Takes the name of the service (possibly only ATS) to reload or restart, and reloads or restarts as necessary. TODO determine if necessary; is this any logic beyond calling service restart and/or traffic_ctl?
```
ocket8888, line 90: What other services are managed by this today? Besides ntpd, which I think we all agree shouldn't be done.

blueprints/ort-rewrite-unix-style.md
ocket8888: It wouldn't need package manager access to do that, just call `trafficserver --version`

ocket8888:
So a lot of those comments were meant to be added to existing conversations and don't make sense without that context. Thanks GitHub.

PR #4634, opened 2020-04-14
Fixing the query to get mid servers

### LINKED ISSUE #3472: Inaccurate query being used to fetch mid servers
I don't believe this query used in this function is quite right:

https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/server/servers.go#L303

It basically evaluates to this:

```
SELECT *
FROM server AS s
WHERE s.id IN (
    SELECT mid.id FROM server mid
    JOIN cachegroup cg ON cg.id IN (
        SELECT cg.parent_cachegroup_id
        FROM server s
        JOIN cachegroup cg ON cg.id = s.cachegroup
        WHERE s.id IN (1,2,3))
    JOIN type t ON mid.type = (SELECT id FROM type WHERE name = 'MID'))
```

which regardless of which server id is fed into it always seems to return ALL the mid servers...

I think it needs to be like this:

```
SELECT *
FROM server AS s 
JOIN type AS t ON s.type = t.id
WHERE t.name = 'MID' AND s.cachegroup IN (
    SELECT cg.parent_cachegroup_id FROM cachegroup AS cg 
    WHERE cg.id IN (
        SELECT s.cachegroup FROM server AS s 
        WHERE s.id IN (1,2,3)))
```

This does these things to fetch the mids associated with a list of edges:

1. gets the cache groups of the edges
2. get the parent cachegroups for those cachegroups (found in 1)
3. get the servers that belong to those  parent cachegroups (found in 2) that are mids

Basically the query is meant to answer the question "if my ds uses these edges, which mids does my ds use based on the relationship of the cache groups"

srijeet0406:

- [x] This PR fixes #3472 

- Traffic Ops

Run the tests in traffic_ops_golang and make sure they all run successfully.
Also, I tested it in CiaB in the following way:
Consider the following configuration.
You have a DS in your CiaB, connected to one Edge Cache Server (say, E1).
Now, say that this Edge Cache Server belongs to group CG1.
Consider there is another mid tier cache group CG2, that is listed as CG1's parent.
Add a new mid cache group (with a new or existing mid tier cache server) that is NOT the parent of any of the edge caches.
Now, run the query without the fix, and by substituting the ID of your edge cache (for your DS, in this case, E1).
Your result should return the details of both the Mid Tier Cache Servers (WRONG).
Now, run it with the fix, and you should just get the parent mid tier cache server as the answer (CORRECT).

- master
- 4.0.1 (RC1)

- [x] This PR includes tests 
- [x]  I have explained why documentation is unnecessary

### DIFF
--- a/traffic_ops/traffic_ops_golang/server/servers.go
+++ b/traffic_ops/traffic_ops_golang/server/servers.go
@@ -335,14 +335,12 @@ func getMidServers(servers []tc.ServerNullable, tx *sqlx.Tx) ([]tc.ServerNullabl
 	edgeIDs := strings.Join(ids, ",")
 	// TODO: include secondary parent?
 	q := selectQuery() + `
-WHERE s.id IN (
-	SELECT mid.id FROM server mid
-	JOIN cachegroup cg ON cg.id IN (
-		SELECT cg.parent_cachegroup_id
-		FROM server s
-		JOIN cachegroup cg ON cg.id = s.cachegroup
-		WHERE s.id IN (` + edgeIDs + `))
-	JOIN type t ON mid.type = (SELECT id FROM type WHERE name = 'MID'))
+JOIN type AS t ON s.type = t.id
+WHERE t.name = 'MID' AND s.cachegroup IN (
+SELECT cg.parent_cachegroup_id FROM cachegroup AS cg
+WHERE cg.id IN (
+SELECT s.cachegroup FROM server AS s
+WHERE s.id IN (` + edgeIDs + `)))
 `
 	rows, err := tx.Queryx(q)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -335,14 +335,12 @@ func getMidServers(servers []tc.ServerNullable, tx *sqlx.Tx) ([]tc.ServerNullabl
...
 	q := selectQuery() + `
-WHERE s.id IN (
-	SELECT mid.id FROM server mid
-	JOIN cachegroup cg ON cg.id IN (
-		SELECT cg.parent_cachegroup_id
-		FROM server s
-		JOIN cachegroup cg ON cg.id = s.cachegroup
-		WHERE s.id IN (` + edgeIDs + `))
-	JOIN type t ON mid.type = (SELECT id FROM type WHERE name = 'MID'))
+JOIN type AS t ON s.type = t.id
```
mitchell852, line 338: I'm getting the following error when trying to hit the endpoint that calls this query:

```
GET /api/2.0/servers?dsId=22 HTTP/1.1" 400
pq: table name "t" specified more than once
```

Issue #4412, opened 2020-02-19
deliveryservices/sslkeys/add with no SSL keys returns Internal Server Error


zrhoffman:
## I'm submitting a ...

- [x] bug report
- [ ] new feature / enhancement request
- [ ] improvement request (usability, performance, tech debt, etc.)
- [ ] other 

## Traffic Control components affected ...

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

Traffic Ops panics if SSL keys are not provided in the POST body to `api/1.4/deliveryservices/sslkeys/add`.

Specifically, it panics dereferencing `*r.AuthType ` when `AuthType` is `nil` [at deliveryservices_ssl_keys.go:139](https://github.com/apache/trafficcontrol/blob/18fe13ac63/lib/go-tc/deliveryservice_ssl_keys.go#L139).

```go
		if r.Certificate.CSR == "" && *r.AuthType != LetsEncryptAuthType {
```

## Expected / new behavior:

Validate `DeliveryServiceSSLKeysReq` and check for `nil` `AuthType` to avoid dereferencing `nil`.

## Minimal reproduction of the problem with instructions:

1. Rebuild the CDN-in-a-Box's `trafficops-go` Docker image at commit 18fe13ac63 or later

2. Start CDN-in-a-Box, wait for Traffic Ops to be available, POST the body given below to `deliveryservices/sslkeys/add` with the Python client for Traffic Ops (from CDN-in-a-Box startup [in `to-access.sh`](https://github.com/apache/trafficcontrol/blob/18fe13ac63/infrastructure/cdn-in-a-box/traffic_ops/to-access.sh#L305), except with empty SSL key files), and pay attention to the `trafficops` container's logs:
    ```shell script
    topost deliveryservices/sslkeys/add '{ "cdn": "CDN-in-a-Box", "certificate": { "crt": "", "csr": "", "key": "" }, "deliveryservice": "demo1", "hostname": "demo2", "key": "demo1", "version": 1 }'
    ```

## Anything else:

*
    <details><summary>Stack trace (click to expand)</summary>

    ```go
[...]

PR #4638, opened 2020-04-15
Fix nil pointer dereference in sslkeys/add validation

rawlinp:
If a request does not contain a CSR _and_ `authType` is nil, TO will dereference the nil pointer and panic, causing it to return a 500 internal server error. This fixes the validation to check for a nil pointer and return a proper 400 bad request.

- Traffic Ops

Run the unit tests, verify they pass.

Example `curl` request that would 500 before but should now 400:
```
curl -kvs -H "$MOJO_COOKIE" -XPOST -d'{"key":"foo-ds","version":1,"certificate":{"crt":"","key":"","csr":""},"expiration":"0001-01-01T00:00:00Z","hostname":"*.foo-ds.mycdn.example.com","cdn":"foo-cdn","deliveryservice":"foo-ds"}' "https://<insert TO FQDN here>/api/2.0/deliveryservices/sslkeys/add"
```

- master

- [x] This PR includes tests
- [x] Bugfix, docs not necessary
- [x] Bug was not released, so changelog is not necessary

### DIFF
--- a/lib/go-tc/deliveryservice_ssl_keys.go
+++ b/lib/go-tc/deliveryservice_ssl_keys.go
@@ -64,7 +64,7 @@ type DeliveryServiceSSLKeysV15 struct {
 }
 
 type DeliveryServiceSSLKeysReq struct {
-	AuthType        *string `json:authType,omitempty`
+	AuthType        *string `json:"authType,omitempty"`
 	CDN             *string `json:"cdn,omitempty"`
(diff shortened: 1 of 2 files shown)


MylesBock:
This appears to resolve #4412 as well

rawlinp:
@MylesBock thank you, you are correct. I will close out #4412.

PR #4643, opened 2020-04-17
Dsstats rename

ocket8888:

This PR changes the misleading response field name `totalBytes` in responses
from `/deliveryservice_stats` to the more appropriate `totalKiloBytes` in
API version 2, and updates the documentation accordingly.

This PR depends on #4642.

- Documentation
- Traffic Ops

Run a CDN, verify that stats output is correct, all tests pass, and it returns
`totalBytes` in API v1 but `totalKiloBytes` in API v2.

- [x] Tests for endpoints depending on external services don't exist

### DIFF
--- a/traffic_ops/traffic_ops_golang/trafficstats/deliveryservice.go
+++ b/traffic_ops/traffic_ops_golang/trafficstats/deliveryservice.go
@@ -186,9 +186,9 @@ func GetDSStats(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if inf.Version.Major > 1 {
-		handleLegacyRequest(w, r, client, c, inf)
-	} else {
 		handleRequest(w, r, client, c, inf)
+	} else {
+		handleLegacyRequest(w, r, client, c, inf)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


traffic_ops/traffic_ops_golang/trafficstats/deliveryservice.go
```diff
@@ -185,44 +185,52 @@ func GetDSStats(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	resp := tc.TrafficDSStatsResponseV1{
-		Source:  tc.TRAFFIC_STATS_SOURCE,
-		Version: tc.TRAFFIC_STATS_VERSION,
+	if inf.Version.Major > 1 {
+		handleLegacyRequest(w, r, client, c, inf)
```
rawlinp, line 189: this seems backwards

ocket8888: boy does it

mitchell852:
@ocket8888 - can you just limit this PR to the name change - totalBytes to totalKiloBytes - and save the refactoring bit for api 3.x?

ocket8888:
That's all I did. I have to make the name change in only 2.x, which means we need different struct tags for each API version, which means we need different structs, which means we need to write different objects to the response depending on the API version. I changed how `getDSSummary` works to avoid copy/pasting the function - the amount of lines changed would have been even greater if I hadn't.

ocket8888:
Done

PR #4644, opened 2020-04-17
Remove cacheurl from Delivery Services

### LINKED ISSUE #1125: cache url expression configs do not get added to mid profiles
When we add a cache url expression to a delivery service in Traffic Ops the config files only get added to the edge profiles. We have to remember to manually add them to the mid profiles other wise ORT will have issues and the configs do not get put on the mids.

### LINKED ISSUE #3225: Remove CacheURL Delivery Service Field

shamrickus:

- [x] This PR begins #3225 and fixes #1125.

This does not finish #3225 as CacheURL exists in ATC5 and TO API v3, so until those are no longer supported CacheURL cannot be completely removed.

- CDN in a Box
- Documentation
- Traffic Control Client 
- Traffic Ops
- Traffic Portal
- Traffic Router

Ensure db migration works
Verify CDN still works (specificially TO, TR, TP)
Verify modify/delete/create of DS in TP/TO


lib/go-atscfg/remapdotconfig.go
```diff
@@ -244,10 +232,7 @@ func BuildRemapLine(cacheURLConfigParams map[string]string, atsMajorVersion int,
 			if _, globalExists := cacheURLConfigParams["location"]; globalExists {
 				log.Warnln("Making remap.config for Delivery Service '" + ds.Name + "': qstring_ignore == 1, but global cacheurl.config param exists, so skipping remap rename config_file=cacheurl.config parameter")
 			} else {
-				qstr, addedCacheURL, addedCacheKey := GetQStringIgnoreRemap(atsMajorVersion)
-				if addedCacheURL {
-					hasCacheURL = true
-				}
+				qstr, _, addedCacheKey := GetQStringIgnoreRemap(atsMajorVersion)
```
alficles, line 235: If you're dropping support for cacheurl, you're dropping support for ATS < 6. GetQStringIgnoreRemap will need to be modified to return a static string.

lib/go-atscfg/cacheurldotconfig.go
```diff
@@ -101,11 +97,10 @@ func MakeCacheURLDotConfig(
 	// TODO verify prefix and suffix exist, and warn if they don't? Perl doesn't
 	dsName := tc.DeliveryServiceName(strings.TrimSuffix(strings.TrimPrefix(fileName, "cacheurl_"), ".config"))
 
-	ds, ok := dses[dsName]
+	_, ok := dses[dsName]
 	if !ok {
 		return text // TODO warn? Perl doesn't
 	}
-	text += ds.CacheURL + "\n"
```
alficles, line 108: This is the only text in the file. And you removed the references to cacheurl.config and cacheurl_qstring.config. Those configs are for the plugin you're removing support for. We should consider removing this whole function, I think. And then look at what's calling it and remove up the chain.

lib/go-tc/deliveryservices.go
```diff
@@ -198,7 +198,6 @@ type DeliveryServiceNullableV11 struct {
 	//
 	Active                   *bool                   `json:"active" db:"active"`
 	AnonymousBlockingEnabled *bool                   `json:"anonymousBlockingEnabled" db:"anonymous_blocking_enabled"`
-	CacheURL                 *string                 `json:"cacheurl" db:"cacheurl"`
```
alficles, line 201: This breaks backward compatibility at a code level. Shouldn't this stay present but empty here and only be removed in later versions of the type? As it is, code that uses the V11 struct will break.

shamrickus: You are correct, I did not mean to commit this change.

lib/go-atscfg/remapdotconfig.go
ocket8888: I didn't realize we still supported ATS < 6 in general

lib/go-atscfg/remapdotconfig_test.go
```diff
@@ -585,209 +578,6 @@ func TestMakeRemapDotConfigDuplicateOrigins(t *testing.T) {
 	}
 }
 
-func TestMakeRemapDotConfigNilMidRewrite(t *testing.T) {
```
ocket8888, line 588: Shouldn't this test still occur, just without setting/checking a CacheURL value?

lib/go-atscfg/remapdotconfig_test.go
```diff
@@ -585,209 +578,6 @@ func TestMakeRemapDotConfigDuplicateOrigins(t *testing.T) {
...
-		t.Errorf("expected no edge header rewrite text for mid server, actual '%v'", txt)
-	}
-
-	if strings.Contains(remapLine, "hdr_rw_") {
-		t.Errorf("expected no edge header rewrite for mid server, actual '%v'", txt)
-	}
-
-}
-
-func TestMakeRemapDotConfigMidHasNoEdgeRewrite(t *testing.T) {
```
ocket8888, line 694: Ditto above question

lib/go-atscfg/remapdotconfig_test.go
shamrickus: I originally had removed this test because the Nil config returns nothing (ignoring comments), I've readded it and just asserted that only the comment is present

lib/go-atscfg/remapdotconfig_test.go
shamrickus: See above

lib/go-atscfg/cacheurldotconfig.go
shamrickus: I believe I have properly removed up the chain, might be some I missed in atstccfg though.

shamrickus:
> So to be clear, this PR removes _support_ for the cacheurl, but does not remove it from the API? Like it'll always show it as `null` in responses because if you do a POST with some cacheurl value that doesn't get unmarshaled and doesn't get stored in the database, but from what I can see the structures marshaled for responses still have it.
Only when querying old API  versions on an update server. V3 should not have these fields while v1/v2 should.

traffic_ops/traffic_ops_golang/ats/atscdn/cacheurldotconfig.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"net/http"
+
+	"github.com/apache/trafficcontrol/lib/go-rfc"
+)
+
+func GetCacheURLDotConfig(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set(rfc.ContentType, rfc.ContentTypeTextPlain)
+	w.Write([]byte("# https://github.com/apache/trafficcontrol/issues/3225"))
+}
```
ocket8888: I think you should just remove the route instead

ocket8888:
So like I was saying before, this removes _support_ for `cacheurl`, but does not actually remove it from the API? I can still see it in v3:
```http
GET /api/3.0/deliveryservices?limit=1 HTTP/1.1
User-Agent: python-requests/2.22.0
Accept: */*
Connection: keep-alive
Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1OTI5NTcyNjgsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--625c11141971227d3862de47c92810086ebd7389

HTTP/1.1 200 OK
Content-Encoding: gzip
Set-Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1OTI5MzkyNjgsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--784bea340410815bda72048f7fa1c6cc6520e590; Path=/; Expires=Tue, 23 Jun 2020 19:07:48 GMT; Max-Age=3600; HttpOnly
Vary: Accept-Encoding
Whole-Content-Sha512: 94qFFF5yps/vfOFe/E3M285fBWJzgmpylXwwdGf4xa8w/J32nPboY2jgd2Ehi52jt1pK0zYtOc+Vtud3f/rVSA==
Date: Tue, 23 Jun 2020 18:07:48 GMT
Content-Length: 806

{
	"response": [
		{
			"active": true,
			"anonymousBlockingEnabled": false,
			"cacheurl": null,
			"ccrDnsTtl": null,
			"cdnId": 2,
			"cdnName": "CDN-in-a-Box",
			"checkPath": null,
			"displayName": "Demo 1",
			"dnsBypassCname": null,
			"dnsBypassIp": null,
			"dnsBypassIp6": null,
			"dnsBypassTtl": null,
			"dscp": 0,
			"edgeHeaderRewrite": null,
			"geoLimit": 0,
			"geoLimitCountries": null,
			"geoLimitRedirectURL": null,
			"geoProvider": 0,
			"globalMaxMbps": null,
			"globalMaxTps": null,
			"httpBypassFqdn": null,
			"id": 1,
			"infoUrl": null,
			"initialDispersion": 1,
			"ipv6RoutingEnabled": true,
			"lastUpdated": "2020-06-23 18:04:18+00",
			"logsEnabled": true,
			"longDesc": "Apachecon North America 2018",
			"longDesc1": null,
			"longDesc2": null,
			"matchList": [
				{
					"type": "HOST_REGEXP",
					"setNumber": 0,
					"pattern": ".*\\.demo1\\..*"
				}
			],
			"maxDnsAnswers": null,
			"midHeaderRewrite": null,
			"missLat": 42,
			"missLong": -88,
			"multiSiteOrigin": false,
			"originShield": null,
			"orgServerFqdn": "http://origin.infra.ciab.test",
			"profileDescription": null,
			"profileId": null,
			"profileName": null,
			"protocol": 2,
			"qstringIgnore": 0,
			"rangeRequestHandling": 0,
			"regexRemap": null,
			"regionalGeoBlocking": false,
			"remapText": null,
			"routingName": "video",
			"signed": false,
			"sslKeyVersion": null,
			"tenantId": 1,
			"type": "HTTP",
			"typeId": 1,
			"xmlId": "demo1",
			"exampleURLs": [
				"http://video.demo1.mycdn.ciab.test",
				"https://video.demo1.mycdn.ciab.test"
			],
			"deepCachingType": "NEVER",
			"fqPacingRate": null,
			"signingAlgorithm": null,
			"tenant": "root",
			"trResponseHeaders": null,
			"trRequestHeaders": null,
			"consistentHashRegex": null,
			"consistentHashQueryParams": [
				"abc",
				"pdq",
				"xxx",
				"zyx"
			],
			"maxOriginConnections": 0,
			"ecsEnabled": false,
			"rangeSliceBlockSize": null,
			"topology": null
		}
	]
}
```

shamrickus:
With my original changes that didn't remove the unused methods this was not possible. Since those have now been removed I can remove it from the v3 api.

ocket8888:
Seeing this in the v1 tests:
```
 deliveryservices_test.go:121: cannot UPDATE DeliveryService by ID: 500 Internal Server Error[500] - Error requesting Traffic Ops https://localhost:6443/api/1.5/deliveryservices/65.json {"alerts":[{"text":"Internal Server Error","level":"error"}]}
         - <nil>
    deliveryservices_test.go:134: results do not match actual: d s 1, expected: something different
    deliveryservices_test.go:135: results do not match actual: 0, expected: 164598
    deliveryservices_test.go:136: results do not match actual: 0, expected: 100
    deliveryservices_test.go:170: cannot UPDATE DeliveryService by ID: 500 Internal Server Error[500] - Error requesting Traffic Ops https://localhost:6443/api/1.5/deliveryservices/65.json {"alerts":[{"text":"Internal Server Error","level":"error"}]}
         - <nil>
```

shamrickus:
v1 issues should be fixed

mitchell852:
just following up on status of this. is this waiting for after 5.0 release?

shamrickus:
At the least, probably best for whatever release perl gets removed in.

shamrickus:
Should be ready for re-review. Not mentioned in the original PR but it is good to also test compatibility with old versions.

lib/go-tc/deliveryservices.go
```diff
@@ -226,11 +225,8 @@ type DeliveryServiceNullableV12 struct {
 // for all fields to be null.
 // TODO move contents to DeliveryServiceNullableV12, fix references, and remove
 type DeliveryServiceNullableV11 struct {
-	// NOTE: the db: struct tags are used for testing to map to their equivalent database column (if there is one)
-	//
 	Active                   *bool                   `json:"active" db:"active"`
 	AnonymousBlockingEnabled *bool                   `json:"anonymousBlockingEnabled" db:"anonymous_blocking_enabled"`
-	CacheURL                 *string                 `json:"cacheurl" db:"cacheurl"`
```
ocket8888, line 233: It looks like you re-added a commit where you took `CacheURL` out of V11 DS structs - there was a previously resolved comment about this so I think it was possibly a rebase issue?

ocket8888 (commented):
I think this should only edit the `tc-fixtures.json` in `v4/` - also, why'd you delete the `userdeliveryservices_test.go` file in `v1/`?

Also, I think you already know this, but the docs have a bunch of references to the now-undefined `ds-cacheurl` label.

lib/go-tc/deliveryservices.go
shamrickus: It was intentional, I readded it for backwards compatibility with older TC servers, but the `json` serializer handles it for us such that the presence or absence of that field does not break anything.

ocket8888: It doesn't break anything as far as JSON serialization, but it is a breaking change to old client call signatures. Changes to old client call signatures must be non-breaking.

traffic_ops/client
```diff
@@ -1 +1 @@
-v3-client
\ No newline at end of file
+v4-client
```
ocket8888, line 1: We can actually just remove this symlink now. It's use was deprecated in the last release.

lib/go-tc/deliveryservices.go
```diff
@@ -173,35 +172,95 @@ type DeliveryServiceV11 struct {
...
-	DeliveryServiceV30
-	MaxRequestHeaderBytes *int `json:"maxRequestHeaderBytes" db:"max_request_header_bytes"`
-}
-
 // DeliveryServiceNullableV30 is the aliased structure that we should be using for all api 3.x delivery structure operations
 // This type should always alias the latest 3.x minor version struct. For ex, if you wanted to create a DeliveryServiceV32 struct, you would do the following:
 // type DeliveryServiceNullableV30 DeliveryServiceV32
 // DeliveryServiceV32 = DeliveryServiceV31 + the new fields
 type DeliveryServiceNullableV30 DeliveryServiceV31
+type DeliveryServiceNullableV40 DeliveryServiceV40
```
ocket8888, line 245: Could we not do this for this API version? We don't need a struct with "Nullable" in the name. It's only there for V1 to distinguish it from the existing `DeliveryService` struct that was not "nullable". The DeliveryServiceV40 struct is already "nullable", so there's no reason to specify.

ocket8888 (commented):
Would you mind adding GoDoc comments to the new exported symbols?

lib/go-tc/deliveryservices.go
shamrickus: Sure can, was just copy and pasting.

traffic_ops/client
shamrickus: Wouldn't we need to keep it until API v3 is no longer used/ATS v5.0 is unsupported?

ocket8888: No, use of the un-versioned `client` import was deprecated in the last major release specifically so we could remove it now. Each version-specific library remains unbroken, you just have to use the correct path. Versioned paths have existed since 4.0 iirc, so people have had plenty of time to stop using this crutch.

lib/go-tc/deliveryservices.go
```diff
@@ -595,6 +836,16 @@ func (ds *DeliveryServiceNullableV30) Validate(tx *sql.Tx) error {
 	return util.JoinErrs(errs)
 }
 
+func (ds *DeliveryServiceV40) validateTopologyFields() error {
```
rawlinp, line 839: ditto my prior comments

traffic_ops/v4-client/deliveryservice.go
```diff
@@ -155,6 +155,19 @@ func (to *Session) GetDeliveryServicesV30WithHdr(header http.Header, params url.
 	return data.Response, reqInf, err
 }
 
+// GetDeliveryServicesV40WithHdr returns all (tenant-visible) Delivery Services that
+// satisfy the passed query string parameters. See the API documentation for
+// information on the available parameters.
+func (to *Session) GetDeliveryServicesV40WithHdr(header http.Header, params url.Values) ([]tc.DeliveryServiceV40, ReqInf, error) {
```
rawlinp, line 161: These methods should also use the `DeliveryServiceV4` alias as input/output. We should also consider implementing the new proposed TO Go client best practices (or make a mental note to come back here and change them once the groundwork is laid out for that).

traffic_ops/v4-client/deliveryservice.go
```diff
@@ -261,6 +274,61 @@ func (to *Session) GetDeliveryServiceByXMLIDNullable(XMLID string) ([]tc.Deliver
 	return ret, reqInf, err
 }
 
+func (to *Session) CreateDeliveryServiceV40(ds tc.DeliveryServiceV40) (tc.DeliveryServiceV40, ReqInf, error) {
```
rawlinp, line 277: ditto prior comment

traffic_ops/v4-client/deliveryservice.go
```diff
@@ -388,7 +456,20 @@ func (to *Session) UpdateDeliveryServiceV30WithHdr(id int, ds tc.DeliveryService
 		return tc.DeliveryServiceNullableV30{}, reqInf, fmt.Errorf("failed to update Delivery Service #%d; response indicated that %d were updated", id, len(data.Response))
 	}
 	return data.Response[0], reqInf, nil
+}
 
+// UpdateDeliveryServiceV40WithHdr replaces the Delivery Service identified by the
+// integral, unique identifier 'id' with the one it's passed.
+func (to *Session) UpdateDeliveryServiceV40WithHdr(id int, ds tc.DeliveryServiceV40, header http.Header) (tc.DeliveryServiceV40, ReqInf, error) {
```
rawlinp, line 463: ditto prior comment

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -852,9 +852,33 @@ func TopologyExists(tx *sql.Tx, name string) (bool, error) {
 	return count > 0, err
 }
 
+// CheckTopologyV30 returns an error if the given Topology does not exist or if one of the Topology's Cache Groups is
+// empty with respect to the Delivery Service's CDN.
+func CheckTopologyV30(tx *sqlx.Tx, ds tc.DeliveryServiceNullableV30) (int, error, error) {
```
rawlinp, line 857: This copy shouldn't be necessary -- all the "real work" of the DS API should be handled at the "v4 (latest) layer." I.e. the v1-3 handlers should all chain up to the v4 handler, where the request is actually handled, then converted back down to the handler at which the request came in.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -97,7 +97,7 @@ func (ds *TODeliveryService) GetType() string {
 
 // IsTenantAuthorized checks that the user is authorized for both the delivery service's existing tenant, and the new tenant they're changing it to (if different).
 func (ds *TODeliveryService) IsTenantAuthorized(user *auth.CurrentUser) (bool, error) {
-	return isTenantAuthorized(ds.ReqInfo, &ds.DeliveryServiceNullableV30)
+	return isTenantAuthorizedV30(ds.ReqInfo, &ds.DeliveryServiceNullableV30)
```
rawlinp, line 100: Version-specific functions like this shouldn't be necessary (see prior comment).

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -450,6 +472,177 @@ func createV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 t
 	return &dsV31, http.StatusOK, nil, nil
 }
 
+func createV40(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV40 tc.DeliveryServiceV40) (*tc.DeliveryServiceV40, int, error, error) {
```
rawlinp, line 475: We shouldn't have to create another copy of this function. Instead, the `createV31` function should call _this_ function, take the result, and convert it back to a V31 struct to return back to the client.

In `createV40` we don't write `CacheURL` to the DB, so that should instead be handled by the `createV31` function after getting a successful return value from `createV40`. If `EnsureParams` handles CacheURL-ey stuff as well, it would go in `createV31` alone.

Also, `createV40` should cast the `tc.DeliveryServiceV40` into the `tc.DeliveryServiceV4` alias at the beginning of the method then cast back to `tc.DeliveryServiceV40` at the end (on L479 and L642, respectively) in order for all the other functions used by this function to accept the `tc.DeliveryServiceV4` alias as input. Otherwise, every new API version requires updating all of these extra functions.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -1066,6 +1282,249 @@ func updateV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 *
 	return dsV31, http.StatusOK, nil, nil
 }
 
+func updateV40(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV40 *tc.DeliveryServiceV40) (*tc.DeliveryServiceV40, int, error, error) {
```
rawlinp, line 1285: See my prior comment about `createV40` -- the same pattern applies to `updateV40` as well.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -1740,8 +2215,21 @@ func GetDSSelectQuery() string {
 	return selectQuery()
 }
 
+// getTenantIDV30 returns the tenant Id of the given delivery service. Note it may return a nil id and nil error, if the tenant ID in the database is nil.
+func getTenantIDV30(tx *sql.Tx, ds *tc.DeliveryServiceNullableV30) (*int, error) {
```
rawlinp, line 2219: per my prior comments, this shouldn't be necessary

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -1785,6 +2273,38 @@ func isTenantAuthorized(inf *api.APIInfo, ds *tc.DeliveryServiceNullableV30) (bo
 	return true, nil
 }
 
+func isTenantAuthorizedV30(inf *api.APIInfo, ds *tc.DeliveryServiceNullableV30) (bool, error) {
```
rawlinp, line 2276: per my prior comments, this shouldn't be necessary

traffic_ops/traffic_ops_golang/server/servers_assignment.go
```diff
@@ -390,12 +389,6 @@ INSERT INTO deliveryservice_server (deliveryservice, server)
 			} else {
 				delete = append(delete, param)
 			}
-			param = getConfigFile(cacheURLPrefix, xmlID.String)
```
rawlinp, line 393: We can't get rid of this functionality yet, right? As long as the 3.1 and below APIs exist, the field still exists, and its associated functionality should be supported?

rawlinp (changes requested):
I added a bunch of comments as I went along, but it might be easier to understand if you go through all the comments first before making changes. Basically it boils down to reducing duplication and making it easier to make future changes to the DS API.

lib/go-tc/deliveryservices.go
```diff
@@ -560,6 +767,40 @@ func (ds *DeliveryServiceNullableV30) validateTypeFields(tx *sql.Tx) error {
 	return nil
 }
 
+func (ds *DeliveryServiceV40) Validate(tx *sql.Tx) error {
```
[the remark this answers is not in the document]
shamrickus: Validate is used externally (DS Requests)

traffic_ops/traffic_ops_golang/server/servers_assignment.go
shamrickus: Correct, I've readdded it

lib/go-tc/deliveryservices.go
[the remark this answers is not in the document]
rawlinp: Ah, I suppose it could be a public function in the `deliveryservice` package, which DSRs could import to use. That would still be better than it living in `lib/go-tc` IMO.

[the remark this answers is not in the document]
ocket8888: There's a TODO somewhere in the codebase about moving validator methods internal to TO; I don't think this is the only handler that has this issue. It is really weird to be requiring a transaction hooked into a TO database in the supposedly generic library code, but moving it to be internal means re-thinking the ParseValidator interface and we were thinking of doing that anyway because of its current inability to return system-level errors. So it might be something to worry about in a later change.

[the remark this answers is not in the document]
rawlinp: `TODeliveryService` doesn't implement the `ParseValidator` interface, so we are free to move this method into TO.

traffic_ops/client
ocket8888, line 1: We can actually just remove this symlink now. Its use was deprecated in the last release.

ocket8888 (approved):
LGTM

This changes DSRs to use APIv4 Delivery Services - across all API versions - which is technically a breaking change. But the only fixes are to either quadruple the DSR handling code or rewrite it to not use the CRUDer. Since the latter is already in progress, I'm inclined to let that be tomorrow's problem.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -278,14 +300,42 @@ func createV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV30 t
 	}
 	return nil, status, userErr, sysErr
 }
+func createV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 tc.DeliveryServiceV31) (*tc.DeliveryServiceV31, int, error, error) {
+	tx := inf.Tx.Tx
+	dsNullable := tc.DeliveryServiceNullableV30(dsV31)
+	ds := dsNullable.UpgradeToV4()
+	res, status, userErr, sysErr := createV40(w, r, inf, tc.DeliveryServiceV40(ds))
+	if res == nil {
```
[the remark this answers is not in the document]
shamrickus: I'm just doing what the previous versions of this function did (but flipped). It looks like if there is an error in createV40, res will always be nil, otherwise it will have a value.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -501,20 +548,22 @@ func (ds *TODeliveryService) Read(h http.Header, useIMS bool) ([]interface{}, er
 	for _, ds := range dses {
 		switch {
 		// NOTE: it's required to handle minor version cases in a descending >= manner
-		case version.Major > 3 && version.Major >= 0:
+		case version.Major > 4 && version.Minor >= 0:
```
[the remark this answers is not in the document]
shamrickus: Good catch.

rawlinp (approved):

PR #4645, opened 2020-04-20
Api v3

ocket8888:

This adds Traffic Ops API version 3.0, including all associated documentation and tests.
Also adds a v3 Go client and makes other components specify the v2 client (for compatibility)

- CDN in a Box
- Documentation
- Traffic Control Client (Go)
- Traffic Ops

Run all unit and client/api integration tests.

- [x] This PR includes tests


grove/grovetccfg/grovetccfg.go
```diff
@@ -35,7 +35,7 @@ import (
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
-	to "github.com/apache/trafficcontrol/traffic_ops/client"
+	to "github.com/apache/trafficcontrol/traffic_ops/v2-client"
```
rawlinp, line 38: Why is this necessary? Shouldn't we get everything to work w/ v3 since it's simply a copy of v2 at this point?

traffic_ops/testing/api/v3/about_test.go
```diff
@@ -0,0 +1,33 @@
+package v2
```
rawlinp, line 1: shouldn't this be v3?

grove/grovetccfg/grovetccfg.go
rawlinp: Although, now that I'm recalling the discussions we've had recently about components speaking two major versions of the TO API, maybe TM and TS _should_ stay on API v2 for the 5.0 release. Although, I'm pretty sure at least TM is going to need v3, right, because of dual-homing changes? So maybe it needs to use v2 for most things then _attempt_ v3 for things it needs for new features but still be able to fall back to v2?

ocket8888: Yeah, honestly I think upgrading all the clients to v2 right away was probably a mistake. Might let us get rid of Perl sooner, though, so well worth it

> _"use v2 for most things then attempt v3"_

That's an option, but tbh it depends on what you mean by "needs". TM changes will be capable of reading either 2.0 or 3.0 response payloads, so TM will continue to function with 2.0 for the forseeable future. But then it won't be able to provide the new feature.

traffic_ops/testing/api/v3/about_test.go
ocket8888: ... no

grove/grovetccfg/grovetccfg.go
rawlinp: Alright, so it sounds like we keep TM at v2 for the time being, but we'll need to figure something out for it soon in order to take advantage of new 3.0 features.

traffic_ops/ort/atstccfg/toreq/toreq.go
```diff
@@ -37,7 +37,7 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/lib/go-util"
-	toclient "github.com/apache/trafficcontrol/traffic_ops/client"
+	toclient "github.com/apache/trafficcontrol/traffic_ops/v2-client"
```
rawlinp, line 40: I don't think this is going to have the desired effect on `atstccfg` due to the way it currently vendors the TO-Go client. I'm pretty sure this will make it use the non-vendored v2 client. I think we need to follow the process of re-vendoring the latest version of the TO-Go client in atstccfg. Maybe that should be done in a separate follow-up PR, and just leave ORT as-is for now?

rob05c: Yeah, this is vendored. It's not necessary to change this, unless API 1 is being removed from Traffic Ops. If so, the `vendor/` dir needs updating along with this import.

The latest client is in https://github.com/apache/trafficcontrol/blob/master/traffic_ops/ort/atstccfg/toreqnew/toreqnew.go

But it's not necessary to update that either unless there's a new feature ORT needs.

ocket8888: ah, I see

mitchell852:
@ocket8888 - why is the `database` label on this PR?

ocket8888:
I think because I did that when this was a draft and I was anticipating having to manipulate seeds.sql for capabilities, but that's not actually how that works

ocket8888:
@rawlinp Those first two were because of rebase issues - should be fixed now - but the bottom one about incorrect arguments is a problem on master currently.

traffic_ops/testing/api/v3/cachegroupsdeliveryservices_test.go
```diff
@@ -0,0 +1,137 @@
...
+func CreateTestCachegroupsDeliveryServices(t *testing.T) {
+	dss, _, err := TOSession.GetDeliveryServiceServers()
+	if err != nil {
+		t.Fatalf("cannot GET DeliveryServiceServers: %v", err)
+	}
+	if len(dss.Response) > 0 {
+		t.Fatalf("cannot test cachegroups delivery services: expected no initial delivery service servers, actual %v", len(dss.Response))
+	}
+
+	dses, _, err := TOSession.GetDeliveryServices()
```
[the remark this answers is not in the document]
ocket8888: I checked and it was there. Also ran the tests - but at any rate I've rebased again since then and #4670 is merged so tests should both compile and pass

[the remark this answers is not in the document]
ocket8888: and also RE: this file in particular - should be recopied now

ocket8888:
Yeah, should be. I just re-copied that one file, but let me do a full folder re-copy to make sure it's all what it's supposed to be.

PR #4647, opened 2020-04-20
Fix /users/register DB query

### LINKED ISSUE #4583: POST request to users/register returns Internal Server Error
## I'm submitting a ...

- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [X] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When making a valid `POST` request to `/users/register` the response returned is:
```http
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```
## Expected / new behavior:

Should return:
```http
HTTP/1.1 200 OK
{
    "alerts": [
        {
            "text": "Sent user registration to {{<email>}} with the following permissions [ role: admin | tenant: root ]",
            "level": "success"
        }
    ]
}
```
## Anything else:

Error Log:
```ERROR: api.go:180: ... Checking for existing user with email {{ <email>}}: missing destination name rolename in *tc.User```

rawlinp:

The query was failing with the following:

    missing destination name rolename in *tc.User

This indicates that the struct was missing a "db:rolename" tag for that
field, but simply adding that tag would break the /users API due to it
using a special struct for GetUsers.

Instead, convert from sqlx.StructScan to plain sql.Scan, which avoids
any unintended side-effects caused by changing the User struct tags.

Fixes #4583.

- Traffic Ops

The the TO API v1 and v2 tests, verify they still pass. Manually register a new user, verify that TO gets to at least the point where the email _would_ be sent if SMTP was enabled. If not enabled, you should see this expected error in the TO API logs instead of the DB query failure:

    SMTP is not enabled; mail cannot be sent

That means it got past the DB error in #4583, but ideally, this would be manually verified in an environment that has SMTP enabled.

- master
- 4.0.0

- [x] Existing TO API tests cover most of this functionality, but they don't support email

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Fix to traffic_ops_ort.pl to strip specific comment lines before checking if a file has changed.  Also promoted a changed file message from DEBUG to ERROR for report mode.
 - Fixed Traffic Portal regenerating CDN DNSSEC keys with the wrong effective date
+- Fixed issue #4583: POST /users/register internal server error caused by failing DB query
 - Type mutation through the api is now restricted to only those types that apply to the "server" table
 - Updated The Traffic Ops Python, Go and Java clients to use API version 2.0 (when possible)
 - Updated CDN-in-a-Box scripts and enroller to use TO API version 2.0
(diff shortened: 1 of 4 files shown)


zrhoffman:
Unit tests pass, API tests pass, code is correctly formatted, `POST /users/register` succeeds with SMTP enabled. The response has not changed, so the docs are up-to-date.

:+1:

PR #4649, opened 2020-04-21
CDN-in-a-Box SMTP server + user registration API test

### LINKED ISSUE #3984: Ciab Register user returns 500 Internal server error
## I'm submitting a ...

- [ ] bug report
- [X] improvement request (usability, performance, tech debt, etc.)
- [ ] other 

- [X] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When we try to register a user from CIAB, 500 internal server error is returned
## Expected / new behavior:

User registration should be successful. 

User Admin-> Users ->Register User-> enter required details-> click send registration

### LINKED ISSUE #3997: Add Email server to Cdn-in-a-Box
## I'm submitting a ...
new feature / enhancement request

- CDN in a Box
- Documentation

## Current behavior:
There are several API endpoints that rely on an external SMTP server for operation. As CDN-in-a-Box has no SMTP server, using these endpoints requires a user to set up their own and edit configuration to accommodate that. 

## Expected / new behavior:
In order for CDN-in-a-Box to more closely portray a fully functioning ATC CDN, it should provide all of the functionality necessary for all TO API endpoints to work. This means adding an SMTP server - possibly as an optional service, though it is this writer's opinion that it ought to be built/started by default, as it is necessary for a subset of TO's base operation.

1. Start CDN-in-a-Box
1. try to hit an endpoint that sends an email e.g. [`/user/reset_password`](https://traffic-control-cdn.readthedocs.io/en/latest/api/user_reset_password.html)
1. Observe the inevitable internal server error (spews HTML in Perl, Go sends an error alert)

zrhoffman:

This PR 
- - [x] fixes #3997, fixes #3984,
- adds a SMTP server to CDN-in-a-Box, and
- adds an API test for user registration.

- CDN in a Box
- Documentation
- Traffic Ops API tests

* Run the API tests, confirm the new test passes (it makes use of the SMTP server)
* Revert 666550760d (#4647), rebuild the Traffic Ops CiaB image, run the API tests, and verify that the new test fails

- [x] This PR includes tests


rawlinp:
I think this is good, but I don't think we want to make a full-featured CIAB a requirement for running the TO API tests. Can we optionally disable certain tests that require external services like an SMTP server?

If we can do that, then we should also consider adding TO API tests for endpoints that use Riak, which should also be optional.

zrhoffman:
Okay, added the `includeSystemTests` option in 95763940d2. It can be set from CLI or in the config.

traffic_ops/testing/api/v2/user_test.go
```diff
@@ -37,6 +39,9 @@ func TestUsers(t *testing.T) {
 		GetTestUsers(t)
 		GetTestUserCurrent(t)
 		UserTenancyTest(t)
+		if includeSystemTests {
+			UserRegistrationTest(t)
```
rawlinp, line 43: since this force-deletes users, we should include a comment that this test needs to run last in the sequence, otherwise tests that come after it may fail, right?

zrhoffman: Yes. Included a comment about this in 53b2e5816.

PR #4653, opened 2020-04-21
Fixed logging.yaml generation.

alficles:

The logging.yaml file was generating a separate logs section for each
log file. This was allowing the yaml parser in ATS to accept the file,
but fail to process any but one of the log files. This corrects that by
putting all the logs in the same object.

- [x] This PR is not related to any existing open issue 

- Traffic Ops ORT

This does not change any existing functionality, it just updates the generation to be correct and to match the existing documentation.

Run ORT against a Traffic Ops instance to generate a config files for a cache that has multiple logging formats configured in a logging.yaml file.

Additionally, this functionality is verified in a unit test as part of lib/go-atscfg, so those unit tests should run and pass.

This fixes a bug that was copied from Traffic Ops in Perl. So no version of Traffic Ops has ever properly rendered this file.

- [x] An update to CHANGELOG.md


ocket8888:
Does this change the [current declared conversion of `logging.yaml` Parameters to chunks of the `logging.yaml` file](https://traffic-control-cdn.readthedocs.io/en/latest/overview/profiles_and_parameters.html#logging-yaml)?

alficles:
@ocket8888 Nope. It makes it match, actually. The documentation indicates, correctly, that `logs:` should be listed only once, but it was actually being generated for every chunk.

ocket8888:
Now that's what I like to hear

jrushford:
I built this, ran unit tests and ort against on a lab machine agains staging and the logging.yaml was built correctly.

Issue #3746, opened 2019-07-23
Fix DS "ACTIVE" flag to work as expected


rgresino:
![image](https://user-images.githubusercontent.com/7713938/61719063-75007100-ad21-11e9-931e-8a3223f592f0.png)
It appears that this is not what this does.  This does not remove traffic from a DS or inactivate it. It removes the DS from the Traffic Router so it is not routed to. Any established connections like linear streams will continue to use this delivery service since they rarely, if ever, go back to the traffic router.We need something that  sets a flag so that the DS is held in TODB, but is not copied into records.config on the caches, so that the DS is removed from the caches, but could be re-enabled quickly, but not deleted from TODB (and the associated Keys and Certs deleted from Traffic Vault.)

rob05c:
Big +1. This is a terrible foot-gun. It's incredibly unintuitive, and further, if the wrong DS were left inadvertently serving at the edge, it could be a security issue.

I suggest we add an additional DS flag, "active-caches," and rename this in the UI to reflect that it is "active-routed." Of course, with documentation and UI tooltips indicating exactly what each does.

The "active-caches" flag to remove a DS from cache configs should be a tiny amount of work, it's basically a single if-statement in remap.config generation, unless I'm mistaken.

I'm also +1 on labelling this 'bug' and 'major'. It may have been by-design, but it's a dangerous and deceptive design, and effectively a user interface bug.

mitchell852:
@rob05c - active-routed AND active-caches? would you ever want active-routed=false, active-caches=true or vice versa? I thought the fix here was simply if DS.active=false exclude from crconfig (snapshot) and any/all applicable cache config files (maybe an oversimplification).

rgresino:
Have to snap before you can remove from ATS.  So it is a two step process.

rob05c:
>would you ever want active-routed=false, active-caches=true or vice versa?

Yes. You might want to stage the removal, to remove a DS from routing, and then make sure the Origin own was right and there actually aren't clients still going to caches, and then remove from the cache maybe a week later.

Likewise, you may want to stage a new deployment to caches, and test that your requests to the cache work as expected, before deploying globally to the Router.

Being able to "stage" the router-cache deployment changes is a big safety win for operations.

ocket8888:
As of today, inactive DSes are already omitted from remap.config generation, I think parent.config is the only place they will have data inserted. I *think*.

mitchell852:
[...]

PR #4654, opened 2020-04-21
Ds Active flag blueprint

ocket8888:

- [x] This PR is related to #3746

Adds a blueprint for changing the "active" property of Delivery Services from a boolean to an enumerated string constant like:

```typescript
/**
 * This defines what other components of ATC will consider a Delivery Service
 * "active".
 *
 * It's not an object exposed through the API in its own right, just a
 * specification of the allowed values.
*/
enum DeliveryServiceActiveState {
	/**
	 * A Delivery Service that is ”active” is one that is functionally
	 * in service, and fully capable of delivering content.
	 *
	 * This means that its configuration is deployed to Cache Servers and it is
	 * available for routing traffic.
	*/
	ACTIVE = 'ACTIVE',
	/**
	 * A Delivery Service that is ”inactive” is not available for
	 * routing and has not had its configuration distributed to its assigned
	 * Cache Servers.
	*/
	INACTIVE = 'INACTIVE',
	/**
	 * A Delivery Service that is ”primed” has had its configuration
	 * distributed to the various servers required to serve its content.
	 * However, the content itself is still inaccessible via routing.
	*/
	PRIMED = 'PRIMED'
}
```

For more details, refer to the blueprint.

None - blueprint.

Read the blueprint.

- [x] Tests are unnecessary

### DIFF
--- a/blueprints/ds-active.md
+++ b/blueprints/ds-active.md
@@ -283,3 +283,14 @@ before then clients using Cache Servers for ongoing content delivery will be
 unaffected by the change. This allows operators to "drain" a Delivery Service
 for a while before setting it to "INACTIVE" and removing it from Cache Server
 configuration, thus minimizing client impact.
+
+Another that has been brought up is changing the name of the current `active`
+field to be `active-routed`, which would mean that Delivery Service will be
+routed, and adding another field called `active-cached` which would mean that
+the Delivery Service's configuration would be disseminated to the cache
+servers. Thus the same amount of flexibility could be achieved without making a
+breaking API change.
+
+The trade-off is that it would allow Delivery Services to be in a state where
+Traffic Router will direct clients to cache servers which are incapable of
+properly serving them content.
--- a/blueprints/ds-active.md
+++ b/blueprints/ds-active.md
@@ -19,42 +19,90 @@ under the License.
 # Delivery Services 'Active' field
 
 ## Problem Description
-<!--
-*What* is being asked for?
-*Why* is this necessary?
-*How* will this be used?
--->
+Setting a Delivery Service to "Inactive" actually only sets it to "not routed".
+Currently, there is no way to create a Delivery Service (with assigned servers)
+that will not be distributed to cache server configuration.
 
 ## Proposed Change
-<!--
-*How* will this be implemented (at a high level)?
--->
+This blueprint proposes changing the `Active` property of Delivery Services
+from a boolean to an enumerated string constant that can represent three
+different "Activity States" for a Delivery Service.
+
+## Data Model Impact
+The proposed new type of the field is expressed below as a TypeScript
+enumeration.
+```typescript
+/**
+ * This defines what other components of ATC will consider a Delivery Service
+ * "active".
+ *
+ * It's not an object exposed through the API in its own right, just a
+ * specification of the allowed values.
+*/
+enum DeliveryServiceActiveState {
+	/**
+	 * A Delivery Service that is ”active” is one that is functionally
+	 * in service, and fully capable of delivering content.
+	 *
+	 * This means that its configuration is deployed to Cache Servers and it is
+	 * available for routing traffic.
+	*/
+	ACTIVE = 'ACTIVE',
+	/**
+	 * A Delivery Service that is ”inactive” is not available for
+	 * routing and has not had its configuration distributed to its assigned
+	 * Cache Servers.
+	*/
+	INACTIVE = 'INACTIVE',
+	/**
+	 * A Delivery Service that is ”primed” has had its configuration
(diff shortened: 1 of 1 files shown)


rawlinp:
This is related to issue #3746

blueprints/ds-active.md
```diff
@@ -0,0 +1,285 @@
...
+updates" that was done would remove it from Cache Server configuration even if
+a Snapshot has not been taken. This results in clients being routed to Cache
+Servers that cannot serve the content they requested.
+
+By contrast, when setting a Delivery Service to "PRIMED", clients will no
+longer be routed for that Delivery Service but even if updates are queued
+before then clients using Cache Servers for ongoing content delivery will be
+unaffected by the change. This allows operators to "drain" a Delivery Service
+for a while before setting it to "INACTIVE" and removing it from Cache Server
+configuration, thus minimizing client impact.
```
rawlinp, line 285: Another alternative that was proposed on the issue I linked was making the existing `active` field really mean `active-routed` and adding a new field called `active-cached`. Personally if we're already going to be making breaking API changes for servers v3 I think we might as well make this breaking change too.

ocket8888: You think I should add that? Or is it sufficient that it was said here?

ocket8888:
I don't have a problem with changing the name of the field. But having discussed this with operators I do think three states are desired.

blueprints/ds-active.md
rawlinp: Seems like it would be worth the add -- considering it would mean the difference between making a breaking API change for this or not.

rob05c:
>I guess the real question for the community is "is there a real need for 3 states?"

I vote yes. 

We need to be able to send config to caches, before it's sent to the router (the existing "Inactive" option). Without it, there's no way to safely stop routing to a DS. So the existing behavior is absolutely necessary.

But currently, there's no way to not send a DS to caches. This means potentially having clients, malicious or accidental, making requests that shouldn't be possible to make, by requesting directly to the cache. It's very common to not yet want to delete the information behind a DS, even when it's no longer active, or not yet active, or temporarily not active. The third "don't send to caches" state makes that possible.

More than that, it's a Footgun. Because "inactive" sounds like it isn't being put anywhere. Unless the user has the TC docs memorized, it's a very easy mistake for new TC operators or tenants to make. Three unambiguous states eliminates that footgun.

rob05c:
I'll also add, as above, a DS shouldn't be immediately removed from (or changed in) routers and caches, without some drain time period, as well as the reverse, making sure a DS is on all Caches before sending to Routers.

In theory, this can be done by waiting long enough between Queue and Snap. But it's safer not to add risk to the already pernicious queue-snap order-of-operations.

So, I think we should make it impossible to change a DS directly from inactive<->active. Both the Portal and the API should require being moved from Active to Primed, and then Primed to Inactive.

Someone who needs to can always do it twice. But that makes it safer, and reduces the chance of operator misunderstanding.

Ideally with good error messages, e.g. "Error: Inactive delivery services must be primed before they're set to active, to make sure they're on the caches before being routed to."

ocket8888:
> _"I think we should make it impossible to change a DS directly from inactive<->active"_

So would that also mean you can never create an ACTIVE DS?

rob05c:
>So would that also mean you can never create an ACTIVE DS?

Yes, +1 on requiring created DSes to be Inactive, or possibly Primed. But, I'd be in favor of removing the Active field from the DS creation form altogether.

In theory, it's probably mostly ok to let someone make a DS as Active, because there are no successful clients anyway.

But it's possible someone making a DS has clients already requesting, and it's possible they handle NXDOMAIN better than 404. For example, maybe they're configured to retry NXDOMAINs as errors, but 404's generally mean something's wrong with the URL so they give the client an error. Worse, maybe they cache that 404 locally.

It's a bit of an edge case, I don't think it's a huge deal either way. But IMO making new DSes always start out Inactive adds an extra little bit of wrong-things-are-impossible safety.

rawlinp:
We should probably make it clear why this is necessary when it seems like we can already accomplish these 3 states today:
```
proposed:      what it means:                 how it's done today:
=================================================================================
active    ==   routed, cached             ==  active,   has ds-server assignments
primed    ==   cached, but not routed     ==  inactive, has ds-server assignments
inactive  ==   not routed, not cached     ==  inactive, does not have ds-server assignments
```
Granted, we're hoping to remove ds-server assignments in the future, so that seems to make this change necessary.

ocket8888:
That depends on what you mean by "accomplish". You can render Delivery Services un-service-able, but that's not _exactly_ the same thing as having their configuration ready to deploy at the push of a button - *one* button.

Issue #3509, opened 2019-04-19
Add Traffic Ops If-Match and If-Unmodified-Since Support in Server and Clients


rob05c:
Currently, there is a race condition if multiple people PUT an object, they will overwrite each-other without realizing it.

HTTP has a well-established solution for this: `If-Match` and `If-Unmodified-Since`.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Unmodified-Since
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match

We should implement this, to prevent races and people accidentally overwriting each other.

It shouldn't be too difficult to implement. We can generate an ETag+Last-Modified header from the resource's `last_updated` field, and likewise compare them when we get a request with IM or IUS.

It's tempting to add the `bug` tag, since this is a logical bug, even if it's not a code bug. I won't, but I wouldn't object if someone else did.

PR #4657, opened 2020-04-22
Allow support for "If-Modified-Since" on GET requests

srijeet0406:

- [x] This PR is related to #3509

- Traffic Ops

You can either test it on your local setup or CiaB.
Here's what I did:
Once you have TO running, make a call to the "GET all" endpoint for any of the resources to get all the items. 
Now, add a new item to the DB either via the TP UI or via a POST call. Now, do the "GET all" call with a header of "If-Modified-Since" specifying a date which is just BEFORE the time of creation of this last resource. Since this is a "GET all" call, it would still return "ALL" of the resources, since its a bulk read. Now, make a GET call for that last resource by ID, with a header of "If-Modified-Since" specifying a date which is just AFTER the time of creation of this last resource. This time, it should just return an empty response with a status code of 304 (NOT MODIFIED).

- [x] This PR includes tests


ocket8888:
Does this fix #3509 ?

traffic_ops/traffic_ops_golang/apitenant/tenant.go
```diff
@@ -122,7 +123,18 @@ func (ten TOTenant) Validate() error {
 
 func (tn *TOTenant) Create() (error, error, int) { return api.GenericCreate(tn) }
 
-func (ten *TOTenant) Read() ([]interface{}, error, error, int) {
+func (ten *TOTenant) Read(h map[string][]string) ([]interface{}, error, error, int) {
+	ims := h["If-Modified-Since"]
```
rob05c, line 127: Would you object to making this a constant in `lib/go-rfc`?

traffic_ops/traffic_ops_golang/apitenant/tenant.go
```diff
@@ -122,7 +123,18 @@ func (ten TOTenant) Validate() error {
...
+	var modifiedSince time.Time
+	modified := false
+	found := false
+	code := http.StatusOK
+
+	if ims != nil && len(ims) != 0 {
+		if t, err := time.Parse(time.RFC1123, ims[0]); err == nil {
```
rob05c, line 134: This isn't sufficient, HTTP Dates can be several formats. There's an existing function you can  use: https://github.com/apache/trafficcontrol/blob/7f5346618daa4d98b7fd1cf9f674650fb7fa6f7f/grove/web/util.go#L185

Which should probably be moved to `lib/go-rfc`.

See https://tools.ietf.org/html/rfc7231#section-7.1.1.1

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
```diff
@@ -100,6 +102,30 @@ func (v *TOServerCapability) Validate() error {
...
-func (v *TOServerCapability) Read() ([]interface{}, error, error, int) { return api.GenericRead(v) }
-func (v *TOServerCapability) Create() (error, error, int)              { return api.GenericCreateNameBasedID(v) }
-func (v *TOServerCapability) Delete() (error, error, int)              { return api.GenericDelete(v) }
+func (v *TOServerCapability) Read(h map[string][]string) ([]interface{}, error, error, int) {
+	ims := h["If-Modified-Since"]
+	var modifiedSince time.Time
+	var res []interface{}
+
+	if ims == nil || len(ims) == 0 {
+		return api.GenericRead(v)
```
rob05c, line 111: There seems to be a lot of very similar code around reads. How difficult would it be to reduce the duplication by putting the "modified" logic inside `api.GenericRead` somehow?

traffic_ops/traffic_ops_golang/cdnfederation/cdnfederations.go
```diff
@@ -143,7 +144,18 @@ func checkTenancy(tenantID *int, tenantIDs []int) bool {
...
-func (fed *TOCDNFederation) Read() ([]interface{}, error, error, int) {
+func (fed *TOCDNFederation) Read(h map[string][]string) ([]interface{}, error, error, int) {
+	ims := h["If-Modified-Since"]
+	var modifiedSince time.Time
+	modified := false
+	found := false
+	code := http.StatusOK
+
+	if ims != nil && len(ims) != 0 {
+		if t, err := time.Parse(time.RFC1123, ims[0]); err == nil {
```
rob05c, line 155: It seems that nearly every endpoint needs this. Should we reduce the duplicated logic by having the HTTP Routing parse this and put it in the HTTP Context?

traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
```diff
@@ -426,10 +439,20 @@ func (cg *TOCacheGroup) Read() ([]interface{}, error, error, int) {
 		}
 		s.LocalizationMethods = &lms
 		s.Fallbacks = &cgfs
+		// In case of a bulk read, even if one of the items has a "lastUpdated" time that is after whats supplied in the request,
+		// we send back the entire array of results
+		if !s.LastUpdated.Before(modifiedSince) {
```
rob05c, line 444: This isn't sufficient. If objects have been deleted, the endpoint won't be flagged as Modified, and will return a 304 when there have actually been changes.

This needs to keep track of deleted rows somehow.

There are several ways to do it. One would be to create a `deleted_foo` table for each existing table, tracking deletions. Possibly with a cleanup routine, and logic to determine "if IMS request  is older than the oldest deletion/cleanup time, consider it modified.

Another way is to make deletions update a "last modified" timestamp  on the endpoint somehow. This could be a simple table of timestamps for each table, or it could artificially update the timestamp of some row on another table which is part of the endpoint (for example, deliveryservices_regexes could update the timestamp of the deliveryservice they belong to, when one  is deleted). Possibly as a Database Trigger.

Those are just ideas. As long as deletions are tracked, I don't have a strong opinion on how.

rob05c:
>Does this fix #3509

It doesn't look like it. It looks like it just adds IMS, not If-Unmodified-Since or If-Match.

But it should make it much easier to, since it adds the modification logic necessary, should just be a matter of adding the header logic for IUS+IM.

Same for ETags and If-None-Match, which allow more precise matching. I have a PR with the code for generating and matching ETags, it shouldn't be too difficult to integrate it with this:
https://github.com/apache/trafficcontrol/blob/5379f97aed6d137cc79b8c517a5601fe3c02beeb/lib/go-rfc/cachecontrol.go

traffic_ops/traffic_ops_golang/apitenant/tenant.go
ocket8888: > _"Which should probably be moved to `lib/go-rfc`"_

:+1:

ocket8888:
Yeah, he PM'd me. Just wanted to be sure that if this did it would auto-close the issue on merge. But it doesn't so all's good.

traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
srijeet0406: Great catch, I'll work on fixing that

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
srijeet0406: Yeah, I was just talking to Brennan about it, and thats the reason I marked this PR as a draft, because I'm trying to see if we can do it all in the GenericReader code. That way, we just make the call to the DB and dont have to copy the same code changes everywhere.

traffic_ops/traffic_ops_golang/apitenant/tenant.go
```diff
@@ -122,7 +123,18 @@ func (ten TOTenant) Validate() error {
...
 
-func (ten *TOTenant) Read() ([]interface{}, error, error, int) {
+func (ten *TOTenant) Read(h map[string][]string) ([]interface{}, error, error, int) {
+	ims := h["If-Modified-Since"]
+	var modifiedSince time.Time
+	modified := false
+	found := false
+	code := http.StatusOK
+
+	if ims != nil && len(ims) != 0 {
```
ocket8888, line 133: `len(nil) == 0` so you don't need to check both

traffic_ops/traffic_ops_golang/apitenant/tenant.go
```diff
@@ -143,10 +155,22 @@ func (ten *TOTenant) Read() ([]interface{}, error, error, int) {
...
 		t.ParentName = &p // copy
 	}
-	return tenants, nil, nil, http.StatusOK
+	// If the modified flag stayed false throughout (meaning that all the items' "lastUpdated" time is before whats supplied in the request),
+	// we send back a 304, with an empty response
+	if modified == false && found == true {
+		code = http.StatusNotModified
+		tenants = []interface{}{}
+	}
+	return tenants, nil, nil, code
```
ocket8888, line 173: this logic looks like if no IMS header is sent in the request it will send back a Not Modified response.

PR #4573, opened 2020-03-30
Update Go to version 1.14

zrhoffman:
## What does this PR (Pull Request) do?

This PR:
* Updates Go to 1.14 for all component
* Enables TLS version 1.1 in Traffic Vault

- [x] This PR is not related to any Issue 

## Which Traffic Control components are affected by this PR?

- CDN in a Box
- Documentation
- Grove
- Traffic Control Client (Go)
- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Traffic Stats
- Traffic Vault

## What is the best way to verify this PR?

- Run unit tests
- Run Traffic Ops API tests
- Make sure Traffic Vault still works
- Check for Go 1.14 deprecations in dependencies:
    ```shell script
    staticcheck -checks SA1019 ./... | grep -vF ': tc.'
    ```
- Make sure docs build with no warnings
  
The MaxTLSVersion config setting for `riak.conf` has a test. Besides that, this PR relies on the tests that are already in place to check for breaking changes when updating the Go version.

## The following criteria are ALL met by this PR

- [x] I have explained why tests are unnecessary
- [x] This PR includes documentation
- [x] This PR includes an update to CHANGELOG.md
- [x] This PR includes any and all required license headers
- [x] This PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #4660, opened 2020-04-23
Traffic Ops riak.conf: Accept a list of TLS versions to enable

zrhoffman:

This PR updates the Traffic Ops config for communicating with Riak to accept a `"TLSVersions"` array instead of the single `"MaxTLSVersion"` value that #4573 added. It leaves TLS 1.1 as the default enabled version for communicating with Riak.

### Motivations for the config option name change
After #4573 was merged, some concerns were brought up about the fact that, even though the config option was named `MaxTLSVersion`, versions below `MaxTLSVersion` but greater than or equal to the default TLS minimum version were not tried. So the question became whether `MaxTLSVersion` was an accurate name for that config option. `TLSVersions` is more straightforward.

### Reasons for checking for a list of versions
- Without any additional config changes after #4573, Traffic Vaults that have TLS 1.0 and TLS 1.2 enabled but not TLS 1.1 are effectively unreachable by Traffic Ops. While we should still leave it up to our users as far as what TLS versions they want to enable, Traffic Ops should not be limited to either succeeding with 1 specific TLS version or failing to reach Traffic Vault.
- Even though Go 1.14.2 uses TLS 1.3 by default, its default `tls.Config.MinVersion` is TLS 1.0, so the fact that [basho/riak-go-client](https://github.com/basho/riak-go-client) doesn't attempt handshakes with TLS versions lower than the maximum version is a bug.

- CDN in a Box
- Documentation
- Traffic Ops

* Run Traffic Ops unit tests
* Run Traffic Ops API tests to check for regressions
* Start CDN-in-a-Box, verify that Traffic Ops stores TLS keys in Traffic Vault
* Build docs, verify there are no warnings but also make sure they look nice

- master (764aaf0abb)

- [x] This PR includes tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added ability to create multiple objects from generic API Create with a single POST.
 - Added debugging functionality to CDN-in-a-Box.
 - Added an SMTP server to CDN-in-a-Box.
+- Added the `"TLSVersions"` config option to Traffic Ops's `riak.conf` to optionally specify exactly which TLS versions should be used for communicating with Riak
 - Traffic Ops Golang Endpoints
   - /api/2.0 for all of the most recent route versions
   - /api/1.1/cachegroupparameters/{{cachegroupID}}/{{parameterID}} `(DELETE)`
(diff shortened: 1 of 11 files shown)


zrhoffman:
Let's not enable TLS 1.0 by default. Removed TLS 1.0 from the default enabled TLS versions for Traffic Ops to try in 4342d18d2.

zrhoffman:
Let's not

PR #4661, opened 2020-04-23
Post ds handle empty xmlid

### LINKED ISSUE #4591: POST request to deliveryservices returns Internal Server Error
## I'm submitting a ...

- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When making a `POST` request to `/deliveryservices` without the `xmlId` in the request body or the `xmlId` set to `null` the response returned is:
```http
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```
However when setting the `xmlId` field to an empty string `i.e. xmlId = ""` the response returned is:
```http
HTTP/1.1 200 OK
{
    "alerts": [
        {
            "text": "Deliveryservice creation was successful.",
            "level": "success"
        }
    ],
    "response": [
        {...}
    ]
}
```
If you try to create another delivery service with another xmlId set to be an empty string you will get an Internal Server Error.
## Expected / new behavior:

Should return:
```http
HTTP/1.1 400 Bad Request
{
    "alerts": [
        {
            "text": "invalid request: 'xmlId' cannot be blank",
            "level": "error"
        }
    ]
}
```
## Anything else:

Error Log:
```ERROR: api.go:180: ... checking tenant: getting tenant ID: delivery service has no ID or XMLID```

### PR DESCRIPTION
**Pull request by shamrickus**

- [x] This PR fixes #4591. The usages of the deprecated `tc.DeliveryService` are all replaced with `tc.DeliveryServiceNullable` in the v2 tests (no test removals).

- Traffic Ops

Verify passing xmlID as null and empty string to post `/deliveryservices/`, it should return a 400 error instead of a 500.

### DIFF
--- a/lib/go-tc/deliveryservices.go
+++ b/lib/go-tc/deliveryservices.go
@@ -448,9 +448,6 @@ func (ds *DeliveryServiceNullable) Validate(tx *sql.Tx) error {
 		"typeId":              validation.Validate(ds.TypeID, validation.Required, validation.Min(1)),
 		"xmlId":               validation.Validate(ds.XMLID, validation.Required, noSpaces, noPeriods, validation.Length(1, 48)),
 	})
-	//if ds.XMLID != nil && *ds.XMLID == "" {
-	//	errs = append(errs, errors.New("xmlId cannot be blank"))
-	//}
 	if err := ds.validateTypeFields(tx); err != nil {
 		errs = append(errs, errors.New("type fields: "+err.Error()))
 	}
--- a/lib/go-tc/deliveryservices.go
+++ b/lib/go-tc/deliveryservices.go
@@ -446,8 +446,11 @@ func (ds *DeliveryServiceNullable) Validate(tx *sql.Tx) error {
 		"remapText":           validation.Validate(ds.RemapText, noLineBreaks),
 		"routingName":         validation.Validate(ds.RoutingName, isDNSName, noPeriods, validation.Length(1, 48)),
 		"typeId":              validation.Validate(ds.TypeID, validation.Required, validation.Min(1)),
-		"xmlId":               validation.Validate(ds.XMLID, noSpaces, noPeriods, validation.Length(1, 48)),
+		"xmlId":               validation.Validate(ds.XMLID, validation.NotNil, noSpaces, noPeriods, validation.Length(1, 48)),
 	})
+	if ds.XMLID != nil && *ds.XMLID == "" {
+		errs = append(errs, errors.New("xmlId cannot be blank"))
+	}
 	if err := ds.validateTypeFields(tx); err != nil {
 		errs = append(errs, errors.New("type fields: "+err.Error()))
 	}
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: traffic_ops/testing/api/v2/tc-fixtures.json [hunk 1]
```diff
@@ -2572,32 +2572,6 @@
...
-        {
-            "dsName": "ds1",
-            "request": {
-                "startTime": "2019-01-01 00:00:00",
-                "ttl": 80,
-                "regex": "/foo"
-            }
-        },
-        {
-            "dsName": "ds2",
```
rawlinp, line 2585: removing this test data is going to break the tests that depend on them -- they might not fail, since they don't always check to make sure their test data exists, but the tests won't actually test what they were meant to test

lib/go-tc/deliveryservices.go
```diff
@@ -446,8 +446,11 @@ func (ds *DeliveryServiceNullable) Validate(tx *sql.Tx) error {
 		"remapText":           validation.Validate(ds.RemapText, noLineBreaks),
 		"routingName":         validation.Validate(ds.RoutingName, isDNSName, noPeriods, validation.Length(1, 48)),
 		"typeId":              validation.Validate(ds.TypeID, validation.Required, validation.Min(1)),
-		"xmlId":               validation.Validate(ds.XMLID, noSpaces, noPeriods, validation.Length(1, 48)),
+		"xmlId":               validation.Validate(ds.XMLID, validation.NotNil, noSpaces, noPeriods, validation.Length(1, 48)),
```
rawlinp, line 449: I think this might need to be `validation.Required`, which covers both the `nil` and `""` cases. So in theory the `if` below is unnecessary.

traffic_ops/testing/api/v2/tc-fixtures.json
shamrickus: You may be correct, regardless I meant to put that it a separate PR.

traffic_ops/testing/api/v2/traffic_control_test.go
```diff
@@ -54,11 +54,5 @@ type TrafficControl struct {
 	SteeringTargets                      []tc.SteeringTargetNullable             `json:"steeringTargets"`
 	Serverchecks                         []tc.ServercheckRequestNullable         `json:"serverchecks"`
 	Users                                []tc.User                               `json:"users"`
-	Jobs                                 []JobRequest                            `json:"jobs"`
```
rawlinp, line 57: I don't think you meant to include this change, right?

shamrickus: Woops, you are correct.

PR #4665, opened 2020-04-27
Support IMS header in Get endpoints

srijeet0406:

- CDN in a Box
- Documentation
- Traffic Ops

You can either test it on your local setup or CiaB.
Here's what I did:
Once you have TO running, make a call to the "GET all" endpoint for any of the resources to get all the items.
Now, add a new item to the DB either via the TP UI or via a POST call. Now, do the "GET all" call with a header of "If-Modified-Since" specifying a date which is just BEFORE the time of creation of this last resource. Since this is a "GET all" call, it would still return "ALL" of the resources, since its a bulk read. Now, make a GET call for that last resource by ID, with a header of "If-Modified-Since" specifying a date which is just AFTER the time of creation of this last resource. This time, it should just return an empty response with a status code of 304 (NOT MODIFIED).
NOTE: To turn the IMS mode on or off, I created a config flag for TO. Its called `use_ims`. 
In order to test out the above functionality, the flag must be set to true.

master

- [x] This PR includes tests 


lib/go-rfc/http.go
```diff
@@ -36,6 +36,7 @@ const (
 	ContentDisposition     = "Content-Disposition"      // RFC6266
 	ApplicationOctetStream = "application/octet-stream" // RFC2046§4.5.2
 	Vary                   = "Vary"                     // RFC7231§7.1.4
+	IfModifiedSince        = "If-Modified-Since"        // RFC1945
```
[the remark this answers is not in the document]
srijeet0406: Done

traffic_ops/app/db/migrations/20200424000000_add_deleted_tables.sql
```diff
@@ -0,0 +1,16 @@
...
+-- SQL in section 'Up' is executed when this migration is applied
+
+-- deleted_type
+CREATE TABLE IF NOT EXISTS deleted_type (
+    id bigint NOT NULL,
+    name text,
+    description text,
+    use_in_table text,
+    last_updated timestamp with time zone DEFAULT now(),
+    deleted_time timestamp with time zone NOT NULL DEFAULT now()
```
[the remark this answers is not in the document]
srijeet0406: Yeah, makes sense to use the updated "last_updated" time from the deleted_table. Done.

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -191,8 +271,28 @@ func GenericOptionsDelete(val GenericOptionsDeleter) (error, error, int) {
...
+	if err != nil {
+		return ParseDBError(err)
+	}
+	return nil, nil, http.StatusOK
+}
+
 // GenericDelete does a Delete (DELETE) for the given GenericDeleter object and type. This exists as a generic function, for the common use case of a simple delete with query parameters defined in the sqlx struct tags.
 func GenericDelete(val GenericDeleter) (error, error, int) {
+	var v interface{}
+	res, err := val.APIInfo().Tx.NamedQuery(val.SelectBeforeDeleteQuery(), val)
```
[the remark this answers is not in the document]
srijeet0406: Done

traffic_ops/traffic_ops_golang/types/types.go
```diff
@@ -158,6 +174,34 @@ func (tp *TOType) loadUseInTable() (error, error, string) {
 	return nil, nil, useInTable
 }
 
+func selectMaxLastUpdatedQuery(where, orderBy, pagination, where2, orderBy2, pagination2 string) string {
```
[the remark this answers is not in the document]
srijeet0406: The reason I did that is because of the way the "ParamColumns" method works. Its creating an alias for the main table and then passing that along to the dbHelpers function to create the "where" clause. If we have a deleted table, we would need to query against that table (with its own alias as well). For ex, this:

`SELECT max(t) from (
		SELECT max(last_updated) as t from type typ WHERE typ.id=12345
		UNION ALL
	SELECT max(last_updated) as t from deleted_type dtyp WHERE dtyp.id=12345
		) as res`
This is why I was thinking of splitting out the two. The only way I could think of was by changing the "ParamColumns" method, but that seemed like a more involved change. Is there a better way of doing this?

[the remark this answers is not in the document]
rob05c: The `UNION ALL` queries can have the same alias, they're effectively separate queries. So
```
`SELECT max(t) from (
		SELECT max(last_updated) as t from type typ ` + where + orderBy + pagination +
		` UNION ALL
	select max(last_updated) as t from deleted_type typ ` + where + orderBy + pagination +
		` ) as res`
```
Works fine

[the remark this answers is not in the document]
srijeet0406: 👍

traffic_ops/app/db/migrations/20200424000000_add_deleted_tables.sql
```diff
@@ -0,0 +1,608 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+
+CREATE TABLE IF NOT EXISTS deleted_api_capability (
+    id bigserial PRIMARY KEY,
```
rawlinp, line 20: I've been trying to follow along, but I might've missed something. Why do we need to copy all the existing tables to `deleted_foo`? At a high level, don't we only care about the answer to the question "have any foos been updated or deleted since this point in time?"

srijeet0406: We need to keep track of the deleted components because, if we just compare the "last_updated" times of the items present in the main tables with what we are supplying in the IMS header, if something got deleted at a time after the IMS time in the request, we would not take that into account and return 304 (which is incorrect, since the endpoint has changed since the IMS time, i.e., one of the objects got deleted).

rob05c: Right, we could only track IDs, but then any change would flag as modified if anything had been deleted, even if it wasn't in the filter. 

E.g. if someone deleted a Parameter on Profile Foo, and someone requested `profiles/name/bar/parameters`, it would flag as modified, even though it wasn't really.

rawlinp: Yeah, I understand that part, so we need _something_ to tell us that rows were deleted. But, why do we need _all_ the table data? Shouldn't all the `deleted_foo` tables only have a single column -- e.g. `deleted_at` -- that we update via an `ON DELETE` trigger similar to the trigger for `last_updated`? It seems like the `deleted_foo` tables would only ever need a single row to track when something was last deleted from its corresponding table. Maybe we only need one `deleted` table, which has a row for every _main_ table, that contains the timestamp of when something was last deleted from that referenced table?

srijeet0406: IMO, it would depend on what all the acceptable params for `DELETE` are. For ex, in some cases, we could be deleting by name, in others, by ID.

rawlinp: > Right, we could only track IDs, but then any change would flag as modified if anything had been deleted, even if it wasn't in the filter.

I think there is a trade-off to be made here in terms of complexity vs performance gain. We can keep it really simple by ignoring the "filter" side of things, and have simple invalidations per table. Or, we can copy all of the existing data, which means we have to keep the columns of both tables in sync forever, not to mention dealing with the additional problems created by keeping deleted data forever.

rob05c: Suppose I make a request to `/api/2.0/deliveryservices?cdn=FOO` at 10:00am.
Someone else then does a `DELETE /api/2.0/deliveryservices/cdn-bar-ds` at 10:05am.
I then do an IMS request to `/api/2.0/deliveryservices?cdn=FOO` at 10:10am, with my `If-Modified-Since: 10:00am`.

The `deleted_deliveryservices` table only has `id | last_updated`. It says "there has been a DS deleted since 10:00am, therefore I must send you the full request back.

When in reality, no DS on CDN "FOO" has been deleted or modified, and it can safely return a 304 Not Modified. But there's no way for the code to determine that, without knowing the data queried on, the cdn id, of the deleted record.

We don't actually need all data, just all columns that can be queried on. But that's probably most of them, it seems easier and safer to add them all.

rawlinp: In your example, I think it would be fine if we returned the full response back because the reality is that things are not deleted very often, so I don't think the theoretical savings of those rare cases is worth the cost of dev/maintenance. IMO I think it would be better to keep the deletion case simple, not fully duplicate tables, and not fully duplicate queries. If anything was deleted from a table since time T, _any_ IMS requests with time < T that involve those tables could simply return the full response. Since deletion is so rare, the extra complexity doesn't seem worth the savings.

rob05c: >I think there is a trade-off to be made here in terms of complexity vs performance gain. 

I think it's a pretty big performance gain, and not too complex. Without it, it'll mean any modification to any profile will make IMS fail on ORT for every server. It's hard to know for sure without looking at detailed change data over a long period of time, but I suspect not handling params will put the IMS hit rate <10%, and with params will be >90%.

> we have to keep the columns of both tables in sync forever

We do. It's a little code, but it doesn't seem unreasonable, it just means adding a column has to add it in both tables.

> not to mention dealing with the additional problems created by keeping deleted data forever.

Personally, I don't believe this will be an issue. Our data is tiny. Speaking from experience, gigabytes of data really isn't big or hard to deal with, and our data is megabytes.

But if it is an issue, it's pretty easy to truncate it, and always return Modified for an IMS older than that.

rob05c: >I don't think the theoretical savings of those rare cases is worth the cost of dev/maintenance. 

I guess we'll have to agree to disagree on that. The savings are not theoretical, and I don't believe the cases are rare. I believe the performance savings are huge. For ORT, TP, and everything else that regularly hits TO.

And I believe the development and maintenance cost are very small.

>fully duplicate tables

I don't see why this is such a big deal. "Duplicate tables" are just a few lines of code. It's a `create table deleted_foo (id, bar, baz...`, and then for migrations,
```
ALTER TABLE deliveryservice ADD COLUMN ecs_enabled boolean;
ALTER TABLE deleted_deliveryservice ADD COLUMN ecs_enabled boolean;
```

It's a very small amount of code, and a very small amount of data. 

Yes, it's a maintenance cost, like every line of code. But why is it large, or difficult, or dangerous, more than any other dozen of code?

>fully duplicate queries

The query is not "fully duplicated" in code, the query parameters are abstracted, so it's

```
SELECT max(t) from (
SELECT max(last_updated) as t from `+ tableName + ` q ` + where + orderBy + pagination +
` UNION ALL
select max(last_updated) as t from deleted_`+ tableName + ` q ` + where + orderBy + pagination +
` ) as res`
```

Rather than 
```
SELECT max(t) from (
SELECT max(last_updated) as t from `+ tableName + ` q ` + where + orderBy + pagination +
` UNION ALL
select max(last_updated) as t from deleted_`+ tableName + ` where id = $1`
` ) as res`
```

The query difference is pretty small.

rawlinp: >  Without it, it'll mean any modification to any profile will make IMS fail on ORT for every server.

But we have a plan to address this in the near future (cache config snapshots), to where this would no longer matter for ORT anyways. So really, the IMS here is really for optimizing things like SREs doing things in Traffic Portal, etc, not for ORT (long-term).

> it doesn't seem unreasonable, it just means adding a column has to add it in both tables

That is where we get into trouble. We already have problems with keeping our nullable and non-nullable Go structs in sync with each other. Unless it's been hardcoded into your brain that you need to update things in all the various places that they're duplicated, you will forget to do so. Duplication is not ideal. It's also an additional, ongoing cost to developing new features, not just maintenance.

>  Our data is tiny.

I'm not really concerned about the size of the data, I'm more concerned about the _contents_ of the data. Data is a liability. We have things like passwords, physical addresses, IP addresses, etc. -- all things that don't really make me comfortable knowing that we'd be holding onto them indefinitely.

> The savings are not theoretical

By theoretical I meant in the sense of the _amount_ of savings. I'm not disagreeing that successful IMS requests are _actual_ savings.

> and I don't believe the cases are rare.

I guess that really depends on what we each think "rare" means. What do you consider "rare"? Once a day? A week? A month?

For reference, in the last ninety days at my company, we only issued 104 `DELETE` TO API requests. Now, that isn't including requests like `POST /profileparameter` that actually delete things under the hood, but we had 12 of those `POST /profileparameter` requests in the last ninety days. So we're talking about slightly over one thing a day on average that could potentially invalidate cache for a certain table. Also, they appeared to mostly be DELETEs of servers and delivery services, which make sense, but it's important to also note that those are also the kinds of resources that are typically asked for in large quantities via GET requests.

Think about opening either of those tables (servers and DSes) in TP -- by default TP requests _all_ of those, and those are easily the two slowest pages in TP. If there were _any_ servers or DSes deleted, it doesn't matter if we tracked every single one deleted or simply tracked that certain tables had been deleted from. That cache would be invalidated the same either way. Why not try to keep the implementation simple? 

> The query is not "fully duplicated" in code, the query parameters are abstracted

I was referring to the `insert into deleted_table` queries -- they contain a lot of duplicated code with their corresponding `insert into table` and `select from table` queries. We would need to be careful to share as much as possible between those three queries rather than having three separate, mostly-duplicated queries, to keep in sync.

rob05c: >we have a plan to address this in the near future (cache config snapshots)

I see solutions like denormalized outside-the-API JSON blob snapshots or intermediary proxies as hacks to work around a broken system, not the ideal. If Traffic Ops were reasonably performant and implemented HTTP standards, and relatively simple performance optimizations like IMS, a 1s cache, and Read-While-Writer on concurrent requests, expensive and unsafe workarounds wouldn't be necessary. 

There are scales where the cost of these things becomes necessary; <1 req/s on a <100MB database are several orders of magnitude below those levels.

>in the last ninety days at my company, we only issued 104 DELETE

Even one DELETE a day, if it's breaking an otherwise successful IMS, means increasing the TO load by thousands of times the necessary, for some brief period. If that change was one cache, it means the server has to handle full requests from every cache in the CDN checking in at once, instead of a single full request and 999 cheap IMS's.

>We would need to be careful to share as much as possible between those three queries rather than having three separate, mostly-duplicated queries

String building isn't hard. In fact, we already have columns duplicated in places, which should really concatenate strings to avoid duplicating all the column names.

At any rate, I'm not going to -1. I'm not a primary developer on Traffic Ops anymore, and I really don't like telling other people how hard something should be for them. If decision is that it's too difficult and expensive to maintain real IMS, well, I'm not writing (much) code on the project anymore to say.

But I will say, Traffic Ops is at the center of everything else in Traffic Control. If TO would do the right thing -- HTTP standards like IMS and Delta, Read-While-Writer, 1s caching, other small things to get decent performance -- these things would be a tiny, tiny fraction of the code TO has to maintain, and would make things better for every other component of Traffic Control, as well as all its operators and users.

The cost of these small features in TO is less than cached blobs and slower configs, and far, far less than additional intermediary proxy service deployments.

TO has had abysmal performance since the beginning. Golang is much better now, but the scalability of TO, how many requests it can handle, is still orders of magnitude lower than it should be.

The things we have to do to work around poor TO performance -- proxies in front of TO, ORT dispersion and backoff and slower config deployment, TM slower health updating -- everything else in TC would be so, so much better if we were just willing to do the work to make TO performant. And the development and operational time of all those things far, far outweighs the time for TO.

ocket8888: Looking around at the way other projects implement this, it looks like they typically do one of two things:

- Implement a `last_deleted` timestamp table with one row for each table that just stores the timestamp of the last deletion. That returns payloads which invalidate unnecessarily sometimes but are still a useful improvement over nothing, or
- Don't support IMS on collections of objects. Just use ETags for collections, and support IMS on single objects.

Those both seem fine to me - and I didn't look too hard at the DB implementation tbh because I thought he wanted me specifically to look at the API code and it's still just a draft - but I share Rawlin's reservations about creating tables to shadow all data that ever existed in the database.

srijeet0406: Refactored the DELETE logic to use just one table with just the last timestamp of the deleted member of that particular "type".

rawlinp: >  I see solutions like denormalized outside-the-API JSON blob snapshots or intermediary proxies as hacks to work around a broken system, not the ideal.

I wouldn't really call it a hack. I think it's actually a step in the right direction that will help us achieve much greater performance and efficiency in dishing out config data to thousands of caches at once. Since caches are by far the biggest client base of the TO API, it makes sense to optimize their requests as much as possible, and I think the cache-config snapshots idea is the simplest way to give us the most optimal solution for those requests. It can even have all those nice features -- perfect IMS, delta encoding, read-while-writer, 1s caching -- at a much cheaper price than implementing them all across the entire API since it's just a single endpoint.

_Back to the PR at hand:_

I've also been thinking, we really need to make sure the IMS DB queries are significantly cheaper compared to the non-IMS DB queries. Currently, I'm pretty sure that they will do a full sequential scan in order to find the most recent `last_updated` because it isn't indexed, and that isn't really much cheaper on the DB than sequentially scanning and returning all of the data. So IMS hits would save bandwidth, but not load on the DB, which I think may be more important. It also basically doubles the cost of an IMS miss, and we don't really know what our IMS hit percentage would be in practice.

I was playing around with indexing `last_updated`, and this seems to make the `select max(last_updated)` queries much cheaper:
```
create index deliveryservice_last_updated_idx on deliveryservice (last_updated DESC NULLS LAST);
```
The `DESC` seems to be important because, in theory, the first timestamp in the index should be the max, so this should be a very cheap query compared to a sequential scan.

There's also the chance that we're wrong about the DB performance in practice, if things are updated often enough to cause way more IMS misses than expected. In that case, it would be nice to have a feature flag to disable IMS handling. That also means we should probably log whether a request was IMS_HIT, IMS_MISS, NON_IMS, etc. Then we will be able to determine what our hit percentage is from the logs.

rob05c: Ah, yes, if `last_updated` doesn't have an index, it definitely needs one for these queries. +1
And +1 on logging

rawlinp:
Just so we don't forget, we probably need to add the `Last-Modified` header to all responses in order for this to be fully supported, right? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified

It seems like we need to set that to the latest `last_modified` in the response body.

rob05c:
>we probably need to add the Last-Modified header to all responses in order for this to be fully supported, right?
We should. It isn't strictly necessary, clients can use the Date. But it's better, and there's no reason not to.

rob05c:
>It seems like we need to set that to the latest last_modified in the response body.

Also, IMS is second-precision: so for headers, as well as all the comparisons to determine if it's been modified, we need to round down to the previous second.

This isn't a huge deal for users, but it is for scripts. If we don't round down, a fast script which updates-then-retrieves could easily not get the change it just made, and break.

Also: when we implement ETags, this isn't an issue, they can have as much precision as we like. We should definitely add ETags and If-None-Match support ASAP, which should be easy with this framework. ETags are always better. They're not only more precise, but they can have IDs, or paths (e.g. to ensure someone doesn't erroneously send the etag from `/servers?name=foo` to `/servers?name=bar`), or any other data we need to put in there.

srijeet0406:
@rawlinp @rob05c I'm close to being done with the logging and indexing changes. I'll also add the `Last-Modified` header to the responses and update the PR. I believe there's another ticket for the `ETags`, not fully sure.

rawlinp:
Since ETags are generally better, would it make sense to just implement them instead of IMS? Do we need both? Or do we just replace IMS support with ETags in the future?

rob05c:
IMO we should do both. All the logic only happens once, it's only a couple more lines to add the IMS and LM headers, and some clients can do IMS but not INM.

srijeet0406:
Here are the fixes in the latest commit:
1.) Now we are logging on an IMS hit/ miss, or on a non IMS vs IMS request 
2.) All the responses now have a `Last-Modified` header which has the "latest" ts of the response set
3.) We are now able to turn the IMS mode on/ off by using the `run_ims` flag (which defaults to `false`)

traffic_ops/traffic_ops_golang/traffic_ops_golang.go
```diff
@@ -59,6 +59,7 @@ func main() {
 	configFileName := flag.String("cfg", "", "The config file path")
 	dbConfigFileName := flag.String("dbcfg", "", "The db config file path")
 	riakConfigFileName := flag.String("riakcfg", "", "The riak config file path")
+	useIMS := flag.Bool("use_ims", false, "Use If-Modified-Since requests")
```
ocket8888, line 62: Can we make this `use-ims` instead? Hyphens are easier to type in command line flags.

Also, the new option should be added to [the Traffic Ops usage section](https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_ops.html#traffic-ops-golang) as well as to the usage example therein.

rawlinp: Rather than being a CLI flag this should probably just be an option in TO's config file instead.

ocket8888: I'm good with that too.

In that case it should be noted in the [TO configuration file section](https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_ops.html#cdn-conf).

srijeet0406: I'll get that changed.

srijeet0406: Done

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -118,28 +125,86 @@ func GenericCreateNameBasedID(val GenericCreator) (error, error, int) {
...
+	} else {
+		query := val.SelectMaxLastUpdatedQuery(where, orderBy, pagination, val.GetType())
+		rows, err := val.APIInfo().Tx.NamedQuery(query, queryValues)
+		defer rows.Close()
+		if err != nil {
+			log.Warnf("Couldn't get the max last updated time: %v", err)
+			return runSecond, max
+		}
+		if err == sql.ErrNoRows {
+			runSecond = false
```
rawlinp, line 151: instead of setting `runSecond = false` before returning it, would it be more clear to create a 2nd variable at the top - `dontRunSecond = false` - and return that instead?

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -118,28 +125,86 @@ func GenericCreateNameBasedID(val GenericCreator) (error, error, int) {
...
+	vals := []interface{}{}
+	code := http.StatusOK
 	where, orderBy, pagination, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(val.APIInfo().Params, val.ParamColumns())
 	if len(errs) > 0 {
-		return nil, util.JoinErrs(errs), nil, http.StatusBadRequest
+		return nil, util.JoinErrs(errs), nil, http.StatusBadRequest, nil
 	}
-
+	runSecond, maxTime := MakeFirstQuery(val, h, where, orderBy, pagination, queryValues)
+	if useIMS {
```
rawlinp, line 182: if `useIMS = false` I think we should probably skip the entire first query to disable as much of the IMS functionality as possible

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -118,28 +125,86 @@ func GenericCreateNameBasedID(val GenericCreator) (error, error, int) {
...
+		if !runSecond {
+			log.Debugln("IMS HIT")
```
rawlinp, line 184: oh, by logging IMS hits vs misses, I was thinking it should be in the TO access logs, similar to the ATS cache result codes: https://docs.trafficserver.apache.org/en/8.0.x/admin-guide/logging/cache-results.en.html#cache-result-codes

We could have IMS_HIT, IMS_MISS, and NON_IMS. Ideally, each result code would be documented (I'm not sure if the TO access log is documented already or not).

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -20,9 +20,15 @@ package api
  */
 
 import (
+	"database/sql"
 	"errors"
 	"fmt"
+	"github.com/apache/trafficcontrol/grove/web"
+	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-rfc"
+	ims2 "github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/util/ims"
```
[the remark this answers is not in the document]
srijeet0406: Aah yeah my goland does this by itself sometimes, I'll change this.

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -118,28 +125,86 @@ func GenericCreateNameBasedID(val GenericCreator) (error, error, int) {
...
+		return runSecond, max
+	}
+	ims = h[rfc.IfModifiedSince]
+	if len(ims) == 0 {
+		return runSecond, max
+	}
+	if l, ok := web.ParseHTTPDate(ims[0]); !ok {
+		log.Warnf("Date not parsable %v", ims[0])
+		return runSecond, max
+	} else {
```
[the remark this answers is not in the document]
srijeet0406: Yeah, but the variable `l` is local to the if/else block scope, hence I had to go this route.

traffic_ops/traffic_ops_golang/api/generic_crud.go
srijeet0406: I was just trying to minimize the number of variables that I use in the function. I always find it easier to read a piece of code if there is one variable tracking the state of the system, rather than multiple. Just a personal preference, but let me know if you'd like me to change it.

traffic_ops/app/db/migrations/20200430000000_add_on_delete_trigger.sql
```diff
@@ -0,0 +1,419 @@
+-- +goose Up
```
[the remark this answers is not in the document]
srijeet0406: The new tables (ip_address and interface) do not have `lastUpdated` fields so I dont think we'll be needing entries for those tables. But I agree about the name of the migration, I have changed that.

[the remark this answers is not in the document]
srijeet0406: Oh nvm, I see the topology and other tables have the lastUpdated field, will update those.

traffic_ops/traffic_ops_golang/api/generic_crud.go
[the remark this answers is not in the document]
rawlinp: Ah, yeah, but `l` doesn't _have_ to be local to the if/else block scope. Also, what about naming it `imsDate`? `l` doesn't really convey what it is and makes `if l.After(v.LatestTime.Time) {` harder to understand.

traffic_ops/traffic_ops_golang/api/generic_crud.go
rawlinp: I find that unnecessary state changes make the code harder to follow -- since it is primarily just the return value, there's no real need for it to change state -- they could be `const` too.

`return dontRunSecond, max` and `return runSecond, max` makes it more clear IMO, but if you don't want to change it that's fine too. Just a suggestion.

srijeet0406: I changed it.

traffic_ops/app/db/migrations/20200520000000_add_deleted_tables.sql
```diff
@@ -0,0 +1,470 @@
+/*
```
rawlinp, line 1: There are some new tables to account for here: https://github.com/apache/trafficcontrol/pull/4687/files

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
```diff
@@ -114,11 +117,20 @@ func ReadDSSHandler(w http.ResponseWriter, r *http.Request) {
 
 	dss := TODeliveryServiceServer{}
 	dss.SetInfo(inf)
-	results, err := dss.readDSS(inf.Tx, inf.User, inf.Params, inf.IntParams, nil, nil)
+	results, err, maxTime := dss.readDSS(nil, inf.Tx, inf.User, inf.Params, inf.IntParams, nil, nil, false)
 	if err != nil {
 		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, err)
 		return
 	}
+	if maxTime != nil {
```
rawlinp, line 125: There are conflicts in this file now due to the total reworking of the servers API here: https://github.com/apache/trafficcontrol/pull/4700/files

rawlinp:
Oh and thinking about DB-related tests to make sure each table is accounted for w.r.t. the deleted_tables table (and indexing last_updated), I think might be able to do that if we get creative with that in the DB tests: https://github.com/apache/trafficcontrol/blob/master/traffic_ops_db/test/docker/run-db-test.sh

Basically, it would be great if that test failed whenever someone forgot to account for the IMS-related behavior, in order to ensure IMS is maintained for new endpoints/resources.

traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
```diff
@@ -436,13 +438,24 @@ func (cg *TOCacheGroup) Read() ([]interface{}, error, error, int) {
 	}
 	where, orderBy, pagination, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(cg.ReqInfo.Params, queryParamsToQueryCols)
 	if len(errs) > 0 {
-		return nil, util.JoinErrs(errs), nil, http.StatusBadRequest
+		return nil, util.JoinErrs(errs), nil, http.StatusBadRequest, nil
+	}
+
+	runSecond, maxTime := ims.TryIfModifiedSinceQuery(cg.APIInfo().Tx, h, queryValues, selectMaxLastUpdatedQuery(where, orderBy, pagination))
+	if useIMS {
```
rawlinp, line 445: So it looks like you addressed this in generic_crud.go but we should also skip the `TryIfModifiedSinceQuery` if IMS is disabled. I noticed this here and in topologies.go, so I imagine there might be other cases to still address in other files as well.

traffic_ops/client/cdn.go
```diff
@@ -100,10 +100,33 @@ func (to *Session) GetCDNByID(id int) ([]tc.CDN, ReqInf, error) {
...
 }
 
+func (to *Session) GetCDNByNameIMS(name string, header http.Header) ([]tc.CDN, ReqInf, error) {
+	url := fmt.Sprintf("%s?name=%s", API_CDNS, url.QueryEscape(name))
+	resp, remoteAddr, err := to.requestIMS(http.MethodGet, url, header)
+	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
+	if resp != nil {
+		reqInf.StatusCode = resp.StatusCode
+	}
+	if resp.StatusCode == http.StatusNotModified {
```
rawlinp, line 110: this should be moved into the previous block because `resp` might be nil here

traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
```diff
@@ -444,7 +445,7 @@ func (cg *TOCacheGroup) Read(h http.Header, useIMS bool) ([]interface{}, error,
 	}
 
 	if useIMS {
-		runSecond, maxTime := ims.TryIfModifiedSinceQuery(cg.APIInfo().Tx, h, queryValues, selectMaxLastUpdatedQuery(where))
+		runSecond, maxTime = ims.TryIfModifiedSinceQuery(cg.APIInfo().Tx, h, queryValues, selectMaxLastUpdatedQuery(where))
```
mitchell852, line 448: ok, just want to make sure that if you don't provide the following header:

```
-H "If-Modified-Since: [date]"
```

then you get back 

```
< last-modified: Mon, 01 Jan 0001 00:00:00 UTC
```

mitchell852: just want to make sure that's the expected result.

srijeet0406: Correct, I set the "last-modified" in the response only if we get a valid IMS request header, and if the "use_ims" flag is set to true in the conf file.

docs/source/admin/traffic_ops.rst
```diff
@@ -443,6 +443,10 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 		:disabled_routes: A list of API route IDs to disable. Requests matching these routes will receive a 503 response. To find the route ID for a given path you would like to disable, run ``./traffic_ops_golang`` using the :option:`--api-routes` option to view all the route information, including route IDs and paths.
 		:ignore_unknown_routes: If ``false`` (default) return an error and prevent startup if unknown route IDs are found. Otherwise, log a warning and continue startup.
 
+:use_ims:
+    .. versionadded:: 5.0
+    This is an optional boolean value to enable the handling of the "If-Modified-Since" HTTP request header. Default: false
```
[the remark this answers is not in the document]
srijeet0406: Done

traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
mitchell852: i guess i expected that the `last-modified` key would not even be in the response at all if IMS was not in the request header

srijeet0406: Well, I have the logic set currently so that if we get a valid (not nil) pointer in the `maxtime` variable, we set the `last-modified` value to that. So, if we get back an empty maxTime (which is the case if you dont supply an IMS header, or if the flag is set to false), we set the last-modified to `Mon, 01 Jan 0001 00:00:00 UTC`.

PR #4666, opened 2020-04-27
Fix segfault in enroller.go

shamrickus:

- [x] This PR fixes a segfault that occurs when enrolling profiles and a given profile is not created successfully.

- CDN in a Box

Ensure that enroller still works properly. If possible, confirm that it no longer segfaults (not sure how to deterministically recreate the issue).

### DIFF
--- a/infrastructure/cdn-in-a-box/enroller/enroller.go
+++ b/infrastructure/cdn-in-a-box/enroller/enroller.go
@@ -560,8 +560,11 @@ func enrollProfile(toSession *session, r io.Reader) error {
 			}
 		}
 		profiles, _, err = toSession.GetProfileByName(profile.Name)
-		if err != nil || len(profiles) == 0 {
+		if err != nil {
 			log.Infof("error getting profile ID from %+v: %s\n", profile, err.Error())
+		}
+		if len(profiles) == 0 {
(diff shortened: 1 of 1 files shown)


infrastructure/cdn-in-a-box/enroller/enroller.go
```diff
@@ -560,9 +560,12 @@ func enrollProfile(toSession *session, r io.Reader) error {
...
 		}
 		profiles, _, err = toSession.GetProfileByName(profile.Name)
-		if err != nil || len(profiles) == 0 {
+		if err != nil {
 			log.Infof("error getting profile ID from %+v: %s\n", profile, err.Error())
 		}
+		if len(profiles) == 0 {
+			log.Infof("no results returned for getting profile ID from %+v", profile)
+		}
 		profile = profiles[0]
```
rawlinp, line 569: This could panic, right? This seems like a fatal error anyways, but maybe the enroller should exit gracefully instead.

shamrickus: I agree, I didn't make the change because for all the other enroller methods, they only log not panic. I can change all of them but I figured something like that is better in it's own PR due to the potential side affects.

PR #4682, opened 2020-05-01
Clean up docker integration test and add documentation

shamrickus:

- [x] This PR is not related to any issue.

This cleans up the dockerfile for the TP integration test, as well as update the documentation with how to run the TO/TP integration tests. It also ignores a folder created for the docker integration tests.

- CDN in a Box
- Documentation

Build docs, follow instructions to run Docker integration containers.

### DIFF
--- a/docs/source/admin/quick_howto/ciab.rst
+++ b/docs/source/admin/quick_howto/ciab.rst
@@ -107,7 +107,7 @@ In order to check the "readiness" of your CDN, you can optionally start the Read
 Integration Tests
 """""""""""""""""
 
-There also exists TP and TO integration tests containers. Both of these containers assume that CDN in a Box is already running on the local system.
+There also exist TP and TO integration tests containers. Both of these containers assume that CDN in a Box is already running on the local system.
 
 .. code-block:: shell
 	:caption: Running TP Integration Tests
--- a/docs/source/admin/quick_howto/ciab.rst
+++ b/docs/source/admin/quick_howto/ciab.rst
@@ -104,6 +104,21 @@ In order to check the "readiness" of your CDN, you can optionally start the Read
 
 	sudo docker-compose -f docker-compose.readiness.yml up
 
+Integration Tests
+"""""""""""""""""
+
+There also exists TP and TO integration tests containers. Both of these containers assume that CDN in a Box is already running on the local system.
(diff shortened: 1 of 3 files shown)


docs/source/admin/quick_howto/ciab.rst
```diff
@@ -104,6 +104,21 @@ In order to check the "readiness" of your CDN, you can optionally start the Read
 
 	sudo docker-compose -f docker-compose.readiness.yml up
 
+Integration Tests
+"""""""""""""""""
+
+There also exists TP and TO integration tests containers. Both of these containers assume that CDN in a Box is already running on the local system.
```
ocket8888, line 110: s/exists/exist/

infrastructure/cdn-in-a-box/traffic_portal_integration_test/Dockerfile
```diff
@@ -31,21 +31,13 @@ gpgkey=https://dl-ssl.google.com/linux/linux_signing_key.pub \n\
 " > /etc/yum.repos.d/google-chrome.repo
 
 RUN yum update -y
-
 RUN yum install -y java-1.8.0-openjdk git google-chrome-stable bind-utils net-tools
-
-RUN curl -LO https://nodejs.org/dist/v12.16.3/node-v12.16.3-linux-x64.tar.xz
```
ocket8888, line 37: Doesn't getting rid of this cause the subsequent `npm install` steps to fail?

shamrickus: `curl -LO` only downloads the file, there is nothing done with it (as far as I can tell) after the download.

ocket8888: It's un-tarred into `/usr/local` on [line 37 of your branch](https://github.com/apache/trafficcontrol/pull/4682/files#diff-25b7184084d757a7bbf533c72127e049R37) - which means that as of now the integration container fails to actually build.

shamrickus: That is interesting, it was definitely still working when I tested. Must have had something cached.

PR #4685, opened 2020-05-06
Parameter filtering should take the cdn name for filtering

### LINKED ISSUE #1903: /api/1.2/cdns/my-cdn-name/configs/monitoring.json is not CDN/Profile specific
If you have multiple profiles on different CDN with different parameters, you'll end up with unwanted results.

For example, if I have 2 CDNs and enable `peers.polling.interval` on only a single RASCAL profile, it will show up on both CDNs Traffic Monitor.

srijeet0406:

- [x] This PR fixes #1903 

- CDN in a Box
- Traffic Monitor
- Traffic Ops

Run the api tests and unit tests.
Test it out on CiaB/ running a query in the DB.
In my CiaB setup, I had a couple of CDNs that were using profiles with common parameters. When I ran the query to get the configs for one of the CDNs, the other params showed up in the results as well. This is fixed with the current behavior.
Query without the fix: 
`SELECT pr.name, pr.value
FROM parameter pr
JOIN profile p ON p.name LIKE 'RASCAL%%'
JOIN profile_parameter pp ON pp.profile = p.id and pp.parameter = pr.id
WHERE pr.config_file = 'rascal-config.txt'`

Query with fix:
`SELECT pr.name, pr.value
FROM parameter pr
JOIN profile p ON p.name LIKE 'RASCAL%%'
JOIN profile_parameter pp ON pp.profile = p.id and pp.parameter = pr.id
JOIN cdn c ON c.id=p.cdn
WHERE pr.config_file = 'rascal-config.txt'
AND c.name='<ENTER YOUR CDN NAME HERE>'`

master

### DIFF
--- a/traffic_ops/traffic_ops_golang/monitoring/monitoring.go
+++ b/traffic_ops/traffic_ops_golang/monitoring/monitoring.go
@@ -115,7 +115,6 @@ func Get(w http.ResponseWriter, r *http.Request) {
 }
 
 func GetMonitoringJSON(tx *sql.Tx, cdnName string) (*Monitoring, error) {
-	fmt.Println("SRIJEET I'm HERE")
 	monitors, caches, routers, err := getMonitoringServers(tx, cdnName)
 	if err != nil {
 		return nil, fmt.Errorf("error getting servers: %v", err)
--- a/traffic_ops/traffic_ops_golang/monitoring/monitoring.go
+++ b/traffic_ops/traffic_ops_golang/monitoring/monitoring.go
@@ -135,7 +135,7 @@ func GetMonitoringJSON(tx *sql.Tx, cdnName string) (*Monitoring, error) {
 		return nil, fmt.Errorf("error getting deliveryservices: %v", err)
 	}
 
-	config, err := getConfig(tx)
+	config, err := getConfig(tx, cdnName)
 	if err != nil {
 		return nil, fmt.Errorf("error getting config: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


traffic_ops/traffic_ops_golang/monitoring/monitoring_test.go
```diff
@@ -454,7 +454,7 @@ func TestGetConfig(t *testing.T) {
 		t.Fatalf("creating transaction: %v", err)
 	}
 
-	sqlConfig, err := getConfig(tx)
+	sqlConfig, err := getConfig(tx, "")
```
zrhoffman, line 457: Instead of using an empty string for the CDN name, the mock query should specify a CDN name as an argument.

srijeet0406: Done

traffic_ops/traffic_ops_golang/monitoring/monitoring.go
```diff
@@ -375,15 +375,17 @@ AND ds.active = true
...
 	query := fmt.Sprintf(`
 SELECT pr.name, pr.value
 FROM parameter pr
 JOIN profile p ON p.name LIKE '%s%%'
 JOIN profile_parameter pp ON pp.profile = p.id and pp.parameter = pr.id
+JOIN cdn c ON c.id=p.cdn
 WHERE pr.config_file = '%s'
-`, tc.MonitorProfilePrefix, MonitorConfigFile)
+AND c.name='%s'
+`, tc.MonitorProfilePrefix, MonitorConfigFile, cdnName)
```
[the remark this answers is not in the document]
srijeet0406: Fixed to use prepared statements.

traffic_ops/traffic_ops_golang/monitoring/monitoring.go
```diff
@@ -115,6 +115,7 @@ func Get(w http.ResponseWriter, r *http.Request) {
 }
 
 func GetMonitoringJSON(tx *sql.Tx, cdnName string) (*Monitoring, error) {
+	fmt.Println("SRIJEET I'm HERE")
```
zrhoffman, line 118: We don't need this line, but you can leave it if you want :smile:

srijeet0406: Lol, no one saw anything 🤫

srijeet0406: Also, fixed the `TestGetConfig()` test

zrhoffman:
Because I did not mention it before: API and unit tests pass and code is correctly-formatted.

PR #4690, opened 2020-05-14
Add multiple interface support to TM CacheStats/cache-statuses

shamrickus:

- [x] This PR is not related to any issue but is one of many in addressing the addition of multiple interfaces to TC. 

- Documentation
- Traffic Monitor
- Traffic Ops

Verify TM/TO functions as usually using the legacy astats ATC plugin. If possible get a build of ATC 
(not released yet) with system_stats.so/stats_over_http.so to test with caches that have multiple interfaces.
Ensure the docs build and make sense.

### DIFF
--- a/traffic_monitor/health/cache.go
+++ b/traffic_monitor/health/cache.go
@@ -274,9 +274,9 @@ func CalcAvailability(results []cache.Result, pollerName string, statResultHisto
 				if status.Why != "" {
 					newWhyText := fmt.Sprintf("%s: %s", interfaceName, status.Why)
 					if aggregateStatus.Why != "" {
-						newWhyText += ", " + newWhyText
+						newWhyText = ", " + newWhyText
 					}
-					aggregateStatus.Why = newWhyText
+					aggregateStatus.Why += newWhyText
 				}
 
 				if status.UnavailableStat != "" {
--- a/traffic_monitor/health/cache.go
+++ b/traffic_monitor/health/cache.go
@@ -269,7 +269,7 @@ func CalcAvailability(results []cache.Result, pollerName string, statResultHisto
 
 				// What does this mean on aggregated data?
 				// For now assume that if any interface was then the aggregate is
-				aggregateStatus.LastCheckedIPv4 = aggregateStatus.LastCheckedIPv4.IPv4 || status.LastCheckedIPv4
+				aggregateStatus.LastCheckedIPv4 = aggregateStatus.LastCheckedIPv4 || status.LastCheckedIPv4
 
 				if status.Why != "" {
 					newWhyText := fmt.Sprintf("%s: %s", interfaceName, status.Why)
--- a/traffic_monitor/health/cache.go
+++ b/traffic_monitor/health/cache.go
@@ -43,20 +43,24 @@ func GetVitals(newResult *cache.Result, prevResult *cache.Result, mc *tc.Traffic
 	// proc.loadavg -- we're using the 1 minute average (!?)
 	newResult.Vitals.LoadAvg = newResult.Statistics.Loadavg.One
 
-	// For now, just getting the first encountered interface in the list of
-	// interfaces parsed out into statistics. In the future, vitals will need
-	// to consider all interfaces configured on a cache server.
-	for name, iface := range newResult.Statistics.Interfaces {
-		log.Infof("Cache '%s' has %d interfaces, we have arbitrarily chosen %s for vitality", newResult.ID, len(newResult.Statistics.Interfaces), name)
-		newResult.Vitals.BytesOut = iface.BytesOut
-		newResult.Vitals.BytesIn = iface.BytesIn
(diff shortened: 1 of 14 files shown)


traffic_monitor/threadsafe/cacheavailablestatus.go
```diff
@@ -7,7 +7,7 @@ package threadsafe
  * regarding copyright ownership.  The ASF licenses this file
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
- * with the License.  You may obtain a copy of the License at
+ * with the License.  You may obtain a copy of the License a{t
```
mattjackson220, line 10: dont think this one was on purpose

shamrickus: You are correct

traffic_monitor/health/cache.go
```diff
@@ -175,65 +180,127 @@ func CalcAvailability(results []cache.Result, pollerName string, statResultHisto
...
+			for interfaceName, status := range localCacheStatuses[tc.CacheName(result.ID)] {
+				if interfaceName == tc.CacheInterfacesAggregate {
+					continue
+				}
+				aggregateStatus.Available.IPv4 = aggregateStatus.Available.IPv4 || status.Available.IPv4
+				aggregateStatus.Available.IPv6 = aggregateStatus.Available.IPv6 || status.Available.IPv6
+
+				// What does this mean on aggregated data?
+				// For now assume that if any interface was then the aggregate is
+				aggregateStatus.LastCheckedIPv4 = aggregateStatus.Available.IPv4 || status.LastCheckedIPv4
```
mattjackson220, line 272: maybe copy leftover? should be `aggregateStatus.LastCheckedIPv4 = aggregateStatus.LastCheckedIPv4 || status.LastCheckedIPv4` right?

shamrickus: Yep, good catch

traffic_monitor/health/cache.go
```diff
@@ -152,6 +156,7 @@ func EvalCache(result cache.ResultInfo, resultStats *threadsafe.ResultStatValHis
 func CalcAvailability(results []cache.Result, pollerName string, statResultHistory *threadsafe.ResultStatHistory, mc tc.TrafficMonitorConfigMap, toData todata.TOData, localCacheStatusThreadsafe threadsafe.CacheAvailableStatus, localStates peer.CRStatesThreadsafe, events ThreadsafeEvents, protocol config.PollingProtocol) {
```
mattjackson220, line 156: in health/cache.go  in CalcAvailability  what are your thoughts on ending this for loop 
`for interfaceName, _ := range result.Interfaces() {`
earlier right before this line
```
// Compute aggregate data based on each interface
			aggregateStatus := cache.AvailableStatus{
```
and then this last line
`			localStates.SetCache(tc.CacheName(result.ID), tc.IsAvailable{IsAvailable: newAvailableState, Ipv4Available: availableTuple.IPv4, Ipv6Available: availableTuple.IPv6})`
should really use the aggregate status right? something like
```
			aggregateAvailableState := processAvailableTuple(aggregateStatus.Available, serverInfo)
			localStates.SetCache(tc.CacheName(result.ID), tc.IsAvailable{IsAvailable: aggregateAvailableState, Ipv4Available: aggregateStatus.Available.IPv4, Ipv6Available: aggregateStatus.Available.IPv6})
```
that way it doesnt calculate the aggregate stats for each loop through the interfaces and the local state of the cache is based on the aggregate instead of on just the last interface it found?

and then this part
```
			if available, ok := localStates.GetCache(tc.CacheName(result.ID)); !ok || available.IsAvailable != newAvailableState {
				protocol := "IPv4"
				if !usingIPv4 {
					protocol = "IPv6"
				}
				log.Infof("Changing state for %s was: %t now: %t because %s poller: %v on protocol %v error: %v", result.ID, available.IsAvailable, newAvailableState, whyAvailable, pollerName, protocol, result.Error)
				events.Add(Event{Time: Time(time.Now()), Description: "Protocol: (" + protocol + ") " + whyAvailable + " (" + pollerName + ")", Name: string(result.ID), Hostname: string(result.ID), Type: toData.ServerTypes[tc.CacheName(result.ID)].String(), Available: newAvailableState, IPv4Available: availableTuple.IPv4, IPv6Available: availableTuple.IPv6})
			}
```
should be moved to the bottom and based on the aggregate as well?

shamrickus: You are correct, originally I thought we were going to need to break out localStates by interface but that is not the case.

mattjackson220:
Looks great! Code looks good, docs build and make sense, TM unit tests pass, TO API and unit tests pass, tested TM with astats and looks good, tested TM with stats_over_http and looks good!

We should have an update in the ChangeLog for this though right?

shamrickus:
Yep you are correct

PR #4694, opened 2020-05-15
More Topologies validation

### LINKED ISSUE #4691: Internal Server Error when create/update a Topologies with invalid parent node
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

Create new topologies with wrong parent node. For example:
Do a POST `api/3.0/topologies` with this payload :
```
{
        "name": "my-topology",
        "description": "This is my topology",
        "nodes": [
                {
                        "cachegroup": "albany-ny-usa",
                        "parents": [
                                7
                        ]
                }
        ]
}
```

API will return Internal Server Error

Same thing for PUT `api/3.0/topologies`.  It also return Internal Server Error.
## Expected / new behavior:

It should return 400 Bad Request.

zrhoffman:

- [x] This PR fixes #4691 by checking whether a parent node's index exists
- This PR fixes a bug by only performing further validation on a topology if all of its nodes' cachegroups exist
- Adds a status code check to the negative API tests to enure that POSTing invalid topologies returns only 400-level response status codes, not 500-level

- Traffic Control Client (Go)
- Traffic Ops

Run Traffic Ops API tests, verify they still pass:
```shell script
docker-compose -f docker-compose.yml -f docker-compose.traffic-ops-test.yml up
```

- master (88f4f2aad0)

- [x] This PR includes tests
- [x] This PR is a bugfix, no documentation necessary

### DIFF
--- a/traffic_ops/client/session.go
+++ b/traffic_ops/client/session.go
@@ -307,7 +307,7 @@ func NewNoAuthSession(toURL string, insecure bool, userAgent string, useCache bo
 	}, useCache)
 }
 
-// ErrUnlessOk returns the response an error if the given Response's status code is anything but 200 OK. This includes reading the Response.Body and Closing it. Otherwise, the given response and a nil error is returned.
+// ErrUnlessOk returns the response an error if the given Response's status code is anything but 200 OK. This includes reading the Response.Body and Closing it. Otherwise, the given response and a nil error are returned.
 func (to *Session) ErrUnlessOK(resp *http.Response, remoteAddr net.Addr, err error, path string) (*http.Response, net.Addr, error) {
 	if err != nil {
 		return resp, remoteAddr, err
--- a/traffic_ops/client/session.go
+++ b/traffic_ops/client/session.go
@@ -394,6 +394,7 @@ func (to *Session) RawRequest(method, path string, body []byte) (*http.Response,
 type ReqInf struct {
 	CacheHitStatus CacheHitStatus
 	RemoteAddr     net.Addr
+	StatusCode     int
 }
 
 type CacheHitStatus string
--- a/traffic_ops/client/session.go
+++ b/traffic_ops/client/session.go
@@ -307,7 +307,7 @@ func NewNoAuthSession(toURL string, insecure bool, userAgent string, useCache bo
 	}, useCache)
 }
 
-// ErrUnlessOk returns nil and an error if the given Response's status code is anything but 200 OK. This includes reading the Response.Body and Closing it. Otherwise, the given response and error are returned unchanged.
+// ErrUnlessOk returns the response an error if the given Response's status code is anything but 200 OK. This includes reading the Response.Body and Closing it. Otherwise, the given response and a nil error is returned.
 func (to *Session) ErrUnlessOK(resp *http.Response, remoteAddr net.Addr, err error, path string) (*http.Response, net.Addr, error) {
 	if err != nil {
 		return resp, remoteAddr, err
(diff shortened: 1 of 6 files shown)


traffic_ops/client/topology.go
```diff
@@ -30,94 +30,118 @@ const (
...
+func (to *Session) CreateTopology(top tc.Topology) (*tc.TopologyResponse, ReqInf, error, int) {
+	var (
+		statusCode = http.StatusNotAcceptable
+		remoteAddr net.Addr
+	)
 	reqBody, err := json.Marshal(top)
 	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
 	if err != nil {
-		return nil, reqInf, err
+		return nil, reqInf, err, statusCode
```
rawlinp, line 41: Based on the mailing list discussion it sounds like consensus has been built around adding the status code to the `ReqInf` struct and returning it that way, so we should start following that pattern for these methods. Then in a separate PR someone could retroactively add the status code to `ReqInf` for the rest of the TO Go client methods.

zrhoffman: Following that pattern in c1140e42f4

rawlinp: Did you not want to start setting the status code in `ReqInf` in these Topologies methods? Or that likely means having to update all of its usage throughout the entire TO client?

zrhoffman: Let's do it. Setting the status code in 481038d58a.

traffic_ops/client/session.go
```diff
@@ -307,7 +307,7 @@ func NewNoAuthSession(toURL string, insecure bool, userAgent string, useCache bo
 	}, useCache)
 }
 
-// ErrUnlessOk returns nil and an error if the given Response's status code is anything but 200 OK. This includes reading the Response.Body and Closing it. Otherwise, the given response and error are returned unchanged.
+// ErrUnlessOk returns the response an error if the given Response's status code is anything but 200 OK. This includes reading the Response.Body and Closing it. Otherwise, the given response and a nil error are returned.
```
srijeet0406, line 310: Should be `the response and an error`
Also, shouldn't the comment say that this function returns three things?

traffic_ops/traffic_ops_golang/topology/validation.go
```diff
@@ -57,9 +57,14 @@ func checkForEdgeParents(nodes []tc.TopologyNode, cachegroups []tc.CacheGroupNul
 	node := nodes[nodeIndex]
 	errs := make([]error, len(node.Parents))
 	for parentIndex := range node.Parents {
-		cacheGroupType := cachegroups[node.Parents[parentIndex]].Type
+		cachegroupIndex := node.Parents[parentIndex]
```
srijeet0406, line 60: This could be a part of the `range` statement above, wherein you'd get the index and the value in the step.

traffic_ops/client/session.go
zrhoffman: Thanks, fixed the typo and mentioned the remote address in 971302817.

traffic_ops/traffic_ops_golang/topology/validation.go
zrhoffman: Changed in 38838fa76

PR #4700, opened 2020-05-19
Servers interfaces

### LINKED ISSUE #4612: Update /servers and /servers/{{ID}} to use data from the interfaces tables
## I'm submitting a ...
- new feature / enhancement request

## Description
The `/servers` endpoint (and `/servers/{{ID}}` which uses the same handler) need to be updated to utilize data from the new tables added by whatever PR resolves #4611.

This depends on #4611

- Documentation
- Traffic Ops
- Traffic Ops client (Go)

## Acceptance Criteria
* `/servers` in API 1.x and 2.x:
	* Uses the tables to provide the right legacy output for the `GET` method (should be unchanged)
	* Properly translates its legacy input into data for the new tables for the `POST` method (behavior should be unchanged)
	* Properly translates its legacy input into data for the new tables for the `PUT` method (behavior should be unchanged)
* `/servers` in API 3.x:
	* Provides the proper representations of cache servers as outlined in the blueprint for the `GET` method
	* Properly validates input to the `POST` and `PUT` methods, according to the rules laid out in the blueprint
	* New behavior adheres to the API guidelines (where applicable)
	* All behavior changes have been documented
* CDN-in-a-Box automated data insertion has been updated to use the new format (probably just need to edit a JSON template)
* The Traffic Ops Go client has been updated (v3 only) to return the new structures
	* Any associated unit and/or client/API integration tests have been updated to
	  test the new structure and all pass.

See the [Additional Server Interfaces blueprint](https://github.com/apache/trafficcontrol/blob/master/blueprints/multi-interface-servers.md#serversserversid) for details.

ocket8888:
- [x] This PR fixes #4612

This PR adds support for the new server interfaces tables to all methods of the
`/servers` and `/servers/{{ID}}` TO API endpoints.

Regarding _"CDN-in-a-Box automated data insertion has been updated to use the
new format..."_ from #4612: The enroller has instead been pinned to version 2.x
of the TO API, which works just as well.

This also adds the necessary documentation.

Finally, this PR removes the configuration file generation route handlers for `parent.config`, `remap.config`, and `ip_allow.config`, as they were in conflict with the changes and have been deprecated for at least a major release.

However, that **does break CDN-in-a-Box for the forseeable future** (to be exact, until #4671 is merged)

- CDN in a Box
- Documentation
- Traffic Control Client
- Traffic Ops

Run all unit and API/client integration tests, perform some manual HTTP requests
to verify expected behavior. Also build and read the documentation.

- [x] This PR includes tests


traffic_ops/traffic_ops_golang/ats/atsserver/ipallowdotconfig.go
```diff
@@ -77,26 +80,45 @@ func GetIPAllowDotConfig(w http.ResponseWriter, r *http.Request) {
...
-    SELECT
-      cg2.id
-    FROM
-     server s2
-     JOIN cachegroup cg2 ON (cg2.parent_cachegroup_id = s2.cachegroup OR cg2.secondary_parent_cachegroup_id = s2.cachegroup)
-    WHERE
-      s2.host_name = $1 )
-  )
-`
+	SELECT
```
rawlinp, line 83: Rather than fixing up deprecated endpoints we should consider just removing these now, since we already know we need to get rid of them. That seems more practical than keeping these alive just to keep the ATS config functionality via TP going and keeping CIAB working (which you seem to have fixed already in your ORT.py PR which seems done and just needs review?).

ocket8888: ugh, yeah, you're right. The thing is that the ORT.py PR is waiting on my giving-ORT-its-own-build PR which in turn is waiting on the apache org to set up some dockerhub repos, which - finally - is waiting on me to give them a list of DockerHub users to add as admins for the repos they will create. Which I was just reminded of seconds ago, as it happens, and am about to do.

rawlinp: If we removed the routes now, how badly would CIAB fail? Obviously the caches wouldn't be able to handle requests, but is there anything else? Does it prevent other service from starting?

If is just breaks the caching in CIAB but all the other services start up fine, I don't really see a problem breaking that functionality temporarily (as well as the TP functionality). Most of the automated testing I know about that depends on CIAB doesn't require the caches to successfully handle requests (unless you count the readiness container).

ocket8888: It is pretty much just the caches. The only thing that's truly concerning, then, is the readiness container - which established that CiaB is "ready" if and only if DS URL curls succeed. I guess it's not a huge deal just to break caching - although that does give me a nice, warm feeling sometimes when I'm testing things with far-reaching changes - but I know that I have tools already that rely on readiness, and I just hope no one else does.

lib/go-tc/servers.go
```diff
@@ -25,7 +30,17 @@ import (
  * under the License.
  */
 
-// ServersResponse is a list of Servers as a response.
+// ServersResponsev3 is the format of a response to a GET request for /servers.
+type ServersResponsev3 struct {
```
shamrickus, line 34: `v3` should probably be capitalized.

lib/go-tc/servers.go
```diff
@@ -49,6 +64,157 @@ type ServersV1DetailResponse struct {
...
+	if !ok {
+		return fmt.Errorf("expected deliveryservice in byte array form; got %T", src)
+	}
+
+	return json.Unmarshal([]byte(b), sii)
+}
+
+// LegacyInterfaceDetails is the details for interfaces on servers for API v1 and v2.
+type LegacyInterfaceDetails struct {
+	InterfaceMtu  *int    `json:"interfaceMtu" db:"interface_mtu"`
```
shamrickus, line 103: `Mtu` should be capitalized

lib/go-tc/servers.go
```diff
@@ -49,6 +64,157 @@ type ServersV1DetailResponse struct {
...
+				}
+			}
+
+			if parsedIp.To4() == nil {
+				legacyDetails.IP6Address = &address
+				legacyDetails.IP6Gateway = gateway
+			} else if mask != nil {
+				legacyDetails.IPAddress = util.StrPtr(parsedIp.String())
+				legacyDetails.IPGateway = gateway
+				legacyDetails.IPNetmask = util.StrPtr(fmt.Sprintf("%d.%d.%d.%d", mask.Mask[0], mask.Mask[1], mask.Mask[2], mask.Mask[3]))
```
shamrickus, line 201: Wouldn't it be possible for the legacy interface to have the netmask field set for a different address/gateway?
E.g.
```
cache
  eth0 
     addr: x.y.w.z
     gateway: x.y.w.z
     netmask: x.y.w.z
  lo:
    addr: i,j,k,l
    gateway: i,j,k,l
```
resulting in a legacy interface that has the `lo` addr/gateway with the `eth0` netmask.

It seems to me that we should unset the IPNetmask in the else below.

lib/go-tc/servers.go
```diff
@@ -49,6 +64,157 @@ type ServersV1DetailResponse struct {
...
+			} else {
+				legacyDetails.IPAddress = util.StrPtr(parsedIp.String())
+				legacyDetails.IPGateway = gateway
+			}
+
+			if intFace.MTU != nil {
+				legacyDetails.InterfaceMtu = util.IntPtr(int(*intFace.MTU))
+			}
+
+			legacyDetails.InterfaceName = &intFace.Name
```
shamrickus, line 211: It might not be worth the effort, but once we have gathered the minimum "full" interface we could short circuit the return here as to prevent useless iterations.

lib/go-tc/servers_test.go
```diff
@@ -0,0 +1,58 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import "fmt"
+
+func ExampleLegacyInterfaceDetails_ToInterfaces() {
```
shamrickus, line 24: This isn't a test

traffic_ops/testing/api/v3/deliveryservices_required_capabilities_test.go
```diff
@@ -181,16 +182,23 @@ func InvalidDeliveryServicesRequiredCapabilityAddition(t *testing.T) {
 	}
 
 	// First assign current capabilities to edge server so we can assign it to the DS
-	servers, _, err := TOSession.GetServerByHostName("atlanta-edge-01")
+	// TODO: DON'T hard-code hostnames!
```
shamrickus, line 185: TODO
`testData.Servers[0].HostName`?

traffic_ops/testing/api/v3/deliveryserviceservers_test.go
```diff
@@ -113,8 +119,10 @@ func CreateTestMSODSServerWithReqCap(t *testing.T) {
 	}
 
 	// Associate origin server to msods1 even though it does not have req cap
-
-	servers, _, err := TOSession.GetServerByHostName("denver-mso-org-01")
+	// TODO: DON'T hard-code server hostnames!
```
shamrickus, line 122: TODO

lib/go-tc/servers.go
```diff
@@ -49,6 +64,157 @@ type ServersV1DetailResponse struct {
 	Alerts
 }
 
+// ServerIpAddress is the data associated with a server's interface's IP address.
+type ServerIpAddress struct {
```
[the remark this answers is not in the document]
ocket8888: Agree

lib/go-tc/servers.go
ocket8888: It should, but it can't be for backward-compatibility. This isn't a new property, it's moved from the old struct.

lib/go-tc/servers.go
ocket8888: This shouldn't be an issue for well-formed `ServerInterfaceInfo`s because they can only legally contain one IPv4 service address, but that's probably still a good idea.

lib/go-tc/servers.go
ocket8888: That's a bit difficult to figure out, best I can do is return if we've already found an IPv6 and an IPv4 - for servers that only have one or the other as a service address, you'll still need to iterate the entire thing.

How's it look now?

lib/go-tc/servers_test.go
ocket8888: [Yes it is](https://blog.golang.org/examples)

shamrickus: I stand corrected, carry on

traffic_ops/traffic_ops_golang/ats/atsserver/ipallowdotconfig.go
ocket8888: Alright, I tore the affected endpoints out.

lib/go-tc/servers.go
shamrickus: Looks good

traffic_ops/testing/api/v1/atsconfig_test.go
```diff
@@ -108,3 +105,83 @@ func GetTestATSConfigs(t *testing.T) {
...
+		t.Errorf("cannot GET Servers: %v", err)
+	}
+	if len(servers) < 1 {
+		t.Error("GET Servers returned no servers, must have at least 1 to test ds-servers")
+	}
+
+	for _, ds := range dses {
+		serverIDs := make([]int, 0, len(servers))
+		for _, server := range servers {
+			if server.Type == "EDGE" && server.CDNName == ds.CDNName {
```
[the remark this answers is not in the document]
ocket8888: That's not my logic, I just moved it because I deleted the file it was originally in but other tests rely on it.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -33,103 +35,360 @@ import (
...
+	type AS server_type_id,
+	upd_pending
+`
 
-func (s *TOServer) Sanitize() {
-	if s.IP6Address != nil && *s.IP6Address == "" {
-		s.IP6Address = nil
-	}
-}
+const insertInterfaceQuery = `
```
shamrickus, line 226: Looks unused?

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -33,103 +35,360 @@ import (
...
+	$4,
+	$5
+)
+`
 
-func (s *TOServer) Validate() error {
-	s.Sanitize()
-	version := s.APIInfo().Version
-	noSpaces := validation.NewStringRule(tovalidate.NoSpaces, "cannot contain spaces")
+const insertIPQuery = `
```
shamrickus, line 242: Looks unused?

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -139,94 +398,224 @@ func (s *TOServer) Validate() error {
...
+	if *m < 1280 {
+		return errors.New("must be at least 1280")
 	}
+	return nil
+}
 
-	var hostName string
-	if s.HostName != nil {
-		hostName = *s.HostName
+func validateGateway(g interface{}) error {
```
shamrickus, line 447: Looks unused?

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -139,94 +398,224 @@ func (s *TOServer) Validate() error {
...
+		api.WriteRespWithSummary(w, r, servers, unfiltered)
+		return
+	}
+
+	if version.Major <= 1 {
+		legacyServers := make([]tc.ServerNullableV11, 0, len(servers))
+		for _, server := range servers {
+			legacyServer, err := server.ToServerV2()
+			if err != nil {
+				api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("Failed to convert servers to legacy format: %v", err))
```
shamrickus, line 555: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -139,94 +398,224 @@ func (s *TOServer) Validate() error {
...
-		// NOTE: it's required to handle minor version cases in a descending >= manner
-		case version.Major >= 2:
-			returnable = append(returnable, server)
-		case version.Major == 1 && version.Minor >= 1:
-			returnable = append(returnable, server.ServerNullableV11)
-		default:
-			return nil, nil, fmt.Errorf("TOServer.Read called with invalid API version: %d.%d", version.Major, version.Minor), http.StatusInternalServerError
+		legacyServer, err := server.ToServerV2()
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("Failed to convert servers to legacy format: %v", err))
```
shamrickus, line 568: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -139,94 +398,224 @@ func (s *TOServer) Validate() error {
...
+		legacyServers = append(legacyServers, legacyServer.ServerNullableV11)
+	}
+	api.WriteAlertsObj(w, r, http.StatusOK, deprecationAlerts, legacyServers)
 }
 
-func getServers(params map[string]string, tx *sqlx.Tx, user *auth.CurrentUser) ([]tc.ServerNullable, error, error, int) {
+func getServers(params map[string]string, tx *sqlx.Tx, user *auth.CurrentUser) ([]tc.ServerNullable, uint64, error, error, int) {
+	var unfiltered uint64
+	if err := tx.QueryRow(unfilteredServersQuery).Scan(&unfiltered); err != nil {
+		return nil, 0, nil, fmt.Errorf("Failed to get servers count: %v", err), http.StatusInternalServerError
```
shamrickus, line 616: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -248,71 +637,103 @@ func getServers(params map[string]string, tx *sqlx.Tx, user *auth.CurrentUser) (
 		// don't allow query on ds outside user's tenant
 		dsID, err := strconv.Atoi(dsIDStr)
 		if err != nil {
-			return nil, errors.New("dsId must be an integer"), nil, http.StatusNotFound
+			return nil, unfiltered, errors.New("dsId must be an integer"), nil, http.StatusNotFound
 		}
 		userErr, sysErr, _ := tenant.CheckID(tx.Tx, user, dsID)
 		if userErr != nil || sysErr != nil {
-			return nil, errors.New("Forbidden"), sysErr, http.StatusForbidden
+			return nil, unfiltered, errors.New("Forbidden"), sysErr, http.StatusForbidden
```
shamrickus, line 644: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,261 +780,478 @@ WHERE s.id IN (` + edgeIDs + `)))
...
-
+func checkTypeChangeSafety(server tc.CommonServerProperties, tx *sqlx.Tx) (error, error, int) {
 	// see if type changed
-	typeID := -1
-	if err := s.APIInfo().Tx.QueryRow("SELECT type FROM server WHERE id = $1", s.ID).Scan(&typeID); err != nil {
+	var typeID int
+	if err := tx.QueryRow("SELECT type FROM server WHERE id = $1", *server.ID).Scan(&typeID); err != nil {
 		if err == sql.ErrNoRows {
-			return errors.New("no server found with this id"), nil, http.StatusNotFound
+			return errors.New("No server found with this ID"), nil, http.StatusNotFound
```
shamrickus, line 788: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,261 +780,478 @@ WHERE s.id IN (` + edgeIDs + `)))
...
-	if typeID != *s.TypeID {
+	if typeID != *server.TypeID {
 		dsIDs := []int64{}
-		if err := s.APIInfo().Tx.QueryRowx("SELECT ARRAY(SELECT deliveryservice FROM deliveryservice_server WHERE server = $1)", s.ID).Scan(pq.Array(&dsIDs)); err != nil && err != sql.ErrNoRows {
+		if err := tx.QueryRowx("SELECT ARRAY(SELECT deliveryservice FROM deliveryservice_server WHERE server = $1)", server.ID).Scan(pq.Array(&dsIDs)); err != nil && err != sql.ErrNoRows {
 			return nil, fmt.Errorf("getting server assigned delivery services: %v", err), http.StatusInternalServerError
 		}
 		if len(dsIDs) != 0 {
-			return errors.New("server type can not be updated when it is currently assigned to delivery services"), nil, http.StatusConflict
+			return errors.New("Server type can not be updated when it is currently assigned to Delivery Services"), nil, http.StatusConflict
```
shamrickus, line 800: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,261 +780,478 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+
+		serviceInterface, err := validateV3(newServer, tx)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusBadRequest, err, nil)
+			return
+		}
+
+		server, err = newServer.ToServerV2()
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("Converting v3 server to v2 for update: %v", err))
```
shamrickus, line 906: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,261 +780,478 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+
+		err := validateV2(&server, tx)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusBadRequest, err, nil)
+			return
+		}
+
+		interfaces, err = server.LegacyInterfaceDetails.ToInterfaces(*server.IPIsService, *server.IP6IsService)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("Converting server legacy interfaces to interface array: %v", err))
```
shamrickus, line 925: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,261 +780,478 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+
+		err := validateV1(legacyServer, tx)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusBadRequest, err, nil)
+			return
+		}
+
+		interfaces, err = legacyServer.LegacyInterfaceDetails.ToInterfaces(true, true)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("Converting server legacy interfaces to interface array: %v", err))
```
shamrickus, line 943: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,261 +780,478 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+
+	var servers []tc.ServerNullable
+	servers, _, userErr, sysErr, errCode = getServers(map[string]string{"id": inf.Params["id"]}, inf.Tx, inf.User)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+
+	if len(servers) < 1 {
+		api.HandleErr(w, r, tx, http.StatusNotFound, fmt.Errorf("No server exists by id #%d", id), nil)
```
shamrickus, line 1217: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,261 +780,478 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+		return
+	}
+	if len(servers) > 1 {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("There are somehow two servers with id %d - cannot delete", id))
```
shamrickus, line 1221: Error string starts with capital

traffic_ops/traffic_ops_golang/server/servers.go
ocket8888: Is that a problem?

shamrickus: Not functionally, but it's an [anti-pattern](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) and internally inconsistent within the file.

traffic_ops/client/server.go
```diff
@@ -33,158 +33,144 @@ const (
...
-	resp, remoteAddr, err := to.request(http.MethodGet, API_SERVERS, nil)
-	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
-	if err != nil {
-		return nil, reqInf, err
+// The 'params' parameter can be used to optionally pass URL "query string
+// parameters" in the request.
+// It returns, in order, the requested servers, any Alerts the API returned,
+// the returned summary.count value, a request info object, and any error that
+// occurred.
+func (to *Session) GetServers(params *url.Values) ([]tc.ServerNullable, tc.Alerts, uint64, ReqInf, error) {
```
rawlinp, line 156: We're starting to get pretty heavy in terms of the number of things these client methods return. Should we just return `ServersV3Response` instead of `[]tc.ServerNullable, tc.Alerts, uint64`, and allow clients to get whatever fields they're interested in from that?

ocket8888: That's fine with me, as long as that's consistently what we wanna do with these.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,267 +737,482 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+			ipArgs = append(ipArgs, ip.Address, ip.Gateway, iface.Name, id, ip.ServiceAddress)
 		}
 	}
 
-	current := TOServer{}
-	err := s.ReqInfo.Tx.QueryRowx(selectV20UpdatesQuery()+` WHERE sv.id=$1`, strconv.Itoa(*s.ID)).StructScan(&current)
+	ifaceQry += strings.Join(ifaceQueryParts, ",")
+	log.Debugf("Inserting interfaces for new server, query is: %s", ifaceQry)
+
+	ifaceRows, err := tx.Query(ifaceQry, ifaceArgs...)
```
rawlinp, line 807: This seems like it should be an `Exec` instead if we really only care about the number of rows affected.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,267 +737,482 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+	insertedIfaces := 0
+	for ifaceRows.Next() {
+		insertedIfaces++
+	}
+	log.Debugf("Inserted %d interfaces", insertedIfaces)
+
+	ipQry += strings.Join(ipQueryParts, ",")
+	log.Debugf("Inserting IP addresses for new server, query is: %s", ipQry)
+
+	ipRows, err := tx.Query(ipQry, ipArgs...)
```
rawlinp, line 821: Similar to the previous comment, this seems like it should be an `Exec`. It's also a bit of a red flag to not iterate over all of the returned `Rows` since later queries in the same transaction will fail if there are `Rows` that haven't been `Next()`'d from a previous query in the transaction.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,267 +737,482 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+	if err != nil {
+		return api.ParseDBError(err)
+	}
+	defer ipRows.Close()
+
+	return nil, nil, http.StatusOK
+}
+
+func deleteInterfaces(id int, tx *sql.Tx) (error, error, int) {
+	if err := tx.QueryRow(deleteIPsQuery, id).Scan(); err != nil && err != sql.ErrNoRows {
```
rawlinp, line 831: Similar to the previous comment, this seems like it should be an `Exec` since the `Scan()` does nothing.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,267 +737,482 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+		return api.ParseDBError(err)
+	}
+
+	if err := tx.QueryRow(deleteInterfacesQuery, id).Scan(); err != nil && err != sql.ErrNoRows {
```
rawlinp, line 835: Similar to the previous comment, this seems like it should be an `Exec` since the `Scan()` does nothing.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,267 +737,482 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+	if len(servers) < 1 {
+		api.HandleErr(w, r, tx, http.StatusNotFound, fmt.Errorf("no server exists by id #%d", id), nil)
+		return
+	}
+	if len(servers) > 1 {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("there are somehow two servers with id %d - cannot delete", id))
+		return
+	}
+
+	userErr, sysErr, errCode = deleteInterfaces(id, tx)
```
rawlinp, line 1186: This seems like it should be handled via `ON DELETE CASCADE` on the interface/ip_address tables, but I see that they are actually `ON DELETE RESTRICT`. At this point I don't think it's a big enough deal to require modifying the db migration.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,267 +737,482 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+
+	if err := tx.QueryRow(deleteServerQuery, id).Scan(); err != nil && err != sql.ErrNoRows {
+		userErr, sysErr, errCode = api.ParseDBError(err)
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
```
rawlinp, line 1196: Should we also handle the `sql.ErrNoRows` error here?

docs/source/glossary.rst
```diff
@@ -302,22 +302,22 @@ Glossary
 				GET /foo/bar/fun.html HTTP/1.1
 				Host: www-origin-cache.cdn.com
 
-		#. The :dfn:`reverse proxy` finds out the URL of the true :term:`origin` - in the case of :abbr:`ATS (Apache Traffic Server)` this is done by looking up ``www-origin-cache.cdn.com`` in its remap rules - and finds that it is ``www.origin.com``.
```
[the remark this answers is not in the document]
ocket8888: Yes, it was just making it very hard to sift through all of the warnings to find out what was wrong with the two files I was changing.

infrastructure/cdn-in-a-box/enroller/Dockerfile
```diff
@@ -20,7 +20,7 @@ FROM golang:1.14.2 AS enroller-builder
 # enroller source and dependencies
 COPY ./lib/ /go/src/github.com/apache/trafficcontrol/lib/
 COPY ./vendor/ /go/src/github.com/apache/trafficcontrol/vendor/
-COPY ./traffic_ops/client/ /go/src/github.com/apache/trafficcontrol/traffic_ops/v2-client/
+COPY ./traffic_ops/v2-client/ /go/src/github.com/apache/trafficcontrol/traffic_ops/v2-client/
```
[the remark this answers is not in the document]
ocket8888: I'm not opposed to rebasing if another PR does it, but I run API tests using CiaB so CiaB must run in order for me to test any PR I write.

traffic_ops/client/server.go
```diff
@@ -33,158 +33,142 @@ const (
...
-func (to *Session) GetServers() ([]tc.Server, ReqInf, error) {
-	resp, remoteAddr, err := to.request(http.MethodGet, API_SERVERS, nil)
-	reqInf := ReqInf{CacheHitStatus: CacheHitStatusMiss, RemoteAddr: remoteAddr}
-	if err != nil {
-		return nil, reqInf, err
+// The 'params' parameter can be used to optionally pass URL "query string
+// parameters" in the request.
+// It returns, in order, the API response that Traffic Ops returned, a request
+// info object, and any error that occurred.
+func (to *Session) GetServers(params *url.Values) (tc.ServersV3Response, ReqInf, error) {
```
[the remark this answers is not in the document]
ocket8888: Either way I'm returning a new structure, so I'm not sure why it matters what's "nullable" or not.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -139,94 +366,213 @@ func (s *TOServer) Validate() error {
...
+
+	if version.Major >= 3 {
+		api.WriteRespWithSummary(w, r, servers, unfiltered)
+		return
 	}
 
+	if version.Major <= 1 {
+		legacyServers := make([]tc.ServerNullableV11, 0, len(servers))
+		for _, server := range servers {
+			legacyServer, err := server.ToServerV2()
```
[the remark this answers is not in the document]
ocket8888: There's no valid reason why a _valid_ V3 server  can't be converted to a V2 or V1 server. Why?

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -248,71 +594,103 @@ func getServers(params map[string]string, tx *sqlx.Tx, user *auth.CurrentUser) (
...
 		userErr, sysErr, _ := tenant.CheckID(tx.Tx, user, dsID)
 		if userErr != nil || sysErr != nil {
-			return nil, errors.New("Forbidden"), sysErr, http.StatusForbidden
+			return nil, unfiltered, errors.New("forbidden"), sysErr, http.StatusForbidden
 		}
 		// only if dsId is part of params: add join on deliveryservice_server table
-		queryAddition = `
-FULL OUTER JOIN deliveryservice_server dss ON dss.server = s.id
-`
+		queryAddition = "\nFULL OUTER JOIN deliveryservice_server dss ON dss.server = s.id\n"
```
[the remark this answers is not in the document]
ocket8888: I don't much care about multiline vs `\n`, I just don't like not being able to fold code. I can put it back to using <kbd>`</kbd> strings but I'd like to add extra indentation in that case so that it clearly, visually falls within its containing block.

traffic_ops/traffic_ops_golang/server/servers.go
ocket8888: TBH I don't even really care about the rows affected, I just ran into problems with not exhausting the rows so I counted them.

docs/source/glossary.rst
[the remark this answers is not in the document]
rawlinp: Yes, it also makes it very hard to sift through all the "non-changes" to determine what documentation changes you are actually trying to make in this PR.

infrastructure/cdn-in-a-box/enroller/Dockerfile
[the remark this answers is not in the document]
rawlinp: Your situation is not unique, I'm just saying there has been at least one other person who hit this issue, and it would be better to fix the bug for everyone rather than tie it to a massive new feature. Separate bugfix PRs are also much easier to backport if necessary.

traffic_ops/client/server.go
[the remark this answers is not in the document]
rawlinp: Reviewing separate changes separately is much easier. Even though you're returning a "new" struct, most of these changes are due to most of the struct fields going from non-nullable to nullable, and that would've been easier to review as a smaller, separate PR.

traffic_ops/traffic_ops_golang/server/servers.go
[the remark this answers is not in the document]
rawlinp: Just wanted to make sure, otherwise someone could submit an invalid v3 server that would cause the v1/v2 APIs to break and become unavailable.

traffic_ops/traffic_ops_golang/server/servers.go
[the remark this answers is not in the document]
rawlinp: That sounds fine to me.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -139,94 +366,213 @@ func (s *TOServer) Validate() error {
...
+		errs = append(errs, tc.EmptyAddressCannotBeAServiceAddressError)
 	}
+	return util.JoinErrs(errs)
+}
 
-	var hostName string
-	if s.HostName != nil {
-		hostName = *s.HostName
+func validateMTU(mtu interface{}) error {
+	m := mtu.(*uint64)
```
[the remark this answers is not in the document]
ocket8888: Sure. It shouldn't be a problem because this is a validation function only called on instances of `*uint64`, but there's no reason not to handle that.

TBH I was sort of counting on the panic because if that's not getting called on a `*uint64` there's some kind of messed up server-side logic and that'd return a 500 error.

traffic_ops/traffic_ops_golang/server/servers.go
ocket8888: Hm. I wanna try to figure out what I was thinking there, but I think my answer is "yes".

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -139,94 +366,213 @@ func (s *TOServer) Validate() error {
...
+			return
+		}
+		legacyServers = append(legacyServers, legacyServer.ServerNullableV11)
+	}
+	api.WriteAlertsObj(w, r, http.StatusOK, deprecationAlerts, legacyServers)
 }
 
-func getServers(params map[string]string, tx *sqlx.Tx, user *auth.CurrentUser) ([]tc.ServerNullable, error, error, int) {
+func getServers(params map[string]string, tx *sqlx.Tx, user *auth.CurrentUser) ([]tc.ServerNullable, uint64, error, error, int) {
+	var unfiltered uint64
```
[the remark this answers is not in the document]
ocket8888: Done. Also fixed it not considering non-pagination query params.

traffic_ops/traffic_ops_golang/server/servers.go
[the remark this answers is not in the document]
ocket8888: Yeah, that should be impossible, barring direct database manipulation.

traffic_ops/traffic_ops_golang/ats/atsserver/ipallowdotconfig.go
```diff
@@ -1,148 +0,0 @@
-package atsserver
```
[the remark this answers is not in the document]
ocket8888: Yeah, that was the plan, but you asked me to delete the routes?

traffic_ops/traffic_ops_golang/server/servers.go
ocket8888: yeah, idk. Fixed.

traffic_ops/traffic_ops_golang/server/servers.go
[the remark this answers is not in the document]
ocket8888: I guess also I should mention that even through direct database manipulation you can't _break_ the V1/V2 APIs, just make them not return IPAddress values even if the server has configured addresses - though they'd be non-service addresses so that's still technically correct.

traffic_ops/traffic_ops_golang/ats/atsserver/ipallowdotconfig.go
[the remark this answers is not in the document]
rawlinp: Yeah, I wasn't really clear about wanting to do that in a separate PR. I'm just trying to throw out ideas for breaking up PRs in the future, since this PR is doing way too many things at this point.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -359,267 +757,475 @@ WHERE s.id IN (` + edgeIDs + `)))
...
+		return
+	}
+
+	userErr, sysErr, errCode = deleteInterfaces(id, tx)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+
+	if err := tx.QueryRow(deleteServerQuery, id).Scan(); err != nil {
```
rawlinp, line 1205: I think this might also need to be an `Exec`, and we might need to handle the "no rows" error separately instead of lumping it in with the other errors. What did the API do in this case previously? 404?

ocket8888: It will 404 before that point, because the first thing it does is fetch it. but also I thought `api.ParseDBError` would handle that distinction?

ocket8888: For whatever reason changing this to explicitly handle ErrNoRows (and use Exec) seems to have fixed the issue.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1202,8 +1202,15 @@ func Delete(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := tx.QueryRow(deleteServerQuery, id).Scan(); err != nil {
-		userErr, sysErr, errCode = api.ParseDBError(err)
+	if _, err := tx.Exec(deleteServerQuery, id); err != nil {
+		if err == sql.ErrNoRows {
```
rawlinp, line 1206:
```
// ErrNoRows is returned by Scan when QueryRow doesn't return a
// row. In such a case, QueryRow returns a placeholder *Row value that
// defers this error until a Scan.
var ErrNoRows = errors.New("sql: no rows in result set")
```
Digging into this a bit more, it seems like `Exec` won't actually return that error if no rows were affected, so it seems like the real fix is to check the result similar this example (though not exactly since it should probably differentiate between "too many rows affected" and "no rows affected"): 
```
result, err := origin.ReqInfo.Tx.NamedExec(deleteQuery(), origin)
	if err != nil {
		return nil, errors.New("origin delete: query: " + err.Error()), http.StatusInternalServerError
	}
	rowsAffected, err := result.RowsAffected()
	if err != nil {
		return nil, errors.New("origin delete: getting rows affected: " + err.Error()), http.StatusInternalServerError
	}
	if rowsAffected != 1 {
		return nil, errors.New("origin delete: multiple rows affected"), http.StatusInternalServerError
	}
```

ocket8888: done

ocket8888:
Yes, unfortunately the only other option would be to query all interface data and re-construct by hand. The `pq` driver does not support nested array values, so we needed to mess with it a bit to get it to spit out the data we need.

rawlinp:
Then I think we might need to do that (query the data then reconstruct in Go).

rawlinp:
> For me, the xmppId fields from https://trafficportal.infra.ciab.test/api/2.0/servers are always empty

That seems to be very problematic. I believe `xmppId` is supposed to be set to the initial hostname when the server is created, as kind of a read-only field, because that is what TR uses for consistent hashing servers. If that functionality is lost, that could seriously break the CDN.

ocket8888:
To be clear, this change does break CiaB. Snapshots won't work properly until #4709 is merged, and caches will never be able to successfully apply queued updates until #4671 is merged.

But the thing about the xmppIds is weird, nothing about them should have changed.

ocket8888:
Oh, I see. The part that sets the XMPPID is receiving a pass-by-value argument of the server, so basically I destroyed its ability to actually affect the structure. Fixed.

ocket8888:
Fixed

ocket8888:
In the future I'll definitely be using feature branches instead. I was just under the impression that was frowned upon prior to starting this effort.

rawlinp:
IMO feature branches are not the solution either. This PR could've done without a bunch of extraneous changes yet included the changes that kept everything functional. For instance: this commit is minuscule in comparison to the size of this PR, yet probably could've been included while other extraneous changes could've been excluded: https://github.com/apache/trafficcontrol/pull/4709/commits/a83545949e5a2af02a902a3d0a2716853accfbb6.

ocket8888:
I think the problem was that I misunderstood what was meant by "feature branch" initially. I think you mean "feature branch" to say "a branch of apache/trafficcontrol that contains the entire body of work necessary to implement servers having multiple interfaces in every facet of every component" whereas I mean "A big branch of changes across multiple components that provides all of the changes necessary to implement a feature - which may or may not mean everything it takes to implement servers having multiple interfaces."

For example, you did a feature branch with topologies. You don't call it a feature branch, but I would, because it was a branch that encapsulated a feature - the feature of including and being able to manipulate ATC topologies. Just because it didn't totally fulfil the requirements of a blueprint doesn't make it not a feature branch, IMO.

To be clear I'm not against using a "feature branch" by either of those definitions. I just thought they were frowned upon equally, so I tried to do things without them. Imagine my surprise when a Topologies feature branch sprung up.

rawlinp:
There was no feature branch for Topologies -- all the work we've done on Topologies so far has been made in atomic PRs to master.

Our definition of feature branches is the same. I'm not specifying "within this repo" either. Everyone creates a "feature branch" in their own forks in order to create a PR. Every PR to master essentially _is_ a feature branch. At the end of the day, I don't think we should ever be opening massive PRs that take a tremendous amount of time in one sitting to review/test/merge. Most new features can always be broken down into smaller, atomic PRs that don't break things when merged on their own. Throughout reviewing this PR, I've tried to describe to you _how_ that could've been done from the get-go, so that we can try to break things down into smaller, atomic PRs in the future. Creating a feature branch goes entirely against that idea, because it means reviewing and merging one massive PR into master that completely implements an entire new feature.

ocket8888:
> _"There was no feature branch for Topologies..."_

> _"Our definition of feature branches is the same."_

Those statements are contradictory (so are _"There was no feature branch for Topologies..."_ and _"Every PR to master essentially is a feature branch."_ lol). What you did for Topologies was what I would call a "feature branch", because it reached across multiple components and/or API endpoints. Its scope was broader than an atomic unit of work done by one person to one aspect of one component. Your definition seems to be more closely related to the "size" of the PR than its "scope".

The only way the `/servers` endpoint could be changed to not break snapshots would be if that was in the same PR (which was a relatively small change, as it turned out), and to avoid breaking Traffic Portal it would need to be in the same PR as well (although in my defence that wouldn't have been the case if TP stayed on the stable 2.0 API, which was a change I deliberately hoped to avoid making for exactly this reason when adding the 3.0 API). Which all sounds fine to me, except that I'd also call that a feature branch.

I understand how to do feature branches - obviously, because I already said I'd prefer to do them going forward. You don't need to convince me, I'm on board. I just needed to know that was an option; this was just a misunderstanding.

rawlinp:
Ok, I think I am beginning to understand your definition of a feature branch. Earlier, you said this:
> A big branch of changes across multiple components that provides all of the changes necessary to implement a feature

Your definition lies in the fact that one PR includes changes to multiple components at once? That is not necessarily true -- you can have a feature branch that only affects a single component.

I am generally in favor of PRs that only affect a single component at a time (assuming those changes can live on their own without breaking things), but that has nothing to do with feature branches. What I mean by a "feature branch" would be this:

1. You create a new branch in this repo called `dual-homing`. This is the feature branch.
2. You implement certain parts of the dual-homing feature and open PRs to the `dual-homing` branch. The size and scope of the PR, how many components they affect, is irrelevant.
3. Three months pass, and multiple PRs have been opened and merged into the `dual-homing` branch. The "dual-homing" feature is now basically done. Meanwhile, `master` has continued to move forward with other changes.
4. Now the `dual-homing` branch is ready to merge into `master` assuming it's been rebased regularly. The diff is massive. Ideally, someone would review the entirety of the diff before merging `dual-homing` into `master`. If the PRs into `dual-homing` were thoroughly reviewed, and `dual-homing` was frequently rebased onto `master`, maybe this final review can be quick.
5. `dual-homing` is merged into `master`. Chaos ensues. Everyone that had open PRs into master now all have merge conflicts, and they're probably not easily resolved since there are three months worth of changes to sift through.

That is essentially what I mean by "feature branches", and I would really like to avoid them. What you're describing as a "feature branch" is just a PR with more than one "atom" of a feature. I think it's fine to have multiple "atoms" in a single PR assuming they can still live on their own without breaking things and aren't so numerous that the PR is a major burden to review due to its size and scope. For me, I'd start to look at breaking up a PR into atoms once it breaks the 500 LOC range. To use TO/TP as an example, if you're adding a new Delivery Service field, you could implement the new field in TO as one PR, then add the new field to TP in a second PR. The TO change can live on its own without breaking things -- yet it can't really be broken down further since it's a simple field addition. That would be one "atom". The TP change would be the second "atom". Both of them combined are probably still a relatively small PR, so you could lump both atoms together in the same PR.

ocket8888:
Yes, that's all exactly correct and highlights the very misunderstanding I only realized after a few PRs.

PR #4701, opened 2020-05-19
Updating the cdn of a server which is already to linked to a DS should be rejected

### LINKED ISSUE #4116: Delivery Services can be assigned servers from a different CDN
## I'm submitting a ...
- bug report

- Traffic Ops

## Current behavior:
Similar to #3756 , the TO API allows servers to be assigned to Delivery Services outside of their respective CDNs via the endpoints:

- `deliveryservices/{{xml_id}}/servers`
- `deliveryserviceserver`

(and also `/servers/{{ID}}/deliveryservices` at the time of this writing, but that's fixed by #3761)

## Expected / new behavior:
Servers outside of a Delivery Service's CDN should not be assignable to said Delivery Service.

1. Create a (Edge-tier cache or Origin) Server in CDN A
1. Create a Delivery Service in CDN B
1. Try to assign the Server to the Delivery Service
1. Notice how that worked even though it shouldn't.

srijeet0406:

- [x] This PR fixes #4116 

- Traffic Ops

Start TO and TP locally. Create atleast two CDNs and create one DS in each of these CDNs.
Now, create and attach a server to the DS of the first CDN.
After this, try to edit the newly created server to update its CDN to be the second one. The request should fail saying that you cannot update the CDN of a server, which is already linked to a DS in a different CDN.
 
master

- [x] This PR includes tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed ORT config generation not using the coalesce_number_v6 Parameter.
 - Removed audit logging from the `POST /api/x/serverchecks` Traffic Ops API endpoint in order to reduce audit log spam
 - Fixed audit logging from the `/jobs` APIs to bring them back to the same level of information provided by TO-Perl
+- Fixed update procedure of servers, so that if a server is linked to one or more delivery services, you cannot change its "cdn".
 
 ### Changed
 - Changed some Traffic Ops Go Client methods to use `DeliveryServiceNullable` inputs and outputs.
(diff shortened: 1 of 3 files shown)


mitchell852:
Maybe major bug fixes warrants a change log entry? Anyone else have an opinion on that?

rob05c:
I'd vote even seemingly minor bugs, if it existed in a previous version, we should note it in the changelog.

Because what's minor to one user might be major to another.

If the bug was introduced since the last release, no changelog needed, of course.

rawlinp:
+1 on changelog entries for all bug fixes no matter the severity (as long as the bug exists in a previous release).

srijeet0406:
Added a changelog entry for this ^^

srijeet0406:
Also closes #5018

PR #4704, opened 2020-05-20
Fix OptionsDeleter audit log keys

rawlinp:
Fixed the audit logging of `OptionsDeleter` to set the keys on the to-be-deleted object so that they are included in the audit log entry when deleted.

- Traffic Ops

Run the unit tests and TO API tests, verify they pass. Observe the deletion audit log entries, verify they contain the key/name of the deleted resource (specifically Topologies and Regions, which both use the `OptionsDeleter` interface).

- master
- 4.1.0-RC0

- [x] This PR includes tests
- [x] unreleased (as of yet), so no changelog necessary

### DIFF
--- a/traffic_ops/traffic_ops_golang/api/shared_handlers.go
+++ b/traffic_ops/traffic_ops_golang/api/shared_handlers.go
@@ -315,25 +315,23 @@ func DeleteHandler(deleter Deleter) http.HandlerFunc {
 			HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
 			return
 		}
-		if !deleteKeyOptionExists {
-			keyFields := obj.GetKeyFieldsInfo() // expecting a slice of the key fields info which is a struct with the field name and a function to convert a string into a interface{} of the right type. in most that will be [{Field:"id",Func: func(s string)(interface{},error){return strconv.Atoi(s)}}]
-			keys := make(map[string]interface{})
-			for _, kf := range keyFields {
-				paramKey := inf.Params[kf.Field]
-				if paramKey == "" {
-					HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("missing key: "+kf.Field), nil)
-					return
-				}
-
-				paramValue, err := kf.Func(paramKey)
-				if err != nil {
-					HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("failed to parse key: "+kf.Field), nil)
-					return
-				}
-				keys[kf.Field] = paramValue
+		keyFields := obj.GetKeyFieldsInfo() // expecting a slice of the key fields info which is a struct with the field name and a function to convert a string into a interface{} of the right type. in most that will be [{Field:"id",Func: func(s string)(interface{},error){return strconv.Atoi(s)}}]
+		keys := make(map[string]interface{})
+		for _, kf := range keyFields {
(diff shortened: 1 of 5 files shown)


traffic_ops/traffic_ops_golang/api/shared_handlers.go
```diff
@@ -315,25 +315,23 @@ func DeleteHandler(deleter Deleter) http.HandlerFunc {
 			HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
 			return
 		}
-		if !deleteKeyOptionExists {
```
rawlinp, line 318: The diff is kind of hard to read here, all I did was remove the `if !deleteKeyOptionExists` so that keys are still set with either Deleter or OptionsDeleter.

mitchell852:
I created and deleted a region and don't see the name in the audit log

image

rawlinp:
Oh, maybe it wasn't clear, it would be either the ID or the name, depending on what the PK of the resource is. So regions uses IDs for deletion whereas Topologies uses the name, but I do agree that when deleting things we should probably always include the name (if it has one). I imagine that is a problem we have with a lot of our APIs currently. But for this PR, it fixes the regions audit log entry to include at least the ID and the topologies log entry to include the name. I think we should address the ID vs name thing separately.

rawlinp:
retest this please

mitchell852:
> Oh, maybe it wasn't clear, it would be either the ID or the name, depending on what the PK of the resource is. So regions uses IDs for deletion whereas Topologies uses the name, but I do agree that when deleting things we should probably always include the name (if it has one). I imagine that is a problem we have with a lot of our APIs currently. But for this PR, it fixes the regions audit log entry to include at least the ID and the topologies log entry to include the name. I think we should address the ID vs name thing separately.

makes sense.

zrhoffman:
Was this reviewed?

traffic_ops/traffic_ops_golang/api/shared_handlers.go
zrhoffman, line 318: We can't remove this condition here without re-adding the logic elsewhere. Normal delete requires all all keys exist, so [`Identifier.GetKeyFieldsInfo()`](https://github.com/apache/trafficcontrol/blob/09bfc37828/traffic_ops/traffic_ops_golang/api/shared_interfaces.go#L50-L51) generally only contains a single key.

[`HasDeleteKeyOptions.DeleteKeyOptions()`](https://github.com/apache/trafficcontrol/blob/09bfc37828/traffic_ops/traffic_ops_golang/api/shared_interfaces.go#L106) is a separate array containing keys that *can* identify a resource, but [only requires](https://github.com/apache/trafficcontrol/blob/09bfc37828/traffic_ops/traffic_ops_golang/api/shared_handlers.go#L122-L129) that one such key exists.

`if !deleteKeyOptionExists {}` is used to keep the handler from exiting if `Identifier.GetKeyFieldsInfo()` keys do not all exist but at least 1 `HasDeleteKeyOptions.DeleteKeyOptions()` key does.

[`DELETE regions?name={{name}}`](https://traffic-control-cdn.readthedocs.io/en/latest/api/v3/regions.html#delete) currently fails with "missing key: id".

zrhoffman: Opened a PR to add an API test for the failing route: #4707

PR #4707, opened 2020-05-20
Add and satisfy OptionsDelete API test for DELETE /regions?name={{name}}

zrhoffman:

- [x] This PR complements #4357's fix for #3819 by adding an API test.

- Traffic Ops

Run the API tests, make sure they pass
```shell script
docker-compose -f docker-compose.yml -f docker-compose.traffic-ops-test.yml up -d && \
  docker-compose -f docker-compose.yml -f docker-compose.traffic-ops-test.yml logs -f integration
```

- master (09bfc37828)

- [x] This PR includes tests
- [x] Documentation already exists, including docs in the PR is unnecessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - Traffic Ops: Added an API 3.0 endpoint, /api/3.0/topologies, to create, read, update and delete flexible topologies.
     - Traffic Ops: Added support for `topology` query parameter to `GET /api/3.0/cachegroups` to return all cachegroups used in the given topology.
     - Traffic Portal: Added the ability to create, read, update and delete flexible topologies.
+- Added an API test for `DELETE /api/3.0/regions?name={{name}}`
 
 ### Fixed
 - Fixed the `GET /api/x/jobs` and `GET /api/x/jobs/:id` Traffic Ops API routes to allow falling back to Perl via the routing blacklist
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - Traffic Ops: Added an API 3.0 endpoint, /api/3.0/topologies, to create, read, update and delete flexible topologies.
     - Traffic Ops: Added support for `topology` query parameter to `GET /api/3.0/cachegroups` to return all cachegroups used in the given topology.
     - Traffic Portal: Added the ability to create, read, update and delete flexible topologies.
-- Added an API test for `DELETE /api/3.0/regions?name={{name}}`
 
 ### Fixed
 - Fixed the `GET /api/x/jobs` and `GET /api/x/jobs/:id` Traffic Ops API routes to allow falling back to Perl via the routing blacklist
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - Traffic Ops: Added an API 3.0 endpoint, /api/3.0/topologies, to create, read, update and delete flexible topologies.
     - Traffic Ops: Added support for `topology` query parameter to `GET /api/3.0/cachegroups` to return all cachegroups used in the given topology.
     - Traffic Portal: Added the ability to create, read, update and delete flexible topologies.
+- Added an API test for `DELETE /api/3.0/regions?name={{name}}`
```
shamrickus, line 14: Should we be adding test changes to the changelog?

zrhoffman: This API test currently fails due to a regression in #4704. This PR will is intended to include a fix, and bugfixes should be included in the changelog. When the fix is pushed, I'll update the existing changelog entry instead of adding a new one.

rawlinp: I don't think so. The audience for the changelog probably doesn't need to know about every single test added ¯\_(ツ)_/¯

rawlinp: Whoops, your comment was added after I started writing mine @zrhoffman. But since this bug is unreleased (as of yet), I don't think the bugfix needs to be in the changelog either.

zrhoffman: Okay, changelog entry gone in da68d318a.

zrhoffman:
This is ready for review.

zrhoffman:
The audit log for `DELETE /regions?name={{name}}` is fixed in d77345814a, and audit log assertions for that route are added to the API tests in a8f72f6d6f.

traffic_ops/traffic_ops_golang/api/shared_handlers.go
```diff
@@ -390,30 +416,34 @@ func DeprecatedDeleteHandler(deleter Deleter, alternative *string) http.HandlerF
...
 		}
-		if !deleteKeyOptionExists {
-			keyFields := obj.GetKeyFieldsInfo() // expecting a slice of the key fields info which is a struct with the field name and a function to convert a string into a interface{} of the right type. in most that will be [{Field:"id",Func: func(s string)(interface{},error){return strconv.Atoi(s)}}]
-			keys := make(map[string]interface{})
-			for _, kf := range keyFields {
-				paramKey := inf.Params[kf.Field]
-				if paramKey == "" {
-					HandleDeprecatedErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("missing key: "+kf.Field), nil, alternative)
-					return
+		keyFields := obj.GetKeyFieldsInfo() // expecting a slice of the key fields info which is a struct with the field name and a function to convert a string into a interface{} of the right type. in most that will be [{Field:"id",Func: func(s string)(interface{},error){return strconv.Atoi(s)}}]
```
rawlinp, line 424: It looks like the DeprecatedDeleteHandler function didn't get all the same treatment as the DeleteHandler. Before, they were basically identical except for `HandleErr` vs `HandleDeprecatedErr` (before I messed).

We should keep these in parity, right? Until we reduce the duplication here.

zrhoffman: Right you are. Parity kept in 0c60881ce.

PR #4708, opened 2020-05-20
Add Blueprint for Cache Config Snapshots

rob05c:

Adds a Blueprint for Cache Config Snapshots

No tests, no documentation, no changelog, it's a blueprint, doesn't affect production code.

None, blueprint.

Read blueprint.

Not a bug fix.

## Additional Information


blueprints/traffic-ops-cache-config-endpoint.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Proposed Change
+
+Traffic Ops will add a “Cache Config Snapshot” feature.
+
+When the cache config is Snapshotted, all data will be serialized and stored, just like the CRConfig.
+
+Snapshot history will be stored. New snapshots will insert with a timestamp, retaining previous versions so they can easily be restored if necessary. Snapshots should be automatically pruned with a configurable number (which maybe be infinity or zero), and/or a configurable amount of time.
+
+The snapshot data will be cached in memory, with a configurable cache time (suggested default: 5s).
```
ocket8888, line 35: TO memory or ORT memory?

rob05c: TO Memory. I changed it to clarify.

TO Memory reduces DB load, by making it so that when the Snapshot changes and 1000 caches request it at once, TO only has to make 1 DB request.

ORT should also cache, once we have TO doing IMS. Which reduces load to each cache when the Snapshot hasn't change; but doesn't obviate the benefit of TO caching for when it has.

blueprints/traffic-ops-cache-config-endpoint.md
```diff
@@ -0,0 +1,155 @@
...
+
+The new API endpoint, and ATS config changes, will have TO API tests and unit tests. See Traffic Ops Impact.
+
+### Performance Impact
+
+No significant performance impact expected. Load impact on Traffic Ops should be reduced.
+
+### Security Impact
+
+n/a
```
rawlinp, line 120: We may want to note here that secrets required by ORT (sslkeys, URI signing keys, etc) would now be stored in the cache-config-snapshot JSON blob in TODB, rather than existing only in Riak, so this new endpoint should require admin privileges.

PR #4713, opened 2020-05-21
Add csv as an output format option for astats via header

ezelkow1:
- [x] This PR is related to the TM 2.0 issue but does not complete it

This adds the ability to output CSV formatted stats via the `Accept` header.  If set to `text/csv` then the output will be in csv format.  Any other values will output in the old `text/json` format.  The `Content-Type` field has also been fixed since previously it was reported as `application/javascript`

- Traffic Monitor (will need future changes to support this)
- ATS

Build astats plugin and install on a cache. Hit the stats endpoint, verify the the old format is output with the new `Content-Type: text/json`.  Add a header of `Accept: text/csv` to a new request and verify that the content comes back in csv format and with `Content-Type: text/csv`.

### DIFF
--- a/traffic_server/plugins/astats_over_http/astats_over_http.c
+++ b/traffic_server/plugins/astats_over_http/astats_over_http.c
@@ -40,6 +40,11 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 
+typedef enum {
+	JSON_OUTPUT,
+	CSV_OUTPUT
+} output_format;
+
 typedef struct {
 	unsigned int recordTypes;
 	char *stats_path;
@@ -82,6 +87,7 @@ typedef struct stats_state_t {
 	char *interfaceName;
 	char *query;
 	unsigned int recordTypes;
+	output_format output;
 } stats_state;
 
 int configReloadRequests = 0;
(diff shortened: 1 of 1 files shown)


traffic_server/plugins/astats_over_http/astats_over_http.c
```diff
@@ -326,7 +383,40 @@ static void appendSystemState(stats_state *my_state) {
...
+			APPEND_STAT_CSV("proc.net.dev", "%s", str);
+		}
+	}
+
+	str = getFile("/proc/loadavg", buffer, sizeof(buffer));
+	if (str) {
+		end = strstr(str, "\n");
+		if (end)
+			*end = 0;
+		APPEND_STAT_CSV("proc.loadavg", "%s", str);
```
jhg03a, line 419: This looks like a lot of copypasta just to change the output functions.  Not a huge issue though.

traffic_server/plugins/astats_over_http/astats_over_http.c
```diff
@@ -193,14 +199,25 @@ stats_add_data_to_resp_buffer(const char *s, stats_state *my_state) {
...
+		switch (my_state->output) {
+			case JSON_OUTPUT:
+				my_state->output_bytes = stats_add_data_to_resp_buffer(RESP_HEADER_JSON, my_state);
+				break;
+			case CSV_OUTPUT:
+				my_state->output_bytes = stats_add_data_to_resp_buffer(RESP_HEADER_CSV, my_state);
+				break;
+			default:
+				TSError("stats_process_read: Unknown output format\n");
+				break;
```
jhg03a, line 219: This might be an issue since it's a change to the default behavior.  Just in case we've got malformed client requests and havn't noticed since it just worked.

traffic_server/plugins/astats_over_http/astats_over_http.c
ezelkow1: Agreed, just trying to leave as much of the original functionality alone as possible.  We can look at refactoring after this if we want

traffic_server/plugins/astats_over_http/astats_over_http.c
ezelkow1: It also shouldnt even change based on client requests since everywhere else it will default to json output, on startup, on request with anything other than `text/csv`, so if we manage to get here and its some unknown type some memory probably got stomped on or something very very bad happened

traffic_server/plugins/astats_over_http/astats_over_http.c
rob05c: +1, to both

traffic_server/plugins/astats_over_http/astats_over_http.c
```diff
@@ -343,25 +433,54 @@ static void json_out_stats(stats_state *my_state) {
...
 		APPEND("\"something\": \"here\"");
 		APPEND("\n  }");
 	}
 
 	APPEND("\n}\n");
 }
 
+static void csv_out_stats(stats_state *my_state) {
+	const char *version;
+	TSDebug(PLUGIN_TAG, "recordTypes: '0x%x'", my_state->recordTypes);
```
zrhoffman, line 451: This can be `'%#x'`

ezelkow1: It can, but leaving it as 0x%x preserves the existing syntax used in the rest of the file

zrhoffman: Fair enough

PR #4721, opened 2020-05-26
Queries with invalid type parameter should return 400s, not 500s

### LINKED ISSUE #4703: Internal Server Error from /cachegroups parameter
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When perform a GET with character type parameter query, the API will return Internal Server Error
`/api/3.0/cachegroups?type=WRONG`

## Expected / new behavior:

Should return 400 Bad Request

srijeet0406:

- [x] This PR fixes #4703  

- CDN in a Box
- Documentation
- Traffic Ops

Setup TO locally and make a query like this `GET /api/3.0/cachegroups?type=WRONG`
This should return a 400, with the error details in it as the response.

master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -357,6 +357,3 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 [4.0.0]: https://github.com/apache/trafficcontrol/compare/RELEASE-3.0.0...RELEASE-4.0.0
 [3.0.0]: https://github.com/apache/trafficcontrol/compare/RELEASE-2.2.0...RELEASE-3.0.0
 [2.2.0]: https://github.com/apache/trafficcontrol/compare/RELEASE-2.1.0...RELEASE-2.2.0
-
-
-[]: https://github.com/apache/trafficcontrol/issues/4703
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Removed audit logging from the `POST /api/x/serverchecks` Traffic Ops API endpoint in order to reduce audit log spam
 - Fixed audit logging from the `/jobs` APIs to bring them back to the same level of information provided by TO-Perl
 - Fixed update procedure of servers, so that if a server is linked to one or more delivery services, you cannot change its "cdn".
+- Fixed `cachegroups` READ endpoint, so that if a request is made with the `type` specified as a non integer value, you get back a `400` with error details, instead of a `500`.
 
 ### Changed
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Removed audit logging from the `POST /api/x/serverchecks` Traffic Ops API endpoint in order to reduce audit log spam
 - Fixed audit logging from the `/jobs` APIs to bring them back to the same level of information provided by TO-Perl
 - Fixed update procedure of servers, so that if a server is linked to one or more delivery services, you cannot change its "cdn".
+- Fixed `cachegroups` READ endpoint, so that if a request is made with the `type` specified as a non integer value, you get back a `400` with error details, instead of a `500`.
```
ocket8888, line 22: I see that you're just following everyone's example here but: Why don't we link to the issue that was resolved?

traffic_ops/traffic_ops_golang/cachegroup/cachegroups_test.go
```diff
@@ -272,3 +273,73 @@ func TestValidate(t *testing.T) {
...
+	}
+	_, userErr, _, sc := obj.Read()
+	if sc != http.StatusBadRequest {
+		t.Errorf("Read expected status code: Bad Request, actual: %v ", sc)
+	}
+	if userErr == nil {
+		t.Errorf("Read expected: user error with details %v, actual: nil", detail)
+	}
+	if userErr.Error() != detail {
+		t.Errorf("Read expected: user error with details %v, actual: %v", detail, userErr.Error())
```
mattjackson220, line 343: Can you change these `Errorf`s to `Fatalf`s?  At least the one on line 340 because otherwise in the case of a failure it will get userErr as nil for line 342 and get a nil exception. but might as well change them both

mattjackson220:
one small change to the test and this looks good to me! unit and API tests pass, code looks good, go fmt is happy, works as expected. nice work!

traffic_ops/traffic_ops_golang/cachegroup/cachegroups_test.go
srijeet0406: Yeah, I meant to do that before creating the PR, but slipped off my mind. Its done now.

mattjackson220:
retest this please

PR #4742, opened 2020-05-29
removes restriction on metric types from /api/deliveryservice_stats

### LINKED ISSUE #4740: GET /deliveryservice_stats limits metric types
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
The Go version of GET /deliveryservice_stats limits the metric types to the following:

```
dsMetricTypes = map[string]interface{}{
	"kbps":      struct{}{},
	"tps_total": struct{}{},
	"tps_2xx":   struct{}{},
	"tps_3xx":   struct{}{},
	"tps_4xx":   struct{}{},
	"tps_5xx":   struct{}{},
}
```

Therefore making it impossible to fetch `metricType=foo` for example. The Perl version did not restrict the metric types but rather it simply added whatever metric type to the influx query.

I would suggest fixing this in 1.x and 2.x (which is already part of TC 4.1) and potentially redesigning the entire endpoint for 3.x (which is part of TC 5.0). I'll created a separate issue for that - #4741 

## Expected / new behavior:
GET /api/deliveryservice_stats does not limit metric types to a predefined set.

GET /api/1.4/deliveryservice_stats?metricType=foo will return a 400 bad request.

mitchell852:
Fixes a regression where metric types were being limited to a predefined set.

- [x] This PR fixes #4740 

- Traffic Ops

Run the API tests.

master
4.0.0
4.1.0 RC1

- [x] This PR includes NO tests as tests for endpoints depending on external services don't exist


ocket8888:
This introduces a vulnerability that allows read-only users to destroy stats data. I don't recommend we include it, at least not this way.

dneuman64:
I understand the need for this and why Jeremy is fixing it.  We should allow some sort of latitude in the API to allow for some customization, but maybe not just strings.  What if we changed the code so the requesting user had to pass in a metric that pattern matched whatever we define?  So instead of ONLY allowing `kbps` we allow `kbps*` so that if we wanted to create a metric called `kbps_custom` it would still work?  I think that would satisfy our need to not have such a strict API while also keeping some safety.

rawlinp:
Tossing another idea out there, could we add a new config option that's just a whitelist of strings of acceptable metric types? That way adding a new metric type is just a reconfiguration, and it doesn't seem like we add new metric types often enough to where the reconfig+restart would be too much to ask.

dneuman64:
That could work though we still have to handle the data coming back which is why I suggested a wildcard.  In other words, if I choose to use `kbps_test` instead of just `kbps` then the data format still has to match.  I guess we could keep those guidelines with a whitelist.

mitchell852:
I think a whitelist probably makes sense. I can add support for that.

ocket8888:
I agree, whitelisting does sound safer/easier

dneuman64:
The downside with a whitelist is having to restart the application for it to take effect.  It would be nice if we could do a reload or have the list loaded dynamically.

ocket8888:
Does TO not support SIGHUP reloads? I feel like I should know that.

rawlinp:
AFAIK only the profiling config is SIGHUP-able. Making other TO config SIGHUP-able likely requires a bit more work, but it's something we should probably do (outside the scope of this issue).

rob05c:
Yes, only profiling currently: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/traffic_ops_golang.go#L225

It shouldn't be too difficult to swap atomic pointers for the HTTP Handler, to reload everything. TO could probably copy-and-paste Grove's code for that:

https://github.com/apache/trafficcontrol/blob/master/grove/cache/handler.go#L47
https://github.com/apache/trafficcontrol/blob/master/grove/grove.go#L296

mitchell852:
this PR now has a "whitelist" although named "supported_ds_metrics" instead. I'll leave all the SIGHUPness for another PR.

PR #4754, opened 2020-06-03
Fix API (deliveryserviceserver and deliveryservices/dsName/servers) should not assign Server from different CDN to Delivery Service

srijeet0406:

- Traffic Ops

Have TO running locally. 
Create a CDN.
Create a DS for this CDN.
Try to add a server, which belongs to a DIFFERENT CDN, to this DS using the backend calls:
`POST api/3.0/deliveryservices/{xml_id}/servers`
and/or
`POST api/3.0/deliveryserviceserver`
This should fail to add the server to the DS.
Now, try adding a server(s), with the same CDN ID as that of the DS, using the same calls as mentioned above.
These calls should now pass.

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
+++ b/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
@@ -598,15 +598,18 @@ func GetServerNameFromID(tx *sql.Tx, id int) (string, bool, error) {
 	return name, true, nil
 }
 
-type ServerHostNameAndType struct {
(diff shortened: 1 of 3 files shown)


shamrickus:
I believe you should also add an entry to `CHANGELOG.md`

srijeet0406:
@shamrickus Well, this is a follow up PR for https://github.com/apache/trafficcontrol/pull/4701, and that has a CHANGELOG entry, so I didnt do it again in this PR.

PR #4757, opened 2020-06-04
Add Raw Remap Range Directive

### LINKED ISSUE #4755: Traffic Ops Range Request Handling Slice Option Breaks With Range-Related Raw Remaps
## I'm submitting a ...
- [x] bug report

- [x] Traffic Ops
- [x] Traffic Ops ORT

## Current behavior:
The new Delivery Service Range Request Handling option for "Slice" added in https://github.com/apache/trafficcontrol/pull/4455 breaks with DS Raw Remaps which need to come before it.

For example, we have a Raw Remap which manipulates the Range header, which needs to come before the Slicer. But the new feature is always hard-coded inserted into the `remap.config`, with Raw Remaps (which always come last) later. This breaks the DS.

This is a major issue for us, and fundamentally breaks the DS Slicer feature for our primary use cases.

## Expected / new behavior:

We need a way to insert the Slicer after Raw Remaps. But Raw Remaps need to be last, and will break most existing services if they don't.

There are several possible solutions, most of which involve a great deal of complexity and operational footgunishness.

I suggest adding a `__SLICER__` directive to the Raw Remap config generation, which if it exists will insert the slicer config into the remap line there; and if it doesn't exist, it's inserted right before the Raw Remap (like it is in `master` now).

That seems like the quickest, safest solution. It's easy to develop and maintain. It's a little "footgunish," but it makes the footgun the exception, instead of the default. All the other options I see make the unintuitive/footgun the default, and most are considerably more work to develop and maintain.

Use the Range Request Handling Slicer option with a Delivery Service which manipulates the Range. Observe catastrophic failure.

rob05c:
Adds a `__RANGE_DIRECTIVE__` to the Raw Remap processing, to insert the
Delivery Service Range Request Handling directives there, instead of
before the Raw Remap.

Normally, Raw Remap always needs to be at the end of the remap line,
but in some cases, specifically when custom remaps are modifying
the Range header, it's necessary to put the Range Request directives
(@plugin=slice.so, @plugin=cache_range_requests.so, etc) after
the custom Raw Remap text.

This makes that possible.

Tested manually, verified range directives are placed over `__RANGE_DIRECTIVE__` in the Raw Remap, range directives are placed as normally if no directive exists, range plugins aren't double-placed. Includes unit tests verifying same.

Fixes #4755 

Includes tests.
Includes docs.
Includes changelog.

- [x] This PR fixes #4755 

- Traffic Ops ORT

Run tests.
Create a Delivery Service with a Range Request Handling: Slice, and a Raw Remap of `sometext __RANGE_DIRECTIVE__`, queue updates, verify the `@plugin=slice.so` is after `sometext` in `remap.config` on the cache.
Remove `__RANGE_DIRECTIVE__` from the Delivery Service, queue, verify Raw Remap Text is at the end of the line like it should be.

This is somewhat between a bug fix and a feature. It's a feature that fixes a critical issue in a new feature in 4.1, #4455

- master 
- 4.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed update procedure of servers, so that if a server is linked to one or more delivery services, you cannot change its "cdn". [Related github issue](https://github.com/apache/trafficcontrol/issues/4116)
 - Fixed `POST /api/x/steering` and `PUT /api/x/steering` so that a steering target with an invalid `type` is no longer accepted. [Related github issue](https://github.com/apache/trafficcontrol/issues/3531)
 - Fixed `cachegroups` READ endpoint, so that if a request is made with the `type` specified as a non integer value, you get back a `400` with error details, instead of a `500`. [Related github issue](https://github.com/apache/trafficcontrol/issues/4703)
+- Added Delivery Service Raw Remap `__RANGE_DIRECTIVE__` directive to allow inserting the Range Directive after the Raw Remap text. This allows Raw Remaps which manipulate the Range.
 
 ### Changed
 - Changed some Traffic Ops Go Client methods to use `DeliveryServiceNullable` inputs and outputs.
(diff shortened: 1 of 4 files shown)


rob05c:
Pinged @traeak to review, since he saw the original issue, and is probably most qualified to say if it's fixed. (Can't actually add as a reviewer, since he's not a committer.)

PR #4763, opened 2020-06-08
Rewrote postinstall to Python (3.6+)

ocket8888:
This PR features a rewrite of postinstall to Python3 (specifically the `_postinstall` Perl script). This is important because the Perl version of the script depended upon multiple libraries installed as dependencies of the Perl version of Traffic Ops - and if we are ever to get rid of that TO version things that depend on its dependencies will need to be reworked.

The new `postinstall.py` script has no dependencies outside of the Python3 standard library, and features several unit tests and an end-to-end test bash script whereas the Perl version had only a single unit test (executed at run time on every run).

To be clear, though, it - like the Perl script it is intended to replace - has dependencies on external programs including `wget`, `openssl` and `psql`. `wget` is only needed for Maxmind download support, so it could possibly be removed as the hard-coded URLs don't work anyway, but is left in for consistency with the old script. Besides that, its only external dependencies are already dependencies of TO - even the Go version - and the Python3 interpreter itself.

Finally, I should mention that this doesn't actually replace the old script, which is totally untouched, but provides the new Python script as an alternative. Many machines running TO in the wild may not have Python3 installed yet, so it may be too soon yet to require that for TO installations/upgrades.

- Traffic Ops

To run the doctests, use `python -m doctest path/to/postinstall.py`. That should output nothing and exit successfully (my `python` is a Python3 interpreter, if yours is Python2 then ~~upgrade please, Python2 is dead~~ you'll need to use `python3 -m doctest path/to/postinstall.py` instead).

To run the end-to-end tests, use the `postinstall.test.sh` script, which should also output nothing (except spitting out some ssl keys/certs/requests in the current directory) and exit successfully.

- [x] This PR includes tests
- [x] Documentation is unnecessary - this isn't actually used yet. It does have docstrings and a `-h`/`--help` option, though

### DIFF
--- a/traffic_ops/install/bin/postinstall.py
+++ b/traffic_ops/install/bin/postinstall.py
@@ -25,9 +25,11 @@ import hashlib
 import json
 import logging
 import os
+import random
 import re
 import shutil
 import stat
+import string
 import subprocess
 import sys
 import typing
@@ -119,6 +121,16 @@ class SSLConfig():
 
 	def params(self) -> str:
 		return f"/C={self.country}/ST={self.state}/L={self.locality}/O={self.company}/OU={self.org_unit}/CN={self.common_name}/"
+
+class CDNConfig():
+
+	def __init__(self, gen_secret: bool, num_secrets: int, port: int, num_workers: int, url: str, ldap_conf_location: str):
+		self.gen_secret = gen_secret
+		self.num_secrets = num_secrets
+		self.port = port
+		self.num_workers = num_workers
+		self.url = url
+		self.ldap_conf_location = ldap_conf_location
 
 # The default question/answer set
 DEFAULTS = {
@@ -547,6 +559,93 @@ def setup_certificates(conf: SSLConfig, root: str, ops_user: str, ops_group: str
 
 	return 0
 
+def random_word(length: int = 12) -> str:
+	word_chars = string.ascii_letters + string.digits + '_'
+	return ''.join(random.choice(word_chars) for _ in range(length))
+
+def generate_cdn_conf(questions:typing.List[Question], fname: str, automatic: bool, root: str):
+	cdn_conf = get_config(questions, fname, automatic)
+
+	if "genSecret" not in cdn_conf:
+		raise ValueError("missing 'genSecret' config_var")
+
+	gen_secret = cdn_conf["genSecret"].casefold() in {'y', 'yes'}
+
+	try:
+		num_secrets = int(cdn_conf["keepSecrets"])
+	except KeyError as e:
+		raise ValueError("missing 'keepSecrets' config_var") from e
+	except ValueError as e:
+		raise ValueError(f"invalid 'keepSecrets' config_var value: {e}") from e
+
+	try:
+		port = cdn_conf.get("port")
+	except KeyError as e:
+		raise ValueError("missing 'port' config_var") from e
+	except ValueError as e:
+		raise ValueError(f"invalid 'port' config_var value: {e}") from e
+
+	try:
+		workers = int(cdn_conf["workers"])
+	except KeyError as e:
+		raise ValueError("missing 'workers' config_var") from e
+	except ValueError as e:
+		raise ValueError(f"invalid 'workers' config_var value: {e}")
+
+	try:
+		url = cdn_conf["base_url"]
+	except KeyError as e:
+		raise ValueError("missing 'base_url' config_var") from e
+
+	try:
+		ldap_loc = cdn_conf["ldap_conf_location"]
+	except KeyError as e:
+		raise ValueError("missing 'ldap_conf_location' config_var") from e
+
+	conf = CDNConfig(gen_secret, num_secrets, port, workers, url, ldap_loc)
+
+	path = os.path.join(root, fname.lstrip('/'))
+	existingConf = {}
+	if os.path.isfile(path):
+		with open(path) as fd:
+			try:
+				existingConf = json.load(fd)
+			except json.JSONDecodeError as e:
+				raise ValueError(f"invalid existing cdn.config at {path}: {e}") from e
+
+	if type(existingConf) is not dict:
+		logging.warning("Existing cdn.conf (at '%s') is not an object - overwriting", path)
+		existingConf = {}
+
+	if conf.gen_secret:
+		if type(existingConf) is dict and "secrets" in existingConf and type(existingConf["secrets"]) is list:
+			logging.debug("Secrets found in cdn.conf file")
+		else:
+			existingConf["secrets"] = []
+			logging.debug("No secrets found in cdn.conf file")
+
+		existingConf["secrets"].insert(0, random_word())
+
+		if conf.num_secrets and len(existingConf["secrets"]) > conf.num_secrets:
+			existingConf["secrets"] = existingConf["secrets"][:conf.num_secrets - 1]
+
+	if conf.url:
+		if "to" not in existingConf or type(existingConf["to"]) is not dict:
+			existingConf["to"] = {}
+		existingConf["to"]["base_url"] = conf.url
+
+	if "traffic_ops_golang" not in existingConf or type(existingConf["traffic_ops_golang"]) is not dict:
+		existingConf["traffic_ops_golang"] = {}
+	existingConf["traffic_ops_golang"]["port"] = conf.port
+	existingConf["traffic_ops_golang"]["log_location_error"] = os.path.join(root, "var/log/traffic_ops/error.log")
+	existingConf["traffic_ops_golang"]["log_location_event"] = os.path.join(root, "var/log/traffic_ops/access.log")
+
+	if "hypnotoad" not in existingConf or type(existingConf["hypnotoad"]) is not dict:
+		existingConf["hypnotoad"]["workers"] = conf.num_workers
+
+	with open(path, "w+") as fd:
+		json.dump(existingConf, fd, indent="\t")
+	logging.info("CDN configuration has been saved")
 
 def main(automatic: bool, debug: bool, defaults: str = None, cfile: str = None, root_dir: str = "/", ops_user: str = "trafops", ops_group: str = "trafops") -> int:
 	"""
@@ -637,6 +736,12 @@ def main(automatic: bool, debug: bool, defaults: str = None, cfile: str = None, 
 			return cert_code
 	except OSError as e:
 		logging.critical("Setting up SSL Certificates: %s", e)
+		return 1
+
+	try:
+		generate_cdn_conf(userInput[CDN_CONF_FILE], CDN_CONF_FILE, automatic, root_dir)
+	except OSError as e:
+		logging.critical("Generating cdn.conf: %s", e)
 		return 1
 
 	return 0
--- a/traffic_ops/install/bin/postinstall.py
+++ b/traffic_ops/install/bin/postinstall.py
@@ -523,7 +523,7 @@ def unmarshal_config(dct: dict) -> typing.Dict[str, typing.List[Question]]:
 		if not isinstance(questions, list):
 			raise ValueError(f"file '{file}' has malformed questions")
 
-		questions = []
+		qstns = []
 		for qstn in questions:
 			if not isinstance(qstn, dict):
 				raise ValueError(f"file '{file}' has a malformed question ({qstn})")
@@ -553,8 +553,8 @@ def unmarshal_config(dct: dict) -> typing.Dict[str, typing.List[Question]]:
 			if qstn:
 				logging.warning("Found unknown extra properties in question in '%s' (%r)", file, qstn.keys())
 
-			questions.append(Question(question, answer, cfg_var, hidden=hidden))
-		ret[file] = questions
+			qstns.append(Question(question, answer, cfg_var, hidden=hidden))
+		ret[file] = qstns
 
 	return ret
 
--- a/traffic_ops/install/bin/postinstall.py
+++ b/traffic_ops/install/bin/postinstall.py
@@ -1,0 +1,81 @@
+#!/usr/bin/python3
+
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import getpass
+import logging
+import sys
+import typing
+
+# Paths for output configuration files
+DATABASE_CONF_FILE = "/opt/traffic_ops/app/conf/production/database.conf"
+DB_CONF_FILE       = "/opt/traffic_ops/app/db/dbconf.yml"
+CDN_CONF_FILE      = "/opt/traffic_ops/app/conf/cdn.conf"
+LDAP_CONF_FILE     = "/opt/traffic_ops/app/conf/ldap.conf"
+USERS_CONF_FILE    = "/opt/traffic_ops/install/data/json/users.json"
+PROFILES_CONF_FILE = "/opt/traffic_ops/install/data/profiles/"
+OPENSSL_CONF_FILE  = "/opt/traffic_ops/install/data/json/openssl_configuration.json"
+PARAM_CONF_FILE    = "/opt/traffic_ops/install/data/json/profiles.json"
+
+CUSTOM_PROFILE_DIR = PROFILES_CONF_FILE + "custom"
+
+# Location of Traffic Ops profiles
+PROFILE_DIR = "/opt/traffic_ops/install/data/profiles/"
+POST_INSTALL_CFG = "/opt/traffic_ops/install/data/json/post_install.json"
+
+# Log file for the installer
+LOG_FILE = "/var/log/traffic_ops/postinstall.log"
+
+# Log file for CPAN output
+CPAN_LOG_FILE = "/var/log/traffic_ops/cpan.log"
+
+# Configuration file output with answers which can be used as input to postinstall
+OUTPUT_CONFIG_FILE = "/opt/traffic_ops/install/bin/configuration_file.json"
+
+def get_field(question: str, config_answer: str, hidden: bool = False) -> str:
+	if hidden:
+		while True:
+			pw = getpass.getpass(question)
+			if pw == getpass.getpass(f"Re-Enter {question}"):
+				return pw
+			print("Error: passwords do not match, try again", file=sys.stderr)
+
+	if config_answer:
+		ipt = input(f"{question} [{config_answer}]: ")
+		return ipt if ipt else config_answer
+	return input(question + ": ")
+
+def get_config(user_input: typing.Dict[str, typing.List[typing.Dict[str, str]]], fname: str, automatic: bool = False) -> dict:
+
+	if fname not in user_input:
+		raise ValueError(f"No {fname} found in config")
+
+	logging.info(f"==========={fname}===========")
+
+	config = {}
+
+	for var in user_input[fname]:
+		question = next(key for key in var.keys() if key != "hidden" and key != "config_var")
+		hidden = var.get("hidden")
+		answer = var.get(question) if automatic else get_field(question, var.get(question), hidden)
+
+		config[var.get(config_var)] = answer
+
+	return config
+
+def generate_db_conf(user_input: dict, fname: str, todb_fname: str, automatic: bool):
+	"""
+	"""
+	db_conf = get_config(user_input, fname, automatic)
--- a/traffic_ops/install/bin/postinstall.py
+++ b/traffic_ops/install/bin/postinstall.py
@@ -1249,15 +1249,15 @@ if __name__ == '__main__':
 	ARGS = PARSER.parse_args()
 
 	USED_LEGACY_ARGS = False
-	DEFAULTS = None
+	DEFAULTS_ARG = None
 	if ARGS.legacy_defaults:
 		if ARGS.defaults:
 			logging.error("cannot specify both '--defaults' and '-defaults'")
 			sys.exit(1)
 		USED_LEGACY_ARGS = True
-		DEFAULTS = ARGS.legacy_defaults
+		DEFAULTS_ARG = ARGS.legacy_defaults
 	else:
-		DEFAULTS = ARGS.defaults
+		DEFAULTS_ARG = ARGS.defaults
 
 	DEBUG = False
 	if ARGS.legacy_debug:
@@ -1293,7 +1293,7 @@ if __name__ == '__main__':
 		EXIT_CODE = main(
 		ARGS.automatic,
 		DEBUG,
-		DEFAULTS,
+		DEFAULTS_ARG,
 		CFILE,
 		os.path.abspath(ARGS.root_directory),
 		ARGS.ops_user,
--- a/traffic_ops/install/bin/postinstall.py
+++ b/traffic_ops/install/bin/postinstall.py
@@ -365,7 +365,7 @@ def generate_users_conf(qstns: typing.List[Question], fname: str, auto: bool, ro
 	path = os.path.join(root, fname.lstrip('/'))
 	with open(path, 'w+') as conf_file:
 		json.dump({"username": config["tmAdminUser"], "password": hashed_pass}, conf_file, indent="\t")
-		print(file=fd)
+		print(file=conf_file)
 
 	return User(config["tmAdminUser"], config["tmAdminPw"])
 
@@ -390,7 +390,7 @@ def generate_openssl_conf(questions: typing.List[Question], fname: str, auto: bo
 	return SSLConfig(gen_cert, cfg_map)
 
 def generate_param_conf(qstns: typing.List[Question], fname: str, auto: bool, root: str) -> dict:
-	conf = get_config(qstns, fname, automatic)
+	conf = get_config(qstns, fname, auto)
 
 	path = os.path.join(root, fname.lstrip('/'))
 	with open(path, 'w+') as conf_file:
@@ -415,24 +415,26 @@ def sanity_check_config(cfg: typing.Dict[str, typing.List[Question]], automatic:
 			logging.warning("File '{}' found in defaults but not config file", fname)
 			cfg[fname] = []
 
-		for defaultValue in file:
-			for configValue in cfg[fname]:
-				if defaultValue.config_var == configValue.config_var:
+		for default_value in file:
+			for config_value in cfg[fname]:
+				if default_value.config_var == config_value.config_var:
 					break
 			else:
-				question = defaultValue.question
-				answer = defaultValue.default
+				question = default_value.question
+				answer = default_value.default
 
 				if not automatic:
 					logging.info("Prompting user for answer")
-					if defaultValue.hidden:
-						answer = defaultValue.ask()
+					if default_value.hidden:
+						answer = default_value.ask()
+				elif default_value.hidden:
+					logging.info("Adding question '{}' with default answer", question)
 				else:
-					logging.info("Adding question '{}' with default answer{}", question, f" {answer}" if not defaultValue.hidden else "")
+					logging.info("Adding question '{}' with default answer {}", question, answer)
 
 				# The Perl here would ask questions, but those would just get asked later
 				# anyway, so I'm not sure why.
-				cfg[fname].append(Question(question, answer, defaultValue.config_var, defaultValue.hidden))
+				cfg[fname].append(Question(question, answer, default_value.config_var, default_value.hidden))
 				diffs += 1
 
 	return diffs
@@ -440,6 +442,7 @@ def sanity_check_config(cfg: typing.Dict[str, typing.List[Question]], automatic:
 def unmarshal_config(dct: dict) -> typing.Dict[str, typing.List[Question]]:
 	"""
 	Reads in a raw parsed configuration file and returns the resulting configuration.
+
 	>>> unmarshal_config({"test": [{"Do the thing?": "yes", "config_var": "thing"}]})
 	{'test': [Question(question='Do the thing?', default='yes', config_var='thing', hidden=False)]}
 	>>> unmarshal_config({"test": [{"foo": "", "config_var": "bar", "hidden": True}]})
@@ -447,21 +450,22 @@ def unmarshal_config(dct: dict) -> typing.Dict[str, typing.List[Question]]:
 	"""
 	ret = {}
 	for file, questions in dct.items():
-		if type(questions) is not list:
+		if isinstance(questions, list):
 			raise ValueError(f"file '{file}' has malformed questions")
 
 		qs = []
 		for q in questions:
-			if type(q) is not dict:
+			if not isinstance(q, dict):
 				raise ValueError(f"file '{file}' has a malformed question ({q})")
 			try:
-				question = next(key for key in q.keys() if q != "hidden" and q != "config_var")
+				question = next(key for key in q.keys() if q not in ("hidden", "config_var"))
 			except StopIteration:
 				raise ValueError(f"question in '{file}' has no question/answer properties ({q})")
 
 			answer = q[question]
-			if type(question) is not str or type(answer) is not str:
-				raise ValueError(f"question in '{file}' has malformed question/answer property ({question}: {answer})")
+			if not isinstance(question, str) or not isinstance(answer, str):
+				errstr = f"question in '{file}' has malformed question/answer property ({question}: {answer})"
+				raise ValueError(errstr)
 
 			del q[question]
 			hidden = False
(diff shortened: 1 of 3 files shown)


traffic_ops/install/bin/postinstall.py
```diff
@@ -0,0 +1,1248 @@
+#!/usr/bin/python3
```
zrhoffman, line 1: If we want to support python2, too, this should be just `python`, not `python3`.

CentOS 7 and CentOS 8 both have a python interpreter at `/usr/libexec/platform-python` (on CentOS 7 it is Python 2.7, on CentOS 8, it is Python 3.6). To accept a binary named `python` if it exists but falls back to `platform-python`, the shebang could be something like

```shell
#!/usr/bin/env -S sh -c 'PATH="${PATH}:/usr/libexec" exec "$(type -p python platform-python | head -n1)" "$0" "$@"'
```

zrhoffman (changes requested):
Python 2.7 is supported by Red Hat [through June 2024](https://access.redhat.com/support/policy/updates/rhel8-app-streams-life-cycle#th-collect19-b). Although we definitely don't *need* to support Python 2.7 just for this script to exist in the repo, there would be a much stronger case for this Python script to replace the Perl one if it could be run using both Python 2.7 and 3.6.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+			"config_var": "dns_subdomain",
+			"hidden": false
+		}
+	]
+}
+EOF
+
+"$MY_DIR/postinstall.py" --no-root --root-directory="$ROOT_DIR" --no-restart-to --no-database --ops-user="$(whoami)" --ops-group="$(whoami)" --automatic --cfile="$ROOT_DIR/defaults.json" --debug 2>"$ROOT_DIR/stderr" | tee "$ROOT_DIR/stdout"
+
+if [[ ! -z "$(grep 'ERROR' $ROOT_DIR/stderr)" ]]; then
```
zrhoffman, line 249: No need for a string comparison, this can just be `if grep -q [...]`

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+ROOT_DIR="$(mktemp -d)";
+
+trap 'rm -rf $ROOT_DIR' EXIT;
+
+mkdir -p "$ROOT_DIR/etc/pki/tls/certs";
+mkdir "$ROOT_DIR/etc/pki/tls/private";
+mkdir -p "$ROOT_DIR/opt/traffic_ops/app/public/routing";
+mkdir "$ROOT_DIR/opt/traffic_ops/app/db";
+mkdir -p "$ROOT_DIR/opt/traffic_ops/app/conf/production";
+echo "{\"hypnotoad\":{\"listen\":[\"https://[::]:60443?cert=$ROOT_DIR/etc/pki/tls/certs/localhost.crt&key=$ROOT_DIR/etc/pki/tls/private/localhost.key\"]}}" > "$ROOT_DIR/opt/traffic_ops/app/conf/cdn.conf";
```
zrhoffman, line 35: This would do better as `cat` with a here document so the double quotes don't need to be backslashed (and, optionally, so the JSON can be formatted)

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+		{
+			"DNS sub-domain for which your CDN is authoritative": "cdn1.kabletown.net",
+			"config_var": "dns_subdomain",
+			"hidden": false
+		}
+	]
+}
+EOF
+
+"$MY_DIR/postinstall.py" --no-root --root-directory="$ROOT_DIR" --no-restart-to --no-database --ops-user="$(whoami)" --ops-group="$(whoami)" --automatic --cfile="$ROOT_DIR/defaults.json" --debug 2>"$ROOT_DIR/stderr" | tee "$ROOT_DIR/stdout"
```
zrhoffman, line 247: `whoami` for the user and group can be `id -un` and `id -gn`, respectively

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+# under the License.
+
+set -e;
+
+if [[ ! -x /usr/bin/python3 ]]; then
+	echo "Python 3.6+ is required to run - or test - postinstall.py" >&2;
+	exit 1;
+fi
+
+ROOT_DIR="$(mktemp -d)";
```
zrhoffman, line 26: `ROOT_DIR` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+mkdir -p "$ROOT_DIR/etc/pki/tls/certs";
+mkdir "$ROOT_DIR/etc/pki/tls/private";
+mkdir -p "$ROOT_DIR/opt/traffic_ops/app/public/routing";
+mkdir "$ROOT_DIR/opt/traffic_ops/app/db";
+mkdir -p "$ROOT_DIR/opt/traffic_ops/app/conf/production";
+echo "{\"hypnotoad\":{\"listen\":[\"https://[::]:60443?cert=$ROOT_DIR/etc/pki/tls/certs/localhost.crt&key=$ROOT_DIR/etc/pki/tls/private/localhost.key\"]}}" > "$ROOT_DIR/opt/traffic_ops/app/conf/cdn.conf";
+mkdir -p "$ROOT_DIR/opt/traffic_ops/install/data/json";
+mkdir "$ROOT_DIR/opt/traffic_ops/install/bin";
+
+MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
zrhoffman, line 39: `MY_DIR` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+if not password.startswith('SCRYPT:16384:8:1:') or len(password.split(':')) != 6:
+	print('Malformed password field in users.json:', password, file=sys.stderr)
+	exit(1)
+
+exit(0)
+"
+
+/usr/bin/python3 -c "$USERS_TEST";
+
+POST_INSTALL_JSON="$ROOT_DIR/opt/traffic_ops/install/data/json/post_install.json";
```
zrhoffman, line 295: `POST_INSTALL_JSON` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+if [[ "$(cat $POST_INSTALL_JSON)" != "{}" ]]; then
+	echo "Incorrect post_install.json, expected: {}, got: $(cat $POST_INSTALL_JSON)" >&2;
+	exit 1;
+fi
+
+PROFILES_JSON_EXPECTED="{
```
zrhoffman, line 301: `PROFILES_JSON_EXPECTED` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+	\"tm.url\": \"https://localhost\",
+	\"cdn_name\": \"kabletown_cdn\",
+	\"dns_subdomain\": \"cdn1.kabletown.net\"
+}";
+
+PROFILES_JSON_ACTUAL="$(cat $ROOT_DIR/opt/traffic_ops/install/data/json/profiles.json)";
```
zrhoffman, line 307: `PROFILES_JSON_ACTUAL` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+if [[ "$PROFILES_JSON_ACTUAL" != "$PROFILES_JSON_EXPECTED" ]]; then
+	echo "Incorrect profiles.json, expected: $PROFILES_JSON_EXPECTED, got: $PROFILES_JSON_ACTUAL" >&2;
+	exit 1;
+fi
+
+DB_CONF_EXPECTED="production:
```
zrhoffman, line 313: `DB_CONF_EXPECTED` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+    driver: postgres
+    open: host=localhost port=5432 user=traffic_ops password=twelve dbname=traffic_ops sslmode=disable";
+
+DB_CONF_ACTUAL="$(cat $ROOT_DIR/opt/traffic_ops/app/db/dbconf.yml)";
```
zrhoffman, line 317: `DB_CONF_ACTUAL` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+if [[ "$DB_CONF_ACTUAL" != "$DB_CONF_EXPECTED" ]]; then
+	echo "Incorrect dbconf.yml, expected:" >&2;
+	echo "$DB_CONF_EXPECTED" >&2;
+	echo "got:" >&2;
+	echo "$DB_CONF_ACTUAL" >&2;
+	exit 1;
+fi
+
+CDN_CONF_TEST="
```
zrhoffman, line 326: `CDN_CON_TEST` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+	print('Incorrect traffic_ops_golang.log_location_event in cdn.conf, expected: $ROOT_DIR/var/log/traffic_ops/access.log, got:', conf['traffic_ops_golang']['log_location_event'], file=sys.stderr)
+	exit(1)
+
+exit(0)
+";
+
+/usr/bin/python3 -c "#CDN_CONF_TEST";
+
+
+DATABASE_CONF_EXPECTED='{
```
zrhoffman, line 389: `DATABASE_CONF_EXPECTED` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+	"type": "Pg",
+	"dbname": "traffic_ops",
+	"hostname": "localhost",
+	"port": "5432",
+	"user": "traffic_ops",
+	"password": "twelve",
+	"description": "Pg database on localhost:5432"
+}';
+
+DATABASE_CONF_ACTUAL="$(cat $ROOT_DIR/opt/traffic_ops/app/conf/production/database.conf)";
```
zrhoffman, line 399: `DATABASE_CONF_ACTUAL` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+if [[ "$DATABASE_CONF_ACTUAL" != "$DATABASE_CONF_EXPECTED" ]]; then
+	echo "Incorrect database.conf, expected: $DATABASE_CONF_EXPECTED, got $DATABASE_CONF_ACTUAL" >&2;
+	exit 1;
+fi
+
+CSR_FILE="$ROOT_DIR/etc/pki/tls/certs/localhost.csr";
```
zrhoffman, line 405: `CSR_FILE` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+CSR_FILE_TYPE="$(file $CSR_FILE)";
```
zrhoffman, line 406: `CSR_FILE_TYPE` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+if [[ "$CSR_FILE_TYPE" != "$CSR_FILE: PEM certificate request" ]]; then
+	echo "Incorrect csr file, expected a PEM certificate request, got: $CSR_FILE_TYPE" >&2;
+	exit 1;
+fi
+
+CERT_FILE="$ROOT_DIR/etc/pki/tls/certs/localhost.crt";
```
zrhoffman, line 412: `CERT_FILE` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+CERT_FILE_TYPE="$(file $CERT_FILE)";
```
zrhoffman, line 413: `CERT_FILE_TYPE` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+if [[ "$CERT_FILE_TYPE" != "$CERT_FILE: PEM certificate" ]]; then
+	echo "Incorrect cert file, expected a PEM certificate, got: $CERT_FILE_TYPE" >&2;
+	exit 1;
+fi
+
+KEY_FILE="$ROOT_DIR/etc/pki/tls/private/localhost.key";
```
zrhoffman, line 419: `KEY_FILE` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+KEY_FILE_TYPE="$(file $KEY_FILE)";
```
zrhoffman, line 420: `KEY_FILE_TYPE` should not be all caps because it is not an environment variable.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+	exit(1)
+
+if not password.startswith('SCRYPT:16384:8:1:') or len(password.split(':')) != 6:
+	print('Malformed password field in users.json:', password, file=sys.stderr)
+	exit(1)
+
+exit(0)
+"
+
+/usr/bin/python3 -c "$USERS_TEST";
```
zrhoffman, line 293: Instead of storing this script to `USERS_TEST` and using it as a really long argument to `python -c`, if this script is a here document, `/usr/bin/python3` can be the command that the here document is standard input for.

traffic_ops/install/bin/postinstall.test.sh
zrhoffman, line 326: Instead of storing this script to `CDN_CONF_TEST` and using it as a really long argument to `python -c`, if this script is a here document, `/usr/bin/python3` can be the command that the here document is standard input for.

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -0,0 +1,424 @@
...
+mkdir -p "$ROOT_DIR/opt/traffic_ops/app/public/routing";
+mkdir "$ROOT_DIR/opt/traffic_ops/app/db";
+mkdir -p "$ROOT_DIR/opt/traffic_ops/app/conf/production";
+echo "{\"hypnotoad\":{\"listen\":[\"https://[::]:60443?cert=$ROOT_DIR/etc/pki/tls/certs/localhost.crt&key=$ROOT_DIR/etc/pki/tls/private/localhost.key\"]}}" > "$ROOT_DIR/opt/traffic_ops/app/conf/cdn.conf";
+mkdir -p "$ROOT_DIR/opt/traffic_ops/install/data/json";
+mkdir "$ROOT_DIR/opt/traffic_ops/install/bin";
+
+MY_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+cat <<- EOF > "$ROOT_DIR/defaults.json"
```
zrhoffman, line 41: `EOF` is not a very descriptive delimiter, is there a delimiter that better describes the here document's contents?

zrhoffman (changes requested):

ocket8888:
> Python 2.7 is supported by Red Hat through June 2024. Although we definitely don't need to support Python 2.7 just for this script to exist in the repo, there would be a much stronger case for this Python script to replace the Perl one if it could be run using both Python 2.7 and 3.6.

But ATC as a project doesn't support Python2.7 as of 4.0.0

zrhoffman:
> But ATC as a project doesn't support Python2.7 as of 4.0.0

Not trying to convince you, just describing the conditions under which it
could replace the Perl postinstall.

traffic_ops/install/bin/postinstall.test.sh
ocket8888: Doing that causes the test to fail to find `users.json`:
```diff
diff --git a/traffic_ops/install/bin/postinstall.test.sh b/traffic_ops/install/bin/postinstall.test.sh
index d98c11645..cb8405c12 100755
--- a/traffic_ops/install/bin/postinstall.test.sh
+++ b/traffic_ops/install/bin/postinstall.test.sh
@@ -32,7 +32,15 @@ mkdir "$ROOT_DIR/etc/pki/tls/private";
 mkdir -p "$ROOT_DIR/opt/traffic_ops/app/public/routing";
 mkdir "$ROOT_DIR/opt/traffic_ops/app/db";
 mkdir -p "$ROOT_DIR/opt/traffic_ops/app/conf/production";
-echo "{\"hypnotoad\":{\"listen\":[\"https://[::]:60443?cert=$ROOT_DIR/etc/pki/tls/certs/localhost.crt&key=$ROOT_DIR/etc/pki/tls/private/localhost.key\"]}}" > "$ROOT_DIR/opt/traffic_ops/app/conf/cdn.conf";
+cat > "$ROOT_DIR/opt/traffic_ops/app/conf/cdn.conf" <<EOF
+{
+       "hypnotoad": {
+               "listen": [
+                       "https://[::]:60443?cert=$ROOT_DIR/etc/pki/tls/certs/localhost.crt&key=$ROOT_DIR/etc/pki/tls/private/localhost.key"
+               ]
+       }
+}
+EOF;
 mkdir -p "$ROOT_DIR/opt/traffic_ops/install/data/json";
 mkdir "$ROOT_DIR/opt/traffic_ops/install/bin";
```

causes:

```
Error loading users.json file: [Errno 2] No such file or directory: '/tmp/tmp.7vg08w3Ow2/opt/traffic_ops/install/data/json/users.json'
```

I don't know how or why but it does

traffic_ops/install/bin/postinstall.test.sh
ocket8888: I'll replace the group one, but the `whoami` manpage says

> Same as id -un.

So there's no reason to replace that one

traffic_ops/install/bin/postinstall.test.sh
ocket8888: says who?

traffic_ops/install/bin/postinstall.test.sh
zrhoffman: If the delimiter ends with a semicolon like `EOF;`, it interprets the semicolon as part of the delimiter.

traffic_ops/install/bin/postinstall.test.sh
ocket8888: Does it need to be descriptive? It is replacing an end-of-file character. To be honest, I find delimiters *besides* EOF confusing.

zrhoffman: The description doesn't need to be in the delimiter. But if it's not, there should be a comment that explains the significance of the content redirected to defaults.json.

traffic_ops/install/bin/postinstall.test.sh
zrhoffman: From [Google's shell style guide](https://github.com/google/styleguide/blob/gh-pages/shellguide.md#constants-and-environment-variable-names):

>### Constants and Environment Variable Names
>
>All caps, separated with underscores, declared at the top of the file.
>
>Constants and anything exported to the environment should be
>capitalized.
>
>```shell
># Constant
>readonly PATH_TO_FILES='/some/path'
>
># Both constant and environment
>declare -xr ORACLE_SID='PROD'
>```

In other words, if you want ROOT_DIR to be a constant, it should
* Be declared using `readonly` or `declare -r`
* Be declared at the top of the file instead of inline

zrhoffman: Granted, declaring inline vs at the top is more subjective.

traffic_ops/install/bin/postinstall.test.sh
ocket8888: I have no idea how that caused that error, but it's fixed now.

traffic_ops/install/bin/postinstall.test.sh
ocket8888: Google ain't the boss of me. But I don't have a problem marking things that should be readonly with `readonly`.

traffic_ops/install/bin/postinstall.test.sh
ocket8888: A comment sounds like a much better place for descriptions of things.

traffic_ops/install/bin/postinstall.test.sh
ocket8888: I resolved all the other conversations about the same thing so we can just discuss it here

zrhoffman:
Everything looks good so far, just need to test

zrhoffman (changes requested):
I'm getting an error on CentOS 8.2 using Python 3.6.8:
```
[root@trafficops-perl app]# /usr/libexec/platform-python ../install/bin/postinstall.py -a --cfile postinstall.json
INFO:root:Starting postinstall
INFO:root:Debug is on
INFO:root:Running in automatic mode
INFO:root:Using input file postinstall.json
INFO:root:Adding question 'Port to serve on?' with default answer 443
INFO:root:Adding question 'ldap.conf location?' with default answer /opt/traffic_ops/app/conf/ldap.conf
INFO:root:Adding question 'LDAP server hostname' with default answer
INFO:root:Adding question 'LDAP Admin DN' with default answer
INFO:root:Adding question 'LDAP Admin Password' with default answer
INFO:root:Adding question 'LDAP Search Base' with default answer
INFO:root:Adding question 'LDAP Search Query' with default answer
INFO:root:Adding question 'LDAP Skip TLS verify' with default answer
INFO:root:Adding question 'LDAP Timeout Seconds' with default answer
INFO:root:Adding question 'Country Name (2 letter code)' with default answer
INFO:root:Adding question 'State or Province Name (full name)' with default answer
INFO:root:Adding question 'Locality Name (eg, city)' with default answer
INFO:root:Adding question 'Organization Name (eg, company)' with default answer
INFO:root:Adding question 'Organizational Unit Name (eg, section)' with default answer
INFO:root:Adding question 'Common Name (eg, your name or your server's hostname)' with default answer
INFO:root:Adding question 'RSA Passphrase' with default answer
INFO:root:File sanity check complete - found 16 differences
INFO:root:===========/opt/traffic_ops/app/conf/production/database.conf===========
INFO:root:Database configuration has been saved
INFO:root:===========/opt/traffic_ops/app/db/dbconf.yml===========
INFO:root:Not setting up ldap
INFO:root:===========/opt/traffic_ops/install/data/json/users.json===========
INFO:root:===========/opt/traffic_ops/install/data/json/openssl_configuration.json===========
INFO:root:===========/opt/traffic_ops/install/data/json/profiles.json===========
INFO:root:Not downloading Maxmind data
INFO:root:Not generating openssl certification
INFO:root:===========/opt/traffic_ops/app/conf/cdn.conf===========
INFO:root:CDN configuration has been saved
INFO:root:paramconf {'tm.url': 'https://trafficops.infra.ciab.test:443', 'cdn_name': 'CDN-in-a-Box', 'dns_subdomain': 'mycdn.ciab.test'}
INFO:root:Setting up the database data
Traceback (most recent call last):
  File "/opt/traffic_ops/install/bin/postinstall.py", line 1241, in <module>
    ARGS.no_database
  File "/opt/traffic_ops/install/bin/postinstall.py", line 1140, in main
    setup_database_data(conn_str, admin_conf, paramconf, root_dir)
  File "/opt/traffic_ops/install/bin/postinstall.py", line 912, in setup_database_data
    if exec_psql(conn_str, tables_found_query) == "t":
  File "/opt/traffic_ops/install/bin/postinstall.py", line 877, in exec_psql
    proc = subprocess.run(cmd, capture_output=True, universal_newlines=True, check=False)
  File "/usr/lib64/python3.6/subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'capture_output'
```

Using this cfile, which works with the Perl `_postinstall` script, and testing on the [postinstall-python-centos8](/zrhoffman/trafficcontrol/commits/postinstall-python-centos8) branch:
```json
{
  "/opt/traffic_ops/app/conf/cdn.conf": [
    {
      "Generate a new secret?": "no",
      "config_var": "genSecret"
    },
    {
      "Number of secrets to keep?": "1",
      "config_var": "keepSecrets"
    },
    {
      "Number of workers?": "12",
      "config_var": "workers"
    },
    {
      "Traffic Ops url?": "https://trafficops.infra.ciab.test:443",
      "config_var": "base_url"
    }
  ],
  "/opt/traffic_ops/app/conf/ldap.conf": [
    {
      "Do you want to set up LDAP?": "no",
      "config_var": "setupLdap"
    }
  ],
  "/opt/traffic_ops/app/conf/production/database.conf": [
    {
      "Database type": "Pg",
      "config_var": "type"
    },
    {
      "Database name": "traffic_ops",
      "config_var": "dbname"
    },
    {
      "Database server hostname IP or FQDN": "db.infra.ciab.test",
      "config_var": "hostname"
    },
    {
      "Database port number": "5432",
      "config_var": "port"
    },
    {
      "Traffic Ops database user": "traffic_ops",
      "config_var": "user"
    },
    {
      "Traffic Ops database password": "twelve",
      "config_var": "password",
      "hidden": "1"
    }
  ],
  "/opt/traffic_ops/app/db/dbconf.yml": [
    {
      "Database server root (admin) username": "traffic_ops",
      "config_var": "pgUser"
    },
    {
      "Database server admin password": "twelve",
      "config_var": "pgPassword",
      "hidden": "1"
    },
    {
      "Download Maxmind Database?": "no",
      "config_var": "maxmind"
    }
  ],
  "/opt/traffic_ops/install/data/json/openssl_configuration.json": [
    {
      "Do you want to generate a certificate?": "no",
      "config_var": "genCert"
    }
  ],
  "/opt/traffic_ops/install/data/json/profiles.json": [
    {
      "Traffic Ops url": "https://trafficops.infra.ciab.test:443",
      "config_var": "tm.url"
    },
    {
      "Human-readable CDN Name.  (No whitespace, please)": "CDN-in-a-Box",
      "config_var": "cdn_name"
    },
    {
      "DNS sub-domain for which your CDN is authoritative": "mycdn.ciab.test",
      "config_var": "dns_subdomain"
    }
  ],
  "/opt/traffic_ops/install/data/json/users.json": [
    {
      "Administration username for Traffic Ops": "admin",
      "config_var": "tmAdminUser"
    },
    {
      "Password for the admin user": "twelve12",
      "config_var": "tmAdminPw",
      "hidden": "1"
    }
  ],
  "/opt/traffic_ops/install/data/profiles/": [
    {
      "Add custom profiles?": "no",
      "config_var": "custom_profiles"
    }
  ]
}
```

traffic_ops/install/bin/postinstall.py
```diff
@@ -0,0 +1,1247 @@
...
+
+It does, however, offer several more command-line flags not present in the original, to aid in
+testing.
+
+-a, --automatic               If there are questions in the config file which do not have answers,
+                              the script will look to the defaults for the answer. If the answer is
+                              not in the defaults the script will exit.
+--cfile [FILE]                An input config file used to ask and answer questions.
+--debug                       Enables verbose logging output.
+--defaults [FILE]             Writes out a configuration file with defaults which can be used as
```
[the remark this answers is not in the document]
ocket8888: I actually think POSIX is fine with that, it's GNU non-compliant. At any rate I'm not sure `argparse` can do that, because it parses strings of characters afters a `-` as separate short-form options as per GNU spec. I'll need to look into that.

ocket8888:
I think that's because the Perl script will silently fail to apply database configuration if no database is present, which is why the Python script now comes with a command-line option to skip database operations. Silent failures are a bug, IMO.

zrhoffman:
> I think that's because the Perl script will silently fail to apply database configuration if no database is present, which is why the Python script now comes with a command-line option to skip database operations. Silent failures are a bug, IMO.

But there *is* a database present, and the Perl script does succeed. Pushed 337785b0f0 to the [postinstall-python-centos8](/zrhoffman/trafficcontrol/commits/postinstall-python-centos8) branch, so now `trafficops-perl` initializes the database using the Perl `_postinstall` so you can observe it succeeding.

traffic_ops/install/bin/postinstall.py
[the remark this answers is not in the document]
ocket8888: [Looks like argparse does, in fact, support this](https://stackoverflow.com/questions/31127366/single-dash-for-argparse-long-options)

ocket8888:
Looks like the error stems from my use of `capture_output`, which was added in Python 3.7. Currently the oldest minor release still receiving security fixes is 3.6 - and that's what's in the CentOS8 `/usr/libexec/` - so we should definitely still support it. Fixed.

Also added support for a Python interpreter at `/usr/libexec/platform-python` - but not Python 2 support. Python 2 is dead and I'm no necromancer. If it would help adoption I would not be in any way opposed to making the script support Python 2 until CentOS 7's end-of-life either after this is merged or through a PR directly to my branch before it's merged, but it's work I don't want to do. I like my type hints, and part of the point of doing this rewrite was to eliminate dependencies outside of the standard library which supporting the two cleanly might require (like `future` or `six`) .

traffic_ops/install/bin/postinstall.py
```diff
@@ -0,0 +1,1344 @@
+#!/usr/bin/env -S sh -c 'PATH="${PATH}:/usr/libexec" exec "$(type -p python platform-python | head -n1)" "$0" "$@"'
```
zrhoffman, line 1: Testing it out, ^ needs a semicolon after the PATH addition in order to work, like

```shell
#!/usr/bin/env -S sh -c 'PATH="${PATH}:/usr/libexec"; exec "$(type -p python3 platform-python | head -n1)" "$0" "$@"'
```

However, apparently `env -S` is not supported on the GNU `env` that CentOS 7 uses (8.22), and since CentOS 8 goes EOL 2.5 years before CentOS 7 does, we should support CentOS 7's `env`. This would work:

```bash
#!/usr/bin/env bash
"exec" "bash" "-c" "PATH+=:/usr/libexec/ exec \$(type -p python3 platform-python | head -n1) \"$0\" $*"
```

zrhoffman, line 1: The Python postinstall script does not support Python 2.7, which is fine, but because it doesn't, it should search the patch for a `python3` explicitly, not `python`, else it would not work on CentOS 7.

zrhoffman (changes requested):

traffic_ops/install/bin/postinstall.py
ocket8888: Turning a Python script into a polyglot bash script seems convoluted. Can I not just use `#!/usr/bin/env python3` or `#!/bin/sh -c 'PATH="${PATH}:/usr/libexec"; exec "$(type -p python3 platform-python | head -n1)" "$0" "$@"'` ? Why does `env` need so much hand-holding? If `platform-python` isn't in the `PATH` should I be using it at all? [The RedHat Developer Blog](https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/) says no:

> _"If you are not re-building the distro, do not use Platform-Python directly. Install `python3` and use that."_

and later reiterates:

> _"Do not use platform-python for your applications. However, use platform-python if you are writing system/admin code for RHEL 8."_

zrhoffman: > Turning a Python script into a polyglot bash script seems convoluted. Can I not just use `#!/usr/bin/env python3` or `#!/bin/sh -c 'PATH="${PATH}:/usr/libexec"; exec "$(type -p python3 platform-python | head -n1)" "$0" "$@"'` ? Why does `env` need so much hand-holding?

It's not `env` that needs the handholding, it's shebang. Shebang does not split arguments by a *space* character (or by anything else), so in the snippet you posted above, `-c 'PATH="${PATH}:/usr/libexec"; exec "$(type -p python3 platform-python | head -n1)" "$0" "$@"'` is all treated as a single argument, which results in an error:

```shell
[root@732dff90b85f a]# ./a.py
/bin/sh: - : invalid option
Usage:  /bin/sh [GNU long option] [option] ...
        /bin/sh [GNU long option] [option] script-file ...
GNU long options:
        --debug
        --debugger
        --dump-po-strings
        --dump-strings
        --help
        --init-file
        --login
        --noediting
        --noprofile
        --norc
        --posix
        --rcfile
        --rpm-requires
        --restricted
        --verbose
        --version
Shell options:
        -ilrsD or -c command or -O shopt_option         (invocation only)
        -abefhkmnptuvxBCHP or -o option
```


> If `platform-python` isn't in the `PATH` should I be using it at all? [The RedHat Developer Blog](https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/) says no:
> 
> > _"If you are not re-building the distro, do not use Platform-Python directly. Install `python3` and use that."_
> 
> and later reiterates:
> 
> > _"Do not use platform-python for your applications. However, use platform-python if you are writing system/admin code for RHEL 8."_

Since there is no Python 2 support currently, yeah, might as well just use `#!/usr/bin/env python3` directly for now.

ocket8888: Done

ocket8888: > _"Since there is no Python 2 support currently, yeah, might as well just use `#!/usr/bin/env python3` directly for now."_

Then that's what I'll do

zrhoffman (approved):
Looks great

PR #4764, opened 2020-06-08
Made admin user compliant with restrictions (API and TP)

ocket8888:

Added an email and full name to the "admin" user created for CDN-in-a-Box, and changes its password to meet the minimum length requirement. Currently in CDN-in-a-Box, if you navigate to the "admin" user's user under "Users", you notice that it's missing the required "Full Name" field and the required "Email" field, and its password isn't long enough. This means that any small change you want to make to the user e.g. for testing purposes must minimally fix those three issues. After this PR, those would no longer be a problem.

Again, though, this _does change the CiaB "admin" user's password_. People using CiaB for testing _may_ be broken, but if done properly by sourcing `variables.env` and using `$TO_ADMIN_PASSWORD` it shouldn't make a difference.

- CDN in a Box

Build and run CDN-in-a-Box; everything should work normally. Try authenticating as `$TO_ADMIN_USER` using `$TO_ADMIN_PASSWORD` and verify that works. Check that user's information to verify that the email and full name are present.

- [x] CiaB has no tests
- [x] Interface doesn't change so docs are unchanged

### DIFF
--- a/infrastructure/cdn-in-a-box/traffic_ops/adduser.pl
+++ b/infrastructure/cdn-in-a-box/traffic_ops/adduser.pl
@@ -24,23 +24,27 @@ use strict;
 use Crypt::ScryptKDF qw{ scrypt_hash };
 
 if ($#ARGV < 2) {
-    die "Usage: $ARGV[0] <username> <password> <role>\n";
+    die "Usage: $ARGV[0] USERNAME PASSWORD [ROLE] [TENANT] [EMAIL] [FULL_NAME]\n";
 }
 
 my $username = shift // 'admin';
 my $password = shift or die "Password is required\n";
 my $role = shift // 'admin';
 my $tenant = shift // 'root';
+my $email = shift // 'admin@no-reply.trafficops.infra.ciab.test';
+my $full_name = shift // 'James Cole'
 
 # Skip the insert if the admin 'username' is already there.
 my $hashed_passwd = hash_pass( $password );
 print <<"EOSQL";
-INSERT INTO tm_user (username, role, local_passwd, confirm_local_passwd, tenant_id)
+INSERT INTO tm_user (username, role, local_passwd, confirm_local_passwd, tenant_id, email, full_name)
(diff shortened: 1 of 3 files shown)


infrastructure/cdn-in-a-box/traffic_ops/adduser.pl
```diff
@@ -24,23 +24,27 @@
...
 if ($#ARGV < 2) {
-    die "Usage: $ARGV[0] <username> <password> <role>\n";
+    die "Usage: $ARGV[0] USERNAME PASSWORD [ROLE] [TENANT] [EMAIL] [FULL_NAME]\n";
 }
 
 my $username = shift // 'admin';
 my $password = shift or die "Password is required\n";
 my $role = shift // 'admin';
 my $tenant = shift // 'root';
+my $email = shift // 'admin@no-reply.trafficops.infra.ciab.test';
```
zrhoffman, line 34: Might as well set `TO_EMAIL` to this value and use that, since `TO_EMAIL` is not currently used anywhere. Perl can be used to retrieve it.

infrastructure/cdn-in-a-box/traffic_ops/adduser.pl
```diff
@@ -24,23 +24,27 @@
...
+my $full_name = shift // 'James Cole'
```
zrhoffman, line 35: `James Cole` should be an environment variable, Perl can be used to retrieve it.

ocket8888: It could be, but that's following the pattern of allowing the script to add a generic user to the database while defaulting to the "admin" user's default information.

What about:
```perl
my $full_name = shift // $ENV{TO_ADMIN_FULL_NAME};
```
? And I'd wanna do the same thing with the email.

zrhoffman: Yeah that looks good.

PR #4765, opened 2020-06-08
Fix a couple TR NullPointerExceptions when no caches are available

rawlinp:

Fixes a Null Pointer Exception (where there are no available caches for a requested DS) and converts some `==` usage to `.equals()` (since I believe we want to compare the actual IP addresses and not the pointer references). Also, change the method for determining whether a given IP address is IPv4 or IPv6 to match what is used elsewhere in similar cases (since `getByName` can have the side-effect of doing DNS lookups).

- Traffic Router

Run the TR unit tests, verify they still pass:
```
mvn clean test -Djava.library.path=/usr/local/opt/tomcat-native/lib
```
Make a request for a DS that doesn't have any available caches, make sure that `editCacheListForIpVersion` does not throw a `NullPointerException`.

- master
- 4.1.0-RC1

- [x] Fairly minor fixes, performed manual tests to validate since there aren't pre-existing unit tests in this specific area (as far as I can tell)
- [x] Bugfix, docs unnecessary
- [x] Unreleased bug, changelog entry unnecessary

### DIFF
--- a/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Node.java
+++ b/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Node.java
@@ -194,24 +194,20 @@ public class Node extends DefaultHashable {
 		}
 
 		for (final InetRecord record : ipAddresses) {
-			if (record.getAddress() == ip4 && !ipv4Available) {
+			if (record.getAddress().equals(ip4) && !ipv4Available) {
 				newlyUnavailable.add(record);
-				LOGGER.debug("disabling IPv4 address " + record.getAddress());
 			}
-			if (record.getAddress() == ip6 && !ipv6Available) {
+			if (record.getAddress().equals(ip6) && !ipv6Available) {
 				newlyUnavailable.add(record);
-				LOGGER.debug("disabling IPv6 address " + record.getAddress());
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


rawlinp:
I rebased to resolve merge conflicts, but I didn't actually have to change any of the code. Cache.java was basically just renamed to Node.java without having to change anything else.

zrhoffman:
> I rebased to resolve merge conflicts, but I didn't actually have to change any of the code. Cache.java was basically just renamed to Node.java without having to change anything else.

I think the intent is to still use Cache because
* Although Cache was renamed to Node, a new class also named Cache was added in the Edge Traffic Routing PR, and it [extends Node](/apache/trafficcontrol/blob/aa03fd37f8/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Cache.java#L30).
* The usage of  Cache [doesn't change](/apache/trafficcontrol/blob/aa03fd37f8/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/config/ConfigHandler.java#L356) in ConfigHandler after the Edge Traffic Routing PR was merged
* Node [looks like](/apache/trafficcontrol/blob/aa03fd37f8/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/config/ConfigHandler.java#L929) it is used for Edge Traffic Routing

PR #4747, opened 2020-06-01
Tp interface servers

ocket8888:
## What does this PR (Pull Request) do?
- [x] This PR fixes #4623 and fixes #4792 

Updates Traffic Portal tables and forms to account for servers having multiple interfaces, including:

- any table that shows servers obtained from `/servers` such as
    - Servers
    - Cache Groups -> View Servers
    - Profiles -> View Servers
    - Delivery Services -> View Servers
- The server create/edit form

## Which Traffic Control components are affected by this PR?
- Traffic Portal

## What is the best way to verify this PR?
- Build Traffic Portal, verify that all the mentioned tables and forms work properly.
- Run the TP integration tests.
- You could also run the UI tests, but I think they're broken on master atm _[citation needed]_, so I'll need to look into that.
## The following criteria are ALL met by this PR
- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR ensures that database migration sequence is correct OR this PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY**

Issue #1801, opened 2018-01-29
ILO server field opens SSH instead of HTTPS link


smalenfant:
In Traffic OPs, the link for ILO is https although looks like the new UI is now sending us to an SSH terminal screen.

Marked as regression due to change in functionality. This feature was added as part of #1399.

Would be nice if you we can have multiple choice on the click like "ssh, https, copy link". Context menu?

jagooch:
+1 this feature. Implementation could be clickable icons near the IP address. 
- A computer terminal icon with an  SSH protocol link to the ILO. 
- A Web icon with an https link, properly formatted for ipv6 https connection to ILO. 


The feature could provide the auto-population of the ILO username/password fields using credentials stored in TODB . An access check would be necessary to control access to the credentials. 

The addition of credentials would enable 1-click login to the ILO Web interface.

rawlinp:
ILO server field link is now https not ssh.

PR #4769, opened 2020-06-09
Servers table switch to ag-grid

### LINKED ISSUE #2326: TP: [RFE] addition of "NOT" (!) operator in search fields
Today, we have the ability to search using `^` and `$` characters. As a user of Traffic Portal, it would also be useful to be able to negate search strings using `!`.

### LINKED ISSUE #2624: Improve TP search query
I'm primarily concerned with the server list, but similar considerations would apply elsewhere.

I'd like the ability to express more complex statements in the TP search box.  Currently you get a substring whitelist across all visible fields.  I'd like to be able to express things like NOT, AND, OR, and groups.  As an additional consideration, being able to supply an optional attribute to each search tuple to specify on what fields it should apply would be nice.  An over the top consideration would be regex support.

### LINKED ISSUE #4354: Improve performance (pagination and search) of TP's tables

ocket8888:
- [x] This PR fixes #4354 and fixes #2624 and fixes #2326

Switches the "/servers" table from jQuery datatables to agGrid, increasing load speed by a factor of about 14.

The new tables have a superset of the functionality of the old tables.

- Traffic Portal

Run Traffic Portal - preferably with a whole ton of servers so you can be impressed by the lightning fast load speed (at least compared to the jQuery data tables).

If #4747 isn't merged yet then you'll need to be using a version of TO that doesn't include the changes from #4700 .

### DIFF
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -71,45 +71,38 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 			headerName: "Cache Group",
 			field: "cachegroup",
 			hide: false,
-			searchable: true
 		},
 		{
 			headerName: "CDN",
 			field: "cdnName",
 			hide: false,
-			searchable: true
 		},
 		{
 			headerName: "Domain",
 			field: "domainName",
 			hide: false,
-			searchable: true
 		},
 		{
 			headerName: "Host",
 			field: "hostName",
 			hide: false,
-			searchable: true
 		},
 		{
 			headerName: "HTTPS Port",
 			field: "httpsPort",
 			hide: true,
-			searchable: false,
 			filter: "agNumberColumnFilter"
 		},
 		{
 			headerName: "ID",
 			field: "id",
 			hide: true,
-			searchable: true,
 			filter: "agNumberColumnFilter"
 		},
 		{
 			headerName: "ILO IP Address",
 			field: "iloIpAddress",
 			hide: true,
-			searchable: true,
 			cellRenderer: "sshCellRenderer",
 			onCellClicked: null
 		},
@@ -117,7 +110,6 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 			headerName: "ILO IP Gateway",
 			field: "iloIpGateway",
 			hide: true,
-			searchable: false,
 			cellRenderer: "sshCellRenderer",
 			onCellClicked: null
 		},
@@ -125,37 +117,31 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 			headerName: "ILO IP Netmask",
 			field: "iloIpNetmask",
 			hide: true,
-			searchable: false
 		},
 		{
 			headerName: "ILO Username",
 			field: "iloUsername",
 			hide: true,
-			searchable: false
 		},
 		{
 			headerName: "Interface Name",
 			field: "interfaceName",
 			hide: true,
-			searchable: false
 		},
 		{
 			headerName: "IPv6 Address",
 			field: "ipv6Address",
 			hide: false,
-			searchable: true
 		},
 		{
 			headerName: "IPv6 Gateway",
 			field: "ipv6Gateway",
 			hide: true,
-			searchable: false
 		},
 		{
 			headerName: "Last Updated",
 			field: "lastUpdated",
 			hide: true,
-			searchable: false,
 			filter: "agDateColumnFilter",
 			valueFormatter: dateCellFormatter
 		},
@@ -163,13 +149,11 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 			headerName: "Mgmt IP Address",
 			field: "mgmtIpAddress",
 			hide: true,
-			searchable: false
 		},
 		{
 			headerName: "Mgmt IP Gateway",
 			field: "mgmtIpGateway",
 			hide: true,
-			searchable: false,
 			filter: true,
 			cellRenderer: "sshCellRenderer",
 			onCellClicked: null
@@ -178,7 +162,6 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 			headerName: "Mgmt IP Netmask",
 			field: "mgmtIpNetmask",
 			hide: true,
-			searchable: false,
 			filter: true,
 			cellRenderer: "sshCellRenderer",
 			onCellClicked: null
@@ -187,7 +170,6 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 			headerName: "Network Gateway",
 			field: "ipGateway",
 			hide: true,
-			searchable: true,
 			filter: true,
 			cellRenderer: "sshCellRenderer",
 			onCellClicked: null
@@ -196,7 +178,6 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 			headerName: "Network IP",
 			field: "ipAddress",
 			hide: false,
-			searchable: true,
 			filter: true,
 			cellRenderer: "sshCellRenderer",
 			onCellClicked: null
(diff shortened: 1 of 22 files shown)


traffic_portal/grunt/browserify2.js
```diff
@@ -89,7 +94,7 @@ module.exports = {
                         cwd: '<%= globalConfig.app %>/bower_components/',
                         src:
                             [
-                                'angular/angular.js',
+                                // 'angular/angular.js',
```
mitchell852, line 97: you want to put this back and we can do this in another PR?

traffic_portal/grunt/copy.js
```diff
@@ -42,6 +42,13 @@ module.exports = {
                     'traffic_portal_release.json',
                     'traffic_portal_properties.json'
                 ]
+            },
+            {
+                expand: true,
```
mitchell852, line 47: i don't _think_ this is necessary

traffic_portal/grunt/copy.js
```diff
@@ -78,6 +85,13 @@ module.exports = {
...
```
mitchell852, line 90: nor this

traffic_portal/grunt/browserify2.js
ocket8888: Yeah, that wasn't meant to be in there.

traffic_portal/grunt/copy.js
ocket8888: It is, unfortunately.

mitchell852:
image

mitchell852:
image

Not sure why "Clear Updates" is disabled.

ocket8888:
"Clear Server Updates" is _meant_ to be disabled when the server isn't a cache server and/or doesn't have pending updates. "Queue Server Updates" is meant to be disabled when the server isn't a cache server and/or _does_ have pending updates. Just to be clear about the intended behavior, because the old context menu would hide those options instead of disabling them.

Clearly, though, that's a cache server, and it _looks_ like it has updates pending. So something's probably wrong.

mitchell852:
> Clearly, though, that's a cache server, and it _looks_ like it has updates pending. So something's probably wrong.

yeah, something is weird there. it's a cache server and updPending=true so "clear updates" should be an option. i looked at the logic and it seemed right so not sure:

```
            <button type="button" ng-click="clearServerUpdates(server, $event)" ng-disabled="!isCache(server) || !server.updPending">Clear Server Updates</button>
```

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -17,405 +17,634 @@
...
-        });
-    });
+var TableServersController = function(servers, $scope, $state, $uibModal, $window, dateUtils, locationUtils, serverUtils, cdnService, serverService, statusService, propertiesModel, messageModel, userModel, $document) {
+	/**** Table cell formatters/renderers ****/
 
+	// browserify can't handle classes...
+	function SSHCellRenderer() {}
+	SSHCellRenderer.prototype.init = function(params) {
+		this.eGui = document.createElement("A");
+		this.eGui.href = "ssh://" + userModel.user.username + "@" + params.value;
```
mitchell852, line 27: not sure if you want to / can address this as part of this PR:

https://github.com/apache/trafficcontrol/issues/1801

ocket8888: I could - and considered it - but I don't want to. For this PR I just want to make sure the new table(s) has/have all the same functionality as was expected of the previous version.

mitchell852:
Maybe in order to get this merged sooner rather than later. just focus on these 3 things i found:

1. cdn column is empty
2. clear updates context menu is disabled
3. ensure that this is not a problem:

image

ocket8888:
"cdn column is empty" is already fixed. Might not have pushed it.
"Failed to load stored column state: one or more columns not found" is also not a real problem. I think that's actually related to the cdns thing. It saved column state with a "cdn" column, and the fix was to make it a "cdnName" column. So you need to clear your local storage. Nobody who didn't use it while it was still in a pull request should see that error.

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -17,405 +17,633 @@
...
+			headerName: "Network Subnet",
+			field: "ipNetmask",
+			hide: true,
+			searchable: false
+		},
+		{
+			headerName: "Offline Reason",
+			field: "offlineReason",
+			hide: true,
+			searchable: false
```
mitchell852: why does it say some are not searchable? looks like they are are all searchable once they are visible.

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -17,405 +17,633 @@
...
+			this.eGui.classList.add("fa-check");
+		}
+	}
+	UpdateCellRenderer.prototype.getGui = function() {return this.eGui;};
+
+	/**
+	 * Gets text with which to file a status tooltip.
+	 * @returns {string | undefined} The offline reason if the server is offline, otherwise nothing.
+	 */
+	function offlineReasonTooltip(params) {
```
mitchell852, line 51: does this work? i can't get it to when i hover over status and tbh not even really sure it's needed since there's an offline reason column anyhow.

mitchell852:
nit: the context menu doesn't close when you select 'open in new tab" or "navigate to fqdn" or "show charts". basically anything that opens a new window/tab

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
ocket8888: those don't belong there. I actually already got rid of those properties by the time you made this review :P

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
ocket8888: It only shows up when the server is set to an "offline" status, e.g. OFFLINE. Works for me. I don't really think it needs to be there either, but the jQuery datatables did it, so I did too.

mitchell852: yep, that makes sense

ocket8888:
That should be fixed now

traffic_portal/package.json
```diff
@@ -30,5 +30,8 @@
     "requirejs": "2.3.6",
     "ssl-root-cas": "1.3.1",
     "time-grunt": "1.3.0"
+  },
+  "dependencies": {
+    "ag-grid-community": "^23.2.0"
```
mitchell852, line 35: i guess i'm less risk tolerant but i'd do `23.2.0` or `~23.2.0` rather than letting it roll ahead to new minor versions, but your call.

ocket8888: That's just the default behavior when you `npm install --save` something.

I don't see a problem with letting it roll ahead to new minor versions; the API won't ever break so it should only come with more benefits. Besides, pinning these things and never coming back is how we wind up on unsupported releases for years.

PR #4788, opened 2020-06-12
Removed unused dependencies, migrated angular from bower to npm

ocket8888:

Removes dependencies on `es5-shim` and `json3`, which are both shims for features supported by major browsers for years. Also moves `angular`, `angular-jwt`, `angular-animate`, and `angular-bootstrap-contextmenu` from bower management into NPM management. Others are probably possible to move as well, but I thought that was enough for one PR.

- Traffic Portal

Build and run Traffic Portal. Observe that angular's still working.

- [x] No functional changes, no tests necessary
- [x] No functional changes, no documentation necessary

### DIFF
--- a/traffic_portal/bower.json
+++ b/traffic_portal/bower.json
@@ -1,8 +1,6 @@
 {
     "name": "traffic_portal",
     "dependencies": {
-        "angular": "1.7.8",
-        "angular-animate": "1.7.8",
         "angular-resource": "1.7.8",
         "angular-route": "1.7.8",
         "angular-sanitize": "1.7.8",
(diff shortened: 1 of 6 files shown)


traffic_portal/bower.json
```diff
@@ -1,23 +1,17 @@
 {
     "name": "traffic_portal",
     "dependencies": {
-        "angular": "1.7.8",
-        "angular-animate": "1.7.8",
         "angular-resource": "1.7.8",
```
mitchell852, line 4: by leaving these other angular ones in bower.json, its still pulling angular in via `bower install`

```
angular-bootstrap#0.14.3 app/bower_components/angular-bootstrap
└── angular#1.7.8
```

so i would suggest either a) limiting this PR to just removing unused deps (es5-shim and json3) or b) moving ALL of the angular* deps to npm.

ocket8888: Not all of "the angular* deps" can be moved over so trivially. Particularly, the bootstrap stuff doesn't load properly when included with browserify from npm. I don't know why, but it might be related to the fact that the version we use from bower is unavailable in npm. Also note, though, that neither available versions are supported, neither package manager offers supported versions, and the version we use is two major versions behind what's current.

With all that in mind I was hoping to just remove angular-bootstrap* entirely, especially if it's only used for the table context menus, since the new ag-grid tables won't use angular-bootstrap-contextmenu.

ocket8888:
I think the differences between this branch and master have drifted to the point where if I really wanted to do this - and at this point I'd rather give up in favor of replacing TP altogether - I'd want to start from scratch.

PR #4797, opened 2020-06-15
Fixed filtering query parameters in the /servers API

### LINKED ISSUE #4795: Fix query parameter for servers API
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

1. When user perform a /GET `https://{{TO_BASE_URL}}/api/3.0/servers?type=MID` with any type like MID,EDGE ...
The API will return 500 Internal Server Error.
```
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```
2. When user perform a /GET `https://{{TO_BASE_URL}}/api/3.0/servers?status=ONLINE` with any valid or invalid status
The API will return 500 Internal Server Error
```
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```
3. When user perform a /GET `https://{{TO_BASE_URL}}/api/3.0/servers?parentCachegroup=1` with valid or invalid parentCachegroup number.
The API will return 500 Internal Server Error
```
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```
## Expected / new behavior:

Should return the servers with correct match

ocket8888:
- [x] This PR fixes #4795

Fixes the filtering query parameters of `/servers`.

- Traffic Ops

Run the API/client integration tests

Make requests that filter servers by type, cdn, cachegroup, profile, physical location, and/or status.

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v3/servers_test.go
+++ b/traffic_ops/testing/api/v3/servers_test.go
@@ -144,6 +144,23 @@ func GetTestServersQueryParameters(t *testing.T) {
 		}
 		params.Del("profileId")
 	}
+
+	cgs, _, err := TOSession.GetCacheGroupsNullable()
+	if err != nil {
+		t.Fatalf("Failed to get Cache Groups: %v", err)
+	}
+	if len(cgs) < 1 {
+		t.Fatal("Failed to get at least one Cache Group")
+	}
+	if cgs[0].ID == nil {
+		t.Fatal("Cache Group found with no ID")
+	}
+
+	params.Add("parentCacheGroup", strconv.Itoa(*cgs[0].ID))
+	if _, _, err = TOSession.GetServers(&params); err != nil {
+		t.Errorf("Error getting servers by parentCacheGroup: %v", err)
+	}
+	params.Del("parentCacheGroup")
 }
 
 func UpdateTestServers(t *testing.T) {
--- a/traffic_ops/testing/api/v3/servers_test.go
+++ b/traffic_ops/testing/api/v3/servers_test.go
@@ -77,10 +77,6 @@ func GetTestServersDetails(t *testing.T) {
 }
 
 func GetTestServersQueryParameters(t *testing.T) {
-	if len(testData.DeliveryServices) < 1 {
-		t.Fatal("Need at least one server to test getting servers with query parameters")
-	}
-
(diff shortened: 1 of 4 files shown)


ocket8888:
Because finding a cache server in the test data with a parentCacheGroup is non-trivial, since it's not a server property, and dsId is currently broken for other reasons. A separate PR that fixes issues with dsId is adding a test for it (#4798).

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -51,6 +51,12 @@ import (
 const serverCountQuery = `
 SELECT COUNT(s.id)
 FROM server AS s
+JOIN cachegroup cg ON s.cachegroup = cg.id
```
rawlinp, line 54: Since this is the exact same `FROM ... JOIN`s as the main `selectQuery` below, should we move this into a constant and share them between both queries?

ocket8888: I don't really have an opinion, but I suppose it could help avoid something like this in the future. That's a good enough reason to do it.

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -101,6 +97,70 @@ func GetTestServersQueryParameters(t *testing.T) {
...
+	}
+
+	if len(resp.Response) < 1 {
+		t.Fatalf("Failed to get at least one server")
+	}
+
+	s := resp.Response[0]
+
+	params.Add("type", s.Type)
+	if _, _, err := TOSession.GetServers(&params); err != nil {
```
rawlinp, line 113: nit: I'm going to merge this anyways, but if you really want to level up this testing you could iterate over the servers and verify that the response only contains the properly filtered servers

ocket8888: ... I know...

PR #4799, opened 2020-06-16
Add grunt task to ensure ruby/compass is installed for TP.

shamrickus:

A grunt task was added to the build-css-dev task that will check if `ruby` and `compass` are installed using `which`.

It also adds a Gemfile for use with the ruby bundler, aptly named Bundler. While this _can_ replace the manual gem installs we use, it _cannot_ be used for production because Bundler requires ruby 2.3.0 and CentOS7 only has 2.0.0.

- CDN in a Box
- Documentation
- Traffic Portal

Verify CDN in a box still builds, and running grunt (in dev) should check if ruby/compass is installed. 
Ensure documentation updates make sense.
These changes are just QOL for devs on TP, as such output code (regardless of prod/dev) should be **unmodified**


mitchell852:
@shamrickus - so i see this now in the console output:

```
Running "run:which_ruby" (run) task
/usr/bin/ruby
/usr/local/bin/compass
```

but if i don't have ruby and/or compass installed, what then? is the dev supposed to call `gem install` or something (sorry, i'm rusty on my ruby)

PR #4801, opened 2020-06-17
Remove API 1.x configfile routes

ocket8888:

This PR removes all server configuration file generation routes from the Traffic Ops API. It also removes the corresponding documentation, Traffic Portal functionality and, of course, the client/API integration tests.

The functionality has not been removed from the v1-client, it's just now untested.

- Documentation
- Traffic Ops Client (Go)
- Traffic Ops
- Traffic Portal

Run all the unit and client/API integration tests. You can also try requesting configuration files and notice that they're served by Perl (if it exists).

- [x] This PR removes tests
- [x] This PR removes documentation


mitchell852:
@ocket8888 - sorry, conflicts.

mitchell852:
going to hold off on merging this until after TC 5.0 branch is cut.

mitchell852:
As discussed, want to change these to simply returning `410 Gone`? And removing any use from TP is fine as is.

ocket8888:
> _"...want to change these to simply returning 410 Gone?"_

Done.

CHANGELOG.md
```diff
@@ -67,10 +67,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Removed
 - Removed deprecated Traffic Ops Go Client methods.
-- Configuration generation logic in the TO API (v1) for:
-  - `ip_allow.config`
-  - `parent.config`
-  - `remap.config`
+- Configuration generation logic in the TO API (v1) for all files and the "meta" route.
```
mitchell852, line 70: can you elaborate on this a bit? i think it's important to make very clear that ORT version whatever that uses these TO api endpoints can no longer be used. instead ORT version whatever that leverages atstccfg needs to be used in conjunction with the current version of TO. I would hate for somebody to upgrade TO to 5.0 and wonder why all their caches are now getting a 410 rather than the ats config files.

ocket8888:
Probably all of those tests could be removed. They have never passed for as long as I've been working on the project.

But also, no. The functionality of the Perl server is unchanged, so there's no reason to destroy the tests. Except to get rid of all of them.

mitchell852:
> Probably all of those tests could be removed. They have never passed for as long as I've been working on the project.
> 
> But also, no. The functionality of the Perl server is unchanged, so there's no reason to destroy the tests. Except to get rid of all of them.

makes sense. what about the genConfigRoutes.py file? won't really work anymore right?

mitchell852:
should this have a different example? 

https://github.com/apache/trafficcontrol/blob/master/traffic_ops_ort/atstccfg/atstccfg.go#L22

ocket8888:
genConfigRoutes.py also generates snapshot routes. It should probably be removed - along with the `compare` tool - just because it doesn't seem to have much value, especially now that the re-write is (nearly) complete. But I don't think the reason for its removal is a necessary consequence of the removal of the API configuration file routes.

mitchell852:
> genConfigRoutes.py also generates snapshot routes. It should probably be removed - along with the `compare` tool - just because it doesn't seem to have much value, especially now that the re-write is (nearly) complete. But I don't think the reason for its removal is a necessary consequence of the removal of the API configuration file routes.

well i _thought_ the compare tool might actually have some value as it can be used to simply compare 2.x/foos and 3.x/foos, right? but any reference to configfiles/ats should be entirely scrubbed w/ this PR imo. including genConfigRoutes.py it would seem.

ocket8888:
> ...it can be used to simply compare 2.x/foos and 3.x/foos, right?

Sure, but you'd expect those to be possibly very different. The purpose of `compare` was to highlight differences in API responses for the same endpoint in the *same* version of the API, to make sure things were properly translated from Perl to Go. When `compare` finds differences, it tells you what route had a difference, and dumps payloads to two files. That means for the purposes of examining differences between different routes it's not really any better than `diff <(toget myroute) <(toget -a 3.0 myroute)`. In fact, it's worse, because it's an extra step before doing the same diff.

mitchell852:
@ocket8888 - this all looks good. but this example bothers me. what do you think?

https://github.com/apache/trafficcontrol/blob/master/traffic_ops_ort/atstccfg/atstccfg.go#L22

ocket8888:
It is confusing, but accurate. That's how you have to use it, and changing that would involve changing not only how atstccfg works, but how the old `traffic_ops_ort.pl` script works. And last time I had to do that, I rewrote it in Python instead. So that probably makes my feelings on the matter pretty clear.

mitchell852:
@ocket8888 - i'm getting this api test failure in TestFederations

`userdeliveryservices_test.go:175: get user delivery services expected >0 actual 0`

ocket8888:
I can't reproduce that test failure:
image

mitchell852:
> I can't reproduce that test failure:
> image

ok, seems ok for me now too

```
$ docker-compose -f docker-compose.traffic-ops-test.yml up --build integration
integration_1  | + '[' 0 -eq 0 ']'
integration_1  | + '[' 0 -eq 0 ']'
integration_1  | + '[' 0 -eq 0 ']'
integration_1  | + echo 'TO API tests success'TO API tests success
```

PR #4802, opened 2020-06-17
Replace `json_build_object` from multi-interface implementations

shamrickus:

Several PRs in the pursuit of implementing multi-interfaces used `json_build_object` in their queries. This is **significantly** slower than building the objects on the go side; as such, this PR replaces all of those occurrences.

- Traffic Ops

Run the Unit/API tests for TO.
Ensure `servers/details`, `deliveryservices/{{id}//servers`, `deliveryservices/{{id}}/servers/eligible` all work with legacy data and multi-interface data.


traffic_ops/traffic_ops_golang/dbhelpers/db_helpers_test.go
```diff
@@ -138,7 +146,133 @@ func TestGetCacheGroupByName(t *testing.T) {
 			}
 		})
 	}
+}
+
+// CreateServerInterfaces takes in a cache id and creates the interfaces/ipaddresses for it
```
srijeet0406, line 151: Nit: should not be a public function in the comment

PR #4805, opened 2020-06-17
Updated list of releases receiving security updates

ocket8888:

Updates the list of ATC releases that are receiving security updates.

None

Read it

- [x] Tests are unnecessary

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -8,7 +8,7 @@
 | 2.2.x   | :x:                |
 | 3.0.x   | :x:                |
 | 3.1.x   | :white_check_mark: |
-| 4.0.x   | :white_check_mark: |
+| 4.0.x   | :x:                |
 | 4.1.x   | :white_check_mark: |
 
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -5,8 +5,11 @@
 | Version | Supported          |
 | ------- | ------------------ |
 | < 2.2   | :x:                |
-| 2.2.x   | :white_check_mark: |
-| 3.0.x   | :white_check_mark: |
+| 2.2.x   | :x:                |
+| 3.0.x   | :x:                |
+| 3.1.x   | :white_check_mark: |
+| 4.0.x   | :white_check_mark: |
```
rawlinp, line 11: I believe now that 4.1.x is released we won't support security fixes to 4.0.x, since I think our policy is to support the latest minor/patch version of the latest two major versions.

dneuman64: +1

PR #4807, opened 2020-06-19
Add new header_rewrite fields for topology-based DSes

rawlinp:
Add the new `firstHeaderRewrite`, `innerHeaderRewrite`, and `lastHeaderRewrite` fields to delivery services. These fields can only be used when the delivery service is assigned to a topology and are mutually exclusive with `edgeHeaderRewrite` and `midHeaderRewrite`, which can **not** be used if the delivery service is assigned to a topology.

**NOTE:** this PR intentionally does *not* create and link "location" parameters for these new header rewrite fields, because this extra information is unnecessary for client-side config generation via `atstccfg` and would just lead to a load of unnecessary complexity. Additionally, I think we should get rid of "location" parameters altogether, so increasing and perpetuating their usage is not ideal.

- Traffic Ops

Run the unit tests and TO API tests, verify they pass. Build the docs, verify nothing broke. View the rendered documentation changes, make sure they look correct.


ocket8888:
When you say that the new fields are "mutually exclusive" with the old ones, do you mean that only one set can be non-null/empty at a time (and presumably must be the topologies ones if it has an assigned topology and vice-versa), or just that only one set will have any effect at one time?

rawlinp:
I believe this should answer that question: https://github.com/apache/trafficcontrol/pull/4807/files#diff-e35176dda777cf2b75d030e86b396b1aR486

PR #4827, opened 2020-06-24
TP: adds first, inner and last header rewrite fields for topologies

mitchell852:
This PR adds 3 new fields to the DNS* and HTTP* delivery service forms:

- firstHeaderRewrite
- innerHeaderRewrite
- lastHeaderRewrite

Validation was also added to:

a. only allow firstHeaderRewrite, innerHeaderRewrite and lastHeaderRewrite values if topology is selected for a ds
b. prevent edgeHeaderRewrite and midHeaderRewrite values if a topology is selected.

All 5 values are displayed to facilitate migration from edge/mid header rewrites to first/inner/last header rewrites when moving from explicit ds / server assignments towards using a topology instead.

- Traffic Portal

- Ensure that at least one topology exists at https://tp.domain.net/#!/topologies
- create/update a DNS* ds
- add values for edge/mid/first/inner/last header rewrites
- toggle between a topology and no topology assigned and notice the validation of those fields.
- ensure edge/mid header rewrites are not allowed when a topology is assigned
- ensure first/inner/last header rewrites are not allowed when NO topology is assigned
- do the same for an HTTP* ds

- [x] This PR does NOT include tests as these fields are optional and only required fields are currently under test.
- [x] This PR does NOT includes documentation changes as each ds form is not documented.


mitchell852:
> How would you feel about hiding the edge/mid header rewrite fields if they're empty and a topology is assigned? And hiding the first/inner/last header rewrite fields if they're empty and a topology isn't assigned?

so i thought about this a lot actually and it gets tricky, stuff starts showing/hiding when you assign/unassign a topology and/or add/remove header rewrite rules so i came to the conclusion that maybe it's just best to show all 5 fields with an validation message that displays if you are using the wrong combination of fields.

mattjackson220:
looks good to me! tested with HTTP and DNS, with topology the first, inner, and last fields work and the edge and mid fields display an error, fields are saved as it is switched between topology and not, without topology first, inner, and last fields display error and edge and mid fields work as expected. works for both new and updating DSes

PR #4830, opened 2020-06-24
Traffic Monitor excessive log errors fix

### LINKED ISSUE #4825: TM error log spam for "incomparable stat type int"
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [x] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

This error message is currently getting spammed in the TM logs:
```
ERROR: stat.go:292: 2020-06-24T15:22:54.408932036Z: Adding result from mycache-01: some stats could not be added: cannot add stat system.inf.speed: incomparable stat type int; cannot add stat system.configReloads: incomparable stat type int; cannot add stat system.lastReload: incomparable stat type int; cannot add stat system.astatsLoad: incomparable stat type int; cannot add stat system.configReloadRequests: incomparable stat type int; cannot add stat system.lastReloadRequest: incomparable stat type int
```
It seems to be new as of TM 4.1.0 and is unclear what the actual effects of the bug are.

## Expected / new behavior:

This error message should be investigated and fixed.

Start up TM (master or 4.1.0) in your environment and wait for it to start polling caches. This error message appears to be printed for every cache polled.

## Anything else:

From @zrhoffman:

> issue is at line 193 in /trafficcontrol/traffic_monitor/threadsafe/resultstathistory.go
it won't let you compare history[0].Val and stat as int type even though they are both ints

ocket8888:
- [x] This PR fixes #4825

- Traffic Monitor

Run Traffic Monitor with some caches to poll, observe no error spam in the logs.

- master
- 4.1

### DIFF
--- a/traffic_monitor/cache/astats.go
+++ b/traffic_monitor/cache/astats.go
@@ -38,7 +38,7 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/traffic_monitor/dsdata"
 	"github.com/apache/trafficcontrol/traffic_monitor/todata"
-	"github.com/json-iterator/go"
+	jsoniter "github.com/json-iterator/go"
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


traffic_monitor/cache/astats.go
```diff
@@ -38,7 +38,7 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/traffic_monitor/dsdata"
 	"github.com/apache/trafficcontrol/traffic_monitor/todata"
-	"github.com/json-iterator/go"
+	jsoniter "github.com/json-iterator/go"
```
rawlinp, line 41: Why is this necessary?

ocket8888: It wasn't, I didn't do it - at least not intentionally. I installed an IDE to help me wade through the confusing naming schemes of TM code, and I'd bet it did that for me automatically.

I can change it back if you want

rawlinp: Ok, it doesn't seem to be affecting anything but it doesn't appear to be used either, which is weird because Go won't let you compile with unused imports. I guess it must be `init`ing something.

PR #4833, opened 2020-06-25
RGB - Lookup Zip Code based on Lat/Long if Zip Code is null

### LINKED ISSUE #4372: RGB - Lookup Zip Code based on Lat/Long if Zip Code is null
## I'm submitting a ...
- [ ] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [x] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Currently when RGB is enabled, a client will be Geo Blocked if their IP does not have a zip code.  Sometime, especially with IP addresses, clients who should be allowed are not.
## Expected / new behavior:
If a client IP does not have a Zip code but does have a lat/long then we should use the lat/long to look up the zip code and then apply RGB logic. 

srijeet0406:

- [x] This PR fixes #4372 

- CDN in a Box
- Documentation
- Traffic Router

Make sure all the unit tests and API tests pass.

Add a line in your regional_geoblock config file under the `geoLocation` key, which has the allowed coordinate ranges like this:

`"geoLocation": {
                "excludePostalCode": [
                    "V5G",
                    "M7A"
                ],
                "coordinateRange": [{"minLat" : -12, "maxLat": 13, "minLon" : 55, "maxLon": 56}, {"minLat" : -13, "maxLat": 14, "minLon" : 55, "maxLon": 56}]
            }`

For a TR request, try spoofing the IP address of the client, such that the client doesn't have a postal code associated with it.
Make sure the min/max lat/long values are valid, and that the user's lat/long (which you get by spoofing the IP) falls within one of these ranges.
This IP should NOT be blocked by RGB.
Now, try to change the coordinate ranges in the config file, such that the request's lat/long do NOT fall in these ranges.
Make the request again.
This time, the IP should be blocked by RGB.

master

- [x] This PR includes tests 


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/loc/RegionalGeo.java
```diff
@@ -219,6 +251,11 @@ private static RegionalGeo parseConfigJson(final JsonNode json) {
                     LOGGER.error("RegionalGeo ERR: geoLocation empty");
                     return null;
                 }
+                // coordinate range
+                final List<RegionalGeoCoordinateRange> coordinateRanges = parseLocationJsonCoordinateRange(locationJson);
+                if (coordinateRanges != null) {
+                    regionalGeo.setCoordinateRanges(coordinateRanges);
```
mattjackson220, line 257: why do we set this here? doesnt look like we ever use the class level variable so maybe we can get rid of it and the getter/setter and this if statement?

srijeet0406: You're correct, I removed it, along with the if statement, and the getters and setters.

PR #4834, opened 2020-06-25
fixed TR NPE where ipAddresses gets set to null

mattjackson220:

This PR fixes an NPE in Traffic Router where an API call can set the ipAddresses to null resulting in a lot of NPEs

- Traffic Router

Verify that TR tests run successfully.
Verify that TR run successfully and when a call to the TR API at /crs/locations/{location}/caches is run, it does not set the ipAddresses to null if there are no A or AAAA records for that cache
Verify that TR successfully gets valid ipAddresses from CrConfig

--master
--4.1

this is a bug fix so it does not have any tests, CHANGELONG updates, or doc updates

### DIFF
--- a/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Node.java
+++ b/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Node.java
@@ -83,12 +83,12 @@ public class Node extends DefaultHashable {
 		return id;
 	}
 
-	public List<InetRecord> getIpAddresses(final JsonNode ttls, final Resolver resolver) {
-		return getIpAddresses(ttls, resolver, true);
+	public List<InetRecord> getIpAddresses(final JsonNode ttls) {
+		return getIpAddresses(ttls, true);
 	}
 
 	@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.NPathComplexity"})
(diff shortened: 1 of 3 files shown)


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Node.java
```diff
@@ -89,9 +89,6 @@ public String getId() {
 
 	@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.NPathComplexity"})
 	public List<InetRecord> getIpAddresses(final JsonNode ttls, final Resolver resolver, final boolean ip6RoutingEnabled) {
```
rawlinp, line 91: I think this is good.

A) we should never allow an external API to change `ipAddresses`.
B) I don't think we ever really have a good reason to use a `Resolver` (like `ZoneManager` here, which will default to using the system resolver if the fqdn is outside of TR's zone).

However, I now notice that `resolver` is unused in this method, so we should remove that parameter and fix all the places this method is called to no longer pass in `ZoneManager` as a resolver.

zrhoffman:
TR unit and integration tests pass, works fine with `resolver` removed. LGTM

mitchell852:
@mattjackson220 - do we need a changelog entry? for the bug fix? i think we're pretty much adding CL entries for everything.

mattjackson220:
I wasnt going to do a changelog entry for this since it shouldnt change the real functionality other than fixing the issue. i could put an entry in basically saying it was updated not to do that but i hadnt seen that in the CL before so i skipped it.

PR #4842, opened 2020-06-30
Extract grove dependency

### LINKED ISSUE #4841: Remove Traffic Ops Golang Dependency on Grove
## I'm submitting a ...

- [x] Grove
- [x] Traffic Control Client
- [x] Traffic Ops

## Current behavior:
`traffic_ops_golang` and `traffic_ops/client` now depend on `grove/web`. Grove is an app, not a library, nothing should depend on it.

The things in `grove/web` that TO needs should be moved to `lib/go-rfc`, to fix the dependency.

## Expected / new behavior:
TO no longer depends on Grove.

Tech debt. Not actively causing a bug, but it's a logistical problem. TO requires Grove to build, build systems, tests, etc must all copy and build grove, etc, etc.

ocket8888:
- [x] This PR fixes #4841

Removes Traffic Ops's dependency on Grove as a library

- Grove
- Traffic Ops

Make sure Traffic Ops and Grove build and pass all of their tests.

### DIFF
--- a/grove/cache/retryinggetter.go
+++ b/grove/cache/retryinggetter.go
@@ -28,6 +28,7 @@ import (
 	"github.com/apache/trafficcontrol/grove/web"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
+	librfc "github.com/apache/trafficcontrol/lib/go-rfc"
 )
 
 const CodeConnectFailure = http.StatusBadGateway
@@ -150,13 +151,13 @@ func GetAndCache(
 		}
 
 		log.Debugf("GetAndCache request returned %v headers %+v (reqid %v)\n", respCode, respHeader, reqID)
-		respRespTime, ok := web.GetHTTPDate(respHeader, "Date")
+		respRespTime, ok := librfc.GetHTTPDate(respHeader, "Date")
 		if !ok {
 			log.Errorf("request %v returned no Date header - RFC Violation! Using local response timestamp (reqid %v)\n", req.RequestURI, reqID)
 			respRespTime = reqRespTime // if no Date was returned using the client response time simulates latency 0
 		}
 
-		lastModified, ok := web.GetHTTPDate(respHeader, "Last-Modified")
+		lastModified, ok := librfc.GetHTTPDate(respHeader, "Last-Modified")
 		if !ok {
 			lastModified = respRespTime
 		}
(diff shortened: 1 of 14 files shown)


grove/cache/retryinggetter.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/apache/trafficcontrol/grove/web"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
+	librfc "github.com/apache/trafficcontrol/lib/go-rfc"
```
rob05c, line 31: Kind of a nitpick, but I'm not a fan of aliases. IMO first-party libraries should have good names that don't need aliases. Could we just use `rfc` here?

ocket8888: no, because that shadows a grove package by the same name

I'd like to probably try to shunt that entire package into the lib, but I don't wanna do that as part of this PR.

rob05c: Ah, I see, this file already has `grove/rfc`. How much work would it be to move all of `grove/rfc` to `lib/go-rfc`?

rob05c: Sorry, replied before I saw your reply.

Yeah, that's fine.

rob05c:
> There is still a dependency on `web.CopyHeaderTo()` in [traffic_ops/client/session.go](https://github.com/apache/trafficcontrol/blob/d0317cdaa3/traffic_ops/client/session.go#L386).

Go 1.13 introduced `http.Header.Clone` which I believe could replace `grove/web.CopyHeaderTo`. I haven't tested though.
 
https://golang.org/pkg/net/http/#Header.Clone

ocket8888:
Should be removed. All the tests are failing because of IMS things, but raw requests seem to be working.

traffic_ops/client/session.go
```diff
@@ -369,12 +368,18 @@ func (to *Session) RawRequest(method, path string, body []byte, header http.Head
 		if err != nil {
 			return nil, remoteAddr, err
 		}
+		if header != nil {
```
rob05c, line 371: Nitpick: these nil checks aren't necessary, because `http.Header.Clone` is defined to: 
> Clone returns a copy of h or nil if h is nil.

ocket8888: No, they are, because if it returns nil I can't do
```go
req.Header = nil
```
then
```go
req.Header.Set("something", "something else")
```
will segfault.

PR #4844, opened 2020-06-30
Fixing migrations for CI build

srijeet0406:

- CDN in a Box
- Traffic Ops

Run the migrations on the DB, and then try seeding the DB multiple times. It should not give you any error.

- master


mitchell852:
do you think it needs to be like this instead?

```
ON CONFLICT (table_name) DO NOTHING
```

or maybe sql is smart enough to know that that this is the same thing because table_name is the primary key?

```
ON CONFLICT DO NOTHING
```

srijeet0406:
I tried running the seeding multiple times and it didn't complain, so I guess it is smart enough, but I could change it pretty easily if you'd like :)

mitchell852:
> I tried running the seeding multiple times and it didn't complain, so I guess it is smart enough, but I could change it pretty easily if you'd like :)

yeah, i'd say change it to the more explicit definition. why not?

PR #4856, opened 2020-07-07
Traffic Portal integration test cleanup

zrhoffman:

This PR does 2 things:
- Reorganizes the Traffic Portal integration test dockerfile so that only TP-related layers need to be rebuilt when files are modified (instead of also rebuilding dependency layers)
- Loads the Protractor config from a JSON file so the Traffic Portal integration test dockerfile can modify it using `jq` instead of sed and perl

- CDN in a Box
- Traffic Portal (tests)

Run the Traffic Portal integration test

In the CDN-in-a-Box directory:
```shell
docker-compose up -d 
docker-compose -f docker-compose.yml -f docker-compose.traffic-portal-test.yml up --build up portal-integration-test
```

- [x] This PR changes tests only, no documentation necessary


zrhoffman:
Rebased onto master

PR #4861, opened 2020-07-08
Add ORT Inferring Location Parameters

rob05c:
Adds ORT location Parameter inference. If location Parameters do not exist
the files are created and added anyway with the directory determined
from the local ATS install, for:
- cache.config
- hosting.config
- ip_allow.config
- parent.config
- plugin.config
- records.config
- remap.config
- storage.config
- volume.config
- Delivery Services with
  - Edge Header Rewrite
  - Mid Header Rewrite,
  - Cache URL
  - URL Sig
  - URI Signing

Note this is not an exhaustive list of required files, and many files
must be dynamic and cannot be inferred.

But this does remove the unnecessary manual configuration for this list.
More files may be added in the future.

This also makes `location` Parameters with Relative Paths append them to the local installed ATS config directory.
So for example, the duplicated `location` `/etc/trafficserver/opt/trafficserver` Parameters everywhere can be changed to `location` value `.`, and the `set_dscp_30.config` `location` `/etc/trafficserver/opt/trafficserver/dscp` Parameters in the Default Profiles can be changed to `set_dscp_30.config` `location` `dscp`.

I've manually tested in a production-like environment, verified ORT places server-wide configs like remap.config, as well as DS-specific configs like hdr_rw_dsname.config, without `location` Parameters. Also tested that relative paths like `""` and `.` work as expected.

Includes tests.
No docs, location Parameters aren't documented (maybe they should be, but that's outside the scope of this PR).
Includes changelog.

- Traffic Ops ORT

Run unit tests.
Run ORT against a Traffic Ops as a Server with no location Parameter on its Profile for remap.config, hdr_rw_dsname.config, and others in the above list, verify configs are still placed properly.

Not a bug fix.

### DIFF
--- a/lib/go-atscfg/meta.go
+++ b/lib/go-atscfg/meta.go
@@ -21,6 +21,8 @@ package atscfg
 
 import (
 	"encoding/json"
+	"errors"
+	"path/filepath"
 	"strings"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
@@ -31,7 +33,6 @@ type ConfigProfileParams struct {
 	FileNameOnDisk string
 	Location       string
 	URL            string
-	APIURI         string
 }
 
 // APIVersion is the Traffic Ops API version for config fiels.
@@ -39,6 +40,23 @@ type ConfigProfileParams struct {
 // Note the version in the meta config is not used by the atstccfg generator, which isn't actually an API.
 // TODO change the config system to not use old API paths, and remove this.
 const APIVersion = "2.0"
+
+// requiredFiles is a constant (because Go doesn't allow const slices).
+// Note these are not exhaustive. This is only used to error if these are missing.
+// The presence of these is no guarantee the location Parameters are complete and correct.
+func requiredFiles() []string {
+	return []string{
+		"cache.config",
+		"hosting.config",
+		"ip_allow.config",
+		"parent.config",
+		"plugin.config",
+		"records.config",
+		"remap.config",
+		"storage.config",
+		"volume.config",
+	}
+}
 
 func MakeMetaConfig(
 	serverHostName tc.CacheName,
@@ -50,10 +68,16 @@ func MakeMetaConfig(
 	scopeParams map[string]string, // map[configFileName]scopeParam
 	dsNames map[tc.DeliveryServiceName]struct{},
 ) string {
-	return MetaObjToMetaConfig(MakeMetaObj(serverHostName, server, tmURL, tmReverseProxyURL, locationParams, uriSignedDSes, scopeParams, dsNames))
-}
-
-func MetaObjToMetaConfig(atsData tc.ATSConfigMetaData) string {
+	// dses are only used when configDir is not empty
+	dses := map[tc.DeliveryServiceName]tc.DeliveryServiceNullable{}
+	for dsName, _ := range dsNames {
+		dses[dsName] = tc.DeliveryServiceNullable{}
+	}
+	configDir := ""
+	atsData, err := MakeMetaObj(serverHostName, server, tmURL, tmReverseProxyURL, locationParams, uriSignedDSes, scopeParams, dses, configDir)
+	if err != nil {
+		return "error creating meta config: " + err.Error()
+	}
 	bts, err := json.Marshal(atsData)
 	if err != nil {
 		// should never happen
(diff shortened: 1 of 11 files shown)


rob05c:
Also, once this is merged, we should make a new Feature Issue to change Traffic Ops Delivery Service creation and modification to not create or update `location` Parameters.

I thought about doing that here, but it seems safer and smaller to do them separately. And it doesn't hurt anything to keep creating them.

Should be straightforward to remove. Code is here: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go#L1476

ocket8888:
> _"If location Parameters do not exist the files are created and added anyway with the directory determined from the local ATS install"_

Does it actually infer based on the local ATS install, or is it just hard-coded to use `/opt/trafficserver/` as the root of the ATS install? If it does infer, how?

rob05c:
ATS Install. It tries `rpm` and falls back to `find`: https://github.com/apache/trafficcontrol/blob/4aa550d319c0269f8e9fbd8fabc0aa757da1a5f9/traffic_ops_ort/traffic_ops_ort.pl#L275

That's important for users who may have ATS installed somewhere other than `/opt/trafficserver`. Which is perfectly reasonable: The LSB makes it clear `/opt` is for people who don't play well with others, standard LSB directories (`/usr/sbin`, `/etc`, `/var/log`, etc.

I believe that's where ATS makefiles install by default, as well.

ocket8888:
> _"I believe that's where ATS makefiles install by default, as well."_

Yes, it is, which is exactly why I ask. Thanks :+1: .

rob05c:
Should also be easy to add other package manager like `apt` in the future

rob05c:
Retest this please.

rob05c:
retest this please

lib/go-atscfg/meta.go
```diff
@@ -40,6 +41,23 @@ type ConfigProfileParams struct {
 // TODO change the config system to not use old API paths, and remove this.
 const APIVersion = "2.0"
 
+// RequiredFiles is a constant (because Go doesn't allow const slices).
```
ocket8888, line 44: nit, but it's `requiredFiles`, not `RequiredFiles`.

lib/go-atscfg/meta.go
```diff
@@ -63,6 +87,146 @@ func MetaObjToMetaConfig(atsData tc.ATSConfigMetaData) string {
...
+		}
+		configFilesM[fileName] = []tc.ATSConfigMetaDataConfigFile{{
+			FileNameOnDisk: fileName,
+			Location:       configDir,
+			Scope:          string(getServerScope(fileName, server.Type, nil)),
+		}}
+	}
+
+	for fileName, fis := range configFilesM {
+		newFis := []tc.ATSConfigMetaDataConfigFile{}
```
ocket8888, line 132: Since you know ahead of time what the size of this needs to be, you can allocate: `newFis := make([]tc.ATSConfigMetaDataConfigFile, 0, len(fis))` to avoid possible multiple reallocations.

traffic_ops_ort/traffic_ops_ort.pl
```diff
@@ -1583,7 +1605,7 @@ sub get_cfg_file_list {
 		$atstccfg_reval_arg = '--revalidate-only';
 	}
 
-	my $result = `$atstccfg_cmd $atstccfg_timeout_arg $atstccfg_arg_disable_proxy --traffic-ops-user='$TO_USER' --traffic-ops-password='$TO_PASS' --traffic-ops-url='$TO_URL' --cache-host-name='$host_name' $atstccfg_reval_arg --log-location-error=stderr --log-location-warning=stderr --log-location-info=null 2>$atstccfg_log_path`;
+	my $result = `$atstccfg_cmd --dir='$ats_config_dir' $atstccfg_timeout_arg $atstccfg_arg_disable_proxy --traffic-ops-user='$TO_USER' --traffic-ops-password='$TO_PASS' --traffic-ops-url='$TO_URL' --cache-host-name='$host_name' $atstccfg_reval_arg --log-location-error=stderr --log-location-warning=stderr --log-location-info=null 2>$atstccfg_log_path`;
```
ocket8888, line 1608: Am I missing something? It doesn't look like you added this option-argument to `traffic_ops_ort/atstccfg/config/config.go`.

rob05c: Yeah, config.go line 103

lib/go-atscfg/meta.go
rob05c: Performance doesn't matter here, and I prefer not to use `make` unless necessary to have a unified variable syntax for readability. But you're right, it is faster. I can change it if you feel strongly about it.

lib/go-atscfg/meta.go
rob05c: Fixed

lib/go-atscfg/meta.go
ocket8888: I guess it doesn't really matter.

traffic_ops_ort/traffic_ops_ort.pl
ocket8888: Oh, okay, I must've been looking at a subset of changes or something.

rob05c:
+1

I'd vote we remove the CiaB location params, as well as change TO to not create DS location params, in subsequent PRs, just to keep the PRs as small as possible

rob05c:
retest this please

mitchell852:
@rob05c - can you create a github issue regarding the removal from TO of the creation of location params? I'm guessing that anywhere in the code that creates a location param can be removed, right?

rob05c:
Not anywhere. There are still both required and custom ("take-n-bake") locations that are necessary.

But any in that list above, yes. I'll make that case, as well as one for the CiaB ort.py inferring /etc/trafficserver

rob05c:
Issues created.

PR #4865, opened 2020-07-08
Update atstccfg to query api 3.0/2.0

shamrickus:

With the upcoming removal of the 1.0 TO api, atstccfg needs to be updated to query 3.0/2.0 instead of 2.0/1.0. 

There was a breaking change in `lib/go-tc` between api 2 and 3. To accommodate the vendorered `client/coordinate.go` file was modified manually to address the issue. In this case the difference does not matter since the original type was incorrect. In the future it might be required to revisit `update-to-client.go` for when more substantial breaking changes are made.

- Traffic Ops ORT

Ensure that atstccfg builds and can run as usual. 
Verify the atstccfg tests pass.
Also confirm that running `update-to-client.go` vendors both the client.


shamrickus:
atstccfg no longer vendors the client

PR #4869, opened 2020-07-08
Fixing /api/1.1/servers/{id} endpoint to not return empty response

### LINKED ISSUE #4867: GET /api/1.1/servers/{id} returns an empty response
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

`GET /api/1.1/servers/{id}` returns an empty response

## Expected / new behavior:

`GET /api/1.1/servers/{id}` should return the same response as `GET /api/1.1/servers?id={id}` but with a deprecation notice

1. Start CDN-in-a-Box and wait for it to finish seeding data
2. GET https://trafficops.infra.ciab.test/api/1.1/servers/1

## Anything else:

Noticed at ff50ddbae4

srijeet0406:

- [x] This PR fixes #4867  

- CDN in a Box
- Traffic Control Client Go 
- Traffic Ops

Run the API v1 tests and make sure they all pass.

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v1/servers_test.go
+++ b/traffic_ops/testing/api/v1/servers_test.go
@@ -16,8 +16,9 @@ package v1
 */
 
 import (
-	"github.com/apache/trafficcontrol/lib/go-tc"
 	"testing"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
 )
 
 func TestServers(t *testing.T) {
@@ -70,18 +71,18 @@ func UpdateTestServers(t *testing.T) {
 	var alert tc.Alerts
 	alert, _, err = TOSession.UpdateServerByID(remoteServer.ID, remoteServer)
 	if err != nil {
-		t.Fatalf("cannot UPDATE Server by ID: %v - %v", err, alert)
+		t.Errorf("cannot UPDATE Server by ID: %v - %v", err, alert)
 	}
 	// Retrieve the server to check rack and interfaceName values were updated
 	resp, _, err = TOSession.GetServerByID(remoteServer.ID)
 	if err != nil {
-		t.Fatalf("cannot GET Server by ID: %v - %v", remoteServer.ID, err)
+		t.Errorf("cannot GET Server by ID: %v - %v", remoteServer.ID, err)
 	}
 
 	respServer := resp[0]
 	if respServer.InterfaceName != updatedServerInterface || respServer.Rack != updatedServerRack {
-		t.Fatalf("results do not match actual: %s, expected: %s", respServer.InterfaceName, updatedServerInterface)
-		t.Fatalf("results do not match actual: %s, expected: %s", respServer.Rack, updatedServerRack)
+		t.Errorf("results do not match actual: %s, expected: %s", respServer.InterfaceName, updatedServerInterface)
+		t.Errorf("results do not match actual: %s, expected: %s", respServer.Rack, updatedServerRack)
 	}
 
 	// Assign server to DS and then attempt to update to a different type
@@ -116,7 +117,7 @@ func UpdateTestServers(t *testing.T) {
 	// Attempt Update - should fail
 	_, _, err = TOSession.UpdateServerByID(remoteServer.ID, remoteServer)
 	if err == nil {
-		t.Fatalf("expected error when updating Server Type of a server assigned to DSes")
+		t.Errorf("expected error when updating Server Type of a server assigned to DSes")
 	}
 
 }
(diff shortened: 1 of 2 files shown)


traffic_ops/testing/api/v1/servers_test.go
```diff
@@ -16,9 +16,8 @@ package v1
 */
 
 import (
-	"testing"
-
 	"github.com/apache/trafficcontrol/lib/go-tc"
+	"testing"
```
ocket8888, line 20: Why'd you re-order these? Normal procedure is to put standard library dependencies first, then in-project dependencies, then 3rd party.

traffic_ops/testing/api/v1/servers_test.go
```diff
@@ -71,19 +70,18 @@ func UpdateTestServers(t *testing.T) {
 	var alert tc.Alerts
 	alert, _, err = TOSession.UpdateServerByID(remoteServer.ID, remoteServer)
 	if err != nil {
-		t.Errorf("cannot UPDATE Server by hostname: %v - %v", err, alert)
+		t.Fatalf("cannot UPDATE Server by ID: %v - %v", err, alert)
```
ocket8888, line 73: Good fix on the typo, but why is that fatal? The one below definitely should be.

traffic_ops/testing/api/v1/servers_test.go
```diff
@@ -71,19 +70,18 @@ func UpdateTestServers(t *testing.T) {
...
-		t.Errorf("cannot GET Server by ID: %v - %v", remoteServer.HostName, err)
+		t.Fatalf("cannot GET Server by ID: %v - %v", remoteServer.ID, err)
 	}
 
 	respServer := resp[0]
 	if respServer.InterfaceName != updatedServerInterface || respServer.Rack != updatedServerRack {
-		t.Errorf("results do not match actual: %s, expected: %s", respServer.InterfaceName, updatedServerInterface)
-		t.Errorf("results do not match actual: %s, expected: %s", respServer.Rack, updatedServerRack)
+		t.Fatalf("results do not match actual: %s, expected: %s", respServer.InterfaceName, updatedServerInterface)
+		t.Fatalf("results do not match actual: %s, expected: %s", respServer.Rack, updatedServerRack)
```
ocket8888, line 84: Why are these fatal now? The rest of the test doesn't panic if the comparison is untrue. In fact, by making the first of these fatal, you've ensured that no one will ever see the second one, meaning that you can get the cryptic failure message:
```
results do not match actual: test, expected: test
```

traffic_ops/testing/api/v1/servers_test.go
```diff
@@ -118,7 +116,7 @@ func UpdateTestServers(t *testing.T) {
 	// Attempt Update - should fail
 	_, _, err = TOSession.UpdateServerByID(remoteServer.ID, remoteServer)
 	if err == nil {
-		t.Errorf("expected error when updating Server Type of a server assigned to DSes")
+		t.Fatalf("expected error when updating Server Type of a server assigned to DSes")
```
ocket8888, line 119: This one doesn't really matter either way, because it's at the end. Just seems a bit unnecessary.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -581,18 +581,28 @@ func ReadID(w http.ResponseWriter, r *http.Request) {
...
 	}
-
 	deprecationAlerts := api.CreateDeprecationAlerts(&alternative)
 
 	// No need to bother converting if there's no data
 	if len(servers) < 1 {
 		api.WriteAlertsObj(w, r, http.StatusOK, deprecationAlerts, servers)
+		return
 	}
+	legacyServers := make([]tc.ServerNullableV11, 0, 1)
```
ocket8888, line 595: as long as you're allocating memory ahead of time, you may as well set the capacity to what we know it needs to be; instead of 1 it'd be much more efficient to go ahead and allocate room for `len(servers)`.

mitchell852:
@ocket8888 - is this related to #4820 in any way? can they be merged together or do they address different problems?

ocket8888:
They're different problems

mitchell852:
@ocket8888 - you want to merge this?

PR #4878, opened 2020-07-14
Use latests grunt-browserify to fix audit issues

shamrickus:

- [x] This PR fixes a large portion of the `npm audit` issues present in TP.

The remaining audit issues require third party libraries to update their dependencies (specifically load-grunt-config and grunt-usemin)

- Traffic Portal

Run both production and dev test environments locally, ensure TP tests pass.

## Additional Information
Audit results before:
```
found 34 vulnerabilities (31 low, 2 high, 1 critical) in 664 scanned packages
16 vulnerabilities require manual review. See the full report for details.
```

Audit results after:
```
found 5 vulnerabilities (3 low, 2 high) in 725 scanned packages  
5 vulnerabilities require manual review. See the full report for details.
```


traffic_portal/package.json
```diff
@@ -1,13 +1,14 @@
...
   "devDependencies": {
+    "browserify": "^16.5.1",
     "connect-livereload": "0.6.1",
     "connect-modrewrite": "0.10.2",
     "connect-timeout": "1.9.0",
     "express": "4.17.1",
     "grunt": "1.0.4",
     "grunt-adam-compass": "0.7.1",
-    "grunt-browserify2": "0.1.8",
+    "grunt-browserify": "^5.3.0",
```
mitchell852, line 11: this always worries me. see how everything else is pinned to a version? i got burned in the past when not pinning to a version.....then again that's a good way to fall behind on libraries. but i suppose our ci pipieline is better now and should catch any problems as a result of library upgrades.

shamrickus: Good point, perhaps `~` (patch releases only) would be better. In any case, I'll make it exact versioning for now as that change should probably be made to the entire file in a separate PR.

PR #4881, opened 2020-07-15
Renaming a host in TC, does not impact hash_id

### LINKED ISSUE #2156: When renaming a host in TC, the hashId retains the value of the previous hostname
Observed in TP 2.3.0-8247.fb69f019.el7

Possible hashId naming collision if previous hostname value is retained.

rimashah25:
- [x] This PR fixes #2156
- On server create, PR sets xmppId = UUID and removes hostname from the equation. 
- On server update, PR prevents xmppId (removing it from update sql query) from ever being changed once set as it is the source of hashId used by traffic router for consistent hashing.

- Documentation
- Traffic Ops

1. Create a server in traffic_portal. Observe the xmpp_id value in db. It should be set to uuid.
2. Edit the hostname of the created server (from step 1)
3. Observe the change of hostname in DB but xmpp_id remains unaffected.

- master 
- 4.1.0

- [x] This PR includes tests

## Additional Information
A new package, uuid, was added to traffic_ops_golang vendor folder.


traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1224,6 +1228,9 @@ func createV3(inf *api.APIInfo, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	str := uuid.New().String()
```
mitchell852, line 1231: sorry, i just did a quick glance at the code. what happens on server update (PUT /servers/{id})? can the user change the XMPPID? If so, that seems bad and should be prevented if the XMPPID is driving the hashId value used in TR.

mitchell852: so i guess what i'm getting at is this:

```
Create a server in traffic_portal. Observe the xmpp_id value in db. It should be set to uuid.
Edit the hostname of the created server (from step 1)
Observe the change of hostname in DB but xmpp_id remains unaffected.
```

but can i update the xmppId on a server via PUT /servers/{id}?

rimashah25: Thanks for the feedback, Jeremy. I have updated the PR to handle change in XMPPID, which would be ignored and ensure hashid remains unaffected.

mitchell852:
@rimashah25 - so you probably want to update the api docs that say:

```
xmppId: An identifier to be used in XMPP communications with the server - in nearly all cases this will be the same as hostName
```

to something like:

```
xmppId: A system-generated UUID used to generate a server hash id. This value is set when a server is created.
```

rimashah25:
> @rimashah25 - so you probably want to update the api docs that say:
> 
> ```
> xmppId: An identifier to be used in XMPP communications with the server - in nearly all cases this will be the same as hostName
> ```
> 
> to something like:
> 
> ```
> xmppId: A system-generated UUID used to generate a server hash id for use in traffic router's consistent hashing algorithm. This value is set when a server is created and cannot be changed afterwards.
> ```
> 
> or something like that.

@mitchell852: Updated the documentation wrt xmppId

traffic_ops/testing/api/v2/servers_test.go
```diff
@@ -101,17 +101,23 @@ func UpdateTestServers(t *testing.T) {
 	hostName := firstServer.HostName
 	// Retrieve the server by hostname so we can get the id for the Update
 	resp, _, err := TOSession.GetServerByHostName(hostName)
+	originalHostname := resp[0].HostName
```
rimashah25, line 104: Test cases to test the following:
1. Change in hostname updated the field hostName but doesn't impact xmppid
2. Also, any change to xmppid, the Update() ignores the change.

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -248,6 +249,8 @@ func UpdateTestServers(t *testing.T) {
 
 	// Retrieve the server by hostname so we can get the id for the Update
 	resp, _, err := TOSession.GetServers(&params, nil)
+	originalHostname := *resp.Response[0].HostName
```
rimashah25, line 252: Test cases to test the following:
1. Change in hostname updated the field hostName but doesn't impact xmppid
2. Also, any change to xmppid, the Update() ignores the change.

Also, the original hostname is retained to ensure all the following test cases pass.

rimashah25, line 252: Test cases to test the following:
1. Change in hostname updated the field hostName but doesn't impact xmppid
2. Also, any change to xmppid, the Update() ignores the change.

Also, the original hostname is retained to ensure all the following test cases pass.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -47,6 +47,7 @@ import (
 
 	"github.com/go-ozzo/ozzo-validation"
 	"github.com/go-ozzo/ozzo-validation/is"
+	"github.com/google/uuid"
```
rimashah25, line 50: Using UUID to set xmppid to a random uuid

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -255,7 +256,6 @@ UPDATE server SET
 	tcp_port=:tcp_port,
 	type=:server_type_id,
 	upd_pending=:upd_pending,
-	xmpp_id=:xmpp_id,
```
rimashah25, line 258: Removed xmppid update to avoid impacting hashId

traffic_ops/testing/api/v2/servers_test.go
srijeet0406, line 104: We should probably do this after we check that the error is nil. In case we dont get back a valid resp, this can throw a seg fault due to dereferencing of a nil ptr.

traffic_ops/testing/api/v3/servers_test.go
srijeet0406, line 252: Same comment here.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -992,6 +992,9 @@ func Update(w http.ResponseWriter, r *http.Request) {
 			api.HandleErr(w, r, tx, http.StatusBadRequest, err, nil)
 			return
 		}
+		if newServer.XMPPID != nil {
+			fmt.Println("Change in xmpp_id was requested but it will be ignored to ensure hashId remains consistent")
```
srijeet0406, line 996: Change to `log.Debugln` or something like that

traffic_ops/testing/api/v2/servers_test.go
rimashah25: Sounds good. Changed it.

traffic_ops/testing/api/v3/servers_test.go
rimashah25: Completed the change.

traffic_ops/traffic_ops_golang/server/servers.go
rimashah25: Changed.

srijeet0406:
Noticing an issue in the following case:
`
1.) Create a server
2.) Change the host name
3.) The xmpp_id doesn't change -> so far it looks great
4.) Do a PUT to change the server's xmpp ID with a json payload that has a different xmpp ID than the one in the database
5.) The Change doesn't take place in the DB (which is the expected behavior), however, we get back a message saying that the server was updated with the new xmpp ID. This is what the response looks like:

{"alerts":[{"text":"Server updated","level":"success"}],"response":{"cachegroup":"edgeserver","cachegroupId":2,"cdnId":1,"cdnName":"ALL","domainName":"rima2","guid":null,"hostName":"rimatest2","httpsPort":443,"id":10,"iloIpAddress":null,"iloIpGateway":null,"iloIpNetmask":null,"iloPassword":null,"iloUsername":null,"lastUpdated":"2020-07-17 14:51:47-06","mgmtIpAddress":null,"mgmtIpGateway":null,"mgmtIpNetmask":null,"offlineReason":null,"physLocation":"phys2","physLocationId":2,"profile":"TRAFFIC_OPS","profileDesc":"Traffic Ops profile","profileId":3,"rack":null,"revalPending":false,"routerHostName":null,"routerPortName":null,"status":"ONLINE","statusId":2,"tcpPort":80,"type":"EDGE","typeId":11,"updPending":false,"xmppId":"5fa51f60-6606-4c9d-8b60-2d78f96018zf","xmppPasswd":null,"interfaces":[{"ipAddresses":[{"address":"192.23.27.67","gateway":"255.255.245.0","serviceAddress":true},{"address":"123.45.67.89","gateway":"255.255.255.0","serviceAddress":false}],"maxBandwidth":5,"monitor":false,"mtu":1500,"name":"rima-interface"}]}}
`
The xmpp ID before this update was `5fa51f60-6606-4c9d-8b60-2d78f96018ae` in the database.
This remains the same even after the `PUT` (which is what we want), however, notice how the response says that the server was updated with the new xmpp ID of `5fa51f60-6606-4c9d-8b60-2d78f96018zf`. We do not want this. The user should be notified that the `server couldn't be updated because change of xmpp_id is not allowed`, or something to that effect.

mitchell852:
> Noticing an issue in the following case:
> 
> 1.) Create a server
> 2.) Change the host name
> 3.) The xmpp_id doesn't change -> so far it looks great
> 4.) Do a PUT to change the server's xmpp ID with a json payload that has a different xmpp ID than the one in the database
> 5.) The Change doesn't take place in the DB (which is the expected behavior), however, we get back a message saying that the server was updated with the new xmpp ID. This is what the response looks like:
> 
> `{"alerts":[{"text":"Server updated","level":"success"}],"response":{"cachegroup":"edgeserver","cachegroupId":2,"cdnId":1,"cdnName":"ALL","domainName":"rima2","guid":null,"hostName":"rimatest2","httpsPort":443,"id":10,"iloIpAddress":null,"iloIpGateway":null,"iloIpNetmask":null,"iloPassword":null,"iloUsername":null,"lastUpdated":"2020-07-17 14:51:47-06","mgmtIpAddress":null,"mgmtIpGateway":null,"mgmtIpNetmask":null,"offlineReason":null,"physLocation":"phys2","physLocationId":2,"profile":"TRAFFIC_OPS","profileDesc":"Traffic Ops profile","profileId":3,"rack":null,"revalPending":false,"routerHostName":null,"routerPortName":null,"status":"ONLINE","statusId":2,"tcpPort":80,"type":"EDGE","typeId":11,"updPending":false,"xmppId":"5fa51f60-6606-4c9d-8b60-2d78f96018zf","xmppPasswd":null,"interfaces":[{"ipAddresses":[{"address":"192.23.27.67","gateway":"255.255.245.0","serviceAddress":true},{"address":"123.45.67.89","gateway":"255.255.255.0","serviceAddress":false}],"maxBandwidth":5,"monitor":false,"mtu":1500,"name":"rima-interface"}]}}`
> 
> The xmpp ID before this update was `5fa51f60-6606-4c9d-8b60-2d78f96018ae` in the database.
> This remains the same even after the `PUT` (which is what we want), however, notice how the response says that the server was updated with the new xmpp ID of `5fa51f60-6606-4c9d-8b60-2d78f96018zf`. We do not want this. The user should be notified that the `server couldn't be updated because change of xmpp_id is not allowed`, or something to that effect.

I'm wondering if rather than simply ignoring the xmppId in the update query (which I believe is what it is doing) maybe it makes sense to return a 400 bad request if request.xmppId does not equal the xmpp_id found in the database. i.e. if the user is trying to update xmppId, return 400 bad request with an alert of type error with a message like "xmppId cannot be changed".

ocket8888:
To keep with our API pattern, the fields should only be ignored if we're gonna remove it from responses. Changing it would be an invalid action on a field we admit exists, so it should be a 4XX error.

rimashah25:
> Noticing an issue in the following case:
> 
> 1.) Create a server
> 2.) Change the host name
> 3.) The xmpp_id doesn't change -> so far it looks great
> 4.) Do a PUT to change the server's xmpp ID with a json payload that has a different xmpp ID than the one in the database
> 5.) The Change doesn't take place in the DB (which is the expected behavior), however, we get back a message saying that the server was updated with the new xmpp ID. This is what the response looks like:
> 
> `{"alerts":[{"text":"Server updated","level":"success"}],"response":{"cachegroup":"edgeserver","cachegroupId":2,"cdnId":1,"cdnName":"ALL","domainName":"rima2","guid":null,"hostName":"rimatest2","httpsPort":443,"id":10,"iloIpAddress":null,"iloIpGateway":null,"iloIpNetmask":null,"iloPassword":null,"iloUsername":null,"lastUpdated":"2020-07-17 14:51:47-06","mgmtIpAddress":null,"mgmtIpGateway":null,"mgmtIpNetmask":null,"offlineReason":null,"physLocation":"phys2","physLocationId":2,"profile":"TRAFFIC_OPS","profileDesc":"Traffic Ops profile","profileId":3,"rack":null,"revalPending":false,"routerHostName":null,"routerPortName":null,"status":"ONLINE","statusId":2,"tcpPort":80,"type":"EDGE","typeId":11,"updPending":false,"xmppId":"5fa51f60-6606-4c9d-8b60-2d78f96018zf","xmppPasswd":null,"interfaces":[{"ipAddresses":[{"address":"192.23.27.67","gateway":"255.255.245.0","serviceAddress":true},{"address":"123.45.67.89","gateway":"255.255.255.0","serviceAddress":false}],"maxBandwidth":5,"monitor":false,"mtu":1500,"name":"rima-interface"}]}}`
> 
> The xmpp ID before this update was `5fa51f60-6606-4c9d-8b60-2d78f96018ae` in the database.
> This remains the same even after the `PUT` (which is what we want), however, notice how the response says that the server was updated with the new xmpp ID of `5fa51f60-6606-4c9d-8b60-2d78f96018zf`. We do not want this. The user should be notified that the `server couldn't be updated because change of xmpp_id is not allowed`, or something to that effect.

@mitchell852 @srijeet0406: I have updated the PR to reflect the changes based on this comment.

mitchell852:
`{"alerts":[{"text":"Bad Request","level":"error"}]}`

@rima - can you add message to let the user know what was wrong with the request? i.e. `xmppId is immutable` or something like that?

rimashah25:
> `{"alerts":[{"text":"Bad Request","level":"error"}]}`
> 
> @Rima - can you add message to let the user know what was wrong with the request? i.e. `xmppId is immutable` or something like that?

@mitchell852 Done.

traffic_ops/testing/api/v2/servers_test.go
```diff
@@ -130,6 +136,33 @@ func UpdateTestServers(t *testing.T) {
...
 	}
 
+	//Check change in hostname with no change to xmppid
+	if originalHostname == respServer.HostName && originalXMPIDD == respServer.XMPPID {
+		t.Errorf("HostName didn't change. Expected: #{updatedHostName}, actual: #{originalHostname}")
+	}
+
+	//Check to verify XMPPID never gets updated
+	changeXMPPID := true
+	if changeXMPPID {
```
srijeet0406, line 146: Won't this always be true?

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -321,6 +332,33 @@ func UpdateTestServers(t *testing.T) {
 		t.Fatalf("Cannot test server type change update; server '%s' had nil type ID", hostName)
 	}
 
+	//Check change in hostname with no change to xmppid
+	if originalHostname == *respServer.HostName && originalXMPIDD == *respServer.XMPPID {
+		t.Errorf("HostName didn't change. Expected: #{updatedHostName}, actual: #{originalHostname}")
+	}
+
+	//Check to verify XMPPID never gets updated
+	changeXMPPID := true
```
srijeet0406, line 341: Same comment here

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -984,6 +984,11 @@ func Update(w http.ResponseWriter, r *http.Request) {
 	}
 	defer inf.Close()
 
+	//Get original xmppid
+	origSer, _, _, _, _, _ := getServers(r.Header, inf.Params, inf.Tx, inf.User, false)
```
srijeet0406, line 988: We should check for the errors here, and only dereference the `origSer` variable in the case when there was no error

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1055,52 +1065,56 @@ func Update(w http.ResponseWriter, r *http.Request) {
...
 
-	rows, err := inf.Tx.NamedQuery(updateQuery, server)
-	if err != nil {
-		userErr, sysErr, errCode = api.ParseDBError(err)
-		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
-		return
-	}
-	defer rows.Close()
+	if changeXMPPID {
+		api.WriteRespAlertObj(w, r, tc.ErrorLevel, fmt.Sprintf("server cannot be updated due to requested XMPPID change. XMPIDD is immutable"), nil)
```
srijeet0406, line 1069: You should return here, after `api.WriteRespAlertObj` because of two reasons:
1.) That way, you dont have to use the `else` block
2.) If you dont return, the `error.log` will complain about response being written twice

traffic_ops/testing/api/v2/servers_test.go
rimashah25: Changed.

traffic_ops/testing/api/v3/servers_test.go
rimashah25: Changed.

traffic_ops/traffic_ops_golang/server/servers.go
rimashah25: Added error check

traffic_ops/traffic_ops_golang/server/servers.go
rimashah25: :) Simplified it.

docs/source/api/v2/servers.rst
```diff
@@ -244,7 +244,7 @@ Request Structure
 
 :typeId:     The integral, unique identifier of the 'type' of this server
 :updPending: A boolean value which, if ``true``, indicates that the server has updates of some kind pending, typically to be acted upon by Traffic Ops ORT
-:xmppId:     An optional identifier to be used in XMPP communications with the server - in nearly all cases this should be the same as ``hostName``
+:xmppId:    A system-generated UUID used to generate a server hashId for use in traffic router's consistent hashing algorithm. This value is set when a server is created and cannot be changed afterwards.
```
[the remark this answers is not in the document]
rimashah25: Done

docs/source/api/v2/servers_id.rst
```diff
@@ -183,7 +183,7 @@ Response Structure
 :type:       The name of the 'type' of this server
 :typeId:     The integral, unique identifier of the 'type' of this server
 :updPending: A boolean value which, if ``true``, indicates that the server has updates of some kind pending, typically to be acted upon by Traffic Ops ORT
-:xmppId:     An identifier to be used in XMPP communications with the server - in nearly all cases this will be the same as ``hostName``
+:xmppId:     A system-generated UUID used to generate a server hashId for use in traffic router's consistent hashing algorithm. This value is set when a server is created and cannot be changed afterwards.
```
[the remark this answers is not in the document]
rimashah25: Done

docs/source/api/v3/servers.rst
```diff
@@ -287,7 +287,7 @@ Request Structure
 
 :typeId:     The integral, unique identifier of the 'type' of this server
 :updPending: A boolean value which, if ``true``, indicates that the server has updates of some kind pending, typically to be acted upon by Traffic Ops ORT
-:xmppId:     An optional identifier to be used in XMPP communications with the server - in nearly all cases this should be the same as ``hostName``
+:xmppId:     A system-generated UUID used to generate a server hashId for use in traffic router's consistent hashing algorithm. This value is set when a server is created and cannot be changed afterwards.
```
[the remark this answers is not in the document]
rimashah25: Done

docs/source/api/v3/servers_id.rst
```diff
@@ -94,7 +94,7 @@ Request Structure
 
 :typeId:     The integral, unique identifier of the 'type' of this server
 :updPending: A boolean value which, if ``true``, indicates that the server has updates of some kind pending, typically to be acted upon by Traffic Ops ORT
-:xmppId:     An optional identifier to be used in XMPP communications with the server - in nearly all cases this should be the same as ``hostName``
```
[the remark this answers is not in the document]
rimashah25: Done

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1055,6 +1069,11 @@ func Update(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if changeXMPPID {
+		api.WriteRespAlertObj(w, r, tc.ErrorLevel, fmt.Sprintf("server cannot be updated due to requested XMPPID change. XMPIDD is immutable"), nil)
```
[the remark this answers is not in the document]
rimashah25: Done

rimashah25:
> Since `xmppId` is always no longer the hostname, CDN-in-a-Box won't finish starting up, because for some reason we were [using xmppIds as hostnames](https://github.com/apache/trafficcontrol/blob/8e4a27b38/infrastructure/cdn-in-a-box/traffic_ops/to-access.sh#L335):
> 
> ```shell
> current_servers_json=$(to-get 'api/'$TO_API_VERSION'/servers' 2>/dev/null | jq -c -e '[.response[] | .xmppId] | sort')
> ```
> 
> Would you please change `.xmppId` there to `.hostName`?

Done

Issue #4824, opened 2020-06-24
New TP server table suggestions


traeak:
## I'm submitting a ...

- [ ] bug report
- [ ] new feature / enhancement request
- [x] improvement request (usability, performance, tech debt, etc.)
- [ ] other 

## Traffic Control components affected ...

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

Just some nits on the new TP servers portal.  Purely my opinion.

- How to search isn't initially obvious.  Per column search is a good feature though.
- Single line table results initially have terrible elision as everything looks the same.
- It would be *lovely* to have tooltip hover for non column expanded items.

mitchell852:
> * How to search isn't initially obvious.  Per column search is a good feature though.

Per column table search makes searching very powerful but i agree, it isn't very obvious how to do it. i also think a global, across all column search (like the old table had) is still very valuable to the non-power user.

rawlinp:
I'll add on that I don't really mind adjusting column widths as long as TP will remember the widths, but it would be nice if adjusting one column's width didn't eat into the width of the column next to it. That would make adjusting the column widths much easier.

ocket8888:
> _"Single line table results initially have terrible elision as everything looks the same."_

> _"In linguistics, an elision or deletion is the omission of one or more sounds in a word or phrase. The word elision is frequently used in linguistic description of living languages, and deletion is often used in historical linguistics for a historical sound change." - Wikipedia article on elision_

I'm not sure how to apply that in this context

> _"It would be lovely to have tooltip hover for non column expanded items."_

What are "non column expanded items"? You want a tooltip on hover that shows all of the columns that are hidden? Or a tooltip that shows cell content when the column is so thin that content gets cut off?

mitchell852:
there's also an issue where if you apply a filter to a column and then hide that column, the filter remains in place. if the column, is hidden the filter should be removed.

traeak:
No a tooltip over the items in a column that have been truncated.  It's possible (but not necessary) to more intelligently shorten what's being displayed.  Fixing that is more of an excuse to write something that uses dynamic programming than anything else.

[...]

PR #4891, opened 2020-07-17
TP: adds ability to perform a quick search, override default pagination size and clear column filters to servers table

mitchell852:

- [x] This PR is related to #4824 

The following improvements were added to the servers table:
- quick search across all columns
- ability to customize page size (override the default of 100)
- ability to clear all column filters
- adds a tooltip to the column values

image

- Traffic Portal

- Run the UI tests and ensure they pass
- Navigate to https://tp.domain.com/#!/servers and try the following:
1. perform a quick search and ensure the rows are filtered appropriately. refresh. ensure value is persistent. 
2. change the pagination page size and ensure the page size changes. refresh. ensure value is persistent.
3. add one or more column filters and then hit the `clear column filters`  menu item under `More` to ensure the column filters are removed. 

### DIFF
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -41,7 +41,7 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		} else {
 			this.eGui.classList.add("fa-check");
 		}
-	}
+	};
 	UpdateCellRenderer.prototype.getGui = function() {return this.eGui;};
 
 	/**
@@ -548,6 +548,13 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		}
 	};
 
+	$scope.onQuickSearchChanged = function() {
+		$scope.gridOptions.api.setQuickFilter(document.getElementById('quickSearch').value);
+	};
+
+	$scope.clearColFilters = function() {
+		$scope.gridOptions.api.setFilterModel(null);
+	};
 
 	/**** Initialization code, including loading user columns from localstorage ****/
 	angular.element(document).ready(function () {
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -552,6 +552,11 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		$scope.gridOptions.api.setQuickFilter(document.getElementById('quickSearch').value);
 	};
 
+	$scope.onPageSizeChanged = function() {
+		let value = document.getElementById('pageSize').value;
+		$scope.gridOptions.api.paginationSetPageSize(Number(value));
+	};
+
 	$scope.clearColFilters = function() {
 		$scope.gridOptions.api.setFilterModel(null);
 	};
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -275,6 +275,8 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		id: -1
 	};
 
+	$scope.pageSize = 100;
+
 	/** Options, configuration, data and callbacks for the ag-grid table. */
 	$scope.gridOptions = {
 		components: {
@@ -294,6 +296,7 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		},
 		rowData: servers,
 		pagination: true,
+		paginationPageSize: $scope.pageSize,
 		rowBuffer: 0,
 		onColumnResized: function(params) {
 			localStorage.setItem("servers_table_columns", JSON.stringify($scope.gridOptions.columnApi.getColumnState()));
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -611,7 +611,7 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 
 		try {
 			const ps = localStorage.getItem("servers_page_size");
-			if (ps) {
+			if (ps && ps > 0) {
 				$scope.pageSize = Number(ps);
 				$scope.gridOptions.api.paginationSetPageSize($scope.pageSize);
 			}
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -275,6 +275,8 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		id: -1
 	};
 
+	$scope.quickSearch = '';
+
 	$scope.pageSize = 100;
 
 	/** Options, configuration, data and callbacks for the ag-grid table. */
@@ -552,13 +554,12 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 	};
 
 	$scope.onQuickSearchChanged = function() {
-		const quickSearchValue = document.getElementById('quickSearch').value;
-		$scope.gridOptions.api.setQuickFilter(quickSearchValue);
-		localStorage.setItem("servers_quick_search", quickSearchValue);
+		$scope.gridOptions.api.setQuickFilter($scope.quickSearch);
+		localStorage.setItem("servers_quick_search", $scope.quickSearch);
 	};
 
 	$scope.onPageSizeChanged = function() {
-		const value = Number(document.getElementById('pageSize').value);
+		const value = Number($scope.pageSize);
 		$scope.gridOptions.api.paginationSetPageSize(value);
 		localStorage.setItem("servers_page_size", value);
 	};
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -52,7 +52,14 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		if (!params.value || !serverUtils.isOffline(params.value)) {
 			return;
 		}
-		return params.data.offlineReason;
+		return params.value + ': ' + params.data.offlineReason;
+	}
+
+	/**
+	 * Gets value to display a default tooltip.
+	 */
+	function defaultTooltip(params) {
+		return params.value;
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -548,6 +551,18 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		}
 	};
 
+	$scope.onQuickSearchChanged = function() {
+		$scope.gridOptions.api.setQuickFilter(document.getElementById('quickSearch').value);
```
ocket8888, line 555: You're using `ng-model`, so you don't need to do `document.getElementById`, just use the model value.

mitchell852: oh duh. i added ng-model later...

traffic_portal/app/src/common/modules/table/servers/table.servers.tpl.html
```diff
@@ -23,31 +23,47 @@
...
+                <div class="btn-group" role="group" uib-dropdown is-open="more.isopen">
+                    <button name="moreBtn" type="button" class="btn btn-default dropdown-toggle" uib-dropdown-toggle aria-haspopup="true" aria-expanded="false">
+                        More&nbsp;
+                        <span class="caret"></span>
+                    </button>
+                    <ul class="dropdown-menu-right dropdown-menu" uib-dropdown-menu>
+                        <li role="menuitem"><a name="createServerMenuItem" href="#!/servers/new">Create New Server</a></li>
+                        <li class="divider"></li>
+                        <li role="menuitem"><a ng-click="confirmCDNQueueServerUpdates(null)">Queue Server Updates</a></li>
+                        <li role="menuitem"><a ng-click="confirmCDNClearServerUpdates(null)">Clear Server Updates</a></li>
```
ocket8888, line 61: You should use buttons for on-click behavior not related to navigation.

traffic_portal/app/src/common/modules/table/servers/table.servers.tpl.html
```diff
@@ -23,31 +23,47 @@
...
+            <div class="form-inline">
+                <div class="input-group text-input">
+                    <span class="input-group-btn">
+                        <button class="btn btn-default" type="button"><i class="fa fa-search"></i></button>
+                    </span>
+                    <input type="text" class="form-control" placeholder="Quick search..." ng-model="quickSearch" ng-change="onQuickSearchChanged()"/>
+                </div>
+                <div class="input-group text-input">
+                    <span class="input-group-btn">
+                        <button class="btn btn-default" type="button">Page Size:</button>
```
ocket8888, line 35: Why is this a button? Should it be a label?

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -548,6 +553,20 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 		}
 	};
 
+	$scope.onQuickSearchChanged = function() {
+		$scope.gridOptions.api.setQuickFilter($scope.quickSearch);
+		localStorage.setItem("servers_quick_search", $scope.quickSearch);
+	};
+
+	$scope.onPageSizeChanged = function() {
+		const value = Number($scope.pageSize);
```
ocket8888, line 562: I could be wrong, but since you guarded in the `localStorage` load and used `input[type=number]` I think this may not be necessary.

mitchell852: maybe not but i recall it not working before the casting. casting a number to a number won't hurt so...

traffic_portal/app/src/common/modules/table/servers/table.servers.tpl.html
mitchell852: image

just a button for style purposes. both the search icon and the `Page Size` text.

mitchell852: i'll fix it

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
mitchell852: i think it's because `input[type=number]` is still a number string

ocket8888: Oh, in general if you access `value` you get back a `USVString` type, but I thought Angular bound form controls to `valueAsNumber` to avoid that for numeric inputs. I could be wrong, might be an Angular thing and not an AngularJS thing.

traffic_portal/app/src/common/modules/table/servers/table.servers.tpl.html
```diff
@@ -23,31 +23,47 @@
...
-                    <li role="menuitem"><a ng-click="confirmCDNQueueServerUpdates(null)">Queue CDN Server Updates</a></li>
-                    <li role="menuitem"><a ng-click="confirmCDNClearServerUpdates(null)">Clear CDN Server Updates</a></li>
-                    <li role="menuitem"><button class="menu-item-button" type="button" ng-click="exportCSV()">Export CSV</button></li>
-                </ul>
+            <div class="form-inline">
+                <div class="input-group text-input">
+                    <span class="input-group-btn">
+                        <label for="quickSearch" class="btn btn-default"><i class="fa fa-search"></i></label>
+                    </span>
+                    <input id="quickSearch" type="text" class="form-control" placeholder="Quick search..." ng-model="quickSearch" ng-change="onQuickSearchChanged()"/>
```
ocket8888, line 31: Inputs without textual labels should have `aria-label` equivalents that describe what the input is (e.g. "quick search servers table"). Alternatively/Additionally, you could make `div.form-inline` into `form.form-inline[role=search]`, which will cause screen readers to announce it as a search form, which makes it more clear what's going to happen on submit (although we're capturing the `change` event of the input itself not the `submit` event of the form, so it's possibly not that helpful).

This could also be an `input[type=search]` instead of `input[type=text]` - the two are functionally identical, but browsers may choose input suggestions differently between the two. that's not really a big deal.

ocket8888: I'm also generally wary of `inputs` with no `name`, but again since there is no owning `form` - and it would never be actually submitted in any case - that's probably not a big deal.

traffic_portal/app/src/common/modules/table/servers/table.servers.tpl.html
```diff
@@ -23,31 +23,47 @@
...
+                </div>
+                <div class="input-group text-input">
+                    <span class="input-group-btn">
+                        <label for="pageSize" class="btn btn-default">Page Size:</label>
+                    </span>
+                    <input id="pageSize" type="number" class="form-control" placeholder="100" ng-model="pageSize" ng-change="onPageSizeChanged()"/>
```
ocket8888, line 37: Is there a maximum we want to consider? Or a "step"? So right now if I hit the up arrow it adds one more entry per page, but I think previously it was restricted to multiples of 25.

I don't really feel strongly about having either of those, just wanted to see if you had any thoughts about them.

Also, the same `name` attribute stuff as above.

mitchell852: yeah, step is probably a good idea. will add.

traffic_portal/app/src/common/modules/table/servers/table.servers.tpl.html
mitchell852: i think i got this all addressed.

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -240,18 +275,21 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
 			headerName: "TCP Port",
 			field: "tcpPort",
 			hide: true,
+			tooltip: defaultTooltip
 		},
 		{
 			headerName: "Type",
 			field: "type",
 			hide: false,
+			tooltip: defaultTooltip
```
ocket8888, line 284: Instead of adding a ton of these tooltips, you should probably just add `tooltip: defaultTooltip` to `gridOptions.defaultColDef`

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -548,6 +591,20 @@ var TableServersController = function(servers, $scope, $state, $uibModal, $windo
...
+	};
+
+	$scope.onPageSizeChanged = function() {
+		const value = Number($scope.pageSize);
+		$scope.gridOptions.api.paginationSetPageSize(value);
+		localStorage.setItem("servers_page_size", value);
+	};
+
+	$scope.clearColFilters = function() {
+		$scope.gridOptions.api.setFilterModel(null);
```
ocket8888, line 606: don't you also wanna clear the filters stored in `localStorage`?

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
mitchell852: oh. yeah, i was wondering if a better way

mitchell852: done

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
mitchell852: this dispatches an event that is already handled with:

```
		$scope.gridOptions.api.addEventListener("filterChanged", function() {
			localStorage.setItem("servers_table_filters", JSON.stringify($scope.gridOptions.api.getFilterModel()));
		});

```

ocket8888: oh, alright.

mitchell852:
> I think it's weird that I can set my page size to 0, -900, etc - but also setting the page size doesn't appear to do anything:
> image
> 
> Also for some reason the labels for page size and quick search (the magnifying glass) all have button-esque styling; they darken on hover, change the cursor to indicate they're click-able, and you can even click them, which looks like clicking a button. But they're actually not click-able in any meaningful sense.

so `input[number]` with a min=1 won't actually prevent a user from entering a 0 or -900, for example, and wasn't really interested in adding additional code to prevent something that a user should not be / probably won't be doing. why would they enter page size=-900? and ag-grid is smart enough to ignore anything less than 1 and just use the default page size of 100. so yeah, a bit weird but not if user doesn't do weird things :)

mitchell852:
> hmm. setting page size doing nothing for you? yeah, you're right. if you set it to -900 and then 10, it's broke.

nm. it's just broken. now sure when/how i broke it.

mitchell852:
> > hmm. setting page size doing nothing for you? yeah, you're right. if you set it to -900 and then 10, it's broke.
> 
> nm. it's just broken. now sure when/how i broke it.

looks like the step is breaking it. removing that.

mitchell852:
> Also for some reason the labels for page size and quick search (the magnifying glass) all have button-esque styling; they darken on hover, change the cursor to indicate they're click-able, and you can even click them, which looks like clicking a button. But they're actually not click-able in any meaningful sense.

removing those "buttons"

PR #4892, opened 2020-07-17
/cdns/capacity returns 500 error: capacity was zero

### LINKED ISSUE #4848: /cdns/capacity returns 500 error: capacity was zero
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

Visiting `/api/3.0/cdns/capacity` returns a 500 error:

```json
{
	"alerts": [
		{
			"level": "error"
		}
	]
}
```

## Expected / new behavior:

Should return a response like

```json
{
	"response": {
		"availablePercent": 89.0939840205533,
		"unavailablePercent": 0,
		"utilizedPercent": 10.9060020300395,
		"maintenancePercent": 1.39494071146245e-05
	}
}
```

1. Start CDN-in-a-Box
2. Log into Traffic Ops
3. Visit https://trafficops.infra.ciab.test/api/3.0/cdns/capacity

## Anything else:

From the error log:
```json
ERROR: api.go:201: 2020-07-01T17:32:06.388024754Z: 192.168.48.10:55050 capacity was zero!
```

srijeet0406:

- [x] This PR fixes #4848  

- CDN in a Box
- Traffic Ops

Two types of tests:
1.) Test in a system where you dont have any monitors for your cdns. In such a system, if you make a call to GET `/cdns/capacity`, it should return a 500, and the error logs should show the reason of the failure as `no monitors available`
2.) Test it in a system that has monitors, etc. (a full setup, like CiaB). Do a snapshot of one of the cdns, and then try to GET the `/cdns/capacity`. You should get a valid response this time.

- master

- [x] This PR includes tests


traffic_ops/traffic_ops_golang/cdn/capacity.go
```diff
@@ -117,10 +119,22 @@ func getMonitorsCapacity(tx *sql.Tx, monitors map[tc.CDNName][]string) (Capacity
 	}, nil
 }
 
+func checkServiceInterface(tx *sql.Tx, name tc.InterfaceName, cacheName tc.CacheName) bool {
+	var result *bool
+	if err := tx.QueryRow(`SELECT i.service_address FROM ip_address AS i JOIN server as s on s.id = i.server WHERE i.interface=$1 AND s.host_name=$2`, name, cacheName).Scan(&result); err != nil {
```
zrhoffman, line 124: Please reformat this query to have more line breaks for readability.

traffic_ops/traffic_ops_golang/cdn/capacity.go
```diff
@@ -166,21 +181,39 @@ func addCapacity(cap CapData, cacheStats CacheStats, crStates tc.CRStates, crCon
...
 			continue
 		}
-		if len(stats.KBPS) < 1 || len(stats.MaxKBPS) < 1 {
-			continue
+		var kbps, maxKbps float64
+		for intName, intStats := range stats {
+			if len(intStats.KBPS) < 1 || len(intStats.MaxKBPS) < 1 {
+				continue
+			}
+			if checkServiceInterface(tx, intName, cacheName) {
```
zrhoffman, line 189: Can this *if* statement's be extracted to save a level of indentation?

Currently:
```golang
if condition {
    /* several lines of code */
}
```

Restructured:
```golang
if ! condition {
   continue
}
 /* several lines of code */
```

zrhoffman, line 189: Can `checkServiceInterface()` be rewritten so it only needs to be called a single time instead of once per cache?

traffic_ops/traffic_ops_golang/cdn/capacity.go
```diff
@@ -117,10 +119,22 @@ func getMonitorsCapacity(tx *sql.Tx, monitors map[tc.CDNName][]string) (Capacity
...
+func checkServiceInterface(tx *sql.Tx, name tc.InterfaceName, cacheName tc.CacheName) bool {
+	var result *bool
+	if err := tx.QueryRow(`SELECT i.service_address FROM ip_address AS i JOIN server as s on s.id = i.server WHERE i.interface=$1 AND s.host_name=$2`, name, cacheName).Scan(&result); err != nil {
+		log.Errorln("Error checking for service interface ", err)
+		return false
+	}
+	if *result == true {
+		return true
+	}
+	return false
```
zrhoffman, line 131: Since we're returning `true` when `*result` is `true` and `false` otherwise, this can just be

```golang
return *result
```

lib/go-tc/enum.go
```diff
@@ -55,6 +55,9 @@ type DeliveryServiceName string
 // CacheType is the type (or tier) of a CDN cache.
 type CacheType string
 
+// InterfaceName is the name of the interface
```
zrhoffman, line 58: This doesn't help a user who doesn't already know what interfaces are and how they are used. Maybe *server interface*?

traffic_ops/traffic_ops_golang/cdn/capacity.go
srijeet0406: You mean once per interface, or once per cache?

zrhoffman: Yes I meant instead of once per interface. Ideally it will only be called once overall per request before entering the cache loop. Workflow could be:

1. From your list of cache servers, get all interfaces by server. `ARRAY_AGG` can be used to list more than 1 interface in a single column so that it's only 1 row per server (with no performance penalty)

2. Store the result to a struct

3. In your interface loop, check the interface using that struct as the data source

srijeet0406: Instead of doing a separate sql query to get the interfaces, we can just make use of the `stats` structure tha we have, where we already have all the interface names. Saves us a DB call. But yeah, I can make it so that the `checkServiceInterface()` function will not need to be called once per interface. Does that make sense?

zrhoffman: Sounds good to me

srijeet0406:
@zrhoffman About the query performance, we have to traverse over a list of interfaces to get the service interface, because the stats associated with the service interface are the only stats that we are interested in. We dont really want any of the other ones. So, even if we get everything except the `aggregate` key, we would still have to run through that list and get the stats associated with only the service interface.

zrhoffman:
/cdns/capacity works, API tests pass, unit tests pass, code is formatted correctly. LGTM!

PR #4907, opened 2020-07-23
Add verification for TO server ips across a profile

### LINKED ISSUE #4888: Can create multiple Servers using the same profile with the same ip address for their service address
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

So right now, user can create multiple server with same profile with same ip address for their service address.
## Expected / new behavior:

It should not allow user to create the second server with same ip address for the same profile.

shamrickus:

- [x] This PR fixes #4888 

- Traffic Ops
- Traffic Portal

Run the TO/TP tests.
Ensure that you cannot update a server to have the same service ip address as another server with the same profile. Both when creating and updating a server.

### DIFF
--- a/traffic_ops/app/db/migrations/20200811082611_add_server_ip_profile_trigger.sql
+++ b/traffic_ops/app/db/migrations/20200811082611_add_server_ip_profile_trigger.sql
@@ -1,0 +1,124 @@
+/*
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION before_server_table()
+    RETURNS TRIGGER AS
+$$
+DECLARE
+    server_count BIGINT;
+BEGIN
+    WITH server_ips AS (
+        SELECT s.id, i.name, ip.address, s.profile
+        FROM server s
+                 JOIN interface i on i.server = s.ID
+                 JOIN ip_address ip on ip.Server = s.ID and ip.interface = i.name
+        WHERE i.monitor = true
+    )
+    SELECT count(*)
+    INTO server_count
+    FROM server_ips sip
+             JOIN server_ips sip2 on sip.id <> sip2.id
+    WHERE sip.id = NEW.id
+      AND sip2.address = sip.address
+      AND sip2.profile = sip.profile;
+
+    IF server_count > 0 THEN
+        RAISE EXCEPTION 'Server [id:%] does not have a unique ip_address over the profile [id:%], [%] conflicts',
+            NEW.id,
+            NEW.profile,
+            server_count;
+    END IF;
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+CREATE OR REPLACE FUNCTION before_ip_address_table()
+    RETURNS TRIGGER
+AS
+$$
+DECLARE
+    server_count   BIGINT;
+    server_id      BIGINT;
+    server_profile BIGINT;
+BEGIN
+    WITH server_ips AS (
+        SELECT s.id as sid, ip.interface, i.name, ip.address, s.profile, ip.server
+        FROM server s
+                 JOIN interface i
+                     on i.server = s.ID
+                 JOIN ip_address ip
+                     on ip.Server = s.ID and ip.interface = i.name
+        WHERE i.monitor = true
+    )
+    SELECT count(sip.sid), sip.sid, sip.profile
+    INTO server_count, server_id, server_profile
+    FROM server_ips sip
+             JOIN server_ips sip2 on sip.sid <> sip2.sid
+    WHERE (sip.server = NEW.server AND sip.address = NEW.address AND sip.interface = NEW.interface)
+      AND sip2.address = sip.address
+      AND sip2.profile = sip.profile
+    GROUP BY sip.sid, sip.profile;
+
+    IF server_count > 0 THEN
+        RAISE EXCEPTION 'ip_address is not unique accross the server [id:%] profile [id:%], [%] conflicts',
+            server_id,
+            server_profile,
+            server_count;
+    END IF;
+    RETURN NEW;
+END;
+$$ LANGUAGE PLPGSQL;
+-- +goose StatementEnd
+
+CREATE TRIGGER before_update_server_trigger
+    BEFORE UPDATE
+    ON server
+    FOR EACH ROW
+    WHEN (NEW.profile <> OLD.profile)
+EXECUTE PROCEDURE before_server_table();
+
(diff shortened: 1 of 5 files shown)


traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -16,13 +16,19 @@ package v3
 */
 
 import (
+
```
zrhoffman, line 19: `gofmt` removes this line

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -16,13 +16,19 @@ package v3
...
 	"net/http"
 	"net/url"
 	"strconv"
 	"testing"
 	"time"
+
+	"github.com/apache/trafficcontrol/lib/go-util"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+
```
zrhoffman, line 30: These newlines between ATC package imports can be removed

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -26,14 +26,15 @@ import (
...
-	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/util/ims"
 	"net"
 	"net/http"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/apache/trafficcontrol/lib/go-rfc"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/util/ims"
+
```
zrhoffman, line 37: These newlines between ATC package imports can be removed

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -485,6 +490,39 @@ func validateV3(s *tc.ServerNullable, tx *sql.Tx) (string, error) {
...
+	} else if rows != nil {
+		defer rows.Close()
+		for rows.Next() {
+			var id int
+			var ipaddress string
+			err = rows.Scan(&id, &ipaddress)
+			if err != nil {
+				errs = append(errs, errors.New("unable to determine service address uniqueness"))
+			} else if (ipaddress == ipv4 || ipaddress == ipv6) && (s.ID == nil || *s.ID != id) {
+				errs = append(errs, errors.New("there exists a server on the same profile that has the same service address"))
```
zrhoffman, line 521: This error should include the IP addresses and IDs being compared.

zrhoffman:
We used to have a database constraint for each set of server IPs and profile being unique, (lost when we moved to using the `ip_address` table for IP addresses). This PR should bring back that constraint.

shamrickus:
I see no reference to that constraint in our migrations?

zrhoffman:
What we had before removing `ip_address` and `ip6_address` from the `server` table (from [traffic_ops/app/db/create_tables.sql#L2412](https://github.com/apache/trafficcontrol/blob/0cf375d2b1/traffic_ops/app/db/create_tables.sql#L2412)):

```pgsql
--
-- Name: idx_89709_ip6_profile; Type: INDEX; Schema: public; Owner: traffic_ops
--

CREATE UNIQUE INDEX idx_89709_ip6_profile ON server USING btree (ip6_address, profile);


--
-- Name: idx_89709_ip_profile; Type: INDEX; Schema: public; Owner: traffic_ops
--

CREATE UNIQUE INDEX idx_89709_ip_profile ON server USING btree (ip_address, profile);
```

shamrickus:
Should be good to go now.

traffic_ops/app/db/migrations/20200811082611_add_server_ip_profile_trigger.sql
```diff
@@ -0,0 +1,124 @@
...
+
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+DROP FUNCTION IF EXISTS before_server_table();
+DROP FUNCTION IF EXISTS before_ip_address_table();
+
+DROP TRIGGER IF EXISTS before_update_server_trigger ON server;
+DROP TRIGGER IF EXISTS before_create_server_trigger ON server;
+DROP TRIGGER IF EXISTS before_update_ip_address_trigger ON ip_address;
+DROP TRIGGER IF EXISTS before_create_ip_address_trigger ON ip_address;
```
zrhoffman, line 123: Goose down fails

```
2020/08/14 20:41:51 FAIL 20200811082611_add_server_ip_profile_trigger.sql (pq: cannot drop function before_server_table() because other objects depend on it), quitting migration.
```

mitchell852: @shamrickus ^^

shamrickus: Will fix

shamrickus:
This change does not add new functionality but port over existing functionality to our [multi-interface initiative](https://github.com/apache/trafficcontrol/blob/master/blueprints/multi-interface-servers.md) (which will be released in ATCv5.0). There already existed a primary key on the server table across [`ip_address` and `profile`](https://github.com/apache/trafficcontrol/blob/0cf375d2b1/traffic_ops/app/db/create_tables.sql#L2415), so if it was working before it should still work now; at least in relation to this PR. That having been said I'm not sure I understand your setup, and it may be that we have made changes in multi-interface such-that your current setup won't work. What version of ATC are you currently using?

bdgranger:
We are currently running ATC 3.1.0 but plan to upgrade in the near future. In our current setup, we are able to create multiple servers in the traffic ops UI, all of which use the same IP address and the same profile, but have different hostnames and different tcpPort and httpsPort, so they can be individually communicated with, even though they are running on the same docker node and share the same external IP address.

I was only concerned since the entire point of #4888 seemed to be to prevent multiple servers with the same IP address.

mitchell852:
@bdgranger - hmm. i'm not really sure when the IP + profile unique constraint was added to the server table (i guess after 3.1 since you can apparently do it in 3.1) and many have questioned it's value. I do know this is being worked on #4940 and will hopefully be in TC 5.0.

ocket8888:
ATCv3.1 has [a constraint that won't let you have multiple servers with the same IP and Profile](https://github.com/apache/trafficcontrol/blob/3.1.x/traffic_ops/app/db/create_tables.sql#L2131) so I'm not sure how you got that set up that way.

bdgranger:
@ocket8888 @mitchell852 - The system was originally running ATC2.2 and that was where I started this work and was able to create the multiple servers with the same IP and profile.  However, when I upgraded to ATC 3.1, it didn't complain about the multiple servers with the same IP address, and even now on the ATC3.1 system (which was upgraded from 2.2) I am still able to create new servers with the same IP address and profile as existing servers.  I wonder if the upgrade somehow didn't apply the constraint change? Not that I'm complaining, since I'm relying on that behavior currently :-)  I guess I'll have to try a "clean" ATC3.1 install and see what happens.

Looks like #4940 is indeed what I really want, if any type of constraint is to be enforced. Is there any estimated date for the TC 5.0 release yet?

mitchell852:
@bdgranger - good question about 5.0 release date. As a community, we were striving for quarterly releases. 4.1 was end of Q1. 5.0 was supposed to be end of Q2 (june 30/july 1) but that didn't happen due to a couple of big features ([flexible topologies](https://github.com/apache/trafficcontrol/blob/master/blueprints/flexible-topologies.md) and [multi-interface support](https://github.com/apache/trafficcontrol/blob/master/blueprints/multi-interface-servers.md)) so hopefully pretty soon?? @ocket8888 is the 5.0 release manager.

PR #4909, opened 2020-07-24
Use TO client library API v3 for the CDN-in-a-Box enroller service

zrhoffman:

When #4645 bumped the latest Traffic Ops API version to 3.0, some ATC components continued using v2.0 for compatibility reasons. This PR updates the CDN-in-a-Box Enroller component to use the v3 version of the TO Go client library. 

This involved dynamically generating server data for the `dns`, `db`, and `enroller` services, since we cannot predict their IP addresses and cannot use pre-generated server data as we have previously done for those services.

- CDN in a Box

Verify that the CDN-in-a-Box delivery service works as intended using the *readiness* container, which indicates that everything was enrolled successfully.
```shell script
docker-compose up -d
docker-compose -f docker-compose.readiness.yml up
```

- [x] CDN-in-a-Box PR, tests are unnecessary beyond the *What is the best way to verify this PR?* section
- [x] CDN-in-a-Box usage is unchanged, no documentation necessary


infrastructure/cdn-in-a-box/traffic_ops/run-db.sh
```diff
@@ -18,5 +18,11 @@
...
 set-dns.sh
 insert-self-into-dns.sh
+set -o monitor #enable job control
+/docker-entrypoint.sh postgres &
 
-exec /docker-entrypoint.sh postgres
+source /to-access.sh
+# Enroll with traffic ops
+TO_URL="https://$TO_FQDN:$TO_PORT"
+to-enroll db ALL '' 5432 5432 || (while true; do echo "enroll failed."; sleep 3 ; done)
```
ocket8888, line 27: I just wanna call this out here; I know this is "standard" in that it's what our other containers do, but we should seriously consider just failing if enrollment fails on non-optional containers.

PR #4746, opened 2020-06-01
Update TM UI for multiple interfaces

shamrickus:
## What does this PR (Pull Request) do?

- [x] This PR is not related to any Issue. 

The Cache Stats tab of TM UI has been updated to reflect multiple interfaces. Each cache server can have any number of interfaces, as such each Cache Stat row now features a hidden table containing the information about it's interfaces. This can be seen (and hidden) by clicking on the parent row.

## Which Traffic Control components are affected by this PR?

- Traffic Monitor

## What is the best way to verify this PR?
Requires #4690 to work

Verify TM UI functions properly and that all interface information can be viewed for each server.  

## The following criteria are ALL met by this PR

- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR ensures that database migration sequence is correct OR this PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #4916, opened 2020-07-27
Multi interface health

### LINKED ISSUE #4618: Enable Traffic Monitor to read in multiple interfaces for a cache server
## I'm submitting a ...
- new feature / enhancement request

## Description
Traffic Monitor needs to be made capable of reading in the new monitoring data format as described in the Additional Server Interfaces blueprint. It must be capable of honoring the polling behavior defined by *either* format.

- Traffic Ops

## Acceptance Criteria
* Traffic Monitor is capable of reading in the new data representations of cache servers as described in the blueprint
	* There are associated - and passing - unit tests that prove this.
* Traffic Monitor is still capable of reading in the legacy data representations of cache servers
	* There are associated - and passing - unit tests that prove this.
* Traffic Monitor respects the new per-interface maximum bandwidth settings
	* There are associated - and passing - tests that prove this.
* Traffic Monitor still respects the normal Parameter-based thresholds
	* There are associated - and passing - tests that prove this.
* The Traffic Monitor UI has been updated to display information about all monitored interfaces of a cache server, as opposed to the "IPv4" and "IPv6" columns of today.
* The new polling behavior - RE: considering multiple interfaces - has been documented.

See the [Additional Server Interfaces blueprint](https://github.com/apache/trafficcontrol/blob/master/blueprints/multi-interface-servers.md#traffic-monitor-impact) for details.

ocket8888:

- [x] This PR fixes #4618

This removes the temporary shim "aggregate" interface used by Traffic Monitor - and Traffic Ops to a much lesser degree - and adds the ability to truly monitor cache servers according to both aggregate statistics as well as network interface thresholds.

Because interface stats now appear in cache stats TM API responses, this also introduces an equivalent to the `stats` query parameter used to filter general cache stats called `interfaceStats`.

This PR also properly wraps use of the Traffic Ops Go client in Traffic Monitor - for tasks requiring APIv3-format data (there are still some places where APIv2 is used explicitly) - such that if login fails with the current API version (3), login will be tried with the older version (2.0) and, if successful, all future requests will use API v2.0, coercing legacy data into modern structures as necessary - and there are unit tests for those coercions.

It also consolidates double-definitions of cache server interface information and IP address structures into one, and consequently moved some functionality between files as appropriate.

It also adds support for signed and unsigned integral values of cache statistics, which previously only allowed floating point-type numbers.

It also adds some table styling to the TM UI to fix error rows not highlighted in red - that's just a temporary, small workaround to ease testing of this PR, and a better fix is coming in #4746.

It also fixes a bug in a "CRConfig" Snapshot API unit test where the test was using servers without any IP addresses - it wasn't failing, but was merely not actually testing as much as it ought to have been.

It also fixes a bug when converting server interface data to the legacy format which would cause them to report an incorrect interface name under certain circumstances, adding a test.

It also fixes a bug parsing [`lib/go-tc.TMParameters`](https://godoc.org/github.com/apache/trafficcontrol/lib/go-tc#TMParameters) structures from JSON with certain comparators in health thresholds, adding a test.

Finally, it also adds a load of GoDoc comments to objects that previously did not have them, and several tests for things that did not previously have those.

- Documentation
- Traffic Monitor
- Traffic Ops

Run all unit tests - for `/lib`, Traffic Monitor, and Traffic Ops. Also ensure all of the Traffic Ops API tests - versions 1-3 - work and pass.

To test the interface monitoring behavior, given some set of cache servers monitored by the Traffic Monitor instance being tested, add dummy interfaces you're sure won't be in the returned health data from the cache servers' health polling endpoints. Ensure that when "monitor" is not true for this/these interface(s) that they don't affect health status setting for the cache servers. Ensure that when this changes ("monitor" is true) the Traffic Monitor marks the cache servers as unhealthy because it can't find any data for the interfaces in the polling data. Also, check that "maxBandwidth" limits on each actually poll-able interface is respected, by setting it to something lower than the interface is actually serving (I use zero, because that's nearly impossible to reach) such that cache servers are marked unavailable when they are exceeded, and are not - at least not for that reason - when they are not exceeded. Finally, check that there is no setting of "maxBandwidth" on any interface with "monitor" set to false that will result in the cache server being marked unavailable (again, I use zero).

- master

- [x] This PR includes tests

### DIFF
--- a/docs/source/api/v3/cdns_name_configs_monitoring.rst
+++ b/docs/source/api/v3/cdns_name_configs_monitoring.rst
@@ -76,7 +76,7 @@ Response Structure
 
 		:health.connection.timeout:                 A timeout value, in milliseconds, to wait before giving up on a health check request
 		:health.polling.url:                        A URL to request for polling health. Substitutions can be made in a shell-like syntax using the properties of an object from the ``"trafficServers"`` array
-		:health.threshold.availableBandwidthInKbps: The total amount of bandwidth that servers using this profile are allowed - across all network interfaces -, in Kilobits per second. This is a string and using comparison operators to specify ranges, e.g. ">10" means "more than 10 kbps"
+		:health.threshold.availableBandwidthInKbps: The total amount of bandwidth that servers using this profile are allowed - across all network interfaces - in Kilobits per second. This is a string and using comparison operators to specify ranges, e.g. ">10" means "more than 10 kbps"
 		:health.threshold.loadavg:                  The UNIX loadavg at which the server should be marked "unhealthy"
 
 			.. seealso:: :manpage:`uptime(1)`
--- a/docs/source/api/v3/cdns_name_configs_monitoring.rst
+++ b/docs/source/api/v3/cdns_name_configs_monitoring.rst
@@ -241,7 +241,6 @@ Response Structure
 					"health.connection.timeout": 2000,
 					"health.polling.url": "http://${hostname}/_astats?application=&inf.name=${interface_name}",
 					"health.threshold.availableBandwidthInKbps": ">1750000",
-					"health.threshold.aggregate.availableBandwidthInKbps": ">17500000",
 					"health.threshold.loadavg": "25.0",
 					"health.threshold.queryTime": 1000,
 					"history.count": 30
@@ -254,7 +253,6 @@ Response Structure
 					"health.connection.timeout": 2000,
 					"health.polling.url": "http://${hostname}/_astats?application=&inf.name=${interface_name}",
 					"health.threshold.availableBandwidthInKbps": ">1750000",
-					"health.threshold.aggregate.availableBandwidthInKbps": ">17500000",
 					"health.threshold.loadavg": "25.0",
 					"health.threshold.queryTime": 1000,
 					"history.count": 30
--- a/lib/go-tc/servers.go
+++ b/lib/go-tc/servers.go
@@ -286,10 +286,14 @@ func InterfaceInfoToLegacyInterfaces(serverInterfaces []ServerInterfaceInfo) (Le
 
 	for _, intFace := range serverInterfaces {
 
+		foundServiceInterface := false
+
 		for _, addr := range intFace.IPAddresses {
 			if !addr.ServiceAddress {
 				continue
 			}
+
+			foundServiceInterface = true
 
 			address := addr.Address
 			gateway := addr.Gateway
@@ -322,17 +326,26 @@ func InterfaceInfoToLegacyInterfaces(serverInterfaces []ServerInterfaceInfo) (Le
 				legacyDetails.InterfaceMtu = util.IntPtr(int(*intFace.MTU))
 			}
 
-			legacyDetails.InterfaceName = &intFace.Name
+			// This should no longer matter now that short-circuiting is better,
+			// but this temporary variable is necessary because the 'intFace'
+			// variable is referential, so taking '&intFace.Name' would cause
+			// problems when intFace is reassigned.
+			name := intFace.Name
+			legacyDetails.InterfaceName = &name
 
 			// we can jump out here since servers can only legally have one
 			// IPv4 and one IPv6 service address
 			if legacyDetails.IPAddress != nil && *legacyDetails.IPAddress != "" && legacyDetails.IP6Address != nil && *legacyDetails.IP6Address != "" {
-				return legacyDetails, nil
+				break
 			}
+		}
+
+		if foundServiceInterface {
+			return legacyDetails, nil
 		}
 	}
 
-	return legacyDetails, nil
+	return legacyDetails, errors.New("no service addresses found")
 }
 
 type Server struct {
--- a/lib/go-tc/servers.go
+++ b/lib/go-tc/servers.go
@@ -420,6 +420,75 @@ type ServerNullableV2 struct {
 	IP6IsService *bool `json:"ip6IsService" db:"ip6_address_is_service"`
 }
 
+// ToNullable converts the Server to an equivalent, "nullable" structure.
+//
+// Note that "zero" values (e.g. the empty string "") are NOT coerced to actual
+// null values. In particular, the only fields that will possibly be nil are
+// FQDN - if the original server had a nil FQDN - and DeliveryServices - which
+// will actually be a pointer to a nil map if the original server had a nil
+// DeliveryServices map.
+// Further note that this makes "shallow" copies of member properties; if
+// reference types (map, slice, pointer etc.) are altered on the original after
+// conversion, the changes WILL affect the nullable copy.
+func (s Server) ToNullable() ServerNullableV2 {
+	return ServerNullableV2{
+		ServerNullableV11: ServerNullableV11{
+			CommonServerProperties: CommonServerProperties{
+				Cachegroup:       &s.Cachegroup,
+				CachegroupID:     &s.CachegroupID,
+				CDNID:            &s.CDNID,
+				CDNName:          &s.CDNName,
+				DeliveryServices: &s.DeliveryServices,
+				DomainName:       &s.DomainName,
+				FQDN:             s.FQDN,
+				FqdnTime:         s.FqdnTime,
+				GUID:             &s.GUID,
+				HostName:         &s.HostName,
+				HTTPSPort:        &s.HTTPSPort,
+				ID:               &s.ID,
+				ILOIPAddress:     &s.ILOIPAddress,
+				ILOIPGateway:     &s.ILOIPGateway,
+				ILOIPNetmask:     &s.ILOIPNetmask,
+				ILOPassword:      &s.ILOPassword,
+				ILOUsername:      &s.ILOUsername,
+				LastUpdated:      &s.LastUpdated,
+				MgmtIPAddress:    &s.MgmtIPAddress,
+				MgmtIPGateway:    &s.MgmtIPGateway,
+				MgmtIPNetmask:    &s.MgmtIPNetmask,
+				OfflineReason:    &s.OfflineReason,
+				PhysLocation:     &s.PhysLocation,
+				PhysLocationID:   &s.PhysLocationID,
+				Profile:          &s.Profile,
+				ProfileDesc:      &s.ProfileDesc,
+				ProfileID:        &s.ProfileID,
+				Rack:             &s.Rack,
+				RevalPending:     &s.RevalPending,
+				RouterHostName:   &s.RouterHostName,
+				RouterPortName:   &s.RouterPortName,
+				Status:           &s.Status,
+				StatusID:         &s.StatusID,
+				TCPPort:          &s.TCPPort,
+				Type:             s.Type,
+				TypeID:           &s.TypeID,
+				UpdPending:       &s.UpdPending,
+				XMPPID:           &s.XMPPID,
+				XMPPPasswd:       &s.XMPPPasswd,
+			},
+			LegacyInterfaceDetails: LegacyInterfaceDetails{
+				InterfaceMtu:  &s.InterfaceMtu,
+				InterfaceName: &s.InterfaceName,
+				IPAddress:     &s.IPAddress,
+				IPGateway:     &s.IPGateway,
+				IPNetmask:     &s.IPNetmask,
+				IP6Address:    &s.IP6Address,
+				IP6Gateway:    &s.IP6Gateway,
+			},
+		},
+		IPIsService:  &s.IPIsService,
+		IP6IsService: &s.IP6IsService,
+	}
+}
+
 // ServerNullable represents an ATC server, as returned by the TO API.
 type ServerNullable struct {
 	CommonServerProperties
--- a/lib/go-tc/servers.go
+++ b/lib/go-tc/servers.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-util"
@@ -203,6 +205,76 @@ func (lid *LegacyInterfaceDetails) ToInterfaces(ipv4IsService, ipv6IsService boo
 
 	iface.IPAddresses = ips
 	return []ServerInterfaceInfo{iface}, nil
+}
+
+// String implements the fmt.Stringer interface.
+func (lid LegacyInterfaceDetails) String() string {
+	var b strings.Builder
+	b.Write([]byte("LegacyInterfaceDetails(InterfaceMtu="))
+
+	if lid.InterfaceMtu == nil {
+		b.Write([]byte("nil"))
+	} else {
+		b.WriteString(strconv.FormatInt(int64(*lid.InterfaceMtu), 10))
+	}
+
+	b.Write([]byte(", InterfaceName="))
+	if lid.InterfaceName != nil {
+		b.WriteRune('\'')
+		b.WriteString(*lid.InterfaceName)
+		b.WriteRune('\'')
+	} else {
+		b.Write([]byte("nil"))
+	}
+
+	b.Write([]byte(", IP6Address="))
+	if lid.IP6Address != nil {
+		b.WriteRune('\'')
+		b.WriteString(*lid.IP6Address)
+		b.WriteRune('\'')
+	} else {
+		b.Write([]byte("nil"))
+	}
+
+	b.Write([]byte(", IP6Gateway="))
+	if lid.IP6Gateway != nil {
+		b.WriteRune('\'')
+		b.WriteString(*lid.IP6Gateway)
+		b.WriteRune('\'')
+	} else {
+		b.Write([]byte("nil"))
+	}
+
+	b.Write([]byte(", IPAddress="))
+	if lid.IPAddress != nil {
+		b.WriteRune('\'')
+		b.WriteString(*lid.IPAddress)
+		b.WriteRune('\'')
+	} else {
+		b.Write([]byte("nil"))
+	}
+
+	b.Write([]byte(", IPGateway="))
+	if lid.IPGateway != nil {
+		b.WriteRune('\'')
+		b.WriteString(*lid.IPGateway)
+		b.WriteRune('\'')
+	} else {
+		b.Write([]byte("nil"))
+	}
+
+	b.Write([]byte(", IPNetmask="))
+	if lid.IPNetmask != nil {
+		b.WriteRune('\'')
+		b.WriteString(*lid.IPNetmask)
+		b.WriteRune('\'')
+	} else {
+		b.Write([]byte("nil"))
+	}
+
+	b.WriteRune(')')
+
+	return b.String()
 }
 
 // InterfaceInfoToLegacyInterfaces converts a ServerInterfaceInfo to an
--- a/lib/go-tc/servers.go
+++ b/lib/go-tc/servers.go
@@ -489,6 +489,32 @@ func (s Server) ToNullable() ServerNullableV2 {
 	}
 }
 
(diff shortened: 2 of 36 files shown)


docs/source/api/v3/cdns_name_configs_monitoring.rst
```diff
@@ -76,7 +76,7 @@ Response Structure
 
 		:health.connection.timeout:                 A timeout value, in milliseconds, to wait before giving up on a health check request
 		:health.polling.url:                        A URL to request for polling health. Substitutions can be made in a shell-like syntax using the properties of an object from the ``"trafficServers"`` array
-		:health.threshold.availableBandwidthInKbps: The total amount of bandwidth that servers using this profile are allowed, in Kilobits per second. This is a string and using comparison operators to specify ranges, e.g. ">10" means "more than 10 kbps"
+		:health.threshold.availableBandwidthInKbps: The total amount of bandwidth that servers using this profile are allowed - across all network interfaces -, in Kilobits per second. This is a string and using comparison operators to specify ranges, e.g. ">10" means "more than 10 kbps"
```
mattjackson220, line 79: remove comma after - at `interfaces -, in Kilobits`

traffic_monitor/health/cache.go
```diff
@@ -144,177 +191,151 @@ func EvalCache(result cache.ResultInfo, resultStats *threadsafe.ResultStatValHis
...
 		serverInfo, ok := mc.TrafficServer[result.ID]
 		if !ok {
 			log.Errorf("Cache %v missing from from Traffic Ops Monitor Config - treating as OFFLINE\n", result.ID)
 		}
 
-		if _, ok := localCacheStatuses[tc.CacheName(result.ID)]; !ok {
-			localCacheStatuses[tc.CacheName(result.ID)] = make(map[string]cache.AvailableStatus)
+		availStatus := cache.AvailableStatus{
+			Available: cache.AvailableTuple{
+				IPv4: result.UsingIPv4,
```
mattjackson220, line 257: the way this is written, IPv4 and IPv6 will never both be labeled as available at the same time. if we are checking ipv4 then should we set IPv6 based on localCacheStatuses[result.ID].Available.IPv6 and vice versa?

lib/go-tc/traffic_monitor_test.go
```diff
@@ -0,0 +1,442 @@
...
+		t.Logf("Got expected error: checking validity of config map with nil TrafficServer: %v", err)
+	}
+
+	mc.TrafficServer = map[string]TrafficServer{}
+	err = mc.Valid()
+	if err == nil {
+		t.Error("Didn't get expected error checking validity of config map with no TrafficServers")
+	} else {
+		t.Logf("Got expected error: checking validity of config map with no TrafficServers: %v", err)
+	}
```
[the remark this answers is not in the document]
ocket8888: The test for a valid case is up on lines 60 thru 76

traffic_monitor/health/cache.go
ocket8888: You'd know better than me, I was never clear on how that's supposed to work.

traffic_monitor/health/cache.go
```diff
@@ -144,177 +191,151 @@ func EvalCache(result cache.ResultInfo, resultStats *threadsafe.ResultStatValHis
...
-// CalcAvailabilityWithStats calculates the availability of each cache in results.
-// statResultHistory may be nil, in which case stats won't be used to calculate availability.
-func CalcAvailability(results []cache.Result, pollerName string, statResultHistory *threadsafe.ResultStatHistory, mc tc.LegacyTrafficMonitorConfigMap, toData todata.TOData, localCacheStatusThreadsafe threadsafe.CacheAvailableStatus, localStates peer.CRStatesThreadsafe, events ThreadsafeEvents, protocol config.PollingProtocol) {
-	localCacheStatuses := localCacheStatusThreadsafe.Get().Copy()
-	statResults := (*threadsafe.ResultStatValHistory)(nil)
-	statResultsVal := (*map[string]threadsafe.ResultStatValHistory)(nil)
-	processAvailableTuple := func(tuple cache.AvailableTuple, serverInfo tc.LegacyTrafficServer) bool {
-		switch protocol {
-		case config.IPv4Only:
+// gets a function to process an availability tuple based on the protocol used.
```
shamrickus, line 201: Missing function name for godocs

traffic_monitor/towrap/towrap.go
```diff
@@ -124,59 +157,114 @@ func (h CRConfigHistoryThreadsafe) Add(i *CRConfigStat) {
...
-// TrafficOpsSessionThreadsafe provides access to the Traffic Ops client safe for multiple goroutines. This fulfills the ITrafficOpsSession interface.
+// TrafficOpsSessionThreadsafe provides access to the Traffic Ops client safe
+// for multiple goroutines. This fulfills the ITrafficOpsSession interface.
 type TrafficOpsSessionThreadsafe struct {
 	session            **client.Session // pointer-to-pointer, because we're given a pointer from the Traffic Ops package, and we don't want to copy it.
+	legacySession      **legacyClient.Session
 	m                  *sync.Mutex
 	lastCRConfig       ByteMapCache
 	crConfigHist       CRConfigHistoryThreadsafe
+	useLegacy          bool
```
shamrickus, line 196: What is the point of this and `legacySession`? It doesn't appear to actually get used anywhere.

traffic_monitor/manager/opsconfig.go
```diff
@@ -169,42 +164,26 @@ func StartOpsConfigManager(
...
-
-					realToSession = to.NewSession(newOpsConfig.Username, newOpsConfig.Password, newOpsConfig.Url, staticAppData.UserAgent, &http.Client{
-						Timeout: trafficOpsRequestTimeout,
-						Transport: &http.Transport{
-							TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-						},
-						Jar: jar,
-					}, useCache)
-					toSession.Set(realToSession)
-					// At this point we have a valid 'dummy' session. This will allow us to pull from disk but will also retry when TO comes up
```
shamrickus, line 194: Are you sure removing this won't break TM when TO is unaccessible but the backup file exists?

traffic_monitor/static/style.css
```diff
@@ -44,11 +44,11 @@ tbody tr:nth-child(odd) {
 	background: #fff;
 }
 .error {
-	background-color: #f00;
+	background-color: #f00!important;
 }
 
 .warning {
-	background-color: #f80;
```
shamrickus, line 51: Is this fixing a bug?

traffic_monitor/threadsafe/resultstathistory.go
```diff
@@ -55,55 +64,212 @@ func (h *ResultInfoHistory) Get() cache.ResultInfoHistory {
...
+	}
+
+	ok, err := newStatEqual(history, stat.Stat)
+	if err != nil {
+		errs.WriteString(stat.InterfaceName)
+		errs.Write([]byte(": cannot add stat "))
+		errs.WriteString(stat.StatName)
+		errs.Write([]byte(": "))
+		errs.WriteString(err.Error())
+		errs.Write([]byte("; "))
```
shamrickus, line 143: Any reason not to use `errs.Write([]bytes(fmt.Sprintf(...))` instead?

traffic_monitor/threadsafe/resultstathistory.go
```diff
@@ -55,55 +64,212 @@ func (h *ResultInfoHistory) Get() cache.ResultInfoHistory {
...
-		return true
+		return f(cacheName, i)
 	})
 }
 
-// ResultStatValHistory is threadsafe for one writer. Specifically, because a CompareAndSwap is not provided, it's not possible to Load and Store without a race condition.
-// If multiple writers were necessary, it wouldn't be difficult to add a CompareAndSwap, internally storing an atomically-accessed pointer to the slice.
+// This is just a convenience structure used only for passing data about a
+// single statistic for a network interface into
+// compareAndAppendStatForInterface.
```
shamrickus, line 120: struct name in godoc.

traffic_monitor/threadsafe/resultstathistory.go
```diff
@@ -216,27 +375,55 @@ func StatsMarshall(statResultHistory ResultStatHistory, statInfo cache.ResultInf
...
 						break
 					}
-					if _, ok := stats.Caches[id][interfaceName]; !ok {
-						stats.Caches[id][interfaceName] = map[string][]cache.ResultStatVal{}
+					if _, ok := stats.Caches[cacheId].Interfaces[interfaceName]; !ok {
+						stats.Caches[cacheId].Interfaces[interfaceName] = map[string][]cache.ResultStatVal{}
 					}
-					stats.Caches[id][interfaceName][stat] = append(stats.Caches[id][interfaceName][stat], val)
+					stats.Caches[cacheId].Interfaces[interfaceName][stat] = append(stats.Caches[cacheId].Interfaces[interfaceName][stat], val)
 					// Todo add for each interface?
```
shamrickus, line 425: This comment (which was mine originally) no longer makes any sense

traffic_monitor/threadsafe/resultstathistory.go
```diff
@@ -253,32 +440,23 @@ func StatsMarshall(statResultHistory ResultStatHistory, statInfo cache.ResultInf
...
 			for stat, statValF := range computedStats {
 				if !filter.UseStat(stat) {
 					continue
 				}
-				if _, ok := stats.Caches[id][tc.CacheInterfacesAggregate]; !ok {
-					stats.Caches[id][tc.CacheInterfacesAggregate] = map[string][]cache.ResultStatVal{}
-				}
-				stats.Caches[id][tc.CacheInterfacesAggregate][stat] = append(stats.Caches[id][tc.CacheInterfacesAggregate][stat],
-					cache.ResultStatVal{Val: statValF(resultInfo, serverInfo, serverProfile, combinedStatesCache), Time: t, Span: 1})
 				// Need to actually handle interfaces, needs vitals to be refactored
```
shamrickus, line 453: Same story as above

traffic_monitor/threadsafe/resultstathistory.go
ocket8888: yeah, string building is up to a thousand times faster than `fmt.Sprintf`, and TM is notoriously resource-hoggy

lib/go-tc/servers_test.go
```diff
@@ -56,3 +59,1323 @@ func ExampleLegacyInterfaceDetails_ToInterfaces() {
...
+	if nullable.XMPPPasswd == nil {
+		t.Error("nullable conversion gave nil XMPPPasswd")
+	} else if *nullable.XMPPPasswd != srv.XMPPPasswd {
+		t.Errorf("Incorrect XMPPPasswd after nullable conversion; want: '%s', got: '%s'", srv.XMPPPasswd, *nullable.XMPPPasswd)
+	}
+}
+
+func TestServerNullableV2_Upgrade(t *testing.T) {
+	fqdn := "testFQDN"
+	srv := Server{
```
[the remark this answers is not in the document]
ocket8888: I'd be more inclined to put the duplicated nil checks in their own function, so that each test is still properly named for what it tests.

lib/go-tc/servers_test.go
```diff
@@ -56,3 +59,1323 @@ func ExampleLegacyInterfaceDetails_ToInterfaces() {
...
+		checkInterfaces = false
+	}
+	if nullable.IPNetmask == nil {
+		t.Error("Unexpectedly nil IPNetmask in nullable-converted server")
+		checkInterfaces = false
+	}
+
+	if checkInterfaces {
+		infLen := len(upgraded.Interfaces)
+		if infLen < 1 {
```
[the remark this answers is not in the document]
ocket8888: we could, but one of those cases is recoverable, whereas the other is not. IDK if it necessarily holds, though that the tests in the recoverable case are any use. I thought so, but I could see why you wouldn't.

traffic_monitor/towrap/towrap.go
```diff
@@ -365,8 +491,9 @@ func (s TrafficOpsSessionThreadsafe) trafficMonitorConfigMapRaw(cdn string) (*tc
 	return configMap, err
 }
 
-// LegacyTrafficMonitorConfigMap returns the Traffic Monitor config map from the Traffic Ops. This is safe for multiple goroutines.
-func (s TrafficOpsSessionThreadsafe) TrafficMonitorConfigMap(cdn string) (*tc.LegacyTrafficMonitorConfigMap, error) {
+// LegacyTrafficMonitorConfigMap returns the Traffic Monitor config map from the
```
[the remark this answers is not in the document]
ocket8888: Yeah, you and Steve are pointing out a lot of those, and to address them once: that's generally because I didn't write it, in cases like this. I just wrapped it so I didn't need to scroll sideways to read it. But I can fix it.

traffic_monitor/towrap/towrap.go
ocket8888: useLegacy? It's used at least three times in that same file.

traffic_monitor/manager/opsconfig.go
ocket8888: No, but tbh TM already seems broken in that case, from my testing with your UI indicator. Have I been doing something wrong? Does it work for you?

traffic_monitor/static/style.css
ocket8888: Yes. The selector is less specific than `tbody tr:nth-child(even)`, so it's getting overridden.

traffic_monitor/towrap/towrap.go
shamrickus: You are correct, originally I thought there were no code paths where legacy would be used, but after a second look it can.

traffic_monitor/manager/opsconfig.go
shamrickus: Depends on what you mean by working. In a strict sense, no it is not. 

After further review you should be good on this, it looks like it was originally just a copy paste of the Login function anyways.

This shouldn't be an issue, but I think it's worth noting: With these changes (assuming TO is down), each attempt will take twice as long since it tries both legacy and new TO.

ocket8888: Yeah, that's unavoidable, since we don't have a way to request which API versions TO supports. There was a proposal for that, but it never really got off the ground.

lib/go-tc/servers.go
```diff
@@ -97,6 +99,30 @@ type ServerInterfaceInfo struct {
...
 
+// GetDefaultAddress returns the IPv4 and IPv6 service addresses of the interface.
+func (i *ServerInterfaceInfo) GetDefaultAddress() (string, string) {
+	var ipv4 string
+	var ipv6 string
+	for _, ip := range i.IPAddresses {
+		if ip.ServiceAddress {
+			address, _, err := net.ParseCIDR(ip.Address)
+			if err != nil || address == nil {
+				continue
```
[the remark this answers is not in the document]
ocket8888: nah. I generally consider importing the logger into the tc, utils, or rfc packages bad form, since they're just libraries, and should be able to run their code in situations where logging is impossible, not set up, or doesn't even make sense. So if anything, it should just return the error and make the caller deal with it.

But also, I didn't write this method. I just changed the receiver; we had two separate, equivalent definitions of server interfaces. The other one was in `traffic_router.go`. I deleted that one and moved this method to its new, more appropriate location. So I don't really wanna change it.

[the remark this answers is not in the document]
rob05c: +1 - libraries should return errors, never log. If this really needs to return a warning, the function can be documented to return a "non-fatal error", or "warnings string array" or something. But that doesn't seem valuable enough to me to warrant the awkwardness. I would vote documenting the behavior in the function doc, though, e.g. "Addresses which aren't valid CIDR are ignored." 

But +1 to everything @ocket8888 said.

traffic_monitor/health/cache.go
ocket8888: How does it look now?

traffic_monitor/health/cache.go
```diff
@@ -144,177 +191,153 @@ func EvalCache(result cache.ResultInfo, resultStats *threadsafe.ResultStatValHis
...
-					if !processAvailableTuple(previousStatus.Available, serverInfo) && previousStatus.UnavailableStat != "" {
-						if !result.HasStat(previousStatus.UnavailableStat) {
-							return
-						}
-					}
-				}
+
+			available, why := EvalInterface(resultInfo.InterfaceVitals, inf)
+			if result.UsingIPv4 {
+				availStatus.Available.IPv4 = availStatus.Available.IPv4 && available
```
mattjackson220, line 277: with the last change and this line, if the previous status is unavailable then it can never get to be available. ill dig deeper today too to come up with another solution but wanted to throw it out there

shamrickus: Does this even need the `availStatus.Available.IPvX &&`? 

Also if I'm set to using IPv4, and in the next cycle I change to using IPv6, Available.IPv4 will remain `true` even though it shouldn't be.

ocket8888: well based on what Matt said, I think it should remain true:

>  _"the way this is written, IPv4 and IPv6 will never both be labeled as available at the same time. if we are checking ipv4 then should we set IPv6 based on `localCacheStatuses[result.ID].Available.IPv6` and vice versa?"_

unless I'm misunderstanding what he's saying

mattjackson220: yea the original plan was for each cycle to only update the status of what its checking. so if we are checking IPv6 then the Available.IPv4 should remain as it was and vice versa. it might be as easy as just removing the `availStatus.Available.IPvX &&` from these lines

shamrickus: Ahh I misread Matt's original comment. In that case then I think removing the && should fix the issue then.

traffic_monitor/manager/opsconfig.go
ocket8888: Issue about TM not working with backup files captured in #4946 .

mattjackson220:
Looks really close to me. TO and TM unit tests pass, API tests pass, docs build and look good, manual testing works as expected.
only thing ive noticed is that all interfaces in `/cache-statuses` have the same bandwidth, making the total bandwidth off too. i think its a bug in cachestate.go lines 181-187.  should use `health[0].InterfaceVitals[inf.Name]` data instead of `lastStat`?

traffic_monitor/cache/cache.go
```diff
@@ -156,121 +145,142 @@ type Vitals struct {
...
+// taken.
 type Stat struct {
 	Time  int64       `json:"time"`
 	Value interface{} `json:"value"`
 }
 
-// Stats is designed for returning via the API. It contains result history for each cache, as well as common API data.
+// StatsCache is a representation of cache server statistics as present in the
+// TM API.
+type StatsCache struct {
```
shamrickus, line 157: This is a bit of a nitpick, but this naming is _very_ confusing to me.
You have a Stats object that has a map of StatsCache that has a map of statistics called Stats.
Perhaps the best solution would be to rename the Stats struct?

ocket8888: Sure - to what?

shamrickus: `ServerStats`?

mattjackson220 (approved):
Looks good to me! lib, monitor, and ops unit tests pass, docs build and look good, v1 v2 and v3 API tests pass, testing in full environment all looks great!

shamrickus:
LGTM
There are still a couple issues, specifically with upgrading ATC (#5001), but those will be resolved in a separate PR.

shamrickus (approved):
LGTM
There are still a couple issues, specifically with upgrading ATC (#5001), but those will be resolved in a separate PR.

PR #4925, opened 2020-07-30
Add indicator when using a UI backup

### LINKED ISSUE #3529: Add Traffic Monitor UI indicator when using a file backup
We added the ability for the Monitor to use a file backup, when it can't reach Traffic Ops.

It logs an error when this happens, but there's no indicator in the web UI. The UI should have some kind of big, visible indicator when that happens, that a file is being used.

This will help administrators more quickly see when something is wrong.

shamrickus:

- [x] This PR fixes #3529 

It adds an indicator to the traffic monitor UI when it is using a disk backup for Traffic Ops.

- Traffic Monitor

Ensure that running TM as usual there is no UI change.
Run TM with a crconfig.backup and tmconfig.backup file present and TO down. Confirm that the UI has an indicator representing the disk backup usage. 

## Additional Information
Currently this backup is only loaded once at startup, as can be seen in `opsconfig.go:StartOpsConfigMapper`. CiaB requires TO to start before it will run TM, so you will need to do something tricky if you want to test in CiaB, I used breakpoints + debugger.

### DIFF
--- a/traffic_monitor/static/script.js
+++ b/traffic_monitor/static/script.js
@@ -109,7 +109,14 @@ function getTrafficOpsUri() {
 
 function getTrafficOpsCdn() {
 	ajax("/publish/ConfigDoc", function(r) {
-		document.getElementById("cdn-name").textContent = JSON.parse(r).cdnName || "unknown";
+		let opsConfig = JSON.parse(r);
+		document.getElementById("cdn-name").textContent = opsConfig.cdnName || "unknown";
+		var diskIconContainer = document.getElementById("icon-disc-holder");
+		if ((opsConfig.usingDummyTO || true) === false) {
+			diskIconContainer.classList.add("hidden");
+		} else {
+			//diskIconContainer.classList.remove("hidden");
+		}
 	});
 }
 
--- a/traffic_monitor/static/script.js
+++ b/traffic_monitor/static/script.js
@@ -108,14 +108,15 @@ function getTrafficOpsUri() {
 }
 
 function getTrafficOpsCdn() {
+	const cdnName = document.getElementById("cdn-name");
+	const discIconContainer = document.getElementById("icon-disc-holder");
 	ajax("/publish/ConfigDoc", function(r) {
 		let opsConfig = JSON.parse(r);
-		document.getElementById("cdn-name").textContent = opsConfig.cdnName || "unknown";
-		var discIconContainer = document.getElementById("icon-disc-holder");
-		if ((opsConfig.usingDummyTO || true) === false) {
-			discIconContainer.classList.add("hidden");
+		cdnName.textContent = opsConfig.cdnName || "unknown";
+		if (opsConfig.usingDummyTO === false) {
+		    discIconContainer.hidden = true;
 		} else {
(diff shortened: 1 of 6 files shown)


traffic_monitor/static/style.css
```diff
@@ -80,6 +80,69 @@ th {
...
 
+#icon-disc-holder.hidden {
+	display: none;
+}
+#icon-disc-holder {
+	display: block;
+	width: 125px;
+	position: relative;
+	margin-top: 5px;
+	cursor: pointer;
```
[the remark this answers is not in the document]
shamrickus: It's what I've always done for icons although you have a point.

Is there any standard docs or something I can go look at for future reference?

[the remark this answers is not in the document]
ocket8888: For cursor keyword purposes? I just look at [the MDN page for the `cursor` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor).

traffic_monitor/static/script.js
```diff
@@ -109,7 +109,14 @@ function getTrafficOpsUri() {
 
 function getTrafficOpsCdn() {
 	ajax("/publish/ConfigDoc", function(r) {
-		document.getElementById("cdn-name").textContent = JSON.parse(r).cdnName || "unknown";
+		let opsConfig = JSON.parse(r);
+		document.getElementById("cdn-name").textContent = opsConfig.cdnName || "unknown";
+		var discIconContainer = document.getElementById("icon-disc-holder");
+		if ((opsConfig.usingDummyTO || true) === false) {
+			discIconContainer.classList.add("hidden");
```
ocket8888, line 116: This is unreachable, because `(opsConfig.usingDummyTO || true) === false` is tautologically false.

Also, you could directly set the element's [`hidden` attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden) to avoid adding extra CSS rules to express that concept.

traffic_monitor/static/script.js
```diff
@@ -109,7 +109,14 @@ function getTrafficOpsUri() {
 
 function getTrafficOpsCdn() {
 	ajax("/publish/ConfigDoc", function(r) {
-		document.getElementById("cdn-name").textContent = JSON.parse(r).cdnName || "unknown";
+		let opsConfig = JSON.parse(r);
```
ocket8888, line 112: could be `const`

traffic_monitor/static/script.js
```diff
@@ -109,7 +109,14 @@ function getTrafficOpsUri() {
...
+		document.getElementById("cdn-name").textContent = opsConfig.cdnName || "unknown";
+		var discIconContainer = document.getElementById("icon-disc-holder");
```
ocket8888, line 114: could be `const`

but also could be calculated once outside of the function instead of on every call.

traffic_monitor/static/script.js
shamrickus: You are correct, I did not mean to push that up. As for the hidden attribute, I was using that originally but it wasn't working, I'll look into it again.

PR #4932, opened 2020-07-31
updated server ip_address usage

### LINKED ISSUE #4924: Internal Server Error when perform snapshot
## I'm submitting a ...

- [x] bug report
-  other 

-  CDN in a Box
-  Documentation
-  Grove
-  Traffic Control Client
-  Traffic Monitor
- [x]  Traffic Ops
-  Traffic Ops ORT
- [x] Traffic Portal
-  Traffic Router
-  Traffic Stats
-  Traffic Vault
-  unknown

## Current behavior:

Snapshot failed when perform.
```
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```

Should return 200 OK

mattjackson220:

- [x] This PR fixes #4924  

updated the places where we still used ip address or interface info from the servers table to use the new tables

- Traffic Ops

Verify unit tests are successful
Verify API tests are successful
Verify that snapshot works correctly and does not return a 500 error
Verify that /caches/stats returns the correct information
Verify that servers/details returns the correct information

this does not include changelog or docs since it is a bug fix
- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/cachesstats/cachesstats.go
+++ b/traffic_ops/traffic_ops_golang/cachesstats/cachesstats.go
@@ -214,7 +214,7 @@ SELECT
   cg.name as cachegroup,
   st.name as status,
   p.name as profile,
-  s.ip_address
+  (select address from ip_address where s.id = ip_address.server and service_address = true AND family(address) = 4) as ip
 FROM
   server s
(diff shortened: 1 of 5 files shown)


mitchell852:
I changed this to major since it broke the snapshot endpoint. Without the snapshot endpoint, TC doesn't really work :)

PR #4933, opened 2020-07-31
Update TO for minimum TLS version

hbeatty:

This update allows an administrator to set minimum TLS version for Traffic Ops.

- Documentation
- Traffic Ops

- Install TO
- Run `docker run -it drwetter/testssl.sh:latest -p <fqdn_of_traffic_ops>` and note the TLS versions supported
  ```

   SSLv2      not offered (OK)
   SSLv3      not offered (OK)
   TLS 1      offered (deprecated)
   TLS 1.2    offered (OK)
   NPN/SPDY   h2, http/1.1 (advertised)
   ALPN/HTTP2 h2, http/1.1 (offered)
  ```
- Set `"min_tls_version": 770` in the `traffic_ops_golang` section of cdn.conf
- restart TO
- Run `docker run -it drwetter/testssl.sh:latest -p <fqdn_of_traffic_ops>` and note that TLS version 1 is now off
  ```

   SSLv2      not offered (OK)
   SSLv3      not offered (OK)
   TLS 1      not offered
   TLS 1.2    offered (OK)
   NPN/SPDY   h2, http/1.1 (advertised)
   ALPN/HTTP2 h2, http/1.1 (offered)
  ```


docs/source/admin/traffic_ops.rst
```diff
@@ -443,6 +443,20 @@ This file deals with the configuration parameters of running Traffic Ops itself.
...
+			| Setting | Value            |
+			+=========+==================+
+			| 769     | TLS v1 (Default) |
+			+---------+------------------+
+			| 770     | TLS v1.1         |
+			+---------+------------------+
+			| 771     | TLS v1.2         |
+			+---------+------------------+
+			| 772     | TLS v1.3         |
+			+---------+------------------+
```
ocket8888, line 458: Tables should be wrapped in a `.. table` directive

mitchell852: @hbeatty ^^

docs/source/admin/traffic_ops.rst
```diff
@@ -443,6 +443,20 @@ This file deals with the configuration parameters of running Traffic Ops itself.
...
+			| Setting | Value            |
+			+=========+==================+
+			| 769     | TLS v1 (Default) |
+			+---------+------------------+
+			| 770     | TLS v1.1         |
+			+---------+------------------+
+			| 771     | TLS v1.2         |
+			+---------+------------------+
+			| 772     | TLS v1.3         |
```
zrhoffman, line 457: We should be accepting TLS version strings like `"1.3"`, `"1.2"`, etc. and translate that string to a valid `int` ourselves when reading the config. This is for `cdn.conf`, but for `riak.conf`, we already agreed to use version strings like that. See @rawlinp's [comment](https://github.com/apache/trafficcontrol/pull/4573#discussion_r400346222) in #4573:

> 769, 770, etc are not very intuitive -- would it be better to use a string or float that we translate into `tls.VersionTLS11` from the `tls` package in the stdlib?

rob05c: Rather than adding `MinTLSVersion`, I think we should put the entire `tls.Config` in the Config. We should do this for all Go apps. There are many config settings in `tls.Config`, such as disabling H2, allows Ciphers, and more. 
If we keep adding individual options, every time we need something else, it's a code change and a new version.

Whereas if we put `tls.Config` in the app config, it's just a config change for users to add whatever they need with respect to TLS.

docs/source/admin/traffic_ops.rst
hbeatty: OK. I'll figure out how to do that.

docs/source/admin/traffic_ops.rst
zrhoffman: `tls.Config` is well-documented in [the tls GoDoc](https://godoc.org/crypto/tls#Config), so I'm fine with sticking to that for now with the caveat that I wouldn't oppose adding `MinTLSVersion` in a future PR.

docs/source/admin/traffic_ops.rst
```diff
@@ -443,6 +447,8 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 		:disabled_routes: A list of API route IDs to disable. Requests matching these routes will receive a 503 response. To find the route ID for a given path you would like to disable, run ``./traffic_ops_golang`` using the :option:`--api-routes` option to view all the route information, including route IDs and paths.
 		:ignore_unknown_routes: If ``false`` (default) return an error and prevent startup if unknown route IDs are found. Otherwise, log a warning and continue startup.
 
+	:tls_config: An optional stanza for TLS configuration. The values of which can be found here: https://golang.org/pkg/crypto/tls/#Config
```
ocket8888, line 450: To link to a GoDoc, use our custom `:godoc:` role. In this case, it'd be:
```rst
The values of which conform to the :godoc:`crypto/tls.Config` structure.
```
That creates an automatic link.

traffic_ops/traffic_ops_golang/config/config.go
```diff
@@ -99,8 +100,9 @@ type ConfigTrafficOpsGolang struct {
 	RiakPort                 *uint                      `json:"riak_port"`
 	WhitelistedOAuthUrls     []string                   `json:"whitelisted_oauth_urls"`
 	OAuthClientSecret        string                     `json:"oauth_client_secret"`
-	RoutingBlacklist         `json:"routing_blacklist"`
-	SupportedDSMetrics       []string `json:"supported_ds_metrics"`
+	RoutingBlacklist                                    `json:"routing_blacklist"`
+	SupportedDSMetrics       []string                   `json:"supported_ds_metrics"`
+	TLSConfig                *tls.Config                `json:"tls_config"`
```
ocket8888, line 105: This file needs to be `go fmt`-ed

docs/source/admin/traffic_ops.rst
ocket8888: pointing to the `net/tls.Config` GoDoc is actually a better solution; this comment is no longer relevant.

docs/source/admin/traffic_ops.rst
```diff
@@ -399,6 +399,10 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 	:db_query_timeout_seconds: An optional field specifying a timeout on database *transactions* (not actually single queries in most cases) within API route handlers. Effectively this is a timeout on a single handler's ability to interact with the Traffic Ops Database. Default if not specified is the value of `DefaultDBQueryTimeoutSecs <https://godoc.org/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/config#pkg-constants>`_.
 	:idle_timeout: An optional timeout in seconds for idle client connections to Traffic Ops. If set to zero, the value of ``read_timeout`` will be used instead. If both are zero, then the value of ``read_header_timeout`` will be used. If all three fields are zero, there is no timeout and connections will be kept alive indefinitely - **not** recommended. Default if not specified is zero.
 	:insecure: An optional boolean which, if set to ``true`` will cause Traffic Ops to skip verification of client certificates whenever necessary/possible. If set to ``false``, the normal verification behavior is exhibited. Default if not specified is ``false``.
+
+		.. deprecated:: 4.2
```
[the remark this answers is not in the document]
hbeatty: :+1:

docs/source/admin/traffic_ops.rst
hbeatty: :+1:

traffic_ops/traffic_ops_golang/config/config.go
hbeatty: Uh, I don't know how to do that. Is that going to keep this PR from being accepted?

ocket8888: > I don't know how to do that.

To `go fmt` a file you just do `go fmt {{file}}`.

> Is that going to keep this PR from being accepted?

Well, yeah, it's listed in [the docs for Traffic Ops Go development requirements](https://traffic-control-cdn.readthedocs.io/en/latest/development/traffic_ops.html#go-implementation-requirements), but it's also an extremely simple fix.

> ... I was following the format that was already there.

Well, you lined up the struct tags. But you can see that [you actually changed the format](https://github.com/apache/trafficcontrol/pull/4933/files#diff-8f0e586303d36096ee3c6ae7a0a7bffdL102-R105):
```diff
-  RoutingBlacklist         `json:"routing_blacklist"`
+  RoutingBlacklist                                    `json:"routing_blacklist"`
```

PR #4942, opened 2020-08-07
Add Traffic Ops If-Match and If-Unmodified-Since Support in Server and Clients

### LINKED ISSUE #3509: Add Traffic Ops If-Match and If-Unmodified-Since Support in Server and Clients

srijeet0406:

- [x] This PR fixes #3509 

- Documentation
- Traffic Ops

Run traffic ops locally.
Now go into the database and check the last_updated time for, say, a server.
Try sending a `PUT` request to modify this server, by specifying an `If-Unmodified-Since` time of something that is BEFORE the last_updated of the existing server. 
In this case, you should get back a response of `412` status code, saying that the resource was modified.
Now, try changing the time to something thats after the last_updated time of the server.
Your update should go through this time.
You can try the same with the `If-Match` header as well, by converting the last_updated time stamp of the resource to an Etag before sending it in.

- master

- [x] This PR includes tests


traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
```diff
@@ -567,6 +582,22 @@ func (cg *TOCacheGroup) Update() (error, error, int) {
 	return nil, nil, http.StatusOK
 }
 
+func CheckIfExistsBeforeUpdate(tx *sqlx.Tx, cgID int) (error, *tc.TimeNoMod) {
```
rob05c, line 585: These functions look identical, except for the table in the select query.
Could they be refactored to use a common func, maybe something like 
```
// GetLastUpdated returns the last updated time of the given ID in the given table, whether the ID existed, and any error.
func GetLastUpdated(tx *sqlx.Tx, dbTable string, id int) (time.Time, bool, error)
```` 

In the `api` package?

File: traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go [hunk 2]
```diff
@@ -226,7 +244,7 @@ func (req *TODeliveryServiceRequest) Update() (error, error, int) {
 	userID := tc.IDNoMod(req.APIInfo().User.ID)
 	req.LastEditedByID = &userID
 
-	return api.GenericUpdate(req)
+	return api.GenericUpdate(nil, req)
```
rob05c, line 247: Why is this passing `nil`, and not the `http.Header`?

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
```diff
@@ -383,6 +401,20 @@ func (req *deliveryServiceRequestAssignment) Update() (error, error, int) {
 	req.DeliveryServiceRequestNullable = current.DeliveryServiceRequestNullable
 	req.AssigneeID = assigneeID
 
+	err, existingLastUpdated := CheckIfExistsBeforeUpdate(req.ReqInfo.Tx, *req.ID)
```
rob05c, line 404: This whole block seems to be identical in a lot of places. Could it be abstracted into something like `func IsUnmodified(tx *sql.Tx, id int, tableName string, hdr http.Header) bool` ?

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
```diff
@@ -383,6 +401,20 @@ func (req *deliveryServiceRequestAssignment) Update() (error, error, int) {
...
+	if err != nil {
+		return errors.New("no delivery service request found with this id"), nil, http.StatusNotFound
+	}
+	_, iumsTime := rfc.GetETagOrIfUnmodifiedSinceTime(h)
+	existingEtag := rfc.ETag(existingLastUpdated.Time)
+
+	if h.Get(rfc.IfMatch) != "" && !strings.Contains(h.Get(rfc.IfMatch), existingEtag) {
+		return errors.New("header preconditions dont match"), nil, http.StatusPreconditionFailed
```
rob05c, line 412: The text here doesn't have to match the code, and since it's a user error, I think it's safe (from a security perspective) to let them know why they're getting a 412. Could we make the message something more informative like "resource was modified" ?

traffic_ops/traffic_ops_golang/deliveryservice/safe.go
```diff
@@ -23,11 +23,10 @@ import (
 	"database/sql"
 	"fmt"
 	"github.com/apache/trafficcontrol/lib/go-log"
-	"net/http"
-
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/tenant"
+	"net/http"
```
rob05c, line 29: We try to separate imports with newlines for "standard library," "third party", and "our own," just to make it a little easier to read. Would you mind moving this up with the other standard library imports? Mind also moving `lib/go-log` down with the other TC ones? (that one wasn't yours, but while you're improving it, might as well do that one too).

traffic_ops/traffic_ops_golang/origin/origins.go
```diff
@@ -309,6 +309,7 @@ func (origin *TOOrigin) Update() (error, error, int) {
 	}
 
 	log.Debugf("about to run exec query: %s with origin: %++v", updateQuery(), origin)
+	// ToDo : Srijeet change here
```
rob05c, line 312: I assume an Unmodified check needs to go here?

traffic_ops/traffic_ops_golang/server/servers_server_capability.go
```diff
@@ -121,7 +121,7 @@ func (ssc TOServerServerCapability) Validate() error {
 	return util.JoinErrs(tovalidate.ToErrors(errs))
 }
 
-func (ssc *TOServerServerCapability) Update() (error, error, int) {
+func (ssc *TOServerServerCapability) Update(h http.Header) (error, error, int) {
```
rob05c, line 124: I know you didn't create this function, but it can be deleted.

This is a legacy from when the "CRUDer" stuff required every func be implemented. This endpoint doesn't have a PUT method, and the "CRUDer" framework no longer requires all funcs, so it can just be deleted now.

traffic_ops/traffic_ops_golang/steeringtargets/steeringtargets.go
```diff
@@ -244,6 +244,7 @@ func (st *TOSteeringTargetV11) Update() (error, error, int) {
 		return userErr, sysErr, errCode
 	}
 
+	// ToDo : Srijeet change here
```
rob05c, line 247: Need an Unmodified check here?

traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
srijeet0406: I have refactored it so that all the places that are NOT part of the CRUDer interfaces, call the common function in the `api` package, by passing in the table name like you suggested. 
For the `GenericUpdate` occurrences, we need it to be specific to each component, because we dont know what "type" is calling the function. Hence, each component that implements the CRUDer interface, has its own implementation of the above function.

rob05c: Could the function also take the type? So the CRUDer funcs all become something like:
```
func (v *TODeliveryServiceRequestComment) CheckIfExistsBeforeUpdate() (time.Time, error) {
  return GetLastUpdated(v.APIInfo().Tx, dbTable string, id int, v.Type())
}
```
?

srijeet0406: Absolutely 👍

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -986,3 +986,36 @@ func CreateDeprecationAlerts(alternative *string) tc.Alerts {
 		return tc.CreateAlerts(tc.WarnLevel, "This endpoint is deprecated, and will be removed in the future")
 	}
 }
+
+// GetLastUpdated checks for the resource in the database, and returns its last_updated timestamp, if available
+func GetLastUpdated(tx *sqlx.Tx, ID int, tableName string) (error, *tc.TimeNoMod) {
```
ocket8888, line 991: I don't think this is a rule or anything, but generally errors are the last returned value of functions

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -986,3 +986,36 @@ func CreateDeprecationAlerts(alternative *string) tc.Alerts {
 		return tc.CreateAlerts(tc.WarnLevel, "This endpoint is deprecated, and will be removed in the future")
 	}
 }
+
+// GetLastUpdated checks for the resource in the database, and returns its last_updated timestamp, if available
```
ocket8888, line 990: nit: GoDocs should be complete sentences that end with proper punctuation.

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -986,3 +986,36 @@ func CreateDeprecationAlerts(alternative *string) tc.Alerts {
...
+	if !rows.Next() {
+		return errors.New("no resource found with this id"), nil
+	}
+	if err := rows.Scan(&lastUpdated); err != nil {
+		return err, nil
+	}
+	return nil, &lastUpdated
+}
+
+// IsUnmodified returns a boolean, saying whether or not the resource in question was modified since the time specified in the headers
```
ocket8888, line 1007: Same as above RE: punctuation.

traffic_ops/traffic_ops_golang/api/generic_crud.go
```diff
@@ -213,15 +215,23 @@ func GenericRead(h http.Header, val GenericReader, useIMS bool) ([]interface{},
...
+
 	rows, err := val.APIInfo().Tx.NamedQuery(val.UpdateQuery(), val)
 	if err != nil {
 		return ParseDBError(err)
 	}
 	defer rows.Close()
 
 	if !rows.Next() {
-		return errors.New("no " + val.GetType() + " found with this id"), nil, http.StatusNotFound
+		return errors.New("resource was modified"), nil, http.StatusPreconditionFailed
```
ocket8888, line 234: I don't think this should've changed, should it?

traffic_ops/testing/api/v3/cachegroups_test.go
```diff
@@ -42,11 +42,43 @@ func TestCacheGroups(t *testing.T) {
 		var header http.Header
 		header = make(map[string][]string)
 		header.Set(rfc.IfModifiedSince, time)
+		header.Set(rfc.IfUnmodifiedSince, time)
```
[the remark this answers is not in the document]
srijeet0406: So currently, we support `If-Modified-Since` in only the `GET` requests, whereas `If-Unmodified-Since` and `If-Match` are supported only in the `PUT` requests. I have set them to the same time (a time in the past) for both the headers, because these headers are being used in different tests, and have no relation to each other as such.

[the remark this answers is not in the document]
ocket8888: I see.

traffic_ops/testing/api/v3/deliveryservice_request_comments_test.go
```diff
@@ -31,13 +31,34 @@ func TestDeliveryServiceRequestComments(t *testing.T) {
...
+		UpdateTestDeliveryServiceRequestCommentsWithHeaders(t, header)
 		GetTestDeliveryServiceRequestComments(t)
 		GetTestDeliveryServiceRequestCommentsIMSAfterChange(t, header)
 	})
 }
 
+func UpdateTestDeliveryServiceRequestCommentsWithHeaders(t *testing.T, header http.Header) {
+	comments, _, err := TOSession.GetDeliveryServiceRequestComments(header)
+
+	firstComment := comments[0]
```
ocket8888, line 49: This'll segfault if len(comments) is < 1

traffic_ops/testing/api/v3/coordinates_test.go
```diff
@@ -33,11 +33,36 @@ func TestCoordinates(t *testing.T) {
...
 	})
 }
 
+func UpdateTestCoordinatesWithHeaders(t *testing.T, header http.Header) {
+	firstCoord := testData.Coordinates[0]
+	resp, _, err := TOSession.GetCoordinateByName(firstCoord.Name, nil)
+	if err != nil {
+		t.Errorf("cannot GET Coordinate by name: %v - %v", firstCoord.Name, err)
+	}
+	coord := resp[0]
```
ocket8888, line 53: Same as above RE: segfault

traffic_ops/testing/api/v3/cachegroups_test.go
```diff
@@ -42,11 +42,43 @@ func TestCacheGroups(t *testing.T) {
...
 	})
 }
 
+func UpdateTestCacheGroupsWithHeaders(t *testing.T, h http.Header) {
+	firstCG := testData.CacheGroups[0]
+	resp, _, err := TOSession.GetCacheGroupNullableByName(*firstCG.Name, h)
+	if err != nil {
+		t.Errorf("cannot GET CACHEGROUP by name: %v - %v", *firstCG.Name, err)
+	}
+	cg := resp[0]
```
ocket8888, line 62: Same as above RE: segfault

traffic_ops/testing/api/v3/cdnfederations_test.go
```diff
@@ -28,9 +32,37 @@ func TestCDNFederations(t *testing.T) {
...
 }
 
+func UpdateTestCDNFederationsWithHeaders(t *testing.T, h http.Header) {
+	for _, id := range fedIDs {
+		fed, _, err := TOSession.GetCDNFederationsByID("foo", id, h)
+		if err != nil {
+			t.Errorf("cannot GET federation by id: %v", err)
+		}
+		expectedCName := "new.cname."
+		fed.Response[0].CName = &expectedCName
```
ocket8888, line 55: Same as above RE: Segfault

traffic_ops/testing/api/v3/cdns_test.go
```diff
@@ -32,12 +32,36 @@ func TestCDNs(t *testing.T) {
...
 }
 
+func UpdateTestCDNsWithHeaders(t *testing.T, header http.Header) {
+	firstCDN := testData.CDNs[0]
+	// Retrieve the CDN by name so we can get the id for the Update
+	resp, _, err := TOSession.GetCDNByName(firstCDN.Name, header)
+	if err != nil {
+		t.Errorf("cannot GET CDN by name: '%s', %v", firstCDN.Name, err)
+	}
+	remoteCDN := resp[0]
```
ocket8888, line 54: Same as above RE: Segfault

traffic_ops/testing/api/v3/divisions_test.go
```diff
@@ -34,12 +34,38 @@ func TestDivisions(t *testing.T) {
...
 }
 
+func UpdateTestDivisionsWithHeaders(t *testing.T, header http.Header) {
+	firstDivision := testData.Divisions[0]
+	// Retrieve the Division by division so we can get the id for the Update
+	resp, _, err := TOSession.GetDivisionByName(firstDivision.Name, header)
+	if err != nil {
+		t.Errorf("cannot GET Division by division: %v - %v", firstDivision.Name, err)
+	}
+	remoteDivision := resp[0]
```
ocket8888, line 56: Same as above RE: Segfault

traffic_ops/testing/api/v3/origins_test.go
```diff
@@ -16,26 +16,61 @@ package v3
...
 }
 
+func UpdateTestOriginsWithHeaders(t *testing.T, header http.Header) {
+	firstOrigin := testData.Origins[0]
+	// Retrieve the origin by name so we can get the id for the Update
+	resp, _, err := TOSession.GetOriginByName(*firstOrigin.Name)
+	if err != nil {
+		t.Errorf("cannot GET origin by name: %v - %v", *firstOrigin.Name, err)
+	}
+	remoteOrigin := resp[0]
```
ocket8888, line 58: Same as above RE: Segfault

traffic_ops/testing/api/v3/parameters_test.go
```diff
@@ -37,12 +37,37 @@ func TestParameters(t *testing.T) {
...
 }
 
+func UpdateTestParametersWithHeaders(t *testing.T, header http.Header) {
+	firstParameter := testData.Parameters[0]
+	// Retrieve the Parameter by name so we can get the id for the Update
+	resp, _, err := TOSession.GetParameterByName(firstParameter.Name, nil)
+	if err != nil {
+		t.Errorf("cannot GET Parameter by name: %v - %v", firstParameter.Name, err)
+	}
+	remoteParameter := resp[0]
```
ocket8888, line 59: Same as above RE: Segfault

traffic_ops/testing/api/v3/phys_locations_test.go
```diff
@@ -35,16 +35,41 @@ func TestPhysLocations(t *testing.T) {
...
 }
 
+func UpdateTestPhysLocationsWithHeaders(t *testing.T, header http.Header) {
+	firstPhysLocation := testData.PhysLocations[0]
+	// Retrieve the PhysLocation by name so we can get the id for the Update
+	resp, _, err := TOSession.GetPhysLocationByName(firstPhysLocation.Name, header)
+	if err != nil {
+		t.Errorf("cannot GET PhysLocation by name: '%s', %v", firstPhysLocation.Name, err)
+	}
+	remotePhysLocation := resp[0]
```
ocket8888, line 57: Same as above RE: Segfault

traffic_ops/testing/api/v3/profiles_test.go
```diff
@@ -30,18 +30,47 @@ func TestProfiles(t *testing.T) {
...
 }
 
+func UpdateTestProfilesWithHeaders(t *testing.T, header http.Header) {
+	firstProfile := testData.Profiles[0]
+	// Retrieve the Profile by name so we can get the id for the Update
+	resp, _, err := TOSession.GetProfileByName(firstProfile.Name, header)
+	if err != nil {
+		t.Errorf("cannot GET Profile by name: %v - %v", firstProfile.Name, err)
+	}
+	remoteProfile := resp[0]
```
ocket8888, line 58: Same as above RE: Segfault

traffic_ops/testing/api/v3/profiles_test.go
```diff
@@ -30,18 +30,47 @@ func TestProfiles(t *testing.T) {
...
 		CopyProfile(t)
+		header = make(map[string][]string)
+		etag := rfc.ETag(currentTime)
+		header.Set(rfc.IfMatch, etag)
+		UpdateTestProfilesWithHeaders(t, header)
 	})
 }
 
+func UpdateTestProfilesWithHeaders(t *testing.T, header http.Header) {
+	firstProfile := testData.Profiles[0]
```
ocket8888, line 52: Same as above RE: Segfault

traffic_ops/testing/api/v3/profiles_test.go
```diff
@@ -30,18 +30,47 @@ func TestProfiles(t *testing.T) {
...
+	// Retrieve the Profile by name so we can get the id for the Update
+	resp, _, err := TOSession.GetProfileByName(firstProfile.Name, header)
+	if err != nil {
+		t.Errorf("cannot GET Profile by name: %v - %v", firstProfile.Name, err)
+	}
+	remoteProfile := resp[0]
+	remoteProfile.Description = "UPDATED"
```
ocket8888, line 59: To cover a corner-case where the description was already "UPDATED", this should append with `+= "UPDATED"` so that it's always changed.

ocket8888: On the other hand, I don't know if that matters for the purposes of this test. But if it doesn't, then why bother changing it at all?

traffic_ops/testing/api/v3/regions_test.go
```diff
@@ -33,17 +33,41 @@ func TestRegions(t *testing.T) {
...
 		DeleteTestRegionsByName(t)
+		header = make(map[string][]string)
+		etag := rfc.ETag(currentTime)
+		header.Set(rfc.IfMatch, etag)
+		UpdateTestRegionsWithHeaders(t, header)
 	})
 }
 
+func UpdateTestRegionsWithHeaders(t *testing.T, header http.Header) {
+	firstRegion := testData.Regions[0]
```
ocket8888, line 50: Same as above RE: Segfault

traffic_ops/testing/api/v3/regions_test.go
```diff
@@ -33,17 +33,41 @@ func TestRegions(t *testing.T) {
...
+	// Retrieve the Region by region so we can get the id for the Update
+	resp, _, err := TOSession.GetRegionByName(firstRegion.Name, header)
+	if err != nil {
+		t.Errorf("cannot GET Region by region: %v - %v", firstRegion.Name, err)
+	}
+	remoteRegion := resp[0]
```
ocket8888, line 56: Same as above RE: Segfault

traffic_ops/testing/api/v3/roles_test.go
```diff
@@ -40,13 +40,38 @@ func TestRoles(t *testing.T) {
...
+	t.Logf("testData.Roles contains: %+v\n", testData.Roles)
+	firstRole := testData.Roles[0]
+	// Retrieve the Role by role so we can get the id for the Update
+	resp, _, status, err := TOSession.GetRoleByName(*firstRole.Name, header)
+	t.Log("Status Code: ", status)
+	if err != nil {
+		t.Errorf("cannot GET Role by role: %v - %v", firstRole.Name, err)
+	}
+	t.Logf("got response: %+v\n", resp)
+	remoteRole := resp[0]
```
ocket8888, line 66: Same as above RE: Segfault

traffic_ops/testing/api/v3/roles_test.go
```diff
@@ -40,13 +40,38 @@ func TestRoles(t *testing.T) {
...
+		header = make(map[string][]string)
+		etag := rfc.ETag(currentTime)
+		header.Set(rfc.IfMatch, etag)
+		UpdateTestRolesWithHeaders(t, header)
 	})
 }
 
+func UpdateTestRolesWithHeaders(t *testing.T, header http.Header) {
+	t.Logf("testData.Roles contains: %+v\n", testData.Roles)
+	firstRole := testData.Roles[0]
```
ocket8888, line 58: Same as above RE: Segfault

traffic_ops/testing/api/v3/staticdnsentries_test.go
```diff
@@ -35,12 +35,35 @@ func TestStaticDNSEntries(t *testing.T) {
...
 		UpdateTestStaticDNSEntriesInvalidAddress(t)
+		header = make(map[string][]string)
+		etag := rfc.ETag(currentTime)
+		header.Set(rfc.IfMatch, etag)
+		UpdateTestStaticDNSEntriesWithHeaders(t, header)
 	})
 }
 
+func UpdateTestStaticDNSEntriesWithHeaders(t *testing.T, header http.Header) {
+	firstStaticDNSEntry := testData.StaticDNSEntries[0]
```
ocket8888, line 51: Same as above RE: Segfault

traffic_ops/testing/api/v3/staticdnsentries_test.go
```diff
@@ -35,12 +35,35 @@ func TestStaticDNSEntries(t *testing.T) {
...
+	// Retrieve the StaticDNSEntries by name so we can get the id for the Update
+	resp, _, err := TOSession.GetStaticDNSEntriesByHost(firstStaticDNSEntry.Host, header)
+	if err != nil {
+		t.Errorf("cannot GET StaticDNSEntries by name: '%s', %v", firstStaticDNSEntry.Host, err)
+	}
+	remoteStaticDNSEntry := resp[0]
```
ocket8888, line 57: Same as above RE: Segfault

traffic_ops/testing/api/v3/statuses_test.go
```diff
@@ -32,12 +32,41 @@ func TestStatuses(t *testing.T) {
...
 		GetTestStatusesIMSAfterChange(t, header)
+		header = make(map[string][]string)
+		etag := rfc.ETag(currentTime)
+		header.Set(rfc.IfMatch, etag)
+		UpdateTestStatusesWithHeaders(t, header)
 	})
 }
 
+func UpdateTestStatusesWithHeaders(t *testing.T, header http.Header) {
+	firstStatus := testData.Statuses[0]
```
ocket8888, line 48: Same as above RE: Segfault

traffic_ops/testing/api/v3/statuses_test.go
```diff
@@ -32,12 +32,41 @@ func TestStatuses(t *testing.T) {
...
+	if firstStatus.Name == nil {
+		t.Fatal("cannot update test statuses: first test data status must have a name")
+	}
+
+	// Retrieve the Status by name so we can get the id for the Update
+	resp, _, err := TOSession.GetStatusByName(*firstStatus.Name, header)
+	if err != nil {
+		t.Errorf("cannot GET Status by name: %v - %v", firstStatus.Name, err)
+	}
+	remoteStatus := resp[0]
```
ocket8888, line 58: Same as above RE: Segfault

traffic_ops/testing/api/v3/deliveryservice_request_comments_test.go
srijeet0406: Yeah I pretty much copied the code from the other `Update` test which does the same thing. I'll fix it in both the places.

traffic_ops/v3-client/deliveryservice.go
```diff
@@ -357,6 +357,45 @@ func (to *Session) CreateDeliveryServiceNullable(ds *tc.DeliveryServiceNullable)
...
+	return data.Response[0], reqInf, nil
+
+}
+
+// UpdateDeliveryServiceV30 replaces the Delivery Service identified by the
+// integral, unique identifier 'id' with the one it's passed.
+// Deprecated: Use UpdateDeliveryServiceV30WithHdr instead.
+func (to *Session) UpdateDeliveryServiceV30(id int, ds tc.DeliveryServiceNullableV30) (tc.DeliveryServiceNullableV30, ReqInf, error) {
+	return to.UpdateDeliveryServiceV30WithHdr(id, ds, nil)
+}
```
ocket8888, line 384: You can just make this take a header instead. Ditto the other functions. They're brand new, so there's no need to deprecate them; you can change these.

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -994,3 +994,57 @@ func CreateDeprecationAlerts(alternative *string) tc.Alerts {
...
+//	if h != nil {
+//		if h.Get(rfc.IfMatch) != "" && !strings.Contains(h.Get(rfc.IfMatch), existingEtag) {
+//			return false
+//		}
+//		if iumsTime != nil && existingLastUpdated.UTC().After(iumsTime.UTC()) {
+//			return false
+//		}
+//	}
+//	return true
+//}
```
ocket8888, line 1031: you wanna remove this now-unused code?

traffic_ops/traffic_ops_golang/api/shared_handlers.go
```diff
@@ -25,15 +25,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/apache/trafficcontrol/lib/go-rfc"
+	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-tc"
```
ocket8888, line 29: Shouldn't move these up into the standard imports, they should stay beneath in a separate set of lines.

lib/go-rfc/cachecontrol_test.go
```diff
@@ -41,4 +41,4 @@ func TestETag(t *testing.T) {
 	if ans.UTC().String() != "2020-08-06 18:11:22.278418 +0000 UTC" {
 		t.Errorf("Expected time %v, actual %v", "2020-08-06 18:11:22.278418 +0000 UTC", ans.UTC().String())
 	}
-}
+}
```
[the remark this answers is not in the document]
rob05c: Moreover, it means the file hasn't been `gofmt`. Running `gofmt` will fix this, and possibly other things.

ocket8888 (approved):
This all looks fine, and all tests are passing.

lib/go-rfc/cachecontrol_test.go
```diff
@@ -0,0 +1,44 @@
+package rfc
+
+import (
+	"testing"
+	"time"
+)
+
+/*
```
rob05c, line 8: Nitpick: the license header should go immediately after the `package`, but before the `import`.

rob05c (changes requested):
Looks great! Code looks good, and this is a great feature to fix operator race conditions. The only thing I'm requesting is the `pq.QuoteIdentifier` to prevent future injection. The other comments would be nice, but I wouldn't hold up this PR for them.

Manually tested, works as expected, sending a IUS header newer than the `date` results in a 412, no IUS or a older IUS PUT succeeds.

Unit tests pass, API tests pass, except 2 failing tests that this doesn't modify and didn't cause.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -999,11 +999,28 @@ func CreateDeprecationAlerts(alternative *string) tc.Alerts {
 	}
 }
 
+// CheckIfUnModified checks to see if the resource was modified since the "If-Unmodified-Since" header value in the request.
+// In case it was, the 412 error code is returned. If some other error was encountered while checking, the appropriate error code along with
+// error details is returned. If the resource was not modified since the specified time, the UPDATE proceeds in the normal fashion.
+func CheckIfUnModified(h http.Header, tx *sqlx.Tx, ID int, tableName string) (error, error, int) {
+	existingLastUpdated, found, err := GetLastUpdated(tx, ID, tableName)
+	if err == nil && found == false {
+		return errors.New("no delivery service request found with this id"), nil, http.StatusNotFound
```
rawlinp, line 1008: The error message should use the `tableName` parameter, or maybe even a different `resourceName` param that is more error message friendly.

srijeet0406: yeah sorry was a copy paste error, will fix it.

rawlinp (approved):

rob05c (approved):
Approving: `pq.QuoteIdentifier` was added since my previous review, which was the only request

PR #4943, opened 2020-08-10
CiaB: Use rpm command instead of yum to avoid fetching repo metadata that will go unused

zrhoffman:

This PR changes CDN-in-a-Box to install the ATC components directly using `rpm` instead of through `yum`, which makes syncing yum repos unnecessary if the RPM installation layers are the only ones needing to be rebuilt. This can save several minutes building CDN-in-a-Box, depending on your Internet speed.

- CDN in a Box

- Ensure CDN-in-a-Box builds successfully and that the readiness container exits successfully

- [x] CDN-in-a-Box change, no tests necessary
- [x] Behavior unchanged, no additional documentation necessary

### DIFF
--- a/infrastructure/cdn-in-a-box/cache/Dockerfile
+++ b/infrastructure/cdn-in-a-box/cache/Dockerfile
@@ -29,7 +29,7 @@ EXPOSE 80
 ADD https://ci.trafficserver.apache.org/RPMS/CentOS7/trafficserver-7.1.4-2.el7.x86_64.rpm /trafficserver.rpm
 ADD https://ci.trafficserver.apache.org/RPMS/CentOS7/trafficserver-devel-7.1.4-2.el7.x86_64.rpm /trafficserver-devel.rpm
 
-RUN yum install -y bind-utils kyotocabinet-libs epel-release initscripts iproute net-tools nmap-ncat gettext autoconf automake libtool gcc-c++ cronie glibc-devel openssl-devel 'perl(JSON)'
+RUN yum install -y bind-utils kyotocabinet-libs epel-release initscripts iproute net-tools nmap-ncat gettext autoconf automake libtool gcc-c++ cronie glibc-devel openssl-devel
 RUN yum install -y /trafficserver.rpm /trafficserver-devel.rpm jq python36-psutil python36-typing python36-setuptools python36-pip logrotate && yum clean all
 RUN python3 -m pip install --upgrade pip && python3 -m pip install requests urllib3 distro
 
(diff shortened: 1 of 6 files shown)


infrastructure/cdn-in-a-box/cache/Dockerfile
```diff
@@ -29,7 +29,7 @@ EXPOSE 80
 ADD https://ci.trafficserver.apache.org/RPMS/CentOS7/trafficserver-7.1.4-2.el7.x86_64.rpm /trafficserver.rpm
 ADD https://ci.trafficserver.apache.org/RPMS/CentOS7/trafficserver-devel-7.1.4-2.el7.x86_64.rpm /trafficserver-devel.rpm
 
-RUN yum install -y bind-utils kyotocabinet-libs epel-release initscripts iproute net-tools nmap-ncat gettext autoconf automake libtool gcc-c++ cronie glibc-devel openssl-devel
+RUN yum install -y bind-utils kyotocabinet-libs epel-release initscripts iproute net-tools nmap-ncat gettext autoconf automake libtool gcc-c++ cronie glibc-devel openssl-devel 'perl(JSON)'
```
ocket8888, line 32: Do we not need `perl(JSON)`?

zrhoffman: The RPM spec requires it. If you don't include it, you get
```
 > [20/20] RUN rpm -Uvh /$(basename infrastructure/cdn-in-a-box/cache/traffic_ops_ort.rpm) &&    rm /$(basename infrastructure/cdn-in-a-box/cache/traffic_ops_ort.rpm):
#26 0.548 error: Failed dependencies:
#26 0.548       perl(JSON) is needed by traffic_ops_ort-4.2.0-10848.7af9dd63.el7.x86_64
```

I agree that we don't really need it. But since the other dependency validation is still useful, maybe we should remove that dependency from the spec when we make the project use the go version of ORT that was added in #4872.

Either way, `perl(JSON)` was already being installed, its installation was just deferred until we installed the ORT RPM.

ocket8888: well, that's a dependency for the Perl script that's unused. Is there a `--no-dependencies` option or something for `rpm`?

zrhoffman: The full dependency list for EL7 is

* perl-JSON
* perl-libwww-perl
* perl-Crypt-SSLeay
* perl-LWP-Protocol-https
* perl-Digest-SHA

So, we wouldn't be missing anything by skipping dependency verification.

Dependencies skipped in af14af15b.

PR #4951, opened 2020-08-13
Use RFC lib and correct Content-Type on TM api

### LINKED ISSUE #2565: Traffic Monitor API gives incorrect Content-Type Header (or response format) for `version` and `traffic-ops-uri`
The TM endpoints at `/api/version` and `/api/traffic-ops-uri` both return plain-text responses, but have a `Content-Type` header that declares it to be `application/json`. The responses should either be wrapped in double quotes to make them valid JSON strings, or the `Content-Type` header should be set to `text/plain`.

shamrickus:

- [x] This PR fixes #2565. 

It replaces the content-type specification in TM with those found in the go-rfc lib. It also changes the `/api/version` and `/api/traffic-ops-uri` TM endpoints to have a Content-Type of `text/plain`/`text/uri-list` (respectively) in place of `application-json` as it does not return a valid JSON string.

- Traffic Monitor

Ensure TM tests pass
Make a request to `api/version` and `api/traffic-ops-uri` and confirm Content-Type

master (9e69ce1)

## Additional Information|

Before:
```
❯ curl -v http://localhost/api/version
*   Trying ::1:80...
* Connected to localhost (::1) port 80 (#0)
> GET /api/version HTTP/1.1
> Host: localhost
> User-Agent: curl/7.71.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Thu, 13 Aug 2020 17:14:52 GMT
< Content-Length: 28
< 
* Connection #0 to host localhost left intact
traffic_monitor-4.2.0.9e69ce
```

After:   
```
❯ curl -v http://localhost/api/version
*   Trying ::1:80...
* Connected to localhost (::1) port 80 (#0)
> GET /api/version HTTP/1.1
> Host: localhost
> User-Agent: curl/7.71.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: text/plain
< Date: Thu, 13 Aug 2020 17:14:52 GMT
< Content-Length: 28
< 
* Connection #0 to host localhost left intact
traffic_monitor-4.2.0.9e69ce   
```

### DIFF
--- a/traffic_monitor/datareq/datareq.go
+++ b/traffic_monitor/datareq/datareq.go
@@ -113,7 +113,7 @@ func MakeDispatchMap(
 		}, rfc.ContentTypeTextPlain)),
 		"/api/traffic-ops-uri": wrap(WrapBytes(func() []byte {
 			return srvAPITrafficOpsURI(opsConfig)
-		}, rfc.ApplicationJSON)),
+		}, rfc.ContentTypeURIList)),
 		"/api/cache-statuses": wrap(WrapErr(errorCount, func() ([]byte, error) {
 			return srvAPICacheStates(toData, statInfoHistory, statResultHistory, healthHistory, lastHealthDurations, localStates, lastStats, localCacheStatus, statMaxKbpses, monitorConfig)
 		}, rfc.ApplicationJSON)),
--- a/traffic_monitor/datareq/datareq.go
+++ b/traffic_monitor/datareq/datareq.go
@@ -30,6 +30,7 @@ import (
 	"unicode"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-rfc"
 	"github.com/apache/trafficcontrol/traffic_monitor/config"
 	"github.com/apache/trafficcontrol/traffic_monitor/health"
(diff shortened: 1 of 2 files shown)


traffic_monitor/datareq/datareq.go
```diff
@@ -72,62 +73,62 @@ func MakeDispatchMap(
...
-		}, ContentTypeJSON)),
+		}, rfc.ApplicationJSON)),
 		"/api/version": wrap(WrapBytes(func() []byte {
 			return srvAPIVersion(staticAppData)
-		}, ContentTypeJSON)),
+		}, rfc.ContentTypeTextPlain)),
 		"/api/traffic-ops-uri": wrap(WrapBytes(func() []byte {
 			return srvAPITrafficOpsURI(opsConfig)
-		}, ContentTypeJSON)),
+		}, rfc.ApplicationJSON)),
```
ocket8888, line 116: As per #2565 (which this PR will fix, btw), `/api/traffic-ops-uri` does not return a JSON-encoded response. `text/plain` would not be incorrect, but [`text/uri-list`](https://tools.ietf.org/html/rfc2483#section-5) would be better.

PR #4953, opened 2020-08-13
GET /api/3.0/servers?dsId=1 500s when requested with If-Modified-Since header

### LINKED ISSUE #4949: GET /api/3.0/servers?dsId=1 500s when requested with If-Modified-Since header
## I'm submitting a ...

-  bug report

-  Traffic Ops

## Current behavior:

`GET /api/3.0/servers?dsId=1` returns a response with an *Internal Server Error* alert and status code 500  when requested with `If-Modified-Since` header.

## Expected / new behavior:

`GET /api/3.0/servers?dsId=1` should return status code 200 with a set of servers matching the query parameter

1. Start CDN-in-a-Box
2. Log into Traffic Portal
3. Visit https://trafficportal.infra.ciab.test/api/3.0/servers?dsId=1
4. Visit https://trafficportal.infra.ciab.test/api/3.0/servers?dsId=1 again, receive error

## Anything else:

Traffic Ops error log:
```go
ERROR: api.go:209: 2020-08-13T15:39:45.268410415Z: 172.16.239.8:49780 querying: pq: current transaction is aborted, commands ignored until end of transaction block
DEBUG: api.go:214: 2020-08-13T15:39:45.268428091Z: Internal Server Error
```

srijeet0406:

- [x] This PR fixes #4949 

- CDN in a Box
- Traffic Ops

Start TO locally
Make a call to get the servers associated with a ds like:
`curl -X GET -k -H "Accept: application/json" -H "If-Modified-Since: Mon, 13 Jul 2020 21:26:00 GMT" "https://localhost:8443/api/3.0/servers?dsId=3"`
This should not give you a 500 error, instead, you should see the servers associated with this DS.

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
@@ -814,7 +814,7 @@ func (dss *TODSSDeliveryService) Read(h http.Header, useIMS bool) ([]interface{}
 
 func selectMaxLastUpdatedQuery(where string) string {
 	return `SELECT max(t) from (
-		SELECT max(dss.last_updated) as t from deliveryservice_server dss ` + where +
+		SELECT max(dss.last_updated) as t from deliveryservice_server dss JOIN deliveryservice ds ON ds.id = dss.deliveryservice ` + where +
 		` UNION ALL
 	select max(last_updated) as t from last_deleted l where l.table_name='deliveryservice_server') as res`
 }
--- a/traffic_ops/traffic_ops_golang/server/servers.go
+++ b/traffic_ops/traffic_ops_golang/server/servers.go
@@ -611,7 +611,8 @@ JOIN cdn cdn ON s.cdn_id = cdn.id
 JOIN phys_location pl ON s.phys_location = pl.id
 JOIN profile p ON s.profile = p.id
 JOIN status st ON s.status = st.id
-JOIN type t ON s.type = t.id ` + where +
+JOIN type t ON s.type = t.id 
+FULL OUTER JOIN deliveryservice_server dss ON dss.server = s.id ` + where +
 		` UNION ALL
 	select max(last_updated) as t from last_deleted l where l.table_name='server') as res`
 }
--- a/traffic_ops/traffic_ops_golang/server/servers.go
+++ b/traffic_ops/traffic_ops_golang/server/servers.go
@@ -648,7 +648,7 @@ JOIN phys_location pl ON s.phys_location = pl.id
 JOIN profile p ON s.profile = p.id
 JOIN status st ON s.status = st.id
 JOIN type t ON s.type = t.id ` +
-queryAddition + where +
+		queryAddition + where +
 		` UNION ALL
 	select max(last_updated) as t from last_deleted l where l.table_name='server') as res`
 }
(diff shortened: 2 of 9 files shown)


traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
```diff
@@ -814,7 +814,7 @@ func (dss *TODSSDeliveryService) Read(h http.Header, useIMS bool) ([]interface{}
 
 func selectMaxLastUpdatedQuery(where string) string {
 	return `SELECT max(t) from (
-		SELECT max(dss.last_updated) as t from deliveryservice_server dss ` + where +
+		SELECT max(dss.last_updated) as t from deliveryservice_server dss JOIN deliveryservice ds ON ds.id = dss.deliveryservice ` + where +
```
zrhoffman, line 817: We don't always need this join, only when dsId is a query parameter. How is the join conditionally added to `serverCountQuery` and `selectQuery`?

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -611,7 +611,8 @@ JOIN cdn cdn ON s.cdn_id = cdn.id
 JOIN phys_location pl ON s.phys_location = pl.id
 JOIN profile p ON s.profile = p.id
 JOIN status st ON s.status = st.id
-JOIN type t ON s.type = t.id ` + where +
+JOIN type t ON s.type = t.id 
+FULL OUTER JOIN deliveryservice_server dss ON dss.server = s.id ` + where +
```
zrhoffman, line 615: As with the comment above, we don't always need this join, only when dsId is a query parameter.

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
srijeet0406: We will actually need this one always, because we look for the DS ID in the params and if its not present, we return there. So if it makes it to this line, we are guaranteed that we do have a delivery service id, denoted by `id` in the params.

traffic_ops/traffic_ops_golang/server/servers.go
srijeet0406: Fixed

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
zrhoffman: I missed that this one was in a completely different file. Sounds good

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -640,14 +640,15 @@ func ReadID(w http.ResponseWriter, r *http.Request) {
...
+func selectMaxLastUpdatedQuery(queryAddition string, where string) string {
 	return `SELECT max(t) from (
 		SELECT max(s.last_updated) as t from server s JOIN cachegroup cg ON s.cachegroup = cg.id
 JOIN cdn cdn ON s.cdn_id = cdn.id
 JOIN phys_location pl ON s.phys_location = pl.id
 JOIN profile p ON s.profile = p.id
 JOIN status st ON s.status = st.id
-JOIN type t ON s.type = t.id ` + where +
+JOIN type t ON s.type = t.id ` +
+queryAddition + where +
```
zrhoffman, line 651: `go fmt` indents this line

PR #4955, opened 2020-08-13
Allow '0' as a valid value for TTL field in DNS static entry

### LINKED ISSUE #3400: Portal: Cannot use "0" as a TTL value for Static DNS entries
This was possible with Traffic Ops although "0" now reports `'ttl' cannot be blank`. TTL of 0 is valid number.

rimashah25:

- [x] This PR fixes #3400 

- Traffic Portal 
- Traffic Ops

**Manual Testing** 
1. In Traffic Portal, create a HTTP delivery service
2. Click on `More` and then add Static DNS entry
3. Select a Type and type ttl value as zero (`0`) 
4. You can create a static DNS entry with a successful message.

**Automation Testing**
1. Run CIAB
2. Run Traffic_Portal locally (make sure the Traffic_Ops url point to CIAB traffic ops url)
3. Ensure protractor and webdriver-manage are installed.
4. Run webdriver-manager start
5. Run protractor config.js from traffic_control/traffic_portal/test/end-to-end/ folder
6. All 73 specs should pass. `73 specs, 0 failures`

- master (e313410)

- [x] This PR includes tests
- [ ] This PR does not include documentation
- [x] This does not include a database migration

### DIFF
--- a/traffic_portal/app/src/common/modules/form/deliveryServiceStaticDnsEntry/form.deliveryServiceStaticDnsEntry.tpl.html
+++ b/traffic_portal/app/src/common/modules/form/deliveryServiceStaticDnsEntry/form.deliveryServiceStaticDnsEntry.tpl.html
@@ -50,9 +50,9 @@ under the License.
             <div class="form-group" ng-class="{'has-error': hasError(dsStaticDnsEntryForm.ttl), 'has-feedback': hasError(dsStaticDnsEntryForm.ttl)}">
                 <label for="ttl" class="control-label col-md-2 col-sm-2 col-xs-12">TTL *</label>
                 <div class="col-md-10 col-sm-10 col-xs-12">
-                    <input id="ttl" name="ttl" type="number" class="form-control" ng-model="staticDnsEntry.ttl" min="0" step="1" required>
+                    <input id="ttl" name="ttl" type="number" class="form-control" ng-model="staticDnsEntry.ttl" min="0" step="1" ng-pattern="/^\d+$/" required>
                     <small class="input-error" ng-show="hasPropertyError(dsStaticDnsEntryForm.ttl, 'required')">Required</small>
-                    <small class="input-error" ng-show="hasPropertyError(dsStaticDnsEntryForm.ttl, 'step') || hasPropertyError(dsStaticDnsEntryForm.ttl, 'min')">Whole Number</small>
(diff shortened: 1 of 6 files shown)


traffic_portal/app/src/common/modules/form/deliveryServiceStaticDnsEntry/form.deliveryServiceStaticDnsEntry.tpl.html
```diff
@@ -50,9 +50,9 @@
             <div class="form-group" ng-class="{'has-error': hasError(dsStaticDnsEntryForm.ttl), 'has-feedback': hasError(dsStaticDnsEntryForm.ttl)}">
                 <label for="ttl" class="control-label col-md-2 col-sm-2 col-xs-12">TTL *</label>
                 <div class="col-md-10 col-sm-10 col-xs-12">
-                    <input id="ttl" name="ttl" type="number" class="form-control" ng-model="staticDnsEntry.ttl" min="0" step="1" required>
+                    <input id="ttl" name="ttl" type="number" class="form-control" ng-model="staticDnsEntry.ttl" ng-pattern="/^\d+$/" required>
```
ocket8888, line 53: Does `min="0" step="1"` not work?

rimashah25: It works but ng-pattern seems to be a better approach.

ocket8888: Why? `min` and `step` are native `input[type=number]` controls that work within the browser. `ng-pattern` not only adds an event listener, but uses userland scripting to do regular expression matching. It's slower and less idiomatic than the native controls.

mitchell852: @ocket8888 - how about just `pattern` instead of `ng-pattern` since the pattern is static? `min` does not prevent the user from entering -1, for example. basically, i just want to make sure an alert shows up if the user enters a non-whole number.

ocket8888: > _"min does not prevent the user from entering -1, for example._"

What do you mean? Like the user can literally type in "-1"? Because you can still do that with `ng-pattern`. Also, without a min, I can use the browser's native controls to enter an invalid value. I can hit the down arrow and get -1. Using `min` and `step` if I put in "-1" I still see the field marked as invalid.

image

PR #4971, opened 2020-08-18
Fix traffic_ops/client changed symbols

### LINKED ISSUE #4969: Fix traffic_ops/client changed symbols
Recently, a `http.Header` was added to the TO client funcs. This is necessary to let clients pass IMS/IUS/etc headers. But we need to add new functions, to avoid breaking existing clients.

So, for example, `traffic_ops/client/cdn.go` was changed from
```
func (to *Session) GetCDNs() ([]tc.CDN, ReqInf, error)
```
To
```
func (to *Session) GetCDNs(header http.Header) ([]tc.CDN, ReqInf, error)
```

This breaks any clients using it. It needs to follow the standard deprecated-for-one-major-version procedure. It should be changed to:

```
// GetCDNs is Deprecated in and will be removed in Traffic Control 6.0.
func (to *Session) GetCDNs() ([]tc.CDN, ReqInf, error) {
  return to.GetCDNsWithHdr(nil)
}
func (to *Session) GetCDNsWithHdr(header http.Header) ([]tc.CDN, ReqInf, error) { 
```

I'm not particular about the name, as long as the existing symbol is preserved to not break users. In fact, if could be `func (to *Session) CDNs(header http.Header) ([]tc.CDN, ReqInf, error)` if people preferred. (I think it used to be, before `GetCDNs` was added with `ReqInf`.)

This needs done for all the changed function names in `traffic_ops/client`.

Note this is concerning the Traffic Control library version, and not breaking users of libraries. It has nothing to do with the Traffic Ops API Version.

## I'm submitting a ...
-  bug report

-  Traffic Control Client

## Current behavior:
Existing function was removed, breaking clients.

## Expected / new behavior:
Existing function continues to exist, pending deprecation-and-removal process.

Try to pull and update the client for an existing app using a changed function, observe build failure.

## Anything else:

srijeet0406:

- [x] This PR fixes #4969  

- Traffic Ops
- Traffic Ops ORT

Make sure the unit tests and api tests pass for TO. Also, for any other apps using TO, verify that they build/ test correctly.

- master

- [x] This PR includes tests
- [x] No documentation necessary


zrhoffman:
Some functions are missing after this PR

```bash
diff \
    <(git show 55b3b03451 | grep '^-func ' | sed 's/^.*Session) \([A-Za-z0-9]\+\).*/\1/i' | sort) \
    <(git show 55b3b03451 | grep '^+func ' | sed 's/^.*Session) \([A-Za-z0-9]\+\).*/\1/i' | sort) |
    grep '^<' | awk '{print $2}'
```
Output:
```
GetDeliveryServicesByServerV30
GetDeliveryServicesV30
```

srijeet0406:
@zrhoffman Those were new functions, and have been replaced by `GetDeliveryServicesByServerV30WithHdr` and `GetDeliveryServicesV30WithHdr` respectively.

PR #4980, opened 2020-08-20
ORT t3c command fixes:

jrushford:
Fixes the following problems found while integration testing the ORT t3c command.
  
  - use the atstccfg --traffic-ops-insecure flag to allow for self signed certs.
  - insure that atstccfg logging locations do not use 'stdout' as using 'stdout'
    will cause json parsing errors.
  - corrects the --help output to correctly show the default logging locations for
    t3c is 'stderr'.

### DIFF
--- a/traffic_ops_ort/t3c/torequest/torequest.go
+++ b/traffic_ops_ort/t3c/torequest/torequest.go
@@ -193,6 +193,24 @@ func (r *TrafficOpsReq) atsTcExec(cmdstr string) ([]byte, error) {
 
 // atsTcExecCommand is used to run the atstccfg command.
 func (r *TrafficOpsReq) atsTcExecCommand(cmdstr string, queueState int, revalState int) ([]byte, error) {
+	// adjust log locations used for atstccfg
+	// cannot use stdout as this will cause json parsing errors.
+	errorLocation := r.Cfg.LogLocationErr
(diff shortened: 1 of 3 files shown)


traffic_ops_ort/t3c/torequest/torequest.go
```diff
@@ -193,16 +193,35 @@ func (r *TrafficOpsReq) atsTcExec(cmdstr string) ([]byte, error) {
...
 		"--traffic-ops-password=" + r.Cfg.TOPass,
 		"--traffic-ops-url=" + r.Cfg.TOURL,
 		"--cache-host-name=" + r.Cfg.CacheHostName,
-		"--log-location-error=" + r.Cfg.LogLocationErr,
-		"--log-location-info=" + r.Cfg.LogLocationInfo,
-		"--log-location-warning=" + r.Cfg.LogLocationWarn,
+		"--log-location-error=" + errorLocation,
+		"--log-location-info=" + infoLocation,
+		"--log-location-warning=" + warningLocation,
+		"--traffic-ops-insecure",
```
rob05c, line 224: I don't think we should always set this. We should use proper HTTPS, unless someone specifically needs it. Especially in production.

Mind adding an arg to `t3c`, to pass through here, that defaults to false?

PR #4985, opened 2020-08-25
Find a better way to determine last status change

### LINKED ISSUE #1636: New TO server field: LastStatusChange
As a user, I want to know when the last status change of a server was so that I can track how long a server has been in a given status such as OFFLINE or ADMIN_DOWN.  The only current way to maybe do it is to parse the entire changelog, but that doesn't work with API changes where the individual changes aren't posted to the changelog.

srijeet0406:

- [x] This PR fixes #1636  

- CDN in a Box
- Traffic Ops
- Traffic Portal
- CI tests

Run local TO and TP and create a server. Notice that the `status_last_updated` field gets set to the current time in the DB.
Now, go to TP, and select the `Status Last Updated` as one of the display columns for the servers table. Notice that it shows the correct time elapsed since you created the server.
Now, update this server changing any field, EXCEPT the `status` of the server. Notice that the `status_last_updated` value DOES NOT change.
Now, update the server by changing the server's status (you will have to do this from the backend call, since TP doesn't let you do it from the front end). 
Notice that the `status_last_updated` time gets changed to the current time this time around.

Also, make sure unit tests and API tests pass.

- master

- [x] This PR includes tests OR

### DIFF
--- a/lib/go-tc/servers.go
+++ b/lib/go-tc/servers.go
@@ -382,7 +382,6 @@ type CommonServerProperties struct {
 	UpdPending        *bool                `json:"updPending" db:"upd_pending"`
 	XMPPID            *string              `json:"xmppId" db:"xmpp_id"`
 	XMPPPasswd        *string              `json:"xmppPasswd" db:"xmpp_passwd"`
-	StatusLastUpdated time.Time            `json:"statusLastUpdated" db:"status_last_updated"`
 }
 
 // ServerNullableV11 is a server as it appeared in API version 1.1.
@@ -396,12 +395,14 @@ type ServerNullableV2 struct {
 	ServerNullableV11
 	IPIsService  *bool `json:"ipIsService" db:"ip_address_is_service"`
 	IP6IsService *bool `json:"ip6IsService" db:"ip6_address_is_service"`
+	StatusLastUpdated time.Time            `json:"statusLastUpdated" db:"status_last_updated"`
 }
 
 // ServerNullable represents an ATC server, as returned by the TO API.
 type ServerNullable struct {
 	CommonServerProperties
 	Interfaces []ServerInterfaceInfo `json:"interfaces" db:"interfaces"`
+	StatusLastUpdated time.Time            `json:"statusLastUpdated" db:"status_last_updated"`
 }
 
 // ToServerV2 converts the server to an equivalent ServerNullableV2 structure,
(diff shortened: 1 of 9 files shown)


lib/go-tc/servers.go
```diff
@@ -342,45 +343,46 @@ type ServerV1 struct {
...
+	RouterPortName    *string              `json:"routerPortName" db:"router_port_name"`
+	Status            *string              `json:"status" db:"status"`
+	StatusID          *int                 `json:"statusId" db:"status_id"`
+	TCPPort           *int                 `json:"tcpPort" db:"tcp_port"`
+	Type              string               `json:"type" db:"server_type"`
+	TypeID            *int                 `json:"typeId" db:"server_type_id"`
+	UpdPending        *bool                `json:"updPending" db:"upd_pending"`
+	XMPPID            *string              `json:"xmppId" db:"xmpp_id"`
+	XMPPPasswd        *string              `json:"xmppPasswd" db:"xmpp_passwd"`
+	StatusLastUpdated time.Time            `json:"statusLastUpdated" db:"status_last_updated"`
```
ocket8888, line 385: This, unfortunately, cannot be done. It changes the structures returned by all client methods. Those need to stay frozen at whatever they were prior to API v3, and this new field must be added only to the APIv3 Server object.

traffic_ops/traffic_ops_golang/server/put_status.go
```diff
@@ -156,10 +157,12 @@ func updateServerStatusAndOfflineReason(serverID, statusID int, offlineReason *s
 	q := `
 UPDATE server
 SET    status = $1,
-       offline_reason = $2
-WHERE  id = $3
+       offline_reason = $2,
+       status_last_updated = $3
```
ocket8888, line 161: You can actually just set this to `NOW()` right in the query, no need to interpolate.

srijeet0406:
It actually does account for the changes made via `PUT /servers/{{ID}}`. But you're right about the user being able to change the `last status change time`. I'll fix that.

srijeet0406:
Actually, scratch that. You cannot change the `status_last_updated` time by doing a `PUT` to `/servers/{{ID}}` because of this part of the code:

`if newServer.StatusID != nil && *newServer.StatusID != originalStatusID {
			newServer.StatusLastUpdated = time.Now()
		} else {
			newServer.StatusLastUpdated = origSer[0].StatusLastUpdated
		}`

srijeet0406:
Closed in favor of https://github.com/apache/trafficcontrol/pull/4989

PR #4986, opened 2020-08-25
Traffic Portal objects blueprint

ocket8888:

This PR adds a blueprint for "breaking out" Traffic Portals into their own API object, out of the generic "server" object.

None

Read the blueprint.

- [x] Tests are unnecessary for blueprints
- [x] Documentation is unnecessary for blueprints
- [x] An update to CHANGELOG.md is not necessary for blueprints

### DIFF
--- a/blueprints/traffic.portal.objects.md
+++ b/blueprints/traffic.portal.objects.md
@@ -1,0 +1,511 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Traffic Portals as API objects
+
+## Problem Description
+Traffic Portals are, today, configured in Traffic Ops as regular servers, with
+all the details that entails. A Traffic Portal can have updates and
+revalidations pending, has a physical location, is associated with a CDN
+despite being primarily for operating arbitrary numbers of CDNs, belongs to a
+Cache Group despite not being a Cache Server, has a Type with no semantic
+meaning beyond identifying it as a Traffic Portal instance, has a Profile with
+no meaningful Parameters, has a HashID despite not being hashed for routing by
+Traffic Router, and can have multiple network interfaces with their own
+bandwidth limits - despite that Traffic Monitor does not monitor Traffic Portals
+at all.
+
+## Proposed Change
+All of these properties are superfluous with no real meaning for Traffic
+Portals, and for that reason alone they'd be better off as their own object.
+The new Traffic Portal object would be compact, containing only the properties
+that are needed by and make sense on Traffic Portal instances. Doing this would
+also be a step toward avoiding complex filtering on servers for processes that
+are looking for specific kinds of servers (though as far as this writer knows,
+there aren't any that look specifically for Traffic Portal servers).
+
+## Data Model Impact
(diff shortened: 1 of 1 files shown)


blueprints/traffic.portal.objects.md
```diff
@@ -0,0 +1,562 @@
...
+at all.
+
+## Proposed Change
+All of these properties are superfluous with no real meaning for Traffic
+Portals, and for that reason alone they'd be better off as their own object.
+The new Traffic Portal object would be compact, containing only the properties
+that are needed by and make sense on Traffic Portal instances. Doing this would
+also be a step toward avoiding complex filtering on servers for processes that
+are looking for specific kinds of servers (though as far as this writer knows,
+there aren't any that look specifically for Traffic Portal servers).
```
jhg03a, line 41: They exist.

blueprints/traffic.portal.objects.md
```diff
@@ -0,0 +1,562 @@
...
+Since Traffic Portals are intended only as a UI, and need not even exist for an
+ATC system to function normally, the impact should be quite limited.
+
+### Traffic Ops Impact
+
+#### Database Impact
+A new table must be created to store Traffic Portal information. It can be
+populated trivially from the information of servers with Type IDs that identify
+Types which have the exact Name "TRAFFIC_PORTAL". Removal of existing data in
+the generic `servers` table should be left up to operators' discretion. The
```
jhg03a, line 110: The existing server object must not be removed or more of its functions must be added to this new object.  TO is the system of record for all infrastructure in ATC.  All the concepts of how operators can access it such as OOB login info or management IPs must be preserved.

blueprints/traffic.portal.objects.md
```diff
@@ -0,0 +1,562 @@
...
+### URL
+This string is a full URL at which the UI is served, including scheme (e.g.
+`http://`) and optionally port (e.g. `80`), e.g.
+`https://trafficportal.infra.ciab.test:443`.
+
+If the port is omitted, it will be guessed based on the protocol indicated by
+the schema, e.g. `80` for `http://`, which indicates the HTTP protocol. The only
+valid schemes are `https://` and `http://`.
+
+This property uniquely identifies a Traffic Portal, and must therefore
```
jhg03a, line 94: Are you also attempting to cover the use case of a LB sitting in front of multiple TP instances?

jhg03a (changes requested):

blueprints/traffic.portal.objects.md
```diff
@@ -0,0 +1,562 @@
...
+that are needed by and make sense on Traffic Portal instances. Doing this would
+also be a step toward avoiding complex filtering on servers for processes that
+are looking for specific kinds of servers (though as far as this writer knows,
+there aren't any that look specifically for Traffic Portal servers).
+
+## Data Model Impact
+The proposed new object type is expressed below as a TypeScript interface.
+
+```typescript
+interface TrafficPortal {
```
rob05c, line 47: This is missing a lot of Asset data about the physical server.
I've said for years that IMO Traffic Ops shouldn't be used for Asset Management. But it's simply a fact that it's designed and used for that today.

A lot of Server fields this removes are cache-specific, but a lot are Asset data. Like the netmask, gateway, ILO fields, physLocation, rack. That's all Asset data people are using to track server machines today, that has to be stored somewhere.

Unless there's a concrete proposal that Traffic Control / Traffic Ops move away from Asset Management, and Users (including us) will be required to migrate their data into new Asset Management systems?

Is the intention to propose and require that?

If not, we need to identify and add the Asset-related fields (the ones I mention aren't exhaustive). Ideally in a shared DB table for normalization.

ocket8888:
Thanks for the reviews, but this was just an example for a proposal on the mailing list. A proposal which appears to be rejected - at least in its current form. This blueprint, therefore, no longer serves any purpose. I had just forgotten to close it.

blueprints/traffic.portal.objects.md
```diff
@@ -0,0 +1,562 @@
...
+This string contains a valid IPv6 Address at which the UI is served. If it is
+`null`, then it is assumed that there is no static IPv6 Address at which
+the Traffic Portal instance may be reached - or it does not communicate over
+IPv6 at all.
+
+### Notes
+This string contains arbitrary text for containing miscellaneous information.
+
+### Online
+This boolean value indicates whether or not this Traffic Portal is online and
```
jhg03a, line 75: This is a functionality reduction from today where we can have TP in custom statuses that match server lifecycle.

PR #4989, opened 2020-08-26
Find a better way to determine last status change

### LINKED ISSUE #1636: New TO server field: LastStatusChange

srijeet0406:

- [x] This PR fixes #1636  

- CDN in a Box
- Traffic Ops
- Traffic Portal
- CI tests

Run local TO and TP and create a server. Notice that the `status_last_updated` field gets set to the current time in the DB.
Now, go to TP, and select the `Status Last Updated` as one of the display columns for the servers table. Notice that it shows the correct time elapsed since you created the server.
Now, update this server changing any field, EXCEPT the `status` of the server. Notice that the `status_last_updated` value DOES NOT change.
Now, update the server by changing the server's status (you will have to do this from the backend call, since TP doesn't let you do it from the front end). 
Notice that the `status_last_updated` time gets changed to the current time this time around.

Also, make sure unit tests and API tests pass.

- master

- [x] This PR includes tests OR


mitchell852:
sorry, conflict in our favorite file :)

srijeet0406:
@mitchell852 fixed 👍

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -102,9 +102,102 @@ SELECT
 	s.type AS server_type_id,
 	s.upd_pending AS upd_pending,
 	s.xmpp_id,
-	s.xmpp_passwd
+	s.xmpp_passwd,
+    s.status_last_updated
```
mitchell852, line 106: format problem?

srijeet0406: hmm, I ran `go fmt` on it 🤔

mitchell852:
for some reason the column filter is not working in TP on this field:

image

I have a server i just changed the status of and it's not showing up. So i'm wondering if the data format needs to be changed to that of LastUpdated:

lastUpdated: "2020-08-28 20:53:55+00",
statusLastUpdated: "2020-08-28T20:53:55.75928Z"

I'm not sure if that will solve the TP grid problem but maybe consistent is better anyhow?

ocket8888:
The data will compare properly if you coerce it to a Date. If you look in the servers table controller, when I set `$scope.servers` I map the returned servers array and apply a conversion to Date objects on the Last Updated field. So no, it's not that the Last Updated format works (in fact it's worse than the statusLastUpdated field, I had to do some string manipulation to get the `Date` constructor to work), it's just that it can only filter Dates by date.

srijeet0406:
Resolved in `97bca2f`

mitchell852:
Can you update `dateCellFormatter` as well to this?

```
	/**
	 * Formats the contents of the 'lastUpdated' and 'statusLastUpdated' column cells as "relative to now".
	 */
	function dateCellFormatter(params) {
		return params.value ? dateUtils.getRelativeTime(params.value) : params.value;
	}
```

so `dateUtils.getRelativeTime` doesn't try to parse a null which displays "Invalid Date"

PR #4993, opened 2020-08-27
Adds support for CSV parsing in astats

ezelkow1:
Adds CSV parsing support to traffic monitor for the astats plugin.
The astats parser will check the incoming content-type header to determine if it should parse the data as json or csv. Added a cfg option of http_polling_format that defaults to "text/json", this is what is sent on http requests to caches. Changing it to "text/csv" will enable the csv output if the cache's astats plugin as has support for CSV, otherwise astats will respond with "text/json" or "text/javascript" depending on the version.

- Documentation
- Traffic Monitor

Set the http_polling_format configuration option in TM to "text/csv" to send that in the accept header to caches.  Using a cache that has a recent version of astats with CSV support then verify that the cache and its stats are reported correctly in TM.

Tests exist in the traffic_monitor/cache directory

## Additional Information

I have also included benchmarks for comparison to json usage:
BenchmarkAstatsJson-8   	    5931	    203527 ns/op	  134475 B/op	    2028 allocs/op
BenchmarkAstatsCSV-8    	    8680	    147683 ns/op	  119839 B/op	     812 allocs/op

The CSV is ~40-50% faster depending on run while also greatly decreasing the amount of allocations per operation which should hopefully help with garbage collection pressures.

Updating with new benchmark numbers after changing to use upfront allocation on the data map:
BenchmarkAstatsJson-8   	    5836	    197696 ns/op	  134473 B/op	    2028 allocs/op
BenchmarkAstatsCSV-8    	   10000	    110413 ns/op	  110065 B/op	     800 allocs/op

The csv is now more efficient both for operations and allocations compared to json doing a single up front allocation of the map based on the number of lines returned from astats


traffic_monitor/cache/astats_csv.go
```diff
@@ -0,0 +1,114 @@
...
+	"strings"
+
+	"github.com/apache/trafficcontrol/lib/go-log"
+)
+
+type astatsDataCsv struct {
+	Ats map[string]interface{}
+}
+
+func astatsCsvParseCsv(cacheName string, data io.Reader, pollCTX interface{}) (Statistics, map[string]interface{}, error) {
```
zrhoffman, line 37: `pollCTX` is unused

traffic_monitor/cache/astats_csv.go
```diff
@@ -0,0 +1,114 @@
...
+		line := scanner.Text()
+		delim := strings.IndexByte(line, ',')
+
+		// No delimiter found, skip this line as invalid
+		if delim < 0 {
+			continue
+		}
+		// Special cases where we just want the string value
+		if strings.Contains(line[0:delim], "proc.") || strings.Contains(line[0:delim], "inf.name") {
+			atsData.Ats[line[0:delim]] = line[delim+1 : len(line)]
```
zrhoffman, line 60: Right-hand value can be just `line[delim+1:]`

traffic_monitor/cache/astats_csv.go
```diff
@@ -0,0 +1,114 @@
...
+		} else {
+			value, err := strconv.ParseFloat(line[delim+1:len(line)], 64)
```
zrhoffman, line 62: Slice can be `line[delim+1:]`

traffic_monitor/cache/astats_csv.go
```diff
@@ -0,0 +1,114 @@
...
+			// Skip values that dont parse
+			if err != nil {
+				continue
+			}
+			atsData.Ats[line[0:delim]] = value
+		}
+	}
+
+	if len(atsData.Ats) < 1 {
+		return stats, nil, errors.New("No 'global' data object found in stats_over_http payload")
```
zrhoffman, line 73: In Go, error messages should not be capitalized.

traffic_monitor/cache/astats_csv.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	statMap := atsData.Ats
+
+	// Handle system specific values and remove them from the map for precomputing to not have issues
+	if stats.Loadavg, err = LoadavgFromRawLine(statMap["proc.loadavg"].(string)); err != nil {
+		return stats, nil, fmt.Errorf("Error parsing loadavg for cache '%s': %v", cacheName, err)
```
zrhoffman, line 80: In Go, error messages should not be capitalized.

traffic_monitor/cache/astats_csv.go
```diff
@@ -0,0 +1,114 @@
...
+	} else {
+		delete(statMap, "proc.loadavg")
+	}
+
+	if err := stats.AddInterfaceFromRawLine(statMap["proc.net.dev"].(string)); err != nil {
+		return stats, nil, fmt.Errorf("Failed to parse interface line for cache '%s': %v", cacheName, err)
```
zrhoffman, line 86: In Go, error messages should not be capitalized.

traffic_monitor/cache/astats.go
```diff
@@ -68,50 +69,60 @@ type Astats struct {
...
+	if ctype == "text/json" || ctype == "text/javascript" || ctype == "" {
+		var astats Astats
+		json := jsoniter.ConfigFastest
+		if err := json.NewDecoder(rdr).Decode(&astats); err != nil {
+			return stats, nil, err
+		}
 
-	stats.NotAvailable = astats.System.NotAvailable
+		if err := stats.AddInterfaceFromRawLine(astats.System.ProcNetDev); err != nil {
+			return stats, nil, fmt.Errorf("Failed to parse interface line for cache '%s': %v", cacheName, err)
```
zrhoffman, line 91: This seems like as good of a time as ever to remove the capitalization from this error message.

traffic_monitor/cache/astats.go
```diff
@@ -68,50 +69,60 @@ type Astats struct {
...
+		astats.Ats["system.lastReloadRequest"] = float64(astats.System.LastReloadRequest)
+		astats.Ats["system.notAvailable"] = stats.NotAvailable
+		astats.Ats["system.proc.loadavg"] = astats.System.ProcLoadavg
+		astats.Ats["system.proc.net.dev"] = astats.System.ProcNetDev
+
+		return stats, astats.Ats, nil
+	} else if ctype == "text/csv" {
+		return astatsCsvParseCsv(cacheName, rdr, pollCTX)
+	} else {
+		return stats, nil, fmt.Errorf("Stats Content-Type (%s) can not be parsed by astats", ctype)
```
zrhoffman, line 124: In Go, error messages should not be capitalized.

traffic_monitor/cache/astats.csv
```diff
@@ -0,0 +1,527 @@
+proxy.process.http.completed_requests,26220072200
```
zrhoffman, line 1: If a license header is added to this file, we don't need to re-add a `.csv` exception to `.dependency_license`, right?

traffic_monitor/cache/stats_types.go
```diff
@@ -105,7 +105,7 @@ type StatsDecoder struct {
 // whatever miscellaneous data was in the payload but not represented by
 // the properties of a Statistics object, so that it can be used in later
 // calculations if necessary.
-type StatisticsParser func(string, io.Reader) (Statistics, map[string]interface{}, error)
+type StatisticsParser func(string, io.Reader, interface{}) (Statistics, map[string]interface{}, error)
```
zrhoffman, line 108: It looks like this is always typecast to `*poller.HTTPPollCtx` type. Any reason we can't skip the typecast and make this `*poller.HTTPPollCtx` directly (though I understand, separately, that `handler.Handler` uses `interface{}` type to avoid an import cycle)?

traffic_monitor/cache/stats_over_http.go
```diff
@@ -19,17 +19,19 @@ package cache
...
+	"strings"
 
-import "github.com/apache/trafficcontrol/lib/go-log"
-import "github.com/apache/trafficcontrol/traffic_monitor/todata"
+	"github.com/apache/trafficcontrol/lib/go-log"
 
-import "github.com/json-iterator/go"
+	"github.com/apache/trafficcontrol/traffic_monitor/todata"
+	jsoniter "github.com/json-iterator/go"
+)
```
zrhoffman, line 34: These imports should be split into (at least) 3 groups:
- core package imports
- ATC imports
- external library imports

traffic_monitor/cache/astats_csv.go
rob05c: To add some context as to why: it's so they compose well. For example, if the parent function did
`return errors.New("doing foo: " + err.Error())`, it ends up as a pretty sentence, "doing foo: parsing loadavg...", and the final `log.Errorln` can/should start with a capital (and could add a period at the end, though people don't usually bother with that).

As a nitpick to the nitpick, the word "Error" here is superfluous for the same reason.

traffic_monitor/cache/astats_test.go
```diff
@@ -20,28 +20,99 @@ package cache
...
+	pl := &poller.HTTPPollCtx{HTTPHeader: http.Header{}}
+	ctx := interface{}(pl)
+	ctx.(*poller.HTTPPollCtx).HTTPHeader.Set("Content-Type", "text/csv")
+	_, thismap, err := astatsParse("testCache", file, ctx)
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	t.Logf("Found %v key/val pairs in ats\n", len(thismap))
```
zrhoffman, line 69: Maybe assert the number of key/val pairs the test is supposed to find?

traffic_monitor/cache/stats_types.go
ezelkow1: I left it as interface{} since this is the generic definition of a parser, so it gets used by any potential stats parser plugins. So specifying it as httppoll ties peoples hands as to what a plugin would be able to accept. So I think its best to leave it as generic as possible

traffic_monitor/cache/astats.csv
ezelkow1: Right but its also not indicative of what the astats plugin would normally return, besides csv was normally excluded to begin with

ezelkow1:
Update to use new allocation method to do the data map allocation all up front, turns out this is more performant reading all of the string data in at once to one dynamically allocated block and then basing the size of the map on that so that the map does not have to be continually resized

BenchmarkAstatsJson-8   	    5836	    197696 ns/op	  134473 B/op	    2028 allocs/op
BenchmarkAstatsCSV-8    	   10000	    110413 ns/op	  110065 B/op	     800 allocs/op

traffic_monitor/cache/astats_test.go
```diff
@@ -60,14 +57,11 @@ func TestAstatsCSV(t *testing.T) {
 	pl := &poller.HTTPPollCtx{HTTPHeader: http.Header{}}
 	ctx := interface{}(pl)
 	ctx.(*poller.HTTPPollCtx).HTTPHeader.Set("Content-Type", "text/csv")
-	_, thismap, err := astatsParse("testCache", file, ctx)
+	_, _, err = astatsParse("testCache", file, ctx)
```
zrhoffman, line 60: We still want to verify the correct size of the map, right?

ezelkow1: its more important that it just doesnt crash. There are already differences between both the json and csv outputs because the CSV one will strip out extraneous fields. Just verifying the size doesnt mean much since you could throw some random data into the files and still have it be allocated in to a proper map

This is also all the original test was doing as well, only checking that it didnt crash, beyond printing a value that wasnt checked or useful for anything, however it was only checking the jsoniter unmarshal capability and not checking that the TM functions themselves didnt cause issues

ezelkow1: That being said I could add some value checks in the map just to test for existence, but Id prefer to do that in the next PR. I plan on working on a follow up directly after this is merged to add similar support to stats_over_http, but that is dependent on this baseline being in since it needs the Accept support and changes to allow the contexts to be passed around

zrhoffman (approved):
- TM tests pass
- TM successfully parses CSV polling responses and reports stats correctly
- Nothing important is on the CSV lines that fail to parse and are ignored
- Docs builds and look good

zrhoffman:
> * [ ]  This PR includes an update to CHANGELOG.md OR such an update is not necessary

Since this is a new feature, would you please add a changelog entry?

ezelkow1:
> > * [ ]   This PR includes an update to CHANGELOG.md OR such an update is not necessary
> 
> Since this is a new feature, would you please add a changelog entry?

Opened https://github.com/apache/trafficcontrol/pull/5009

PR #4994, opened 2020-08-27
change server primary key to only id

### LINKED ISSUE #4941: Fix Traffic Ops Server Primary key to be ID
The TO Primary Key is currently `(id, cachegroup, type, status, profile)`.

That is absolutely broken. It must've slipped in a migration or something, but the PK needs to be `id`, and tons of things will break if IDs aren't unique. 

If necessary, we can consider adding the other columns there to their own unique key. But I don't think they need to be.

Personally, I don't like artificial IDs. They're unnecessary and violate, confuse, and obfuscate relationality. Using the real Candidate Key is much better. But as long as we have them, it's imperative they're unique.

## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
TO Server ID is not PK

## Expected / new behavior:
TO Server ID is PK

Insert a Server into the Database with the same ID as another. Observe chaos.

## Anything else:

mattjackson220:

- [x] This PR fixes #4941 

This PR updates the server table so that only the id is the primary key instead of the id, cachegroup, type, status, and profile

- Traffic Ops

verify that all unit and API tests are successful, verify that adding, editing, and deleting a server works as before except that servers cannot have the same id twice

There are no test updates because it is covered by existing tests
There are no CHANGELOG updates since it is not a functionality change
There are no doc updates since it is just a db change

### DIFF
--- a/traffic_ops/app/db/migrations/2020082700000000_server_id_primary_key.sql
+++ b/traffic_ops/app/db/migrations/2020082700000000_server_id_primary_key.sql
@@ -1,0 +1,41 @@
+/*
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+		http://www.apache.org/licenses/LICENSE-2.0
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+-- +goose Up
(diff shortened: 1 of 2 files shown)


rimashah25:
Successful manual tested as follows:
1. Ran a POST /api/3.0/servers to create a new server. Returned 200 response
2. Got the ID (eg 15) from the newly created server.
3. Created another server with the ID (eg 15) received from step 2.
4. Got the following error with a 400 response.
`{"alerts":[{"text":"server with id 15 already exists. Please do not provide an id.","level":"error"}]}`
5. I deleted the existing server (ID:15)
6. Ran a POST call (/api/3.0/servers) with the same ID (15) and was able to create a new server.

Unit tests pass 
API tests pass (v2/v3)
Code is formatted correctly
LGTM

PR #5003, opened 2020-09-02
Missing ProfileId during server creation returns Bad Request error

### LINKED ISSUE #4979: Internal Server Error when create Server without profileId
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
The `POST` handler of the `/servers` endpoint returns a `500 Internal Server Error` response when the request payload did not specify a `profileId`.

## Expected / new behavior:
It used to return 400 with a message to the effect that `profileId` is required.

Create a Server using `https://{{TO_BASE_URL}}/api/3.0/servers` without `profileId`. API will return Internal Server Error.
```json
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```

rimashah25:

- [x] This PR fixes #4979 

- Traffic Ops

1. Run Traffic_Ops locally
2. Run POST call to https://TO's-url:port/api/3.0/servers/ with a missing profileId or profile, or profileDesc
Note: You should the following response
```
{"alerts":
    [{
        "text":"'profileId' is required",
        "level":"error"}
    ]}
}
```

- master (d1a50b0)

- [x] This PR includes tests

## Additional Information

Documentation has the correct example (a successful POST)


traffic_ops/testing/api/v3/tc-fixtures.json
```diff
@@ -3368,4 +3368,4 @@
             "description": "bar"
         }
     ]
-}
+}
```
ocket8888, line 3371: Removing this trailing newline makes this not a text file

rimashah25: aha, I see.. i'll add it back

traffic_ops/v3-client/server.go
```diff
@@ -117,6 +117,9 @@ func (to *Session) CreateServer(server tc.ServerNullable) (tc.Alerts, ReqInf, er
 	resp, remoteAddr, err := to.request(http.MethodPost, API_SERVERS, reqBody, nil)
 	reqInf.RemoteAddr = remoteAddr
 	if err != nil {
+		if strings.Contains(err.Error(), "400 Bad Request[400]") {
+			reqInf.StatusCode = 400
+		}
```
ocket8888, line 122: We should just set `reqInf.StatusCode` to `resp.StatusCode` so that it's directly captured whatever value it happens to have, and doesn't rely on string pattern matching.

rimashah25: Ok. Updated

traffic_ops/v3-client/server.go
```diff
@@ -117,6 +117,9 @@ func (to *Session) CreateServer(server tc.ServerNullable) (tc.Alerts, ReqInf, er
 	resp, remoteAddr, err := to.request(http.MethodPost, API_SERVERS, reqBody, nil)
 	reqInf.RemoteAddr = remoteAddr
 	if err != nil {
+		if strings.Contains(err.Error(), "400 Bad Request[400]") {
+			reqInf.StatusCode = resp.StatusCode
+		}
```
ocket8888, line 122: I think you misunderstood what I was saying here. Instead of checking the error at all, before you even check if it's `nil` you can just do `reqInf.StatusCode = resp.StatusCode`. That way the status code is reported faithfully no matter what it is. Like:
```go
//...
resp, remoteAddr, err := to.request(http.MethodPost, API_SERVERS, reqBody, nil)
reqInf.RemoteAddr = remoteAddr
reqInf.StatusCode = resp.StatusCode
if err != nil {
	return alerts, reqInf, err
}
defer resp.Body.Close()
//...
```

zrhoffman: Also `reqInf.StatusCode = resp.StatusCode` should be inside a `nil` check for `resp`.

rimashah25: Updated.

ocket8888 (approved):
Looks good, tests all pass.

PR #5011, opened 2020-09-03
Enhanced Geo Location bypass in Traffic Router

### LINKED ISSUE #4912: Enhanced Geo Location bypass in Traffic Router
## I'm submitting a ...

- [ ] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [x] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

Currently if a client/resolver IP cannot be resolved via CZF and Maxmind does not know the location, Maxmind defaults to sending a lat/long that is exactly the center of the country.  Traffic Router then uses this to route to the closest cache group, which ends up being whichever is closest to the center of the country.  This can result in a non-trivial amount of traffic being incorrectly routed to one cache group causing potential capacity issues.

## Expected / new behavior:

Traffic Control can be configured such that Traffic Router knows what the default location is for a country and will use the Geo Miss Default Latitude  and Geo Miss Default Longitude for a DS when Maxmind returns the default lat/long for a country.  This will allow the traffic to be sent to a cache group that makes the best sent and can alleviate pressure on the most central cache group. 

When this scenario is hit, TR should log a new rtype so that this traffic can be seen.  It should be something like GEO_DS

How to validate this is working as expected: 

- DS1 is configured with a default lat/long
- TR is configured to know to use the DS default lat/long for DS1
- TR is configured to know the default lat/long for the US (37.751,-97.822)
- If a request for DS1 comes into TR and has the default location, TR uses the default lat/long configured on the DS to choose a cache
- TR logs the message with a new rtype to distinguish this from a regular GEO hit

srijeet0406:

- [x] This PR fixes #4912  

- Documentation
- Traffic Router

There are a few pre reqs to test this PR.
1.) You need to be in an environment that has the `geolocation.polling.url` set to `http://cdn-dczf.cdn.comcast.net/maxmind/GeoIP2-City.mmdb.gz` if you're on comcast VPN/ network. If on the public network, you can set it to `https://mailfud.org/geoip-legacy/GeoIPCity.dat.gz`.
NOTE: CiaB does NOT use this version of Maxmind, and instead uses the `Lite` version of that database. This will not work for this PR. You could use staging/ nightly if you're on the comcast VPN/ network. You can also override the param value to point to the new value, and then do a snapshot of the CDN.
2.) You will also need to set the `maxmind.default.override` for the `CRConfig.json` to point to the default coordinates of the country. This can be done for multiple countries. The value for my setup is `US;37.751,-97.822`.
3.) Now, you will need to route traffic through this TR, using a spoofed IP. This IP should be one that has the city and postal code, and no subdivisions, and should be in the US (since our override is configured with the default coordinates of the US). I use this IP for my tests `103.111.32.61` 
4.) Tail the logs of the TR docker container
5.) The way I use this spoofed IP in my CiaB environment (which has all the above mentioned changes) is this:
Exec into the enroller container of CiaB: `docker-compose exec enroller /bin/bash`
Curl the DS URL of CiaB, with a spoofed IP: `curl -H x-mm-client-ip:103.111.32.61 -L http://video.demo1.mycdn.ciab.test`
6.) You should see a `rType` of `GEO_DS` in the logs, for your client IP. This is the new rType that says that the default miss location of the DS was used.

- master

- [x] This PR includes tests


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
```diff
@@ -513,8 +514,9 @@ public Geolocation getLocation(final String clientIP, final DeliveryService deli
 			}
 		}
 
-		if (clientLocation.isDefaultLocation() && defaultGeolocationsOverride.containsKey(clientLocation.getCountryCode())) {
-			clientLocation = defaultGeolocationsOverride.get(clientLocation.getCountryCode());
+		if (clientLocation.isDefaultLocation() && getDefaultGeoLocationsOverride().containsKey(clientLocation.getCountryCode())) {
+			clientLocation = deliveryService.getMissLocation();
```
limited, line 518: What happens if `missLocation` is null? 

I think the most consistent behavior for a null `missLocation` would be to continue using 
```
defaultGeolocationsOverride.get(clientLocation.getCountryCode());
````

File: traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java [hunk 2]
```diff
@@ -523,7 +530,11 @@ public Geolocation getLocation(final String clientIP, final DeliveryService deli
 			track.setResultDetails(ResultDetails.GEO_NO_CACHE_FOUND);
 		}
 
-		track.setResult(ResultType.GEO);
+		if (useDSDefaults) {
```
mattjackson220, line 533: would this be able to go into the if statement above? instead of having the useDSDefaults boolean just set the result type in that if statement instead? then we could get rid of the SuppressWarnings too

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/StatTracker.java
```diff
@@ -184,7 +184,13 @@ public void setRegionalAlternateCount(final int regionalAlternateCount) {
...
 			 * This value indicates that routing was blocked in accordance with anonymous blocking configurations.
 			 */
-			ANON_BLOCK
+			ANON_BLOCK,
+			/**
+			 * This value indicates that routing was based on the default lat/long of the delivery service, because maxmind
+			 * returned the centre of the country as the client location, due to the CZF not being able to resolve the client IP
+			 * to a valid location.
+			 */
+			GEO_DS
```
mattjackson220, line 193: do you think there would be any gain to add another result type that let us know that it is just using the default from maxmind (like if it doesnt have a DS default and the client ip didnt resolve to a valid location)? not sure if theres any gain there but might just give us more info quickly?

srijeet0406: This was part of the ticket description from @dneuman64. I think he wants it so that the Maple team can easily see how many events matched this type.

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/TrafficRouter.java
srijeet0406: Done 👍

traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/StatTracker.java
mattjackson220: yea thats why i was thinking it might be good to add another one in there that tells us that it is just using the maxmind defaults. eh if theres value there itll be an easy change to add in if people decide they want it

mattjackson220 (approved):
Looks good! All tests pass, docs build and look good, manual testing in ciab works as expected. nice work!

phusinh:
can you tell me if http://video.demo1.mycdn.ciab.test is the origin server or the edge server or the fqdn of the traffic router

PR #5027, opened 2020-09-15
Fixed reference URLs on delivery service (DNS, HTTP) pages

rimashah25:

- [x] This PR fixes reference URLs on Delivery Service Page, specifically for type: HTTP and DNS 

- Traffic Portal

1. Go to traffic_portal's create a delivery service page.
2. Select either HTTP, or DNS type
3. Go to section Cache Configuration Settings, and hover over every field in that section and click on the links present in help tool tip.
4. No link should show a `page does not exists yet` error.

master (d809e9a)

## Additional Information

The PR doesn't need documentation or tests since it only updates the Reference URL path in Traffic Portal. No logic has been implemented.

### DIFF
--- a/traffic_portal/conf/config.js
+++ b/traffic_portal/conf/config.js
@@ -21,8 +21,8 @@
 module.exports = {
     timeout: '120s',
     useSSL: true, // set to true if you plan to use https (self-signed or trusted certs).
-    port: 80, // set to http port
-    sslPort: 443, // set to https port
+    port: 8080, // set to http port
+    sslPort: 9090, // set to https port
     // if useSSL is true, generate ssl certs and provide the proper locations.
     ssl: {
         key:    '/etc/pki/tls/private/localhost.key',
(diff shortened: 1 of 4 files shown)


traffic_portal/conf/config.js
```diff
@@ -21,8 +21,8 @@
 module.exports = {
```
ocket8888, line 21: I don't think you meant to commit this file

rimashah25: oops..

ocket8888 (approved):
This looks like it works for what it changes - however, HTTP and DNS Delivery Service pages in TP still have two broken references. The first is in the "Deep Caching" field's tooltip, where it should probably be `https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_router.html#deep-caching` to link to the correct anchor. The other is in the tooltip for "Geo Limit Redirect URL" and to be honest I'm not sure where that's supposed to take you.

Neither of these land on 404 pages, they're just non-existent anchors on an existent page. So you don't need to fix that if you don't want to, but I thought I'd bring it up.

rimashah25:
> This looks like it works for what it changes - however, HTTP and DNS Delivery Service pages in TP still have two broken references. The first is in the "Deep Caching" field's tooltip, where it should probably be `https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_router.html#deep-caching` to link to the correct anchor. The other is in the tooltip for "Geo Limit Redirect URL" and to be honest I'm not sure where that's supposed to take you.
> 
> Neither of these land on 404 pages, they're just non-existent anchors on an existent page. So you don't need to fix that if you don't want to, but I thought I'd bring it up.

I updated both these reference urls.

ocket8888 (approved):

PR #5034, opened 2020-09-16
updated traffic_ops_ort.pl to use multistep mv for config shuffle

### LINKED ISSUE #4311: Fix ORT to atomically place files.
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [x] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
ORT does not atomically place new files. It uses `cp`, which is not atomic. 

This means if somethings fails – Perl crashes, the OS crashes, the syscall fails for some reason – the file will be half-written, causing ATS reloading to fail.

## Expected / new behavior:
The proper procedure is to make a new temporary file, and then call `mv`, which is guaranteed to be atomic, and either succeed or fail.

Code is here:

https://github.com/apache/trafficcontrol/blob/master/traffic_ops/ort/traffic_ops_ort.pl#L1575

Difficult to reproduce. You'd have to run ORT and kill the Perl when it was in the middle of a file write. But it would definitely happen. And then your ATS reload or restart would fail.

MylesBock:
Closes #4311 

* Replaced /bin/cp with /bin/mv and changed target to same wd as target with new suffix

* Added error handling surrounding `system` use in that block

* Added second step to mv old config to the same wd with `.old` suffix, then mv from `.new`

First PR opened here, so may need guidance on how to test this

- [x] This PR fixes #4311 OR is not related to any Issue 

- CDN in a Box
- Documentation
- Grove
- Traffic Control Client 
- Traffic Monitor
- Traffic Ops
- [x] Traffic Ops ORT
- Traffic Portal
- Traffic Router
- Traffic Stats
- Traffic Vault
- CI tests

Validate a new configuration can be generated and that the old can be supplanted via ORT


MylesBock:
Drafting.

https://github.com/MylesBock/trafficcontrol/tree/experimental/restructure

Will return when this project is more conducive of outside help, cya

PR #5041, opened 2020-09-17
Change ORT to not update ip_allow except badass

rob05c:
ATS has a known bug where changing ip_allow.config causes random
blocking on config reload. We changed ORT a while back to not reload
when it changes, but other files can later trigger a reload.

This changes ORT to not update the file at all, and log an error.
This will cause any added servers to not be added to the allow,
likely breaking Edges. But breaking an Edge is better than
breaking a Mid.

Further, the error log will allow users to create alarms, so
they know to go in and manually badass and restart the machine.

I've manually tested, verified server IP changes log the error but don't update the file, changes to other files apply as expected, badass mode updates ip_allow.config as expected.

Includes changelog.
No docs, ORT config file details are not documented.
No tests, ORT itself doesn't have an integration test framework yet. 

- Traffic Ops ORT

Change or add a Server in Traffic Ops with a new IP, run ORT in syncds mode, verify an Error is logged but the file is not updated. Change something else, like a DS Origin, verify change still applies correctly. Run ORT in badass move, verify ip_allow.config is updated.

It's behaved this way forever

## Additional Information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Changed Traffic Portal to use the more performant and powerful ag-grid for all server tables.
 - Changed ORT Config Generation to be deterministic, which will prevent spurious diffs when nothing actually changed.
 - Changed ORT to find the local ATS config directory and use it when location Parameters don't exist for many required configs, including all Delivery Service files (Header Rewrites, Regex Remap, URL Sig, URI Signing).
+- Changed ORT to not update ip_allow.config but log an error if it needs updating in syncds mode, and only actually update in badass mode.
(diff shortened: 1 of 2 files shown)


ezelkow1 (approved):

dneuman64:
Since this only affects Mids and not Edges, is there a way to only apply this to Mids?  Maybe we can make it a config or flag and we can set it up such that edges update the file and mids do not?

rob05c:
Edges have the same problem. They just don't update their config very often.
But if an Edge ip_allow.config changes, ATS starts blocking random addresses the same.

dneuman64:
Are you sure because I don't think we have seen that issue on edges before?  We have only seen this issue on Mids.
I still think we should consider making this a flag.  It would be beneficial for if/when the bug gets fixed to be able to put ip_allow.config in place without having to BA ATS every time

rob05c:
I am positive. We haven't seen it on Edges because they're basically open to the internet, and almost never change. If `ip_allow.config` changed on an Edge for any reason, ATS would start blocking random addresses.

I can add a flag.

ezelkow1 (approved):

Issue #5021, opened 2020-09-09
Internal Server Error when /GET asns API with cachegroup query parameter


dpham692:
## I'm submitting a ...

-  bug report

## Traffic Control components affected ...

-  Traffic Ops

## Current behavior:

Perform a /GET `https://{{TO_BASE_URL}}/api/3.0/asns?cachegroup={{cachegroupName}}`
API will return Status Code 500
```
{
    "alerts": [
        {
            "text": "Internal Server Error",
            "level": "error"
        }
    ]
}
```

## Expected / new behavior:

API should return correct ASNs that link with the cache group.

srijeet0406:
This is expected behavior.
When you search by cachegroup name, you need to use `cachegroupName` as the query param.
If you want to use cachegroup ID, use `cachegroup`.
So the following two calls should work:
`GET /api/3.0/asns?cachegroupName=<your cg name here>` and
`GET /api/3.0/asns?cachegroup=<your cg ID here>`.

I spoke to @dpham692 and he was ok with closing this issue.

zrhoffman:
Since the `cachegroup` query parameter should be an ID, it should be validated and return 400 bad request instead of 500 internal server error when the user provides a string.

PR #5043, opened 2020-09-17
Creating an ASN with the same number and same cache group should not be allowed

### LINKED ISSUE #5020: Bug that let user create two ASNs with same number in the same cachegroup
## I'm submitting a ...

-  bug report

-  Traffic Ops

## Current behavior:

1. Go to Traffic Portal
2. Click on Topology -> ASNs.
3. Create a ASNs with valid number.
4. Create another ASNs with same number in the same cache group like step 3. Notice that TP let user create it.
## Expected / new behavior:

TO and TP should not let user create 2 ASNs with same number in the same cachegroup

srijeet0406:

- [x] This PR fixes #5020 #5021  

- Traffic Ops

Make sure the unit tests and API tests pass
Run TO and TP locally, and try to create two asns with the same number and same cachegroup ID. This should fail.
Also, try to GET an asn by specifying a non numeric ID for the cachegroup ID as the query param. You should get a 400 instead of a 500 now.

- master

- [x] This PR includes tests

### DIFF
--- a/docs/source/api/v3/asns.rst
+++ b/docs/source/api/v3/asns.rst
@@ -106,6 +106,7 @@ Response Structure
 ``POST``
 ========
 Creates a new :abbr:`ASN (Autonomous System Number)`.
+.. note:: There cannot be two different ASN object with the same ``asn``.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
--- a/docs/source/api/v3/asns.rst
+++ b/docs/source/api/v3/asns.rst
@@ -106,6 +106,7 @@ Response Structure
 ``POST``
 ========
 Creates a new :abbr:`ASN (Autonomous System Number)`.
+
 .. note:: There cannot be two different ASN object with the same ``asn``.
 
 :Auth. Required: Yes
--- a/traffic_ops/traffic_ops_golang/asn/asns_test.go
+++ b/traffic_ops/traffic_ops_golang/asn/asns_test.go
@@ -134,3 +134,39 @@ func TestValidate(t *testing.T) {
 		t.Errorf(`expected %v,  got %v`, expected, errs)
 	}
 }
+
+func TestValidateSameAsnAndCachegroup(t *testing.T) {
+	expected := `an asn with the specified number and cachegroup already exists`
+	mockDB, mock, err := sqlmock.New()
+	if err != nil {
+		t.Fatalf("an error '%s' was not expected when opening a stub database connection", err)
+	}
+	defer mockDB.Close()
+
+	db := sqlx.NewDb(mockDB, "sqlmock")
+	defer db.Close()
+
+	cols := []string{"id"}
+	rows := sqlmock.NewRows(cols)
+	rows = rows.AddRow(
+		1,
+	)
+	mock.ExpectBegin()
+	mock.ExpectQuery("SELECT").WillReturnRows(rows)
+	mock.ExpectCommit()
+
+	reqInfo := api.APIInfo{Tx: db.MustBegin(),}
+	asnNum := 2
+	cachegroupID := 10
+	asn := TOASNV11{
+		api.APIInfoImpl{&reqInfo},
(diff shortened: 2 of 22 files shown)


jhg03a:
It's also invalid to assign an ASN to more than one cachegroup

mattjackson220:
I think it might be worth throwing a note in the ASNs docs to call this out as well? or is it a common enough rule that people will know?

rawlinp:
@jhg03a is that actually invalid or just something that seems invalid in our environment? I could imagine an environment having the same ASN on multiple cachegroups.

jhg03a:
It's required for us, but I think it ought to be required generally even though it's not been enforced today.  The goal of tracking the mapping of ASNs to Cachegroups is so that a transitive mapping between subnets and cachegroups can be created as part of CZF file generation.  Trying to assign the same subnet to multiple cachegroups doesn't make a ton of sense to me.  Keep in mind too that some entities don't break their subnets into geographically based ASNs which CDNs have no control over or input on.  So in this case the argument for multiple mapping could be made but it's still just guessing at that point as the CZF file used by TR will only ever use one path to the best of my knowledge.  I think that's still the right behavior as well because you want IPs to land in cachegroups by way of CZF hits consistently.  Having multiple options there could lead to indeterminate behavior also making cachegroup capacity planning more difficult.

rawlinp:
CZF generation doesn't really have a bearing on how this API should be restricted since that is external to the API, and adding new restrictions to the API like that might violate existing data, causing the DB upgrade to fail.

srijeet0406:
So I'm assuming that the scope of this ticket will be reverted to `Creating an ASN with the same number and same cache group should not be allowed`?

rob05c:
Does it ever make sense to assign the same subnet to multiple cachegroups?

What will Traffic Router do if that happens? Is it ever useful to do that?

>adding new restrictions to the API like that might violate existing data, causing the DB upgrade to fail (assuming it would be a DB constraint).

Can we at least make it an API constraint to prevent new ones, if it's always wrong?

rawlinp:
I disagree for the reasons I've already stated, but if @rob05c and @jhg03a feel strongly enough for us to add API restrictions for external business reasons, then feel free.

rob05c:
I don't know enough about the issue to feel strongly. That's why I was asking, I honestly don't know the answer to those questions.

jhg03a:
I do believe we should for reasons I've already stated.

srijeet0406:
If thats the case then no more changes are required to this PR.

docs/source/api/v3/asns.rst
```diff
@@ -106,6 +106,7 @@ Response Structure
 ``POST``
 ========
 Creates a new :abbr:`ASN (Autonomous System Number)`.
```
mattjackson220, line 108: needs a blank line before the note as well

traffic_ops/traffic_ops_golang/asn/asns_test.go
```diff
@@ -134,3 +134,39 @@ func TestValidate(t *testing.T) {
...
+	reqInfo := api.APIInfo{Tx: db.MustBegin()}
+	asnNum := 2
+	cachegroupID := 10
+	asn := TOASNV11{
+		api.APIInfoImpl{&reqInfo},
+		tc.ASNNullable{ASN: &asnNum, CachegroupID: &cachegroupID},
+	}
+	err = asn.ASNExists()
+	if err == nil {
+		t.Fatalf("expected no error but got %v", err.Error())
```
mattjackson220, line 167: this error message is flipped. should be "expected an error but got error = nil"

ocket8888: it's also going to segfault whenever the test fails

mattjackson220 (approved):
Looks good! unit tests pass, API v1, v2, and v3 tests pass, docs build and look good, manual testing works as expected.

PR #5045, opened 2020-09-17
Fix ORT remap.config cachekey to be deterministic

rob05c:
Fixes ORT remap.config generation to be deterministic, with respect to cache key plugin parameters.

No changelog, changelog already exists for ORT deterministic config gen, this was just an oversight.
No new tests, remap.config already has extensive unit tests.
No docs, no interface change.

- Traffic Ops ORT

Run unit tests.
Run ORT, verify config is semantically unchanged. Run ORT multiple times with cache key parameters, verify output is always the same.

Not a bug fix.

## Additional Information


alficles (approved):
Yeah, this looks right. At a minimum, consistent is better than arbitrary.

PR #5049, opened 2020-09-18
4.1.x Backport: Fix ORT parent orders to be deterministic

rob05c:
Backported from Github #5048


alficles (changes requested):
I think this will incorrectly sort these values that should already be sorted by ParentInfoSortByRank. Maybe that function needs more sort specificity?

rob05c:
>I think this will incorrectly sort these values that should already be sorted by ParentInfoSortByRank. Maybe that function needs more sort specificity?
Fixed

alficles (approved):
Looks good now.

PR #5052, opened 2020-09-18
TP: adds warning message for large IP CIDR

### LINKED ISSUE #5038: Traffic Portal warning when user enters unusually large CIDR
## I'm submitting a ...

-  Documentation
-  Traffic Portal

## Current behavior:

Currently, a user can add an IP for a server with an unreasonable CIDR.  This is specifically true for IPv6 where CIDRs are commonly included with the IP Address. We would usually expect a /64 with an IP but a user can enter something like a /12.  This leads to issues with the way ip_allow.config is generated on caches which can result in requests being incorrectly denied.

## Expected / new behavior:

If a user enters less than a /64 for IPv6 and less than a /27 for Ipv4, Traffic Portal should show the user a warning and make sure they really want to enter a CIDR that large.

Create a new server, add an IPv6 with a /30 and a IPv4 with a /24.

mitchell852:
^^ the PR title sums it up.

- [x] This PR fixes #5038 

- Traffic Portal

Navigate to trafficportal.domain.com/#!/servers/new and create a server interface with an ipv4 or ipv6 address with the following:

- if ipv4, enter CIDR under 27. i.e. 1.1.1.1/8
- if ipv6, enter CIDR under 64. i.e. 2001:558:fe2a:30::2/32

and you'll see a warning message like such:

image

### DIFF
--- a/traffic_portal/app/src/common/modules/form/server/FormServerController.js
+++ b/traffic_portal/app/src/common/modules/form/server/FormServerController.js
@@ -140,6 +140,24 @@ var FormServerController = function(server, $scope, $location, $state, $uibModal
         getProfiles($scope.server.cdnId); // and get a new list of profiles (for the selected cdn)
     };
 
+    $scope.isLargeCIDR = function(address) {
+        if($scope.IPWithCIDRPattern.test(address) && /.+\/(\d+)/.test(address)) {
+            let ip = address.split("/")[0],
+                cidr = parseInt(address.split("/")[1],10);
+
+            if ($scope.IPv4Pattern.test(ip)) {
+                if (cidr < 27) {
+                    return true;
+                }
+            } else {
+                if (cidr < 64) {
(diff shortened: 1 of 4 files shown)


traffic_portal/app/src/common/modules/form/server/FormServerController.js
```diff
@@ -140,6 +140,24 @@ var FormServerController = function(server, $scope, $location, $state, $uibModal
         getProfiles($scope.server.cdnId); // and get a new list of profiles (for the selected cdn)
     };
 
+    $scope.isLargeCIDR = function(address) {
+        if($scope.IPWithCIDRPattern.test(address) && /.+\/(\d+)/.test(address)) {
```
mitchell852, line 144: quite sure there is a more elegant way to do this :P

shamrickus (approved):
LGTM

PR #5058, opened 2020-09-21
changed OAuth workflow to allow local accounts too

mattjackson220:

This PR allows for local accounts to be used when OAuth is enabled. This is useful for service accounts and for a backup in case the OAuth provider fails for whatever reason. 

- Traffic Ops
- Traffic Portal

Set up OAuth with an OAuth provider and enable it in the traffic_portal_properties.json file with the correct info for your OAuth provider. 
Verify that with that enabled, TP shows the "Log in with SSO" button in addition to the normal login screen.  
Verify that you can log in with SSO as before.  
Verify that you can log in with a local account as before.

No test are updated as this is just combining previous functionality.


shamrickus (approved):
LGTM:
* Able to login with SSO
* Still able to login with local logins

PR #5064, opened 2020-09-23
Sorting Traffic Ops GET /api/*/cdns alphabetically

### LINKED ISSUE #3455: Traffic Ops GET /api/*/cdns should be sorted alphabetically by name by default
When clicking the snapshot button (the camera icon in the top right) in Traffic Portal, it shows the list of CDNs to choose from. Today, these CDNs appear to be sorted ascending by their numerical ID rather than alphabetically by their name. With a high number of CDNs to choose from, it would be easier to sift through the list alphabetically.

The CDN table (viewed by clicking the "CDNs" tab on the left) should also be sorted alphabetically by name. This appears to be a minor regression from the Perl TO API implementation which sorted by name by default (if no other `orderby` param is specified in the request URL): https://github.com/apache/trafficcontrol/blob/master/traffic_ops/app/lib/API/Cdn.pm#L34

rimashah25:

- [x] This PR fixes #3455 

- Traffic Ops

Login to TP UI and click on CDNs tab to ensure the CDNs are listed alphabetically
OR
Login to TP UI and click on Snapshot or Queue Server update and the drop down list should be sorted alphabetically
OR
Curl https://localhost:8443/api/3.0/cdns and the response list should be sorted alphabetically.

master (7fa607e)

- [x]  This PR includes tests

## Additional Information

PR testing can be done using API testing or visually looking at the TP UI (CDN tab or Snapshot tab)
Documentation for GET API calls is up to date.

### DIFF
--- a/traffic_ops/testing/api/v3/cdns_test.go
+++ b/traffic_ops/testing/api/v3/cdns_test.go
@@ -102,12 +102,11 @@ func SortTestCDNs(t *testing.T) {
 		sortedList = append(sortedList, resp[i].Name)
 	}
 
-	res1 := sort.SliceIsSorted(sortedList, func(p, q int) bool {
+	res := sort.SliceIsSorted(sortedList, func(p, q int) bool {
 		return sortedList[p] < sortedList[q]
 	})
-
-	if res1 != true {
-		t.Errorf("Slice is not sorted by their names")
+	if res != true {
+		t.Errorf("list is not sorted by their names: %v", sortedList)
 	}
 }
 
--- a/traffic_ops/testing/api/v3/cdns_test.go
+++ b/traffic_ops/testing/api/v3/cdns_test.go
@@ -32,6 +32,7 @@ func TestCDNs(t *testing.T) {
 		var header http.Header
 		header = make(map[string][]string)
 		header.Set(rfc.IfModifiedSince, time)
+		SortTestCDNs(t)
 		UpdateTestCDNs(t)
 		GetTestCDNs(t)
 		GetTestCDNsIMSAfterChange(t, header)
@@ -90,6 +91,16 @@ func CreateTestCDNs(t *testing.T) {
 		}
 	}
 
+}
+
+func SortTestCDNs(t *testing.T) {
+	var header http.Header
+	for _, cdn := range testData.CDNs {
+		_, reqInf, err := TOSession.GetCDNByNameWithHdr(cdn.Name, header)
+		if reqInf.StatusCode == http.StatusInternalServerError {
+			t.Errorf("has issue sorting CDN list %v:", err)
+		}
+	}
 }
 
 func UpdateTestCDNs(t *testing.T) {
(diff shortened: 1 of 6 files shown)


rawlinp (changes requested):
This seems kind of a roundabout way of providing a default sort for these various endpoints. Most of these support an `orderby` query parameter, and I think at least some of our endpoints will basically just choose a "default" `orderby` -- e.g. name, xmlID, etc. See traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go L999

rawlinp (changes requested):
Ok, since these all use `GenericRead`, that previous advice is not super helpful.

For the handlers that call `GenericRead`, I think they can add the default `orderby` parameter to `APIInfo.Params` (if one isn't present already), before calling `GenericRead`. Looking at the code, that _should_ work.

rimashah25:
> Ok, since these all use `GenericRead`, that previous advice is not super helpful.
> 
> For the handlers that call `GenericRead`, I think they can add the default `orderby` parameter to `APIInfo.Params` (if one isn't present already), before calling `GenericRead`. Looking at the code, that _should_ work.

Completed

mitchell852:
should this PR be named something different since it's more than that?

rawlinp:
It was more before, but we discussed the previous version offline and changed the approach. I believe @rimashah25 is going to add default sorting to the other API endpoints separately.

traffic_ops/traffic_ops_golang/cdn/cdns.go
```diff
@@ -136,6 +137,8 @@ func (cdn *TOCDN) Read(h http.Header, useIMS bool) ([]interface{}, error, error,
 	// sorting alphabetically by cdn name.
 	if _, ok := cdn.APIInfo().Params["orderby"]; !ok {
 		cdn.APIInfo().Params["orderby"] = "name"
+	} else {
+		return nil, fmt.Errorf("unable to sort the list"), nil, http.StatusInternalServerError, nil
```
ocket8888: Doesn't this mean that if I try to set `orderby` myself in the URL it'll give me back an internal server error?

rimashah25: Resolved conversation based on offline chat.

traffic_ops/testing/api/v3/cdns_test.go
```diff
@@ -92,6 +93,16 @@ func CreateTestCDNs(t *testing.T) {
 
 }
 
+func SortTestCDNs(t *testing.T) {
+	var header http.Header
+	for _, cdn := range testData.CDNs {
+		_, reqInf, err := TOSession.GetCDNByNameWithHdr(cdn.Name, header)
```
rawlinp, line 99: So this doesn't really verify that the response is sorted alphabetically. Maybe we should take the response of all the CDNs, iterate over them, and use `strings.Compare` to make sure the response is sorted as expected.

rimashah25: Fixed test case.

rawlinp:
Since we can't reorder the existing CDNs in tc-fixtures.json, can we add a _new_ CDN named something like `a-cdn` on the end?

rimashah25:
> Since we can't reorder the existing CDNs in tc-fixtures.json, can we add a _new_ CDN named something like `a-cdn` on the end?

I can try.. but messing with tc_fixtures is kinda harrowing experience.

rawlinp (approved):
LGTM, checks passed

PR #5068, opened 2020-09-24
[4.1.x] Make 4.1.x TM backwards compatible with pre-4.0 TO

### LINKED ISSUE #5005: Traffic Monitor Cannot Be Upgraded Independently of Traffic Ops
Traffic Monitor uses the unvendored Traffic Ops client. This means if the TO API version changes, the client automatically uses the latest, and TM will fail to connect to TO if TM is upgraded first.

ATC supports upgrading in any order, which is very frequently necessary for innumerable reasons. So, this is a bug.

It's been a bug for a while, but it most recently came up because of another bug in 4.0.x, which being able to upgrade TM first would have worked around.

We should fix this by vendoring the client, using the new client and falling back if any new data is necessary (like ORT does, see https://github.com/apache/trafficcontrol/tree/master/traffic_ops_ort/atstccfg/toreq and https://github.com/apache/trafficcontrol/tree/master/traffic_ops_ort/atstccfg/toreqnew).

This should be fixed in `master` and backported to 4.1.x and a new 4.1.1 Release made with the fix.

I recommend coordinating with #5006 to fix both in the same release.

## I'm submitting a ...
-  bug report

-  Traffic Monitor

## Current behavior:
Traffic Monitor can not be upgraded independently of Traffic Ops, arbitrarily depending on the ATC version and whether the API was changed.

## Expected / new behavior:
Traffic Monitor can be upgraded before or after Traffic Ops, as necessary.

Upgrade Traffic Monitor from 3.0 to 4.0 before upgrading Traffic Ops 3.0, observe failure to connect.

rawlinp:
NOTE: this is not a true backport, but these changes were extracted from
similar changes made to TM in master.

- [x] This PR fixes #5005 

- Traffic Monitor

With a pre-4.0 TO, upgrade TM to this version, verify that TM attempts to use TO API 2.0 before falling back to 1.3.

- 4.0
- 4.1

- [x] This PR is a pseudo-backport, but this change had no corresponding tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Added
 - Added the ability to set TLS config provided here: https://golang.org/pkg/crypto/tls/#Config in Traffic Ops
 
+### Fixed
+- Fixed #5005: Traffic Monitor cannot be upgraded independently of Traffic Ops
+
 ### Deprecated
 - Deprecated the `insecure` option in `traffic_ops_golang` in favor of `"tls_config": { "InsecureSkipVerify": <bool> }`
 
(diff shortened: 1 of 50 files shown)


rawlinp:
Merging this per the testing feedback from @smalenfant.

PR #5071, opened 2020-09-25
Delivery Service Request Overhaul

### LINKED ISSUE #5012: Internal Server Error when missing key in POST to /deliveryservice_requests
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
When passing a request that's missing the `deliveryservice` property to `/deliveryservice_request`'s POST handler, the request's goroutine will segfault and panic.

## Expected / new behavior:
Invalid requests should not be capable of causing a server-side segfault.

POST to `/deliveryservice_request` with a request body not containing a `deliveryservice` property (probably `"deliveryservice": null` works too, but I haven't checked). Simplest body to reproduce the error is just `{}`.

## Anything else:
<details>
<summary>Full HTTP stack</summary>

```http
POST /api/2.0/deliveryservice_requests HTTP/1.1
User-Agent: python-requests/2.22.0
Accept: */*
Connection: keep-alive
Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1OTkxODIwNjQsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--fd12037b48601a4f29a16fd06a9881c19be9456b
Content-Length: 2

{}
```

```http
Content-Encoding: gzip
Set-Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1OTkxNjQwNjQsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--de133c58fd1afe4b253fe6d920be7002dfdb4513; Path=/; Expires=Thu, 03 Sep 2020 20:14:24 GMT; Max-Age=3600; HttpOnly
Vary: Accept-Encoding
Whole-Content-Sha512: /HT8WHLIsAIPgTPjZAa4GmpUwYtwph3fcU1//Cr7W4rkL1lKRW3kUEDTIz1DoA3/zUm3oa0elunc5VHcism6Qg==
Date: Thu, 03 Sep 2020 19:14:24 GMT
Content-Length: 83

{
	"alerts": [
		{
			"level": "error"
		}
	]
}
```

</details>

<details>
<summary>Traffic Ops Logs</summary>

```
INFO: routing.go:249: 2020-09-03T19:05:31.140864324Z: POST /api/2.0/deliveryservice_requests? handling (reqid 2750)
DEBUG: plugin.go:190: 2020-09-03T19:05:31.140914016Z: DEBUG plugins.OnRequest calling 0 plugins
ERROR: api.go:222: 2020-09-03T19:05:31.151730185Z: 192.168.176.1:41492 panic: (err: runtime error: invalid memory address or nil pointer dereference) stacktrace:
goroutine 21149 [running]:
github.com/apache/trafficcontrol/lib/go-util.Stacktrace(0xc000fed398, 0xce4900, 0x15d8b00)
	/tmp/go/src/github.com/apache/trafficcontrol/lib/go-util/util.go:32 +0x9d
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware.WrapPanicRecover.func1.1(0xf74760, 0xc0016a17d0, 0xc001368400)
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go:152 +0x54
panic(0xce4900, 0x15d8b00)
	/usr/local/go/src/runtime/panic.go:969 +0x166
github.com/apache/trafficcontrol/lib/go-tc.(*DeliveryServiceNullable).Validate(0x0, 0xc000f98100, 0x3, 0x4)
	/tmp/go/src/github.com/apache/trafficcontrol/lib/go-tc/deliveryservices.go:543 +0x263
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice/request.(*TODeliveryServiceRequest).Validate(0xc000eda000, 0xda2260, 0xc000eda000)
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice/request/validate.go:63 +0x43e
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api.decodeAndValidateRequestBody(0xc001368400, 0x7f9b62f5e2b8, 0xc000eda000, 0x0, 0x0)
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api/shared_handlers.go:115 +0x13d
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api.CreateHandler.func1(0xf74760, 0xc0016a17d0, 0xc001368400)
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api/shared_handlers.go:615 +0x10da
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware.AuthBase.GetWrapper.func1.1(0xf74760, 0xc0016a17d0, 0xc001368400)
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go:107 +0x276
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware.WrapPanicRecover.func1(0xf74760, 0xc0016a17d0, 0xc001368400)
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go:156 +0x79
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware.WrapHeaders.func1(0xf775e0, 0xc00333c420, 0xc001368400)
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go:136 +0x643
net/http.HandlerFunc.ServeHTTP(0xc00047e9b0, 0xf775e0, 0xc00333c420, 0xc001368400)
	/usr/local/go/src/net/http/server.go:2012 +0x44
net/http.(*timeoutHandler).ServeHTTP.func1(0xc00333c540, 0xc000191c40, 0xc00333c420, 0xc001368400, 0xc000114360)
	/usr/local/go/src/net/http/server.go:3238 +0x7f
created by net/http.(*timeoutHandler).ServeHTTP
	/usr/local/go/src/net/http/server.go:3232 +0x226
```

</details>

### LINKED ISSUE #5014: Creating a Delivery Service Request with an ID fails
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
Because of the way workflow transitions are verified, creating a DSR with any non-null value in the `id` property will cause the request to fail. If `id` was an integer, though, it will give back the error-level alert: `sql: no rows in result set`, which is both cryptic and - looking at the code - unnecessary.

## Expected / new behavior:
If IDs have special meaning to requests, they should be documented - but it's more likely that the intended behaviour is just to ignore the ID, since it can't reasonably be known before the object is created.

Submit a POST request to `/deliveryservice_requests` with an integral ID value. The simplest request body to reproduce the error is just `{"id": -1}` (which should work regardless of extant data).
## Anything else:
<details>
<summary>Full HTTP Stack</summary>

```http
POST /api/2.0/deliveryservice_requests HTTP/1.1
User-Agent: python-requests/2.22.0
Accept: */*
Connection: keep-alive
Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1OTkxOTI1ODIsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--2d0e7a169ef17f538d58830ee8451e2113562119
Content-Length: 9

{"id": 2}
```
```http
HTTP/1.1 400 Bad Request
Content-Encoding: gzip
Set-Cookie: mojolicious=eyJhdXRoX2RhdGEiOiJhZG1pbiIsImV4cGlyZXMiOjE1OTkxNzQ1ODIsImJ5IjoidHJhZmZpY2NvbnRyb2wtZ28tdG9jb29raWUifQ--88ac7381f3f3d92bb3fd8dc8975be3896a2acc3d; Path=/; Expires=Thu, 03 Sep 2020 23:09:42 GMT; Max-Age=3600; HttpOnly
Vary: Accept-Encoding
Whole-Content-Sha512: kLdrEUgBgalHQ/z01sSUnd5k/TCwjNOnjY6tZWxY88JIhEsiOPoACLxUIGPad+ErIRJ+l2mpC5g8/89l34o0jA==
Date: Thu, 03 Sep 2020 22:09:42 GMT
Content-Length: 88

{
	"alerts": [
		{
			"text": "sql: no rows in result set",
			"level": "error"
		}
	]
}
```

</details>

### LINKED ISSUE #5015: Creating a Delivery Service Request has issues verifying tenant access
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
When submitting a POST request to `/deliveryservice_requests`, if the request body's `deliveryservice` object has no `tenantId` property (or probably also `null` but I haven't checked) the handler will return a 500 Internal Server Error to the client.

Also, the Tenant ID checked for user access is whatever the user submitted, - **not** the actual Tenant of the Delivery Service. In this way, a `change` or `delete` request can be submitted by a user who does not have permission to modify that Delivery Service. So a user with a Tenant ID of 3, which is a child of Tenant 1 can legally submit a request for any Delivery Service accessible to 1 by simply replacing the `tentantId` property of the `deliveryservice` object with their own Tenant's ID.

## Expected / new behavior:
If a property is required, then omitting it should cause a 400 Bad Request response with an appropriate `alerts` entry describing the problem.

A resource's actual Tenancy should be checked - returning a 403 Forbidden response in the event that access is not allowed - instead of whatever the client asks to be checked.

Make a POST request to `/deliveryservice_requests` containing a syntactically valid DSR, but one wherein the `tenantId` of the `deliveryservice` object is undefined.

Make a POST request to `/deliveryservice_requests` containing a `deliveryservice` inaccessible to the requesting Tenant, but with the `deliveryservice`'s `tentantId` modified to an accessible value.

### LINKED ISSUE #5033: Internal Server Error when assigning a non-existent DSR
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
Trying to assign a Delivery Service Request which does not exist to any user will result in an Internal Server error.

## Expected / new behavior:
If a resource is not found, the API should communicate that with the `404 Not Found` HTTP status code.

Given that no Delivery Service Request exists with the ID 1, attempt to assign Delivery Service Request 1 to any (existing) user. I used my own user (who had the ID 9 in the following example) and it looks like this:

### Request
```http
PUT /api/2.0/deliveryservice_requests/1/assign HTTP/1.1
User-Agent: python-requests/2.22.0
Accept: */*
Connection: keep-alive
Cookie: mojolicious=...
Content-Length: 15

{"assigneeId": 9}
```

(Note also that the documentation specifies a required "id" property of the request body - that is not correct. The DSR to be assigned is fully identified by the request URI.)

### Response
```http
Content-Encoding: gzip
Set-Cookie: mojolicious=...
Vary: Accept-Encoding
Date: Tue, 15 Sep 2020 23:34:43 GMT
Content-Length: 83

{
	"alerts": [
		{
			"level": "error"
		}
	]
}
```
## Anything else:
From the stack trace I found:
```
INFO: routing.go:249: 2020-09-15T23:26:42.644209507Z: PUT /api/2.0/deliveryservice_requests/1/assign? handling (reqid 33624)
DEBUG: plugin.go:190: 2020-09-15T23:26:42.644259514Z: DEBUG plugins.OnRequest calling 0 plugins
ERROR: api.go:209: 2020-09-15T23:26:42.651463829Z: 172.20.0.1:47534 dsr assignment querying existing: sql: no rows in result set
DEBUG: api.go:214: 2020-09-15T23:26:42.651514639Z: Internal Server Error
172.20.0.1:47534 - admin [15/Sep/2020:23:26:42 +0000] "PUT /api/2.0/deliveryservice_requests/1/assign? HTTP/1.1" 500 83 7 "python-requests/2.22.0" 2703160290 NON_IMS
INFO: routing.go:252: 2020-09-15T23:26:42.652031369Z: PUT /api/2.0/deliveryservice_requests/1/assign? handled (reqid 33624) in 7.774269ms
```
which suggests that a `database/sql.ErrNoRows` isn't being handled somewhere.

ocket8888:

- [x] This PR fixes #5033 and fixes #5015 and fixes #5014 and fixes #5012 

This PR overhauls how DSRs work (in APIv3). Instead of a single `deliveryservice` property, DSRs will now have `requested` and `original`.

`requested` is the Delivery Service as it is requested to exist. For `delete`-type DSRs, this property will not exist in responses and will be ignored in requests. In most senses, it's pretty analogous to the old `deliveryservice` property.

`original` is the Delivery Service as it exists/existed before the DSR is/was/would have been applied. This property does not exist on `create`-type DSRs.  For non-`create` DSRs that are "open", this is either given in the request (for `delete` DSRs) or generated by Traffic Ops for responses from the current DS that matches the DSR's `requested`'s XMLID (for `update` DSRs). When a DSR is "closed", the original is saved from the current DS so that even if that DS changes in the future, the actual change enacted by the now-closed DSR is preserved.

Also, a couple more changes: Assignee ID, Author ID, and Last Edited By ID are no longer present on objects returned by the API, the `deliveryservice_requests/{{ID}}/assign` endpoint now allows assigning by username as well as by ID (ID takes precedence when both are provided), and both the `deliveryservice_requests/{{ID}}/assign` and `deliveryservice_requests/{{ID}}/status` now have simple GET handlers that just return the assignee's username (or `null` if there isn't one) and the DSR's status, respectively, to be compliant with the API guidelines.

For a (hopefully) complete explanation of the new model and interface, refer to the PR's docs changes.

Finally, this makes TP use the stable API 2.0 endpoints for DSR-related operations, to avoid shoving TP work into this PR.

- Documentation
- Traffic Ops Client (Go)
- Traffic Ops
- Traffic Portal

Run the associated unit and client/api integration tests. Build and read the docs. Run TP end-to-end tests and verify that the interface works as it used to.

- [x] This PR includes tests


mitchell852:
```
$ db/admin upgrade
Running goose up...
goose: migrating db environment 'development', current version: 2020082700000000, target: 2020083100000000
2020/09/29 16:41:53 FAIL 2020083100000000_dsrs_originals.sql (pq: check constraint "appropriate_requested_and_original_for_change_type" is violated by some row), quitting migration.
```

traffic_ops/testing/api/v1/deliveryservice_requests_test.go
```diff
@@ -148,27 +154,43 @@ func TestDeliveryServiceRequestWorkflow(t *testing.T) {
 			t.Errorf("Expected no entries in DeliveryServiceRequest slice -- got %d", len(dsrs))
 		}
 
+		if len(testData.DeliveryServiceRequests) < 1 {
+			t.Fatalf("Need at least one DSR to test workflow")
```
zrhoffman, line 158: Nit: *Need* can be uncapitalized

zrhoffman (changes requested):

traffic_ops/testing/api/v3/deliveryservice_requests_test.go
```diff
@@ -209,29 +405,43 @@ func TestDeliveryServiceRequestWorkflow(t *testing.T) {
...
 			t.Errorf("expected status=submitted,  got %s", string(dsr.Status))
 		}
 	})
 }
 
-func updateDeliveryServiceRequestStatus(t *testing.T, dsr tc.DeliveryServiceRequest, newstate string) (tc.Alerts, tc.DeliveryServiceRequest) {
-	ID := dsr.ID
+func updateDeliveryServiceRequestStatus(t *testing.T, dsr tc.DeliveryServiceRequestV30, newstate string) (tc.Alerts, tc.DeliveryServiceRequestV30) {
+	if dsr.ID == nil {
+		t.Error("Cannot update DSR with no ID")
```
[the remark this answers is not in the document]
ocket8888: Why not?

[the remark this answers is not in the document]
zrhoffman: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.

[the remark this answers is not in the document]
ocket8888: That's not an error string, it's a testing error, which is printed directly, not returned to the caller to be possibly appended to something else

traffic_ops/testing/api/v1/todb_test.go
```diff
@@ -333,6 +333,7 @@ func Teardown(db *sql.DB) error {
 	DELETE FROM profile;
 	DELETE FROM parameter;
 	DELETE FROM profile_parameter;
+	DELETE FROM topology;
```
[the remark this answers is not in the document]
ocket8888: Because if you run the APIv1 against an "un-clean" database with DSes that have topologies, the tests will to clean up DSes because it violates a foreign key constraint.

lib/go-tc/deliveryservice_requests_test.go
```diff
@@ -105,3 +107,116 @@ func TestRequestStatusJSON(t *testing.T) {
...
+	// Output: Parsed DSRCT: 'create'
+	// Error: invalid Delivery Service Request changeType: 'something invalid'
+}
+
+func ExampleDeliveryServiceRequestV30_String() {
+	var dsr DeliveryServiceRequestV30
+	fmt.Println(dsr.String())
+
+	// Output: DeliveryServiceRequestV30(Assignee=<nil>, AssigneeID=<nil>, Author="", AuthorID=<nil>, ChangeType="", CreatedAt=0001-01-01T00:00:00Z, ID=<nil>, LastEditedBy="", LastEditedByID=<nil>, LastUpdated=0001-01-01T00:00:00Z, Status="")
+}
```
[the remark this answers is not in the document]
ocket8888: I generally like simple positive-case to be in [testable examples](https://blog.golang.org/examples)

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
```diff
@@ -20,451 +22,843 @@ package request
...
-// Update implements the tc.Updater interface.
-//all implementations of Updater should use transactions and return the proper errorType
-//ParsePQUniqueConstraintError is used to determine if a request with conflicting values exists
-//if so, it will return an errorType of DataConflict and the type should be appended to the
-//generic error message returned
-func (req *TODeliveryServiceRequest) Update() (error, error, int) {
-	if req.ID == nil {
-		return errors.New("missing id"), nil, http.StatusBadRequest
+// Warning: this assumes inf isn't nil, and neither is dsr, inf.Tx or inf.User or inf.Tx.Tx.
+func insert(dsr *tc.DeliveryServiceRequestV30, inf *api.APIInfo) (int, error, error) {
```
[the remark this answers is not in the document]
ocket8888: Just a convention. `golint` says that errors should always be the last values returned from a function - which is a convention followed throughout the standard library.

We have a golangci-lint Dockerfile in the repository, but we don't as a project actually enforce any go linting or vetting beyond `go fmt`, so if you want me to change it for consistency, I will.

lib/go-tc/deliveryservice_requests_test.go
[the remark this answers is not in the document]
zrhoffman: Oh interesting, I didn't know you you could run these as tests.

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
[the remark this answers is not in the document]
zrhoffman: IMO `(error, error, int)` should be exempt from that rule because an HTTP status code is effectively an error code. I can see why you would choose the order that you did though, so go with what you think is best.

It would be nice to enforce linting at some point. In order for `golangci-lint` to ignore that particular `golant` rule, we should exclude it in `issues.exclude-rules` in our `.golangci.yml`.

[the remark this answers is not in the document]
ocket8888: > It would be nice to enforce linting at some point

It might, but I doubt we ever will. Go fmt is already enough to annoy people. They'll be very mad if we add enforcements of vetting or linting. It's not a hill I'm willing to die on, is what I'm saying.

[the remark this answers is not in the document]
zrhoffman: Oh, I agree we shouldn't add it all at once. We could disable all the rules that fail so that the master branch passes and only slowly enabling more rules 1 by 1 as we improve the codebase.

mitchell852 (changes requested):
I'd like to give this a test drive before it gets merged. thanks!

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -423,6 +425,34 @@ func GetServerCapabilitiesFromName(name string, tx *sql.Tx) ([]string, error) {
...
+func DSRExistsWithXMLID(xmlid string, tx *sql.Tx) (bool, error) {
+	if tx == nil {
+		return false, errors.New("checking for DSR with nil transaction")
+	}
+
+	var exists bool
+	if err := tx.QueryRow(dsrExistsQuery, xmlid).Scan(&exists); err != nil {
+		return false, err
+	}
+	return exists, nil
```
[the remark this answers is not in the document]
ocket8888: no, that would only return an error. It's not true that `exists` is always true when `err == nil`, so that would not be enough information.

Unless you mean
```go
err := tx.QueryRow(dsrExistsQuery, xmlid).Scan(&exists)
return exists, err
```
?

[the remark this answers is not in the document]
zrhoffman: Whoops, yeah that's what I meant

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
```diff
@@ -20,451 +22,843 @@ package request
...
 }
 
-// GetType is part of the tc.Identifier interface
-func (req TODeliveryServiceRequest) GetType() string {
-	return "deliveryservice_request"
+// getOriginals fetches the Delivery Services identified in 'ids' and sets
+// them as originals on the Delivery Services to which each ID maps in
+// needOriginals. It returns a response code to use if an error occurred, in
+// which case it also returns a user error and a system error.
+func getOriginals(ids []int, tx *sqlx.Tx, needOriginals map[int][]*tc.DeliveryServiceRequestV30) (int, error, error) {
```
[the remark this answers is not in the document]
ocket8888: Discussed below

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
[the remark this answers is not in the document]
ocket8888: Well anyway, I'd like to leave it as-is for now. It passes linting and that pleases me.

Also, though, I think it'd be nice to do something like 
```go
type Errors {
    Code             int
    SystemError error
    UserError     error
}

func (e Errors) Occurred() bool {
    return e.SystemError != nil || e.UserError != nil
}
```

Because I, for one, always have to look up the function I'm calling to figure out which error is for the system and which is for the user, and I think this would make things a lot easier, and solve the linting problems at the same time.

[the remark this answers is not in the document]
zrhoffman: From @rob05c in #3349:

> The `tc.ApiErrorType` is a legacy artifact that needs removed. TO Golang has standardized around the `userErr, sysErr, errCode` idiom, which is more powerful and flexible, and easier to work with (no type casting). The few remaining functions in TOGo using `tc.ApiErrorType` need changed to use the `userErr, sysErr, errCode` idiom, and the symbol removed.

A struct seems like an improvement over the original iota enum, and yeah it would satisfy the `error should be the last type when returning multiple items` golint rule. However, since, as #3349 mentions, `error, error, int` is the current standard,
* `error, error, int` should be used here. Project standard > project-agnostic standard
* We should reach some degree of consensus about whether we want to make that switch. Let's open an issue for it in the meantime, to document it as a possibility.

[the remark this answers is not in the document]
ocket8888: Well nothing not discussed on the mailing list is a project standard.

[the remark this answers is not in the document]
zrhoffman: Ha, fair

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -528,6 +541,36 @@ func NewInfo(r *http.Request, requiredParams []string, intParamNames []string) (
 	}, nil, nil, http.StatusOK
 }
 
+const createChangeLogQuery = `
+INSERT INTO log (
+	level,
+	message,
+	tm_user
+) VALUES (
+	'` + ApiChange + `',
```
zrhoffman, line 550: This should be a prepared statement parameter

zrhoffman (changes requested):
Building the v3 API tests fails

```shell
[user@computer api]$ go test -c ./v3
v3/deliveryservices_test.go:94:31: undefined: tc.DeliveryServiceNullableV30
```

docs/source/api/v3/deliveryservice_requests_id_assign.rst
```diff
@@ -13,255 +13,287 @@
...
-	:ipv6RoutingEnabled:            A boolean that defines the :ref:`ds-ipv6-routing` setting on this :term:`Delivery Service`
-	:lastHeaderRewrite:             A set of :ref:`ds-last-header-rw-rules`
-	:lastUpdated:                   The date and time at which this :term:`Delivery Service` was last updated, in :rfc:`3339` format
-	:logsEnabled:                   A boolean that defines the :ref:`ds-logs-enabled` setting on this :term:`Delivery Service`
-	:longDesc:                      The :ref:`ds-longdesc` of this :term:`Delivery Service`
-	:longDesc1:                     An optional field containing the :ref:`ds-longdesc2` of this :term:`Delivery Service`
-	:longDesc2:                     An optional field containing the :ref:`ds-longdesc3` of this :term:`Delivery Service`
-	:matchList:                     The :term:`Delivery Service`'s :ref:`ds-matchlist`
+It is not required to send both of these; either property is sufficient to determine an :ref:`dsr-assignee`. In most cases, it's easier to use just `assignee`. If both *are* given, then `assigneeId` will take precedence in the event that the two properties do not refer to the same user.
```
zrhoffman, line 91: This should also mention that providing neither property unassigns a DSR from a user.

traffic_ops/traffic_ops_golang/deliveryservice/request/assign.go
```diff
@@ -0,0 +1,243 @@
...
+	if r == nil || tx == nil {
+		return "", http.StatusInternalServerError, nil, errors.New("nil transaction or assignment request")
+	}
+
+	var message string
+	if r.AssigneeID != nil {
+		r.Assignee = new(string)
+		if err := tx.QueryRow(`SELECT username FROM tm_user WHERE id = $1`, *r.AssigneeID).Scan(r.Assignee); err == sql.ErrNoRows {
+			userErr := fmt.Errorf("no such user #%d", *r.AssigneeID)
+			return "", http.StatusConflict, userErr, nil
```
zrhoffman, line 114: Should not use 409 Conflict here for reasons already discussed

traffic_ops/traffic_ops_golang/deliveryservice/request/assign.go
```diff
@@ -0,0 +1,243 @@
...
+		} else if err != nil {
+			sysErr := fmt.Errorf("getting username for assignee ID (#%d): %v", *r.AssigneeID, err)
+			return "", http.StatusInternalServerError, nil, sysErr
+		}
+		message = fmt.Sprintf("Changed assignee of '%s' Delivery Service Request to '%s'", xmlID, *r.Assignee)
+	} else if r.Assignee != nil {
+		r.AssigneeID = new(int)
+		if err := tx.QueryRow(`SELECT id FROM tm_user WHERE username=$1`, *r.Assignee).Scan(r.AssigneeID); err == sql.ErrNoRows {
+			userErr := fmt.Errorf("no such user '%s'", *r.Assignee)
+			return "", http.StatusConflict, userErr, nil
```
zrhoffman, line 124: Should not use 409 Conflict here

traffic_ops/traffic_ops_golang/deliveryservice/request/assign.go
```diff
@@ -0,0 +1,243 @@
...
+	if dsr.ChangeType == tc.DSRChangeTypeUpdate {
+		query := deliveryservice.SelectDeliveryServicesQuery + `WHERE xml_id=:XMLID`
+		originals, userErr, sysErr, errCode := deliveryservice.GetDeliveryServices(query, map[string]interface{}{"XMLID": dsr.XMLID}, inf.Tx)
+		if userErr != nil || sysErr != nil {
+			api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+			return
+		}
+		if len(originals) < 1 {
+			userErr = fmt.Errorf("cannot update non-existent Delivery Service '%s'", dsr.XMLID)
+			api.HandleErr(w, r, tx, http.StatusConflict, userErr, nil)
```
zrhoffman, line 218: Should use a different response status code than 409 Conflict

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
```diff
@@ -20,451 +22,843 @@ package request
...
-		return errors.New("Missing status for DeliveryServiceRequest"), nil, http.StatusBadRequest
+	if dsr.ChangeType == tc.DSRChangeTypeUpdate {
+		query := deliveryservice.SelectDeliveryServicesQuery + `WHERE xml_id=:XMLID`
+		originals, userErr, sysErr, errCode := deliveryservice.GetDeliveryServices(query, map[string]interface{}{"XMLID": dsr.XMLID}, inf.Tx)
+		if userErr != nil || sysErr != nil {
+			return errCode, userErr, sysErr
+		}
+		if len(originals) < 1 {
+			userErr = fmt.Errorf("cannot update non-existent Delivery Service '%s'", dsr.XMLID)
+			return http.StatusConflict, userErr, nil
```
zrhoffman, line 363: The response status code should be something other than 409 Conflict

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
```diff
@@ -20,451 +22,843 @@ package request
...
-	userID := tc.IDNoMod(req.APIInfo().User.ID)
-	req.AuthorID = &userID
-	req.LastEditedByID = &userID
+	if ok, err = dbhelpers.DSRExistsWithXMLID(dsr.XMLID, tx); err != nil {
+		err = fmt.Errorf("checking for existence of DSR with xmlid '%s'", dsr.XMLID)
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, err)
+		return
+	} else if ok {
+		userErr := fmt.Errorf("an open Delivery Service Request for XMLID '%s' already exists", dsr.XMLID)
+		api.HandleErr(w, r, tx, http.StatusConflict, userErr, nil)
```
zrhoffman, line 444: 409 Conflict is not the right response status code to use here

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
```diff
@@ -20,451 +22,843 @@ package request
...
+	if dsr.ChangeType == tc.DSRChangeTypeUpdate {
+		query := deliveryservice.SelectDeliveryServicesQuery + `WHERE xml_id=:XMLID`
+		originals, userErr, sysErr, errCode := deliveryservice.GetDeliveryServices(query, map[string]interface{}{"XMLID": dsr.XMLID}, inf.Tx)
+		if userErr != nil || sysErr != nil {
+			api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+			return
+		}
+		if len(originals) < 1 {
+			userErr = fmt.Errorf("cannot update non-existent Delivery Service '%s'", dsr.XMLID)
+			api.HandleErr(w, r, tx, http.StatusConflict, userErr, nil)
```
zrhoffman, line 718: Please find a response status code to use here that is not 409 Conflict

traffic_ops/traffic_ops_golang/deliveryservice/request/requests.go
```diff
@@ -20,451 +22,843 @@ package request
...
+	if !authorized {
+		api.HandleErr(w, r, tx, http.StatusForbidden, errors.New("not authorized on this tenant"), nil)
+		return
 	}
 
-	if err = req.APIInfo().Tx.QueryRowx(selectDeliveryServiceRequestsQuery()+` WHERE r.id = $1`, *req.ID).StructScan(req); err != nil {
-		return nil, errors.New("dsr status update querying: " + err.Error()), http.StatusInternalServerError
+	if !current.IsOpen() {
+		userErr = fmt.Errorf("cannot change DeliveryServiceRequest in '%s' status", current.Status)
+		api.HandleErr(w, r, tx, http.StatusConflict, userErr, nil)
```
zrhoffman, line 840: Should be changed from response status code 409 Conflict to something else

traffic_ops/traffic_ops_golang/deliveryservice/request/status.go
```diff
@@ -0,0 +1,267 @@
...
+	if dsr.IsOpen() && dsr.ChangeType == tc.DSRChangeTypeUpdate {
+		query := deliveryservice.SelectDeliveryServicesQuery + `WHERE xml_id=:XMLID`
+		originals, userErr, sysErr, errCode := deliveryservice.GetDeliveryServices(query, map[string]interface{}{"XMLID": dsr.XMLID}, inf.Tx)
+		if userErr != nil || sysErr != nil {
+			api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+			return
+		}
+		if len(originals) < 1 {
+			userErr = fmt.Errorf("cannot update non-existent Delivery Service '%s'", dsr.XMLID)
+			api.HandleErr(w, r, tx, http.StatusConflict, userErr, nil)
```
zrhoffman, line 245: The response status code should not be 409 Conflict

traffic_ops/traffic_ops_golang/deliveryservice/request/validate.go
```diff
@@ -20,22 +20,27 @@ package request
...
+	}
+
 	fromStatus := tc.RequestStatusDraft
-	if req.ID != nil && *req.ID > 0 {
-		err := req.APIInfo().Tx.Tx.QueryRow(`SELECT status FROM deliveryservice_request WHERE id=` + strconv.Itoa(*req.ID)).Scan(&fromStatus)
+	if dsr.ID != nil && *dsr.ID > 0 {
+		err := tx.QueryRow(`SELECT status FROM deliveryservice_request WHERE id=$1`, *dsr.ID).Scan(&fromStatus)
 
 		if err != nil {
 			return err
```
zrhoffman, line 46: Should not be returning SQL errors to the user

lib/go-tc/deliveryservice_requests.go
```diff
@@ -547,3 +625,372 @@ func (r RequestStatus) ValidTransition(to RequestStatus) error {
...
+func (dsr *DeliveryServiceRequestV30) Validate(tx *sql.Tx) error {
+	if tx == nil {
+		return errors.New("unknown error")
+	}
+
+	fromStatus := RequestStatusDraft
+	if dsr.ID != nil && *dsr.ID > 0 {
+		if err := tx.QueryRow(`SELECT status FROM deliveryservice_request WHERE id=$1`, *dsr.ID).Scan(&fromStatus); err != nil {
+			return err
+		}
```
zrhoffman, line 893: The user should not be able to see SQL errors.

zrhoffman (changes requested):
I still get an internal server error when leaving out the `"deliveryService"` key for API versions 1.x and 2.0.

traffic_ops/traffic_ops_golang/deliveryservice/request/validate.go
ocket8888: The only option, then, would be to swallow it. Validation pipelines do not allow system vs user-facing errors. FWIW, that behavior existed before I touched this code.

lib/go-tc/deliveryservice_requests.go
ocket8888: Generally I agree. This was done to mimic the behavior of the old API versions, which do/did return SQL errors back to the user. So I think it'd be best to settle the above discussion before getting into this one, since depending on the resolution this might no longer be an issue, and/or they'll be changed in the same way at the same time.

traffic_ops/traffic_ops_golang/deliveryservice/request/validate.go
ocket8888: that's a good idea

lib/go-tc/deliveryservice_requests.go
ocket8888: So should I do what I did above here, or keep the existing error message? Clients shouldn't be looking for specific error messages, but I know they do. Even our own API tests look for exact wording in some places.

ocket8888: nvm, I've got that backwards. I think I'll just not use the `ParseValidator` interface.

zrhoffman: Our API tests looking for exact wording was before the TO API tests had access to the response status codes. No reason the wording should matter now that we've got those.

zrhoffman:
This is all working for me, and all review comments are addressed. Once the merge conflicts are fixed, it should be good to go.

zrhoffman:
API tests fail

```go
2020-10-15T19:29:20.0475519Z === RUN   TestCreateDeletionRequest
2020-10-15T19:29:20.0476243Z     types_test.go:89: ---- CreateTestTypes ----
2020-10-15T19:29:20.1630263Z     cdns_test.go:89: Response:  {[{cdn was created. success}]}
2020-10-15T19:29:20.1657562Z     cdns_test.go:89: Response:  {[{cdn was created. success}]}
2020-10-15T19:29:20.1681491Z     cdns_test.go:89: Response:  {[{cdn was created. success}]}
2020-10-15T19:29:20.1706272Z     cdns_test.go:89: Response:  {[{cdn was created. success}]}
2020-10-15T19:29:20.1732972Z     cdns_test.go:89: Response:  {[{cdn was created. success}]}
2020-10-15T19:29:20.4677684Z     topologies_test.go:50: could not CREATE topology: 400 Bad Request[400] - Error requesting Traffic Ops https://localhost:6443/api/3.0/topologies {"alerts":[{"text":"'empty cachegroups' cachegroups with no servers in them: cachegroup3, parentCachegroup, secondaryCachegroup","level":"error"}]}
2020-10-15T19:29:20.4702457Z     topologies_test.go:186: cannot DELETE topology: 400 Bad Request[400] - Error requesting Traffic Ops https://localhost:6443/api/3.0/topologies?name=my-topology {"alerts":[{"text":"cannot find exactly 1 topology with the query string provided","level":"error"}]}
```

mitchell852 (changes requested):
still getting:

```
2020/10/21 13:59:43 FAIL 2020083100000000_dsrs_originals.sql (pq: check constraint "appropriate_requested_and_original_for_change_type" is violated by some row), quitting migration.
```

mitchell852:
@ocket8888 - sorry, conflicts.

ocket8888:
There've been conflicts for a while. I'm not looking into it until 5.0.0 is released.

mitchell852:
> There've been conflicts for a while. I'm not looking into it until 5.0.0 is released.

Ok, that's right. I forgot you're going to have to bump the API version, right?

ocket8888:
Yes, plus this PR changed the name of `DeliveryServiceNullableV30` to just `DeliveryServiceV30`, which not only won't make sense to do when it's actually going into APIv4, but probably is where nearly all of those conflicts are coming from.

traffic_ops/v3-client/session.go
```diff
@@ -353,6 +353,48 @@ func (to *Session) request(method, path string, body []byte, header http.Header)
...
+// expired not by examining the cookie itself, but by recognizing a
+// 401 Unauthorized response status code from the server. It returns, in order,
+// a pointer to an HTTP response - which, if there is no error, contains a
+// non-nil Body the caller is expected to close -, the remote address of the
+// Traffic Ops server used to service the request, and any error that occurred.
+// Note that any response code greater than or equal to 400 will result in an
+// error - and that means that it cannot be assumed that the response object is
+// nil if there is an error! This is to provide greater visibility to
+// HTTP-level errors while still giving back as much information as possible.
+func (to *Session) requestWithoutClosingBody(method, path string, body []byte, header http.Header) (*http.Response, net.Addr, error) {
```
rawlinp, line 367: I don't really want us to add yet another underlying method to the client that is only used by a small fraction of top-level methods. It makes the client inconsistent and confusing. Instead, is there a way to incorporate these changes into the single shared method I've created [here](https://github.com/apache/trafficcontrol/pull/5334/files#diff-cbf10b941a6c6105c02e8e50de8be60c46c104c53f2f531f1105dffbacd49737R339)? I'm open for suggestions if you'd like to review that approach. I'd just like the top-level method bodies to be concise and use the same underlying method for consistency.

rawlinp (changes requested):

traffic_ops/v3-client/session.go
ocket8888: That'll work fine. It's a much better approach, I just didn't want to go to the trouble of refactoring the whole client in this already large PR. When I finally re-base/do this, I'll use the new method.

ocket8888:
The rebase got very messy, so I'm splitting this up into two parts, the first of which is open for review as #5489 .

PR #5077, opened 2020-09-28
Go fmt action

ocket8888:

This PR adds a new check to GHA for Go Format problems. It outputs annotations containing a diff chunk at each problematic part of the codebase. It also adds a supporting python script for the action (which could potentially be used with other linters, Pylint, ESLint etc.) and a badge for the action to the repository's README.

- CI tests

As it happens, there is currently a file in violation of `go fmt` in the `t3c` tool, so I fixed it with this PR and now you can see the action run and passing on this PR.

To see it fail, make a branch from this PR's branch, make some go-fmt invalidating change and push.

The Python script comes with some doctests, so to test it run `python -m doctest path/to/parse_diffs.py`.

- [x] This PR includes tests
- [x] This PR includes documentation in the form of docstrings


.github/actions/go-fmt/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+FROM golang:1.14.2-alpine
+
+RUN apk add --no-cache git python3
+
+COPY entrypoint.sh /entrypoint.sh
+RUN chmod a+x /entrypoint.sh
```
zrhoffman, line 23: If the file is executable, this layer can be removed

misc/parse_diffs.py
```diff
@@ -0,0 +1,262 @@
...
+		raise ValueError(f"'{contents}' does not represent a file diff in git format")
+
+	match = FILE_HEADER_PATTERN.match(lines[0])
+	if not match or len(match.groups()) != 2:
+		raise ValueError(f"invalid git diff file header: '{lines[0]}'")
+
+	fname = match.groups()[1]
+
+	lines = lines[4:]
+	contents = "\n".join(lines)
```
zrhoffman, line 155: This variable is unused

.github/actions/go-fmt/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+
+GOPATH="$(mktemp -d)"
+SRCDIR="$GOPATH/src/github.com/apache"
+mkdir -p "$SRCDIR"
+ln -s "$PWD" "$SRCDIR/trafficcontrol"
+cd "$SRCDIR/trafficcontrol"
+
+/usr/local/go/bin/go fmt ./...
+DIFF_FILE="$(mktemp)"
+git --no-pager diff >"$DIFF_FILE"
```
zrhoffman, line 29: `--no-pager` is not necessary, it only adds the pager when stdout is a TTY

.github/actions/go-fmt/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [ -s "$DIFF_FILE" ]; then
+	./misc/parse_diffs.py <"$DIFF_FILE";
+	rm "$DIFF_FILE";
+	exit 1;
+fi
+echo "No diff found, exiting" >&2
+exit 0;
```
zrhoffman, line 37: `exit 0` is not necessary, 0 is the default exit code

misc/parse_diffs.py
```diff
@@ -0,0 +1,262 @@
...
+	Annotation(level=Level.error, file='test', line=2)
+	>>> print(ann)
+	::error file=test,line=2::Format error
+	```diff
+	@@ -1,3 +1,3 @@
+	 {
+	+    "test": "quest"
+	-    "foo": "bar"
+	 }
+	```
```
zrhoffman, line 90: Annotations are not markdown-formatted, so <code>```diff</code> is visible in the annotation itself

zrhoffman (changes requested):

.github/actions/go-fmt/Dockerfile
ocket8888: I've had problems with that in the past, but I suppose the only relevant question is: does it matter to the GHA runner? We'll soon find out.

misc/parse_diffs.py
ocket8888: Annotations aren't markdown-formatted _today_, but they could be. I'm sort of hoping they add support, although [the related support community post](https://github.community/t/checks-ability-to-include-markdown-in-line-annotations/14736/3) only has three hearts, so that may be a pipe dream. I can get rid of the formatting if you think it's ugly.

.github/actions/go-fmt/entrypoint.sh
ocket8888: right. That if statement logic was reversed in an earlier revision.

misc/parse_diffs.py
zrhoffman: Fair enough

zrhoffman (approved):
This all works. I'll wait until all of the CI stuff passes to merge.

PR #5080, opened 2020-09-29
Remove dependencies on internal-only Sun packages

ARMmaster17:
Replaces all references to the internal-only Sun Java packages with BouncyCastle. Besides rebasing the cert code on a supported library, these changes are pre-requisites to a future PR that will upgrade Java from 8 to 11 LTS.

- Traffic Router

Documentation is not required as no features/dependencies were changed. All references to Sun.* were changed to use BouncyCastle, which is already included in ATC.

Tests are not included because at the function and class level, nothing should be different, so the existing tests should be sufficient.

- Run tests and confirm success
- Run CiaB and confirm that certs are loaded and used correctly when browsing HTTPS endpoints.

## Additional Information
[mailing list discussion](https://lists.apache.org/thread.html/r083c16c4feb33824e6638c43808d0866c30f63e64d31a2c0fa176001%40%3Cdev.trafficcontrol.apache.org%3E)


smalenfant (approved):
Tested the code using a backport to 3.1 and did not run into any issues.
I also upgraded my openjdk to 11 and it worked as well. Although there is a few warnings due to some old java libraries we are using like "spring". OpenJDK8 worked fine.

zrhoffman:
FYI GitHub does not notify repo watchers when a PR that was initially marked as a draft is ready for review. So, when marking other PRs as ready for review, please also comment that it is, too.

PR #5082, opened 2020-09-29
Fix querying by non-topology-based delivery service ID

### LINKED ISSUE #5079: Getting delivery service servers returns empty array for non-topology based delivery services
## I'm submitting a ...

-  bug report

-  CDN in a Box
-  Traffic Ops

## Current behavior:

For a non topology based DS that has servers assigned, a call to `GET api/3.0/servers?dsId=<your ds ID>` returns an empty response.

## Expected behavior:
We should be able to see the servers associated with the delivery service in the response.

Run TO (and TP) locally.
Create a non-topology based delivery service.
Assign a server to this ds.
Now try to make a call to `GET api/3.0/servers?dsId=<your ds ID>`
Notice how the response has an empty array

zrhoffman:

- [x] This PR fixes #5079
- Makes querying by non-topology-based delivery service ID work again
- Adds an assertion to the test for querying by non-topology-based delivery service ID. Before, it was only asserting that there were no errors, but now it makes sure that servers were returned, as well.

- Traffic Ops

Run the servers API tests, make sure they pass

 - master (e8bb3e90ab)

- [x] This PR includes tests
- [x] Intended behavior is unchanged, no documentation is necessary

### DIFF
--- a/traffic_ops/testing/api/v3/servers_test.go
+++ b/traffic_ops/testing/api/v3/servers_test.go
@@ -316,7 +316,7 @@ func GetTestServersDetails(t *testing.T) {
 }
 
 func GetTestServersQueryParameters(t *testing.T) {
-	dses, _, err := TOSession.GetDeliveryServicesV30WithHdr(nil, url.Values{})
+	dses, _, err := TOSession.GetDeliveryServicesV30WithHdr(nil, url.Values{"xmlId": []string{"ds1"}})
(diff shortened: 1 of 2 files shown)


srijeet0406 (approved):
Changes look good.
Unit and API tests pass.
Tested manually by GETting delivery service servers and made sure it returns the right results for topology based and non-topology based servers.

PR #5084, opened 2020-09-30
Fix remaining `go vet` issues with string(int)

shamrickus:

- Traffic Ops

In the root of the repo, run `go vet ./... 2>&1 | grep rune` and verify there are no issues.
Also make sure that changes correctly address the issue.

- master (435a8b0)

### DIFF
--- a/traffic_ops/traffic_ops_golang/federations/ds.go
+++ b/traffic_ops/traffic_ops_golang/federations/ds.go
@@ -45,7 +45,7 @@ func PostDSes(w http.ResponseWriter, r *http.Request) {
 	fedID := inf.IntParams["id"]
 	fedName, ok, err := getFedNameByID(inf.Tx.Tx, fedID)
 	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("getting federation cname from ID '"+strconv.Itoa(fedID)+"': "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, fmt.Errorf("getting federation cname from ID '%v': %v", fedID, err))
 		return
 	} else if !ok {
 		api.HandleErr(w, r, inf.Tx.Tx, http.StatusNotFound, fmt.Errorf("federation %v not found", fedID), nil)
--- a/traffic_ops/traffic_ops_golang/federations/ds.go
+++ b/traffic_ops/traffic_ops_golang/federations/ds.go
@@ -28,7 +28,6 @@ import (
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
-
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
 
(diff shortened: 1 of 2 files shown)


zrhoffman (approved):
Awesome! Now the Traffic Ops Golang tests compile (and pass) in Go 1.15.

traffic_ops/traffic_ops_golang/federations/ds.go
```diff
@@ -46,7 +45,7 @@ func PostDSes(w http.ResponseWriter, r *http.Request) {
 	fedID := inf.IntParams["id"]
 	fedName, ok, err := getFedNameByID(inf.Tx.Tx, fedID)
 	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("getting federation cname from ID '"+string(fedID)+"': "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("getting federation cname from ID '"+strconv.Itoa(fedID)+"': "+err.Error()))
```
ocket8888, line 48: this should really just use `fmt.Errorf`

rob05c: I try to discourage unnecessary format strings, because they're 2-100x slower (depending on context) than string concatenation and simple conversions. They were a huge scalability issue in Grove and Traffic Monitor.

The performance probably doesn't matter here, but IMO it's good practice to just always avoid them, and then it's never an issue.

ocket8888: it's already being done elsewhere in the file and handler, and I'd expect the bottleneck in the handler to be network ops like reaching out to the database - but I haven't measured.

I guess it doesn't matter, but `string + strconv.Itoa + string + err.Error()` is real ugly.

rob05c: I agree. It makes me sad Go format strings are slow, they're so much cleaner.
To be clear, I'm not an outright -1 here since the performance shouldn't matter, just giving my 2¢.

zrhoffman: > I guess it doesn't matter, but `string + strconv.Itoa + string + err.Error()` is real ugly.

Does that write off all use of the string concatenation operator? if the goal really is to not concatenate, you can also build the string with something like
```go
strings.Join([]string{"string1", strconv.Itoa(number), "string2", err.Error()}, "")
```

ocket8888: tbh I think that's even harder to read. I'm not against string concatenation in general - and I definitely have done things like this line in Traffic Monitor where every little performance gain you can make is important - but I just think ideally you concatenate actual strings. It's the concatenating strings with function calls to convert other types to strings that I think looks nasty.

zrhoffman: I too agree that formatted strings are nicer-looking than concatenation (or joining) and prefer it when performance is not an issue. Benchmarking locally, when making an error 100000000 times,
```go
errors.New("getting federation cname from ID '"+strconv.Itoa(fedID)+"': "+err.Error())
```
takes around 9 seconds, and
```go
fmt.Errorf("getting federation cname from ID %d: %s", fedID, err.Error())
```
takes around 22 seconds. IMO a ~2x slowdown for this single line is acceptable in this case, especially since that factor will not change based on the value of the `fedId` or the message of `err`.

ocket8888: I'd leave it up to shamrickus. Whatever you wanna do will get an approving "review" from me so we can merge this.

shamrickus: I will make the change, I like the readability changes and the error case shouldn't be happening often enough for performance to be a concern.

ocket8888 (approved):

PR #5085, opened 2020-09-30
Fix bug that doesn't let you change the cachegroup of an ASN

### LINKED ISSUE #5081: Cannot change cache group of an ASN
## I'm submitting a ...

-  bug report

-  Traffic Portal

## Current behavior:

Right now, user cannot change the cache group of an ASNs. Even when it is the only one ASN.
## Expected behavior:

TP should let user change the cachegroup of the ASN.

1. Go to TP, create an ASN.
2. Try to change the cache group of the ASN.
3. The error message will display "an asn with the specified number already exists" but this is the only ASN was created.

srijeet0406:

- [x] This PR fixes #5081  

- Traffic Ops

Make sure api and unit tests pass
Also test using the following steps:
Start TO(and TP) locally
Create an asn with a number and cachegroup
Make sure that you cannot create another asn that has the same number OR cachegroup
Try changing the cachegroup of this asn, make sure you can do that
Now, create another asn with a different number and cachegroup
Try changing the cachegroup of this second asn to that of the first asn
Make sure this step errors out
 
- master 

- [x] This PR includes tests

### DIFF
--- a/docs/source/api/v3/asns.rst
+++ b/docs/source/api/v3/asns.rst
@@ -107,7 +107,7 @@ Response Structure
 ========
 Creates a new :abbr:`ASN (Autonomous System Number)`.
 
-.. note:: There cannot be two different ASN object with the same ``asn``.
+.. note:: There cannot be two different ASN object with the same ``asn``. An ASN may only belong to one cachegroup, but a cachegroup can have zero or more ASNs.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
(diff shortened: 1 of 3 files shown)


zrhoffman (changes requested):
IMO the PR title should should be related to what you did to fix the problem, not just the original issue's title.

srijeet0406:
Changed the title

mattjackson220 (approved):
Looks good! Docs build and look good, API tests pass, Unit tests pass, manual testing works as expected!

PR #5087, opened 2020-09-30
 Add default sort functionality to all read()

rimashah25:

- [x] This PR is not related to any Issue. It adds default sort functionality to all generic read API calls 

- Traffic Ops

1. Run v3 API tests
2. Visually check TP UI for all read calls

master (2f5268c)

- [x] This PR includes tests

## Additional Information

Documentation is up to date.


traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -994,3 +994,11 @@ func CreateDeprecationAlerts(alternative *string) tc.Alerts {
 		return tc.CreateAlerts(tc.WarnLevel, "This endpoint is deprecated, and will be removed in the future")
 	}
 }
+
+// Sorting alphabetically for a given readertype (eg: TOCDN, TODeliveryService, TOOrigin etc).
+func Sort(readerType *APIInfo, param string) {
```
rawlinp, line 999: Should we call this `DefaultSort`? Also, the godoc comment should start with the function name.

rimashah25: Sure, I am ok with changing the name of the function.

traffic_ops/v3-client/asn.go
```diff
@@ -68,24 +68,40 @@ func (to *Session) UpdateASNByID(id int, entity tc.ASN) (tc.Alerts, ReqInf, erro
 }
 
 // Returns a list of ASNs
-func (to *Session) GetASNs() ([]tc.ASN, ReqInf, error) {
-	resp, remoteAddr, err := to.request(http.MethodGet, API_ASNS, nil, nil)
+func (to *Session) GetASNsWithHeader(header http.Header) ([]tc.ASN, ReqInf, error) {
```
rawlinp, line 71: I think since we're updating the ASNs client methods here, we should make it take a `*url.Values` as the first parameter, which allows the user to pass the ID or ASN in the request. Then we no longer need `GetASNByIDWithHeader` and `GetASNByASNWithHeader`. I believe that is the pattern we're trying to move towards these days.

rawlinp (changes requested):
Looks good, just have a minor comment on the ASNs client methods. Also, the github checks are reporting a TO API v3 test failure, but it seems unrelated to this PR.

traffic_ops/v3-client/asn.go
rimashah25: Ah ok.. I had no idea, but will update accordingly.

rawlinp (approved):
LGTM, checks passed

PR #5091, opened 2020-10-01
Validate DS Topologies when deleting a server capability

rawlinp:
There must be at least one server in a cachegroup that can satisfy the
required capabilities of the delivery services the cachegroup is
assigned to via topologies. Otherwise, the topologies would no longer be
valid w.r.t. the required capabilities of the delivery services.

P.S. don't be alarmed at the size of the PR, most of it is from the additions to `tc-fixtures.json` and associated whitespace changes.

- [x] This PR  is not related to any issue

- Traffic Ops

Review the added TO API tests, verify the check passes.

- [x] added validation, no docs necessary


traffic_ops/traffic_ops_golang/server/servers_server_capability.go
```diff
@@ -137,46 +138,145 @@ JOIN server s ON sc.server = s.id ` + where + orderBy + pagination +
...
+	defer log.Close(dsRows, "closing dsRows in checkTopologyBasedDSRequiredCapabilities")
+
+	xmlidToTopology := make(map[string]string)
+	xmlidToTenantID := make(map[string]int)
+	xmlidToReqCaps := make(map[string][]string)
+	for dsRows.Next() {
+		xmlID := ""
+		topology := ""
+		tenantID := 0
+		reqCaps := []string{}
```
zrhoffman, line 176: These can all be `var` declarations without assignment

traffic_ops/traffic_ops_golang/server/servers_server_capability.go
```diff
@@ -137,46 +138,145 @@ JOIN server s ON sc.server = s.id ` + where + orderBy + pagination +
...
+	serverRows, err := ssc.APIInfo().Tx.Tx.Query(getServerCapabilitiesOfCachegoupQuery(), ssc.ServerID, ssc.ServerCapability)
+	if err != nil {
+		return nil, fmt.Errorf("querying server capabilitites of server %d's cachegroup: %v", *ssc.ServerID, err), http.StatusInternalServerError
+	}
+	defer log.Close(serverRows, "closing serverRows in checkTopologyBasedDSRequiredCapabilities")
+
+	serverIDToCapabilities := make(map[int]map[string]struct{})
+	for serverRows.Next() {
+		serverID := 0
+		capabilities := []string{}
```
zrhoffman, line 197: These can also be `var` declarations without assignment

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/server/servers_server_capability.go
rawlinp: I kind of prefer the type inference because I can see the zero-values.

traffic_ops/traffic_ops_golang/server/servers_server_capability.go
```diff
@@ -137,46 +138,145 @@ JOIN server s ON sc.server = s.id ` + where + orderBy + pagination +
...
+	var dsStrings []string
+	dsStrings = make([]string, 0, len(xmlidToTopology))
+	if len(serverIDToCapabilities) == 0 {
+		for ds, top := range xmlidToTopology {
+			if _, ok := accessibleTenants[xmlidToTenantID[ds]]; ok {
+				dsStrings = append(dsStrings, "(xml_id = "+ds+", topology = "+top+")")
+			}
+		}
+		return fmt.Errorf("this capability is required by delivery services, but there are no other servers in this server's cachegroup to satisfy them %s", strings.Join(dsStrings, ", ")), nil, http.StatusBadRequest
+	}
```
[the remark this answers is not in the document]
rawlinp: I think this was necessary in an earlier version, but re-reading my code here I believe I agree with you. I'll remove it in the next revision.

traffic_ops/traffic_ops_golang/server/servers_server_capability.go
rob05c: I also like `:=` for that reason, and also to use a unified declaration everywhere, instead of having 2 different ways to declare variables

zrhoffman: I can see the value in that

traffic_ops/testing/api/v3/withobjs_test.go
```diff
@@ -78,38 +80,40 @@ type TCObjFuncs struct {
...
+	ServerServerCapabilitiesForTopologies: {CreateTestServerServerCapabilities, DeleteTestServerServerCapabilitiesForTopologies},
+	Servers:                               {CreateTestServers, DeleteTestServers},
+	ServiceCategories:                     {CreateTestServiceCategories, DeleteTestServiceCategories},
+	Statuses:                              {CreateTestStatuses, DeleteTestStatuses},
+	StaticDNSEntries:                      {CreateTestStaticDNSEntries, DeleteTestStaticDNSEntries},
+	SteeringTargets:                       {SetupSteeringTargets, DeleteTestSteeringTargets},
+	Tenants:                               {CreateTestTenants, DeleteTestTenants},
+	ServerCheckExtensions:                 {CreateTestServerCheckExtensions, DeleteTestServerCheckExtensions},
+	Topologies:                            {CreateTestTopologies, DeleteTestTopologies},
+	TopologyBasedDeliveryServiceRequiredCapabilities: {CreateTestTopologyBasedDeliveryServicesRequiredCapabilities, DeleteTestDeliveryServicesRequiredCapabilities},
```
zrhoffman, line 116: Nit: If this `TCObj` identifier is 40 characters or less (like `TopologyBasedDeliveryServiceRequiredCaps` or `TopologyBasedDeliverySvcReqdCapabilities`), `go fmt` will align the `TCObjFuncs`. Hard to say if that makes it more or less readable overall, I'll leave that call up to you.

rawlinp: While it is deeply regrettable that the columns are no longer fully lined up, I think I prefer reading it unabbreviated (at least when looking at the test file that uses it) since its counterpart (`DeliveryServicesRequiredCapabilities`) is also unabbreviated. When we go full topology, we will be able to remove the `TopologyBased` prefix anyways, and keeping it sticking out like a sore thumb like this is a good reminder to follow through on that.

zrhoffman: Works for me

zrhoffman (approved):
Tests pass, code is formatted correctly, manual testing confirms the validation works.

PR #5092, opened 2020-10-01
Add warning when creating/updating jobs that overlap.

shamrickus:

Currently, an invalidation request with _identical_ regex's that overlap in time is able to be created (Currently ATS will ignore the new entry, but this is subject to change). This PR changes it so that warning messages are returned.

- Traffic Ops
- Traffic Portal

Run TO/TP tests
Verify that creating/updating jobs through TO API with conflicting jobs will return relevant warnings.
Verify that creating jobs through TP with conflicting jobs will show the warnings.

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
@@ -902,7 +902,7 @@ func updateV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, ds *tc.
 		newCDNName, err = getCDNName(*ds.CDNID, tx)
 		newHostName, err = getHostName(ds.Protocol, *ds.Type, *ds.RoutingName, *ds.MatchList, cdnDomain)
 		if err != nil {
-			return nil, http.StatusInternalServerError, nil, errors.New("getting hostname after update: " + err.Error())
+			return nil, http.StatusInternalServerError, nil, errors.New("getting cdnname/hostname after update: " + err.Error())
 		}
 	}
 
--- a/traffic_portal/app/src/common/service/utils/ServerUtils.js
+++ b/traffic_portal/app/src/common/service/utils/ServerUtils.js
@@ -67,24 +67,6 @@ var ServerUtils = function($window, propertiesModel, userModel) {
 		);
 	};
 
-	// Takes a response object from TO and converts all of its alerts to messages acceptable by the messageModel
-	this.convertResponseToMessages = function(response) {
-		let messages = [];
-		if (response.data !== undefined && response.data.alerts !== undefined) {
-			for (let i = 0; i < response.data.alerts.length; ++i) {
-				let text = response.data.alerts[i].text;
-				text = text.charAt(0).toUpperCase() + text.slice(1);
-				messages.push({
-					level: response.data.alerts[i].level,
-					text: text
-				});
-			}
-		} else {
-			console.warn("Malformed response object: " + response);
-		}
-		return messages;
-	};
-
 	/**
 	 * Converts a server's interfaces into legacy IP information. (primarily
 	 * for use in tables)
--- a/traffic_portal/app/src/common/service/utils/ServerUtils.js
+++ b/traffic_portal/app/src/common/service/utils/ServerUtils.js
@@ -67,6 +67,24 @@ var ServerUtils = function($window, propertiesModel, userModel) {
 		);
 	};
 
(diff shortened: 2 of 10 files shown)


traffic_portal/app/src/common/service/utils/ServerUtils.js
```diff
@@ -67,6 +67,24 @@ var ServerUtils = function($window, propertiesModel, userModel) {
 		);
 	};
 
+	// Takes a response object from TO and converts all of its alerts to messages acceptable by the messageModel
+	this.convertResponseToMessages = function(response) {
```
mitchell852, line 71: pretty sure you don't have to do this. messagemodel takes in an array or messages so all you have to do is

```
					messageModel.setMessages(result.alerts, true);
```

or 

```
messageModel.setMessages(result.data.alerts, true);
```

mitchell852: actually, i guess i'm not sure what you're doing. if the server returns 1 alert, TP will show 1. If it returns 5, TP will show 5...

mitchell852: it looks like you return 2 alerts, right? one is a success and one is a warning. an alerts array of 2 _should_ work fine...

shamrickus: You are correct, when I started writing the method I assumed they were differently structured, and I never revisited. It sounds like it would be a better idea to have that method combine all the alerts with the same level instead .

mitchell852: > It sounds like it would be a better idea to have that method combine all the alerts with the same level instead 

Maybe. Or just leave as it is...

mitchell852:
@shamrickus - you'll need to make a change to FormNewJobController.js as well as the response message is currently hardcoded in there.

You may not have been aware that you can create jobs here in TP as well: https://tp.domain.tld/#!/jobs

mitchell852:
Nit: How about getting these messages in line with each other?

image

I.e.

```
Invalidation request duplicate found for http://foo.bar.com/bingo, start:2020-10-07 21:15:47.444 +0000 UTC end:2020-10-09 21:15:47.444 +0000 UTC
```

```
Invalidation request created for http://foo.bar.com/bingo, start:2020-10-07 21:15:47.444 +0000 UTC end:2020-10-09 21:15:47.444 +0000 UTC
```

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -902,7 +902,7 @@ func updateV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, ds *tc.
 		newCDNName, err = getCDNName(*ds.CDNID, tx)
 		newHostName, err = getHostName(ds.Protocol, *ds.Type, *ds.RoutingName, *ds.MatchList, cdnDomain)
 		if err != nil {
-			return nil, http.StatusInternalServerError, nil, errors.New("getting hostname after update: " + err.Error())
+			return nil, http.StatusInternalServerError, nil, errors.New("getting cdnname/hostname after update: " + err.Error())
```
mitchell852, line 905: is this related to this PR?

mitchell852:
weird branch name but no worries just threw me off a bit :)

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
shamrickus: Yep, I added getCDNName but didn't update the message

shamrickus:
> weird branch name but no worries just threw me off a bit :)

Yeah...I definitely got my wires crossed on this one haha

mitchell852:
```
Invalidation request created for /dup, start:2020-10-08 15:34:03.464 +0000 UTC end 2020-10-09 15:34:03.464 +0000 UTC
```
^^ can you make sure the origin is prepended to the regex in the success message?

mitchell852:
I'm also wondering if the changelog message should be different if it is a "duplicate". here's what it looks like now:

```
Created content invalidation job - ID: 3817 DS: xml-id URL: 'http://origin.tld/dup' Params: 'TTL:24h'
```

maybe something like:

```
Created content invalidation job (duplicate) - ID: 3817 DS: xml-id URL: 'http://origin.tld/dup' Params: 'TTL:24h'
```

mitchell852 (approved):
works as expected. lgtm.

PR #5101, opened 2020-10-05
Miscellaneous fixes

ocket8888:

This PR fixes some experimental/miscellaneous Go tools and/or their tests that had compilation issues. They were:

- experimental TR in Go
- test/router's DNSSEC test suite
- the `traffic_vault_util` program under `tools/`
- the semi-experimental `t3c` tool in `traffic_ops_ort/`

The size of this PR is nearly exclusively due to moving vendored packages from `traffic_ops/vendor` into `vendor/` so that `traffic_vault_util` would compile. They might look like big changes, but the files were just renamed without changes and the Go tests/CiaB Action should demonstrate this had no effect on compiling Traffic Ops.

- Traffic Ops ORT (t3c thing)

The checks should be able to do the heavy lifting, but you could also check each tool I listed above and make sure it compiles/passes its own tests/ whatever else. Or don't; they're largely experimental.

- master

- [x] This PR fixes existing tests
- [x] Documentation is unnecessary for bug fixes
- [x] An update to CHANGELOG.md is not necessary for changes to experimental things.


test/router/dnssec/dnssec_suite_test.go
```diff
@@ -20,14 +20,14 @@ package dnssec_test
  */
 
 import (
-	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
+	// _ "github.com/onsi/ginkgo"
+	// _ "github.com/onsi/gomega"
```
zrhoffman, line 24: These can just be removed, right?

test/router/dnssec/dnssec_suite_test.go
```diff
@@ -39,17 +39,32 @@ func init() {
...
+// 	Expect(nameserver).ToNot(Equal("changeit"), "Pass in a ns flag with the hostname of the traffic router")
+// 	Expect(deliveryService).ToNot(Equal("changeit"), "Pass in a ds flag with the dns label for a DNS delivery service")
+// 	log.Println("Nameserver", nameserver)
+// 	log.Println("DeliveryService", deliveryService)
+// })
+
+// func TestDnssec(t *testing.T) {
+// 	RegisterFailHandler(Fail)
+// 	RunSpecs(t, "Dnssec Suite")
+// }
```
zrhoffman, line 55: Can these commented lines be removed?

zrhoffman (changes requested):

test/router/dnssec/dnssec_suite_test.go
ocket8888: Maybe. I'm not totally convinced that my new code is doing exactly the same thing as that old code is meant to do. But I am sure it compiles now. Since I'm not sure what the intention was, and I don't feel the need to make a test for something experimental very clean, I'm inclined to leave it in.

test/router/dnssec/dnssec_suite_test.go
ocket8888: Actually, based on what I said below, I think I should change those back to `.` imports. `.` and `_` don't do the same thing for imports, which is what I thought when I changed that. When you use `.` it imports all symbols from the package into the current namespace, which might be necessary for the code below to compile (although it actually didn't compile). So if I'm leaving in the below code, I'll wanna fix this to just be commented-out and not also changed.

test/router/dnssec/dnssec_suite_test.go
zrhoffman: Sounds good to me

zrhoffman (approved):
- Confirmed traffic_vault_util compiles
- traffic_router_golang tests pass (there are 2 of them)
- The Dockerized tests at `tools/golang`, `traffic_ops/app/bin/tests` (after changing `traffic_ops/ort` to `traffic_ops_ort`), and `infrastructure/cdn-in-a-box/traffic_ops_integration_test` seem to handle the vendored package move just fine.
- `test/router/dnssec` passes after downloading the external `github.com/onsi/ginkgo` dependency

Everything looks like an objective improvement, and existing Traffic Control components still work. LGTM

PR #5107, opened 2020-10-05
Fix Traffic Ops looking for influxdb.conf in the wrong place

### LINKED ISSUE #4292: Traffic Ops Golang isn't looking for influxdb.conf in the right place
## I'm submitting a ...
- bug report

- Documentation
- Traffic Ops

## Current behavior:
Traffic Ops (Go implementation) looks for `influxdb.conf` by default in the same directory as `cdn.conf`.

## Expected / new behavior:
The Perl behavior was to look in `conf/$MOJO_MODE/` in the working directory. Traffic Ops should match the legacy behavior (for now).

Put your influxdb.conf file in the usual location, start Traffic Ops, try to request an endpoint that will connect to an Influx server

ocket8888:

- [x] This PR fixes #4292 

With this PR, Traffic Ops will look for influxdb.conf in `./conf/$MOJO_MODE/` if `influxdb_conf_path` is not given in `cdn.conf` and the `MOJO_MODE` environment variable is set to any non-empty string.

- Traffic Ops

Set `MOJO_MODE`, then start Traffic Ops from somewhere such that there is a valid influxdb.conf at `./conf/$MOJO_MODE/influxdb.conf`, then verify that the therein contained was picked up by Traffic Ops by requesting some endpoint that needs Traffic Stats and watching the logs output.

- 4.1.0
- master

### DIFF
--- a/docs/source/admin/traffic_ops.rst
+++ b/docs/source/admin/traffic_ops.rst
@@ -324,7 +324,7 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 	:workers:           This is used only by the `Legacy Perl Script`_, and sets the number of concurrent HTTP server workers allowed.
 
 :inactivity_timeout: This was used by the `Legacy Perl Script`_ to set timeouts on idle client connections to Traffic Ops - the exact operation (and even units) of this configuration option is unknown. `traffic_ops_golang`_ ignores this field.
-:influx_db_conf_path: An optional field which gives `traffic_ops_golang`_ the absolute or relative path to an `influxdb.conf`_ file. Default if not specified is a file named ``influxdb.conf`` in the same directory as this ``cdn.conf`` file.
+:influxdb_conf_path: An optional field which gives `traffic_ops_golang`_ the absolute or relative path to an `influxdb.conf`_ file. Default if not specified is to first check if the :envvar:`MOJO_MODE` environment variable is set. If it is, then Traffic Ops will look in the current working directory for a subdirectory named ``conf/``, then inside that for a subdirectory with the name that is the value of the :envvar:`MOJO_MODE` variable, and inside that directory for a file named ``influxdb.conf``. If :envvar:`MOJO_MODE` is *not* set, then Traffic Ops will look for a file named ``influxdb.conf`` in the same directory as this ``cdn.conf`` file.
 
 	.. versionadded:: 4.0
 
--- a/traffic_ops/traffic_ops_golang/config/config.go
+++ b/traffic_ops/traffic_ops_golang/config/config.go
@@ -298,13 +298,7 @@ func LoadConfig(cdnConfPath string, dbConfPath string, riakConfPath string, appV
 
 	idbPath := cfg.InfluxDBConfPath
 	if idbPath == "" {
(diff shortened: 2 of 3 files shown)


traffic_ops/traffic_ops_golang/config/config.go
```diff
@@ -298,7 +298,19 @@ func LoadConfig(cdnConfPath string, dbConfPath string, riakConfPath string, appV
 
 	idbPath := cfg.InfluxDBConfPath
 	if idbPath == "" {
-		idbPath = filepath.Join(filepath.Dir(cdnConfPath), "influxdb.conf")
+		var mojoMode string
+		for _, v := range os.Environ() {
```
rawlinp, line 302: Why not `os.Getenv("MOJO_MODE")` instead?

rawlinp (changes requested):

docs/source/admin/traffic_ops.rst
```diff
@@ -324,7 +324,7 @@ This file deals with the configuration parameters of running Traffic Ops itself.
...
 
 :inactivity_timeout: This was used by the `Legacy Perl Script`_ to set timeouts on idle client connections to Traffic Ops - the exact operation (and even units) of this configuration option is unknown. `traffic_ops_golang`_ ignores this field.
-:influx_db_conf_path: An optional field which gives `traffic_ops_golang`_ the absolute or relative path to an `influxdb.conf`_ file. Default if not specified is a file named ``influxdb.conf`` in the same directory as this ``cdn.conf`` file.
+:influxdb_conf_path: An optional field which gives `traffic_ops_golang`_ the absolute or relative path to an `influxdb.conf`_ file. Default if not specified is to first check if the :envvar:`MOJO_MODE` environment variable is set. If it is, then Traffic Ops will look in the current working directory for a subdirectory named ``conf/``, then inside that for a subdirectory with the name that is the value of the :envvar:`MOJO_MODE` variable, and inside that directory for a file named ``influxdb.conf``. If :envvar:`MOJO_MODE` is *not* set, then Traffic Ops will look for a file named ``influxdb.conf`` in the same directory as this ``cdn.conf`` file.
```
rawlinp, line 327: This reminds me, in Perl if `MOJO_MODE` was unset, it would assume `MOJO_MODE=development`.

ocket8888: Should we emulate that? In 4.1.0 TO will look default to that last behavior by checking the `cdn.conf` directory. Using an implicit `MOJO_MODE` when it's not set effectively walks that back - and might make it harder to break away from `MOJO_MODE` in the future, if only slightly.

rawlinp: I would if it were me, since that was the pre-existing behavior, but I don't feel too strongly about changing the default location.

ocket8888: after taking, apparently, nearly a week to think about it, I think I'd like to leave the logic as-is. Because hopefully we're looking at excising the Perl entirely and it feels weird to have remnants like this still hanging around.

rawlinp (approved):
LGTM

PR #5111, opened 2020-10-06
Set root ownership of TM logrotate file

### LINKED ISSUE #2712: Logrotate for Traffic Monitor as wrong user
The logrotate.d config file for traffic monitor is being installed as the traffic_monitor user instead of root which means it's not being used.

rawlinp:
The TM logrotate config file needs to be owned by root in order for
logrotate to rotate the TM logs.

- [x] This PR fixes #2712

- Traffic Monitor

Install the resulting TM rpm, verify that the ownership of `/etc/logrotate.d/traffic_monitor` is `root:root`.

- master
- 4.x
- 3.x

- [x] rpm spec file change, no tests
- [x] no docs necessary


zrhoffman (approved):
* Confirmed that `/etc/logrotate.d/traffic_monitor` is owned by `root:root` when creating a new file
* Confirmed that `/etc/logrotate.d/traffic_monitor` is owned by `root:root` for upgrades when it was previously owned by `traffic_monitor:traffic_monitor`

PR #5121, opened 2020-10-08
When updating a DS, validate its topology w.r.t. required capabilities

### LINKED ISSUE #5099: Validate topology field on PUT /deliveryservices with respect to required capabilities
## I'm submitting a ...

-  Traffic Ops

## Current behavior:

We do not have validation on `PUT /api/3.0/deliveryservices/:id` when the DS has a topology and the DS has required capabilities.

## New behavior:

As a CDN operator, I'd like ensure that the topology I've chosen for my delivery service is valid w.r.t its required capabilities.

Each cachegroup in the assigned Topology should contain at least one server (in the delivery service's CDN) with the delivery service's required capabilities, otherwise the request should fail

1. Add required capabilities to a delivery service
2. Add a topology to that delivery service and PUT

rawlinp:
A DS's topology must contain at least one server in the DS's CDN in each
of the topology's cachegroups that has all the required capabilities.

- [x] This PR fixes #5099 

- Traffic Ops

- Review the added TO API test case, verify the TO API tests pass
- create a new server capability (and do not assign it to any servers), add it as a required capability to a non-topology-based DS, then try to assign a topology to that DS. The update should be prohibited.

- [x] This PR includes tests
- [x] no docs necessary
- [x] The changelog already contains an entry for this validation (added for a different route)

### DIFF
--- a/lib/go-tc/deliveryservices.go
+++ b/lib/go-tc/deliveryservices.go
@@ -532,37 +532,6 @@ func (ds *DeliveryServiceNullableV30) validateTypeFields(tx *sql.Tx) error {
 	return nil
 }
 
-func (ds *DeliveryServiceNullable) Validate(tx *sql.Tx) error {
-	neverOrAlways := validation.NewStringRule(tovalidate.IsOneOfStringICase("NEVER", "ALWAYS"),
-		"must be one of 'NEVER' or 'ALWAYS'")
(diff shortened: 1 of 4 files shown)


zrhoffman (approved):
API tests pass and manual testing confirms the validation works as intended. Nice job!

As you mentioned the changelog already covers this PR.

PR #5133, opened 2020-10-09
TR: set cache to unavailable if not found in TM health data

rawlinp:
In Traffic Router, set a cache to unavailable if it's not found in the TM health data (CrStates). Otherwise, when a cache is set to OFFLINE and removed from TM health data, there is a short window where TR sets it to available until it has finished processing the new snapshot. OFFLINE caches should never be considered available.

- Traffic Router

This is a race condition, which makes it difficult to verify manually, but it can be attempted by:
1. Assign a DS to a single cache
2. Repeatedly request the DS
3. Set the cache to ADMIN_DOWN, snapshot (TR should be returning 5xx shortly)
4. Set the cache to OFFLINE, snapshot

Without this fix, you might be able to get TR to 302 to the offlined cache for a short period until TR has finished processing the new snapshot. With this fix, that should not happen.

- master
- 4.x
- 3.x

- [x] This section of the code currently lacks tests, but I did manually verify the expected behavior before/after
- [x] bug fix, no docs necessary

### DIFF
--- a/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Node.java
+++ b/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Node.java
@@ -200,9 +200,18 @@ public class Node extends DefaultHashable {
 
 	@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.NPathComplexity"})
 	public void setState(final JsonNode state) {
-		final boolean isAvailable = JsonUtils.optBoolean(state, "isAvailable", true);
-		final boolean ipv4Available = JsonUtils.optBoolean(state, "ipv4Available", true);
-		final boolean ipv6Available = JsonUtils.optBoolean(state, "ipv6Available", true);
+		final boolean ipv4Available;
+		final boolean ipv6Available;
+		if (state == null) {
+			LOGGER.warn("got null health state for " + fqdn + ". Setting it to unavailable!");
+			isAvailable = false;
+			ipv4Available = false;
+			ipv6Available = false;
+		} else {
+			isAvailable = JsonUtils.optBoolean(state, "isAvailable", true);
+			ipv4Available = JsonUtils.optBoolean(state, "ipv4Available", true);
+			ipv6Available = JsonUtils.optBoolean(state, "ipv6Available", true);
+		}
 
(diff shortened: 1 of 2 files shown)


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/edge/Node.java
```diff
@@ -200,9 +200,18 @@ public InetAddress getIp6() {
...
+		if (state == null) {
+			LOGGER.warn("got null health state for " + fqdn + ". Setting it to unavailable!");
+			isAvailable = false;
+			ipv4Available = false;
+			ipv6Available = false;
+		} else {
+			isAvailable = JsonUtils.optBoolean(state, "isAvailable", true);
+			ipv4Available = JsonUtils.optBoolean(state, "ipv4Available", true);
+			ipv6Available = JsonUtils.optBoolean(state, "ipv6Available", true);
+		}
```
zrhoffman, line 214: Previously, if `state` was `null`, `JsonUtils.optBoolean()` would return the provided default value. So does this produce a different result than

```java
isAvailable = JsonUtils.optBoolean(state, "isAvailable", false);
ipv4Available = JsonUtils.optBoolean(state, "ipv4Available", false);
ipv6Available = JsonUtils.optBoolean(state, "ipv6Available", false);
```
?

zrhoffman: With the `if` statement we still get truthy defaults in case the TM is <= 4.0.0.

rawlinp: The behavior is slightly different if we were to do that instead, because there is a difference between the entire state being null and the state missing one of the `ipv4Available` or `ipv6Available` keys (which were added in 4.0 I believe). Upon upgrading, those two would need to default to `true` if missing (since they're new); otherwise, TR would mistakenly treat every cache as unavailable.

rawlinp: Yep, that is correct.

zrhoffman (approved):
* Verified the cache mentioned in the *What is the best way to verify this PR?* section is immediately unavailable
* External tests pass
* Integration tests pass
* Unit tests pass.

PR #5134, opened 2020-10-11
Added `--traffic_ops_insecure=<0|1>` optional option to traffic_ops_ort.pl

hbeatty:
Includes changelog.

- Traffic Ops ORT

Install on a CDN where Traffic Ops uses a self-signed cert and run ORT.

It seems the ORT documentation refers to a Python ORT. I didn't know what to do with that.


rob05c (changes requested):
Looks good! Big +1, this is something I've been wanting and just haven't had the time, great feature to have.

My only request is if we can avoid that big command twice everywhere. Would you object to adding a global:
```
my $atstccfg_insecure_arg = "";
if ($traffic_ops_insecure == 1) {
  $atstccfg_insecure_arg = "--traffic-ops-insecure";
}
```
and then changing all the existing command calls to:
```
$response = `$atstccfg_cmd $atstccfg_insecure_arg $atstccfg_timeout_arg ...`;
```
to avoid having 2 copies of that big command everywhere?

rob05c (approved):
Looks Great! Code looks good, tested, works as expected.

No tests, but ORT doesn't have an integration test framework yet and this can't be reasonably unit tested.
No docs, ORT flags aren't documented (outside the app usage printing).

rob05c:
@hbeatty Looks like there's a changelog conflict, I'll merge as soon as you resolve that

PR #5142, opened 2020-10-12
Return last-modified header only in case of IMS requests

srijeet0406:

- Traffic Ops

Make sure that the unit tests and API tests pass.
If you're sending a GET request without an IMS header, you should not see a `LAST-MODIFIED` header in the response.
The `LAST-MODIFIED` header should only be present in case there is an IMS header supplied, and the `use_ims` flag is set to true in TO.

- master

- [x] This PR includes tests


rawlinp:
Part of the problem is probably that when use_ims = false, we're always returning `last-modified` with epoch 0. So that is what's changing -- we're no longer sending a default epoch 0 `last-modified` when use_ims = false.

Maybe once use_ims = true has been thoroughly vetted, we can just enable that by default and always return a valid `last-modified`?

mitchell852 (commented):
LGTM. Also, disabled caching in TP until it utilizes the IMS functionality that TO now provides.

PR #5146, opened 2020-10-13
Service Categories tenancy cross referencing DS tenancy

mattjackson220:

This PR fixes a bug where a service category belonging to a higher tenant could be assigned to a delivery service with a lower tenant

- Traffic Ops
- Traffic Portal

Verify that all unit and API tests pass
Verify that a DS cannot be created with a service category with a higher tenant
Verify that a DS cannot be updated to include a service category with a higher tenant
Verify that a DS can be created and updated if the service category has an equal or lower tenant
Verify that the /service_categories endpoint when given the dsID query parameter filters correctly to only show the appropriate service categories
Verify that TP only shows available service categories

No tests, docs, or changelogs because this is a bug fix

### DIFF
--- a/traffic_ops/traffic_ops_golang/tenant/tenancy.go
+++ b/traffic_ops/traffic_ops_golang/tenant/tenancy.go
@@ -229,13 +229,13 @@ func GetDSTenantIDByIDTx(tx *sql.Tx, id int) (*int, bool, error) {
 }
 
 // GetServiceCategoryTenantIDByNameTx returns the tenant ID, whether the service category exists, and any error.
-func GetServiceCategoryTenantIDByNameTx(tx *sql.Tx, name string) (*int, bool, error) {
-	tenantID := (*int)(nil)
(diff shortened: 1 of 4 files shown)


traffic_ops/traffic_ops_golang/tenant/tenancy.go
```diff
@@ -227,3 +227,32 @@ func getDSTenantIDByIDTx(tx *sql.Tx, id int) (*int, bool, error) {
 	}
 	return tenantID, true, nil
 }
+
+// GetServiceCategoryTenantIDByNameTx returns the tenant ID, whether the service category exists, and any error.
+func GetServiceCategoryTenantIDByNameTx(tx *sql.Tx, name string) (*int, bool, error) {
```
ocket8888, line 232: it's not possible for a service category to have no tenant ID, so you can't ever see `nil, true, nil` returned from this function. Since when the other two values are `false` and/or some non-`nil` error, it seems like there's no reason to allow this to return a nil ID. Why not just have this return `int, bool, error`?

ocket8888 (approved):
All tests passing, tested manually and it works.

PR #5155, opened 2020-10-14
Disallow deleting or moving the last server in a Cache Group used by a topology

### LINKED ISSUE #5154: Traffic Ops should prohibit deleting the last Server in a Cache Group that is assigned to a Topology
## I'm submitting a ...

-  Traffic Ops

## Current behavior:

Servers can be deleted or removed from a Cache Group used by a Topology, even if it means that Cache Group would no longer have any Servers in it.

#5144 already added this validation for `POST` and `PUT /api/3.0/topologies`, so if such a Server was deleted or moved to a different Cache Group, updating that Topology would then fail, as long as the Topology still contains that Cache Group.

## New behavior:

As a CDN operator, I would like to prevent deleting the last server in a Cache Group that is assigned to any Topology. With no Servers in a Cache Group assigned to a Topology, that Topology is no longer valid, so the Cachegroup should be removed from the Topology before the last server can be deleted from it.

For DELETE:
1. Make a Cache Group
2. Add only 1 Server to the Cache Group
3. Make a Topology with that Cache Group
4. Delete the Server (should fail)

For PUT:
1. Make a Cache Group
2. Add only 1 Server to the Cache Group
3. Make a Topology with that Cache Group
4. Move the Server to a different Cache Group (should fail)

zrhoffman:

- [x] This PR fixes #5154 OR is not related to any Issue 
- Keeps you from deleting a Server from a CacheGroup used by a Topology or moving a Server from a CacheGroup used by a Topology if it means that Cache Group would no longer have any servers in it
- Adds the `cachegroupName` query parameter to `GET /api/3.0/servers`

- Documentation
- Traffic Control Client 
- Traffic Ops

1. Run the API tests

2. Test manually:
    For `DELETE /api/3.0/servers/{{id}}`:
    1. Make a Cache Group
    2. Add only 1 Server to the Cache Group
    3. Make a Topology with that Cache Group
    4. Delete the Server (should fail)

    For `PUT /api/3.0/servers/{{id}}`:
    1. Make a Cache Group
    2. Add only 1 Server to the Cache Group
    3. Make a Topology with that Cache Group
    4. Move the Server to a different Cache Group (should fail)

- [x] This PR includes tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added the ability to view Hash ID field (aka xmppID) on Traffic Portals' server summary page
 - Added the ability to delete invalidation requests in Traffic Portal
 - Added the ability to set TLS config provided here: https://golang.org/pkg/crypto/tls/#Config in Traffic Ops
-- Added an indiciator to the Traffic Monitor UI when using a disk backup of Traffic ops.
+- Added support for the `cachegroupName` query parameter for `GET /api/3.0/servers` in Traffic Ops
+- Added an indiciator to the Traffic Monitor UI when using a disk backup of Traffic Ops.
 - Added debugging functionality to CDN-in-a-Box for Traffic Stats.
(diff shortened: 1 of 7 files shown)


mitchell852 (commented):
performed manual tests via TP via the API. worked great. didn't look at code however.

rawlinp (approved):

PR #5168, opened 2020-10-19
Enable includeSystemTests in the API tests GitHub Action

zrhoffman:

This PR
- [x] is not related to any Issue 
- adds an SMTP server to the Traffic Ops API tests GitHub Actions workflow
- adds a Riak server to the Traffic Ops API tests GitHub Action
- Enables the `"includeSystemTests"` in the Traffic Ops Test config used by the Traffic Ops API tests GitHub Action

- Traffic Ops (only API tests fixtures changed)
- CI tests

- Check the result of the *Traffic Ops Go client/API integration tests* GitHub Action
- Look at `.github/actions/to-integration-tests/README.md` and check if it still makes sense

- [x] This PR includes tests
- [x] Affects only our CI tests, no changelog entry necessary


zrhoffman:
Rebased onto master

.github/actions/to-integration-tests/entrypoint.sh
```diff
@@ -26,6 +26,46 @@ download_go() {
...
+		# Do not wait for CDN in a Box to generate SSL keys
+		sed -i '0,/^update-ca-certificates/d' /etc/riak/prestart.d/00-config.sh;
+
+		# Do not try to source to-access.sh
+		sed -i '/to-access\.sh/d' /etc/riak/{prestart.d,poststart.d}/*
+	BASH_LINES
+
+	DOCKER_BUILDKIT=1 docker build "$ciab_dir" -f "${ciab_dir}/traffic_vault/Dockerfile" -t "$trafficvault";
+	network="$(docker network ls --quiet --filter name=github_network)";
+	if [[ "$INPUT_VERSION" -lt 3 ]]; then
```
shamrickus, line 50: Looking at the logs for the V1/V2 tests (specifically the TO error logs):
```
ERROR: riak_services.go:374: 2020-10-19T17:24:58.753104875Z: GetPooledCluster failed, returning nil cluster
ERROR: deleteoldcerts.go:112: 2020-10-19T17:24:58.753126176Z: deleting old DS certificates: getting riak ds keys: with cluster error: getting riak pooled cluster: GetPooledCluster unable to return cluster
ERROR: riak_services.go:374: 2020-10-19T17:24:58.894709833Z: GetPooledCluster failed, returning nil cluster
```
Perhaps it would be good to have it enabled for old versions too? It's not causing any test failures, but those might be hiding legitimate regression bugs.

zrhoffman: Having TV enabled the whole time would add 4-5 minutes to the workflow because
* Right now, it builds Traffic Vault while running the API tests
* It starts Traffic Vault in the background while running the API tests andit sort works out that it's up and running before it's used by `TestDeliveryServices` in v3

While log messages are useful, if there really is a bug related to that, ideally we would have an API for it instead of relying on scanning the logs.

One thing that makes scanning the logs trickier is that the Traffic Ops error log is only shown after the tests have finished running, so seeing log messages that correspond with certain points in the tests is not straightforward. I have changed 2 things in 9c915d9eb5 to make seeing Traffic Ops errors and warnings easier:
* Now the logs show up in real time instead of after tests have run
* Traffic Ops warnings are shown with a red background and warnings are shown with a yellow, so they are easy to spot (not sure why, but I don't see any TO warnings anyway).

Will that work?

shamrickus: That's fine, I don't think these particular errors are anything significant, just something that I noticed

shamrickus:
ciab-build timed out

zrhoffman:
Passed on re-run. It failed before due to #5182 which is not caused by this PR.

.github/actions/to-integration-tests/dist/to-integration-test.js
```diff
@@ -0,0 +1,31 @@
+/*
```
shamrickus, line 1: Did you mean to commit this file?

zrhoffman: Yes. TypeScript actions typically commit the transpiled JS file, the action itself is not responsible for transpiling.

shamrickus: This file is named `to-integration-test.js` there is another in dist called `to-integration-tests.js` which corresponds to the `to-integration-tests.ts` file.

zrhoffman: oh

zrhoffman: Removed in 9bee219470

shamrickus (approved):
LGTM
* TO API tests still pass locally
* GHA properly transpiles to js
* TV is started and system tests are ran

shamrickus:
Could you add `.github/actions/to-integration-tests/node_modules` to the `.gitignore`?

zrhoffman:
> Could you add `.github/actions/to-integration-tests/node_modules` to the `.gitignore`?

Gitignored in 8d06b84b28.

zrhoffman:
Changed tabs to spaces in c10adf023b.

Since `"strict"` is already set in `tsconfig.json`, `"strictNullChecks"` is [already set as well](https://github.com/microsoft/TypeScript-Website/commit/66f17ad4d0fa6b2b9fee791b0b7a729710aeb907#diff-c61f3a65b5d5f3b9d067e04829015516162b7aa9f9ececb29e3c138f0dd873caR86
).

> --strict`                                     | `boolean` | `false`                        | Enable all strict type checking options. <br/>Enabling `--strict` enables `--noImplicitAny`, `--noImplicitThis`, `--alwaysStrict`, `--strictBindCallApply`, `--strictNullChecks`, `--strictFunctionTypes` and `--strictPropertyInitialization`.

zrhoffman:
The API tests failure is legitimate. See #5203

zrhoffman:
Rebased after #5203 was closed.

ocket8888 (approved):
Tests now pass

Issue #1124, opened 2017-09-05
Traffic Ops "maintenance mode"


dangogh:
some situations require that traffic ops db not be modified for a period of time,  e.g. when upgrading to a newer version.  In-process items should be finished up,  but no new actions started.

Create a "maintenance mode" that can be set by an admin.   This should be shown somehow on the UI and any modification actions disallowed until changed back to the "normal mode".

mitchell852:
without the protection this provides, you could corrupt TO which would have a very negative impact on traffic control.

rawlinp:
Here's an idea:
- some kind of global "maintenance" flag in the DB
- some way to grant "maintenance" privileges to TO users

If a user does not have maintenance privileges, and TO is in maintenance mode, all their requests get 503'd.

mitchell852:
> Here's an idea:
> 
> * some kind of global "maintenance" flag in the DB
> * some way to grant "maintenance" privileges to TO users
> 
> If a user does not have maintenance privileges, and TO is in maintenance mode, all their requests get 503'd.

would it make sense to use a couple new parameters that are part of the global profile? i.e. 

1. name=maintenance, config_file=global, value=true|false and 
2. name=maintenance.users, config_file=global, value=comma-separated list of usernames that can interact w/ TO when maintenance=true

or is that taking parameters too far?

ocket8888:
I think that's "parameter abuse". Especially since `value =true|false` cannot be enforced.

rawlinp:
Yeah, I'd lean more towards something like a new `maintenance_mode_users` table. If there are any rows, TO is in maintenance mode, and each row contains a user ID that is allowed to perform operations. But any admin can take us in or out of maintenance mode.

mitchell852:
wondering if this is needed anymore since "cdn locks" was introduced. basically, you can apply an exclusive lock (aka hard lock) on one or all of the cdns which will prevent anyone except for you changing the state of the cdn.

rawlinp:
I think cdn locks pretty much cover this use case.

zrhoffman:
> I think cdn locks pretty much cover this use case.

Agreed. If anyone feels it doesn't, feel free to comment so we can reopen.

PR #5179, opened 2020-10-21
Adds the ability to create a CDN-level notification via TP/TO API 

mitchell852:

This feature provides users w/ operations or admin role the ability to add a notification to a cdn which does 2 things:

1. creates a notification in the database for the CDN
2. displays a persistent notification for each CDN in TP as such:

image

3. allows anyone w/ operations or admin role to delete the CDN notification

- [x] This PR is related to #1124 

- Traffic Ops
- Traffic Portal

- Launch TP
- Click on a cdn notification icon in the header of TP to create a notification
- Navigate to different views of TP to ensure that the notification is persistent
- Launch a 2nd instance of TP
- In the 2nd instance, click the same cdn notification icon in the header of TP to delete the notification
- Navigate back to TP instance 1 and wait 30 seconds, notification should disappear.


docs/source/api/v4/cdns_id_notification.rst
```diff
@@ -0,0 +1,123 @@
...
+.. distributed under the License is distributed on an "AS IS" BASIS,
+.. WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+.. See the License for the specific language governing permissions and
+.. limitations under the License.
+..
+
+.. _to-api-cdns-id-notification:
+
+****************************
+``cdns/{{ID}}/notification``
```
ocket8888, line 19: Can you not GET the notification?

From [the API guidelines](https://traffic-control-cdn.readthedocs.io/en/latest/development/api_guidelines.html#get):

> _"All endpoints dealing with the manipulation or fetching representations of “Traffic Control Objects” MUST support this method."_

docs/source/api/v4/cdns_id_notification.rst
```diff
@@ -0,0 +1,123 @@
...
+	Access-Control-Allow-Origin: *
+	Content-Type: application/json
+	Set-Cookie: mojolicious=...; Path=/; Expires=Mon, 18 Nov 2019 17:40:54 GMT; Max-Age=3600; HttpOnly
+	Whole-Content-Sha512: rBpFfrrP+9IFkwsRloEM+v+I8MuBZDXqFu+WUTGtRGypnAn2gHooPoNQRyVvJGjyIQrLXLvqjEtve+lH2Tj4uw==
+	X-Server-Name: traffic_ops_golang/
+	Date: Wed, 14 Nov 2018 21:02:07 GMT
+	Content-Length: 41
+
+	{ "response": {
+		"cdnId": 2,
```
ocket8888, line 77: I have to know the CDN ID to call this endpoint, I'm not sure I need this in the response

docs/source/api/v4/cdns_id_notification.rst
ocket8888: Also, instead of ID did you consider using the Name instead? It's just much easier for humans to use

mitchell852: yeah, you're right. name is probably a better option. i was just following the pattern but it looks like there is a pattern for both methods :)

mitchell852: > Can you not GET the notification?

no, but i can add it

docs/source/api/v4/cdns_id_notification.rst
mitchell852: gone

docs/source/api/v4/cdns_id_notification.rst
mitchell852: > Can you not GET the notification?

a notification is attached to the cdn so you "get" it  by calling GET /cdns. think that's ok? i hate to add a route that has no value add.

ocket8888: Well by the same token can you not just create a notification by updating the CDN with a PUT?

traffic_portal/app/src/common/modules/form/cdn/form.cdn.tpl.html
```diff
@@ -41,6 +41,9 @@
                     <span class="caret"></span>
                 </button>
                 <ul class="dropdown-menu-right dropdown-menu" uib-dropdown-menu>
+                    <li ng-if="!cdn.notificationCreatedBy" role="menuitem"><a ng-click="toggleNotification(cdn)">Create CDN Notification</a></li>
+                    <li ng-if="cdn.notificationCreatedBy" role="menuitem"><a ng-click="toggleNotification(cdn)">Delete CDN Notification</a></li>
```
ocket8888, line 45: Anchor tags should be links - these aren't performing navigation, so they ought to be `button[type="button"]`s

traffic_portal/app/src/common/modules/notifications/notifications.tpl.html
```diff
@@ -0,0 +1,24 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<div id="notificationsContainer">
+    <div ng-show="cdn.notificationCreatedBy" class="alert alert-info" ng-repeat="cdn in cdns">
+        <div ng-bind-html="cdn.name + ': ' + cdn.notification + ' (' + cdn.notificationCreatedBy + ')' | linky:'_blank'"></div>
```
ocket8888, line 22: What is the `linky` pipe doing here?

ocket8888 (commented):
It looks like currently this change is adding two fields to CDNs in all API versions - is that correct?

traffic_portal/app/src/common/modules/notifications/notifications.tpl.html
mitchell852: converts urls to links. i.e. you can have a notification like this:

```
Please go to http://www.website.com for more information
```

and it will turn that into a link.

mitchell852:
> It looks like currently this change is adding two fields to CDNs in all API versions - is that correct?

yes, that is correct.

traffic_portal/app/src/common/modules/notifications/notifications.tpl.html
ocket8888: Is that safe? I think using invalidated user input to provide functionality like this could technically be an XSS vulnerability (is "XSS" appropriate here? Not sure)

ocket8888:
Doesn't that technically require a minor version bump for each?

mitchell852:
> Doesn't that technically require a minor version bump for each?

well i don't want that. i'll look into it.

traffic_portal/app/src/common/modules/notifications/notifications.tpl.html
mitchell852: not sure. i could remove it i guess but seemed like nice functionality. i'll see if the library does any sanitation.

traffic_portal/app/src/common/modules/form/cdn/form.cdn.tpl.html
mitchell852: removing. these were supposed to be removed but i forgot about them.

traffic_portal/app/src/common/modules/notifications/notifications.tpl.html
mitchell852: https://docs.angularjs.org/api/ngSanitize/filter/linky

rawlinp:
If `GET /cdns` returns new fields, but you can't use `PUT /cdns` or `POST /cdns` to update/create those fields, I think that might be fine to do without a minor version bump.

IMO minor version bumps are all about "round-trip safety". I.e. can an "old" client do a `GET`, followed by a `PUT` using that response and not accidentally delete any new data fields. If the new data fields are essentially read-only, I don't see a real need for bumping the minor version.

jhg03a:
Unless the lack of those fields in POST/PUT cause TO to reject the request.

rawlinp:
Right, if that were the case, it would be a _major_ version bump (adding new required fields).

rob05c:
What happens if a new client does a `GET /cdns` from an old TO, silently saves the empty string, and then does a `POST /notifications` with that data (possibly modified) later after TO is upgraded, or maybe to different HA TO?

The client had no way to know the TO was a different version. It's impossible for the client to even know the issue is going to happen, much less prevent it. Because the TO's have the same version, even though they're actually different.

This is one of the fundamental error cases that minor versioning prevents. Giving the same data different GET and POST paths doesn't somehow make that issue not exist. It just confuses it.

I disagree that versioning should only ensure "round trip safety." Versioning prevents clients and servers from breaking for a number of reasons, that's only one of them.

Giving the same data different GET and POST paths doesn't somehow make those errors not happen.

mitchell852:
it's alright. i'm going to rethink this and probably just create a GET/POST/PUT/DELETE 4.0/cdn_notifications endpoint

rawlinp:
> What happens if a new client does a GET /cdns from an old TO, silently saves the empty string, and then does a POST /notifications with that data (possibly modified) later after TO is upgraded, or maybe to different HA TO?

I'm sorry, but that is an extremely contrived example of something a user would most likely never have a valid reason to even do. You can't stop clients from misusing an API like that, even with 100% semantic versioning.

rob05c:
It is not contrived, and it is not misuse by the client. 

Caching is something we have frequently done in the past. ORT has done it, the Monitor has done it. We just implemented IMS and IUS in TO (which does _not_ solve this problem either). TO can only serve so many requests per second, and even if it can network and data is slow.

It's perfectly reasonable for a client to get data, and expect to be able to modify that data and push it back. It's also reasonable to be able to upgrade TO in-place, without breaking ongoing CDN traffic and operational scripts.

Minor versioning in many cases, including this one, makes accidental errors impossible. Yes, the client could willfully ignore the error, like deserializing an older version to a newer struct and then posting that back. That would be misuse. And you're right, we can't prevent intentional bad behavior. But requesting data and posting it back is not misuse, and versioning prevents accidental and unknowable errors. It's not possible to prevent willful bad behavior, but it is possible to prevent accidental mistakes, and make it possible to detect and prevent them.

docs/source/api/v4/cdn_notifications.rst
```diff
@@ -0,0 +1,201 @@
...
+			"level": "success"
+		}
+	],
+	"response": {
+		"cdn": "cdn1",
+		"lastUpdated": "2019-12-02 21:49:08+00",
+		"notification": "the content of the notification",
+		"user": "username123",
+	}
+}
```
ocket8888, line 150: This closing brace needs to be indented. Or probably just at the end of the previous line.

docs/source/api/v4/cdn_notifications.rst
```diff
@@ -0,0 +1,201 @@
...
+:Roles Required: "admin" or "operations"
+:Response Type: ``undefined``
+
+Request Structure
+-----------------
+
+.. code-block:: http
+	:caption: Request Example
+
+	DELETE /api/4.0/cdn_notifications?cdn=cdn1 HTTP/1.1
```
ocket8888, line 167: Looks like an undocumented, required query parameter here.

docs/source/api/v4/cdn_notifications.rst
```diff
@@ -0,0 +1,201 @@
...
+
+.. note:: Currently only one notification per CDN is supported.
+
+:Auth. Required:	Yes
+:Roles Required:	"admin" or "operations"
+:Response Type:		Object
+
+Request Structure
+-----------------
+:cdn:			The name of the CDN to which the notification shall belong
```
ocket8888, line 97: I don't think it's actually a problem, but there's a lot of weird whitespace between these field properties and their values. Looks like a tab character or three.

lib/go-tc/cdn_notification.go
```diff
@@ -0,0 +1,58 @@
...
+// CDNNotificationRequest encodes the request data for the POST
+// cdn_notifications endpoint.
+type CDNNotificationRequest struct {
+	CDN          string `json:"cdn"`
+	Notification string `json:"notification"`
+}
+
+// CDNNotification is a notification created for a specific CDN
+type CDNNotification struct {
+	CDN          *string    `json:"cdn" db:"cdn"`
```
ocket8888, line 46: Is this allowed to be `null`?

lib/go-tc/cdn_notification.go
```diff
@@ -0,0 +1,58 @@
...
+	LastUpdated  *TimeNoMod `json:"lastUpdated" db:"last_updated"`
```
ocket8888, line 47: Can we stop using `TimeNoMod`? If you just use `time.Time` it'll automatically use RFC3339, which is a lot easier to deal with.

lib/go-tc/cdn_notification.go
```diff
@@ -0,0 +1,58 @@
...
+	User         *string    `json:"user" db:"user"`
+}
+
+// Validate validates the CDNNotification request is valid for creation.
+func (n *CDNNotification) Validate(tx *sql.Tx) error {
+	errs := validation.Errors{
+		"cdn": validation.Validate(n.CDN, validation.Required),
+	}
+	return util.JoinErrs(tovalidate.ToErrors(errs))
+}
```
ocket8888, line 58: With the recent DS CacheURL removal, we've sort of started putting these into Traffic Ops itself, since it's weird to expose this thing that requires a transaction into the TO database to work.

traffic_ops/traffic_ops_golang/cdnnotification/cdnnotifications.go
```diff
@@ -0,0 +1,190 @@
...
+import (
+	"database/sql"
+	"errors"
+	"fmt"
+	"github.com/apache/trafficcontrol/lib/go-util"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
+	"net/http"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
```
ocket8888, line 31: project-internal imports should be grouped beneath standard library imports

traffic_ops/traffic_ops_golang/cdnnotification/cdnnotifications.go
```diff
@@ -0,0 +1,190 @@
...
+func Create(w http.ResponseWriter, r *http.Request) {
+	inf, sysErr, userErr, errCode := api.NewInfo(r, nil, nil)
+	tx := inf.Tx.Tx
+	if sysErr != nil || userErr != nil {
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+	defer inf.Close()
+
+	var n tc.CDNNotification
```
ocket8888, line 122: Shouldn't this be a `tc.CDNNotificationRequest`? I mean, I think this works since you're using basic `QueryRow`, but I just assume this is the reason you made that struct.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/cdnnotification"
```
ocket8888, line 26: project-local imports should be grouped beneath standard library imports

traffic_ops/v4-client/cdn_notifications.go
```diff
@@ -0,0 +1,49 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package client
+
+import (
+	"fmt"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"net/http"
```
ocket8888, line 21: project-local imports should be grouped beneath standard library imports

traffic_ops/v4-client/cdn_notifications.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+const (
+	API_CDN_NOTIFICATIONS = apiBase + "/cdn_notifications"
```
ocket8888, line 25: Can you add a GoDoc to this exported symbol? Or just un-export it, I don't think anything needs it. Although I know (almost) all other client files use this casing for their request paths.

traffic_ops/v4-client/cdn_notifications.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// Returns a list of CDN Notifications.
```
ocket8888, line 28: GoDoc should begin with the name of the documented symbol.

traffic_portal/app/src/common/api/CDNService.js
```diff
@@ -222,7 +222,42 @@ var CDNService = function($http, locationUtils, messageModel, ENV) {
...
+            },
+            function(err) {
+                messageModel.setMessages(err.data.alerts, false);
+                throw err;
+            }
+        );
+    };
+
+    this.deleteNotification = function(cdn) {
+        return $http.delete(ENV.api['root'] + 'cdn_notifications?cdn=' + cdn.name).then(
```
ocket8888, line 251: You should either pass the `cdn` parameter in the `$http.delete`'s query string parameters optional argument, or use `encodeURIComponent` to ensure proper URI component encoding.

traffic_portal/app/src/common/modules/form/cdn/FormCDNController.js
```diff
@@ -17,7 +17,7 @@
  * under the License.
  */
 
-var FormCDNController = function(cdn, $scope, $location, $uibModal, formUtils, stringUtils, locationUtils, cdnService) {
+var FormCDNController = function(cdn, $scope, $location, $state, $uibModal, formUtils, stringUtils, locationUtils, cdnService, messageModel) {
```
ocket8888, line 20: The only changes I see to this file are adding these  dependency injections and removing a blank line. Does this controller use those injections already, and it was a bug that they weren't being injected?

traffic_portal/app/src/common/modules/header/HeaderController.js
```diff
@@ -64,6 +78,68 @@ var HeaderController = function($rootScope, $scope, $state, $uibModal, $location
...
+        });
+        modalInstance.result.then(function(notification) {
+            cdnService.createNotification(cdn, notification).
+                then(
+                    function() {
+                        $rootScope.$broadcast('headerController::notificationCreated');
+                    }
+                );
+        }, function () {
+            // do nothing
```
ocket8888, line 112: Do you wanna just `console.error` the error?

traffic_portal/app/src/common/modules/header/HeaderController.js
```diff
@@ -64,6 +78,68 @@ var HeaderController = function($rootScope, $scope, $state, $uibModal, $location
...
+        });
+        modalInstance.result.then(function() {
+            cdnService.deleteNotification(cdn).
+                then(
+                    function() {
+                        $rootScope.$broadcast('headerController::notificationDeleted');
+                    }
+                );
+        }, function () {
+            // do nothing
```
ocket8888, line 139: Same as above RE: `console.error`

traffic_portal/app/src/common/modules/header/header.tpl.html
```diff
@@ -40,7 +40,7 @@
             <li role="presentation" class="dropdown" ng-if="hasCapability('change-logs-read')">
                 <div class="btn-group" title="Change Logs" uib-dropdown is-open="alerts.isopen">
                     <button id="alertsButton" type="button" class="btn btn-link" ng-click="getChangeLogs()" uib-dropdown-toggle>
-                        <i class="fa fa-comment-o"></i>
+                        <i class="fa" ng-class="{ 'fa-comment': newLogCount() > 0, 'fa-comment-o': newLogCount() == 0 }"></i>
```
ocket8888, line 43: `ng-class` supports expresions that evaluate to strings and arrays of strings as well as object property mappings. What I'm getting at is I think it's simpler - and calls `newLogCount` less times - to use `ng-class="{ newLogCount() > 0 ? 'fa-comment' : 'fa-comment-o'}"`

traffic_portal/app/src/common/api/CDNService.js
mitchell852: oh right. i meant to fix that.

traffic_portal/app/src/common/modules/form/cdn/FormCDNController.js
mitchell852: i'll check

traffic_portal/app/src/common/modules/header/HeaderController.js
mitchell852: sure

traffic_portal/app/src/common/modules/header/header.tpl.html
mitchell852: yep, that's better

lib/go-tc/cdn_notification.go
mitchell852: sure. i really don't know the difference tbh. i was just following a pattern that someone introduced.

lib/go-tc/cdn_notification.go
mitchell852: ah i see

traffic_ops/v4-client/cdn_notifications.go
mitchell852: so api_cdn_notifications instead? sorry, i'm used to constants being uppercase, exported (public) or not (private).

lib/go-tc/cdn_notification.go
mitchell852: no, can't be null. so just remove the pointer?

traffic_portal/app/src/common/modules/header/HeaderController.js
mitchell852: oh, this is when you cancel the dialog window. there is no error to throw or log. i mean i could do this:

```
console.log('dialog dismissed')
```

and i considered that at one time but it pollutes the dev console. if you feel strongly about not doing `do nothing`, maybe we can create an issue to fix this globally?

traffic_portal/app/src/common/modules/header/header.tpl.html
mitchell852: unfortunately this syntax is not welcome so i'll just leave it as is

```
<i class="fa" ng-class="{ newLogCount() &gt 0 ? 'fa-comment' : 'fa-comment-o'}"></i>
```

lib/go-tc/cdn_notification.go
ocket8888: The intention was for you to not be able to modify the `last_updated` time manually with a `POST` or `PUT`, but that's only a problem if you read into the entire struct and then use `sqlx.QueryRowx` to dump a struct's properties into a values row. But since you're using plain `sql.QueryRow` and have `CDNNotificationRequest` it's not going to cause any issues.

I believe the proprietary format is meant to mirror Perl's returned date/time format, which wasn't RFC3339 for whatever reason.

traffic_portal/app/src/common/modules/header/HeaderController.js
ocket8888: Oh, I thought this was related to the `cdnService.createNotification` call. But it is not. Nvm.

traffic_portal/app/src/common/modules/header/header.tpl.html
ocket8888: Oh, that's because it doesn't need the curly braces. It's not an object anymore. My bad.

traffic_ops/v4-client/cdn_notifications.go
ocket8888: Yeah, me too, that's why I casefolded them all to that. But they *will* be exported if we keep doing this, whether that's what we want or not. I think snake_case is ugly, so I'd do `apiCDNNotification`, but `go fmt` won't care either way, so neither do I.

traffic_ops/traffic_ops_golang/cdnnotification/cdnnotifications.go
mitchell852: done

traffic_ops/v4-client/cdn_notifications.go
rawlinp: FYI the client changes here will need to be updated once https://github.com/apache/trafficcontrol/pull/5552 is merged. That PR replaces all the snake casing w/ more "Go idiomatic" casing -- e.g. `APICDNNotifications`.

lib/go-tc/cdn_notification.go
```diff
@@ -0,0 +1,59 @@
...
+
+// CDNNotificationRequest encodes the request data for the POST
+// cdn_notifications endpoint.
+type CDNNotificationRequest struct {
+	CDN          string `json:"cdn"`
+	Notification string `json:"notification"`
+}
+
+// CDNNotification is a notification created for a specific CDN
```
ocket8888, line 45: GoDoc should end with a period

traffic_ops/traffic_ops_golang/cdnnotification/cdnnotifications.go
```diff
@@ -0,0 +1,191 @@
...
+		userErr, sysErr, errCode = api.ParseDBError(err)
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
+
+	changeLogMsg := fmt.Sprintf("CDN_NOTIFICATION: %s, CDN: %s, ACTION: Created", *resp.Notification, resp.CDN)
+	api.CreateChangeLogRawTx(api.ApiChange, changeLogMsg, inf.User, tx)
+
+	alertMsg := fmt.Sprintf("CDN notification created [ User = %s ] for CDN: %s", resp.User, resp.CDN)
+	api.WriteRespAlertObj(w, r, tc.SuccessLevel, alertMsg, resp)
```
ocket8888, line 140: creation should use a `201 Created` response code and ideally set a `Location` header with to a request path where the new object representation can be fetched.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -355,6 +356,11 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		{api.Version{4, 0}, http.MethodPost, `coordinates/?$`, api.CreateHandler(&coordinate.TOCoordinate{}), auth.PrivLevelOperations, Authenticated, nil, 44281121573},
 		{api.Version{4, 0}, http.MethodDelete, `coordinates/?$`, api.DeleteHandler(&coordinate.TOCoordinate{}), auth.PrivLevelOperations, Authenticated, nil, 43038498893},
 
+		//CDN notification
+		{api.Version{4, 0}, http.MethodGet, `cdn_notifications/?$`, cdnnotification.Read, auth.PrivLevelReadOnly, Authenticated, nil, 2221224514},
+		{api.Version{4, 0}, http.MethodPost, `cdn_notifications`, cdnnotification.Create, auth.PrivLevelOperations, Authenticated, nil, 2765223513},
```
ocket8888, line 361: This one can't end with a `/`? why not? Should be anchored with a `$` either way - we've had that introduce weird bugs in the past when that wasn't done.

traffic_ops/v4-client/cdn_notifications.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+)
+
+const (
+	API_CDN_NOTIFICATIONS = apiBase + "/cdn_notifications"
+)
+
+// GetCDNNotificationsWithHdr returns a list of CDN Notifications.
+func (to *Session) GetCDNNotificationsWithHdr(cdnName string, header http.Header) ([]tc.CDNNotification, ReqInf, error) {
```
ocket8888, line 30: This doesn't need to include `WithHdr` - there's no alternative without a header. You might also consider just allowing a `net/url.Values` instead of `cdnName`, so that the endpoint's full functionality is exposed, including pagination and sorting.

traffic_ops/v4-client/cdn_notifications.go
```diff
@@ -0,0 +1,50 @@
...
+// CreateCDNNotification creates a CDN notification.
+func (to *Session) CreateCDNNotification(notification tc.CDNNotificationRequest) (tc.Alerts, ReqInf, error) {
+	var alerts tc.Alerts
+	reqInf, err := to.post(API_CDN_NOTIFICATIONS, notification, nil, &alerts)
+	return alerts, reqInf, err
+}
+
+// DeleteCDNNotification deletes a CDN Notification by CDN name.
+func (to *Session) DeleteCDNNotification(cdnName string) (tc.Alerts, ReqInf, error) {
+	route := fmt.Sprintf("%s?cdn=%s", API_CDN_NOTIFICATIONS, cdnName)
```
ocket8888, line 46: cdnName should be properly encoded. I think `net/url` exports some way to do that besides putting it in a `Values` and using its `Encode` method, but I can't remember for sure.

traffic_ops/v4-client/cdn_notifications.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+const (
+	API_CDN_NOTIFICATIONS = apiBase + "/cdn_notifications"
+)
+
+// GetCDNNotificationsWithHdr returns a list of CDN Notifications.
+func (to *Session) GetCDNNotificationsWithHdr(cdnName string, header http.Header) ([]tc.CDNNotification, ReqInf, error) {
+	var data tc.CDNNotificationsResponse
+	route := fmt.Sprintf("%s?cdn=%s", API_CDN_NOTIFICATIONS, cdnName)
```
ocket8888, line 32: This also needs to be encoded

traffic_ops/traffic_ops_golang/routing/routes.go
mitchell852: what's the problem? this is how all GETs look

mitchell852: what _should_ it look like?

mitchell852: oh sorry, maybe i'm looking at the wrong one

mitchell852: fixed. will push shortly.

traffic_ops/v4-client/cdn_notifications.go
mitchell852: fixing. going to leave out godoc as nowhere else does that.

traffic_ops/traffic_ops_golang/cdnnotification/cdnnotifications.go
mitchell852: where can i find an example of that?

ocket8888: `traffic_ops/traffic_ops_golang/acme/acme_account.go`

traffic_ops/testing/api/v4/cdnnotifications_test.go
```diff
@@ -0,0 +1,61 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+import (
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"testing"
```
ocket8888, line 20: standard library imports should be grouped before project-local imports

traffic_ops/testing/api/v4/cdnnotifications_test.go
```diff
@@ -0,0 +1,61 @@
...
+func GetTestCDNotifications(t *testing.T) {
+	for _, cdn := range testData.CDNs {
+		resp, _, err := TOSession.GetCDNNotifications(cdn.Name, nil)
+		if err != nil {
+			t.Errorf("cannot GET cdn notification for cdn: %v - %v", err, resp)
+		}
+		if len(resp) > 0 {
+			respNotification := resp[0]
+			expectedNotification := "test notification: " + cdn.Name
+			if *respNotification.Notification != expectedNotification {
```
ocket8888, line 38: This will segfault if the CDN has no notification, should check `respNotification.Notification != nil` first

traffic_ops/v4-client/cdn_notifications.go
```diff
@@ -0,0 +1,55 @@
...
+*/
+
+package client
+
+import (
+	"fmt"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/toclientlib"
+	"net/http"
+	"net/url"
```
ocket8888, line 23: standard library imports should be grouped before project-local imports

traffic_ops/testing/api/v4/cdnnotifications_test.go
mitchell852: i guess my "optimize imports" button doesn't work :(

ocket8888: That's a weird-sounding button

ocket8888 (approved):
Seems to work. `linky` behavior seems safe-ish.

PR #5184, opened 2020-10-21
Remove unnecessary JOIN

### LINKED ISSUE #3528: Fix Traffic Ops monitoring.json missing DeliveryServices
Traffic Ops `monitoring.json` is missing DeliveryServices. This is mostly ok, because the Monitor is smart enough to copy them in from the CRConfig. But, the CRConfig doesn't have thresholds. So, Delivery Service thresholds are broken, until this is fixed.

It looks like the bug is that the endpoint is only using DS Profiles which are also a Router Profile. See https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/monitoring/monitoring.go#L350

It may be as simple as removing that line, and the join above it.

### LINKED ISSUE #5180: Global Max Mbps and Tps is not send to TM
## I'm submitting a ...

-  bug report

-  Traffic Monitor
-  Traffic Ops
-  Traffic Portal

## Current behavior:

When a value for Global Max Mbps or Global Max Tps is defined through TP these values is not sent to TM and can not be seen at "api/monitor-config", consequently when the amount of traffic which is being delivered for DS reaches those values extra traffic won't be redirected to HTTP or DNS bypass FQDN or IP.
## Expected behavior:

Values which are defined for Global Max Mbps or Global Max should be sent to TM. As a result when current DS traffic reaches those values TM is supposed to mark that DS unavailable and sent extra traffic to HTTP or DNS bypass FQDN or IP.

ocket8888:
- [x] This PR fixes #5180 and fixes #3528.

- Traffic Ops

Make a GET request to `/cdn/{{Some CDN}}/configs/monitoring` where `Some CDN` is the name of some CDN containing at least one Delivery Service. Notice the array of Delivery Services is not empty.

- master
- 4.1
- 4.0
- 5.0 (RC0)

- [x] This PR includes tests
- [x] No documentation changes for bug fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed #5102 - Python client scripts fail silently on authentication failures
 - Fixed #5103 - Python client scripts crash on connection errors
 - Fixed matching of wildcards in subjectAlternateNames when loading TLS certificates
+- Fixed #5180 - Global Max Mbps and Tps is not send to TM
+- Fixed #3528 - Fix Traffic Ops monitoring.json missing DeliveryServices
 
 ### Changed
 - Changed some Traffic Ops Go Client methods to use `DeliveryServiceNullable` inputs and outputs.
--- a/traffic_ops/traffic_ops_golang/monitoring/monitoring.go
+++ b/traffic_ops/traffic_ops_golang/monitoring/monitoring.go
@@ -1,3 +1,5 @@
+// Package monitoring contains handlers and supporting logic for the
+// /cdns/{{CDN Name}}/configs/monitoring Traffic Ops API endpoint.
 package monitoring
 
 /*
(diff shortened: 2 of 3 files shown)


mitchell852:
2 birds w/ one stone. i like it!

mitchell852:
seems like more than a `minor`

ocket8888:
idk, things seem to have been working fine without this fix for the past year and a half, so I think it's minor. But I don't care what you do to the labels, make it whatever you want

mitchell852:
> idk, things seem to have been working fine without this fix for the past year and a half, so I think it's minor. But I don't care what you do to the labels, make it whatever you want

well ds thresholds have been broken, right? seems more than minor for anyone wanting to use ds thresholds :)

zrhoffman (approved):
- Confirmed that the correct values for `TotalTpsThreshold and `TotalKbpsThreshold` show up in
    - TM's `/api/monitor-config`
    - API versions 1.1, 2.0, and 3.0 of `/cdns/{name}/configs/monitoring`

We don't lose any specificity by removing the JOIN, since a Delivery Service (optionally unless using MSO) has only 1 profile.

PR #5185, opened 2020-10-22
POC: skaffold environment

tenhaus:
This is a proof of concept to help me learn every bit of the TC platform and ultimately provide a development environment that hot-reloads on code changes, doesn't rely on RPM building to start, hacks around the GOPATH-iness of the repository and provides a rough sample implementation of deploying to Kubernetes.

Dependencies:
- docker for desktop installed with the kubernetes option enabled
- skaffold (`brew install skaffold`)

To run:
`skaffold dev` from the root of the repository

TODO:
- [ ] Convert services to Go modules because running `go get` to vender packages on my machine takes ~10 seconds. In a Go module the Dockerfile could be structured to only install dependencies once and this would make local redeploys near instant
- [ ] Strip out traffic ops perl. This takes a really long time to build even if it is just a one-time thing. The latest working group meeting mentioned taking it out for good.
- [ ] Determine if certs are necessary for the services to run. They add overhead and complexity that might not be needed for a development environment.
- [ ] Populate the database
- [ ] Configs populated from environment variables
- [ ] Investigate the scope of router DNS interaction and whether the built-in CoreDNS is suitable
- [ ] Determine if kustomize is useful or overkill. As of now it would be useful for managing configuration but I'm not sure the complexity, although minimal, is worth it atm
- [ ] The other services
- [ ] What makes cdn-in-a-box use so much CPU and how can I possibly avoid it?


tenhaus:
Closing this to focus my time and attention here https://github.com/tenhaus/hightower

PR #5209, opened 2020-10-27
Perlectomy

### LINKED ISSUE #3229: Include carton stdout & stderr output in postinstall.log
The postinstall perl script doesn't appear to be capturing carton's stdout or stderr in postinstall.log.  This would be immensely helpful when triaging missing perl modules once TO is installed, but only partially functional.

### LINKED ISSUE #3724: Traffic ops Api - doesn't handle the invalid values for fallbackId gracefully.
Api version: 1.4 
component: Traffic OPS
**Scenario1:** 
Request: 
/api/1.4/cachegroup_fallbacks?cacheGroupId=<<valid cache group id >> &fallbackId=!131
expected: invalid value - list accepted value type
Actual: refer to the screen shot
Screen Shot 2019-07-15 at 1 52 47 PM

**Scenario2:** 
/api/1.4/cachegroup_fallbacks?fallbackId=`<<valid id>>`
Expected: all the cachegroup that has fallback id associated with it ?!
actual: 
```
{"alerts":[{"level":"error","text":"Invalid cachegroup id, should be an integer"}]}
```
as per the definition here -> https://traffic-control-cdn.readthedocs.io/en/latest/api/cachegroup_fallbacks.html#id3 the request expects atleast one of the two params.

ocket8888:
- [x] This PR closes #3229 and fixes #3724

This PR removes the Traffic Ops Perl codebase. It does _not_ remove API v1 from the Go server.

- CDN in a Box
- Documentation
- Traffic Ops

Make sure the tests pass; the CiaB build proves that at least a minimal CDN can run without Perl.

- [x] Tthis PR does not include a database migration


infrastructure/cdn-in-a-box/traffic_ops/to-access.sh
```diff
@@ -183,8 +182,7 @@ to-enroll() {
 	export MY_GATEWAY="$(route -n | grep $MY_NET_INTERFACE | grep -E '^0\.0\.0\.0' | tr -s ' ' | cut -d ' ' -f2)"
 	MY_NETMASK="$(ifconfig $MY_NET_INTERFACE | grep 'inet ' | tr -s ' ' | cut -d ' ' -f 5)"
 	export MY_NETMASK=${MY_NETMASK#"Mask:"}
-	MY_IP6_ADDRESS="$(ifconfig $MY_NET_INTERFACE | grep inet6 | grep -i global | sed 's/addr://' | awk '{ print $2 }')"
-	export MY_IP6_ADDRESS=${MY_IP6_ADDRESS%%/*}
```
rimashah25, line 187: Why was this removed?

infrastructure/cdn-in-a-box/traffic_vault/poststart.d/02-add-search-schema.sh
```diff
@@ -16,7 +16,7 @@
 # under the License.
 source /to-access.sh
 
-curl -kvs -XPUT -H 'Content-Type:application/xml' "https://$TV_ADMIN_USER:$TV_ADMIN_PASSWORD@$TV_FQDN:$TV_HTTPS_PORT/search/schema/sslkeys" -d @/sslkeys.xml 
+curl -kvs -XPUT -H 'Content-Type:application/xml' "https://$TV_ADMIN_USER:$TV_ADMIN_PASSWORD@$TV_FQDN:$TV_HTTPS_PORT/search/schema/sslkeys" -d @/sslkeys.xml || cat /var/log/riak/*
```
rimashah25, line 19: I am not sure I understand why are we choosing data set as ssl keys or riak log? shouldn't it be only ssl keys unless riak logs = ssl keys

traffic_control/clients/python/pylint.rc
```diff
@@ -259,7 +259,7 @@ redefining-builtins-modules=six.moves,past.builtins,future.builtins,builtins,io
 
 # Format style used to check logging format string. `old` means using %
 # formatting, while `new` is for `{}` formatting.
-logging-format-style=new
+logging-format-style=old
```
rimashah25, line 262: shouldn't we let it be new?

infrastructure/cdn-in-a-box/traffic_ops/to-access.sh
ocket8888: Well it wasn't removed, exactly, it was just changed from defining the variable on one line and exporting it on the next.

I'm not sure exactly what the second line was supposed to do (@zrhoffman would know) but it was crashing the script so I moved the export up, removed that line and now it works.

infrastructure/cdn-in-a-box/traffic_vault/poststart.d/02-add-search-schema.sh
ocket8888: The or here is not being applied to what data to use, it's the second-lowest shell operator binding. So this isn't
```bash
curl -kvs -XPUT -H 'Content-Type:application/xml' "https://$TV_ADMIN_USER:$TV_ADMIN_PASSWORD@$TV_FQDN:$TV_HTTPS_PORT/search/schema/sslkeys" -d (@/sslkeys.xml || cat /var/log/riak/*)
```
(which I don't think would even actually work like that), it's
```bash
(curl -kvs -XPUT -H 'Content-Type:application/xml' "https://$TV_ADMIN_USER:$TV_ADMIN_PASSWORD@$TV_FQDN:$TV_HTTPS_PORT/search/schema/sslkeys" -d @/sslkeys.xml) || (cat /var/log/riak/*)
```
that is, the second command runs if the first one fails.

traffic_control/clients/python/pylint.rc
ocket8888: The new logging format isn't actually supported, Python the project has said they're just going to continue using the old format to avoid breaking people. I don't remember specifics, but I do remember being extremely annoyed by an inability to use the "new" format.

infrastructure/cdn-in-a-box/traffic_vault/poststart.d/02-add-search-schema.sh
rimashah25: oh ok.

traffic_control/clients/python/pylint.rc
rimashah25: copy!

infrastructure/cdn-in-a-box/traffic_ops/to-access.sh
rimashah25: ok. Thanks for the explanation.

traffic_ops/build/traffic_ops.spec
```diff
@@ -206,23 +203,16 @@ if [ "$1" = "0" ]; then
...
 %license ../LICENSE
 %defattr(644,root,root,755)
-%attr(755,%{TRAFFIC_OPS_USER},%{TRAFFIC_OPS_GROUP}) %{PACKAGEDIR}/app/bin/config2json
-%attr(755,%{TRAFFIC_OPS_USER},%{TRAFFIC_OPS_GROUP}) %{PACKAGEDIR}/app/bin/extensions
-%attr(755,%{TRAFFIC_OPS_USER},%{TRAFFIC_OPS_GROUP}) %{PACKAGEDIR}/app/bin/osversions-convert.pl
-%attr(755,%{TRAFFIC_OPS_USER},%{TRAFFIC_OPS_GROUP}) %{PACKAGEDIR}/app/bin/routes.pl
-%attr(755,%{TRAFFIC_OPS_USER},%{TRAFFIC_OPS_GROUP}) %{PACKAGEDIR}/app/bin/start.pl
 %attr(755,%{TRAFFIC_OPS_USER},%{TRAFFIC_OPS_GROUP}) %{PACKAGEDIR}/app/bin/traffic_ops_golang
-%attr(755,%{TRAFFIC_OPS_USER},%{TRAFFIC_OPS_GROUP}) %{PACKAGEDIR}/app/script/cdn
 %attr(755,%{TRAFFIC_OPS_USER},%{TRAFFIC_OPS_GROUP}) %{PACKAGEDIR}/app/script/detect10ginterfaces.pl
```
rimashah25, line 214: Don't we need to delette these .pl files too?

ocket8888: I don't think we should, those scripts might still be in use. The `cdn` script was deleted because it was the entrypoint for Traffic Ops Perl - or one of them, anyway. I think there were like 3 ways to start it.

rimashah25:
Do we need these files in the repo?
image

rimashah25:
Also, the log files for perl? Should we delete them?
image

traffic_ops/build/traffic_ops.spec
rimashah25: Is there a way we can confirm whether these scripts are really in use?

ocket8888:
Yes, we do need the checks, still. People use those - for what I'm not sure, but they do run. They just make calls to the Traffic Ops API, so they should be unaffected if they're using supported API versions.

I do think we can get rid of `log4perl` configuration files under `traffic_ops/app/conf/` - so I'll do that - but I don't know what getting rid of that Jinja2 template will do to the Ansible stuff. Those playbooks are pretty much "experimental" in my estimation, so I'm happy to leave them alone for now.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -1826,32 +1804,18 @@ func DBStatsHandler(db *sqlx.DB) http.HandlerFunc {
...
+}
 
-	managerHandler := CreateThrottledHandler(loggingProxyHandler, d.BackendMaxConnections["mojolicious"])
-	return managerHandler
+// rootHandler returns the / handler for the service, which simply returns a "not found" response.
+func rootHandler(d ServerData) http.Handler {
+	return root{}
 }
 
 // notImplementedHandler returns a 501 Not Implemented to the client. This should be used very rarely, and primarily for old API Perl routes which were broken long ago, which we don't have the resources to rewrite in Go for the time being.
```
rimashah25, line 1821: Is this function (notImplementedHandler) required? since it was primarily used for Perl routes?

rimashah25:
Since perl is removed, we need to removed it from README too
image

Also, build-traffic_ops.logs and build-traffic_ops_ort.log should remove installation steps for perl
image

ciab-rst needs a cleanup too
image

traffic_ops.rst under admin and development folder need cleanup wrt traffic_ops-perl reference

need to remove perl module installs from ciab/mid/Dockerfile

ocket8888:
Perl module installs in the CiaB's "mid" Dockerfile are unrelated to Traffic Ops Perl, they're dependencies of the `traffic_ops_ort` RPM.

The `build-traffic_ops.log` and build-traffic_ops_ort.log` files are not checked into the repository. They're generated by `pkg` when you build RPMs; they're literally just the logs from the Docker containers that build each component.

traffic_ops/traffic_ops_golang/routing/routes.go
ocket8888: Yeah, that's going to need to stick around as long as APIv1 exists. It's basically used to fill "holes" that the Go implementation has in the v1 API.

traffic_ops/build/traffic_ops.spec
ocket8888: We could send out an email on the users mailing list, see if anybody responds in like a week

traffic_ops/traffic_ops_golang/routing/routes.go
rimashah25: ok.. should we add a TODO? since v1API is on its way out soon..

traffic_ops/build/traffic_ops.spec
rimashah25: I like the idea. Lets email and wait until Friday EOB.

traffic_ops/traffic_ops_golang/routing/routes.go
ocket8888: Yeah, I can do that.

rimashah25:
Also, are these two folders required?
image


What about this image?
image

rimashah25:
LGTM
Successfully tested api/v2/v3/v4 TO tests and TP/end-to-end on local CIAB with latest rpm and images.
Required Perl files have been removed and code modified correctly to ensure TO-Perl is no longer required.
PR can be merged once all the checks have successfully passed.

zrhoffman (changes requested):
Building Traffic Ops:

```
 + /usr/bin/cp install/data/perl/osversions.cfg /tmp/go/src/github.com/apache/trafficcontrol/rpmbuild/BUILDROOT/traffic_ops-5.1.0-1.c40273f8.el8.x86_64/var/www/files/.
cp: cannot stat 'install/data/perl/osversions.cfg': No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.oib3Qs (%install)
```

rimashah25:
PR is good for merge.

rawlinp (approved):
LGTM in terms of ripping out the perl bypass functionality from the Routing Blacklist.

rimashah25 (approved):

dewrich:
👍 Nice work!

infrastructure/cdn-in-a-box/optional/docker-compose.debugging.yml
```diff
@@ -81,16 +72,6 @@ services:
       args:
```
zrhoffman, line 72: Same as above: The `-go` in `trafficops-go-nondebug`, `trafficops-go`, `trafficops-go-debug`, and `Dockerfile-go` can be removed.

zrhoffman, line 72: The context needs to be `../..` now

zrhoffman (changes requested):
CDN in a Box fails to build when the debugging yml is included:

```shell
[user@computer cdn-in-a-box]$ docker-compose -f docker-compose.yml -f optional/docker-compose.debugging.yml
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
Building trafficops-go-nondebug
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /home/user/go/src/github.com/apache/trafficcontrol/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile-go: no such file or directory
Service 'trafficops-go-nondebug' failed to build
[user@computer cdn-in-a-box]$
```

dneuman64:
RIP Perl

infrastructure/cdn-in-a-box/optional/docker-compose.debugging.yml
```diff
@@ -62,13 +62,13 @@ services:
...
-    image: trafficops-go
+  # The trafficops-nondebug service ensures that the trafficops base image
+  # exists before trying to build trafficops-debug.
+  trafficops-nondebug:
+    image: trafficops
     build:
-      context: .
-      dockerfile: traffic_ops/Dockerfile-go
+      context: ../../
+      dockerfile: traffic_ops/Dockerfile
```
zrhoffman, line 71: Building `trafficops-nondebug` fails:
```dockerfile
Building trafficops-nondebug
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /home/user/go/src/github.com/apache/trafficcontrol/traffic_ops/Dockerfile: no such file or directory
Service 'trafficops-nondebug' failed to build
```

You'll need to change the rpm path, too.

infrastructure/cdn-in-a-box/optional/docker-compose.debugging.yml
```diff
@@ -32,13 +32,13 @@ services:
       - trafficmonitor-nondebug
   trafficops:
     build:
-      context: .
-      dockerfile: traffic_ops/Dockerfile-go-debug
-    image: trafficops-go-debug
+      context: ../..
```
zrhoffman: Context for this one can stay `.`. Otherwise, it won't find the Dockerfile

zrhoffman (changes requested):

infrastructure/cdn-in-a-box/optional/docker-compose.debugging.yml
ocket8888: The RPM path in `infrastructure/cdn-in-a-box/traffic_ops/Dockerfile` is `infrastructure/cdn-in-a-box/traffic_ops/traffic_ops.rpm` - I think that's correct?

zrhoffman: Yep

zrhoffman: Both `dockerfile` and `TRAFFIC_OPS_RPM` can have the same values as the `trafficops` service in `docker-compose.yml` has. It only needs to be specified again because the service name changes.

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
```diff
@@ -20,129 +20,76 @@
...
+	dnf -y install "https://download.postgresql.org/pub/repos/yum/reporpms/EL-${RHEL_VERSION%%.*}-x86_64/pgdg-redhat-repo-latest.noarch.rpm"; \
+	dnf -y $use_repo -- install postgresql96; \
+	dnf -y install epel-release; \
+	dnf -y install      \
+		jq              \
+		bind-utils      \
+		net-tools       \
+		gettext         \
+		perl-JSON-PP    \
+		perl-Crypt-ScryptKDF  \
```
TaylorCFrey, line 49: Do we still need these perl packages?

OIC, I think this is required because of the components like `adduser.pl` since it's requiring the _Crypt::ScriptKDF_ library.

Should we look into removing the perl scripts as a separate issue in the future?

TaylorCFrey (commented):
Does this mean we will also migrate `traffic_ops_golang` to just `traffic_ops` at some point?

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
zrhoffman: > Should we look into removing the perl scripts as a separate issue in the future?

For sure. All of the Perl cryptography functions it uses are reimplemented in Go and used by traffic_ops_golang, so we should have everything we need to rewrite it.

traffic_ops/app/conf/cdn.conf
```diff
@@ -36,7 +36,6 @@
         "oauth_client_secret": "",
         "routing_blacklist": {
             "ignore_unknown_routes": false,
-            "perl_routes": [],
```
jhg03a, line 39: Can you make a simlilar update to the ansible TO config templates?

jhg03a (commented):
Aside from the minor change to cdn.conf, are there any procedural updates necessary to install/setup traffic_ops?  Is postinstall.pl still in the mix and does it still work after the perlectomy?

traffic_ops/app/conf/cdn.conf
rimashah25: I believe postinstall.pl is replaced by postinstall.py in this PR: https://github.com/apache/trafficcontrol/pull/4763

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
ocket8888: There are also a bunch of other scripts - like the checks - that have their own dependencies which were once part of the carton dependencies of the Perl version of Traffic Ops that we'll probably want to re-write (or remove, as appropriate) before too long.

traffic_ops/app/conf/cdn.conf
ocket8888: So, `postinstall` is unchanged. It even still calls `_postinstall` with the changes made in #4763. The postinstall.py is _added_ as an alternative only, but the default is to still use the Perl version.

Installation is still `yum install trops.rpm` and then run `postinstall`. configuration is mostly untouched, just some things don't have any purpose anymore. I guess the biggest change is that some scripts may expect that by setting their `PERL5LIB` path (or whatever it's called) that their dependencies will be satisfied, but that won't actually be true for new installations. A lot of those aren't used anymore afaik, and many others have active PRs rewriting them. Oh, osversion.cfg is gone, only osversions.json is supported now.

> Can you make a simlilar update to the ansible TO config templates?

I can try, but I don't know much about Ansible - I kind of assume removing the Perl service might just break it all in half.

ocket8888:
> Does this mean we will also migrate traffic_ops_golang to just traffic_ops at some point?

I have a branch for that - I think it's called "making-go-a-first-class-citizen" or something - but it's woefully out-of-date.

Upshot is: I'm gonna try, at some point.

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
```diff
@@ -20,129 +20,76 @@
...
-    infrastructure/cdn-in-a-box/traffic_ops/run.sh \
-    infrastructure/cdn-in-a-box/traffic_ops/config.sh \
-    infrastructure/cdn-in-a-box/traffic_ops/adduser.pl \
-    infrastructure/cdn-in-a-box/traffic_ops/to-access.sh \
-    infrastructure/cdn-in-a-box/traffic_ops/generate-certs.sh \
-    infrastructure/cdn-in-a-box/traffic_ops/trafficops-init.sh \
-    infrastructure/cdn-in-a-box/variables.env \
-    /
+COPY $TRAFFIC_OPS_RPM /traffic_ops.rpm
+RUN rpm --install --nodeps --verbose --hash /traffic_ops.rpm && \
```
zrhoffman, line 76: `--nodeps` was in here for the `trafficops-go` image but not for `trafficops-perl` because the `trafficops-perl` service handled all of the database setup. Now that TO Perl is gone, we should remove `--nodeps` to assert that all of the necessary dependencies have been installed.

zrhoffman (changes requested):

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
zrhoffman: With `--nodeps` removed, there are unmet dependencies for both CentOS 7 and CentOS 8.

Dependencies that CentOS 7 complains about but CentOS 8 does not:
```shell
golang >= 1.12 is needed by traffic_ops-5.1.0-11232.dc08115a.
openssl-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
```

Dependencies that CentOS 8 complains about but CentOS 7 does not:
```shell
perl is needed by traffic_ops-5.1.0-11232.dc08115a.
```
---
The full lists:

With `RHEL_VERSION=7` in the environment and a CentOS 7 Traffic Ops RPM:
```shell
 => ERROR [11/12] RUN rpm --install --verbose --hash /traffic_ops.rpm &&  rm /traffic_ops.rpm                     1.0s
------
 > [11/12] RUN rpm --install --verbose --hash /traffic_ops.rpm &&       rm /traffic_ops.rpm:
error: Failed dependencies:
cpanminus is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
expat-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
gcc-c++ is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
golang >= 1.12 is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
libcurl-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
libidn-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
libpcap-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
openssl-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-DBD-Pg is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-DBI is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-Digest-SHA1 is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-JSON is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-TermReadKey is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-Test-CPAN-Meta is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-WWW-Curl is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-core is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
perl-libwww-perl is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
postgresql96-devel >= 9.6.2 is needed by traffic_ops-5.1.0-11232.dc08115a.el7.x86_64
```

With `RHEL_VERSION=8` in the environment and a CentOS 8 Traffic Ops RPM:
```
 > [11/12] RUN rpm --install --verbose --hash /traffic_ops.rpm &&       rm /traffic_ops.rpm:
error: Failed dependencies:
cpanminus is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
expat-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
gcc-c++ is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
libcurl-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
libidn-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
libpcap-devel is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-DBD-Pg is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-DBI is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-Digest-SHA1 is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-JSON is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-TermReadKey is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-Test-CPAN-Meta is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-WWW-Curl is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-core is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
perl-libwww-perl is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
postgresql96-devel >= 9.6.2 is needed by traffic_ops-5.1.0-11232.dc08115a.el8.x86_64
```

traffic_ops/app/conf/cdn.conf
ocket8888: So, I can't find any reference to this configuration file option in the Ansible files:

```shell-session
$ grep -r 'perl_routes' ./infrastructure/ansible | wc -l
0
```

.github/workflows/ciab.yaml
```diff
@@ -229,11 +229,6 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           path: ${{ github.workspace }}/dist/
-      - name: Cache Perl modules
-        uses: actions/cache@v2
-        with:
-          path: ${{ github.workspace }}/infrastructure/cdn-in-a-box/traffic_ops/perl-cache/centos-${{ env.RHEL_VERSION }}/local
```
zrhoffman, line 235: Nit: The `infrastructure/cdn-in-a-box/traffic_ops/perl-cache/centos-7/local/.gitkeep` and `infrastructure/cdn-in-a-box/traffic_ops/perl-cache/centos-7/local/.gitkeep` files can also be removed.

zrhoffman (changes requested):

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
```diff
@@ -73,7 +78,7 @@ ADD infrastructure/cdn-in-a-box/traffic_ops_data /traffic_ops_data
 ARG TRAFFIC_OPS_RPM=infrastructure/cdn-in-a-box/traffic_ops/traffic_ops.rpm
 
 COPY $TRAFFIC_OPS_RPM /traffic_ops.rpm
-RUN rpm --install --verbose --hash /traffic_ops.rpm && \
+RUN yum -y install /traffic_ops.rpm && \
```
zrhoffman, line 81: On CentOS 8, `yum` is just a symlink to `dnf`

.github/workflows/ciab.yaml
ocket8888: I thought I got those already... done

infrastructure/cdn-in-a-box/optional/docker-compose.debugging.yml
```diff
@@ -71,25 +62,12 @@ services:
...
-  # exists before trying to build trafficops-go-debug.
-  trafficops-go-nondebug:
-    image: trafficops-go
+  # The trafficops-nondebug service ensures that the trafficops base image
+  # exists before trying to build trafficops-debug.
+  trafficops-nondebug:
+    image: trafficops
     build:
-      context: .
-      dockerfile: traffic_ops/Dockerfile-go
```
zrhoffman, line 80: It's still not finding the Dockerfile because the Dockerfile doesn't exist in the directory of the build context. Its path needs to be specified.

```
[user@computer cdn-in-a-box]$ docker-compose -f docker-compose.yml -f optional/docker-compose.debugging.yml build trafficops-nondebug
Building with native build. Learn about native build in Compose here: https://docs.docker.com/go/compose-native-build/
Building trafficops-nondebug
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /home/user/go/src/github.com/apache/trafficcontrol/Dockerfile: no such file or directory
```

zrhoffman (changes requested):

infrastructure/cdn-in-a-box/optional/docker-compose.debugging.yml
zrhoffman: I could have sworn this was working... maybe it was lost in the rebase.

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
ocket8888: interesting

zrhoffman (changes requested):
Building the `trafficops` image fails for CentOS 7, it fails to find the `powertools` repo. That is why the `trafficops-perl` image only added `--enablerepo=powertools` if it was using a CentOS 8 base image:

https://github.com/apache/trafficcontrol/blob/01aa2b9d298356218b5c99067eadd78797f8f946/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile#L65-L70

```dockerfile
  dnf-plugins-core.noarch 0:4.0.2.2-3.el7_6                                     

Dependency Installed:
  python-dateutil.noarch 0:1.5-7.el7                                            
  python2-dnf-plugins-core.noarch 0:4.0.2.2-3.el7_6                             

Complete!
Error: No matching repo to modify: powertools.
The command '/bin/sh -c set -o nounset -o errexit && 	mkdir -p /etc/cron.d; 	if [[ "${RHEL_VERSION%%.*}" -eq 7 ]]; then 		use_repo=''; 	else 		use_repo='--repo=pgdg96'; 	fi; 	yum install -y dnf-plugins-core; 	dnf config-manager --set-enabled powertools; 	dnf -y install "https://download.postgresql.org/pub/repos/yum/reporpms/EL-${RHEL_VERSION%%.*}-x86_64/pgdg-redhat-repo-latest.noarch.rpm"; 	dnf -y $use_repo -- install postgresql96; 	dnf -y install epel-release; 	dnf -y install      		jq              		bind-utils      		net-tools       		gettext         		perl-JSON-PP    		perl-Crypt-ScryptKDF  		mkisofs         		isomd5sum       		nmap-ncat       		openssl         		python3         		golang          		git             		libidn-devel    		libpcap-devel   		perl-Digest-SHA1		rsync;          	dnf clean all' returned a non-zero code: 1
Service 'trafficops' failed to build
```

zrhoffman:
Full command to reproduce that error:

```shell
RHEL_VERSION=7 docker-compose build trafficops
```

TaylorCFrey (approved):
Code review looks good to me.

traffic_ops/app/conf/cdn.conf
jhg03a: You're right, it looks like it was added and will be removed before anyone uses it in https://github.com/jhg03a/trafficcontrol/blob/ansible.refactor/infrastructure/ansible/roles/traffic_ops/templates/cdn.conf.j2

jhg03a: Good point though on the PERL5Lib setup though if you'd like to remove it from https://github.com/apache/trafficcontrol/blob/5397a4fb17a24c255d698ce9594f37562b531c89/infrastructure/ansible/roles/traffic_ops/tasks/traffic_ops.yml#L127.

ocket8888: I think it's fine to leave that there. If it's an existing install that's where the libraries will already be, and if it's a new install I don't think that'll hurt anything.

zrhoffman (approved):
CDN in a Box works as expected on both CentOS 7 and CentOS 8.

dewrich:
Nice work!

PR #5187, opened 2020-10-23
License go-acme/lego dependencies

zrhoffman:
## What does this PR (Pull Request) do?

- [x] This PR is not related to any Issue 

For awhile now, we have had many vendored packages in [traffic_ops/traffic_ops_golang/vendor/github.com/go-acme/lego/vendor/](https://github.com/apache/trafficcontrol/tree/18fe13ac63/traffic_ops/traffic_ops_golang/vendor/github.com/go-acme/lego/vendor). Because the `go-acme/lego` license is applied to all files in the `go-acme/lego` directory,

https://github.com/apache/trafficcontrol/blob/132b56f5c7bfafa63394b07fc85e97b6cae7f315/LICENSE#L465-L467

Weasel has not been checking the licenses of these packages. This PR
- Removes an unused package, `cloud.google.com/go/compute/metadata`
- Removes all vendored `golang.org/x/*` packages because we intentionally do not track those
- Adds a missing license
- Copies the licenses for `github.com/BurntSushi/toml/cmd/tomlv` (from a file named `COPYING`) and `github.com/go-errors/errors` (from a file named `LICENSE.MIT`) into new files named `LICENSE` so Weasel can recognize them (see Comcast/weasel#19)
- Removes duplicate packages that exist elsewhere in the repo
- Moves all remaining packages from `traffic_ops/traffic_ops_golang/vendor/github.com/go-acme/lego/vendor/` to `traffic_ops/traffic_ops_golang/vendor/` and registers them with Weasel in our main `LICENSE` file

The only additions in this PR are
- Modifying the main `LICENSE` file
- Adding 3 other `LICENSE` files
- Adding 2 packages to our list of golang/x dependencies in a few places:
    * tools/golang/Dockerfile
    * traffic_ops/build/build_rpm.sh
    * tools/golang/unit-tests.sh

Everything else in this PR is deletions and moving files.
## Which Traffic Control components are affected by this PR?

- Traffic Ops
- Traffic Ops Auth (experimental)
- Traffic Stats

## What is the best way to verify this PR?

1. <span id="1">Check</span> that each case of removing a duplicate package makes sense. The selections in 196a225205 and 5174c44120 are visualized in the tables below.

    The package versions not removed in 196a225205 have **bolded** git refs:

    Package | traffic_ops/traffic_ops_golang/vendor | traffic_ops/traffic_ops_golang/vendor/github.com/go-acme/lego/vendor | vendor
    --- |:--- |:--- |:---
[...]

PR #5220, opened 2020-10-29
Fix compilation error in experimental tool

ocket8888:

This PR fixes a compilation error in an experimental tool that was introduced when its dependency version changed in #5187.

None

compile the affected package

- master

- [x] Tests are unnecessary for experimental things
- [x] Documentation is unnecessary for experimental things
- [x] An update to CHANGELOG.md is not necessary for experimental things


zrhoffman (approved):
Verified that `traffic_ops_auth` compiles with this patch.

PR #5225, opened 2020-10-29
removed tenancy from service categories

mattjackson220:

This PR removes tenancy from Service Categories

- Documentation
- Traffic Ops
- Traffic Portal

verify that unit and API tests pass
verify that GET, POST, PUT, and DELETE still work for /service_categories
verify that all tenants can see all service categories
verify that when a service category that has already been assigned to a delivery service is updated that that delivery service is also updated
verify that a service category assigned to a delivery service cannot be deleted

### DIFF
--- a/traffic_ops/traffic_ops_golang/routing/routes.go
+++ b/traffic_ops/traffic_ops_golang/routing/routes.go
@@ -420,7 +420,7 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 
 		//ServiceCategories
 		{api.Version{3, 0}, http.MethodGet, `service_categories/?$`, api.ReadHandler(&servicecategory.TOServiceCategory{}), auth.PrivLevelReadOnly, Authenticated, nil, 1085181543, noPerlBypass},
-		{api.Version{3, 0}, http.MethodPut, `service_categories/{name}/?$`, api.UpdateHandler(&servicecategory.TOServiceCategory{}), auth.PrivLevelOperations, Authenticated, nil, 306369141, noPerlBypass},
+		{api.Version{3, 0}, http.MethodPut, `service_categories/{name}/?$`, servicecategory.Update, auth.PrivLevelOperations, Authenticated, nil, 306369141, noPerlBypass},
 		{api.Version{3, 0}, http.MethodPost, `service_categories/?$`, api.CreateHandler(&servicecategory.TOServiceCategory{}), auth.PrivLevelOperations, Authenticated, nil, 553713801, noPerlBypass},
 		{api.Version{3, 0}, http.MethodDelete, `service_categories/{name}$`, api.DeleteHandler(&servicecategory.TOServiceCategory{}), auth.PrivLevelOperations, Authenticated, nil, 1325382238, noPerlBypass},
 
--- a/traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
+++ b/traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
@@ -26,7 +26,6 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/lib/go-tc/tovalidate"
 	"github.com/apache/trafficcontrol/lib/go-util"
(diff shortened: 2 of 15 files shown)


traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -420,7 +420,7 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 
 		//ServiceCategories
 		{api.Version{3, 0}, http.MethodGet, `service_categories/?$`, api.ReadHandler(&servicecategory.TOServiceCategory{}), auth.PrivLevelReadOnly, Authenticated, nil, 1085181543, noPerlBypass},
-		{api.Version{3, 0}, http.MethodPut, `service_categories/{name}/?$`, api.UpdateHandler(&servicecategory.TOServiceCategory{}), auth.PrivLevelOperations, Authenticated, nil, 306369141, noPerlBypass},
+		{api.Version{3, 0}, http.MethodPut, `service_categories/{name}/?$`, servicecategory.Update, auth.PrivLevelOperations, Authenticated, nil, 306369141, noPerlBypass},
```
rawlinp, line 423: Why was this change necessary? Looking at `servicecategory.Update`, it doesn't appear to validate or create a changelog entry, both of which this should be doing via `api.UpdateHandler` already.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/routing/routes.go
ocket8888: I also just finished testing, and trying to update a Service Category with a name of an SC that already exists results in an Internal Server Error

mattjackson220: ill add in validation and changelog entries for it. this was necessary because the service category name is the primary key. Using the UpdateHandler doesnt work because it has to update using the previous information. So when i was using it, it returned that the service category doesnt exist because it was looking for the new name

rawlinp: Ah, thanks for the info -- I didn't catch that on the first pass. That makes sense.

mitchell852:
should there be any validation on name like alphanumerics and dashes?

mitchell852 (approved):
works as i'd expect. just the name validation question i had.

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -123,89 +118,78 @@ func (serviceCategory *TOServiceCategory) Create() (error, error, int) {
...
-			}
+	var origSC TOServiceCategory
+	if err := inf.Tx.QueryRow(`SELECT name FROM service_category WHERE name = $1`, name).Scan(&origSC.Name); err != nil {
+		if err == sql.ErrNoRows {
+			api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("no service category found with name "+name), nil)
+			return
 		}
-
-		return filteredSCByDS, nil, nil, errCode, maxTime
+		log.Errorf("MATT JACKSON in here")
```
rawlinp, line 157: I think you forgot to remove this

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -123,89 +118,78 @@ func (serviceCategory *TOServiceCategory) Create() (error, error, int) {
...
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, err)
+		return
 	}
+	log.Errorf("MATT JACKSON or here")
```
rawlinp, line 161: and this

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -123,89 +118,78 @@ func (serviceCategory *TOServiceCategory) Create() (error, error, int) {
...
-			return true
-		}
+	resp, err := inf.Tx.Tx.Exec(updateQuery(), newSC.Name, name)
+	if err != nil {
+		userErr, sysErr, errCode = api.ParseDBError(err)
+		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
+		return
 	}
-	return false
+	api.CreateChangeLogRawTx(api.ApiChange, api.Updated+" Service Category: "+newSC.Name, inf.User, inf.Tx.Tx)
```
rawlinp, line 169: Since we have the data, we could include the old name as well as the new name here.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
mattjackson220: how embarrassing

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
rawlinp: Whoops, I meant in the changelog entry

rawlinp (approved):
LGTM, works like a charm

PR #5226, opened 2020-10-29
Move clients out of empty directory

ocket8888:

This PR just moves the Python and Java clients out of the annoying, empty, top-level directory `traffic_control`, and also updates references throughout the repository to correctly specify the client path.

Draft because I don't want to change this for 5.0.0

- CDN in a Box
- Documentation
- CI tests

There should be no functionality changes, so just verify that the installation instructions work as now written.

- [x] Tests are unnecessary when functionality isn't changing
- [x] Documentation changes are unnecessary when functionality isn't changing


rawlinp:
_Making a mental note to go fix the myriad things that will break because of this_

ocket8888:
Oh yeah. That's why it's a draft, I think people maintaining production deployments will need more than a couple days heads-up.

jhg03a:
So what's the actual gain here?  Are we considering breaking how people have adopted these SDK just because we don't like the directory structure in the repo?

PR #5227, opened 2020-10-29
Fix CIAB mid enrollment

rawlinp:
The `hostname` command, depending on particular CentOS versions, will sometimes return the FQDN or the short hostname. This fixes the mid enrollment to use `--short` in order to not try to find cachegroups with the suffix `infra.ciab.test` (which don't exist).

- CDN in a Box

Verify the CIAB check completes successfully. If you _really_ want to verify, modify the `mid` Dockerfile to use CentOS 7.6.1810 (the version I happen to be using).

- master
- 5.0

- [x] ciab is a test
- [x] docs unnecessary
- [x] changelog unnecessary


zrhoffman (approved):
Tests pass, this looks like an improvement.

PR #5228, opened 2020-10-30
Migration fixes

ocket8888:

This PR fixes an issue where downgrading the TO DB with null properties on server interfaces would fail, and another issue that cause one migration to sometimes run out of its proper order - or sometimes not at all.

- Traffic Ops Database
- CI tests

The PR includes some CI tests that you should be able to see passing.

- master
- 5.0.0 (RC1)

- [x] This PR includes tests


rawlinp (approved):
LGTM, migration applies in the expected order, and the new coalescence works as expected.

PR #5229, opened 2020-10-30
apache/trafficcontrol is a Go module

zrhoffman:
## What does this PR (Pull Request) do?

- [x] This PR switches us to the Go module system, which makes fixing #4835 or #4380 possible (although this PR does NOT fix either of those issues). 
    * #4380 would be fixed as soon as a vX.Y.Z tag is pushed for a release that includes this PR.
    * This PR preserves *all* of our existing module versions. In order for #4380 to be fixed, a future PR would need to update our dependencies to use v2+ libraries. See afb7bddb8b for all modifications made to our vendored packages in this PR. The only changes in Go files are trivial comments.
    * 4 new dependencies were added (0d536bd851). Note that these are not actually new dependencies, they are previously-unvendored dependencies that only [`/test/router/dnssec`](/apache/trafficcontrol/tree/f1a06fce50/test/router/dnssec) uses:
        - github.com/nxadm/tail
        - github.com/onsi/ginkgo
        - github.com/onsi/gomega
        - gopkg.in/tomb.v1
    * Using Go modules means only vendoring the packages necessary to satisfy dependencies, even if that means vendoring an incomplete repo (although it always includes licenses). Before this PR, there were 102 vendored dependencies. This PR cuts it down to 41:
        ```shell script
        ls -d vendor/{?[^i],github.com/}*/* | wc -l
        ```

## Which Traffic Control components are affected by this PR?

- CDN in a Box
- Documentation
- Grove
- Traffic Control Client (Go)
- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Traffic Stats
- CI tests

## What is the best way to verify this PR?

* Make sure Traffic Control still all builds and works
* Run the various tests affected by this PR:
    - grove/integration_test
    - infrastructure/cdn-in-a-box/traffic_ops_integration_test
    - test/router/dnssec
    - tools/golang
    - traffic_monitor/tests
    - traffic_ops/app/bin/tests
    - traffic_ops/testing/compare

## The following criteria are ALL met by this PR

- [x] This PR updates tests
- [x] This PR includes documentation
- [x] This PR includes an update to CHANGELOG.md
- [x] This PR includes any and all required license headers
- [x] This PR ensures that database migration sequence is correct
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #5230, opened 2020-10-30
Use the standard Traffic Ops v1 client for atstccfg legacy toreq

zrhoffman:

- Updates the `atstccfg/toreq` package to use the standard Traffic Ops v1 client.

- Traffic Ops ORT

- Run the `atstccfg/toreq` unit test suite, make sure it passes
- Run `atstccfg`, make sure it works

- [x] This PR includes tests
- [x] Behavior is unchanged, no documentation necessary


ocket8888:
Shouldn't this be using the v2 client now? Unless we're trying to get this into 5.0.0

rawlinp:
4.1 tries /api/2.0 then /api/1.3.

It seems like 5.0 should be trying /api/3.0 then /api/2.0, right? Unless it doesn't need 2.0 yet and is still good with 1.4?

zrhoffman:
Discussed this with @rob05c. It sounds like he's -1 on using non-vendored TO clients for the legacy `atstccfg/toreq` package unless the client library adds the feature of being able to specify the minor version to use for requests. This is so you can upgrade from the earliest minor release of a last major release instead of needing to first upgrade to the last minor of the last major release (example: Supporting upgrading from 3.0 to 4.1, not just 3.1 to 4.1).

@rob05c does that sound right?

rawlinp:
> unless the client library adds the feature of being able to specify the minor version to use for requests

That smells sort of like an anti-pattern, but would that even be necessary anymore per https://github.com/apache/trafficcontrol/pull/5229#issuecomment-719869508? Sounds like you already figured it out.

zrhoffman:
> > unless the client library adds the feature of being able to specify the minor version to use for requests
> 
> That smells sort of like an anti-pattern, but would that even be necessary anymore per [#5229 (comment)](https://github.com/apache/trafficcontrol/pull/5229#issuecomment-719869508)? Sounds like you already figured it out.

Right. AFAIK, because no active PR depends on #5230, we can take our time to decide what we want to do with the vendored TO client library without holding up anything else.

rob05c:
>Discussed this with @rob05c. It sounds like he's -1 on using non-vendored TO clients for the legacy atstccfg/toreq package unless the client library adds the feature of being able to specify the minor version to use for requests. This is so you can upgrade from the earliest minor release of a last major release instead of needing to first upgrade to the last minor of the last major release (example: Supporting upgrading from 3.0 to 4.1, not just 3.1 to 4.1).
>
>@rob05c does that sound right?

I think the right solution is to make the `v1-client` (and other TO clients) figure out the right minor version from TO to use.

As-is, changing ORT's legacy client to `v1-client` technically doesn't break upgrades, but only because TO 5.0 didn't add any minors to 1.3. But it will as soon as we do.

As soon as ORT's client is on the latest and next-latest versions, it's going to break. For example, suppose ORT had `v2-client` and `v3-client`, and Traffic Control 4.0 had `/api/2.0`, ATC 4.1 added `/api/2.1`, and ATC 5.0 adds `/api/3.0`. Then, if a user had Traffic Ops running 4.0, and used Go Modules (or otherwise updated to the latest), and updated their ORT, it's going to update `v2-client` to the latest, ATC 4.1, which will request `/api/2.1`, which doesn't exist on their TO. But Go Modules promised they could upgrade minors without breaking things, which they're supposed to be able to do safely. Except we broke them.

This isn't limited to ORT: fundamentally, if the TO `vx-client` libraries request the latest minor, and don't fall back, they're going to break Go Modules for everyone using them. So, I think making the TO clients fall back to previous minors is the right solution for Go Modules, and to generally work with older TO's anyway.

That should be safe, as long as we're only adding new symbols in new ATC minors in `lib/go-tc`, and making sure their default/zero/null values make everything behave like it did without them.

We should be able to either use `/ping` today to figure out the latest version the TO supports (I actually have a personal script to do that, because I needed it), or we could add the `/api/x/versions` endpoint @jhg03a has been asking for for years. And if `lib/go-tc` is compatible like it should be, we should be able to just change the request path and nothing else.

zrhoffman:
Supersede by #5247

rob05c:
Bearing in mind what I mentioned above: we can't use v2-client until it falls back to the oldest minor.

zrhoffman:
Understood. But since [`/traffic_ops_ort/atstccfg/toreq/vendor`](/apache/trafficcontrol/commits/master/traffic_ops_ort/atstccfg/toreq/vendor) is gone, it no longer blocks #5229, the Go modules PR.

PR #5231, opened 2020-11-02
Fix service categories if-not-modified/if-match functionality

rawlinp:
This functionality was accidentally lost due to a recent change. Also,
add the missing last_deleted functionality in order for deletions to
count as modifications.

- Traffic Ops

Verify the checks pass. Run the DB upgrade and downgrade, verify they're successful. Make `if-not-modified` and `if-match` `PUT /3.0/service_categories/{name}` with given timestamps and ETags, verify either 200 or 412 status codes as expected (this is what the added test does).

- master
- 5.0.x

- [x] This PR includes tests
- [x] bug fix, no docs necessary
- [x] bug not yet released, no changelog necessary

### DIFF
--- a/traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
+++ b/traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
@@ -144,7 +144,7 @@ func Update(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if !api.IsUnmodified(r.Header, origSC.LastUpdated.Time) {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusPreconditionFailed, errors.New("service category has been modified"), nil)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusPreconditionFailed, errors.New("service category could not be modified because the precondition failed"), nil)
 		return
 	}
 
--- a/traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
+++ b/traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
@@ -40,21 +40,7 @@ type TOServiceCategory struct {
 }
 
 func (v *TOServiceCategory) GetLastUpdated() (*time.Time, bool, error) {
-	found := false
-	lastUpdated := time.Time{}
-	rows, err := v.APIInfo().Tx.Query(`select last_updated from service_category where name=$1`, v.Name)
-	if err != nil {
-		return nil, found, errors.New("querying last_updated: " + err.Error())
-	}
-	defer rows.Close()
-	if !rows.Next() {
-		return nil, found, errors.New("no resource found with this id")
-	}
-	found = true
-	if err := rows.Scan(&lastUpdated); err != nil {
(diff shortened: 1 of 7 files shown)


traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
```diff
@@ -149,14 +135,18 @@ func Update(w http.ResponseWriter, r *http.Request) {
...
+	if err := inf.Tx.QueryRow(`SELECT name, last_updated FROM service_category WHERE name = $1`, name).Scan(&origSC.Name, &origSC.LastUpdated); err != nil {
 		if err == sql.ErrNoRows {
 			api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("no service category found with name "+name), nil)
 			return
 		}
 		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, err)
 		return
 	}
+	if !api.IsUnmodified(r.Header, origSC.LastUpdated.Time) {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusPreconditionFailed, errors.New("service category has been modified"), nil)
```
srijeet0406, line 147: The API error should not be "service category has been modified", rather it should be something like "service category could not be modified because the precondition failed", or something to that effect.

srijeet0406 (changes requested):

traffic_ops/traffic_ops_golang/servicecategory/servicecategories.go
rawlinp: Fixed in https://github.com/apache/trafficcontrol/pull/5231/commits/7230f480a8a75cdf9c14c8ceb2f95f88ee02b26d

traffic_ops/v3-client/serviceCategory.go
```diff
@@ -51,20 +51,22 @@ func (to *Session) CreateServiceCategory(serviceCategory tc.ServiceCategory) (tc
 }
 
 // UpdateServiceCategoryByName updates a service category by its unique name.
-func (to *Session) UpdateServiceCategoryByName(name string, serviceCategory tc.ServiceCategory) (tc.Alerts, ReqInf, error) {
-
+func (to *Session) UpdateServiceCategoryByName(name string, serviceCategory tc.ServiceCategory, header http.Header) (tc.Alerts, ReqInf, error) {
```
srijeet0406, line 54: Is changing the function signature acceptable? I know we had this discussion a while back about the consistency of function signatures and structs, for external clients' use cases not breaking.

rawlinp: I might have been away when that discussion took place, but it has been discussed in the past before with most devs seemingly okay with changing the client method signatures without a deprecation period.

That said, this method is being introduced in the 5.0.0 release, which means we can still change the signature without breaking any clients (assuming this PR gets accepted into 5.0.0 which it should).

srijeet0406 (approved):
LGTM
Built and tested manually
Verified that all the unit tests and API tests pass
Code changes look good too 👍

PR #5243, opened 2020-11-04
Fix IPv6 Address validation in geniso

### LINKED ISSUE #5237: Generate ISO broken with IPv6 addresses
## I'm submitting a ...

-  bug report

-  Traffic Ops

## Current behavior:

When trying to use the /isos 3.0 interface with both an IPv4 and IPv6 IP specified, TO returns an error message back to TP indicating the IPv6 address is invalid when a CIDR netmask is included like normal.  Looks like at some point we accidentally made the form validation in TO no longer support IPv6 addresses with CIDR netmask.

## Expected behavior:

TO API should supply the generate script with all the necessary information

Try to create an iso with a /64 netmask at the end of an IPv6 address like normal.

## Anything else:

```
alerts	[ {…} ]
0	Object { text: "decoding: invalid IP address: ffff:3333:fe2e:b::2/64", level: "error" }
text	"decoding: invalid IP address: ffff:3333:fe2e:b::2/64"
level	"error"
```

ocket8888:
- [x] This PR fixes #5237

Fixes `/isos` not accepting valid IPv6 addresses containing CIDR-notation network prefixes.

- Traffic Ops

Run the validation unit tests. Also try to generate an ISO using an IPv6 address like `2001:DB8::1/63` - observe that it passes validation (in CDN-in-a-Box that'll just get you an internal server error, but that does mean that it got past the validation step).

- master
- 5.0.0RC1

- [x] This PR includes tests


rawlinp (approved):
code looks good, checks passed

PR #5246, opened 2020-11-04
Fix GHA running when PR title or description changed

ocket8888:

This PR fixes the GHA behavior where all checks would re-run when editing only PR metadata - the title or description. The downside to this is that the same trigger that fires when those events occur is also responsible for changes in the base branch of a PR. So from now on if you change the branch into which your PR goes (e.g. 4.1.x -> master) then you'll need to re-run the actions manually, if applicable. I do think that case is much less common, though.

This also disables running actions against PRs in draft mode.

- CI tests

I'm gonna change the PR's title after actions are done running, and you should be able to look through the action history to see that they only ran once (unless I messed something up and have to push more than the initial commit).

- master
- 5.0.0 (RC1)

- [x] This PR modifies testing automation

### DIFF
--- a/.github/workflows/ciab.yaml
+++ b/.github/workflows/ciab.yaml
@@ -55,10 +55,11 @@ on:
       - 'misc/**'
       - 'NOTICE'
       - 'traffic_control/java'
-    types: [ opened, reopened, edited, synchronize ]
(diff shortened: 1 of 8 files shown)


zrhoffman (approved):
Confirmed that edits to the PR like adding a label no longer re-runs the GitHub actions.

PR #5251, opened 2020-11-05
t3c should send atstccfg the default ATS configuration directory

jrushford:

This PR fixes a bug where t3c does not send the default ATS config
directory to atstccfg.  The default ATS configuration directory is used
for config files that have a relative or nil 'Path'.  Also, this re-orders the rpm
package check to occur before any config files are request from
traffic ops.

No issue was created because t3c doesn't have integration tests yet
and is not yet used in production.  This problem was detected while
developing tests for t3c.

- Traffic Ops ORT

t3c is not currently used in production and an integration test framework is still being
developed for it so this tests are forthcoming along with documentation and there is
no need to update the CHANGELOG at this time.

### DIFF
--- a/traffic_ops_ort/t3c/config/config.go
+++ b/traffic_ops_ort/t3c/config/config.go
@@ -33,6 +33,7 @@ import (
 )
 
 var TSHome string = "/opt/trafficserver"
+var TSConfigDir string = "/opt/trafficserver/etc/trafficserver"
 
 const (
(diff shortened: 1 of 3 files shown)


rob05c (approved):
Looks good! I manually tested, t3c drops all configs as-expected. Removed ATS, watched it default to `/opt/trafficserver`. Tested location params still place where their value is, custom `--trafficserver-home` places in that dir/etc/trafficserver as expected.

PR #5253, opened 2020-11-05
GitHub Actions: Cache CPAN modules

zrhoffman:

This PR caches the CPAN modules built for Traffic Ops Perl during the `build-ciab` GitHub action for later reuse.

It shaves about 4 minutes off of the time required to build all of the CDN in a Box images.

- CDN in a Box
- Documentation
- CI tests

* Run CDN in a Box, make sure it works
* Using the instructions added to the Docs, copy the Perl lib out of the Traffic Ops Perl container while it is running, then rebuild the `trafficops-perl` image
  ```shell script
  docker cp $(docker-compose ps -q trafficops-perl):/opt/traffic_ops/app/local ../../traffic_ops/app
  ```
  and make sure you don't need to rebuild all of the images again
* Make sure the *CDN-in-a-Box CI* GHA workflow passes

- [x] This PR updates tests

### DIFF
--- a/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
+++ b/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
@@ -97,7 +97,7 @@ RUN rpm -Uvh /$(basename $TRAFFIC_OPS_RPM) && \
 
 # Run carton again, in case the cpanfile included in the RPM differs from the one used earlier in the
 # build (should never happen - Perl is supposed to be going away)
-RUN POSTGRES_HOME=/usr/pgsql-9.6 PERL5LIB=$(pwd)/local/lib/perl5 ./local/bin/carton && \
+RUN POSTGRES_HOME=/usr/pgsql-9.6 PERL5LIB=$(pwd)/local/lib/perl5 carton && \
     rm -rf $HOME/.cpan* /tmp/Dockerfile /tmp/local.tar.gz
 
 ADD infrastructure/cdn-in-a-box/enroller/server_template.json \
--- a/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
+++ b/infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
@@ -79,6 +79,9 @@ ADD traffic_router/core/src/test/resources/geo/GeoLite2-City.mmdb.gz /opt/traffi
 
 WORKDIR /opt/traffic_ops/app
 ADD traffic_ops/app/cpanfile traffic_ops/install/bin/install_goose.sh ./
+
+# Start with the existing traffic_ops/app/local directory
+COPY traffic_ops/app/local local
 RUN cpanm -l ./local Carton && \
     POSTGRES_HOME=/usr/pgsql-9.6 PERL5LIB=$(pwd)/local/lib/perl5 ./local/bin/carton  && \
     rm -rf $HOME/.cpan* /tmp/Dockerfile /tmp/local.tar.gz ./cpanfile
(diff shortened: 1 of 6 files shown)


infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
```diff
@@ -79,6 +79,9 @@ ADD traffic_router/core/src/test/resources/geo/GeoLite2-City.mmdb.gz /opt/traffi
 
 WORKDIR /opt/traffic_ops/app
 ADD traffic_ops/app/cpanfile traffic_ops/install/bin/install_goose.sh ./
+
+# Start with the existing traffic_ops/app/local directory
+COPY traffic_ops/app/local local
```
rawlinp, line 84: Wouldn't this try to copy my macOS-based `local` dir into a centOS container?

zrhoffman: It will, but `carton` replaces anything it cannot use.

zrhoffman: Granted, I haven't tested on macOS yet. Doing that now

zrhoffman: Tested on Mac. If the wrong versions of CPAN modules or their dependencies are installed, Carton installs the correct version for the platform (since we don't track a `cpanfile.snapshot`). Architecture-specific components are stored in `./local/lib/perl5/darwin-thread-multi-2level` for my Mac, in `./local/lib/perl5/x86_64-linux-thread-multi` for CDN in a Box, etc., so builds for more than one architecture do not conflict.

The only catch is that CPANminus does not have those checks. Installing Carton in the system-wide Perl library in e8ea729ce1. Works for me when CPAN dependencies are vendored locally on macOS using default Perl, v5.18.4, and on Linux.

rawlinp (changes requested):
Looks like it fails to run carton the 2nd time:
```
Step 17/21 : RUN POSTGRES_HOME=/usr/pgsql-9.6 PERL5LIB=$(pwd)/local/lib/perl5 ./local/bin/carton &&     rm -rf $HOME/.cpan* /tmp/Dockerfile /tmp/local.tar.gz
 ---> Running in 90e3632283e7
/bin/sh: ./local/bin/carton: No such file or directory
ERROR: Service 'trafficops-perl' failed to build: The command '/bin/sh -c POSTGRES_HOME=/usr/pgsql-9.6 PERL5LIB=$(pwd)/local/lib/perl5 ./local/bin/carton &&     rm -rf $HOME/.cpan* /tmp/Dockerfile /tmp/local.tar.gz' returned a non-zero code: 127
```

zrhoffman:
> Looks like it fails to run carton the 2nd time:

Fixed the second `carton` command in 85e9c391df. Weird that the `ciab-build` action did not fail.

zrhoffman:
Oh right, it's using cached dependencies, and the local Carton module was cached. Good thing Perl doesn't change very often.

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
zrhoffman: I was mistaken, `x86_64-linux-thread-multi` is not specific enough. Using CPAN modules built with CentOS 7 on CentOS 8, I get

```
/usr/bin/perl: symbol lookup error: /opt/traffic_ops/app/local/lib/perl5/x86_64-linux-thread-multi/auto/List/Util/Util.so: undefined symbol: Perl_xs_apiversion_bootcheck
```

Opened #5262 to cache modules by OS version and to use a `local` directory specific to CDN-in-a-Box.

PR #5261, opened 2020-11-07
Experimental TP update

ocket8888:

This PR does two main things:
1. Update the experimental TP's Angular version to 10
2. Provide a new Servers page with an as-yet incomplete Ag-Grid table of servers.

but it also contains a ton of other changes that I couldn't possibly summarize well enough for anyone to read.

None

The best way is to not. It's experimental, and the tests don't even pass yet.

- [x] This PR includes tests that don't pass but also aren't necessary


mitchell852:
@ocket8888 - looks like a weasel failure. want to fix and i'll get this merged?

PR #5262, opened 2020-11-07
GHA: Cache CPAN modules by OS version

zrhoffman:

As a follow-up to #5253, this PR
- adds CentOS version to the cache keys
- Uses `infrastructure/cdn-in-a-box/traffic_ops/local` as the modules cache location
- Removes copied-in modules and reruns `carton` if modules were copied in and `carton` failed

Caching modules by the version of the OS they were built on is necessary to avoid issues related to version differences in Perl and dynamically-linked libraries.

- CDN in a Box
- Documentation
- CI tests

* Run CDN in a Box, make sure it works
* Using the instructions added to the Docs, copy the Perl lib out of the Traffic Ops Perl container while it is running, then rebuild the `trafficops-perl` image
  ```shell script
  docker cp $(docker-compose ps -q trafficops-perl):/opt/traffic_ops/app/local traffic_ops
  ```
  and make sure you don't need to rebuild all of the modules again
* Make sure the *CDN-in-a-Box CI* GHA workflow passes

 - master (bfcf04a3f1)

- [x] This PR updates tests

### DIFF
--- a/.github/actions/run-ciab/run-ciab.sh
+++ b/.github/actions/run-ciab/run-ciab.sh
@@ -33,7 +33,7 @@ docker_compose='docker-compose -f ./docker-compose.yml -f ./docker-compose.readi
 $docker_compose up -d $logged_services $other_services;
 $docker_compose logs -f $logged_services &
 # Copy built Perl modules for caching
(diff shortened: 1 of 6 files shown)


rawlinp (approved):
Caching worked as expected, CIAB came up fine.

PR #5269, opened 2020-11-10
fixed federations/all IMS

### LINKED ISSUE #5191: Traffic Ops IMS query for /federations/all doesn't work
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
When requesting the `GET /federations/all` route with `use_ims = true` in cdn.conf, the following warning message is printed:
```
WARNING: allfederations.go:242: 2020-10-26T13:13:22.87762-06:00: Couldn't get the max last updated time: pq: got 1 parameters but the statement requires 0
```
This is because the query doesn't use a parameter, but a parameter `""` is passed anyways: 
```
    imsQuery := `SELECT max(t) from (
        SELECT max(fds.last_updated) as t FROM
  federation_deliveryservice fds
  JOIN deliveryservice ds ON ds.id = fds.deliveryservice
  JOIN federation fd ON fd.id = fds.federation
    UNION ALL
    select max(last_updated) as t from last_deleted l where l.table_name='federation_deliveryservice') as res`

    if useIMS {
        runSecond, maxTime = tryIfModifiedSinceQuery(header, tx, "", imsQuery)
```

## Expected behavior:
The query should not be passed a parameter in this case and should succeed.

Set `use_ims = true` in cdn.conf and request `GET /federations/all`.

mattjackson220:

- [x] This PR fixes #5191  

This PR fixes IMS for /federations/all

- Traffic Ops

verify that all TO unit and API tests pass
run TO with IMS enabled. request GET /federations/all with an old If-Modified-Since header and verify that all data is returned successfully and the warning seen previously is not seen any more in TO logs
request GET /federations/all with a recent If-Modified-Since header and verify that a 304 Not Modified is returned and the warning seen previously is not seen any more in TO logs

Previous error in the logs was 
`WARNING: allfederations.go:242: 2020-10-26T13:13:22.87762-06:00: Couldn't get the max last updated time: pq: got 1 parameters but the statement requires 0`

no test, doc, changelog, or license updates since this is a bug fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     to Basho's repo responding with 402 Payment Required. The repo has been removed from the image.
 - [#5069](https://github.com/apache/trafficcontrol/issues/5069) - For LetsEncryptDnsChallengerWatcher in Traffic Router,
     the cr-config location is configurable instead of only looking at `/opt/traffic_router/db/cr-config.json`
-- Fixed #5191 - Error from IMS requests to /federations/all [Related github issue](https://github.com/apache/trafficcontrol/issues/5191)
+- [#5191](https://github.com/apache/trafficcontrol/issues/5191) - Error from IMS requests to /federations/all
 
 ## [5.0.0] - 2020-10-20
 ### Added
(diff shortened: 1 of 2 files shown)


shamrickus (changes requested):
I believe this needs a CHANGELOG.md entry

shamrickus (approved):
LGTM

PR #5272, opened 2020-11-11
Adding plugin config fields to cdn.conf file

srijeet0406:

This PR adds the place holders in the `cdn.conf` for any plugins that users wish to add. Plugins are intended to override existing routes functionality, or creating new functionality specific to a particular user base. The two field `plugins` and `plugin_config` added in this PR will make `cdn.conf` recognize the plugins and route traffic through the new(or overridden) route. For example, say that you want to add a new plugin named `foo`, with a config that has a username and password in it. Your `cdn.conf` file would look like this:

`"plugins": ["foo"],
  "plugin_config": {"foo": {"user": "abc","password": "xyz"}},`

- Traffic Ops

Make sure nightly builds fine with these config changes.

- master

### DIFF
--- a/infrastructure/ansible/roles/traffic_ops/templates/cdn.conf.j2
+++ b/infrastructure/ansible/roles/traffic_ops/templates/cdn.conf.j2
@@ -77,6 +77,6 @@
       "write_timeout" : {{ to_go_write_timeout }},
       "supported_ds_metrics" : {{ to_go_supported_ds_metrics | to_json }},
       "plugins": {{ to_plugin_config.keys() | list | to_json }},
-      "plugin_config" : "{{ to_plugin_config | to_nice_json(indent=2) }}"
+      "plugin_config" : {{ to_plugin_config | to_nice_json(indent=2) }}
    }
 }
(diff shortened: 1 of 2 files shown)


infrastructure/ansible/roles/traffic_ops/templates/cdn.conf.j2
```diff
@@ -75,6 +75,8 @@
       "read_header_timeout" : {{ to_go_read_header_timeout }},
       "read_timeout" : {{ to_go_read_timeout }},
       "write_timeout" : {{ to_go_write_timeout }},
-      "supported_ds_metrics" : {{ to_go_supported_ds_metrics | to_json }}
+      "supported_ds_metrics" : {{ to_go_supported_ds_metrics | to_json }},
+      "plugins": {{ to_plugin_config.keys() | list | to_json }},
+      "plugin_config" : "{{ to_plugin_config | to_nice_json(indent=2) }}"
```
jhg03a, line 80: Remove the quotes around this interpolation or it will try to treat the json as a string

jhg03a (changes requested):

jhg03a (approved):

PR #5280, opened 2020-11-13
Fix LetsEncryptDnsChallengeWatcher config location

### LINKED ISSUE #5069: SteeringTest fails when LetsEncryptDnsChallengeWatcher hangs
## I'm submitting a ...

-  bug report

-  Traffic Router

## Current behavior:

When running the Traffic Router tests in `ExternalTestSuite`, SteeringTest fails with

```
Expected: a value greater than <986>
     but: <986> was equal to <986>
	at com.comcast.cdn.traffic_control.traffic_router.core.external.SteeringTest.z_itemsMigrateFromSmallerToLargerBucket(SteeringTest.java:356)
```

## Expected behavior:

The tests in `SteeringTest` should pass

Run the external tests. For convenience, I have made a branch that runs `SteeringTest` only in CDN-in-a-Box
https://github.com/zrhoffman/trafficcontrol/commits/steering-test

You can use it to run the Traffic Router tests, including external tests, as follows:
```shell
docker-compose up -d
docker-compose -f docker-compose.readiness.yml up readiness # Waits for Traffic Monitor to be usable by Traffic Routers
docker-compose -f docker-compose.traffic-router-test.yml up tr-integration-test
```

## Anything else:

SteeringTest has failed since 18fe13ac63/#3534 (and snuck into 4.1.0). The only 2 assertions that fail are [these](https://github.com/apache/trafficcontrol/blob/386cf14296/traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/external/SteeringTest.java#L356-L357):
```java
		assertThat(rehashedPaths.get(smallerTarget).size(), greaterThan(hashedPaths.get(smallerTarget).size()));
		assertThat(rehashedPaths.get(largerTarget).size(), lessThan(hashedPaths.get(largerTarget).size()));
```

I think this is because SteeringWatcher doesn't get to update [after the POST request to the mock TO server to change the steering attributes](https://github.com/apache/trafficcontrol/blob/386cf14296/traffic_router/core/src/test/java/com/comcast/cdn/traffic_control/traffic_router/core/external/SteeringTest.java#L325-L327).

The reason I don't think SteeringWatcher updates after that point is that `LetsEncryptDnsChallengeWatcher`'s `ScheduledExecutorService` hangs when trying to update `LetsEncryptDnsChallengeWatcher`'s database [from `https://${toHostname}/api/2.0/letsencrypt/dnsrecords/`](https://github.com/apache/trafficcontrol/blob/386cf14296/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/ds/LetsEncryptDnsChallengeWatcher.java#L38).

For whatever reason, `Fetcher`'s request for `LetsEncryptDnsChallengeWatcher` never makes it to HttpDataServer (the mock TO server) even though the host and port seem to line up. But, if you make `Fetcher` return for `LetsEncryptDnsChallengeWatcher` requests right before [it sends the request](https://github.com/apache/trafficcontrol/blob/386cf14296/traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/util/Fetcher.java#L135):
```java
			if (url.contains("letsencrypt")) {
				return http;
			}
```

then `SteeringTest` succeeds. If you don't, the request doesn't seem to ever complete.

--

Other notes:
* Changing `DEFAULT_LE_DNS_CHALLENGE_URL` to `"https://${toHostname}/api/2.0/letsencrypt/dnsrecords"` so it doesn't end in a slash, then [adding a mock response for it](https://github.com/apache/trafficcontrol/tree/386cf14296/traffic_router/core/src/test/resources/api/2.0) does not keep `LetsEncryptDnsChallengeWatcher`'s `Fetcher` request from hanging.
* One `AbstractResourceWatcher` hanging should not make another `AbstractResourceWatcher` hang. If Traffic Router does this outside of tests, it's a serious bug that we should fix.

shamrickus:

- [X] This PR fixes #5069 

This fixes the LEDnsChallengerWatcher when the cr-config file is not located at `/opt/traffic_router/db/cr-config.son`.

- Traffic Router

Make sure `/opt/traffic_router/db/cr-config.json` does not exist and run the TR tests. 
Alternatively, you can run this branch [shamrickus/tr-int-test-2](https://github.com/shamrickus/trafficcontrol/commits/tr-int-test-2) and follow [these](https://github.com/apache/trafficcontrol/issues/5069#issuecomment-717411843) instructions

- master (2b0fad0)
- 4.1.0
- 4.1.1 (RC2)
- 5.0.0 (RC1)

### DIFF
--- a/traffic_router/core/src/main/webapp/WEB-INF/applicationContext.xml
+++ b/traffic_router/core/src/main/webapp/WEB-INF/applicationContext.xml
@@ -89,14 +89,15 @@
 		<property name="steeringRegistry" ref="steeringRegistry" />
 	</bean>
 
-    <bean id="letsEncryptDnsChallengeWatcher" class="com.comcast.cdn.traffic_control.traffic_router.core.ds.LetsEncryptDnsChallengeWatcher">
-        <property name="executorService" ref="ScheduledExecutorService" />
-        <property name="databasesDirectory" ref="databasesDir" />
-        <property name="databaseName" value="$[cache.letsencrypt.database:letsencrypt.json]" />
-        <property name="trafficOpsUtils" ref="trafficOpsUtils" />
-        <property name="trafficRouterManager" ref="trafficRouterManager" />
-		<property name="pollingInterval" value="60000" />
-        <property name="configHandler" ref="ConfigHandler" />
+  <bean id="letsEncryptDnsChallengeWatcher" class="com.comcast.cdn.traffic_control.traffic_router.core.ds.LetsEncryptDnsChallengeWatcher">
+      <property name="executorService" ref="ScheduledExecutorService" />
+      <property name="databasesDirectory" ref="databasesDir" />
+      <property name="databaseName" value="$[cache.letsencrypt.database:letsencrypt.json]" />
+      <property name="trafficOpsUtils" ref="trafficOpsUtils" />
+      <property name="trafficRouterManager" ref="trafficRouterManager" />
+      <property name="pollingInterval" value="60000" />
+      <property name="configFile" value="$[cache.config.json:cr-config.json]" />
+      <property name="configHandler" ref="ConfigHandler" />
 	</bean>
 
 	<bean id="certificatesQueue" class="java.util.concurrent.ArrayBlockingQueue" >
--- a/traffic_router/core/src/main/webapp/WEB-INF/applicationContext.xml
+++ b/traffic_router/core/src/main/webapp/WEB-INF/applicationContext.xml
@@ -89,15 +89,15 @@
 		<property name="steeringRegistry" ref="steeringRegistry" />
 	</bean>
 
-  <bean id="letsEncryptDnsChallengeWatcher" class="com.comcast.cdn.traffic_control.traffic_router.core.ds.LetsEncryptDnsChallengeWatcher">
-      <property name="executorService" ref="ScheduledExecutorService" />
-      <property name="databasesDirectory" ref="databasesDir" />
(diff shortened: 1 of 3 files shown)


traffic_router/core/src/main/webapp/WEB-INF/applicationContext.xml
```diff
@@ -89,14 +89,15 @@
...
-        <property name="configHandler" ref="ConfigHandler" />
+  <bean id="letsEncryptDnsChallengeWatcher" class="com.comcast.cdn.traffic_control.traffic_router.core.ds.LetsEncryptDnsChallengeWatcher">
+      <property name="executorService" ref="ScheduledExecutorService" />
+      <property name="databasesDirectory" ref="databasesDir" />
+      <property name="databaseName" value="$[cache.letsencrypt.database:letsencrypt.json]" />
+      <property name="trafficOpsUtils" ref="trafficOpsUtils" />
+      <property name="trafficRouterManager" ref="trafficRouterManager" />
+      <property name="pollingInterval" value="60000" />
+      <property name="configFile" value="$[cache.config.json:cr-config.json]" />
+      <property name="configHandler" ref="ConfigHandler" />
```
rawlinp, line 100: The formatting here is off

rawlinp (changes requested):
Code looks good, unit and integration tests pass. Just one formatting issue

rawlinp (approved):

PR #5282, opened 2020-11-13
Fix an issue with API v1/2 tests not cleaning up certain v3 structures (topologies)

ocket8888:

This PR fixes an issue when API v1 and v2 tests would fail to run against a database containing Topologies and Delivery Services using some or all of those Topologies, e.g. when running API v3 tests followed by v2 and/or v1.

- Traffic Ops Client (Go) - tests

Run the API v3 tests, then the v1 or v2 tests, verify the database cleanup is successful and testing can proceed.

- master
- 5.0.0 (RC1)

- [x] This PR fixes tests


zrhoffman (approved):
This is perfect

PR #5293, opened 2020-11-17
DNSSEC Refresh check compile fix

ocket8888:

This PR fixes the [DNSSEC Refresh program](https://github.com/apache/trafficcontrol/tree/master/traffic_ops/app/bin/checks/DnssecRefresh) to compile locally. It was failing to compile because it depends on [`github.com/pborman/getopt/v2`](https://github.com/pborman/getopt/v2) which was outside of its vendor tree.

This PR also adds `go test`-ing checks and changes that check to only `go get` the `golang.org/x/*` dependencies we consider external, so that hopefully vendoring issues can be caught earlier from now on.

- Traffic Ops (checks)
- CI tests

Hopefully the GHA checks will take care of verification entirely.

- master

- [x] This PR includes tests


shamrickus (approved):
LGTM
* Builds locally
* Unvendored packages are caught in [GHA](https://github.com/shamrickus/trafficcontrol/actions?query=branch%3AVendor-test)

PR #5297, opened 2020-11-17
Disallow updating regexes that have set number 0 and type HOST_REGEXP

### LINKED ISSUE #2805: Traffic Ops Does Not Update DNSSEC Keys on Delivery Service Update
Traffic Ops Does Not Update DNSSEC Keys on Delivery Service Update.

As far as I can tell, this has been a bug forever. The old Perl endpoint doesn't update them either, and it appears it never has.

This means updating anything in a DS which changes the served FQDNs (e.g. Routing Name, Regexes) will break DNSSEC.

### LINKED ISSUE #5296: Cannot update any regex in Traffic Ops/ Traffic Portal
## I'm submitting a ...

-  bug report

-  CDN in a Box
-  Traffic Ops

## Current behavior:

If you have a regex for a delivery service, you cannot update it.
## Expected behavior:

All regexes should be updateable, with the exception of the following:
1.) If the regex in question has set number `0` and type `HOST_REGEXP`
2.) If there already is another regex with the same order.

Go to Traffic Portal.
In an existing or new delivery service, try changing the pattern of a regex thats associated with this DS.
You'll notice that it fails stating that there already is another regex with the same order.
This is incorrect, because "updating" something means that it has to be already available.

srijeet0406:

- [x] This PR fixes #5296 and potentialy #2805   

- CDN in a Box
- Traffic Ops

Make sure all the unit/ API tests pass
Go to Traffic Portal
Try modifying a regex for any DS
This should be possible as long as the following two conditions are met:
1.) The regex in question does not have set number `0` and type `HOST_REGEXP`
2.) No other regex exists with the set number that you're trying to update the current regex to

- master

- [x] This PR includes tests 


mitchell852:
fyi the text needs to look like this:

```
This PR fixes #2805 and fixes #5296
```

to autoclose both tickets.

mitchell852 (changes requested):
I found a bug. I was able to change the regex type of the position 0 regex from HOST_REGEXP to:

image

srijeet0406:
@mitchell852 Fixed in the latest commit

mitchell852 (approved):
LGTM!

PR #5299, opened 2020-11-18
Add validation to topology updates and server updates/deletions

### LINKED ISSUE #5171: Need to validate PUT/DELETE /servers satisfying a Topology-based Delivery Service's Required Capabilities
## I'm submitting a ...

-  Traffic Ops

## Current behavior:

On PUT /servers, it is possible to change the CDN of a server to unsatisfy a Topology-based Delivery Service's Required Capabilities.

## New behavior:

We need validation to prevent that.

Scenario 1:
1. Have a Topology-based Delivery Service
2. Create a second CDN
3. Move the last Server in a Cache Group used by that Top-based DS to CDN 2.

Scenario 2:
1. Have a Topology-based Delivery Service
3. Delete the last Server in a Cache Group used by that Top-based DS

Validation should prevent both scenarios.

Discovered by @dpham692

### LINKED ISSUE #5286: Server update Topology safeties can be bypassed at lower API versions
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
When modifying a server's Cache Group assignment, the API only checks if this invalidates any Topologies if the client made the request at API version 3.

## Expected behavior:
This restriction should be in place unilaterally across API versions.

In CDN-in-a-Box, create a new EDGE Cache Group and move the `edge` server into it; notice this operation fails in API v3, but not if you request API v2 or lower.

rawlinp:
Adds the following validation to topology updates and server updates/deletions:

Topologies must contain at least 1 server per cachegroup in each of the
CDNs of any assigned delivery services. Otherwise, a delivery service's
topology is not truly representative of reality.

- [x] This PR fixes #5171 and fixes #5286 

- Traffic Ops

Review the added tests, ensure they pass.

- [x] This PR includes tests
- [x] validation, no docs necessary

### DIFF
--- a/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
+++ b/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
@@ -881,6 +881,51 @@ WHERE
 	return dses, nil
 }
 
+// GetDeliveryServiceCDNsByTopology returns a slice of CDN IDs for all delivery services
+// assigned to the given topology.
+func GetDeliveryServiceCDNsByTopology(tx *sql.Tx, topology string) ([]int, error) {
+	q := `
+SELECT
+  COALESCE(ARRAY_AGG(DISTINCT d.cdn_id), '{}'::BIGINT[])
+FROM
+  deliveryservice d
+WHERE
+  d.topology = $1
+`
+	cdnIDs := []int64{}
+	if err := tx.QueryRow(q, topology).Scan(pq.Array(&cdnIDs)); err != nil {
+		return nil, fmt.Errorf("in GetDeliveryServiceCDNsByTopology: querying deliveryservices by topology '%s': %v", topology, err)
+	}
+	res := make([]int, len(cdnIDs))
+	for i, id := range cdnIDs {
+		res[i] = int(id)
+	}
+	return res, nil
+}
+
+// CheckCachegroupHasTopologyBasedDeliveryServicesOnCDN returns true if the given cachegroup is assigned to
+// any topologies with delivery services assigned on the given CDN.
+func CachegroupHasTopologyBasedDeliveryServicesOnCDN(tx *sql.Tx, cachegroupID int, CDNID int) (bool, error) {
+	q := `
+SELECT EXISTS(
+  SELECT
+    1
+  FROM cachegroup c
+  JOIN topology_cachegroup tc on c.name = tc.cachegroup
+  JOIN topology t ON tc.topology = t.name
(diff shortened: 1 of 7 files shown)


ocket8888 (approved):
This all looks solid to me.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -881,6 +881,51 @@ WHERE
...
+	res := make([]int, len(cdnIDs))
+	for i, id := range cdnIDs {
+		res[i] = int(id)
+	}
+	return res, nil
+}
+
+// CheckCachegroupHasTopologyBasedDeliveryServicesOnCDN returns true if the given cachegroup is assigned to
+// any topologies with delivery services assigned on the given CDN.
+func CachegroupHasTopologyBasedDeliveryServicesOnCDN(tx *sql.Tx, cachegroupID int, CDNID int) (bool, error) {
```
zrhoffman, line 908: Godoc doesn't match function name

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -881,6 +881,51 @@ WHERE
...
+	q := `
+SELECT EXISTS(
+  SELECT
+    1
```
zrhoffman, line 912: The `1` is unnecessary, can be just `SELECT FROM`

zrhoffman (commented):
I'm running into a bug where, if a Topology's Cache Group is already empty with respect to the CDN used by a Delivery Service that uses that Topology, you cannot move servers into that Cache Group.

Steps to reproduce in CDN in a Box:

1. Clone `ATS_MID_TIER_CACHE` to a new profile for the `ALL` CDN

2. In the database, make server `mid-01` be in CDN `ALL`
    ```sql
    UPDATE "server" s
    SET cdn_id = (SELECT id FROM cdn WHERE cdn."name" = 'ALL')
    WHERE s.host_name = 'mid-01'
    ```

3. Go to Servers -> `mid-01` and change the CDN to `CDN-in-a-Box`, click `Update`

4. Receive error:
    ```
    server is the last one in its cachegroup, which is used by a topology, so it cannot be moved to another cachegroup: cachegroups with no servers in them: CDN_in_a_Box_Mid-01 (in topologies: demo1-top)
    ```

However, if the `CDN_in_a_Box_Mid-01` Cache Group is not empty with respect to the `CDN-in-a-Box` CDN before attempting the update, the update will succeed.

PR #5308, opened 2020-11-19
Change ORT to use API /update endpoint

### LINKED ISSUE #4438: ORT uses non-standard API endpoint /update/{{host name}}
## I'm submitting a ...
- improvement request (usability, performance, tech debt, etc.)

- CDN-in-a-Box (because of ORT.py)
- Traffic Ops ORT (and also the Python version used by CDN-in-a-Box)

## Current behavior:
To report that updates and/or revalidations were successfully applied on cache servers, ORT calls the `/update/{{host name}}` endpoint, which is not grounded in an API version nor documented anywhere, nor handled in Go.

## Expected / new behavior:
ORT should use only real, documented, supported, versioned, API endpoints. I'm not personally very particular how that's done, though I prefer the bottom-most option (unrealistic though that may be).

• The call could be forwarded to ATSTCCFG which could be updated to it using standard API calls
• ORT itself could be changed to make the request using the standard API
• The endpoint could be rewritten in Go and added to the 2.0 API.

Run ORT, observe the call to an unsupported, non-standard API

rob05c:
Change ORT to use API /update endpoint, instead of the non-API /update.

No tests, HTTP req needs an integration test framework that's still being developed. I manually tested queueing and unqueueing against our prod-like environment, everything worked as expected.
No docs, no interface change.

- [x] This PR fixes #4438 

- Traffic Ops ORT

Run `./atstccfg --set-queue-status --set-reval-status`, verify it works as expected.

Not a bug. But it fixes a "bug" in TO 5.0 calling a broken Perl endpoint, so it should probably be backported to 5.0.

### DIFF
--- a/traffic_ops_ort/atstccfg/getdata/getdata.go
+++ b/traffic_ops_ort/atstccfg/getdata/getdata.go
@@ -35,6 +35,8 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/traffic_ops_ort/atstccfg/config"
+	"github.com/apache/trafficcontrol/traffic_ops_ort/atstccfg/toreq"
+	"github.com/apache/trafficcontrol/traffic_ops_ort/atstccfg/toreqnew"
 )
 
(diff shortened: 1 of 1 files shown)


ezelkow1 (approved):
Tested against staging and saw statuses properly change for chosen host

ocket8888:
CiaB timed out; re-runnig

PR #5309, opened 2020-11-19
Fix Traffic Monitor Log Spam

### LINKED ISSUE #5306: TM repeatedly logging events for ADMIN_DOWN caches
## I'm submitting a ...
-  bug report

-  Traffic Monitor

## Current behavior:
If a cache is `ADMIN_DOWN`, it appears that TM will log an event for the cache every single time it is polled:
```
1605800485.067 host="my-edge", type=EDGE, available=false, msg="Protocol (IPv4) ADMIN_DOWN - available (health) "
1605800487.140 host="my-edge", type=EDGE, available=false, msg="Protocol (IPv6) ADMIN_DOWN - available (health) "
1605800489.045 host="my-edge", type=EDGE, available=false, msg="Protocol (IPv4) ADMIN_DOWN - available (health) "
1605800489.977 host="my-edge", type=EDGE, available=false, msg="Protocol (IPv6) ADMIN_DOWN - available (stat) "
1605800490.857 host="my-edge", type=EDGE, available=false, msg="Protocol (IPv6) ADMIN_DOWN - available (health) "
```

## Expected behavior:
Before 5.0.x, TM would not repeatedly log events for `ADMIN_DOWN` caches every time they were polled. I believe TM should only be logging events for changes in availability, which is how it has always logged events. Since `ADMIN_DOWN` caches are always unavailable for routing, these events should not be logged.

Set a cache to `ADMIN_DOWN`, snapshot, observe TM event logs.

## Anything else:
This appears to affect version 5.0.x/master, compared to 4.1.x.

ocket8888:
- [x] This PR fixes #5306 

- Traffic Monitor

Start CDN-in-a-Box, set one (or more, if you like) cache server to ADMIN_DOWN, notice that only the availability change is logged, not every time it's polled.

- master
- 5.0.x (RC1)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,9 @@ local.tar.gz
 .projectile
 *.bak
 *.iml
+.anjuta/
+.anjuta_sym_db.db
+*.anjuta
 # built local binaries
 ./traffic_ops_golang
 # Created by TP/TO docker integration tests
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ local.tar.gz
 .anjuta/
 .anjuta_sym_db.db
 *.anjuta
+*.kateproject
 # built local binaries
 ./traffic_ops_golang
 # Created by TP/TO docker integration tests
--- a/traffic_monitor/health/cache.go
+++ b/traffic_monitor/health/cache.go
@@ -87,6 +87,9 @@ func GetVitals(newResult *cache.Result, prevResult *cache.Result, mc *tc.Traffic
 
 }
 
+// EvalCacheWithStatusInfo evaluates whether the given cache should be marked
+// available, taking the server's configured Status into account as well as its
+// polling information.
 func EvalCacheWithStatusInfo(result cache.ResultInfo, mc *tc.TrafficMonitorConfigMap, status tc.CacheStatus, serverStatus string) (bool, string, string) {
 	availability := AvailableStr
 	if !result.Available {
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -65,6 +65,10 @@ local.tar.gz
 .projectile
 *.bak
 *.iml
+.anjuta/
+.anjuta_sym_db.db
+*.anjuta
+*.kateproject
```
shamrickus, line 71: What are these for?

traffic_monitor/health/cache.go
```diff
@@ -322,7 +327,7 @@ func CalcAvailability(
 			Ipv6Available: availStatus.Available.IPv6,
 		})
 
-		if available, ok := localStates.GetCache(tc.CacheName(result.ID)); !ok || !available.IsAvailable || !availStatus.ProcessedAvailable {
```
shamrickus, line 325: Originally this was `available.IsAvailable != availStatus.ProcessedAvailable` (4.1). I think it should use that instead of `lastStatus`.
I'm not seeing a log entry when going from ADMIN_DOWN -> ONLINE and I assume this is the reason.

shamrickus (changes requested):

traffic_monitor/health/cache.go
ocket8888: `available` no longer exists

shamrickus: I am now seeing it from that path - which is to say you are correct

shamrickus (approved):
LGTM
* Fixes log pollution

PR #5313, opened 2020-11-19
Add ability to filter DS's by the active flag

### LINKED ISSUE #5305: Add "active=true" query param to deliveryservices API
## I'm submitting a ...

Add the ability to query for only active delivery services.

-  Traffic Ops

## Current behavior:

Can't query for only active DSs.

## New behavior:

Can query for only active DSs :smile:

srijeet0406:

- [x] This PR fixes #5305  

- Documentation
- Traffic Ops

Run TO locally and make sure you can query for active/ inactive delivery services.
For ex, `GET /api/3.0/deliveryservices?active=true` should only return the active DS entries.
Similarly, `GET /api/3.0/deliveryservices?active=false` should only return the inactive DS entries.
`GET /api/3.0/deliveryservices` should return ALL the DS entries(active + inactive).

- master

- [x] This PR includes tests 


docs/source/api/v3/deliveryservices.rst
```diff
@@ -66,6 +66,8 @@ Request Structure
 	| page              | no       | Return the n\ :sup:`th` page of results, where "n" is the value of this parameter, pages are ``limit`` long and the first page is 1.    |
 	|                   |          | If ``offset`` was defined, this query parameter has no effect. ``limit`` must be defined to make use of ``page``.                       |
 	+-------------------+----------+-----------------------------------------------------------------------------------------------------------------------------------------+
+    | active            | no       | Show only the :term:`Delivery Services` that have :ref:`ds-active` set or not based on this boolean(whether or not they are active)     |
```
rawlinp, line 69: I didn't view the rendered docs, but I think this may not render properly (without warnings) because you used spaces instead of tabs. The columns have to line up properly. Also, there should be a space before the `(`.

traffic_ops/testing/api/v3/deliveryservices_test.go
```diff
@@ -396,6 +397,31 @@ func GetTestDeliveryServices(t *testing.T) {
...
+	}
+	totalDSes := len(actualDSes)
+	params := url.Values{}
+	params.Set("active", strconv.FormatBool(false))
+	inactiveDSes, _, err := TOSession.GetDeliveryServicesV30WithHdr(nil, params)
+	if err != nil {
+		t.Errorf("cannot GET DeliveryServices: %v - %v", err, actualDSes)
+	}
+	if len(inactiveDSes) != 0 {
+		t.Errorf("expected none of the delivery services to be inactive, but got %v inactive", len(inactiveDSes))
```
rawlinp, line 413: We shouldn't hardcode data assumptions like this in the tests. Someone might need to add a test in the future that would require an inactive DS, and this test would fail for no reason. Ideally, we should add at least one `active: false` delivery service to tc-fixtures.json, then iterate over all returned DSes and check whether they are all inactive or active, based on the query parameter that was used.

rawlinp (changes requested):

rawlinp (approved):
Test looks good now, and the table renders properly even though there are a couple other lines that also use space indentation.

PR #5314, opened 2020-11-20
Improve validation error message for regions with no division name

### LINKED ISSUE #5285: Uninformative error message creating region without division
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
When creating a new region without specifying a division, the API reports:

> _"division not found"_

## Expected behavior:
Should say e.g. _"'division' is required".

Try to create a region without specifying a division, e.g.

```json
{
    "name": "anything"
}
```

MylesBock:

Improves Validation and Validation error messages for Regions, now when a `POST` is fired at that endpoint without a division name it returns an appropriate error message.

This may require additional documentation changes (swagger docs, maybe)

- [x] This PR fixes #5285 

- Traffic Ops

Follow repro steps in #5285 description, ensure new error message is displayed

### DIFF
--- a/traffic_ops/traffic_ops_golang/region/regions.go
+++ b/traffic_ops/traffic_ops_golang/region/regions.go
@@ -97,8 +97,8 @@ func (region *TORegion) Validate() error {
 	if len(region.Name) < 1 {
 		return errors.New(`region 'name' is required`)
 	}
-	if region.DivisionName == "" || region.Division == 0 {
-		return errors.New(`region 'division' and 'divisionName' are required`)
+	if region.Division == 0 {
+		return errors.New(`region 'division' is are required`)
 	}
 	return nil
 }
--- a/traffic_ops/traffic_ops_golang/region/regions.go
+++ b/traffic_ops/traffic_ops_golang/region/regions.go
@@ -98,7 +98,7 @@ func (region *TORegion) Validate() error {
 		return errors.New(`region 'name' is required`)
 	}
 	if region.Division == 0 {
-		return errors.New(`region 'division' is are required`)
+		return errors.New(`region 'division' is required`)
 	}
 	return nil
 }
--- a/traffic_ops/traffic_ops_golang/region/regions.go
+++ b/traffic_ops/traffic_ops_golang/region/regions.go
@@ -95,7 +95,10 @@ func (region *TORegion) GetType() string {
 
 func (region *TORegion) Validate() error {
 	if len(region.Name) < 1 {
-		return errors.New(`Region 'name' is required.`)
+		return errors.New(`region 'name' is required`)
+	}
+	if region.DivisionName == "" || region.Division == 0 {
+		return errors.New(`region 'division' and 'divisionName' are required`)
 	}
 	return nil
 }
--- a/traffic_ops/traffic_ops_golang/region/regions_test.go
+++ b/traffic_ops/traffic_ops_golang/region/regions_test.go
@@ -20,13 +20,15 @@ package region
  */
 
 import (
+	"testing"
+	"time"
+
+	"github.com/jmoiron/sqlx"
+	sqlmock "gopkg.in/DATA-DOG/go-sqlmock.v1"
+
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/test"
-	"github.com/jmoiron/sqlx"
-	sqlmock "gopkg.in/DATA-DOG/go-sqlmock.v1"
-	"testing"
-	"time"
 )
 
 func getTestRegions() []tc.Region {
--- a/traffic_ops/traffic_ops_golang/region/regions_test.go
+++ b/traffic_ops/traffic_ops_golang/region/regions_test.go
@@ -20,14 +20,13 @@ package region
  */
 
 import (
-	"testing"
-	"time"
-
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/test"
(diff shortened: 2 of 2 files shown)


traffic_ops/traffic_ops_golang/region/regions_test.go
```diff
@@ -109,3 +109,30 @@ func TestInterfaces(t *testing.T) {
...
+		DivisionName: "west",
+		ID:           1,
+		Name:         "region1",
+		LastUpdated:  tc.TimeNoMod{Time: time.Now()},
+	}
+	testTORegion := TORegion{Region: testRegion}
+	errs := test.SortErrors(test.SplitErrors(testTORegion.Validate()))
+	expected := []error{}
+
+	if reflect.DeepEqual(expected, errs) {
```
ocket8888, line 123: You should avoid using `reflect.DeepEqual` to compare error values in tests - and actually I think `golang-lint` warns about this by default. The reason is that errors are consistently appended to outer errors to build an idea of the scope of the error and at what level it occurred. Because of that or just adding more context, error strings change all the time, so using `reflect.DeepEqual` makes tests brittle.

Plus in this particular case, you're only looking for an empty set of errors, so you can easily check `len(errs) != 0` instead.

MylesBock: copypasted test, will update it as it seemed weird where i grabbed it from (`divisions_test.go`)

traffic_ops/traffic_ops_golang/region/regions.go
```diff
@@ -95,7 +95,11 @@ func (region *TORegion) GetType() string {
 
 func (region *TORegion) Validate() error {
 	if len(region.Name) < 1 {
-		return errors.New(`Region 'name' is required.`)
+		return errors.New(`region 'name' is required`)
+	}
+	if region.DivisionName == "" || region.Division == 0 {
+		// golang style guides say error strings shouldn't be capitalized or end with punctuation
```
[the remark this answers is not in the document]
MylesBock: fair enough, was just providing my rationale for changing the existing error message

traffic_ops/traffic_ops_golang/region/regions_test.go
```diff
@@ -109,3 +108,30 @@ func TestInterfaces(t *testing.T) {
...
+		DivisionName: "west",
+		ID:           1,
+		Name:         "region1",
+		LastUpdated:  tc.TimeNoMod{Time: time.Now()},
+	}
+	testTORegion := TORegion{Region: testRegion}
+	errs := test.SortErrors(test.SplitErrors(testTORegion.Validate()))
+
+	if len(errs) > 0 {
+		t.Errorf(`expected no errors,  got %v`, errs)
```
ocket8888, line 122: This expects no errors because you provided `DivisionName`, but the validation requires `Division` *and* `DivisionName` as it's currently written. Which is why the test fails.

ocket8888 (changes requested):

traffic_ops/traffic_ops_golang/region/regions_test.go
MylesBock: Yes, had changes to do that but pivoted because it was pretty late

Test fixture updated to include a division, error message updated

traffic_ops/traffic_ops_golang/region/regions_test.go
```diff
@@ -20,14 +20,13 @@ package region
...
-	"testing"
-	"time"
-
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/test"
 	"github.com/jmoiron/sqlx"
 	sqlmock "gopkg.in/DATA-DOG/go-sqlmock.v1"
+	"testing"
+	"time"
```
ocket8888, line 29: Import grouping should be:

1. Standard library
1. `golang.org/x/*`
1. External/vendored libraries (e.g. `sqlx`)
1. Project-local packages

MylesBock: Done, thanks

MylesBock: Nevermind definitely not resolved, gofmt wont let me follow that convention as-is, goland also formats it the way I pushed it

MylesBock: I went with what gofmt did to it locally, is this okay to squash and proceed?

MylesBock:
@ocket8888 merged your PR

traffic_ops/traffic_ops_golang/region/regions_test.go
MylesBock: What is the intended functionality here for this validation? Should both Division and DivisionName be required?

ocket8888: No, I don't think so. I think only the ID should be required, as the docs say:

> _"The only “division” key that actually matters in the request body is `division`; `divisionName` is not validated and has no effect - particularly not the effect of re-naming the division - beyond changing the name in the API response to this request. Subsequent requests will reveal the true name of the division. Note that if `divisionName` is not present in the request body it will be null in the response, but again further requests will show the true division name (provided it has been assigned to a division)._"

MylesBock: Removed divisionName from the validation, updated test slightly

traffic_ops/traffic_ops_golang/region/regions.go
```diff
@@ -95,7 +95,10 @@ func (region *TORegion) GetType() string {
 
 func (region *TORegion) Validate() error {
 	if len(region.Name) < 1 {
-		return errors.New(`Region 'name' is required.`)
+		return errors.New(`region 'name' is required`)
+	}
+	if region.Division == 0 {
+		return errors.New(`region 'division' is are required`)
```
ocket8888, line 101: nit:

```diff
-		return errors.New(`region 'division' is are required`)
+		return errors.New(`region 'division' is required`)
```

ocket8888 (approved):
Works

PR #5322, opened 2020-11-21
TP: Use developer conf for dev build

shamrickus:

`process.env.NODE_ENV` is the canonical way of checking for which environment the server is running. By default, this is nothing and in that case there should be no effect. Running the grunt dev task sets this to `dev`

This change accomplishes two things:
1) Only the certs need to be generated to run the dev environment (everything else is the default per CiaB).
2) Previously, changing config to allow for local development breaks the CiaB TP as the paths are incorrect. Now it will be unaffected.

- Traffic Portal

Build default task, confirm it loads configDev instead of config
Build for prod, confirm it loads config as normal
Run CiaB, ensure TP works.
Run TP Integration tests.

### DIFF
--- a/docs/source/development/traffic_portal.rst
+++ b/docs/source/development/traffic_portal.rst
@@ -103,14 +103,13 @@ Installing The Traffic Portal Developer Environment
 #. Run ``bower install`` to install client-side dependencies into ``traffic_portal/app/bower_components``. Only needs to be done the first time unless ``traffic_portal/bower.json`` changes.
 #. Make sure that compass is installed and functioning correctly by running ``compass version``. If compass is not available, then it can be installed following the instructions under :ref:`dev-tp-compass`.
 
-#. Run ``grunt`` to package the application into ``traffic_portal/app/dist``, start a local HTTPS server (Express), and start a file watcher.
 #. Modify ``traffic_portal/conf/configDev.js``:
-
 	#. Valid SSL certificates and keys are needed for Traffic Portal to run. Generate these (e.g. using `this SuperUser answer <https://superuser.com/questions/226192/avoid-password-prompt-for-keys-and-prompts-for-dn-information#answer-226229>`_) and update ``ssl``.
 	#. Modify ``api.base_url`` to point to your Traffic Ops API endpoint.
 	#. Modify ``files.static`` to be ``./app/dist/public``.
 	#. Modify ``log.stream`` to be ``./server/log/access.log``. If you have created a ``traffic_portal/log`` directory with ``access.log`` file, use that location.
     #. Ensure ca_bundle.crt is present at ``/etc/pki/tls/certs``. If not present, use the cmd: ``sudo touch ca_bundle.crt``
+#. Run ``grunt`` to package the application into ``traffic_portal/app/dist``, start a local HTTPS server (Express), and start a file watcher.
 #. Navigate to http(s)://localhost:[port|sslPort defined in ``traffic_portal/conf/configDev.js``]
 
 .. note:: The Traffic Portal consumes the Traffic Ops API. Modify traffic_portal/conf/configDev.js to specify the location of Traffic Ops.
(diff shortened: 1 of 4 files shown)


traffic_portal/conf/configDev.js
```diff
@@ -0,0 +1,46 @@
...
+
+// this is the config that is consumed by /server.js on traffic portal startup (sudo service traffic_portal start)
+module.exports = {
+    timeout: '120s',
+    useSSL: true, // set to true if you plan to use https (self-signed or trusted certs).
+    port: 80, // set to http port
+    sslPort: 443, // set to https port
+    // if useSSL is true, generate ssl certs and provide the proper locations.
+    ssl: {
+        key:    '../localhost.key',
```
mitchell852, line 28: this requires it outside the TP directory. do you think it should be `./localhost.key` instead to keep it contained to TP?

traffic_portal/conf/configDev.js
```diff
@@ -0,0 +1,46 @@
...
+        cert:   '../localhost.cert',
```
mitchell852, line 29: is it typically named `localhost.cert` or `localhost.crt`. I ask because i have no idea although i guess it doesn't really matter.

traffic_portal/conf/configDev.js
shamrickus: Yes, I reuse that key in other parts of the CDN so that's where I have it but having it in the TP dir makes more sense.

traffic_portal/conf/configDev.js
shamrickus: I think I've seen it as both, I'll change it to `crt` to be more inline with what we have elsewhere.

docs/source/development/traffic_portal.rst
```diff
@@ -104,13 +104,13 @@ Installing The Traffic Portal Developer Environment
 #. Make sure that compass is installed and functioning correctly by running ``compass version``. If compass is not available, then it can be installed following the instructions under :ref:`dev-tp-compass`.
 
 #. Run ``grunt`` to package the application into ``traffic_portal/app/dist``, start a local HTTPS server (Express), and start a file watcher.
```
mitchell852, line 106: can you take a look at these steps. they don't seem quite right. 

image

i.e. 6 should be after 7 and the subbullets are screwed up

mitchell852: otherwise, all looks good.

docs/source/development/traffic_portal.rst
```diff
@@ -103,14 +103,13 @@ Installing The Traffic Portal Developer Environment
...
 #. Make sure that compass is installed and functioning correctly by running ``compass version``. If compass is not available, then it can be installed following the instructions under :ref:`dev-tp-compass`.
 
-#. Run ``grunt`` to package the application into ``traffic_portal/app/dist``, start a local HTTPS server (Express), and start a file watcher.
 #. Modify ``traffic_portal/conf/configDev.js``:
-
 	#. Valid SSL certificates and keys are needed for Traffic Portal to run. Generate these (e.g. using `this SuperUser answer <https://superuser.com/questions/226192/avoid-password-prompt-for-keys-and-prompts-for-dn-information#answer-226229>`_) and update ``ssl``.
 	#. Modify ``api.base_url`` to point to your Traffic Ops API endpoint.
 	#. Modify ``files.static`` to be ``./app/dist/public``.
 	#. Modify ``log.stream`` to be ``./server/log/access.log``. If you have created a ``traffic_portal/log`` directory with ``access.log`` file, use that location.
     #. Ensure ca_bundle.crt is present at ``/etc/pki/tls/certs``. If not present, use the cmd: ``sudo touch ca_bundle.crt``
```
mitchell852, line 111: can this line be deleted? looks like `ca_bundle.crt` is not needed.

mitchell852 (approved):
looks good

PR #5327, opened 2020-11-24
4954/psql to pgx admin

MylesBock:
Replaces use of `psql` and `dropuser` with `dburl` and `lib/pq`.
Test coverage is unknown

- Traffic Ops

Run the `traffic_ops/db/admin` utility against a postgres instance, ensure the multiline SQL queries do not break anything

### DIFF
--- a/traffic_ops/app/db/admin.go
+++ b/traffic_ops/app/db/admin.go
@@ -20,7 +20,7 @@ package main
  */
 
 import (
-	"bytes"
+	"database/sql"
 	"errors"
 	"flag"
 	"fmt"
@@ -29,6 +29,8 @@ import (
 	"os/exec"
 	"strings"
 
+	_ "github.com/lib/pq"
+	"github.com/xo/dburl"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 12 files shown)


MylesBock:
Converted to draft, need to figure out what's up with CIAB starting up

traffic_ops/app/db/admin.go
```diff
@@ -171,68 +173,73 @@ func parseDBConfig() error {
...
-	fmt.Printf("%s", out)
-	if err != nil {
-		die("Can't create db " + DBName)
+func connectAndExecute(query string, queryErrPrefix string, returnRows bool) *sql.Rows {
+	//todo prepared statements
+	//todo inspect results of executing the above query for all possible cases (db exists, db dne, db inaccessible)
+	//todo determine test coverage of this portion
+	if db, err := dburl.Open(fmt.Sprintf("pg://%s:%s@%s:%s/", DBSuperUser, DBPassword, HostIP, HostPort)); err == nil {
+		if !returnRows {
+			if _, err := db.Exec(query); err != nil {
```
rawlinp, line 182: So some of our files like `seeds.sql` contain multiple sql queries. Doesn't this only execute a single sql query?

MylesBock: Correct, but in this case it's reading an entire file in for things like patch, seed, create and executing it all at once

This has come up apparently in a few discussions on `lib/pq` and other repositories related to golang and postgres, and one project offers a solution (https://github.com/gchaincl/dotsql) but it requires more effort

MylesBock: If it's preferred I could parse in the DML/DDL files and break them down into individual queries

rawlinp: Wow, I thought `db.Exec` would return after running only a single query, hence making it not work for files like `seeds.sql` which have multiple queries. If that works, I'm fine with that -- no need to break them down into individual queries.

MylesBock: I'm 99% sure this is why ciab doesn't come up in the pipeline though, so I'll break it down into individual queries or use dotsql

MylesBock:
Drafting.

https://github.com/MylesBock/trafficcontrol/tree/experimental/restructure

Will return when this project is more conducive of outside help, cya

PR #5329, opened 2020-11-24
Adds a more powerful UI grid for changelogs and removes changelog entry for unqueuing servers updates/revals

mitchell852:
This PR does the following:
1. replaces the UI table used for change logs from the slow jquery datatables to the fast/more powerful ag-grid

image

2. allows the number of days of change logs that TP fetches to be configurable (default is 7 days) in traffic_portal_properties.json and overridden by the user in TP
3. removes the change log entry created when server updates/revals are unqueued (to eliminate the "spam")

- Traffic Ops
- Traffic Portal

- Navigate to https://tp.domain.tld/#!/change-logs, use the table
- Optionally change the value in traffic_portal_properties.json `changeLogs.days`
- Run API tests 
- Run UI tests

- [x] Existing API / UI tests will test for regressions
- [x] No documentation to change

### DIFF
--- a/traffic_ops/traffic_ops_golang/server/update.go
+++ b/traffic_ops/traffic_ops_golang/server/update.go
@@ -102,20 +102,6 @@ func UpdateHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	err = api.CreateChangeLogBuildMsg(
-		api.ApiChange,
-		api.Updated,
-		inf.User,
-		inf.Tx.Tx,
-		"server-update-status",
-		hostName,
-		map[string]interface{}{"host_name": hostName},
-	)
-	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("writing changelog: "+err.Error()))
-		return
-	}
-
 	respMsg := "successfully set server '" + hostName + "'"
 	if hasUpdated {
 		respMsg += " updated=" + strconv.FormatBool(strToBool(updated))
--- a/traffic_portal/app/src/common/models/PropertiesModel.js
+++ b/traffic_portal/app/src/common/models/PropertiesModel.js
@@ -21,6 +21,7 @@ var PropertiesModel = function() {
 	this.loaded = false;
 
 	this.setProperties = function(properties) {
+		if (this.loaded == true) return;
 		this.properties = properties;
 		this.loaded = true;
 	};
--- a/traffic_portal/app/src/common/models/PropertiesModel.js
+++ b/traffic_portal/app/src/common/models/PropertiesModel.js
@@ -21,7 +21,7 @@ var PropertiesModel = function() {
 	this.loaded = false;
 
 	this.setProperties = function(properties) {
-		if (this.loaded == true) return;
+		if (this.loaded) return;
 		this.properties = properties;
 		this.loaded = true;
 	};
--- a/traffic_portal/app/src/common/modules/dialog/input/dialog.input.tpl.html
+++ b/traffic_portal/app/src/common/modules/dialog/input/dialog.input.tpl.html
@@ -24,7 +24,7 @@ under the License.
 <div class="modal-body">
     <p ng-bind-html="params.message"></p>
     <form name="inputForm" novalidate>
-        <input type="text" class="form-control" ng-model="inputValue" ng-maxlength="256">
+        <input type="{{params.type ? params.type : 'text'}}" class="form-control" ng-model="inputValue" ng-maxlength="256">
     </form>
 </div>
 <div class="modal-footer">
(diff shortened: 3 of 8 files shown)


traffic_portal/app/src/common/models/PropertiesModel.js
```diff
@@ -21,6 +21,7 @@ var PropertiesModel = function() {
 	this.loaded = false;
 
 	this.setProperties = function(properties) {
+		if (this.loaded == true) return;
```
ocket8888, line 24: comparisons should use `===` - but also you don't need to compare booleans to constant values. `if (this.loaded)` works just fine.

traffic_portal/app/src/common/modules/table/changeLogs/table.changeLogs.tpl.html
```diff
@@ -20,35 +20,45 @@
 <div class="x_panel">
     <div class="x_title">
         <ol class="breadcrumb pull-left">
-            <li class="active">Change Logs</li>
+            <li class="active">Change Logs <small>[ last <a ng-click="changeDays()">{{days}}</a> days ]</small></li>
```
ocket8888, line 23: If this isn't a link, should use `button[type="button"]` instead of `a`.

traffic_portal/app/src/modules/private/changeLogs/list/index.js
```diff
@@ -27,8 +27,12 @@ module.exports = angular.module('trafficPortal.private.changeLogs.list', [])
 						templateUrl: 'common/modules/table/changeLogs/table.changeLogs.tpl.html',
 						controller: 'TableChangeLogsController',
 						resolve: {
-							changeLogs: function(changeLogService) {
-								return changeLogService.getChangeLogs({ days: 3 });
+							tableName: function() {
+								return 'changeLogs';
+							},
+							changeLogs: function(changeLogService, propertiesModel) {
+								let days = (propertiesModel.properties.changeLogs) ? propertiesModel.properties.changeLogs.days : 7;
```
ocket8888, line 34: `days` is never reassigned, could be `const`

traffic_portal/app/src/common/modules/dialog/input/dialog.input.tpl.html
```diff
@@ -24,7 +24,7 @@ <h4 class="modal-title">{{::params.title}}</h4>
 <div class="modal-body">
     <p ng-bind-html="params.message"></p>
     <form name="inputForm" novalidate>
-        <input type="text" class="form-control" ng-model="inputValue" ng-maxlength="256">
+        <input type="{{params.type ? params.type : 'text'}}" class="form-control" ng-model="inputValue" ng-maxlength="256">
```
ocket8888, line 27: This will not enforce a min or max on numeric inputs, although your message states that log days should be between 1 and 365.

Also `ng-maxlength="{{some constant value}}"` can just be `maxlength="{{some constant value}}"`

traffic_ops/traffic_ops_golang/server/update.go
```diff
@@ -102,20 +102,6 @@ func UpdateHandler(w http.ResponseWriter, r *http.Request) {
...
-		inf.Tx.Tx,
-		"server-update-status",
-		hostName,
-		map[string]interface{}{"host_name": hostName},
-	)
-	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("writing changelog: "+err.Error()))
-		return
-	}
-
```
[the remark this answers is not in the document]
mitchell852: i'll look again but i was pretty sure this is just when Servers are "unqueued" or "unrevaled". So basically if the upd_pending or reval_pending flags are set back to false.

traffic_portal/app/src/common/models/PropertiesModel.js
mitchell852: oh right. i like the `(this.loaded)` syntax

traffic_portal/app/src/common/modules/dialog/input/dialog.input.tpl.html
mitchell852: right but this is a generic component so didn't want to overload it.

traffic_portal/app/src/common/modules/table/changeLogs/table.changeLogs.tpl.html
mitchell852: yes

traffic_ops/traffic_ops_golang/server/update.go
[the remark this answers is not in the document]
mitchell852:
```
// UpdateHandler implements an http handler that updates a server's upd_pending and reval_pending values.
func UpdateHandler(w http.ResponseWriter, r *http.Request) {
```

traffic_portal/app/src/common/modules/dialog/input/dialog.input.tpl.html
ocket8888: I was going to suggest you use an `*ngIf`-`else` to insert different inputs depending on the type, but it turns out that only works in Angular, not AngularJS. In AngularJS you'd just need to do two `ng-if`s which is kind of a cludgy solution. So never mind, I guess.

PR #5330, opened 2020-11-24
Switch to Sqitch - Sqitch DB management proof-of-concept

ocket8888:
This should never be merged. It breaks CDN-in-a-Box and TO and TP in general. That's because it adds a sample migration to change DS Active properties from a boolean to an enum. The tests run only far enough to prove that the database changes take place properly.

This PR changes the DB management system from the abandoned Goose project to one recommended by PostgreSQL itself: Sqitch.

With Sqitch, you don't need to worry about naming your files with a timestamp, or ensuring they have right precision or even that they run in the right order - dependencies are explicit and the plan file keeps track of them, and it allows you to write verification queries to test that your changes really worked. This PR includes one such test for the migration it includes. It also provides functionality akin to the `admin.go` utility out-of-the-box - although this PR also updates that utility to use Sqitch migrations when provided the `--use-sqitch` option-argument on the command-line.

For more information about Sqitch, refer to https://sqitch.org

(The only reason this isn't a draft is so you can see the CI tests run)


rawlinp:
Where does PostgreSQL recommend this? I'm curious to know why.

ocket8888:
They list a bunch of suggested tools on [their wiki page about them](https://wiki.postgresql.org/wiki/Change_management_tools_and_techniques) - this is just one such tool. The one I like the most.

ocket8888:
We moved to Migrate instead - I still like Sqitch better, but at this point it seems this PR has been superseded.

PR #5334, opened 2020-11-25
Refactor TO Go client internals to use a consistent, shared method

rawlinp:
Refactor the TO Go client methods to all use the same set of internal methods (`to.post()`, `to.put()`, `to.get()`, `to.del()`, ) that each funnel down into the same shared method (`to.makeRequestWithHeader()`).

Before this PR, there were a handful of different ways the TO Go client methods made requests, each of them slightly different and inconsistent in their own ways. By funneling all the top-level methods down into the same, shared, consistent method, all the top-level methods now make requests in the same consistent manner. By doing this, it allows us to remove a tremendous amount of duplicated code that has been copied and pasted into new methods as the client has grown.

Additionally, this PR fixes countless little bugs among various methods in the TO Go client:
- returning `nil` instead of the `err` from decoding the response body
- not using `url.QueryEscape()` or `url.Values{}.Encode()` to properly escape query strings
- unchecked errors
- not setting `ReqInf.RemoteAddress` properly

The vast majority of top-level methods returned an error if the http status code was not 304 or < 300. There were a few I had to change in order to make them consistent with the entire client:
- deliveryservice_requests
- roles
- staticdnsentries

These client methods now return an error if the status code is not 304 or < 300, so I had to modify a few TO API tests to reflect that new change in behavior. I don't expect this change in behavior to impact client users very much, since most users (if any) will be checking `err != nil` anyways. In general, the consistency that this change brings to the client will make it less surprising and easier to use.

Note: this PR doesn't change the signatures of the top-level client methods, but it does make `ErrUnlessOkOrNotModified` private (as it should've always been), since it should not be exposed and used publicly.

- Traffic Control Client (Go)

Verify the TO API tests pass, since they heavily exercise the TO Go client.

- [x] This PR includes tests
- [x] internal refactor, no docs necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - #2881 Some API endpoints have incorrect Content-Types
 - [#5311](https://github.com/apache/trafficcontrol/issues/5311) - Better TO log messages when failures calling TM CacheStats
 
+### Changed
+- Refactored the Traffic Ops Go client internals so that all public methods have a consistent behavior/implementation
+
 ## [5.0.0] - 2020-10-20
 ### Added
 - Traffic Ops Ort: Disabled ntpd verification (ntpd is deprecated in CentOS)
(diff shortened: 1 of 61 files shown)


rob05c (approved):
Looks great! This is a solid improvement for reducing duplicate code.

Tests are passing, and I manually tested with ORT and it generated the same config.

PR #5343, opened 2020-12-02
TR log warnings for 503 status codes from TM

rawlinp:
When TR receives a 503 from TM, log it instead of silently retrying.

- Traffic Router

Run TR but point it at a TM (or a mock http server) that returns a 503 (python script below can be used).

- [x] no docs necessary

## Additional Information
This python script can be used to return 503s:
```
#!/usr/bin/env python3

from http.server import BaseHTTPRequestHandler, HTTPServer

class MyHandler(BaseHTTPRequestHandler):
  def do_GET(self):
    try:
      self.send_response(503) # let bots know whats up
      self.end_headers()
      return
    except IOError:
      self.send_error(500,'Internal error')

class TimeoutingHTTPServer(HTTPServer):
  def finish_request(self, request, client_address):
    request.settimeout(5) # Really short timeout as there is only 1 thread
    HTTPServer.finish_request(self, request, client_address)

def main():
  try:
    server = TimeoutingHTTPServer(('',8081), MyHandler)
    print('Listening...')
    server.serve_forever()
  except KeyboardInterrupt:
    print('quit...')
    server.socket.close()

if __name__ == '__main__':
  main()
```


mattjackson220 (approved):
looks good to me! logs the code as expected, tests pass

Issue #5244, opened 2020-11-04
Limit request_header_max_size per DS


dsouza93:
## I'm submitting a ...

-  new feature / enhancement request

## Traffic Control components affected ...

-  Documentation
-  Traffic Ops
-  Traffic Ops ORT
-  Traffic Portal

## Current behavior:

Currently, ATS' request_header_max_size is only configurable globally. It is not set on a per delivery service basis. Delivery service owners should have the ability to increase their header_max_size above the global limit if their origin is capable of handling that request and it is required for their delivery lane.

## New behavior:

It would be ideal if Request Max Header Size was an additional configurable value in the Delivery Service config and integrated into Traffic Portal as a field.  

The request_header_max_size value is not overridable using header rewrite, so ATC would likely need to configure ATS under the hood by injecting: 

cond %{REMAP_PSEUDO_HOOK}
cond %<cqhl> > {Configured Value in Bytes} 
set-status 400
into the delivery services header rewrite. 

It is also worth noting that we would keep the global variable in play, set higher than our default as a last line of defense. If the Delivery service configured value is higher than the global, it will be useless as the global takes precedence. It would be nice if either the Portal or TO could prevent the user from that misconfiguration.

rob05c:
+1 - this is something every DS owner should at least think about. Setting it poorly is a security vulnerability on their end.
The options are DS field, Parameter, or custom Header Rewrite text. The latter two are unintuitive and dangerous. IMO this is important enough to warrant a DS field.

srijeet0406:
Talked to @dsouza93 and @rob05c .
Dylan said that the check to see if the DS specific request header size is > the global max value is not needed.
Rob mentioned that the ORT effort should be its own separate issue/ ticket. So the PR https://github.com/apache/trafficcontrol/pull/5345 just deals with adding an extra field to the DS structure in TO/TP.

PR #5345, opened 2020-12-03
Add a new field to the Delivery Service to accommodate for "max_request_header_size" 

srijeet0406:

- Documentation
- Traffic Control Client 
- Traffic Ops

Run the `goose` migrations for both dev and test
Make sure all the unit tests/ API tests pass
Try creating a new DS in TP, and make sure the `Max Request Header Size` field shows up.
Make sure you can also add and remove that field from the TP DS table view.
Try updating that field and make sure that the updates take effect.

- master

- [x] This PR includes tests
- [x] This PR includes database migrations

## Additional Information

This PR is the prerequisite for ORT to complete https://github.com/apache/trafficcontrol/issues/5244
Once this is merged, ORT will be required to inject the header rewrite rules to enforce that the constraint is followed


lib/go-tc/deliveryservices.go
```diff
@@ -172,11 +174,12 @@ type DeliveryServiceV11 struct {
...
-	FirstHeaderRewrite *string `json:"firstHeaderRewrite" db:"first_header_rewrite"`
-	InnerHeaderRewrite *string `json:"innerHeaderRewrite" db:"inner_header_rewrite"`
-	LastHeaderRewrite  *string `json:"lastHeaderRewrite" db:"last_header_rewrite"`
-	ServiceCategory    *string `json:"serviceCategory" db:"service_category"`
+	Topology             *string `json:"topology" db:"topology"`
+	FirstHeaderRewrite   *string `json:"firstHeaderRewrite" db:"first_header_rewrite"`
+	InnerHeaderRewrite   *string `json:"innerHeaderRewrite" db:"inner_header_rewrite"`
+	LastHeaderRewrite    *string `json:"lastHeaderRewrite" db:"last_header_rewrite"`
+	ServiceCategory      *string `json:"serviceCategory" db:"service_category"`
+	MaxRequestHeaderSize *int    `json:"maxRequestHeaderSize" db:"max_request_header_size"`
```
rawlinp, line 182: Since 5.0 is leaving the station, we probably need a new `DeliveryServiceNullableV31` struct and all the other changes that come with that, since this will likely be 3.1. Right @ocket8888? This shouldn't be backported.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -858,6 +859,7 @@ func updateV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, ds *tc.
 		&ds.InnerHeaderRewrite,
 		&ds.LastHeaderRewrite,
 		&ds.ServiceCategory,
+		&ds.MaxRequestHeaderSize,
```
rawlinp, line 862: The way the DS API works, we also need to query the existing value of `max_request_header_size` in `updateV15`, in order for 1.5 requests to be properly _upgraded_ to 3.0 requests. Otherwise, a pre-3.0 request would blow away the existing value.

The [README](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/README.md#updating-a-minor-version) explains this a bit better, but might be out of date now that `DeliveryServiceNullable` is deprecated :(

rawlinp (changes requested):

lib/go-tc/deliveryservices.go
ocket8888: Well hopefully the next release can include the DS `active` flag change, so we may actually want to go to 4.0. That'll be a breaking change that needs a new major version bump.

srijeet0406: The DS `active` flag change(that I worked on) is not modifying the structure. It just lets us query by the `active` flag.

rawlinp: He's referring to https://github.com/apache/trafficcontrol/blob/master/blueprints/ds-active.md which hasn't been started yet. That blueprint would call for a major version bump.

_This_ PR could add it as 3.1, then if we get to ☝️ before 5.1 is released, that API could still be added as 4.0. Then 3.1 == 4.0 like how 1.5 == 2.0 (if that makes sense). It _would_ be cleaner if we were able to have just one or the other, but I don't think it's a big deal if we end up having to add both.

lib/go-tc/deliveryservices.go
```diff
@@ -179,6 +188,11 @@ type DeliveryServiceNullableV30 struct {
 	ServiceCategory    *string `json:"serviceCategory" db:"service_category"`
 }
 
+type DeliveryServiceV31 struct {
+	DeliveryServiceNullableV30
+	MaxRequestHeaderSize *int `json:"maxRequestHeaderSize" db:"max_request_header_size"`
+}
+
 // Deprecated: Use versioned structures only from now on.
 type DeliveryServiceNullable DeliveryServiceNullableV15
```
rawlinp, line 197: So I had some thoughts @ocket8888 @rob05c. I think we may want to continue using type aliases like this, but only per major version.

Otherwise, let's say we end up with another 5 minor versions in 3.x. That means we will have 5 different `GetDeliveryServiceV*`, `UpdateDeliveryServiceV*`, etc, all in the same client package. I don't think that's what we want the client to look like, right? As a user, I wouldn't want to have to pick between all those minor version methods. I thought we wanted it to auto-downgrade minor versions, which means all the client functions should be using a type alias per major version, e.g. `type DeliveryServiceV3 DeliveryServiceV31`, which we continually update to alias to new minor versions, e.g. `type DeliveryServiceV3 DeliveryServiceV32`. Does that make sense?

Since we already have `DeliveryServiceNullableV30`, maybe that should be the current alias. Then we don't need to add new client methods for V31.

lib/go-tc/deliveryservices.go
```diff
@@ -335,6 +349,46 @@ func ParseOrgServerFQDN(orgServerFQDN string) (*string, *string, *string, error)
 	return &protocol, &FQDN, port, nil
 }
 
+func (ds *DeliveryServiceV31) Sanitize() {
```
rawlinp, line 352: I think we might not need this if we alias `type DeliveryServiceV30 DeliveryServiceV31` like I mention in the previous comment. But if we don't want to do that for some reason, I think the matching `DeliveryServiceV30` methods would no longer be in use and could be removed. We should only be calling `Sanitize`, `Validate`, etc. in the _latest_ handlers.

traffic_ops/app/db/migrations/2020113000000000_add_max_request_header_size_delivery_service.sql
```diff
@@ -0,0 +1,22 @@
...
+    limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE deliveryservice ADD COLUMN max_request_header_size int DEFAULT 131072;
+
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+ALTER TABLE deliveryservice DROP COLUMN max_request_header_size;
```
rawlinp, line 22: missing newline at end of file

traffic_ops/testing/api/v3/deliveryservices_test.go
```diff
@@ -69,6 +71,21 @@ func TestDeliveryServices(t *testing.T) {
 	})
 }
 
+func GetTestDeliveryServicesV31(t *testing.T) {
```
rawlinp, line 74: We shouldn't need new minor-version specific tests if we follow my first comment above. We could just check this new field in the existing `GetTestDeliveryServices` test (similar for update).

traffic_ops/testing/api/v3/traffic_control_test.go
```diff
@@ -31,6 +31,7 @@ type TrafficControl struct {
 	DeliveryServiceRequests                           []tc.DeliveryServiceRequest             `json:"deliveryServiceRequests"`
 	DeliveryServiceRequestComments                    []tc.DeliveryServiceRequestComment      `json:"deliveryServiceRequestComments"`
 	DeliveryServices                                  []tc.DeliveryServiceNullableV30         `json:"deliveryservices"`
+	DeliveryServicesV31                               []tc.DeliveryServiceV31                 `json:"deliveryservicesV31"`
```
rawlinp, line 34: We probably wouldn't need this and the new additions to tc-fixtures.json either (assuming we alias like I mention in the first comment).

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -144,6 +144,13 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		/**
 		 * 3.x API
 		 */
+		////DeliveryServices
+		{api.Version{3, 1}, http.MethodGet, `deliveryservices/?$`, api.ReadHandler(&deliveryservice.TODeliveryService{}), auth.PrivLevelReadOnly, Authenticated, nil, 22383173943, noPerlBypass},
```
rawlinp, line 148: We should only need 3.1 routes for POST and PUT. The GET handler should have the minor-version specific handling, and DELETE is version-agnostic.

traffic_ops/v3-client/deliveryservice.go
```diff
@@ -34,13 +34,25 @@ const (
 	// See Also: https://traffic-control-cdn.readthedocs.io/en/latest/api/v3/deliveryservices.html
 	API_DELIVERY_SERVICES = apiBase + "/deliveryservices"
 
+	// API_DELIVERY_SERVICES_V31 is the API path on which Traffic Ops serves Delivery Service
+	// information. More specific information is typically found on sub-paths of this.
+	// See Also: https://traffic-control-cdn.readthedocs.io/en/latest/api/v3/deliveryservices.html
+	API_DELIVERY_SERVICES_V31 = apiBaseV31 + "/deliveryservices"
```
rawlinp, line 40: If we follow my aliasing comment above, I think we would just update `apiBase` to 3.1, and we wouldn't need the new 3.1-specific client methods below.

traffic_portal/app/src/common/api/DeliveryServiceService.js
```diff
@@ -20,7 +20,7 @@
 var DeliveryServiceService = function($http, locationUtils, messageModel, ENV) {
 
     this.getDeliveryServices = function(queryParams) {
-        return $http.get(ENV.api['root'] + 'deliveryservices', {params: queryParams}).then(
+        return $http.get(ENV.api['deliveryServices'] + 'deliveryservices', {params: queryParams}).then(
```
rawlinp, line 23: I think we would just update `ENV.api['root']` to 3.1, right @mitchell852? I think we would prefer to have TP using the same version for all requests if possible.

rawlinp (changes requested):
I'd like to get some feedback from @ocket8888 and @rob05c on this as well. We may need to rethink the whole "deprecate the `DeliveryServiceNullable` alias" thing.

lib/go-tc/deliveryservices.go
```diff
@@ -335,6 +349,46 @@ func ParseOrgServerFQDN(orgServerFQDN string) (*string, *string, *string, error)
...
+	if ds.MaxOriginConnections == nil || *ds.MaxOriginConnections < 0 {
+		ds.MaxOriginConnections = util.IntPtr(0)
+	}
+	if ds.DeepCachingType == nil {
+		s := DeepCachingType("")
+		ds.DeepCachingType = &s
+	}
+	*ds.DeepCachingType = DeepCachingTypeFromString(string(*ds.DeepCachingType))
+	if ds.MaxRequestHeaderSize == nil {
+		ds.MaxRequestHeaderSize = util.IntPtr(131072)
```
TaylorCFrey, line 388: Is there something special about this value (131072)? Should we asign it as a variable to avoid magic number confusion in the future?

lib/go-tc/deliveryservices.go
```diff
@@ -32,6 +32,8 @@ import (
 */
 
 const DefaultRoutingName = "cdn"
+const MaxUint = ^uint(0)
```
TaylorCFrey, line 35: Are these max values used?

lib/go-tc/deliveryservices.go
```diff
@@ -179,6 +188,11 @@ type DeliveryServiceNullableV30 struct {
 	ServiceCategory    *string `json:"serviceCategory" db:"service_category"`
 }
 
+type DeliveryServiceV31 struct {
```
TaylorCFrey, line 191: Do we need to signal if this, too, is `nullable` or not, like the nested struct?

lib/go-tc/deliveryservices.go
```diff
@@ -532,6 +586,121 @@ func (ds *DeliveryServiceNullableV30) validateTypeFields(tx *sql.Tx) error {
...
+			validation.NewStringRule(validateOrgServerFQDN, "must start with http:// or https:// and be followed by a valid hostname with an optional port (no trailing slash)")),
+		"rangeSliceBlockSize": validation.Validate(ds,
+			validation.By(func(dsi interface{}) error {
+				ds := dsi.(*DeliveryServiceV31)
+				if ds.RangeRequestHandling != nil {
+					if *ds.RangeRequestHandling == 3 {
+						return validation.Validate(ds.RangeSliceBlockSize, validation.Required,
+							// Per Slice Plugin implementation
+							validation.Min(262144),   // 256KiB
+							validation.Max(33554432), // 32MiB
```
TaylorCFrey, line 638: Would these also be considered _"magic numbers"_ even though they are explained in comments?

lib/go-tc/deliveryservices.go
ocket8888: That should work out fine, since the only kinds of changes that should break alias property access are the same kinds that would necessitate a major version bump.

lib/go-tc/deliveryservices.go
rawlinp: I believe the general rule is that new structs moving forward should be nullable, so it's not really necessary to call that out for new structs anymore.

traffic_portal/app/src/common/api/DeliveryServiceService.js
ocket8888: That's a possibility, but it will mean needing to add routes to `routes.go` for 3.1 for all API routes.

rawlinp: That's not the way routes work unless it was changed. If there is a single 3.1 route in existence, any handlers for 3.0 will implicitly handle 3.1 requests.

lib/go-tc/deliveryservices.go
srijeet0406: This is the default value used for our header sizes, I have made it into a constant.

lib/go-tc/deliveryservices.go
srijeet0406: Made constants for these values as well.

lib/go-tc/deliveryservices.go
TaylorCFrey: Cool. I meant is there anything unique about 131072 vs, say, 32768? Is that bits or bytes? According to https://stackoverflow.com/questions/686217/maximum-on-http-header-values it looks like the max header size is 4K to 8K, but it depends on the underlying server or page files, etc. Not sure if this that is even relevant.

lib/go-tc/deliveryservices.go
```diff
@@ -179,6 +189,13 @@ type DeliveryServiceNullableV30 struct {
 	ServiceCategory    *string `json:"serviceCategory" db:"service_category"`
 }
 
+type DeliveryServiceV31 struct {
+	DeliveryServiceV30
+	MaxRequestHeaderSize *int `json:"maxRequestHeaderSize" db:"max_request_header_size"`
```
rob05c, line 194: Would you object to changing this to `MaxRequestHeaderBytes`?
I think it makes it easier, so people don't have to look up the documentation to know the units, and don't assume the wrong unit and make a mistake.

lib/go-tc/deliveryservices.go
rob05c: FWIW, I ended up making an alias in `lib/go-atscfg` for both the latest DS and Server. Just for stability, because it kept changing, and requiring updating hundreds of lines of code. Where using the alias lets me just update that single location, and the few places that broke.

I don't currently have a strong opinion on this. But I think we should do whatever is easy for users to work with. From my recent experience, that seems to be an alias.

> That should work out fine, since the only kinds of changes that should break alias property access are the same kinds that would necessitate a major version bump.

Right, that's the big issue: how does an updating alias work with Go Modules and the library Version Promise?
It's a little bit nebulous what constitutes "breaking." The updating minor alias will only break if people use the nested initialization, e.g. `sv := tc.Server{tc.ServerV3{tc.ServerV2{Foo: foo}, Bar: bar}}` vs `sv := tc.Server{}; sv.Foo = foo; sv.Bar = bar`.

I think I'd be ok with declaring that unsupported in terms of the Go Modules and the library compatibility promise, and documenting so a symbol comment. "Struct alias initialization is only guaranteed forward-compatible for default-initialization."

rawlinp: That sounds good to me.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -629,117 +860,352 @@ WHERE
...
+	}
+
+	res, status, userErr, sysErr := updateV13(w, r, inf, &dsV13)
+	if res != nil {
+		return &res.DeliveryServiceNullableV12, status, userErr, sysErr
+	}
+	return nil, status, userErr, sysErr
+}
+
+func updateV13(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, reqDS *tc.DeliveryServiceNullableV13) (*tc.DeliveryServiceNullableV13, int, error, error) {
```
rawlinp, line 874: This is kind of confusing to review since it seems like this was all added as new code, but I think things were just moved around.

For the update* functions, you should just rename `updateV30` to `updateV31`, and change it to take in `DeliveryServiceV31` as input and output. Then _within_ the function you should immediately cast it to `DeliveryServiceNullableV30`. At the end of the function, you cast it back to `DeliveryServiceV31` before returning it.

Then you can basically copy `updateV15` to `updateV30`, change it to take `DeliveryServiceV30` as input and output, then change it to read the new `max_request_header_bytes` from the DB before passing it to `updateV31`, where the real handling is done.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
srijeet0406: Yeah this is not a new change, it was just moved around.

lib/go-tc/deliveryservices.go
```diff
@@ -428,8 +445,8 @@ func (ds *DeliveryServiceNullable) validateTypeFields(tx *sql.Tx) error {
 					if *ds.RangeRequestHandling == 3 {
 						return validation.Validate(ds.RangeSliceBlockSize, validation.Required,
 							// Per Slice Plugin implementation
-							validation.Min(262144),   // 256KiB
-							validation.Max(33554432), // 32MiB
+							validation.Min(MinRangeSliceBlockSize), // 256KiB
```
rawlinp, line 448: We should do some basic validation for `MaxRequestHeaderBytes` as well -- like it should be greater than zero.

lib/go-tc/deliveryservices.go
```diff
@@ -32,6 +32,9 @@ import (
 */
 
 const DefaultRoutingName = "cdn"
+const DefaultMaxHeaderSize = 131072
```
rawlinp, line 35: we should rename this `DefaultMaxRequestHeaderBytes` to match the new name

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -239,91 +261,101 @@ func createV15(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, reqDS t
...
+	dsV31 := tc.DeliveryServiceV31{DeliveryServiceV30: ds}
+	res, status, userErr, sysErr := createV31(w, r, inf, dsV31)
+	if res != nil {
+		return &res.DeliveryServiceV30, status, userErr, sysErr
+	}
+	return nil, status, userErr, sysErr
+}
+
+// create creates the given ds in the database, and returns the DS with its id and other fields created on insert set. On error, the HTTP status code, user error, and system error are returned. The status code SHOULD NOT be used, if both errors are nil.
+func createV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, ds tc.DeliveryServiceV31) (*tc.DeliveryServiceV31, int, error, error) {
```
rawlinp, line 274: We should rename `ds` to `dsV13`, then we can use `ds` instead of `dsNullableV30` on L278. Essentially, we don't want to be distracted by the versioning stuff throughout the method. Also it cuts down on the number of lines we have to change when adding new versions.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -699,62 +759,301 @@ WHERE
 }
 
 func updateV15(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, reqDS *tc.DeliveryServiceNullableV15) (*tc.DeliveryServiceNullableV15, int, error, error) {
```
rawlinp, line 761: So this function should actually stay exactly the same as it was before. Basically, we want to rename `updateV30` to `updateV31`, change it to accept and return V31 structs, then copy `updateV15` to `updateV30`, change it to accept V30 structs, upgrade them to V31 by changing the query to only select the new V31 field from the DB, then pass it to `updateV31` for the actual handling.

It's kind of hard to draw in a review, but the functions are designed to chain together like so:
```
UpdateV12 -> updateV12
              |     ^
              V     |
UpdateV13 -> updateV13
              |     ^
              V     |
   ...          ...
UpdateV31 -> updateV31  <------(this latest handler does all the actual handling)
```
So each API version is like a link in a chain. For adding a new version, you just add a new link to the end of the chain, and you don't have to worry about earlier links in the chain (other than the one you're linking to). Does that make sense?

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -699,62 +759,301 @@ WHERE
...
-	res, status, userErr, sysErr := updateV30(w, r, inf, &dsV30)
+	res, status, userErr, sysErr := updateV31(w, r, inf, &dsV31)
 	if res != nil {
 		return &res.DeliveryServiceNullableV15, status, userErr, sysErr
 	}
 	return nil, status, userErr, sysErr
 }
 
-func updateV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, ds *tc.DeliveryServiceNullableV30) (*tc.DeliveryServiceNullableV30, int, error, error) {
+func updateV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, ds *tc.DeliveryServiceV31) (*tc.DeliveryServiceV31, int, error, error) {
```
rawlinp, line 800: similar to my comment for `createV31`, we should rename this parameter to `dsV31`, then use `ds` instead of `dsNullableV30` on L803

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -699,62 +759,301 @@ WHERE
...
+	}
+
+	if err := api.CreateChangeLogRawErr(api.ApiChange, "Updated dsNullableV30: "+*dsNullableV30.XMLID+" id: "+strconv.Itoa(*dsNullableV30.ID), user, tx); err != nil {
+		return nil, http.StatusInternalServerError, nil, errors.New("writing change log entry: " + err.Error())
+	}
+	ds = (*tc.DeliveryServiceV31)(&dsNullableV30)
+	return ds, http.StatusOK, nil, nil
+}
+
+func updateV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, ds *tc.DeliveryServiceV30) (*tc.DeliveryServiceV30, int, error, error) {
```
rawlinp, line 1030: Per my earlier comment above, this function should look just like `createV30`, except with the DB query that reads the existing value of the new field.

rawlinp (changes requested):

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -796,11 +880,6 @@ func updateV30(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, ds *tc.
 				return nil, status, userErr, sysErr
 			}
 		}
-
-		userErr, sysErr, status := dbhelpers.CheckOriginServerInCacheGroupTopology(tx, *ds.ID, *ds.Topology)
```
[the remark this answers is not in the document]
rawlinp: So I figured out why the test for this wasn't failing and am about to open a PR. This should be added back.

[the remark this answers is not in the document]
srijeet0406: Whoops sorry, I must have fat fingered a copy paste. fixed now.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -400,6 +400,22 @@ <h3>Current Value</h3>
                             </aside>
                         </div>
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(cacheConfig.maxRequestHeaderBytes), 'has-feedback': hasError(cacheConfig.maxRequestHeaderBytes)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="maxRequestHeaderBytes">Max Request Header Size<div class="helptooltip">
+                                <div class="helptext">
+                                    The maximum size of the request headers for this particular delivery service.
```
rawlinp, line 406: this should probably say "The maximum size (in bytes) of the ..." (as well as the tooltip on the other form)

rawlinp (changes requested):

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -400,6 +400,22 @@ <h3>Current Value</h3>
...
                     </div>
+                    <div class="form-group" ng-class="{'has-error': hasError(cacheConfig.maxRequestHeaderBytes), 'has-feedback': hasError(cacheConfig.maxRequestHeaderBytes)}">
+                        <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="maxRequestHeaderBytes">Max Request Header Bytes<div class="helptooltip">
+                                <div class="helptext">
+                                    The maximum size (in bytes) of the request headers for this particular delivery service.
+                                </div>
+                            </div>
+                        </label>
+                        <div class="col-md-10 col-sm-10 col-xs-12">
+                            <input id="maxRequestHeaderBytes" name="maxRequestHeaderBytes" type="number" class="form-control" ng-model="deliveryService.maxRequestHeaderBytes" step="1" min="1" value="131072">
```
mitchell852, line 411: the value attribute is not needed here. if this is a new ds, the value of this field will be derived from deliveryService.maxRequestHeaderBytes that was set in traffic_portal_properties.json. if this is an existing ds, it will show the ds's value.

mitchell852 (changes requested):
looks good. one thing however.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
srijeet0406: Made the change

mitchell852:
sorry if this was already discussed but was it considered using a value of 0 to mean "no max"? We have a precedent with that. See max dns answers and max origin connections.

srijeet0406:
@mitchell852 No, 0 is considered an invalid value. It wont let you create a DS with the max header bytes set to 0 in TP or in the backend TO.

mitchell852:
> @mitchell852 No, 0 is considered an invalid value. It wont let you create a DS with the max header bytes set to 0 in TP or in the backend TO.

honestly, i don't like 0 to mean "no max" but it's kinda weird that some properties on a ds, you can use 0 and others you cannot but i'll defer to you guys.

rawlinp:
Hmm... I suppose if 0 meant "no max", that would most likely be a lie, because by default ATS sets the max to `131072` (which is why we're also setting that as the default here).

srijeet0406:
> > @mitchell852 No, 0 is considered an invalid value. It wont let you create a DS with the max header bytes set to 0 in TP or in the backend TO.
> 
> honestly, i don't like 0 to mean "no max" but it's kinda weird that some properties on a ds, you can use 0 and others you cannot but i'll defer to you guys.

Yeah, 0 header size would just mean that you cannot add a header rewrite rule in there, which is not what we want. You will always have to specify a value > 0.

rawlinp (approved):
Code looks good. Automated tests passed. Manual testing worked. 👍

mitchell852:
@srijeet0406 - since you changed traffic_portal_properties.json, you'll need to make a change like this to account for the new field: https://github.com/apache/trafficcontrol/pull/5379/commits/a1e504c6468f679d077c1192821ab498b0899be5

mitchell852:
image

Think you could expand this to explain what leaving it blank/null will do?

srijeet0406:
> image
> 
> Think you could expand this to explain what leaving it blank/null will do? For example, if empty is there a) no max or b) the global max (if defined) kicks in?

It actually cannot be empty, the min value of 1 will always be needed.

rawlinp:
If you mean from the API perspective, it will sanitize a null value to the default.

mitchell852:
> If you mean from the API perspective, it will sanitize a null value to the default.

Oh i see. so maybe the tooltip should say something like "leaving this value empty will result in the default: 131072". By the way what's the reasoning behind the 131072?

mitchell852:
oh geez. i see...it can't be empty. i should actually read.

mitchell852:
so i can clear the value

image

but then the default comes back so the tooltip should probably say that.

mitchell852:
bug: I entered 2147483648 (1 higher than int range). i guess 2147483647 should be enforced as the max in UI/API

srijeet0406:
> bug: I entered 2147483648 (1 higher than int range). i guess 2147483647 should be enforced as the max in UI/API

Fixed in the latest commit

rawlinp:
> By the way what's the reasoning behind the 131072?
That's the default that ATS uses for this.

mitchell852:
@srijeet0406 - isn't this related to #5244? want to add that to the description? also, want to share this with/demo this to @dsouza93 before merging to see if he has any feedback?

mitchell852:
@srijeet0406 - sorry, this is not really related to your PR but can you make a tiny change for me? I found a small bug in table.serverCapabilityDeliveryServices.tpl.html. line 37 should be `<th>XML ID (Key)</th>` instead of `<th>Host</th>`. Think you can make that super tiny change?

srijeet0406:
@mitchell852 No problem at all, I've fixed it in the latest commit. Also, I did have a talk with Dylan and showed him a quick demo earlier. He seemed happy with the changes. So if you feel it looks good, I'd say we are good here :)

srijeet0406:
> @srijeet0406 - isn't this related to #5244? want to add that to the description? also, want to share this with/demo this to @dsouza93 before merging to see if he has any feedback? or I suppose it could be demo'd after merging but any changes would require a new PR. just let me know. this all looks good. i'm ready to merge on your request.

I've mentioned it in the "additional info" section of the PR

TaylorCFrey:
> @srijeet0406 - sorry, this is not really related to your PR but can you make a tiny change for me? I found a small bug in table.serverCapabilityDeliveryServices.tpl.html. line 37 should be `<th>XML ID (Key)</th>` instead of `<th>Host</th>`. Think you can make that super tiny change?

FWIW I think this particular sized bug is fine to squeeze into a PR like this, but I do think a quick GH issue would be nice to ensure there was some level of tracking. Not necessarily a separate PR, but if the issue was tracked and included via reference in this PR then we should be good to go. What do you think @mitchell852?

mitchell852:
> > @srijeet0406 - sorry, this is not really related to your PR but can you make a tiny change for me? I found a small bug in table.serverCapabilityDeliveryServices.tpl.html. line 37 should be `<th>XML ID (Key)</th>` instead of `<th>Host</th>`. Think you can make that super tiny change?
> 
> FWIW I think this particular sized bug is fine to squeeze into a PR like this, but I do think a quick GH issue would be nice to ensure there was some level of tracking. Not necessarily a separate PR, but if the issue was tracked and included via reference in this PR then we should be good to go. What do you think @mitchell852?

normally i'd agree but this is so tiny/cosmetic seems like a waste of time to create a new issue/PR and really tracking this level of change is probably unnecessary and provides almost no value.

traffic_ops/app/db/migrations/2020113000000000_add_max_request_header_size_delivery_service.sql
```diff
@@ -0,0 +1,22 @@
+/*
```
rawlinp, line 1: Another migration has been merged, so this date will need to be updated to be later than the recently merged one.

mitchell852 (approved):
all looks good to me!

PR #5350, opened 2020-12-03
Validate DS's ORG server's cachegroup with topology's cachegroup

rimashah25:

- Traffic Ops

Method 1:
1. Assign ORG servers to a DS
2. Assign a topology which has origin cachegroup as one of the nodes.
3. Now, remove origin cachegroup from a given DS's topology and it should not allow you to do it.

Method2:
1. Remove ORG server part of DS1. 
2. Assign Topo1 to DS2 (which has ORG servers). 
3. Go to DS1's, manage topo and remove ORG CG from topo1. It should still error since topo1 is part of DS2, which has ORG server assignment.

Also, run the api/v3 test cases.

- [x] This PR includes tests

## Additional Information

API route hasn't changed. Hence the documentation is not updated.


traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -150,6 +150,20 @@ func (topology *TOTopology) Validate() error {
...
 	rules["required capabilities"] = topology.validateDSRequiredCapabilities()
 
+	for _, cdnId := range dsCDNs {
+		err, dsId := dbhelpers.GetDSByCDNIdTopology(topology.ReqInfo.Tx.Tx, cdnId, topology.Name)
+		if err != nil {
+			return fmt.Errorf("failed to get a response from DB: %s", err)
+		}
+		for _, id := range dsId {
+			topoCachegroupNames := topology.getCachegroupNames()
+			userErr, sysErr, _ := dbhelpers.CheckOriginServerInCacheGroupTopology(topology.ReqInfo.Tx.Tx, id, topology.Name, topoCachegroupNames)
```
rawlinp, line 160: Ok, I think I might've been wrong about being able to reuse this same `CheckOriginServerInCacheGroupTopology` function for this validation. When we update a topology, we don't want to make N DB queries (where N is the number of delivery services assigned to the topology). Preferably, we would make a single DB query that gets all the DS data we need, then we can do the validation from there.

rawlinp (changes requested):

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     - Traffic Router: Added support for topology-based delivery services
     - Traffic Monitor: Added the ability to mark topology-based delivery services as available
     - CDN-in-a-Box: Add a second mid to CDN-in-a-Box, add topology `demo1-top`, and make the `demo1` delivery service topology-based
+    - Traffic Ops: Added validation to ensure assigned ORG server cachegroups are in the topology when updating a delivery service
+    - Traffic Ops: Added validation to ensure that the cachegroups of a delivery services' assigned ORG servers are present in the topology
```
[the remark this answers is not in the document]
rimashah25: done.

traffic_ops/testing/api/v3/topologies_test.go
```diff
@@ -240,6 +241,39 @@ func UpdateTestTopologies(t *testing.T) {
...
+	}
+	var p []int
+	newNodes := []tc.TopologyNode{{Id: 0, Cachegroup: "topology-edge-cg-01", Parents: p, LastUpdated: nil}}
+	if *remoteDS[0].Topology == resp.Name {
+		resp.Nodes = newNodes
+	}
+	_, _, err = TOSession.UpdateTopology(*remoteDS[0].Topology, *resp)
+	if err == nil {
+		t.Fatalf("shouldnot UPDATE topology:%v, but update was a success", *remoteDS[0].Topology)
+	}
```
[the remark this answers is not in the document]
rimashah25: done.

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: Done.

traffic_ops/testing/api/v3/topologies_test.go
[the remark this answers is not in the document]
rawlinp: This doesn't seem to be resolved yet

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1150,3 +1151,68 @@ func CheckOriginServerInCacheGroupTopology(tx *sql.Tx, dsID int, dsTopology stri
...
+	rows, err := tx.Query(q, pq.Array(dsIDs), tc.OriginTypeName)
+	if err != nil {
+		return nil, fmt.Errorf("querying deliveryservice origin server: %s", err), http.StatusInternalServerError
+	}
+	defer log.Close(rows, "error closing rows")
+	for rows.Next() {
+		if err := rows.Scan(&dsName, &serverName, &cacheGroupName); err != nil {
+			return nil, fmt.Errorf("querying deliveryservice origin server: %s", err), http.StatusInternalServerError
+		}
+		servers[cacheGroupName] = serverName
```
rawlinp, line 1199: We should also have a map of server ID to DS xml_id, so that we can include the DS xml_id in the error message on L1211.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1150,3 +1151,68 @@ func CheckOriginServerInCacheGroupTopology(tx *sql.Tx, dsID int, dsTopology stri
...
+		res[i] = int(id)
+	}
+	return err, res
+}
+
+// CheckTopologyOrgServerCGInDSCG checks if ORG server are part of DS. IF they are then the user is not allowed to remove the ORG servers from the associated DS's topology
+func CheckTopologyOrgServerCGInDSCG(tx *sql.Tx, dsIDs []int, dsTopology string, topologyCGNames []string) (error, error, int) {
+	// get servers and respective cachegroup name that have ORG type for evert delivery service
+	q := `
+		SELECT d.xml_id, s.host_name, c.name
```
rawlinp, line 1178: I just remembered that `host_name` is not unique (sad, I know). So we should basically replace the host_name with the server ID throughout this function.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1150,3 +1151,68 @@ func CheckOriginServerInCacheGroupTopology(tx *sql.Tx, dsID int, dsTopology stri
...
+	var offendingDSSerCG []string
+	// put slice values into map for Topology's validation
+	topoCacheGroupNames := make(map[string]string)
+	for _, currentCG := range topologyCGNames {
+		topoCacheGroupNames[currentCG] = ""
+	}
+	for cg, s := range servers {
+		_, currentTopoCGOk := topoCacheGroupNames[cg]
+		if !currentTopoCGOk {
+			offendingDSSerCG = append(offendingDSSerCG, fmt.Sprintf("%s (%s)", cg, s))
```
rawlinp, line 1211: With the server IDs change mentioned above and a map of server ID to DS xml_id, we should format the message like this:
```
ORG servers are assigned to delivery services that use this topology, and their cachegroups cannot be removed: my-cachegroup (server ID = x, delivery service = y)
```

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -150,6 +150,19 @@ func (topology *TOTopology) Validate() error {
 	rules["empty cachegroups"] = topology_validation.CheckForEmptyCacheGroups(topology.ReqInfo.Tx, cacheGroupIds, dsCDNs, false, nil)
 	rules["required capabilities"] = topology.validateDSRequiredCapabilities()
 
+	for _, cdnId := range dsCDNs {
+		err, dsId := dbhelpers.GetDSByCDNIdTopology(topology.ReqInfo.Tx.Tx, cdnId, topology.Name)
```
rawlinp, line 154: Instead of looping over each CDN and running the check for each CDN, we should just pass the `dsCDNs` slice into `CheckTopologyOrgServerCGInDSCG` and update its query to replace `WHERE ds.deliveryservice =ANY($1)` with `WHERE ds.cdn_id = ANY($1)`. We would also then have to add `INNER JOIN topology top ON ds.topology = top.name` and add `AND top.name = $3` to the `WHERE` clause. That would allow us to get all the data in a single query, and we could delete the `GetDSByCDNIdTopology` function.

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -150,6 +150,19 @@ func (topology *TOTopology) Validate() error {
...
+		if err != nil {
+			return fmt.Errorf("failed to get a response from DB: %s", err)
+		}
+		//Get current Topology-CG for the requested change.
+		topoCachegroupNames := topology.getCachegroupNames()
+		userErr, sysErr, _ := dbhelpers.CheckTopologyOrgServerCGInDSCG(topology.ReqInfo.Tx.Tx, dsId, topology.Name, topoCachegroupNames)
+		if userErr != nil || sysErr != nil {
+			return fmt.Errorf("%s", userErr.Error())
```
rawlinp, line 162: If `userErr != nil`, we should just `return userErr`. But if `sysErr != nil`, we should log the error then `return errors.New("unable to validate topology")`

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: Done.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rimashah25: :(

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: This `INNER JOIN topology top ON ds.topology = top.name` doesn't make any sense, because `topology` table doesn't have any column that would help with delivery service results

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rimashah25: sigh.. ok

traffic_ops/testing/api/v3/topologies_test.go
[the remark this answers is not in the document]
rimashah25: How? I tested it on my local machine.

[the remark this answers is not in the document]
rawlinp: The test isn't doing this:
> unassign the denver-mso-org-01 server from ds-top

traffic_ops/traffic_ops_golang/topology/topologies.go
rawlinp: Ah, it's hard to come up with good SQL queries in a review comment. You're right. The `INNER JOIN topology` would not be required, since you can add `AND d.topology = $3` to the `WHERE` clause instead.

traffic_ops/testing/api/v3/topologies_test.go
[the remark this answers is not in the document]
rimashah25: Found a tiny bug:
`Rima Shah: It seems the DS is ok with no servers assigned to it.. but the minute you assign one, you can’t take it away because of this condition in delete function: checkLastServer()
Rawlin Peters: ah, sounds like that checkLastServer function should ignore ORG servers.`
Fix for the bug in this PR.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rimashah25: Logic will be a map of serverID to a slice of xmlIDs.

traffic_ops/testing/api/v3/deliveryserviceservers_test.go
```diff
@@ -111,17 +111,17 @@ func TryToRemoveLastServerInDeliveryService(t *testing.T) {
 
 	_, _, err = TOSession.DeleteDeliveryServiceServer(*ds.ID, *server.ID)
 	if err == nil {
-		t.Error("Didn't get expected error trying to remove the only server assigned to a Delivery Service")
+		t.Logf("Didn't get expected error trying to remove the only server assigned to a Delivery Service")
```
rawlinp, line 114: Why did you swap these errors to logs and vice versa? I think the logic was correct before.

traffic_ops/testing/api/v3/topologies_test.go
```diff
@@ -240,6 +241,49 @@ func UpdateTestTopologies(t *testing.T) {
...
+	if *remoteDS[0].Topology == resp.Name {
+		resp.Nodes = newNodes
+	}
+	_, _, err = TOSession.UpdateTopology(*remoteDS[0].Topology, *resp)
+	if err == nil {
+		t.Fatalf("shouldnot UPDATE topology:%v to %v, but update was a success", *remoteDS[0].Topology, newNodes[0].Cachegroup)
+	}
+
+	// Remove org server assignment and reset DS back to as it was for further testing
+	serverResp, _, err := TOSession.GetDeliveryServiceServersWithHdr(nil)
```
rawlinp, line 280: I think we need to use `GetServersWithHdr` here, including `host_name = denver-mso-org-01` in the `url.Values`, then use that ID in `DeleteDeliveryServiceServer` below instead. When you use `GetDeliveryServiceServersWithHdr`, it returns _all_ deliveryservice-server assignments, and it's not guaranteed that the assignment you're trying to delete is the first in the response.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1150,3 +1151,52 @@ func CheckOriginServerInCacheGroupTopology(tx *sql.Tx, dsID int, dsTopology stri
...
+func CheckTopologyOrgServerCGInDSCG(tx *sql.Tx, cdnIds []int, dsTopology string, topologyCGNames []string) (error, error, int) {
+	// get servers and respective cachegroup name that have ORG type for evert delivery service
+	q := `
+		SELECT ARRAY_AGG(d.xml_id), s.id, c.name
+		FROM server s
+			INNER JOIN deliveryservice_server ds ON ds.server = s.id
+			INNER JOIN deliveryservice d ON d.id = ds.deliveryservice
+			INNER JOIN type t ON t.id = s.type
+			INNER JOIN cachegroup c ON c.id = s.cachegroup
+		WHERE d.cdn_id =ANY($1) AND t.name=$2 AND topology=$3
```
rawlinp, line 1165: instead of `topology` by itself we should include the table alias `d.topology`

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1150,3 +1151,52 @@ func CheckOriginServerInCacheGroupTopology(tx *sql.Tx, dsID int, dsTopology stri
...
+		GROUP BY s.id, c.name
+	`
+	serverId := ""
+	cacheGroupName := ""
+	dsNames := []string{}
+	serversCG := make(map[string]string)
+	serversDS := make(map[string]interface{})
```
rawlinp, line 1172: why `interface{}` instead of `[]string`?

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1150,3 +1151,52 @@ func CheckOriginServerInCacheGroupTopology(tx *sql.Tx, dsID int, dsTopology stri
...
+	var offendingDSSerCG []string
+	// put slice values into map for Topology's validation
+	topoCacheGroupNames := make(map[string]string)
+	for _, currentCG := range topologyCGNames {
+		topoCacheGroupNames[currentCG] = ""
+	}
+	for cg, s := range serversCG {
+		_, currentTopoCGOk := topoCacheGroupNames[cg]
+		if !currentTopoCGOk {
+			offendingDSSerCG = append(offendingDSSerCG, fmt.Sprintf("cachegroup=%s (serverID=%s, delivery_service=%s)", cg, s, serversDS[s]))
```
rawlinp, line 1195: since there could be multiple DSes, `delivery_service` should be plural

traffic_ops/traffic_ops_golang/deliveryservice/servers/delete.go
```diff
@@ -45,10 +45,11 @@ func DeleteDeprecated(w http.ResponseWriter, r *http.Request) {
 }
 
 const lastServerQuery = `
-SELECT COUNT(*) = 0
-FROM deliveryservice_server
-WHERE deliveryservice = $1
-	AND server <> $2
+	SELECT COUNT(*)
```
rawlinp, line 48: Why remove the `= 0`? The query is expected to return a boolean value not an actual count. As-is, does this actually scan successfully into a bool?

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -150,6 +150,17 @@ func (topology *TOTopology) Validate() error {
 	rules["empty cachegroups"] = topology_validation.CheckForEmptyCacheGroups(topology.ReqInfo.Tx, cacheGroupIds, dsCDNs, false, nil)
 	rules["required capabilities"] = topology.validateDSRequiredCapabilities()
 
+	//Get current Topology-CG for the requested change.
+	topoCachegroupNames := topology.getCachegroupNames()
+	userErr, sysErr, _ = dbhelpers.CheckTopologyOrgServerCGInDSCG(topology.ReqInfo.Tx.Tx, dsCDNs, topology.Name, topoCachegroupNames)
+	if userErr != nil {
+		return fmt.Errorf("%s", userErr.Error())
```
rawlinp, line 157: this can just be `return userErr`

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -150,6 +150,17 @@ func (topology *TOTopology) Validate() error {
...
+	}
+	if sysErr != nil {
+		fmt.Println("System Error", sysErr)
```
rawlinp, line 160: This should use `log.Errorf` instead and should include a more precise context (e.g. `log.Errorf("validating topology: %s", sysErr.Error())`

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rimashah25: we can for clarity.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rimashah25: since I am storing a list to map-key (ARRAY_AGG(d.xml_id))

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: fmt.println was supposed to be cleaned up. That is my mistake. And i'll add a better log line.

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: ok.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rawlinp: Does `[]string` not work? Generally `interface{}` should only be used where necessary.

traffic_ops/testing/api/v3/topologies_test.go
rimashah25: I see your point. Since my test had only one server assignment, both functions GetServersWithHdr() and GetDeliveryServiceServersWithHdr were getting the correct serverID. But if we ever add more servers then things might go wrong. So, i'll change it.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rimashah25: I didn't test bu why?

traffic_ops/testing/api/v3/deliveryserviceservers_test.go
rimashah25: Logic wasn't making sense to me but I have swapped them back to work on later.

traffic_ops/traffic_ops_golang/deliveryservice/servers/delete.go
rimashah25: `= 0` gives true even when it shouldn't. So I replaced it with` > 0`

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -150,6 +150,16 @@ func (topology *TOTopology) Validate() error {
 	rules["empty cachegroups"] = topology_validation.CheckForEmptyCacheGroups(topology.ReqInfo.Tx, cacheGroupIds, dsCDNs, false, nil)
 	rules["required capabilities"] = topology.validateDSRequiredCapabilities()
 
+	//Get current Topology-CG for the requested change.
+	topoCachegroupNames := topology.getCachegroupNames()
+	userErr, sysErr, _ = dbhelpers.CheckTopologyOrgServerCGInDSCG(topology.ReqInfo.Tx.Tx, dsCDNs, topology.Name, topoCachegroupNames)
+	if userErr != nil {
+		return fmt.Errorf("%s", userErr)
```
rawlinp, line 157: This can just be `return userErr`.

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -150,6 +150,16 @@ func (topology *TOTopology) Validate() error {
...
+	}
+	if sysErr != nil {
+		log.Errorf("error while validate topology: %s", sysErr.Error())
```
rawlinp, line 160: In addition to logging the sysErr, we should `return errors.New("unable to validate topology")`. We don't want to expose system errors to the user, so we return a generic error here instead (because this method only supports returning a single error currently).

traffic_ops/testing/api/v3/topologies_test.go
```diff
@@ -240,6 +241,52 @@ func UpdateTestTopologies(t *testing.T) {
...
+	// remove org server cachegroup
+	var p []int
+	newNodes := []tc.TopologyNode{{Id: 0, Cachegroup: "topology-edge-cg-01", Parents: p, LastUpdated: nil}}
+	if *remoteDS[0].Topology == resp.Name {
+		resp.Nodes = newNodes
+	}
+	_, _, err = TOSession.UpdateTopology(*remoteDS[0].Topology, *resp)
+	if err == nil {
+		t.Fatalf("shouldnot UPDATE topology:%v to %v, but update was a success", *remoteDS[0].Topology, newNodes[0].Cachegroup)
+	}
```
rawlinp, line 277: Because I ran into a test issue like this the other day, we should add an `else if` to check that the error message contains `"ORG servers are assigned to delivery services that use this topology, and their cachegroups cannot be removed"`.

We want to make sure we're testing _that_ validation specifically, and that the topology update would otherwise be successful if not for that specific validation.

rawlinp (changes requested):
Just a few small issues but otherwise looks good. I'll pull this down and test manually while those get fixed.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1150,3 +1151,52 @@ func CheckOriginServerInCacheGroupTopology(tx *sql.Tx, dsID int, dsTopology stri
...
+		topoCacheGroupNames[currentCG] = ""
+	}
+	for cg, s := range serversCG {
+		_, currentTopoCGOk := topoCacheGroupNames[cg]
+		if !currentTopoCGOk {
+			offendingDSSerCG = append(offendingDSSerCG, fmt.Sprintf("cachegroup=%s (serverID=%s, delivery_services=%s)", cg, s, serversDS[s]))
+		}
+	}
+	if len(offendingDSSerCG) > 0 {
+		return errors.New("ORG servers are assigned to delivery services that use this topology, and their cachegroups cannot be removed:" + strings.Join(offendingDSSerCG, ", ")), nil, http.StatusBadRequest
```
rawlinp, line 1199: there should be a space after `removed:`

rimashah25: Done.

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: ok.

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: Understood.

traffic_ops/testing/api/v3/topologies_test.go
```diff
@@ -274,6 +274,8 @@ func UpdateValidateTopologyORGServerCacheGroup(t *testing.T) {
 	_, _, err = TOSession.UpdateTopology(*remoteDS[0].Topology, *resp)
 	if err == nil {
 		t.Fatalf("shouldnot UPDATE topology:%v to %v, but update was a success", *remoteDS[0].Topology, newNodes[0].Cachegroup)
+	} else if !strings.Contains(err.Error(), "ORG servers are assigned to delivery services that use this topology, and their cachegroups cannot be removed:") {
+		t.Errorf("ORG servers are assigned to delivery services that use this topology, and their cachegroups cannot be removed")
```
rawlinp, line 278: Test failure output should be informative like `"expected: error messaging containing '...', actual: '...'"` just to help whoever ends up getting this test failure in the future

rimashah25: Done.

rawlinp (approved):

PR #5357, opened 2020-12-07
Removed extra newline in LE cert chain

### LINKED ISSUE #5331: Let's Encrypt certs have a blank line in the chain
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:

When using Let's Encrypt there is a blank line inserted in the certificate chain. This causes TR to not be able to use the certificate.

## Expected behavior:

There should not be a blank line in the cert chain

Create a cert using Let's Encrypt.

Version:
4.1.1

mattjackson220:

- [x] This PR fixes #5331  

This PR removes the extra new line in the LE cert chain

- Traffic Ops

Generate new certs with Let's Encrypt, verify that the extra line is removed, verify that the cert can be deployed to TR / edges and is functioning as before

small bug fix so no tests, docs, changelog updates

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/letsencryptcert.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/letsencryptcert.go
@@ -355,7 +355,10 @@ func GetLetsEncryptCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEn
 	}
 	keyPem := keyBuf.Bytes()
 
-	dsSSLKeys.Certificate = tc.DeliveryServiceSSLKeysCertificate{Crt: string(EncodePEMToLegacyPerlRiakFormat(certificates.Certificate)), Key: string(EncodePEMToLegacyPerlRiakFormat(keyPem)), CSR: ""}
+	// remove extra line if LE returns it
+	trimmedCert := bytes.ReplaceAll(certificates.Certificate, []byte("\n\n"), []byte("\n"))
+
+	dsSSLKeys.Certificate = tc.DeliveryServiceSSLKeysCertificate{Crt: string(EncodePEMToLegacyPerlRiakFormat(trimmedCert)), Key: string(EncodePEMToLegacyPerlRiakFormat(keyPem)), CSR: ""}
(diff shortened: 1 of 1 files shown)


traffic_ops/traffic_ops_golang/deliveryservice/letsencryptcert.go
```diff
@@ -355,7 +355,10 @@ func GetLetsEncryptCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEn
 	}
 	keyPem := keyBuf.Bytes()
 
-	dsSSLKeys.Certificate = tc.DeliveryServiceSSLKeysCertificate{Crt: string(EncodePEMToLegacyPerlRiakFormat(certificates.Certificate)), Key: string(EncodePEMToLegacyPerlRiakFormat(keyPem)), CSR: ""}
+	// remove extra line if LE returns it
+	trimmedCert := bytes.ReplaceAll(certificates.Certificate, []byte("\n\n"), []byte("\n"))
```
TaylorCFrey, line 359: Any reason to ever expect a Microsoft style carriage return that would also need to be filtered/removed?

Also, the `EncodePEMToLegacyPerlRiakFormat` (hilarious comments BTW) reintroduces newline characters. Any chance they could end up doubled up again? Or maybe we don't need newlines at all before being passed into `EncodePEMToLegacyPerlRiakFormat`?

mattjackson220: since this only interacts with Let's Encrypt i wouldnt expect it to ever be different unless they change on their side. so i dont think we need to account for those but i guess it also wouldnt hurt

I think `EncodePEMToLegacyPerlRiakFormat` only adds new line characters between every 76 chars and at the very end so that shouldnt impact this. this is fixing because the LE certs had a new line in the middle of the cert chain that was causing issues.  at least i havent seen any issues with it in testing

also i with i could take credit for the funny comments haha. not from me but i really appreciate them

TaylorCFrey: @mattjackson220 So I can see the (_obvious_) fix, but how do I go about testing this? Is there someway for me to generate a LE cert and verify the newlines before and after?

TaylorCFrey (approved):
Reviewed code, addressed my comments.
Did peer screen share for verification and validation as well.

PR #5358, opened 2020-12-07
Tm csv nilcheck

ezelkow1:

This fixes an issue where if a cache returns no values for its proc entries and TM was set to use CSV data, it could cause a crash. Now instead it will return an error and be logged in the TM interface (currently the json format happily accepts empty entries with no error)

- Traffic Monitor

Test against a cache that does not return any proc data in its stats to verify that it does not crash when using the "http_polling_format": "text/csv" option in the config.

- master
- 5.0

## Additional Information

One other thing to be aware of is that the current astats json parsing has no issues with this. It blindly accepts empty values (or no values) coming from the astats plugin on caches. So in these scenarios there is no warning/error that entries do not exist. So in this case the csv version actually provides some more useful information for health


zrhoffman (approved):
Looks good, this fixes the panic if any of these stats are missing.

zrhoffman:
@ocket8888 #5358 should land in 5.0.0

PR #5366, opened 2020-12-08
Show correct CDN ID in Changelog

### LINKED ISSUE #5195: Change log / audit log error when snapshotting a cdn
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Expected behavior:
The following change/audit log entry is created when a cdn is snapshotted:

```
CDN: cdn-name, ID: 0, ACTION: Snapshot of CRConfig and Monitor
```

Should be:

```
CDN: cdn-name, ID: {{cdn-id}}, ACTION: Snapshot of CRConfig and Monitor
```

PUT /api/version/snapshot and view the created change/audit log entry.

TaylorCFrey:
The Changelog contains informative data to help with review and audit of actions that have been taken. However the CDN ID that is recorded when a CDN snapshot is performed always reads "0". This PR addresses this bug by ensuring the correct ID is displayed in the Changelog.

- [X] This PR fixes #5195 

- Traffic Ops
- Traffic Portal

This bug fix addresses what should have been a previously functional changelog statement. There are no references to specific or direct entries in the changelog in documentation, especially since it's subject to change.

~While I like to think I'm the first to promote automated tests, I believe the effort to add unit tests here would outweigh and overshadow the bug fix itself. Instituting automated tests to validate a text string for a log entry would require a somewhat significant undertaking, mocking at least a dozen DB calls to fake a snapshot, potential refactors, accounting for both current and deprecated paths, etc., would lead to fragile tests to ensure a string that is subject to change is correct.~

**EDIT** I've since added a check in the tests for one of the existing unit tests to ensure the ID value is validated as well as the previous implementations checking for errors.

Manual Testing can be performed to verify the fix.
1. Through the Traffic Portal, select CDNs from the menu
2. Select a CDN from the table
    2a. If none exist, create a CDN with +
> Note the ID value in the URL (2 in this case): https://localhost:9090/#!/cdns/2
3. Click the button labeled "Diff CDN Config Snapshot"
4. Click the button labeled "Perform Snapshot"
5. In the popup alert, click "Yes" to perform snapshot
6. Verify the ID is correct in the Changelog entry (The URL will be something like: https://localhost:9090/#!/change-logs)
    6a. The entry should contain the correct ID as noted from above

CDN: cdn-name, ID: **2**, ACTION: Snapshot of CRConfig and Monitor
-------------------^

- master (9597ead3)

### DIFF
--- a/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers_test.go
+++ b/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers_test.go
@@ -381,14 +381,14 @@ func TestGetCDNIDFromName(t *testing.T) {
 				mock.ExpectQuery("SELECT").WillReturnRows(rows)
 			}
 			mock.ExpectCommit()
-			_, exists, err := GetCDNIDFromName(db.MustBegin().Tx, "testCdn")
-			if testCase.storageError != nil && err == nil {
+			id, exists, err := GetCDNIDFromName(db.MustBegin().Tx, "testCdn")
+			if testCase.storageError != nil && err == nil && id != 0 {
 				t.Errorf("Storage error expected: received no storage error")
 			}
-			if testCase.storageError == nil && err != nil {
+			if testCase.storageError == nil && err != nil && id != 0 {
 				t.Errorf("Storage error not expected: received storage error")
 			}
-			if testCase.found != exists {
+			if testCase.found != exists && id == 0 {
 				t.Errorf("Expected return exists: %t, actual %t", testCase.found, exists)
 			}
 		})
--- a/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers_test.go
+++ b/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers_test.go
@@ -389,6 +389,9 @@ func TestGetCDNIDFromName(t *testing.T) {
 			if testCase.storageError == nil && err != nil && id == expectedIDValue {
 				t.Errorf("Storage error not expected: received storage error")
 			}
+			if exists && testCase.storageError == nil && err == nil && id != expectedIDValue {
+				t.Errorf("Expected ID %d, but got %d", expectedIDValue, id)
+			}
 			if testCase.found != exists && id == 0 {
 				t.Errorf("Expected return exists: %t, actual %t", testCase.found, exists)
 			}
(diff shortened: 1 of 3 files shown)


ocket8888:
I'm gonna re-run the CiaB check, might've just timed out due to ephemeral conditions.

srijeet0406:
The code changes look good. I also tested and verified manually that we can see the CDN ID in the TP changelog msgs now.
The only change I'm requesting is adding some tests around this functionality. I know the actual method that writes the changelog doesn't return anything so its difficult to write tests for it. However, maybe we could add a test for the `GetCDNIDFromName` method, and make sure we hit a cse where we get back a valid CDN ID, and another case where we get an error instead.

srijeet0406:
On a closer look, it appears that we do have a test for the `GetCDNIDFromName` function, but its not testing the CDN ID in there. So maybe you could just add that part.

TaylorCFrey:
@srijeet0406 I've added checks for the ID value within the existing unit tests for `GetCDNIDFromName`.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers_test.go
```diff
@@ -381,14 +381,14 @@ func TestGetCDNIDFromName(t *testing.T) {
...
 			}
 			mock.ExpectCommit()
-			_, exists, err := GetCDNIDFromName(db.MustBegin().Tx, "testCdn")
-			if testCase.storageError != nil && err == nil {
+			id, exists, err := GetCDNIDFromName(db.MustBegin().Tx, "testCdn")
+			if testCase.storageError != nil && err == nil && id != 0 {
 				t.Errorf("Storage error expected: received no storage error")
 			}
-			if testCase.storageError == nil && err != nil {
+			if testCase.storageError == nil && err != nil && id != 0 {
```
srijeet0406, line 388: In this case and the one above it, you can actually check if the `id` value is `1` or not. It should be 1 for the cases where `found` is true, because you specify it in this line:
`mock.ExpectQuery("SELECT").WillReturnRows(rows)`

TaylorCFrey: Good call. I've since made the expected ID value explicit.

srijeet0406 (approved):
LGTM! 
Code looks good.
Made sure the unit tests/ api tests pass.
Did manual testing and made sure the correct cdn ID shows up in the changelog.
Good job! 👍

PR #5367, opened 2020-12-10
CDN Snapshots and Monitoring Payloads rework (blueprint)

ocket8888:

This PR adds a blueprint for reworking Snapshots and Monitoring payloads to only containing the information necessary for Traffic Router and Traffic Monitor to do their jobs, respectively. For details read the blueprint.

None.

Don't. It's just a blueprint.

- [x] Tests are unnecessary

### DIFF
--- a/blueprints/snapshots.and.monitoring.md
+++ b/blueprints/snapshots.and.monitoring.md
@@ -1,0 +1,828 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+# Snapshots and Monitoring Separation
+
+## Problem Description
+Traffic Monitor uses data partially from CDN Snapshots and partially from its
+own Monitoring payloads to determine what to monitor and how. This leads to
+race conditions and certain parts of it being out-of-sync with others as one
+data set is polled while another grows stale.
+
+## Proposed Change
+Traffic Router should receive Snapshots directly from Traffic Ops, while
+Traffic Monitor should only fetch data relevant to its operation through the
+Monitoring payloads.
+
+## Data Model Impact
+The new structures for the Monitoring payloads and Snapshots are stripped down
+to only what each service needs, as well as having some properties renamed to
+be more consistent with the rest of the API. The proposed structures for each
+are given below as TypeScript<sup>[1](#typescript)</sup> interfaces.
+
+### Monitoring Payloads
+```typescript
+interface IPAddress {
+	address: string;
+	gateway: string | null;
+	serviceAddress: boolean;
+}
+
+interface Interface {
+	ipAddresses: Array<IPAddress>;
+	maxBandwidth: number | null;//integer
+	monitor: boolean;
+	mtu: number;//integer
+	name: string;
+}
+
+interface CacheServer {
+	deliveryServices: Array<string>;
+	httpsPort: number;//integer
+	interfaces: Array<Interface>;
+	profile: string;
+	status: string;
+	// formerly "port"
+	tcpPort: number;//integer
+	type: "EDGE" | "MID";
+}
+
+interface Monitoring {
+	// This set should only contain MID_LOC and EDGE_LOC cache servers
+	cacheGroups: {
+		[cacheGroupName: string]: {
+			// formerly a top-level property of monitoring payloads named
+			// "trafficServers"
+			cacheServers: {
+				[hostName: string]: CacheServer;
+			};
+			// latitude and longitude used to be inside a "coordinates"
+			// property which only had those two properties
+			latitude: number;
+			longitude: number;
+		}
+	};
+	config: {
+		// formerly 'health.event-count'
+		eventCount: number;//integer
+		// formerly 'health.timepad'
+		healthPad: number;
+		// formerly 'health.polling.interval'
+		healthPollingInterval: number;
+		// formerly 'heartbeat.polling.interval'
+		heartbeatPollingInterval: number;
+		// formerly 'peers.polling.interval'
+		peerPollingInterval: number;
+		// formerly 'tm.polling.interval'
+		configPollingInterval: number;
+	};
+	deliveryServices: {
+		[xmlID: string]: {
+			// formerly totalTpsThreshold
+			globalMaxTps: number | null;
+			// formerly totalKbpsThreshold
+			globalMaxMbps: number | null;
+		}
+	};
+	// This set should only include Profiles for Edge-tier and Mid-tier Cache
+	// Servers
+	profiles: {
+		[profileName: string]: {
+			parameters: {
+				[parameterName: string]: string;
+			};
+			thresholds: {
+				[thresholdName: string]: {
+					value: number;
+					comparator: "<" | ">" | "<=" | ">=" | "=" | "!=";
+				}
+			}
+		};
+	};
+	topologies: {
+		[topologyName: string]: {
+			nodes: Array<string>;
+		};
+	};
+	trafficMonitors: {
+		[hostName: string]: {
+			fqdn: string;
+			ipv4Address: string | null;
+			ipv6Address: string | null;
+			port: number;
+			profile: string;
+			status: string;
+		};
+	}
+}
+```
+
+... and what follows is an example payload from the default CDN-in-a-Box data
+set.
+
+```json
+{ "response": {
+	"cacheGroups": {
+		"CDN_in_a_Box_Mid-01": {
+			"cacheServers": {
+				"mid-01": {
+					"deliveryServices": [],
+					"httpsPort": 443,
+					"interfaces": [{
+						"ipAddresses": [{
+							"address": "172.20.0.5",
+							"gateway": "172.20.0.1",
+							"serviceAddress": true
+						}],
+						"maxBandwidth": null,
+						"monitor": true,
+						"mtu": 1500,
+						"name": "eth0"
+					}],
+					"profile": "ATS_MID_TIER_CACHE",
+					"status": "REPORTED",
+					"tcpPort": 80,
+					"type": "MID"
+				}
+			},
+			"latitude": 38.897663,
+			"longitude": -77.036574
+		},
+		"CDN_in_a_Box_Mid-02": {
+			"cacheServers": {
+				"mid-02": {
+					"httpsPort": 443,
+					"interfaces": [{
+						"ipAddresses": [{
+							"address": "172.20.0.3",
+							"gateway": "172.20.0.1",
+							"serviceAddress": true
+						}],
+						"maxBandwidth": null,
+						"monitor": true,
+						"mtu": 1500,
+						"name": "eth0"
+					}],
+					"profile": "ATS_MID_TIER_CACHE",
+					"status": "REPORTED",
+					"tcpPort": 80,
+					"type": "MID"
+				}
+			},
+			"latitude": 38.897663,
+			"longitude": -77.036574
+		},
+		"CDN_in_a_Box_Edge": {
+			"cacheServers": {
+				"edge": {
+					"httpsPort": 443,
+					"interfaces": [{
+						"ipAddresses": [{
+							"address": "172.20.0.2",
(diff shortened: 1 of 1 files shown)


blueprints/snapshots.and.monitoring.md
```diff
@@ -0,0 +1,828 @@
...
+		// used to be named "match-type"
+		type: "HOST" | "PATH" | "HEADER"
+	}>;
+}
+
+interface BaseDeliveryService {
+	// used to be a string containing a boolean, instead of just a boolean.
+	anonymousBlockingEnabled: boolean;
+	// Formerly 'ttl' - now consistent with the DS property it represents.
+	ccrDNSTTL: null | number;//integer
```
limited, line 354: rename this to get rid of ccr?

blueprints/snapshots.and.monitoring.md
```diff
@@ -0,0 +1,828 @@
...
+	consistentHashQueryParams: Array<string>;
+	coverageZoneOnly: boolean;
```
limited, line 356: isn't this currently an int?

blueprints/snapshots.and.monitoring.md
```diff
@@ -0,0 +1,828 @@
...
+
+Traffic Monitor's parsing of monitoring payloads will also need to be updated
+to handle the new format - but as always, a previous API version must still be
+supported.
+
+## Traffic Router Impact
+Traffic Router will need to be updated to handle the new Snapshot format, as
+well as maintaining compatibility with the old format. It will also need to be
+updated to fetch Snapshots from Traffic Ops directly, rather than through
+Traffic Monitor.
```
limited, line 760: TR has the nice ability today to be able to retrieve a snapshot from any of the Traffic Monitors. How does replacing n TMs with 1 TO impact the resiliency of config update process?

blueprints/snapshots.and.monitoring.md
```diff
@@ -0,0 +1,828 @@
...
+"cacheGroup"/"locationID" property to an entry in a 'cacheGroups' map) will be
+done on the Traffic Ops side, which incurs a negligible cost to TO itself but
+saves potentially significant time reproducing those mappings manually on the
+Traffic Monitor and Traffic Router side. Overall the expected performance
+impact is positive.
+
+## Security Impact
+There should be no changes from a security standpoint.
+
+## Upgrade Impact
```
limited, line 800: Will TM/TR first try to request the v4 API and failing that retry with v3 APIs calls?

ocket8888: Yes. TM already does this (currently v3->v2). I don't know if TR already does but it should because all components are supposed to do that to maintain backward-compatibility.

blueprints/snapshots.and.monitoring.md
ocket8888: No. According to [the docs](https://traffic-control-cdn.readthedocs.io/en/latest/api/v3/cdns_name_snapshot_new.html):

> _"[`coverageZoneOnly` is] A string containing a boolean that tells whether or not this Delivery Service routes traffic based only on its Coverage Zone File_"

blueprints/snapshots.and.monitoring.md
ocket8888: I'd love to- but not until that can be done on Delivery Services themselves. This is the same name as the property on the DS and I think that consistency is more important than a good name, because it makes it very obvious where that data came from.

blueprints/snapshots.and.monitoring.md
```diff
@@ -0,0 +1,828 @@
...
+		healthPollingInterval: number;
+		// formerly 'heartbeat.polling.interval'
+		heartbeatPollingInterval: number;
+		// formerly 'peers.polling.interval'
+		peerPollingInterval: number;
+		// formerly 'tm.polling.interval'
+		configPollingInterval: number;
+	};
+	deliveryServices: {
+		[xmlID: string]: {
```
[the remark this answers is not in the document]
ocket8888: We could, but it'd be superfluous. Only Delivery Services which are actively routed appear in monitoring payloads.

blueprints/snapshots.and.monitoring.md
```diff
@@ -0,0 +1,828 @@
...
+The new structures for the Monitoring payloads and Snapshots are stripped down
+to only what each service needs, as well as having some properties renamed to
+be more consistent with the rest of the API. The proposed structures for each
+are given below as TypeScript<sup>[1](#typescript)</sup> interfaces.
+
+### Monitoring Payloads
+```typescript
+interface IPAddress {
+	address: string;
+	gateway: string | null;
```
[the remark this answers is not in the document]
ocket8888: It doesn't, it's only included to keep the structure of a server's interfaces consistent across the API.

blueprints/snapshots.and.monitoring.md
```diff
@@ -0,0 +1,828 @@
...
+Traffic Monitor will need to be overhauled; Snapshot dependency is ingrained
+fairly deep in it. It will need to build out its configuration and stats from
+only the monitoring payload. It will also need to keep serving the Snapshots,
+however, to support old versions of Traffic Router.
+
+Traffic Monitor's parsing of monitoring payloads will also need to be updated
+to handle the new format - but as always, a previous API version must still be
+supported.
+
+## Traffic Router Impact
```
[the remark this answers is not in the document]
ocket8888: > _"... is there now a new condition that TR may receive configuration for servers for which TM does not yet have any health data?"_

That's actually already possible. So, no, that _new_ condition is not being introduced.

limited:
Thanks for the quick responses. 

Overall, I think this is a great improvement. I've seen some strange cases where TRs dont always update from TMs and I think this will be in general more reliable. 

Also, selfishly, this opens up potential for Traffic Ops to select which snapshot version a particular TR receives, which can be used as part of a staged or canary deployment.

blueprints/snapshots.and.monitoring.md
```diff
@@ -0,0 +1,828 @@
...
+# Snapshots and Monitoring Separation
+
+## Problem Description
+Traffic Monitor uses data partially from CDN Snapshots and partially from its
+own Monitoring payloads to determine what to monitor and how. This leads to
+race conditions and certain parts of it being out-of-sync with others as one
+data set is polled while another grows stale.
+
+## Proposed Change
+Traffic Router should receive Snapshots directly from Traffic Ops, while
```
rob05c, line 28: > Traffic Router should receive Snapshots directly from Traffic Ops

The TO->TM->TR flow is very intentional, to make sure the Monitor is monitoring what the Router is serving.

Removing that flow path and making independent TO->TM and TO->TR flows will create more even worse race conditions than the one mentioned above.

What does TR do if it gets CRConfig changes before the Monitor? So health data doesn't exist for a cache? What if something important, like a Profile, is changed? So the thing being monitored, the server, thresholds, deliveryservices, anything, is different than what the Router needs to serve? Any and all data--routing, thresholding, services, anything--is now a potential race condition, and can lead to strange and very-difficult-to-debug errors, including outages. Worse, if either the Monitor or Router is having trouble getting the latest CRConfig for any reason, those inscrutable errors could be indefinite.

> Traffic Monitor uses data partially from CDN Snapshots and partially from its own Monitoring payloads to determine what to monitor and how. This leads to race conditions

This can be solved either put all Monitoring data in the CRConfig, or to add a "bulk fetch" method to TO to get both objects in one request. 

If necessary, the CRConfig Monitoring data TR doesn't need could be stripped before sending it to TR. But I suspect it's a tiny size increase, and not nearly worth the safety and correctness benefit of having the Monitor directly serve the raw CRConfig bytes it got from TO. It also gives us more flexibility in the future to make TR use that data somehow.

ocket8888: Looks like you missed the conversation we had about this in Slack. This was @rawlinp's response to more or less exactly what you just said:

>>  If we are getting CrConfig from TO and CrStates from TM, it is possible that a cache that has been set ADMIN_DOWN could still be health in CrStates.

> that race doesn't seem like a big issue, because once TM has processed the ADMIN_DOWN-ing of the cache, it will set it unhealthy in CrStates, but in the meantime TR will still be sending traffic to it. This is no different than TR waiting to get the latest CRConfig+health directly from TM -- it will still send traffic to the cache until TM has done its thing.

>>  It's also possible that a cache that has been moved to OFFLINE and should not be in the CrConfig is still in CrStates.

> This isn't really an issue either, because TR only looks up health state for caches that are in its crconfig. I was just in that code a month ago to fix the bug where if a cache went from admin_down to offline, TM removed it from the crstates, causing TR to treat it as available.On the flipside, for caches going from offline to reported/admin_down, if TR has the cache in its crconfig but not in crstates, TR should treat it as unavailable until it gets a valid state.

which I think is at least as well as I could've put it. Originally the plan was actually to have TM generate configuration for TR, but this convinced me and @dneuman64 that it was okay to let TR go directly to TO.

Frankly I don't really care one way or the other, though as @limited has pointed out the latter has some advantages that would be nice to take advantage of.

rob05c: The Server.Status is not the only potential race. As I said

> What if something important, like a Profile, is changed? So the thing being monitored, the server, thresholds, deliveryservices, anything, is different than what the Router needs to serve? Any and all data--routing, thresholding, services, anything--is now a potential race condition

We've discussed adding smarter health, a percentage instead of a boolean, better thresholding, and many other features. Every new feature like those is another potential outage-inducing race, if TM and TR are racing for the same data instead of passing it through a flow.

blueprints/snapshots.and.monitoring.md
rob05c: In terms of resiliency,
If TO fails, the TMs won't be able to get config either. So it's not really introducing a single point of failure for new config. That already exists.

But what it does do is introduce a point-of-failure which can make the config race condition permanent. That is, if TM gets a new config, and TO goes down before TR can get the same config (or vica versa), what was a temporary race can become a long-standing mismatch, potentially outage-inducing, until TO is back up.

Where with the TO->TM->TR flow, no race exists, so TR will always match TM's config. If TO goes down, it may not be the latest, but it won't be an outage-inducing mismatch.

blueprints/snapshots.and.monitoring.md
rawlinp: The only data both TM and TR need is which caches to monitor/route to. If those get out of sync, TR has contingencies in place to not route traffic to caches it has no health information about. Basically any cache it knows about via the CRConfig that _doesn't have health info yet_ is considered unavailable. And if TR gets health information about a cache it doesn't know about (because it's not in its CRConfig), it ignores it. So although TM/TR get out of sync with each other until they've both processed the config, it's not a problem.

TM should only be telling TR which caches are healthy or not. I don't see how changing profiles, thresholds, etc., is a potential outage-inducing race condition. Those would only affect how TM determines cache availability, which is all that TR really cares about from TM.

blueprints/snapshots.and.monitoring.md
rawlinp: Per my comment on the main thread, I think TM and TR can safely get out of sync without causing an outage.

blueprints/snapshots.and.monitoring.md
rob05c: The Monitor also has Stat data, and does Delivery Service Thresholding, as well as serving stats to be aggregated via Traffic Stats to an upstream location, which can potentially be critical that it be correct. For example, if it were used for Billing.

What happens if that Delivery Service stat and thresholding is updated on the Router before the Monitor? Now we're routing to things the Monitor doesn't know about, and doesn't know to aggregate, and thus that traffic data is lost upstream, with potential financial or legal ramifications.

> TM should only be telling TR which caches are healthy or not. I don't see how changing profiles, thresholds, etc., is a potential outage-inducing race condition. Those would only affect how TM determines cache availability, which is all that TR really cares about from TM.

TM also serves Delivery Service availability to TR, as above. Moreover, as I said, we'd like to be able to add more complex health, such as DS-specific availability for each cache. These races make that either impossible, or worse, someone might not realize their subtlety and implement constantly racing and error-prone things.

rawlinp: TR wouldn't be getting any kind of stat or thresholding data from TO or TM. TM processes that data then tells TR simply "this cache is unavailable" or "this delivery service is unavailable". If we wanted to enhance that to have TM say "this cache is 50% available" for instance, then that still wouldn't be a race w/ TR. That's just TR waiting to get health data from TM as always.

blueprints/snapshots.and.monitoring.md
ocket8888: I'm just going to resolve this in favor of the longer discussion below

mitchell852:
@rawlinp - i believe we  have gone in a different but similar direction already regarding TR and TM configs. Do you think this should be closed?

rawlinp:
I think the compromise of https://github.com/apache/trafficcontrol/pull/6625 is that TM will get the data it needs from monitoring.json rather than relying on both monitoring.json _and_ CRConfig.json, but it will still synchronize the distribution of CRConfig to TR because that is one of the contentious points in this blueprint. If that's agreeable to @ocket8888 and others I think this could be closed.

PR #5369, opened 2020-12-10
Fixed /ping route having incorrect content-type

### LINKED ISSUE #2881: Some API Responses use wrong Content-Type
The only concrete example with which I can come up is `/api/1.3/servers/<hostname>/configfiles/ats`, which responds with `Content-Type: text/plain;charset=UTF-8`, which is the correct character encoding, but the data is a JSON object so this header should be `Content-Type: application/json;charset=UTF-8`. However, I seriously doubt that out of all of TO's hundreds of API endpoints, this is the only one to suffer from this header issue.

ocket8888:
- [x] This PR fixes #2881 by setting the `Content-Type` property to its appropriate value (`application/json`) for the `/ping` route.

- Traffic Ops

Using CiaB or a locally running Traffic Ops, send a GET request to `/ping` and note that it has the appropriate `Content-Type` header in the response to all API versions.

- master
- 5.0.0 (RC3)
- 4.1.1
- 4.0
- 3.1


shamrickus (changes requested):
This should have a CHANGELOG entry

rawlinp (approved):
works

PR #5372, opened 2020-12-11
Fix TR warning for caches without ipv4 addresses

rawlinp:
It is valid for a cache to not have an ipv4 address, so we should not
log a warning in that case.

- Traffic Router

In an environment where there are caches with no IPv4 addresses (i.e. they're IPv6-only), run TR _without_ this PR and you'll see warnings like:
```
WARN  2020-12-11T16:36:59.705 [New I/O worker #1] com.comcast.cdn.traffic_control.traffic_router.core.dns.ZoneManager - java.lang.IllegalArgumentException: unknown address family : null
```
_With_ this PR you shouldn't see those warnings.

- master
- 4.x
- 5.x

- [x] no new tests, but the unit and integration tests pass (except for `itRejectsCrConfigWithMissingCert` which has been failing for quite some time now)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - [#5195](https://github.com/apache/trafficcontrol/issues/5195) - Correctly show CDN ID in Changelog during Snap
+- Fixed Traffic Router logging unnecessary warnings for IPv6-only caches
 - [#5294](https://github.com/apache/trafficcontrol/issues/5294) - TP ag grid tables now properly persist column filters
     on page refresh.
 - [#5295](https://github.com/apache/trafficcontrol/issues/5295) - TP types/servers table now clears all filters instead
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - [#5195](https://github.com/apache/trafficcontrol/issues/5195) - Correctly show CDN ID in Changelog during Snap
+- Fixed Traffic Router logging unnecessary warnings for IPv6-only caches
```
TaylorCFrey, line 14: Should this have the `[#5372]` in the line?

rawlinp: We don't normally link the changelog entry to the PR, but we do sometimes link to the associated issue. In this case, there is no associated issue since I didn't create one 😄

TaylorCFrey: oh duh.

zrhoffman (approved):
* Confirmed that you get
  ```
  com.comcast.cdn.traffic_control.traffic_router.core.dns.ZoneManager - java.lang.IllegalArgumentException: unknown address family : null
  ```
  if you have ipv6-only caches

* Confirmed that this PR fixes that error.

Code looks good also

PR #5379, opened 2020-12-15
Removes invalid leaf node UI validation from topology builder

### LINKED ISSUE #5370: Topologies: Invalid error in TP when a mid or an org cache has ZERO primary children but 1+ secondary children
## I'm submitting a ...
-  bug report

-  Traffic Portal

## Current behavior:
Invalid error in TP when a mid or an org cache has ZERO primary children but 1+ secondary children:

image

## Expected behavior:
A mid or org cache group cannot be a "leaf" node in a topology and the following should make it not a leaf node:
- it has 1+ primary children and/or
- it has 1+ secondary children

create a topology like in the screen shot ^^

mitchell852:
In a topology, leaf nodes are defined as a cache group with zero primary OR secondary cache group children. MID_LOC and ORG_LOC cache groups cannot be leaf nodes in a topology. Only EDGE_LOC cachegroups can be a leaf node. The UI was not taking into consideration secondary children when identifying a leaf node so it would block you from creating a valid topology (a topology that had a MID_LOC or ORG_LOC with ONLY secondary children). This leaf node validation has been removed from the UI and is delegated to the API instead.

There are no additional UI tests or documentation attached to this PR as:
a. there is no real documentation at this point for topologies
b. the existing UI topology tests are not this granular

- [x] This PR fixes #5370 

- Traffic Ops
- Traffic Portal

1. Create a topology with 2 mid cache groups. assign edges directly to one mid but not the other. save. you shoud get an api validation error regarding leaf node error.
2. set the secondary parent of the edges to the empty mid cache group. save. no error.

- master (df92555)
- 5.0.0 (RC4)

### DIFF
--- a/infrastructure/ansible/roles/traffic_portal/defaults/main.yml
+++ b/infrastructure/ansible/roles/traffic_portal/defaults/main.yml
@@ -207,6 +207,9 @@ tp_default_properties_template:
           can the charts be found? xml id will be appended to the provided url.
         show: false
         baseUrl: "{{ tp_ts_base_url }}/dashboard/script/traffic_ops_server.js?which="
+    changeLogs:
+      _comment: Change log settings
+      days: 7
     customMenu:
       _comments: These are custom items you want to add to the menu. 'items' is an array
         of hashes where each hash has 'name' (the menu item name), 'embed' (true|false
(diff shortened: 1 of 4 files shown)


rawlinp (approved):
looks good, verification steps worked

PR #5383, opened 2020-12-18
Fix parent.config generation for MSO with required capabilities

rawlinp:
Server capabilities do not apply to ORG servers (origins), so don't
check for them.

- Traffic Ops ORT

Create an MSO delivery service (`multiSiteOrigin: true`) and a required capability. Assign 2 ORG servers and at least 1 EDGE server to the delivery service. The EDGE server must have the required capability. Make sure at least 1 MID in the CDN also has the required capability. On that MID, run ORT and verify that the line in `parent.config` for this delivery service contains the 2 ORG servers.

Next, create a topology with your edge -> mid -> origin cachegroups, and assign it to your MSO delivery service. On the same MID as before, run ORT and verify that the line in `parent.config` for this delivery service contains the 2 ORG servers.

- master
- 5.0.x
- 4.x

- [x] This PR includes tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - [#5195](https://github.com/apache/trafficcontrol/issues/5195) - Correctly show CDN ID in Changelog during Snap
 - Fixed Traffic Router logging unnecessary warnings for IPv6-only caches
 - Fixed parent.config generation for topology-based delivery services (inline comments not supported)
+- Fixed parent.config generation for MSO delivery services with required capabilities
 - [#5294](https://github.com/apache/trafficcontrol/issues/5294) - TP ag grid tables now properly persist column filters
     on page refresh.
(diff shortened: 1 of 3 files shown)


ocket8888:
re-running ciab check

zrhoffman (approved):
* `TestMakeParentDotConfigTopologiesMSOWithCapabilities` and `TestMakeParentDotConfigMSOWithCapabilities` fail without the fixes included in `parentdotconfig.go` (but pass with fixes included)
* Without the patch:
    - A MSO DS whose origin servers do not satisfy its required capabilities yields an empty `parent.config` (other than the `# DO NOT EDIT` comment) for servers in the child cachegroup of the ORG_LOC cachegroup.
    - `parent.config` is successfully populated if the origin servers do satisfy the DS's required capabilities (by inserting server server capabilities directly into the DB).
    - This is the case for topology-based and non-topology-based delivery services.

* With the patch, the `parent.config` has a `dest_domain=` line that includes the origin servers assigned to the MSO DS, for both topology-based and non-topology-based DSes.

The tests include redundant types, but that is a code style choice and is fine because it is more explicit that way.

PR #5164, opened 2020-10-16
Topologies support for assigning ORG servers to Delivery Services

zrhoffman:
## What does this PR (Pull Request) do?

- [x] This PR fixes both #5151 and #5157
- Excludes ORG-type servers from `GET /api/3.0/servers?dsId=#` for Topology-based Delivery Services unless the ORG server is assigned to that Delivery Service.
- Adds to TP the ability to assign ORG servers to topology-based delivery services

## Which Traffic Control components are affected by this PR?

- CDN in a Box
- Documentation
- Traffic Ops
- Traffic Portal

## What is the best way to verify this PR?

- Run the Traffic Ops API tests
- Run the Traffic Portal UI tests
- Try to assign an Origin server to a Topology-based Delivery Service in Traffic Portal
- Make sure that Origin server is not included in `GET /servers?dsId=#` unless it is assigned

## The following criteria are ALL met by this PR

- [x] This PR includes tests
- [x] This PR includes documentation
- [x] This PR includes an update to CHANGELOG.md
- [x] This PR includes any and all required license headers
- [x] This PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details) <div align="right"><sup><sup><details><summary></summary><div>closes #5151</div><div>closes #5157</div></summary></sup></sup></div>

PR #5386, opened 2020-12-18
Add a second origin server to CDN in a Box

zrhoffman:

This PR

- [x] is not related to any Issue 
- Renames the CDN in a Box `origin` server to `origin-01`
- Adds an `origin-02` server to CDN in a Box
- Enables the multi-site origin feature for CDN in a Box's `demo1` delivery service (was meant to be done in #5164)
- Fixes a bug where the Traffic Ops server objects for CiaB's origin server did not include an IPv6 address because the BusyBox implementation of the `route` command that Alpine uses by default does not include the `-6` flag. Fixed by installing the actual `route` command.

- CDN in a Box
- Documentation
- Traffic Ops (v1 API tests)
- CI tests

- Run CDN in a Box, make sure the readiness container exits successfully
- Build the documentation
- Check the result of the *CDN-in-a-Box CI* workflow

For the origin `route` command fix:

- master
- 5.0.0 (RC5)
- 4.1.1

- [x] This PR includes tests


ocket8888:
CDN-in-a-Box is getting pretty large... I, personally, think it ought to be kept to the absolute minimum to get a working CDN running. That was the original intention. It does support loading (somewhat) arbitrary data sets for testing purposes, but honestly I think it's high time we revisited the idea of splitting it into two. One setup for dev/demo and another for testing that can be more expansive. As a developer, I want CiaB to spin up fast and use the smallest amount of resources possible. Every new service we add impacts those two desires to some degree.

zrhoffman:
The origin servers run on a 26.1MB Alpine-based image. I can understand wanting to avoid adding big, bulky things to CDN in a Box, but that's not what this.

Also, servers queue for enrollment in parallel, so the additional startup time added by this PR is the time that a single POST request takes, a fraction of a second.

zrhoffman:
Having a second origin server in CiaB is not so important to me. PRed the bugfixes that this PR included in #5388, and closing this one.

PR #5389, opened 2020-12-18
Topology based DS's + reqd capabilities + ORG servers should show ORG servers, alongwith the other servers when queried for the linked servers

### LINKED ISSUE #5380: ds topology + required server capabilities + assigned ORG servers not working correctly
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
GET /api/3.0/servers?dsId={{id}} does not return ORG servers directly assigned to the DS if:
1. a topology is assigned to the DS and
2. one ore more ORG servers are directly assigned to the DS and
3. the ds requires server capabilities that the ORG server(s) does not have

In fact, I don't even think you are allowed to assign server capabilities to an ORG so you'd never be able to accomplish #3.

Note: this issue also seems to affect ORT/atstccfg as the applicable line for the ds in parent.config file is missing altogether when the only parent(s) are ORG servers that don't have the necessary server capability. ORG servers are exempt from server capability filtering and should ALWAYS show up for a ds in which they are directly assigned (server capability or not) but looks like this was addressed with PR #5383 

## Expected behavior:
If I recall correctly, server capabilities are not applicable to ORG servers, therefore, for a ds with a topology AND required server capabilities:
a. show all the servers in the topology in the same cdn as the ds
b. minus the ones w/o the ds's required server caps...except for ORG servers which server caps do not apply to (if assigned, show ORG servers regardless)

1. assign a topology to a DS
2. assign an ORG to the DS
3. create a server capability (https://tp.domain.tld/#!/server-capabilities) and add it to the DS as a required capability at https://tp.domain.tld/#!/delivery-services/999/required-server-capabilities
4. View the servers associated with the DS: https://tp.cdn.comcast.net/#!/delivery-services/999/servers - despite the assigned ORG server not having the DS's required server capability, it should be displayed.
5. add the server capability to one or more of the topology's servers. Now https://tp.cdn.comcast.net/#!/delivery-services/999/servers  should show the ORG plus the edges/mids that you assigned the server capability to.

srijeet0406:

- [x] This PR fixes #5380  

- Documentation
- Traffic Ops
- CI tests

Create a topology based DS with some required capabilities
Assign some ORG servers directly to this DS

Before the fix:
If you query for the servers of this DS using `GET /api/3.0/servers?dsId=<YOUR DS ID>`, you will only get the edges and mids that are:
1.) In the topology AND
2.) In the same CDN as the DS AND
3.) Have the server capabilities as specified in the DS
You will notice that the ORG servers DO NOT show up in the list.

After the fix:
If you perform the same steps, you will notice that the ORG servers also show up in the result now.

- master

- [x] This PR includes tests


traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -846,6 +875,15 @@ func getServers(h http.Header, params map[string]string, tx *sqlx.Tx, user *auth
...
 
 	where, orderBy, pagination, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(params, queryParamsToSQLCols)
+
+	// get the origin server count
+	if _, ok := params["dsId"]; ok {
+		serverCountOrigin, err = getServerCount(tx, serverCountQuery+originServerQuery, queryValues)
+		if err != nil {
+			return nil, 0, nil, fmt.Errorf("failed to get origin servers count: %v", err), http.StatusInternalServerError, nil
+		}
+	}
```
zrhoffman, line 885: Why is it necessary to count origin servers separately? We should avoid adding additional queries when we can.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -855,22 +893,12 @@ func getServers(h http.Header, params map[string]string, tx *sqlx.Tx, user *auth
...
-		if err = countRows.Scan(&serverCount); err != nil {
-			return nil, 0, nil, fmt.Errorf("failed to read servers count: %v", err), http.StatusInternalServerError, nil
-		}
-		rowsAffected++
-	}
-	if rowsAffected != 1 {
-		return nil, 0, nil, fmt.Errorf("incorrect rows returned for server count, want: 1 got: %v", rowsAffected), http.StatusInternalServerError, nil
-	}
 
+	serverCount += serverCountOrigin
```
zrhoffman, line 901: Tested this against CDN in a Box. In the case of no required capabilities, the original server count (6) is accurate, so the reported server count ends up being 7 instead of 6.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -767,6 +773,26 @@ JOIN type t ON s.type = t.id ` +
...
+	for countOrigRows.Next() {
+		if err = countOrigRows.Scan(&serverCount); err != nil {
+			return 0, fmt.Errorf("failed to read servers count: %v", err)
+		}
+		rowsAffected++
+	}
+	if rowsAffected != 1 {
+		return 0, fmt.Errorf("incorrect rows returned for server count, want: 1 got: %v", rowsAffected)
+	}
+	return serverCount, nil
```
zrhoffman, line 793: A scalar value is returned from getServerCount(). No need to defer or loop, this can compute the server count in SQL itself and return the result.

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/server/servers.go
srijeet0406: I've made the change so that we are accounting for the origin servers in the same query.

traffic_ops/traffic_ops_golang/server/servers.go
srijeet0406: Fixed the count now. I'm querying for the ORGs separately only when its specific to a DS that has reqd capabilities.

traffic_ops/traffic_ops_golang/server/servers.go
srijeet0406: Fixed to use a NamedStatement instead.

traffic_ops/testing/api/v3/deliveryservices_required_capabilities_test.go
```diff
@@ -45,9 +46,44 @@ func TestDeliveryServicesRequiredCapabilities(t *testing.T) {
...
+		t.Fatalf("no ID in the resulting delivery service")
+	}
+	dsID := *resp[0].ID
+	params := url.Values{}
+	_, _, err = TOSession.AssignServersToDeliveryService([]string{"denver-mso-org-01"}, "ds-top-req-cap2")
+	if err != nil {
+		t.Errorf("assigning ORG server to ds-top delivery service: %v", err.Error())
+	}
+	params.Add("dsId", strconv.Itoa(dsID))
+	params.Add("type", "ORG")
```
zrhoffman, line 71: `"ORG"` should be `tc.OriginTypeName`

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -399,6 +399,12 @@ RETURNING
 	upd_pending
 `
 
+const originServerQuery = `
+JOIN deliveryservice_server dsorg 
+ON dsorg.server = s.id 
+WHERE t.name='ORG'
```
zrhoffman, line 405: `ORG` should be `tc.OriginTypeName`

zrhoffman (changes requested):

zrhoffman (approved):

PR #5392, opened 2020-12-22
Updating a non existent DS should return a 404, instead of a 500

### LINKED ISSUE #5378: Trying to update a DS that doesn't exist returns an Internal Server Error
## I'm submitting a ...

-  bug report

-  CDN in a Box
-  Traffic Ops

## Current behavior:

Try making an API call to `PUT` (update) a delivery service that doesn't exist.
You will get back an internal server error (status code 500).
If you tail the logs, you'll see that this happens because the error is `sql: no rows in result set` (which is expected)
## Expected behavior:

This action should return a 404 (not found) if the DS was not found, or maybe a 400 (bad request).

Same as mentioned above
## Anything else:

I haven't checked if the other resources show a similar kind of behavior, maybe we need to check everything for this.

srijeet0406:

- [x] This PR fixes #5378  

- CDN in a Box
- Traffic Ops
- CI tests

Make sure all unit and API tests pass.

Try making an API call to PUT (update) a delivery service that doesn't exist.
You will get back a 404 instead of 500.

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
@@ -849,10 +849,12 @@ func updateV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 *
 	oldCDNName := ""
 	oldRoutingName := ""
 	errCode := http.StatusOK
+	var userErr error
+	var sysErr error
 	if dsType.HasSSLKeys() {
-		oldHostName, oldCDNName, oldRoutingName, err, errCode = getOldDetails(*ds.ID, tx)
-		if err != nil {
-			return nil, errCode, nil, errors.New("getting existing delivery service hostname: " + err.Error())
+		oldHostName, oldCDNName, oldRoutingName, userErr, sysErr, errCode = getOldDetails(*ds.ID, tx)
+		if userErr != nil || sysErr != nil {
+			return nil, errCode, userErr, sysErr
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -1179,7 +1180,7 @@ func selectMaxLastUpdatedQuery(where string) string {
 	select max(last_updated) as t from last_deleted l where l.table_name='deliveryservice') as res`
 }
 
-func getOldDetails(id int, tx *sql.Tx) (string, string, string, error) {
+func getOldDetails(id int, tx *sql.Tx) (string, string, string, error, int) {
```
zrhoffman, line 1183: Previously, the assumption was that any error returned from this function would be 500-level, so returning a single error would be sufficient. Now that errors can be 400-level or 500-level, this function should return both a user-level error and a system-level error.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -1194,25 +1195,28 @@ WHERE ds.id=$1
 	cdnDomain := ""
 	cdnName := ""
 	if err := tx.QueryRow(q, id).Scan(&xmlID, &protocol, &dsTypeStr, &routingName, &cdnDomain, &cdnName); err != nil {
-		return "", "", "", fmt.Errorf("querying delivery service %v host name: "+err.Error()+"\n", id)
+		if err == sql.ErrNoRows {
+			return "", "", "", fmt.Errorf("querying delivery service %v host name: "+err.Error(), id), http.StatusNotFound
+		}
```
zrhoffman, line 1200: Once this error is user-visible, the user should not see any database-related error like `no rows in result set`.

zrhoffman (changes requested):

zrhoffman (approved):
This all looks good.

PR #5401, opened 2021-01-05
Add atscfg ipallow topology children

rob05c:
Adds Topology children to atscfg ip_allow.config generation.

Labelling "bug" because without this, if a server is a child and happens to not be a child cachegroup of the parent server, and the IP happens to not coalesce, then it won't get added to the ip_allow of the Mid tier, resulting in the "mid" blocking the "edge." So it's effectively a bug, even though it's adding something, in that Topologies won't work right without this.

Includes tests.
No docs, no interface change.
No changelog, no interface change (might need a changelog 5.0 is released without this).

- Traffic Ops ORT

Configure Topologies, make sure your "edge" is not a child of the "mid"'s CacheGroup and its IP is unique enough to not coalesce with other edges. Without this PR, the IP shouldn't be added to the ip_allow.config for the "mid." With, it should.

- master

## Additional Information

### DIFF
--- a/lib/go-atscfg/atscfg.go
+++ b/lib/go-atscfg/atscfg.go
@@ -366,6 +366,36 @@ func makeTopologyNameMap(topologies []tc.Topology) map[TopologyName]tc.Topology 
 	return topoNames
 }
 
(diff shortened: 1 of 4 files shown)


rawlinp (approved):
Looks good, makes sense, tests pass

PR #5402, opened 2021-01-05
Cascade server delete to ip address and interfaces

### LINKED ISSUE #5364: TODB should cascade delete interfaces and IP addresses when deleting a server via SQL
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
When trying to delete a server via `DELETE FROM server WHERE ...`, you are restricted due to the `ON DELETE RESTRICT` FK clauses on the `interface` and `ip_address` tables. Additionally, deleting server interfaces is restricted via the `ON DELETE RESTRICT` FK clause on the `ip_address` table.

## Expected behavior:
If you delete a server via `DELETE FROM server WHERE ...`, it should cascade delete all the server's associated interfaces and IP addresses. If you delete an interface, it should cascade delete all the interface's associated IP addresses.

Basically these FK clauses should be `ON DELETE CASCADE` instead of `ON DELETE RESTRICT`.

For a server that has at least one interface, attempt to delete it via the SQL query: `DELETE FROM server WHERE id = ...`. For an interface with at least one IP address, attempt to delete it via the SQL query: `DELETE FROM interface WHERE server = ... AND name = ...`.

mattjackson220:

- [x] This PR fixes #5364  

This PR updates the ip_address and interface table foreign keys to cascade deletes so that when a server is deleted, the related ip addresses and interfaces are as well

- Traffic Ops

Verify that when a server object is deleted, the corresponding ip addresses and interfaces are also deleted.
Verify that GET, PUT, POST, and DELETE work as expected (behavior is unchanged) for servers, interfaces, and ip addresses

Bug fix so tests and docs do not need updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
     of just column filters
 - #2881 Some API endpoints have incorrect Content-Types
 - [#5311](https://github.com/apache/trafficcontrol/issues/5311) - Better TO log messages when failures calling TM CacheStats
+- [#5364](https://github.com/apache/trafficcontrol/issues/5364) - Cascade server deletes to delete corresponding IP addresses and interfaces
 - [#5390](https://github.com/apache/trafficcontrol/issues/5390) - Improve the way TO deals with delivery service server assignments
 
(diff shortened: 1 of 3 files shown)


shamrickus (approved):
LGTM

PR #5404, opened 2021-01-06
TO unable to unassign ORG server from delivery services if it's the last assigned

### LINKED ISSUE #5390: TO unable to unassign ORG server from delivery services if it's the last assigned
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
As of https://github.com/apache/trafficcontrol/pull/5328, TO prohibits unassigning all servers from an active delivery service, including `ORG` servers.

## Expected behavior:
This validation should be limited to server types that begin with `EDGE`, since the validation should be designed to prevent the accidental removal of all `EDGE`-type servers from an active delivery service.

Create an active delivery service and assign an `ORG` server to it. Try to remove the `ORG` server from the active delivery service.

## Anything else:
Once fixed, [this TODO](https://github.com/apache/trafficcontrol/pull/5350/files#diff-8bed268f92ae2156b9b90850085c66313c485e2700c74d509c6f3a52b60f7bf3R282) should be completed so that the test properly cleans up after itself.

srijeet0406:

- [x] This PR fixes #5390  

- Traffic Ops
- CI tests

Create a new active DS in TP
Try to assign a server to this DS that is not in the ONLINE/ REPORTED state -> this should fail
Now, try to assign an EDGE that is in ONLINE/ REPORTED state -> this should pass
Try deleting this EDGE server from TP and also from the backend API call -> both these actions should be prevented

Now create a new active DS in TP
Assign an ONLINE/ REPORTED ORG server to this DS from TP -> this should pass
Now, try to unassign this server from TP -> this should be prevented, because TP uses the `replace=true` mode, and you cannot unassign(or replace) all servers from an active DS
Try to delete this server from the backend call -> this should pass, since this is an ORG server

Note: to make a backend call to the DELETE api, use this endpoint:
`DELETE https://<TO HOST:PORT>/api/3.0/deliveryserviceserver/<DS ID>/<SERVER ID>`

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v3/topologies_test.go
+++ b/traffic_ops/testing/api/v3/topologies_test.go
@@ -279,15 +279,13 @@ func UpdateValidateTopologyORGServerCacheGroup(t *testing.T) {
 
 	}
 
-	//TODO: Need to fix the query in deliveryservice/servers/delete.go for DeleteDeliveryServiceServer() to work correctly.
-
-	// Remove org server assignment and reset DS back to as it was for further testing
-	//params.Set("hostName", "denver-mso-org-01")
-	//serverResp, _, err := TOSession.GetServersWithHdr(&params, nil)
-	//_, _, err = TOSession.DeleteDeliveryServiceServer(*remoteDS[0].ID, *serverResp.Response[0].ID)
-	//if err != nil {
-	//	t.Errorf("cannot delete assigned server from Delivery Services: %v", err)
-	//}
+	//Remove org server assignment and reset DS back to as it was for further testing
+	params.Set("hostName", "denver-mso-org-01")
+	serverResp, _, err := TOSession.GetServersWithHdr(&params, nil)
+	_, _, err = TOSession.DeleteDeliveryServiceServer(*remoteDS[0].ID, *serverResp.Response[0].ID)
+	if err != nil {
+		t.Errorf("cannot delete assigned server from Delivery Services: %v", err)
+	}
 }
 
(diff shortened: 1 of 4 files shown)


traffic_ops/testing/api/v3/topologies_test.go
```diff
@@ -279,15 +279,13 @@ func UpdateValidateTopologyORGServerCacheGroup(t *testing.T) {
...
-	//params.Set("hostName", "denver-mso-org-01")
-	//serverResp, _, err := TOSession.GetServersWithHdr(&params, nil)
-	//_, _, err = TOSession.DeleteDeliveryServiceServer(*remoteDS[0].ID, *serverResp.Response[0].ID)
-	//if err != nil {
-	//	t.Errorf("cannot delete assigned server from Delivery Services: %v", err)
-	//}
+	//Remove org server assignment and reset DS back to as it was for further testing
+	params.Set("hostName", "denver-mso-org-01")
+	serverResp, _, err := TOSession.GetServersWithHdr(&params, nil)
+	_, _, err = TOSession.DeleteDeliveryServiceServer(*remoteDS[0].ID, *serverResp.Response[0].ID)
```
ocket8888, line 285: This could segfault in two different ways: `serverResp.Response` could be nil so indexing with `0` would panic, and `serverResp.Response[0].ID` could be nil so dereferencing it would panic. Those conditions should be checked.

ocket8888 (approved):
Change makes sense, tests are passing

PR #5406, opened 2021-01-06
Add TO validation to ensure a parent.config weight parameter is a float

### LINKED ISSUE #3514: TO should validate that a parent.config weight parameter is a float.
A Parameter config file `parent.config` name `weight` must be a float, for ATS. Traffic Ops should validate this, when parameters are created.

This will prevent bad data being inserted, causing bad `parent.config` files to be generated, causing ATS to reject configs, refuse to start, or crash.

zrhoffman:

- [x] This PR fixes #3514 by adding validation to ensure a parent.config weight parameter is a float

- Documentation
- Traffic Ops

* Add a parameter
    - named `weight`
    - with config file `parent.config`
    - with a valid float value

* Make sure you cannot create or update a parameter to be
  - named `weight`
  - with config file `parent.config`
  - have a value that does not parse to a valid float

- [x] This PR includes tests


lib/go-tc/tovalidate/rules.go
```diff
@@ -15,9 +15,12 @@ package tovalidate
 import (
 	"errors"
 	"fmt"
+	validation "github.com/go-ozzo/ozzo-validation"
```
ocket8888, line 18: This should come after the standard library imports

zrhoffman: Reordered imports in bbd66b5111.

traffic_ops/testing/api/v3/withobjs_test.go
```diff
@@ -32,9 +32,15 @@ func WithObjs(t *testing.T, objs []TCObj, f func()) {
 			withFuncs[obj].Delete(t)
 		}
 	}()
-	for _, obj := range objs {
+	var obj TCObj
+	for _, obj = range objs {
 		withFuncs[obj].Create(t)
 	}
+
+	if _, exists := withNegativeFuncs[obj]; exists {
```
rawlinp, line 40: Do we really need a new negative fixtures json file and all this extra `WithObjs` stuff that goes with it? It seems like unnecessary complexity when negative test data has always been defined in the tests themselves.

traffic_ops/testing/api/v3/negative_fixtures_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+
+	"github.com/apache/trafficcontrol/lib/go-log"
+)
+
+// LoadFixtures ...
```
ocket8888, line 30: GoDoc should be a complete sentence starting with the documented entity and ending with punctuation.

traffic_ops/testing/api/v3/traffic_control_test.go
```diff
@@ -59,3 +59,11 @@ type TrafficControl struct {
 	Users                                             []tc.User                               `json:"users"`
 	InvalidationJobs                                  []tc.InvalidationJobInput               `json:"invalidationJobs"`
 }
+
+// NegativeFixtures - maps to the tc-negative-fixtures.json file
```
ocket8888, line 63: GoDoc should be a complete sentence starting with the documented entity and ending with punctuation.

traffic_ops/traffic_ops_golang/parameter/parameters.go
```diff
@@ -21,6 +21,7 @@ package parameter
 
 import (
 	"errors"
+	"github.com/apache/trafficcontrol/lib/go-atscfg"
```
ocket8888, line 24: These imports ought to be separate from the standard library

ocket8888 (commented):
The GoDoc comments can be ignored if you decide to do as rawlinp suggests and just define the "negative fixtures" on a per-test basis instead of an externally loaded file.

traffic_ops/testing/api/v3/traffic_control_test.go
zrhoffman: Made this into a sentence in 5bd6414471.

traffic_ops/testing/api/v3/negative_fixtures_test.go
zrhoffman: Added a sentence in 5bd6414471.

traffic_ops/traffic_ops_golang/parameter/parameters.go
zrhoffman: Separated that import from the standard library in 63eb27d310

traffic_ops/testing/api/v3/withobjs_test.go
zrhoffman: I can see how generalizing *negative tests* introduces tech debt if I don't move the negative test data to `tc-negative-fixtures.json` at the same time. Since doing that would be outside the scope of this PR:

* Moved the *invalid parameters* data to `CreateNegativeTestParameters()` in c364546143
* Removed the generalized *negative tests* stuff in 04b2fb10ae

It can be added back in a future PR that also moves existing negative test data to the JSON file that the PR would add.

ocket8888 (approved):
This looks sound to me and all the tests pass.

zrhoffman:
Rebased to resolve merge conflicts

PR #5414, opened 2021-01-07
Add atscfg parent.config topology and delivery service comment

rob05c:
Adds the #topology comment back to parent.config on its own line (it was removed because it was inline, which parent.config doesn't support).

Also adds the Delivery Service name to the same comment line, which should also be very helpful for production debugging.

Includes tests.
No docs, no interface change.

- Traffic Ops ORT

Run ORT, verify parent.config has comments with delivery service and topology names.

Not a bug.

## Additional Information

### DIFF
--- a/lib/go-atscfg/parentdotconfig_test.go
+++ b/lib/go-atscfg/parentdotconfig_test.go
@@ -28,7 +28,7 @@ import (
 )
 
 func TestMakeParentDotConfig(t *testing.T) {
-	hdr := "myHeaderComment"
+	hdr := ParentConfigOpts{AddComments: false, HdrComment: "myHeaderComment"}
 
 	ds0 := makeParentDS()
 	ds0Type := tc.DSTypeHTTP
@@ -132,7 +132,7 @@ func TestMakeParentDotConfig(t *testing.T) {
 	}
 	txt := cfg.Text
 
-	testComment(t, txt, hdr)
+	testComment(t, txt, hdr.HdrComment)
 
 	if !strings.Contains(txt, "dest_domain=ds0.example.net") {
 		t.Errorf("expected parent 'dest_domain=ds0.example.net', actual: '%v'", txt)
@@ -146,7 +146,7 @@ func TestMakeParentDotConfig(t *testing.T) {
 }
 
 func TestMakeParentDotConfigCapabilities(t *testing.T) {
-	hdr := "myHeaderComment"
+	hdr := ParentConfigOpts{AddComments: false, HdrComment: "myHeaderComment"}
 
 	ds0 := makeParentDS()
 	ds0Type := tc.DSTypeHTTP
@@ -254,7 +254,7 @@ func TestMakeParentDotConfigCapabilities(t *testing.T) {
 	}
 	txt := cfg.Text
 
-	testComment(t, txt, hdr)
+	testComment(t, txt, hdr.HdrComment)
 
 	lines := strings.Split(txt, "\n")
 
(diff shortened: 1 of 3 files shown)


lib/go-atscfg/parentdotconfig_test.go
```diff
@@ -2188,6 +2188,292 @@ func TestMakeParentDotConfigNoSecondaryMode(t *testing.T) {
...
+			Value:      tc.AlgorithmConsistentHash,
+			Profiles:   []byte(`["serverprofile"]`),
+		},
+		tc.Parameter{
+			Name:       ParentConfigParamQString,
+			ConfigFile: "parent.config",
+			Value:      "myQstringParam",
+			Profiles:   []byte(`["serverprofile"]`),
+		},
+	}
```
ocket8888, line 2363: nit but when you do `[]T{` you don't need to have `T{` for each element because it's trivially inferred from the collection type. i.e.

```go
stuff := []T{
	{
		someProp: 5,
	},
}
```
works just as well as
```go
stuff := []T{
	T{
		someProp: 5,
	},
}
```

you've got that in a few places, I'm not going to bother marking them all. Change if you want, or don't.

lib/go-atscfg/parentdotconfig_test.go
```diff
@@ -2188,6 +2188,292 @@ func TestMakeParentDotConfigNoSecondaryMode(t *testing.T) {
...
+	if !strings.Contains(txt, "qstring=myQStringHandlingParam") {
+		t.Errorf("expected qstring from param 'qstring=myQStringHandlingParam', actual: '%v'", txt)
+	}
+	if strings.Contains(txt, "secondary_mode") {
+		t.Errorf("expected no secondary_mode for DSes without ParentConfigParamSecondaryMode parameter, actual: '%v'", txt)
+	}
+	if !strings.Contains(txt, `# ds 'ds1' topology 't0'`) {
+		t.Errorf("expected comment with delivery service and topology, actual: '%v'", txt)
+	}
```
ocket8888, line 2476: I wanted to look at the whole output here so I did a `t.Log(txt)` and I see this:
```
dest_domain=ds0.example.net port=80 parent="mymid.mydomain.example.net:80|0.999;"  secondary_parent="mymid1.mydomain.example.net:80|0.999;" round_robin=consistent_hash go_direct=false qstring=myQStringHandlingParam
# ds 'ds1' topology ''
dest_domain=ds1.example.net port=80 parent="mymid.mydomain.example.net:80|0.999" secondary_parent="mymid1.mydomain.example.net:80|0.999" round_robin=consistent_hash go_direct=false qstring=myQStringHandlingParam
# ds 'ds1' topology 't0'
dest_domain=. parent="mymid.mydomain.example.net:80|0.999;" secondary_parent="mymid1.mydomain.example.net:80|0.999;" round_robin=consistent_hash go_direct=false qstring=myQstringParam
```

There's only one Topology here, so I think I'd only expect it to output one comment? the top line looks just like the third line except it has an extra `;` after each 0.999 and an extra ` ` before `secondary_parent` and `qstring`. Plus the top line comes from a Topology with no name? I think there's a bug here, but I'm not sure where it's coming from.

rob05c: I intentionally made the comment for non-topology lines `topology ''` so it'd be easier for log parsers to handle.

The trailing `;` is how Perl always did it, so I made Go do the same. I'm not sure whether ATS requires it, but it works.

The extra space could be removed, but I don't think this PR added it.

The top line missing the comment is the issue John mentioned. I'll look into it.

rob05c: Ah, it's the bottom line, not the top. Missing comment issue is that the comments are after the lines, not before.
I'll fix.

rob05c: Fixed

lib/go-atscfg/parentdotconfig_test.go
rob05c: Changed

ocket8888 (approved):
Seems to be working fine now that I know why sometimes Topologies are `''`.

PR #5416, opened 2021-01-08
Limit which Topology parent CacheGroup Servers to check for updates

### LINKED ISSUE #5413: Traffic Ops /api/3.0/servers/:host_name/update_status returns pending true for OFFLINE topology parents
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
/api/3.0/servers/:host_name/update_status (but not /api/2.0 it appears) returns parent_pending and parent_reval_pending true if Topology parents that aren't ONLINE or REPORTED are pending (which are likely offline, and will never update).

This breaks ORT, which will never update or reval those children.

It appears the problem is here https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/server/servers_update_status.go#L56

I think I see what's happening in the query.
The `parentservers` CTE is checking `WHERE pstatus.name != $1`
But the `server_topology_ancestors` doesn't, nor do either of the final 2 subqueries that use it,
`SELECT sta.upd_pending FROM server_topology_ancestors sta WHERE sta.base_server_id = s.id`.
I think it can be fixed by either adding the check to `parentservers`, or else the 2 places it's used:
```
SELECT sta.upd_pending FROM server_topology_ancestors sta WHERE sta.base_server_id = s.id and sta.status <> (select id from status where name = $1)
```

Also, the query only checks `<> OFFLINE`. I think it really needs to exclude everything not ONLINE or REPORTED. So, I think everything `<> OFFLINE` needs changed to `NOT IN (ONLINE, REPORTED)`.

## Expected behavior:

/api/3.0/servers/:host_name/update_status only returns parent_pending or parent_reval_pending true if ONLINE or REPORTED parents are upd_pending or reval_pending.

Set up topologies, set a Mid in the Topology to OFFLINE and queue updates on it. Queue a child, and then run ORT on that child, or request

zrhoffman:

This PR
- - [x] Fixes #5413 by only checking the update status for Topology parent CacheGroup Servers with `ONLINE` or `REPORTED` status.
- Restricts Topology parent CacheGroup types to check pending Server updates for to `EDGE_LOC` and `MID_LOC` so that if a server in an `ORG_LOC` cachegroup ever has pending updates or revalidation, caches in Topology child Cache Groups will not have `parent_pending` = `true` or `parent_reval_pending` = `true`.
- Makes API v1 and API v2 use the API v3 handler for GET `servers/{host_name}/update_status`

- Traffic Ops

- Run the API tests, make sure they pass

- master (36a1a847b7)
- 5.0.0 (RC6)

- [x] This PR includes tests
- [x] Bugfixes only, documentation is unnecessary


traffic_ops/traffic_ops_golang/server/servers_update_status.go
```diff
@@ -51,39 +53,44 @@ func getServerUpdateStatus(tx *sql.Tx, cfg *config.Config, hostName string) ([]t
...
+	WHERE s.host_name = $5
 UNION ALL
 /* Find all direct topology parent nodes tc of a given topology ancestor ta. */
 	SELECT tcp.parent, tc.cachegroup, ta.base_server_id
 	FROM topology_ancestors ta, topology_cachegroup_parents tcp
 	JOIN topology_cachegroup tc ON tcp.parent = tc.id
+	JOIN cachegroup c ON tc.cachegroup = c."name"
+	JOIN "type" t ON c."type" = t.id
 	WHERE ta.parent = tcp.child
+	AND t."name" = ANY($4::TEXT[])
```
rawlinp, line 76: I think we may want to support cachegroup types other than EDGE_LOC/MID_LOC, but still require them to have either the EDGE or MID prefix. E.g. EDGE_FOO_LOC or MID_BAR_LOC. So we should probably start just checking for the EDGE/MID prefix now, similar to how we check the EDGE/MID prefix for server types.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/server/servers_update_status.go
zrhoffman: Checking for Cache Group types that start with `EDGE` or `MID` in eaf05d2e54

traffic_ops/traffic_ops_golang/server/servers_update_status.go
```diff
@@ -127,7 +127,7 @@ ORDER BY s.id
 `
 
 	cacheStatusesToCheck := []tc.CacheStatus{tc.CacheStatusOnline, tc.CacheStatusReported}
-	cacheGroupTypes := []string{tc.CacheGroupEdgeTypeName, tc.CacheGroupMidTypeName}
+	cacheGroupTypes := []tc.CacheType{tc.CacheTypeEdge + "%", tc.CacheGroupMidTypeName + "%"}
```
rawlinp, line 130: `CacheTypeMid`?

traffic_ops/traffic_ops_golang/server/servers_update_status.go
```diff
@@ -97,29 +104,31 @@ SELECT
...
 LEFT JOIN parentservers ps ON ps.cachegroup = cg.parent_cachegroup_id
 	AND ps.cdn_id = s.cdn_id
-WHERE s.host_name = $4
+WHERE s.host_name = $5
 GROUP BY s.id, s.host_name, type.name, server_reval_pending, use_reval_pending.value, s.upd_pending, status.name
 ORDER BY s.id
 `
 
-	rows, err := tx.Query(selectQuery, tc.CacheStatusOffline, tc.UseRevalPendingParameterName, tc.GlobalConfigFileName, hostName)
+	cacheStatusesToCheck := []tc.CacheStatus{tc.CacheStatusOnline, tc.CacheStatusReported}
```
rawlinp, line 129: I think we may also want to consider `ADMIN_DOWN`, since that is the existing behavior today. Thinking about revalidations, we'd still want to have an `ADMIN_DOWN` mid revalidate before its children, otherwise when it comes back up it could serve invalidated content to the edges still.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/server/servers_update_status.go
zrhoffman: Considering `ADMIN_DOWN` in 03c22553ad

traffic_ops/traffic_ops_golang/server/servers_update_status.go
zrhoffman: Using `MidTypePrefix` in 88cbe8eb3d

rawlinp (approved):
Looks good, tested the query manually against an environment currently experiencing this issue, and it worked.

PR #5417, opened 2021-01-08
Fixed https status code when updating CDN for a DS for existing SSL keys

### LINKED ISSUE #5341: PUT /api/3.0/deliveryservices/{id} with changed cdn returns 403 forbidden
## I'm submitting a ...
-  bug report

-  Traffic Ops
-  Traffic Portal

## Current behavior:
Changing the CDN for a delivery service returns a 403 with 

```
{"alerts":[{"text":"Forbidden","level":"error"}]}
```

403s are for operations that fail role or tenancy check.

## New behavior:
This needs to be a 400 Bad Request instead with a message like:

```
{"alerts":[{"text":"Delivery service CDN is immutable","level":"error"}]}
```

Navigate to a DS in TP and change the CDN value. Save.

## Anything else:
If the intent is to make CDN immutable on a DS, the cdn field of an existing DS should be disabled as well to prevent the user from changing it.

rimashah25:

- [x] This PR fixes #5341 

- Traffic Ops
- Traffic Portal

Run TP end-to-end tests and TO API tests

- master(7a3506a)

## Additional Information

The PR doesn't have tests or documentation update as the logic hasn't changed.


mitchell852 (commented):
do u want to create an api test that attempts to change the cdn of a ds and ensure that a 400 is returned?

mitchell852 (commented):
also, in the delivery-services-spec.js file, can you add a check to each of the update tests to ensure the cdn dropdown is enabled?

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -1018,7 +1018,7 @@ func updateV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 *
 	}
 
 	if newDSType.HasSSLKeys() && (oldHostName != newHostName || oldCDNName != newCDNName || oldRoutingName != *ds.RoutingName) {
-		return nil, http.StatusForbidden, nil, errors.New("delivery service has ssl keys that cannot be automatically changed")
+		return nil, http.StatusBadRequest, errors.New("delivery service CDN is immutable"), errors.New("delivery service has ssl keys that cannot be automatically changed")
```
mitchell852, line 1021: is this supposed to return 2 errors? because i'm only seeing one.

mitchell852: maybe the error message needs to be something like - "delivery service has ssl keys that cannot be automatically changed therefore, cdn and routing name are immutable"

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -171,7 +171,7 @@ <h3>Current Value</h3>
                         </div>
                         </label>
                         <div class="col-md-10 col-sm-10 col-xs-12">
-                            <select id="cdn" name="cdn" class="form-control" ng-model="deliveryService.cdnId" ng-options="cdn.id as cdn.name for cdn in cdns" required>
+                            <select id="cdn" name="cdn" class="form-control" ng-model="deliveryService.cdnId" ng-options="cdn.id as cdn.name for cdn in cdns" required ng-disabled="!settings.isNew">
```
mitchell852, line 174: i guess you should leave this enabled as you can change the cdn on a DS IF the ds has no ssl keys...

rimashah25: The reason I added it because of this statement, "_If the intent is to make CDN immutable on a DS, the cdn field of an existing DS should be disabled as well to prevent the user from changing it._"

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
rimashah25: The error I added is user error. The system error in my opinion is not needed but I maybe wrong.

rimashah25:
> do u want to create an api test that attempts to change the cdn of a ds and ensure that a 400 is returned?

I'll add one. That is a good idea.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
mitchell852: where did u see that statement?

mitchell852: i _think_ cdn and routing name are only immutable if the ds has ssl keys, otherwise, make the change...

mitchell852: as a user, i would be annoyed if i created a delivery service and put it in the wrong cdn and was blocked from changing it..

rimashah25: Under **_Anything Else_** section in issue: https://github.com/apache/trafficcontrol/issues/5341

rimashah25: I'll change it. and make it immutable only if DS has ssl key.

mitchell852: yeah, i guess that was a bad assumption on my part. I _think_ if a ds has no ssl keys, then changing the cdn and/or routing name is perfectly fine.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
rimashah25: Done.

mitchell852:
> > do u want to create an api test that attempts to change the cdn of a ds and ensure that a 400 is returned?
> 
> I'll add one. That is a good idea.

actually, only a ds WITH ssl keys should return a 400 when cdn and/or routing name is changed right? not sure how you'll know if it has ssl keys as the api framework is not hooked into riak...

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.anyMap.tpl.html
```diff
@@ -156,7 +156,7 @@ <h3>Current Value</h3>
                         </div>
                         </label>
                         <div class="col-md-10 col-sm-10 col-xs-12">
-                            <select id="cdn" name="cdn" class="form-control" ng-model="deliveryService.cdnId" ng-options="cdn.id as cdn.name for cdn in cdns" required ng-disabled="!settings.isNew">
+                            <select id="cdn" name="cdn" class="form-control" ng-model="deliveryService.cdnId" ng-options="cdn.id as cdn.name for cdn in cdns" required>
```
mitchell852: do u want to disable this one?

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -746,7 +746,7 @@ <h3>Current Value</h3>
                         </label>
                         <div class="col-md-10 col-sm-10 col-xs-12">
                             <small class="input-warning" ng-show="!settings.isNew && routingConfig.routingName.$dirty">Warning: Changing the routing name may require SSL certificates to be updated.</small>
```
mitchell852, line 748: you can get rid of these warnings on all the routing name fields that should no longer be relevant

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -1017,8 +1025,8 @@ func updateV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 *
 		}
 	}
 
-	if newDSType.HasSSLKeys() && (oldHostName != newHostName || oldCDNName != newCDNName || oldRoutingName != *ds.RoutingName) {
-		return nil, http.StatusBadRequest, errors.New("delivery service CDN is immutable"), errors.New("delivery service has ssl keys that cannot be automatically changed")
+	if newDSType.HasSSLKeys() && exists && (oldHostName != newHostName || oldCDNName != newCDNName || oldRoutingName != *ds.RoutingName) {
```
mitchell852, line 1028: kind of confused why `newDSType.HasSSLKeys()` is not enough to determine if ssl keys exist or not for this ds...

mitchell852: to me this looks like "if has key and keys exist and....". seems redundant.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -851,11 +851,19 @@ func updateV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 *
...
 		oldHostName, oldCDNName, oldRoutingName, userErr, sysErr, errCode = getOldDetails(*ds.ID, tx)
 		if userErr != nil || sysErr != nil {
 			return nil, errCode, userErr, sysErr
 		}
+		exists, err = GetSSLVersion(*ds.XMLID, tx)
+		if err != nil {
+			return nil, http.StatusInternalServerError, nil, fmt.Errorf("querying delivery service failed: %s", err)
+		}
+		if exists {
+			return nil, http.StatusBadRequest, errors.New("delivery service has ssl keys that cannot be automatically changed, therefore CDN and routing name are immutable"), nil
```
mitchell852, line 865: i'm confused why this error is being called twice. here and on line 1029

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.anyMap.tpl.html
rimashah25: Nope, since SSLkeys are part of only 3 services: HTTP, DNS and Steering

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
rimashah25: Aha.. that makes sense.., i'll do that.

shamrickus (approved):
LGTM 
* GHA issues are transient and unrelated

PR #5431, opened 2021-01-12
Maintain file permissions in ATS config directory

jpappa200:

- [X] This PR is not related to any Issue. Although it has been requested that files in the config directory are only read/write by the owner octal 600. It was also requested that *records.config* be world readable. I've included an exemption hash. The key is the filename and the value is the required permission in octal format. 

- Traffic Ops ORT

Put *traffic_ops_ort.pl* on an offline cache, queued updates in TP, and ran it manually.
Checked that the files in */opt/trafficserver/etc/trafficserver* had the proper permissions.
Restarted trafficserver to insure there were no file permission issues on start/restart.

This is not a bug fix. 


jpappa200:
These changes were made in perl for file permissions. Per is deprecated and will be removed in the next major version. This PR is no longer needed.

PR #5435, opened 2021-01-13
Traffic Stats for Docker

smalenfant:

- CDN in a Box
- Documentation
- Grove
- Traffic Control Client 
- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Traffic Portal
- Traffic Router
- Traffic Stats
- Traffic Vault
- CI tests

### DIFF
--- a/traffic_monitor/Dockerfile
+++ b/traffic_monitor/Dockerfile
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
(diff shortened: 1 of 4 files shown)


zrhoffman:
We already have a (pre-CDN in a Box) attempt at a Docker environment for Apache Traffic Control at https://github.com/apache/trafficcontrol/tree/c29dba212d/infrastructure/docker (see: all subdirectories except for `build`).

All of them are pretty outdated, but if you're building a Docker environment for Traffic Stats, maybe you can replace the old Traffic Stats Docker image.

rob05c:
+1 for infrastructure/docker

Some of them are outdated, there aren't many people using them. I don't use Traffic Stats much. But I use the DB+TO+TP+TV almost daily. Just because they're faster than the CiaB, and more persistent.

I think updating or replacing infrastructure/docker/traffic_stats is the right answer here.

smalenfant:
I think I'm trying to do something a little different. This runs grafana + influxdb + traffic_stats. Traffic Stats should be it's own component. If somebody needs to run InfluxDB and Grafana, that should be separate (use docker-compose).

I'm also trying not to depend on the RPM build. It's just too many steps that are not needed. That's why the Dockerfile is in the "traffic_monitor" directory. I'm open to suggestions but really I want to be able to use `apache/traffic_stats:$RELEASE` from docker hub.

rob05c:
Ah, that makes sense. So yeah, it should be its own thing. 

I'd vote we be pretty open about accepting things like this. If Docker Hub is what you need, sure, we can add that. We can always remove it later if nobody is using it.

I'd still lean toward putting Docker stuff in `/infrastructure`

ocket8888:
> I'm also trying not to depend on the RPM build. It's just too many steps that are not needed.

I been saying that for years, godspeed

PR #5446, opened 2021-01-19
Specify nodejs version in TP spec

### LINKED ISSUE #5438: TP rpmspec does not specify nodejs version
## I'm submitting a ...

-  bug report

-  Traffic Portal

## Current behavior:
Any version of nodejs is accepted as a dependency to TP

## Expected behavior:
The nodejs version should either be pinned or restricted to prevent major updates to nodejs which might break TP. Currently the documentation says nodejs 12 or above is required.

shamrickus:

- [x] This PR fixes #5438

- Traffic Portal

Build the TP RPM and verify that it depends on Nodejs 12 instead of any nodejs version.

- master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - [#5378](https://github.com/apache/trafficcontrol/issues/5378) - Updating a non existent DS should return a 404, instead of a 500
 - [#5380](https://github.com/apache/trafficcontrol/issues/5380) - Show the correct servers (including ORGs) when a topology based DS with required capabilities + ORG servers is queried for the assigned servers
 - [#5195](https://github.com/apache/trafficcontrol/issues/5195) - Correctly show CDN ID in Changelog during Snap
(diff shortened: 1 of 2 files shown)


zrhoffman:
> Currently the documentation says nodejs 12 or above is required.

https://github.com/apache/trafficcontrol/blob/f2cb7e12d2b437480e1b5b16c2f6f6429fe8ce61/traffic_portal/build/traffic_portal.spec#L28

Does TP not run correctly on Node 13 or above? If so, that should be documented.

shamrickus:
There's no reason it shouldn't, but I think we want to pin our nodejs to a tested version.

zrhoffman:
> [...] I think we want to pin our nodejs to a tested version.

That's an environment decision. The NodeSource yum repos are separated by Node.js major version (example: `https://rpm.nodesource.com/pub_12.x/el/8/x86_64/nodesource-release-el8-1.noarch.rpm`), so there's no danger of upgrading an existing `nodejs` package to a new major version. Since the EPEL `nodejs` versions for CentOS 7 and CentOS 8 are both under 12 anyway, specifying minimum version only should be sufficient.

zrhoffman (approved):

mitchell852 (approved):
lgtm

PR #5450, opened 2021-01-21
When adding multiple interfaces on a server, a router host and router description should also be created per interface

### LINKED ISSUE #5316: When adding multiple interfaces on a server, a router host and router description shoudl also be created per interface

-  Traffic Ops
-  Traffic Portal

## Current behavior:
When creating a second, third etc interface in to 5 only a single router hostname and router port exists.

## New behavior:
adding more interfaces require more router(s) port(s) TP should reflect this

srijeet0406:

- [x] This PR fixes #5316  

- Documentation
- Traffic Ops
- Traffic Portal
- CI tests

Make sure all the unit tests and api tests pass. Note that there are new v4 api tests now.
Go to TP, and try to add a server. You will notice that the router host name and ports are now present on a per interface basis, rather than a per server basis. 

- master

- [x] This PR includes tests


docs/source/api/v4/servers_id.rst
```diff
@@ -0,0 +1,509 @@
...
+
+	:maxBandwidth: The maximum healthy bandwidth allowed for this interface. If bandwidth exceeds this limit, Traffic Monitors would have considered the entire server unhealthy - which includes *all* configured network interfaces. If this was ``null``, it has the meaning "no limit". It had no effect if ``monitor`` was not true for this interface.
+
+		.. seealso:: :ref:`health-proto`
+
+	:monitor: A boolean which describes whether or not this interface should have been monitored by Traffic Monitor for statistics and health consideration
+	:mtu:     The :abbr:`MTU (Maximum Transmission Unit)` of this interface. If it is ``null``, it may be assumed that the information was either not available or not applicable for this interface.
+	:name:    The name of the interface. It is the same as the network interface's device name on the server, e.g. ``eth0``.
+    :routerHostName: The human-readable name of the router responsible for reaching this server's interface.
+    :routerPort: The human-readable name of the port used by the router responsible for reaching this server's interface.
```
ocket8888, line 388: Same as above RE: indentation

docs/source/api/v4/servers_id.rst
```diff
@@ -0,0 +1,509 @@
...
+						"gateway": "::2",
+						"serviceAddress": true
+					}
+				],
+				"maxBandwidth": null,
+				"monitor": true,
+				"mtu": 1500,
+				"name": "bond0",
+		        "routerHostName": "",
+		        "routerPort": ""
```
ocket8888, line 504: same as above RE: formatting

lib/go-tc/deliveryservice_servers.go
```diff
@@ -105,14 +105,144 @@ type DSServerBase struct {
...
 // DSServerV11 contains the legacy format for a Delivery Service Server.
 type DSServerV11 struct {
 	DSServerBase
 	LegacyInterfaceDetails
 }
 
-// DSServer contains information for a Delivery Service Server.
-type DSServer struct {
+// DSServerV3 contains information for a Delivery Service Server.
+type DSServerV3 struct {
```
ocket8888, line 155: You can't get rid of DSServer because existing clients depend on it

lib/go-tc/servers.go
```diff
@@ -495,6 +745,48 @@ type CommonServerProperties struct {
 	XMPPPasswd       *string              `json:"xmppPasswd" db:"xmpp_passwd"`
 }
 
+// CommonServerPropertiesV40 is just the collection of properties which are
+// shared by all servers across API versions >= 4.0.
+type CommonServerPropertiesV40 struct {
```
ocket8888, line 750: I don't think this needs to be separate from the ServerV4 definition; instead I'd just remove the router fields from `CommonServerProperties` and add them to the older objects that still need them. The point of `CommonServerProperties` is just to not re-iterate properties shared by servers of all versions - but people are supposed to access those properties like `server.Prop` not `server.CommonServerProperties.Prop` so it's fine to change it as long as the overall Server structure isn't changed.

traffic_ops/app/db/migrations/2021011500000000_update_interfaces_multiple_routers.sql
```diff
@@ -0,0 +1,53 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE interface ADD COLUMN router_host_name text DEFAULT '';
+ALTER TABLE interface ADD COLUMN router_port text DEFAULT '';
```
ocket8888, line 19: I know that this is currently a text field, but isn't it actually a literal port? If so, seems like it should be an integer

traffic_ops/app/db/migrations/2021011500000000_update_interfaces_multiple_routers.sql
```diff
@@ -0,0 +1,53 @@
...
+
+UPDATE interface
+SET router_host_name = CASE
+WHEN server.router_host_name IS NULL THEN ''
+ELSE interface.router_host_name || server.router_host_name
+END
```
ocket8888, line 25: This looks like it'll work fine, but for what it's worth it looks like you wanted a `COALESCE` here. Since you just created the router_port column of an interface, they can't already have any text, so this should be equivalent to
```sql
UPDATE interface
SET router_host_name = CASE
WHEN server.router_host_name IS NULL THEN ''
ELSE server.router_host_name
END
FROM server
WHERE server = server.id;
```

which, in turn, is equivalent to

```sql
UPDATE interface
SET router_host_name = COALESCE(server.router_host_name, '')
FROM server
WHERE server = server.id;
```

... but also, can't `interface.router_host_name` itself be NULL? Then maybe you don't need to check at all. If the intention was to make it NOT NULL then I think you missed a constraint.

lib/go-tc/servers.go
```diff
@@ -71,6 +80,12 @@ type ServerDetailV30 struct {
 	ServerInterfaces *[]ServerInterfaceInfo `json:"interfaces"`
 }
 
+// ServerDetailV40 is the details for a server for API v4
+type ServerDetailV40 struct {
```
ocket8888, line 84: Why not just have this one type declaration instead of a "base" and this combination?

ocket8888 (commented):
Haven't quite finished looking through it, but this is what I noticed so far

traffic_ops/app/db/migrations/2021011500000000_update_interfaces_multiple_routers.sql
srijeet0406: Checked with Jeff Howell, and he suggested keeping it as a string. I changed the name to be `router_port_name` though.

docs/source/api/v4/servers_id.rst
```diff
@@ -49,21 +49,21 @@ Request Structure
...
-    :routerHostName: An optional string containing The human-readable name of the router responsible for reaching this server's interface.
-    :routerPort: An optional string containing The human-readable name of the port used by the router responsible for reaching this server's interface.
+	:monitor:           A boolean which describes whether or not this interface should be monitored by Traffic Monitor for statistics and health consideration.
+	:mtu:               The :abbr:`MTU (Maximum Transmission Unit)` of this interface. If it is ``null``, it may be assumed that the information is either not available or not applicable for this interface. This unsigned integer must not be less than 1280.
+	:name:              The name of the interface. No two interfaces of the same server may share a name. It is the same as the network interface's device name on the server, e.g. ``eth0``.
+    :routerHostName:    An optional string containing The human-readable name of the router responsible for reaching this server's interface.
+    :routerPort:        An optional string containing The human-readable name of the port used by the router responsible for reaching this server's interface.
```
ocket8888, line 66: these are still indented with spaces instead of tabs

ocket8888 (commented):
You need to change all of the labels in the API docs. `to-api-servers-id`, for example, is always the latest version of that endpoint. That way, when we change API versions, we don't need to comb through the documentation and find every reference to every endpoint and update them all. The downside is that anything that isn't the latest version needs to be updated when you do the copy into v4 - so that label would need to become `to-api-v3-servers-id`.

docs/source/api/v3/about.rst
```diff
@@ -13,7 +13,7 @@
 .. limitations under the License.
 ..
 
-.. _to-api-v3-about:
+.. _to-api-v3-v3-about:
```
ocket8888, line 16: This one apparently already had a `-v3-` segment, so now it's got two

docs/source/api/v3/cdns_dnsseckeys_generate.rst
```diff
@@ -13,7 +13,7 @@
 .. limitations under the License.
 ..
 
-.. _to-api-v3-cdns-dnsseckeys-generate:
+.. _to-api-v3-v3-cdns-dnsseckeys-generate:
```
ocket8888, line 16: same as above RE: double-segment.

docs/source/api/v3/deliveryservice_requests_id_assign.rst
```diff
@@ -13,7 +13,7 @@
 .. limitations under the License.
 ..
 
-.. _to-api-v3-deliveryservice_requests-id-assign:
+.. _to-api-v3-v3-deliveryservice_requests-id-assign:
```
ocket8888, line 16: Same as above RE: two`v3` segments

docs/source/api/v3/deliveryservice_requests_id_status.rst
```diff
@@ -13,7 +13,7 @@
 .. limitations under the License.
 ..
 
-.. _to-api-v3-deliveryservice_requests-id-status:
+.. _to-api-v3-v3-deliveryservice_requests-id-status:
```
ocket8888, line 16: Same as above RE: two`v3` segments

docs/source/api/v3/federations_all.rst
```diff
@@ -13,7 +13,7 @@
 .. limitations under the License.
 ..
 
-.. _to-api-v3-federations-all:
+.. _to-api-v3-v3-federations-all:
```
ocket8888, line 16: Same as above RE: two`v3` segments

docs/source/api/v3/ping.rst
```diff
@@ -13,7 +13,7 @@
 .. limitations under the License.
 ..
 
-.. _to-api-v3-ping:
+.. _to-api-v3-v3-ping:
```
ocket8888, line 16: Same as above RE: two`v3` segments

docs/source/api/v3/servers.rst
```diff
@@ -39,7 +39,7 @@ Request Structure
 	| cachegroupName | no       | Return only those servers within the :term:`Cache Group` that has this :ref:`cache-group-name`                    |
 	+----------------+----------+-------------------------------------------------------------------------------------------------------------------+
 	| dsId           | no       | Return only those servers assigned to the :term:`Delivery Service` identified by this integral, unique identifier.|
-	|                |          | If the Delivery Service has a :term:`Topology` assigned to it, the :ref:`to-api-servers` endpoint will return     |
+	|                |          | If the Delivery Service has a :term:`Topology` assigned to it, the :ref:`to-api-v3-servers` endpoint will return     |
```
ocket8888, line 42: Malformed table

docs/source/api/v3/steering.rst
```diff
@@ -13,7 +13,7 @@
 .. limitations under the License.
 ..
 
-.. _to-api-v3-steering:
+.. _to-api-v3-v3-steering:
```
ocket8888, line 16: Same as above RE: two`v3` segments

ocket8888 (commented):
Because those already had `-v3-` in them, the v4 documentation probably also has `-v3-` segments in the same endpoints, which should be removed.

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -220,7 +220,7 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
 		},
 		{
 			headerName: "Router Port Name",
-			field: "routerPortName",
+			field: "routerPort",
```
ocket8888, line 223: This field won't work anymore, since there's no field on the input data named `routerPort` (or `routerPortName`). I think we may just have to remove it.

srijeet0406: I'm a UI rookie, but isn't the field `interfaceName` working in the same way? It comes from the `interface` struct inside the `server` struct, as do `routerHostName` and `routerPortName`.

ocket8888: Yeah, I guess we don't actually need to remove it. It's fair enough to represent the router that's used for the IP addresses that we're adding to it.

But I still think the name is wrong. The property name that gets added from `toLegacyIPInfo` looks like `routerPortName`, so that'd need to be the name of the field to display it properly.

srijeet0406: yeah I think I changed it to `routerPortName` in my latest commit

ocket8888 (commented):
Everything looks fine, I still need to test it. There's just one thing; we have a bunch of broken references in the docs now - they're not in files you changed which makes them easy to miss. But I think the thing is, when you make a new API version you have to copy over ALL of the old docs into the new `vN` directory. That way links like e.g. `to-api-user-login-oauth` on the `oauth_login.rst` page don't break and always point to the latest API version.

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
```diff
@@ -268,6 +253,15 @@ <h6>Deprecated</h6>
...
+                        <small class="input-error" ng-show="hasPropertyError(serverForm[inf.name+'-router-hostname'], 'maxlength')">Too long</small>
+                        <small class="input-error" ng-show="hasPropertyError(serverForm[inf.name+'-router-hostname'], 'pattern')">No Spaces</small>
+                        <label for="{{inf.name}}-router-portname" ng-class="{'has-error': hasError(serverForm[inf.name+'-router-portname'])}">Router Port Name</label>
+                        <input id="{{inf.name}}-router-portname" ng-model="inf.routerPortName" ng-class="{'has-error': hasError(serverForm[inf.name+'-router-portname'])}" type="text" name="{{inf.name}}-router-portname" maxlength="256" pattern="\S*"/>
+                        <small class="input-error" ng-show="hasPropertyError(serverForm[inf.name+'-router-port'], 'maxlength')">Too long</small>
+                        <small class="input-error" ng-show="hasPropertyError(serverForm[inf.name+'-router-port'], 'pattern')">No Spaces</small>
```
ocket8888, line 263: These errors aren't showing up when their conditions occur - I think it's because you're checking `inf.name+'router-port'` and the input's `id` is actually `{{inf.name}}-router-portname`

ocket8888 (commented):
This seems to be working mostly very well, but there's a small UI issue

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
srijeet0406: aah yeah, copy paste error. Fixed now

ocket8888 (commented):
Everything seems to be working fine. There are, unfortunately, more docs problems, though. These ones didn't get caught earlier because they're in those extra v4 copies you just made:

- `cache_stats-get-request-example`
- `deliveryservice_stats-get-request-example`
- `response-structure`

All appear in the v3 and v4 docs, making them duplicates. They aren't prefixed with `to-api` which is probably why they didn't get caught in the find/replace.

Also, in `v3/deliveryserviceserver.rst` line 58 this label doesn't exist: `to-api-v3-response-structure`. I think it's supposed to be just `to-api-response-structure` - which is fine because it references the "Response Structure" section of the _main_ API docs page and therefore isn't versioned.

srijeet0406:
> Everything seems to be working fine. There are, unfortunately, more docs problems, though. These ones didn't get caught earlier because they're in those extra v4 copies you just made:
> 
> * `cache_stats-get-request-example`
> * `deliveryservice_stats-get-request-example`
> * `response-structure`
> 
> All appear in the v3 and v4 docs, making them duplicates. They aren't prefixed with `to-api` which is probably why they didn't get caught in the find/replace.
> 
> Also, in `v3/deliveryserviceserver.rst` line 58 this label doesn't exist: `to-api-v3-response-structure`. I think it's supposed to be just `to-api-response-structure` - which is fine because it references the "Response Structure" section of the _main_ API docs page and therefore isn't versioned.

Good eye, I've fixed them now.

ocket8888:
GHA was experiencing an outage last night; I restarted all the jobs. Assuming they pass like they have before, it'll be good to merge.

PR #5452, opened 2021-01-21
Not updating registration_sent value when updating user.

### LINKED ISSUE #5445: PUT /users/:id sets tm_user.registration_sent (date) to null
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
PUT /users/:id sets tm_user.registration_sent (date) to null

## Expected behavior:
PUT /users/:id should NOT touch tm_user.registration_sent (date) in any way. if it has a value, leave it. if it has no value, leave it. 

tm_user.registration_sent is set if a user is created via `POST users/register` and should never be nulled once it is set. it can, however, be updated if `POST users/register` is called again with the same email BUT it should never be nulled.

- Create a user via `POST users/register` and provide email, tenant and role
- check the database for the newly created user. see that registration_sent is set to current date/time
- note the id of the newly created user
- update the newly created user via `PUT /user/:id` and notice that registration_sent is set to null. this is wrong.

rimashah25:

- [x] This PR fixes #5445  

- Traffic Ops

Run API tests

master (9930731)

## Additional Information

No change in logic to update tests and documentation


traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -271,6 +271,8 @@ func (user *TOUser) Update(h http.Header) (error, error, int) {
 		return userErr, sysErr, errCode
 	}
 
+	user.RegistrationSent = &tc.TimeNoMod{Time: time.Now()}
```
mitchell852: confused. updating user (PUT /users/:id) should not update registration_sent value.

mitchell852: the ONLY way to change that value is by calling `POST users/register`

mitchell852: oh wait, nm. i guess i'm looking at outdated code :)

mitchell852: can u change the title of this PR? it confused me.

rimashah25: this is old code.

mitchell852 (approved):
lgtm

PR #5455, opened 2021-01-21
Change lib/go-log to always log UTC

rob05c:
Without this, log time format will be arbitrary dependent on the OS.

This came out of a conversation with our logs team. It's easier for log tools if the prefix format is consistent. We noticed it was different on dev laptops versus production CentOS servers. Because without this fix, the format is arbitrary depending on how the OS reports the timezone to Go.

ATC officially supports CentOS, which appears to report UTC. So this shouldn't change production servers. But it will guarantee the format is always UTC going forward.

Includes tests.
No docs, log format is not documented.
No changelog, log format is not documented, and this shouldn't change the supported CentOS behavior.

- library; error logs of all apps using lib/go-log (TO, TM, ORT)

Run tests. Print a log from lib/go-log on a machine reporting local time, like a laptop, before this fix, verify format is local, build again with this fix, verify time format is UTC.

Not a bug fix.

### DIFF
--- a/lib/go-log/log.go
+++ b/lib/go-log/log.go
@@ -76,14 +76,14 @@ func logf(logger *log.Logger, format string, v ...interface{}) {
 	if logger == nil {
 		return
 	}
-	logger.Output(stackFrame, time.Now().Format(timeFormat)+": "+fmt.Sprintf(format, v...))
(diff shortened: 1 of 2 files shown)


shamrickus (approved):
LGTM - Logs are now in UTC and tests pass

PR #5457, opened 2021-01-21
Fix changelog to account for things that actually went into 5.0

ocket8888:

Fixes the changelog having things that went into 5.0.0 being under 'unreleased'.

None

Make sure the changelog reflects the things that actually went into 5.0.0

- master

- [x] Tests are unnecessary
- [x] This PR is an update to CHANGELOG.md


rawlinp (approved):
LGTM

PR #5458, opened 2021-01-21
Allow max request header bytes to be 0

srijeet0406:

- CDN in a Box
- Traffic Ops
- Traffic Portal

Go to TP and try setting the maxRequestHeaderBytes propert of a DS to 0 -> this should be allowed.
Also, while creating a new DS, the default value of this field should be 0.
Make sure the unit tests and API tests pass.

- master

- [x] This PR does not include tests

### DIFF
--- a/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
+++ b/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
@@ -404,10 +404,8 @@ under the License.
                         <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="maxRequestHeaderBytes">Max Request Header Bytes<div class="helptooltip">
                                 <div class="helptext">
                                     The maximum size (in bytes) of the request headers for this particular delivery service. Clearing this will result in the default <code>(0)</code> being set for this DS.
-                                    Having <code>0</code> in this field will not apply any limit on the request headers.
-                                    <br>
-                                    <br>
-                                    <a href="https://docs.trafficserver.apache.org/en/8.0.x/admin-guide/files/records.config.en.html#proxy-config-http-request-header-max-size" target="_blank">See proxy.config.http.request_header_max_size</a>
+                                    Having <code>0</code> in this field will not apply any limit on the request headers, and will use the value configured in
+                                    <a href="https://docs.trafficserver.apache.org/en/8.0.x/admin-guide/files/records.config.en.html#proxy-config-http-request-header-max-size" target="_blank">proxy.config.http.request_header_max_size</a>
                                 </div>
                             </div>
                         </label>
--- a/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
+++ b/traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
@@ -403,13 +403,13 @@ under the License.
                     <div class="form-group" ng-class="{'has-error': hasError(cacheConfig.maxRequestHeaderBytes), 'has-feedback': hasError(cacheConfig.maxRequestHeaderBytes)}">
                         <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="maxRequestHeaderBytes">Max Request Header Bytes<div class="helptooltip">
                                 <div class="helptext">
-                                    The maximum size (in bytes) of the request headers for this particular delivery service. Clearing this will result in the default (131072) being set for this DS.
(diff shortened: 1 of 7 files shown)


rawlinp:
Should we bother updating the DB schema to make 0 the default there as well?

srijeet0406:
@rawlinp Yeah I was thinking about that, wasn't sure if the outcome is worth having yet another migration. Personally, I try to keep the migrations to a minimum, but I dont mind adding one for the default value if you'd like. IMO the main functionality is to "allow" 0s, which we can do with this PR. But as I said, I'm fine either way.

rawlinp:
Well, migrations are cheap, and it might be worth it for consistency's sake. It might be confusing to see the default as 0 in the API/UI but not in the DB.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -403,13 +403,13 @@ <h3>Current Value</h3>
                     <div class="form-group" ng-class="{'has-error': hasError(cacheConfig.maxRequestHeaderBytes), 'has-feedback': hasError(cacheConfig.maxRequestHeaderBytes)}">
                         <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="maxRequestHeaderBytes">Max Request Header Bytes<div class="helptooltip">
                                 <div class="helptext">
-                                    The maximum size (in bytes) of the request headers for this particular delivery service. Clearing this will result in the default (131072) being set for this DS.
+                                    The maximum size (in bytes) of the request headers for this particular delivery service. Clearing this will result in the default (0) being set for this DS.
```
mitchell852, line 406: You might want to tell the user what a value of 0 means exactly. Does it mean no max?

rawlinp: Good call, we might even want to reference the ATS parameter that this corresponds to: https://docs.trafficserver.apache.org/en/8.0.x/admin-guide/files/records.config.en.html#proxy-config-http-request-header-max-size.

mitchell852:
image

If you set it to zero on the DS, does it then use the value found in `proxy.config.http.request_header_max_size`?

srijeet0406:
> image
> 
> If you set it to zero on the DS, does it then use the value found in `proxy.config.http.request_header_max_size`?

Yes, you're right. I've made the doc change.

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.HTTP.tpl.html
```diff
@@ -403,13 +403,15 @@ <h3>Current Value</h3>
                     <div class="form-group" ng-class="{'has-error': hasError(cacheConfig.maxRequestHeaderBytes), 'has-feedback': hasError(cacheConfig.maxRequestHeaderBytes)}">
                         <label class="has-tooltip control-label col-md-2 col-sm-2 col-xs-12" for="maxRequestHeaderBytes">Max Request Header Bytes<div class="helptooltip">
                                 <div class="helptext">
-                                    The maximum size (in bytes) of the request headers for this particular delivery service. Clearing this will result in the default (131072) being set for this DS.
+                                    The maximum size (in bytes) of the request headers for this particular delivery service. Clearing this will result in the default <code>(0)</code> being set for this DS.
```
mitchell852, line 406: sorry to nit. How about this?

```The maximum size (in bytes) of the request headers for this particular delivery service. The default value is <code>0</code> which will limit the request header size to the value defined by the <a href="https://docs.trafficserver.apache.org/en/8.0.x/admin-guide/files/records.config.en.html#proxy-config-http-request-header-max-size" target="_blank">proxy.config.http.request_header_max_size</a> parameter.
```

srijeet0406: No worries, made that change.

mitchell852:
ok, it looks like your previous migration  - `2020121800000000_add_max_request_header_size_delivery_service.sql` that actually created the `max_request_header_bytes` field set the value to `131072` (the default) for all ds's.  So my question is - do you need an update query in `2021012200000000_max_request_header_bytes_default_zero.sql` to set them all to zero? I don't know the answer, just asking the question.

srijeet0406:
> ok, it looks like your previous migration - `2020121800000000_add_max_request_header_size_delivery_service.sql` that actually created the `max_request_header_bytes` field set the value to `131072` (the default) for all ds's. So my question is - do you need an update query in `2021012200000000_max_request_header_bytes_default_zero.sql` to set them all to zero? I don't know the answer, just asking the question.

Yes, you actually do need another migration to set the default to 0, for all new DS's.

mitchell852 (approved):
works as expected and resets all the values to 0 on migration. will merge once checks pass.

PR #5459, opened 2021-01-21
Code QL workflow

ocket8888:

This PR adds a new GHA workflow that scans the repository for common security vulnerabilities like e.g.

```
func Handle(w http.ResponseWriter, r *http.Request) {
    globalDBHandle.Exec("DELETE FROM " + r.Body.Read())
}
```

or something. It scans Go code, Javascript, and Python - and the tool has support for Java as well, but TR's build process couldn't be done automatically by the "Autobuild" step and I didn't bother trying to get it to work until I know if anyone's even interested in this.

None

Note the Action results on this PR. Should pass (because it did on my fork's `master` branch)

- [x] This PR is a test


zrhoffman:
Results are at [https://github.com/apache/trafficcontrol/security/code-scanning?query=ref:refs/pull/5459/merge](/apache/trafficcontrol/security/code-scanning?query=ref:refs/pull/5459/merge) but write access to the repo [is required](https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization#permission-requirements-for-security-features) in order to see them.

> The tools has support for Java as well, but TR's build process couldn't be done automatically by the "Autobuild" step and I didn't bother trying to get it to work until I know if anyone's even interested in this.

This is a cool thing, any chance we can get it working for Java also?

ocket8888:
There's always a chance. I just don't really know how. I took a glance at the build output and didn't see anything that pointed me to the obvious problem, but I'd bet just about anything that it's because of Tomcat not being present.

.github/workflows/codeql.analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  schedule:
+    - cron: '32 11 * * 0'
```
zrhoffman, line 26: A comment that describes this CRON expression would make it more maintainable

zrhoffman (changes requested):

zrhoffman:
> There's always a chance. I just don't really know how. I took a glance at the build output and didn't see anything that pointed me to the obvious problem, but I'd bet just about anything that it's because of Tomcat not being present.

Okay, it can always be added later.

.github/workflows/codeql.analysis.yml
ocket8888: that was actually auto-generated so I'll need to look it up (because I always need to, which is why this is a good idea) but I can do that.

.github/workflows/codeql.analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # CodeQL also supports Java, but our build system is too esoteric for them.
+        language: [ 'go', 'javascript', 'python' ]
```
zrhoffman, line 37: This GHA is set to run for commits to master and on PRs. As-is, all 3 languages are scanned each time. What do you think about breaking this workflow file into 3 separate workflow files so each language's workflow is only triggered if sources for that language are changed?

zrhoffman (changes requested):

.github/workflows/codeql.analysis.yml
ocket8888: So I think the reason it isn't is because when you run the cron task you want it to hit everything, since nothing in particular has changed, possibly. I could split it into 4 separate worflows to keep that behavior and also only run what's needed on PRs.

ocket8888: So I think the reason it isn't is because when you run the cron task you want it to hit everything, since nothing in particular has changed, possibly. I could split it into 4 separate worflows to keep that behavior and also only run what's needed on PRs.

zrhoffman: Yeah, with that explanation, 4 workflows makes the most sense since we still want the weekly job to hit all languages.

.github/workflows/codeql.go.yml
```diff
@@ -0,0 +1,47 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
zrhoffman, line 1: https://github.com/apache/trafficcontrol/actions/runs/514388427

> 1 issue was detected with this workflow: Please specify an on.push hook so that Code Scanning can compare pull requests against the state of the base branch.

.github/workflows/codeql.javascript.yml
```diff
@@ -0,0 +1,47 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
zrhoffman, line 1: https://github.com/apache/trafficcontrol/actions/runs/514388423

>  1 issue was detected with this workflow: Please specify an on.push hook so that Code Scanning can compare pull requests against the state of the base branch.

.github/workflows/codeql.python.yml
```diff
@@ -0,0 +1,47 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
zrhoffman, line 1: https://github.com/apache/trafficcontrol/actions/runs/514388430

> 1 issue was detected with this workflow: Please specify an on.push hook so that Code Scanning can compare pull requests against the state of the base branch.

> Unable to download and extract the tools needed for installing the python dependencies. You can call this action with 'setup-python-dependencies: false' to disable this process.

zrhoffman, line 1: https://github.com/apache/trafficcontrol/actions/runs/514388430

> Unable to download and extract the tools needed for installing the python dependencies. You can call this action with 'setup-python-dependencies: false' to disable this process.

zrhoffman (changes requested):
The new workflow files have warnings.

ocket8888:
So, regarding

> 1 issue was detected with this workflow: Please specify an on.push hook so that Code Scanning can compare pull requests against the state of the base branch.

if I add `on.push` hooks to all of those then we'll be double-testing every push to master with each of those workflows _and_ the main/scheduled one.

If it's comparing against the state of master, then maybe we didn't need to split these into separate workflows? Because then the runs for languages that have no changes won't actually be doing anything, if I'm understanding that correctly. The CodeQL web pages aren't helping me understand the action's behavior very much, because they're very focused on the parsing tool and don't mention this GHA at all, but I can probably contact GitHub support for a more thorough explanation

zrhoffman:
> So, regarding
> 
> > 1 issue was detected with this workflow: Please specify an on.push hook so that Code Scanning can compare pull requests against the state of the base branch.
> 
> if I add `on.push` hooks to all of those then we'll be double-testing every push to master with each of those workflows _and_ the main/scheduled one.

IMO we should just nix the scheduled workflow. The scheduled workflow alone is not adequate to provide a commit baseline for the per-language workflows that we want to run against PRs.

> If it's comparing against the state of master, then maybe we didn't need to split these into separate workflows? Because then the runs for languages that have no changes won't actually be doing anything, if I'm understanding that correctly.
Once the runs start. they won't get far before exiting if nothing has changed, sure, but in cases where Apache has surpassed the *concurrent jobs* limit, which happens daily, each of these languages adds an additional job to the queue that holds up other Apache projects' GitHub Action runs that were queued after that job, even if nothing has changed for that language.

So, it's still advantageous to split the workflow by language, since each workflow ignores different paths.

ocket8888:
Per github/codeql#5048 we should just get rid of the fourth workflow altogether - schedule _and_ push -, and run each of the language-specific ones on both push and pull, otherwise CodeQL won't run properly. I'll do that.

zrhoffman (approved):
This looks good. Even though [it says](https://github.com/apache/trafficcontrol/pull/5459/checks?check_run_id=1811405280)

> Warning: Code scanning cannot determine the alerts introduced or fixed by this pull request, because 3 analyses were not found.

That warning should go away once CodeQL runs directly against the master branch.

PR #5465, opened 2021-01-25
Update ORT to handle max_request_header_size

### LINKED ISSUE #5244: Limit request_header_max_size per DS

jpappa200:

- [X] This PR closes #5244 and will add a header rw to control max_haeder_request_size if the field in TP is not nil and greater than 0 . However, if the size added in TP exceeds the global setting, a header rw will be added with a comment that it is greater than the global setting and will be ignored.

- Traffic Ops ORT

Tested in CDN nightly lab. Updated ORT RPM on test cache. Update Max Request Header Size field in test delivery service and queue updates on an edge cache. 

This is not a bug fix. 


jpappa200:
This PR is obsolete due to changes in headerrewritedotconfig.go

PR #5467, opened 2021-01-26
T3c: DS parameter plugin args for slice, cache_range_requets, cachekey, url_sig, background_fetch

### LINKED ISSUE #5267: "cachekey.config" Parameter ConfigFile is undocumented
`## I'm submitting a ...
-  bug report

-  Documentation

## Current behavior:
Parameters with the ConfigFile `cachekey.config` are handled specially by configuration generation - but this behavior is undocumented.

## Expected behavior:
Special behavior needs to be documented.

Check the "Profiles and Parameters" overview section, notice no entry for `cachekey.config`.

traeak:

- [x] This PR partly allows 4759
- [x] This PR fixes #5267

This PR adds support for adding @pparam args for plugins using DS parameters:
- background_fetch.so - background_fetch.pparam
- cachekey.so - cachekey.pparam
- cache_range_requests.so - cache_range_requests.pparam
- slice.so - slice.pparam
- url_sig.so - url_sig.pparam

To specify multiple plugin arguments, multiple pparam args should be assigned to that Delivery Service.  Order of arguments should be considered random.

This PR expects "cachekey.config" DS parameters to at some point be converted over to "cachekey.pparam".  atstccfg will still load the cachekey.config DS params but will print a WARNING deprecated message.

Also note these pparam parameters will apply at all cdn layers, not just edge or mid (see below for tier specific proposal).

cachekey.config parameters will work as before.

To create different parameters (Name Config Value Secure) use:

@pparam=--someparam=true
```cachekey.pparam remap.config --someparam=true false```

@pparam=--separator=
```cachekey.pparam remap.config --separator= false```

@pparam=-o
```cachekey.pparam remap.config -o false```

For the future *if necessary* I would suggest extending this to something like:
* <plugin>.pparam.edge
* <plugin>.pparam.inner
or whatever the new convention is for the different cdn layers.

- Documentation
- Traffic Ops ORT

For both cache_range_requests and slice plugin create parameters and associate with a DS profile and select to use slice or cache_range_requests for range handling.

Example:
```
cache_range_requests.pparam --consider-ims remap.config false
cache_range_requests.pparam remap.config --key=value false
```

Run atstccfg and pull remap.config
With the above you would see:
```
@plugin=cache_range_requests.so @pparam=--consider-ims @pparam=--key=value
```

To enforce order one might do the following:
```
slice.pparam
remap.config
-i @pparam=\.mp4$
false
```

Which would result in remap.line section:
```
@plugin=slice.so @pparam=--blockbytes=<whatever default> @pparam=-i @pparam=\.mp4$
```

Also do similar with cachekey.pparam.  This should be tested to check for deprecation message of cachekey.config and also for proper loading of cachekey.config and cachekey.pparam.

Additional test case is to associate both cackekey.config and cachekey.pparam with a DS profile.  Ensure the cachekey.config entry items are ignored and only the cachekey.pparam items are used.

### CODE REVIEW

**Comment by mitchell852** [comment]
> ... @plugin=cache_range_requests.so @pparam=--consider-ims @pparam=--key=value ...

did you mean

```
@plugin=cache_range_requests.so @pparam=--consider-ims=false @pparam=--key=value ...
```

**Comment by mitchell852** [comment]
what if it's something like this instead? cache_range_requests.config makes me think that a cache_range_requests.config file is being created.

```
Example:
consider-ims cache_range_requests.plugin false
key cache_range_requests.plugin value false
```

File: cache-config/t3cutil/getdatacfg.go [hunk 1]
```diff
@@ -638,29 +642,55 @@ func GetConfigData(toClient *toreq.TOClient, disableProxy bool, cacheHostName st
...
+	cacheKeyConfigParamsF := func() error {
+		defer func(start time.Time) { log.Infof("cacheKeyConfigParamsF took %v\n", time.Since(start)) }(time.Now())
 		{
 			reqHdr := (http.Header)(nil)
 			if oldCfg != nil {
-				reqHdr = MakeReqHdr(oldCfg.MetaData.CacheKeyParams)
+				reqHdr = MakeReqHdr(oldCfg.MetaData.CacheKeyConfigParams)
 			}
-			params, reqInf, err := toClient.GetConfigFileParameters(atscfg.CacheKeyParameterConfigFile, reqHdr)
+			params, reqInf, err := toClient.GetConfigFileParameters("cachekey.config", reqHdr)
```
rob05c, line 653: Why change the constant to a literal?

traeak: should one be added for remap.config and parent.config to make it uniform?

rob05c (approved):
LGTM

rob05c:
This doesn't touch TM, I don't think it broke the TM tests, and this is an important feature we need, so I'm merging

cache-config/t3cutil/getdatacfg.go
rob05c: That would be ideal, but IMO not important enough to hold up merging

PR #5469, opened 2021-01-27
Remove test that checks system-dependent value

ocket8888:

This PR removes a test from Traffic Monitor that was checking that the machine running the tests was using more than 1MiB of memory - as that was fairly unreliable (particularly when running in GitHub Actions).

- Traffic Monitor
- CI tests (indirectly)

Notice that the tests all pass in GHA checks

- [x] This PR includes tests

### DIFF
--- a/traffic_monitor/datareq/datareq_test.go
+++ b/traffic_monitor/datareq/datareq_test.go
@@ -30,8 +30,7 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-util"
 	"github.com/apache/trafficcontrol/traffic_monitor/config"
 	"github.com/apache/trafficcontrol/traffic_monitor/peer"
-
-	"github.com/json-iterator/go"
+	jsoniter "github.com/json-iterator/go"
 )
 
 type AvailabilityType string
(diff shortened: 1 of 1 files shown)


zrhoffman:
@ezelkow1 Since you authored this test, do you have any thoughts on removing it?

ezelkow1:
> @ezelkow1 Since you authored this test, do you have any thoughts on removing it?

I did not author this test, I would ask @rob05c

zrhoffman:
Ah, gotcha. My mistake, I misremembered.

ocket8888:
Yeah, I requested his review, which should (hopefully) get his attention

rob05c:
It's been a while, but I think this was just added to make sure the code to get system data worked. I don't _think_ there was a bigger reason or issue than that.

I won't object to its removal.

Though I do disagree with the characterization of "system has more than a megabyte of memory" as "wildly varying system-dependent"

ocket8888:
Well the value is wildly varying, but whether or not it varies above 1MiB is not very wild at all, to be sure.

mattjackson220 (approved):
Looks good! code looks good and tests pass as expected

PR #5470, opened 2021-01-27
Adding atscfg overrides to insert the Release version in via string

ezelkow1:

The release version can be a user definable build option of ATS, instead of the actual ATS version number. So being able to set this as the via string can be useful for both debug to get a specific build value that was in use as well as obfuscation of the ATS version in use

This also adds atscfg and ort command line options for both the release version via string as well as a previous option
to disable per line comments in parent.config

- CDN in a Box
- Traffic Ops ORT

Run atstccfg with the `--via_string_release` option, or set that as an option in the crontab for ciab. Then check the via string output from a cache request in the headers and verify that it now shows the release value from the rpm instead of the ATS version

### DIFF
--- a/infrastructure/cdn-in-a-box/ort/traffic_ops_ort/__init__.py
+++ b/infrastructure/cdn-in-a-box/ort/traffic_ops_ort/__init__.py
@@ -472,6 +472,14 @@ def main() -> int:
 	                    help="Sets the timeout in milliseconds for requests made to Traffic Ops.",
 	                    type=int,
 	                    default=None)
+	parser.add_argument("--via_string_release",
+			    		help="set the ATS via string to the package release instead of version",
+			    		type=int,
+			    		default=0)
+	parser.add_argument("--disable-parent-config-comments",
+						help="Do not insert comments in parent.config files",
+						type=int,
+						default=0)
 	parser.add_argument("-v", "--version",
 	                    action="version",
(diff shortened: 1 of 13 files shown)


infrastructure/cdn-in-a-box/ort/traffic_ops_ort/__init__.py
```diff
@@ -472,6 +472,14 @@ def main() -> int:
 	                    help="Sets the timeout in milliseconds for requests made to Traffic Ops.",
 	                    type=int,
 	                    default=None)
+	parser.add_argument("--via_string_release",
```
ocket8888, line 475: Not that you need to bother updating ORT.py at all (bless you), but in most other places you call it `--via-string-release` - which I think is better, honestly - but here you have underscores instead of dashes.

ezelkow1: 👍 will do. Yea I used CIAB as the test environment for this, so to be able to pass the option around I had to add it everywhere, including ort.py :)

rob05c (approved):
Looks good! Code looks good. I tested, worked as expected. Ran ORT as before, didn't insert via strings. Ran with `--via_string_release`, ATS version was added to records.config, requested ATS, observed ATS yum version in via header in response.

PR #5471, opened 2021-01-27
Steering returns ordered results

### LINKED ISSUE #5461: TO /api/x/steering output should be sorted
## I'm submitting a ...
-  bug report

-  Traffic Ops
-  Traffic Router

## Current behavior:
Currently, the output of `/api/x/steering` is unsorted. This causes the response to be different each time it's requested, which causes TR to reload the steering data even though the steering data has not actually changed. This leads to a lot of unnecessary/misleading log output in TR.

## Expected behavior:
Responses from `/api/x/steering` should be sorted by `xml_id` so that they only change if the actual steering data has changed.

Request `/api/3.0/steering` multiple times. Observe that the ordering changes with each request.

mattjackson220:

- [x] This PR fixes #5461  

This PR sorts the results from the /steering endpoint to be sorted alphabetically by delivery service

- Traffic Ops

do a GET on the /steering endpoint on all API versions and make sure that the results are always ordered the same.
Verify that unit and API tests still pass

no test or doc updates since this is a small bug fix


TaylorCFrey (approved):
Pulled down proposed PR and ran locally. Confirmed the results were _not_ sorted previously. Then verified that the results are returned in ascending alphabetical order every time with all results still appearing.

LGTM

PR #5480, opened 2021-01-28
Added a check against cdn.id when updating DS with existing SSL keys

### LINKED ISSUE #5478: Given a DS with SSL Keys assigned, the CDN value shouldn't update.
## I'm submitting a ...

-  bug report

-  Traffic Ops

## Current behavior:

1. Create a DS
2. Generate SSL keys
3. Using PUT API (/api/3.0/deliveryservices/<dsid>), update DS's cdnName using cdn.id
4. It updates the DS.

## Expected behavior:

Given a DS with SSL Keys assigned, the CDN value shouldn't update. It should give a 400 error response.

rimashah25:

- [x] This PR fixes #5478 

- Traffic Ops

1. Create a DS
2. Generate SSL keys (If using TP, go to mange SSL keys in a DS and then click on generate SSK keys OR use SSL Keys generate API)
3. Edit DS using PUT API (/api/3.0/deliveryservices/<ds-id>) and change the cdn id to a different value.
4.  You should see an error as follows:

``` 
{
   "alerts":[{
          "text":"delivery service has ssl keys that cannot be automatically changed, therefore CDN and routing name are immutable",
          "level":"error"
 }]}
```

and a 400 HTTP status code

- master (456f7620)
 
- [x] This PR includes tests
- [x] have explained why documentation is unnecessary
- [x] This PR does not includes an update to CHANGELOG.md as it is not necessary

## Additional Information

API Tests already present. No documentation change.


traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices_test.go
```diff
@@ -70,6 +70,9 @@ func TestGetDetails(t *testing.T) {
 	if oldDetails.OldCdnName != "foo" {
 		t.Errorf("expected old cdn name to be foo, but got %v", oldDetails.OldCdnName)
 	}
+	if *oldDetails.OldCdnId != 1 {
```
srijeet0406, line 73: You need to add a `nil` check for `oldDetails.OldCdnId` here. Basically something like is being done in line 61.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -862,8 +863,8 @@ func updateV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 *
 		if err != nil {
 			return nil, http.StatusInternalServerError, nil, fmt.Errorf("querying delivery service with sslKeyVersion failed: %s", err)
 		}
-		if ds.CDNName != nil && ds.RoutingName != nil {
-			if sslKeysExist && (oldDetails.OldCdnName != *ds.CDNName || oldDetails.OldRoutingName != *ds.RoutingName) {
+		if (ds.CDNName != nil || ds.CDNID != nil) && ds.RoutingName != nil {
+			if sslKeysExist && (*oldDetails.OldCdnId != *ds.CDNID || oldDetails.OldCdnName != *ds.CDNName || oldDetails.OldRoutingName != *ds.RoutingName) {
```
srijeet0406, line 867: If `oldDetails.OldCdnId` is `nil`, this will seg fault here.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -862,8 +863,8 @@ func updateV31(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV31 *
 		if err != nil {
 			return nil, http.StatusInternalServerError, nil, fmt.Errorf("querying delivery service with sslKeyVersion failed: %s", err)
 		}
-		if ds.CDNName != nil && ds.RoutingName != nil {
-			if sslKeysExist && (oldDetails.OldCdnName != *ds.CDNName || oldDetails.OldRoutingName != *ds.RoutingName) {
+		if (ds.CDNName != nil || ds.CDNID != nil) && ds.RoutingName != nil {
```
srijeet0406, line 866: If one of ds.CDNName or ds.CDNID is nil, this will cause the next line to seg fault

srijeet0406 (changes requested):

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
rimashah25: I feel that I am complicating the solution by defining CDNID as a pointer of type int instead of int.

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices_test.go
```diff
@@ -70,7 +70,7 @@ func TestGetDetails(t *testing.T) {
 	if oldDetails.OldCdnName != "foo" {
 		t.Errorf("expected old cdn name to be foo, but got %v", oldDetails.OldCdnName)
 	}
-	if *oldDetails.OldCdnId != 1 {
+	if oldDetails.OldCdnId != 1 {
 		t.Errorf("expected old cdn id to be 1, but got %v", oldDetails.OldCdnName)
```
srijeet0406, line 74: This needs to be `oldDetails.OldCdnId`

rimashah25: Done

srijeet0406:
Tested this in CiaB. I noticed an issue:
Consider a DS that has SSL keys associated with it.
If I try to update this DS with something that has a different CDN ID, but no CDN Name or Routing Name, it gives me the success message, which it shouldn't. I think it has got to do with the way you are handling the conditional statements in `deliveryservices.go`.
I would do something like this:
`Check for ds.CDNId to not be nil`
`Now, put all the other conditionals under if hasSSLKeys {...}`
`Inside this conditional, you can check for the old and new CDN IDs to be equal`
`Next, you would check for if the ds.RoutingName is nil or not, and if it isn't, proceed to check if its equal to the old routing name(this check goes within the if hasSSLKeys conditional)`
`Next, you would check for if the ds.CDNName is nil or not, and if it isn't, proceed to check if its equal to the old CDN name (this check goes within the if hasSSLKeys conditional)`

srijeet0406 (approved):
Code looks good. Unit tests and API tests pass.
Tested in CiaB and made sure manual tests pass. 👍

PR #5490, opened 2021-02-03
TP Remember Pagination on ag-grids

### LINKED ISSUE #5384: TP grids should retain pagination positioning
## I'm submitting a ...

-  Traffic Portal

## Current behavior:
Pagination is not retained in any of the following tables when navigating away and then back to the table.

- servers
- ds requests

## New behavior:
Pagination IS retained in any of the following tables when navigating away and then back to the table.

- servers
- ds requests

- Navigate to https://tp.domain.tld/#!/servers or https://tp.domain.tld/#!/delivery-service-requests (make sure ds requests are enabled in traffic_portal_properties.json) and paginate to page 2
- click on a row to navigate away from the table
- hit the back button
- notice that you are no longer on page 2 but on page 1 instead.

shamrickus:

- [x] This PR fixes #5384

- Traffic Portal

Visit the Servers and DS X tables and verify that they will remember your page number when navigating back.

- master
- 5.0.0
- 4.1.0

### DIFF
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -417,12 +417,12 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
 			
 			try {
 				const page = localStorage.getItem(tableName + "_table_page");
-				const pageSize = $scope.gridOptions.api.paginationGetPageSize();
-				if (page !== undefined && page > 0 && page <= pageSize-1) {
+				const totalPages = $scope.gridOptions.api.paginationGetTotalPages();
+				if (page !== undefined && page > 0 && page <= totalPages-1) {
 					$scope.gridOptions.api.paginationGoToPage(page);
 				}
 			} catch (e) {
-				console.error("failed to load stored page number:", e);
+				console.error("Failed to load stored page number:", e);
 			}
 
 			$scope.gridOptions.api.addEventListener("sortChanged", function() {
--- a/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
+++ b/traffic_portal/app/src/common/modules/table/servers/TableServersController.js
@@ -416,7 +416,7 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
 			}
 			
 			try {
-				const page = localStorage.getItem(tableName + "_table_page");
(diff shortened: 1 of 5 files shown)


traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -414,6 +414,16 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
 			} catch (e) {
 				console.error("Failure to load stored page size:", e);
 			}
+			
+			try {
+				const page = localStorage.getItem(tableName + "_table_page");
+				const pageSize = $scope.gridOptions.api.paginationGetPageSize();
+				if (page !== undefined && page > 0 && page <= pageSize-1) {
```
mattjackson220, line 421: i think the `page <= pageSize-1` part is breaking it if the pageSize is too low. I set pageSize to 2 rows and then went to page 5 which means it wont go to that page again since it fails that part.  maybe use `$scope.gridOptions.api.paginationGetTotalPages()` instead?

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -414,6 +414,16 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
...
+					$scope.gridOptions.api.paginationGoToPage(page);
+				}
+			} catch (e) {
+				console.error("failed to load stored page number:", e);
```
mattjackson220, line 425: only cause i had another change, can you capitalize "failed" for consistency?

mattjackson220 (changes requested):

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
mattjackson220: i think we will also have to do `$scope.gridOptions.api.paginationSetPageSize(pageSize);` inside this if after the page change

shamrickus: Woops, I changed it `pagenationGetTotalPages` but I must've undid that change at some point.

shamrickus: Page size shouldn't need to be set as it was set before and it specifies the `max` number of rows to display.

mattjackson220 (approved):
looks good to me! tested and works as expected!

PR #5492, opened 2021-02-03
Adding nil checks before dereferencing, to prevent Traffic Monitor null pointer crash

### LINKED ISSUE #5125: Traffic Monitor Null Pointer Crash
Traffic Monitor is accessing `ReqInf.RemoteAddr` without checking for `nil` here:
https://github.com/apache/trafficcontrol/blob/2609e6fc59d98ae1985f0424f28956c6970b585d/traffic_monitor/towrap/towrap.go#L346
https://github.com/apache/trafficcontrol/blob/2609e6fc59d98ae1985f0424f28956c6970b585d/traffic_monitor/towrap/towrap.go#L355

The `traffic_ops/*client*` returned `ReqInf.RemoteAddr` is guaranteed non-nil if `err==nil`, but if `err != nil`, it may be nil, depending how far the client got before it errored. It must be checked for nil before accessing. 

This causes panics and applications crashes when certain types of errors happen.

## I'm submitting a ...
-  bug report

-  Traffic Monitor

## Current behavior:
Traffic Monitor crashes on certain Traffic Ops request errors.

## Expected behavior:
Traffic Monitor does not crash.

Induce particular types of Traffic Ops connection errors. This was observed after downgrading a Traffic Ops from 4.1 to 3.0. But that may or may not be consistently reproducible, and is certainly not the only case which will cause a crash.

## Anything else:

srijeet0406:

- [x] This PR fixes #5125  

- Traffic Monitor

Make sure the unit tests and api tests pass.

- master

## Additional Information

Just adding a nil check before dereferencing a variable, tests and docs not needed.

### DIFF
--- a/traffic_monitor/towrap/towrap.go
+++ b/traffic_monitor/towrap/towrap.go
@@ -343,7 +343,9 @@ func (s TrafficOpsSessionThreadsafe) CRConfigRaw(cdn string) ([]byte, error) {
 		b, reqInf, e := ss.GetCRConfig(cdn)
 		err = e
 		data = b
-		remoteAddr = reqInf.RemoteAddr.String()
+		if reqInf.RemoteAddr != nil {
+			remoteAddr = reqInf.RemoteAddr.String()
+		}
 	} else {
 		ss := s.get()
 		if ss == nil {
@@ -352,7 +354,9 @@ func (s TrafficOpsSessionThreadsafe) CRConfigRaw(cdn string) ([]byte, error) {
 		b, reqInf, e := ss.GetCRConfig(cdn)
 		err = e
 		data = b
-		remoteAddr = reqInf.RemoteAddr.String()
+		if reqInf.RemoteAddr != nil {
+			remoteAddr = reqInf.RemoteAddr.String()
+		}
 	}
 
 	if err == nil {


traffic_monitor/towrap/towrap.go
```diff
@@ -352,7 +354,9 @@ func (s TrafficOpsSessionThreadsafe) CRConfigRaw(cdn string) ([]byte, error) {
 		b, reqInf, e := ss.GetCRConfig(cdn)
 		err = e
 		data = b
-		remoteAddr = reqInf.RemoteAddr.String()
+		if reqInf.RemoteAddr != nil {
+			remoteAddr = reqInf.RemoteAddr.String()
```
TaylorCFrey, line 358: This code looks good to me, but what happens if we don't provide a value for `remoteAddr` later on down the line? According to the GitHub issue this will likely fix that particular panic, but do we end up with a panic later since the `remoteAddr` is potentially nil? Should we give it an empty string or use `localHostIP` like it does later if there was an error?

traffic_monitor/towrap/towrap.go
```diff
@@ -343,7 +343,9 @@ func (s TrafficOpsSessionThreadsafe) CRConfigRaw(cdn string) ([]byte, error) {
...
```
TaylorCFrey, line 347: Should we log the error here and below? If there's an error, should we try to recover or let it fall through?

traffic_monitor/towrap/towrap.go
srijeet0406: Well, `remoteAddr` is a string, so it cannot be `nil`. It will default to an empty string if no value is provided.

traffic_monitor/towrap/towrap.go
srijeet0406: I don't think that value being `nil` hampers the code after it. I could add an error here, but the only other place that `ReqAddr` (which gets set with the `remoteAddr` value) is being used is in the comparison on line 152 in the same file.

TaylorCFrey (approved):
Code LGTM and directly solves the related issue.

PR #5495, opened 2021-02-03
Ability to change the name of a topology

### LINKED ISSUE #5361: Provide the ability to change the name of a topology
## I'm submitting a ...

-  Traffic Ops
-  Traffic Portal

## Current behavior:
Currently you cannot change the name of a topology

## New behavior:
Allow the user to change the name of a topology thru the API and cascade that change to all appropriate areas. I.e. ds.toplogy

In TP, navigate to an existing topology and notice that the name field is disabled.

## Anything else:
Changing a topology name will probably dirty the snapshot for that CDN in 2 places: the topology section and the delivery service section (if the topology is assigned to a ds)

rimashah25:

- [x] This PR fixes #5361 

- Traffic Ops
- Traffic Portal

Two ways to test the PR:
1. Method 1:
Using curl API: api/3.0/topologies/edit?name=<existing_topology_name> and a json request body (like below), change the name field to a new name (for eg: test):

```
{
	"name":"test",
	"description":"west topology-test",
	"nodes":[{
		"cachegroup":"test",
		"parent":{"name":"","type":""},
		"secParent":{"name":"","type":""},
		"parents":[]
	}]
}
```
You should see a success message.

2. Method 2:
Using TP, edit an existing topology. The name field shouldn't be disabled and a user can edit the name and then click on update button.

- master (82314b1)

- [x] This PR includes tests

## Additional Information

No structure changes have occurred to require changing the documentation.


traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -394,8 +394,10 @@ func (topology TOTopology) GetKeys() (map[string]interface{}, bool) {
 
 // SetKeys is a requirement of the api.Updater interface and is called by
 // api.UpdateHandler().
+// SetKeys function will no longer makes the name key immutable by setting it to
+// query param. On 2/3/21, the requirement was added to change the topology name.
 func (topology *TOTopology) SetKeys(keys map[string]interface{}) {
-	topology.Name, _ = keys["name"].(string)
+	//topology.Name, _ = keys["name"].(string)
```
rawlinp, line 400: So, I'm not positive, but I _think_ the `Validation()` method might be doing validation based on the _requested_ name, rather than the _current_ name, which would be bad. Can you double-check if that's happening or not?

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -572,10 +574,11 @@ func (topology *TOTopology) addParents() (error, error, int) {
 	return nil, nil, http.StatusOK
 }
 
-func (topology *TOTopology) setDescription() (error, error, int) {
-	rows, err := topology.ReqInfo.Tx.Query(updateQuery(), topology.Description, topology.Name)
+func (topology *TOTopology) setNameDescription() (error, error, int) {
```
srijeet0406, line 577: I think we should just change the name of this function to `setTopologyDetails` or something on those lines.

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: Good catch.. I updated the Validate()

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: Good idea!

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -781,13 +785,13 @@ WHERE tcp.child IN
...
-description = $1
-WHERE t.name = $2
+name = $1,
+description = $2
+WHERE t.name = $3
 RETURNING t.name, t.description, t.last_updated
 `
 	return query
 }
-
```
rawlinp, line 790: gofmt will probably complain about this line missing

traffic_portal/app/src/common/modules/form/topology/edit/FormEditTopologyController.js
```diff
@@ -37,11 +37,12 @@ var FormEditTopologyController = function(topologies, cacheGroups, $scope, $cont
...
-	$scope.save = function(name, description, topologyTree) {
-		let normalizedTopology = topologyUtils.getNormalizedTopology(name, description, topologyTree);
-		topologyService.updateTopology(normalizedTopology).
+	$scope.save = function(currentName, newName, description, topologyTree) {
+		let normalizedTopology = topologyUtils.getNormalizedTopology(newName, description, topologyTree);
+		topologyService.updateTopology(normalizedTopology, currentName).
 			then(function(result) {
-				messageModel.setMessages(result.data.alerts, false);
+				messageModel.setMessages(result.data.alerts, true);
+				locationUtils.navigateToPath('/topologies');
```
rawlinp, line 45: Why navigate to `/topologies`? Don't we want to stay on the same page?

traffic_portal/app/src/common/modules/form/topology/form.topology.tpl.html
```diff
@@ -87,7 +87,8 @@
             </div>
             <div class="modal-footer">
                 <button type="button" class="btn btn-danger" ng-show="!settings.isNew" ng-click="confirmDelete(topology)">Delete</button>
-                <button type="button" class="btn btn-success" ng-disabled="topologyForm.$pristine || topologyForm.$invalid" ng-click="save(topology.name, topology.description, topologyTree)">{{settings.saveLabel}}</button>
+                <button ng-if="settings.isNew" type="button" class="btn btn-success" ng-disabled="topologyForm.$pristine || topologyForm.$invalid" ng-click="save(topology.name, topology.description, topologyTree)">{{settings.saveLabel}}</button>
```
rawlinp, line 90: Shouldn't this call to `save` have all 4 params?

rawlinp (changes requested):

traffic_portal/app/src/common/modules/form/topology/form.topology.tpl.html
rimashah25: nope, the first save is for create button and second save is for update button.

traffic_portal/app/src/common/modules/form/topology/edit/FormEditTopologyController.js
rimashah25: We can't since the `query param` in the url link will have old param. So, we need to go back to /topologies, and then click on a topology, to update the `query param` to new name.

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: oh ok.. i'll add it.

traffic_portal/app/src/common/modules/form/topology/form.topology.tpl.html
rawlinp: Ah, I see, they use the same function name -- that's confusing. Thank you for clarifying.

traffic_portal/app/src/common/modules/form/topology/edit/FormEditTopologyController.js
rawlinp: I'm not much of a javascripter, but is it possible to use the `result.data.response.name` to navigate back to the original page?

traffic_portal/app/src/common/modules/form/topology/form.topology.tpl.html
rimashah25: Anytime.

traffic_portal/app/src/common/modules/form/topology/edit/FormEditTopologyController.js
rimashah25: I am not sure if that would work but I can try.

rimashah25: Updated.

rawlinp (approved):
This LGTM, appears to work fine in manual testing, and checks passed.

PR #5502, opened 2021-02-05
Updated MTU check constraint in DB to allow 1280 as a legal value

### LINKED ISSUE #5288: Edge case in checking server interface MTU
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
If you create a server having an interface with an MTU of exactly 1280, you get back a 500 Internal Server Error response. This is because the validation in the API checks for MTU >= 1280, but the database check is strictly MTU > 1280 (or NULL).

## Expected behavior:
The API and database constraints should be consistent - I think the intended behavior is for 1280 to be a legal value.

Create a server having an interface with an MTU of exactly 1280.

rimashah25:

- [x] This PR fixes #5288 

- Traffic Ops

1. Create a JSON request where interface.mtu is set to 1280
2. curl using POST API /servers
3. Get a success (200 OK) with a server response struct
For eg:
```
curl -X POST -k -H "Accept: application/json" --cookie "mojolicious=<mojolicious id>" --data @post_server_ciab.json https://localhost:6443/api/3.0/servers
{"alerts":[{"text":"Server created","level":"success"}],"response":{"cachegroup":"CDN_in_a_Box_Mid-02","cachegroupId":7,"cdnId":2,"cdnName":"CDN-in-a-Box","domainName":"ciab.com","guid":null,"hostName":"bar","httpsPort":443,"id":19,"iloIpAddress":null,"iloIpGateway":null,"iloIpNetmask":null,"iloPassword":null,"iloUsername":null,"lastUpdated":"2021-02-05 20:45:48+00","mgmtIpAddress":null,"mgmtIpGateway":null,"mgmtIpNetmask":null,"offlineReason":null,"physLocation":"CDN_in_a_Box","physLocationId":2,"profile":"ATS_EDGE_TIER_CACHE","profileDesc":"Edge Cache - Apache Traffic Server","profileId":9,"rack":null,"revalPending":false,"status":"OFFLINE","statusId":1,"tcpPort":80,"type":"EDGE","typeId":11,"updPending":false,"xmppId":"e6966f42-b6be-414e-9534-f04a853aa12a","xmppPasswd":null,"routerHostName":null,"routerPortName":null,"interfaces":[{"ipAddresses":[{"address":"1.2.3.4","gateway":"1.0.0.1","serviceAddress":true}],"maxBandwidth":50,"monitor":true,"mtu":1280,"name":"int0"}],"statusLastUpdated":"2021-02-05T20:45:48.6267179Z"}}
```

- master (daa6a55)

- [x] This PR includes tests

## Additional Information

The documentation is correct and doesn't need an update
```
The MTU of this interface. If it is null, it may be assumed that the information is either not available or not applicable for this interface. This unsigned integer may not be less than 1280.
```


traffic_ops/app/db/migrations/2021020500000000_change_mtu_check_interfaces.sql
```diff
@@ -0,0 +1,21 @@
...
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE interface DROP CONSTRAINT interface_mtu_check;
+ALTER TABLE interface ADD CONSTRAINT interface_mtu_check CHECK (((mtu IS NULL) OR (mtu >= 1280)));
+
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+ALTER TABLE interface DROP CONSTRAINT interface_mtu_check;
+ALTER TABLE interface ADD CONSTRAINT interface_mtu_check CHECK (((mtu IS NULL) OR (mtu > 1280)));
```
ocket8888, line 21: Missing newline at EOF

traffic_ops/testing/api/v4/servers_test.go
```diff
@@ -885,6 +886,7 @@ func UpdateTestServers(t *testing.T) {
 
 	originalHostname := *resp.Response[0].HostName
 	originalXMPIDD := *resp.Response[0].XMPPID
+	originalMTU := *resp.Response[0].Interfaces[0].MTU
```
ocket8888, line 889: This check should go down below the check for `len(infs)` on line 896 to avoid a segfault. For the same reason, should also check that MTU is not `nil` before dereferencing it

traffic_ops/testing/api/v4/servers_test.go
```diff
@@ -950,6 +954,11 @@ func UpdateTestServers(t *testing.T) {
 		t.Fatalf("Cannot test server type change update; server '%s' had nil type ID", hostName)
 	}
 
+	// Check to verify mtu changed
+	if originalMTU == *respServer.Interfaces[0].MTU {
```
ocket8888, line 958: Need to check that `respServer.Interfaces` has a length of at least one, and that the 0<sup>th</sup> element has a non-`nil` MTU before dereferencing like this, to avoid a segfault.

traffic_ops/testing/api/v4/servers_test.go
```diff
@@ -950,6 +954,11 @@ func UpdateTestServers(t *testing.T) {
...
+		t.Errorf("MTU value didn't update. Expected: #{*updatedMTU}, actual: #{*originalMTU}")
```
ocket8888, line 959: I don't think this is going to print out the values, I think this will just literally print out
```
MTU value didn't update. Expected: #{*updatedMTU}, actual: #{*originalMTU}
```

traffic_ops/testing/api/v4/servers_test.go
rimashah25: ok

traffic_ops/testing/api/v4/servers_test.go
rimashah25: ok

ocket8888 (approved):
Confirmed this fixes the issue

PR #5506, opened 2021-02-09
Workflow to check for and PR minor Go revision updates

zrhoffman:

This PR adds a GitHub Actions workflow that checks once a day if a newer minor revision exists for our Go major version (currently `1.15`). If so, the workflow will
- Create a new branch in the *apache/trafficcontrol* repo and update the `GO_VERSION` file on that branch
- Update golang.org/x/ dependencies
- Open a pull request from that branch to *apache/trafficcontrol*'s `master`  branch.

Note that the workflow will *not* create a pull request for major version updates like `1.15.7` -> `1.16.0`.

- CI tests

1. Test the workflow
   1. Push the `go-versions-prs` branch to your own fork
   2. Go to Actions -> *Is it time to update Go?* -> *Run workflow* and manually run the workflow
   3. Verify that the pull request created in your fork contain the changes necessary to update the Go version for your fork.

2. Run the tests:
   ```shell
   cd .github/actions/pr-to-update-go
   python3 -m unittest discover ./tests
   ```

- [x] This PR includes tests


.github/actions/pr-to-update-go/pr_to_update_go/pr_template.md
```diff
@@ -0,0 +1,120 @@
...
+replacing the text in the checkbox item with the Issue number e.g.
+
+- [x] This PR fixes #9001 OR is not related to any Issue
+
+^ This will automatically close Issue number 9001 when the Pull Request is
+merged (The '#' is important).
+
+Be sure you check the box properly, see the "The following criteria are ALL
+met by this PR" section for details.
+-->
```
ocket8888, line 18: I don't think you need comments like these, they're for humans to read

.github/actions/pr-to-update-go/pr_to_update_go/__main__.py
```diff
@@ -0,0 +1,32 @@
...
+    try:
+        github_token: str = os.environ[ENV_GITHUB_TOKEN]
+    except KeyError:
+        print(f'Environment variable {ENV_GITHUB_TOKEN} must be defined.')
+        sys.exit(1)
+    gh = Github(login_or_token=github_token)
+    GoPRMaker(gh).run()
+
+
+main()
```
ocket8888, line 32: I _think_ this should be wrapped in an `if __name__ == "__main__":` but I could be wrong; I'm not entirely sure the interpreter sets `__name__` like that when the module is called with `python -m`.

It's not normal to import modules beginning with `_` from packages - and nothing is meant to use this - so it could be safe, but it damages my delicate sensibilities to have large routines running as side effects to importing a module.

.github/actions/pr-to-update-go/setup.cfg
```diff
@@ -0,0 +1,37 @@
...
+    requests
+
+[options.entry_points]
+console_scripts = pr-to-update-go = pr_to_update_go:main
+
+[options.extras_require]
+test = unittest
+
+[options.package_data]
+pr_to_update_go = pr_template.md
```
ocket8888, line 37: Setuptools will read this in lieu of passing the info as args? That's pretty neat. Wonder how long that's been the case.

.github/actions/pr-to-update-go/README.rst
```diff
@@ -0,0 +1,73 @@
...
+	    GO_VERSION_FILE: GO_VERSION
+
+Tests
+=====
+
+To run the unit tests:
+
+.. code-block:: shell
+
+	python3 -m unittest discover ./tests
```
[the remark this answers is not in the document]
ocket8888: If the intended parser is PyPi it also won't work (I know from my experience packaging the TO Python client), but I think that's probably less likely.

.github/actions/pr-to-update-go/pr_to_update_go/__main__.py
zrhoffman: If the filename is `__main__.py`, then `__name__` is always equal to `'__main__'`.

.github/actions/pr-to-update-go/pr_to_update_go/pr_template.md
zrhoffman: Removed comments in de4b55e9f3

.github/actions/pr-to-update-go/pr_to_update_go/GoPRMaker.py
```diff
@@ -0,0 +1,210 @@
...
+    gh: Github
+    latest_go_version: str
+    repo: Repository
+
+    def __init__(self, gh: Github) -> None:
+        self.gh = gh
+        repo_name: str = self.get_repo_name()
+        self.repo = self.get_repo(repo_name)
+
+    def run(self) -> None:
```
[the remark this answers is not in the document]
zrhoffman: I like `-> None:` because then PyCharm gives you warnings when trying to a value returned by a function whose return type is `None` without needing to analyze control flow. For example, on line 6 of this snippet,

```python
def a() -> None:
	if False:
		return 1
	print('Returns nothing')

b: int = a()
```

it warns
> Expected type 'int', got 'None' instead.

But that warning is gone if `-> None` is removed.

.github/actions/pr-to-update-go/README.rst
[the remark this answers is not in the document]
zrhoffman: > However, GitHub's renderer isn't Sphinx. I think it's just standard RST - and the `code-block` directive is provided by Sphinx, so if the intention was for this to be readable on GitHub I don't think it'll work.

I wasn't able to find anything rendering incorrectly at https://github.com/apache/trafficcontrol/blob/b3a6c32500/.github/actions/pr-to-update-go/README.rst.

Sphinx doesn't seem to care about about a blank line after

https://github.com/apache/trafficcontrol/blob/b3a6c32500f0c69eb81b480cba9d98da91dfa568/.github/actions/pr-to-update-go/README.rst#L73

When using Sphinx, the code block renders correctly with or without a newline at the end of the file.

.github/actions/pr-to-update-go/README.rst
```diff
@@ -0,0 +1,73 @@
...
+***************
+pr-to-update-go
+***************
+
+Opens a PR if a new minor Go revision is available.
+
+For example, it will open a PR if the ``GO_VERSION`` contains ``1.14.7`` but Go versions 1.15.1 and 1.14.8 are available, it will
+
+1. Create a branch named ``go-1.14.8`` to update the repo's Go version to 1.14.8
+2. Open a PR targeting the ``master`` branch from branch ``go-1.14.8``
```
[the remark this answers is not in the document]
zrhoffman: Sentence restructured in 366a14ebf

.github/actions/pr-to-update-go/README.rst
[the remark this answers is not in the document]
ocket8888: That's very interesting, because if you look at e.g. [the source for the `/cdns` API endpoint docs in APIv4](https://github.com/apache/trafficcontrol/blob/b3a6c32500f0c69eb81b480cba9d98da91dfa568/docs/source/api/v4/cdns.rst) you'll notice it isn't rendering request or response examples.

Matter of syntax support? You'd think it just falls back to regular monospace...

Regardless, you're right, it's rendering properly.

.github/actions/pr-to-update-go/pr_to_update_go/GoPRMaker.py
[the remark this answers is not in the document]
ocket8888: So Pycharm treats implicit return types as [Any](https://docs.python.org/3/library/typing.html#typing.Any) ? As a Typescript programmer I turn my nose up at such sloppy inference!

But that's fine.

It's weird to me that `typing` has an `any` and a `never` but not a `void`. Ideally you'd be able to do
```python3
from typing import Any, Void

def a() -> Void:
    pass

b: Any = a()
```

and that would show up in linters/IDEs/whathaveyou as an invalid assignment of void to a value. But I guess the technology just isn't there yet.

.github/actions/pr-to-update-go/pr_to_update_go/__main__.py
ocket8888: Oh, right. That's an interesting edge case.

zrhoffman:
Updated the workflow so that it additionally updates `golang.org/x/...` dependencies used by the project, since before Go modules, we always used the latest versions, and with Go modules, their versions are pinned.

.github/actions/pr-to-update-go/pr_to_update_go/go_pr_maker.py
```diff
@@ -0,0 +1,391 @@
...
+	def update_golang_org_x(self, previous_commit: Commit) -> Union[GitCommit, None]:
+		"""
+		:param previous_commit:
+		:type previous_commit:
+		:return:
+		:rtype: Union[GitCommit, None]
+		"""
+		subprocess.run(['git', 'fetch', 'origin'], check=True)
+		subprocess.run(['git', 'checkout', previous_commit.sha], check=True)
+		script_path: str = f'.github/actions/pr-to-update-go/update_golang_org_x.sh'
```
ocket8888, line 329: This f-string is unnecessary

.github/actions/pr-to-update-go/pr_to_update_go/go_pr_maker.py
```diff
@@ -0,0 +1,391 @@
...
+import re
+import subprocess
+import sys
+from typing import Union
+
+import requests
+from github.Branch import Branch
+from github.Commit import Commit
+from github.GitCommit import GitCommit
+from github.GitRef import GitRef
```
ocket8888, line 33: This import is never used

.github/actions/pr-to-update-go/pr_to_update_go/go_pr_maker.py
```diff
@@ -0,0 +1,391 @@
...
+		tree_hash = subprocess.check_output(
+			['git', 'log', '-1', '--pretty=%T', previous_commit.sha]).decode().strip()
+		base_tree: GitTree = self.repo.get_git_tree(sha=tree_hash)
+		tree: GitTree = self.repo.create_git_tree(tree_elements, base_tree)
+		commit_message: str = f'Update golang.org/x/ dependencies for go{self.latest_go_version}'
+		previous_git_commit: GitCommit = self.repo.get_git_commit(previous_commit.sha)
+		git_commit: GitCommit = self.repo.create_git_commit(message=commit_message, tree=tree,
+			parents=[previous_git_commit],
+			author=self.author, committer=self.author)
+		print(f'Updated golang.org/x/ dependencies')
```
ocket8888, line 354: This also doesn't need to be an f-string

zrhoffman: Removed *f* in 978e5f94cd

.github/actions/pr-to-update-go/pr_to_update_go/go_pr_maker.py
zrhoffman: Removed unused import in b8a2a9e6eb

.github/actions/pr-to-update-go/pr_to_update_go/go_pr_maker.py
zrhoffman: *f* removed in 978e5f94cd

ocket8888 (approved):
Verified this works on my fork

PR #5508, opened 2021-02-09
Add 'value' query parameter to /parameters

ocket8888:

This PR adds a `value` query parameter to `/parameters` and makes the APIv4 client use it (since API v3 and under can't be guaranteed to be running against a TO that supports the parameter, they don't get the same treatment).

- Traffic Ops Client (Go)
- Traffic Ops

Verify the Go client/API integration tests work.

- [x] This PR includes tests


srijeet0406 (commented):
Code looks good. Could you pls add some tests around this? Maybe just an API test where you query by value and check the expected vs actual result.

ocket8888:
There actually already is a test for this, because the tests use `client.GetParameterByNameAndConfigFileAndValueWithHdr` and I changed that to use the new query parameter instead of just filtering client-side

srijeet0406 (approved):
Code looks good. API and unit tests pass. Manual tests pass as well.

PR #5509, opened 2021-02-10
Update to DB migration server_id_primary_key.sql

hbeatty:

I was unable to run the dbadmin upgrade because this update was trying to add 2 primary keys (one for each schema).

I encountered this while trying to upgrade from 4.1.1 to 5.0.0.

- [x] Is not related to any Issue

- Traffic Ops

I'm not sure how to test this. My DB has been around since the 1.x days. It has 2 schemas (traffic_ops and public). This SQL statement was trying to create a primary key in both schemas.

- 4.1.1 -> 5.0.0 upgrade


ocket8888 (approved):
Checked migration up/down, works without breaking, finished database works with TO on master (checked DS content available through CiaB).

PR #5511, opened 2021-02-10
Update add_deleted_tables.sql

hbeatty:

The create function call was not adding the function to the public schema
which was causing the subsequent "ALTER FUNCTION" call to fail.

- Traffic Ops

I'm not sure how to test this. My DB has been around since the 1.x days. It has 2 schemas (traffic_ops and public).

- 4.1.1 -> 5.0.0 upgrade


ocket8888 (approved):
Verified db upgrade/downgrade works without issue, CiaB tests show that fully upgraded DB works fine for TO.

this is ready to go once conflicts are resolved

PR #5513, opened 2021-02-11
Log when getting errors from TO but backup configs exist

rawlinp:
Currently, TM doesn't log an error if it got an invalid monitoring snapshot from TO and a backup file exists. Also TM doesn't log an error if it gets an error fetching a CRConfig snapshot from TO and a backup file exists. This PR makes TM log errors in those cases, because we would need to know the cause of these errors if they were to occur.

Also, this PR makes TM consider a monitoring snapshot invalid if there are no delivery services.

- Traffic Monitor

1. Starting with a valid crconfig so that TM will create a backup of it, shut down TO so that TM fails to fetch it. TM should log an informative error in this case.
2. Starting with a valid monitoring snapshot so that TM will create a backup of it, delete all delivery services from TO, then snapshot the CDN. This should produce an invalid monitoring snapshot, upon which TM should log an informative error (even though it has a valid backup).

- master
- 4.1.x
- 5.0.x

- [x] This PR includes tests


ocket8888 (approved):
LGTM; verified new log line appears when a backup is in use.

PR #5514, opened 2021-02-11
Acme auto renew

mattjackson220:

This PR updates the certificate autorenewal script to include ACME providers (previously only did Let's Encrypt)

- Documentation
- Traffic Ops

Set up the necessary information in cdn.conf to allow automatic renewals
Make sure you have delivery services that have ACME certificates (and other types as well for testing)
Do a POST against the `acme_autorenew`endpoint or put in the username and password and setup the cron job to run
Verify that all expiring certs for ACME and Let's Encrypt are renewed and a summary email is sent if that is setup
Verify that docs build and look good

Tests are not included since this is entirely dependent on 3rd parties


docs/source/admin/traffic_ops.rst
```diff
@@ -349,9 +356,7 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 	.. versionadded:: 4.1
 
 	:user_email: A required email address to create an account with Let's Encrypt or to receive expiration updates. If this is not included then `rate limits <https://letsencrypt.org/docs/rate-limits>`_ may apply for the number of certificates.
-	:send_expiration_email: A boolean option to send email summarizing certificate expiration status
 	:convert_self_signed: A boolean option to convert self signed to Let's Encrypt certificates as they expire. This only works for certificates labeled as Self Signed in the Certificate Source field.
-	:renew_days_before_expiration: Set the number of days before expiration date to renew certificates.
```
ocket8888, line 354: Removing these configuration file properties is a breaking change that needs a major release cycle deprecation period.

docs/source/api/v1/letsencrypt_autorenew.rst
```diff
@@ -36,54 +36,20 @@ No parameters available
...
+	{ "alerts": [
+		{
+			"text": "This endpoint is deprecated, please use letsencrypt/autorenew instead",
+			"level": "warning"
+		},
+		{
+			"text": "Beginning async call to renew certificates.  This may take a few minutes.",
+			"level": "success"
+		}
+	]}
```
ocket8888, line 55: You can't change the response structure of old API versions. If there's really no way to meaningfully reproduce it, then you can just use meaningless data, but this is a breaking client change to a released API version.

docs/source/api/v4/acme_autorenew.rst
```diff
@@ -0,0 +1,49 @@
...
+
+	HTTP/1.1 200 OK
+	Content-Type: application/json
+
+	{ "alerts": [
+      {
+         "text": "Beginning async call to renew certificates.  This may take a few minutes.",
+         "level": "success"
+      }
+	]}
```
ocket8888, line 49: indentation within code blocks can use spaces if you want, but the first indent that associates the code with the `.. code-block` directive **must** be a tab. I think in this particular case this won't render properly.

docs/source/api/v4/acme_autorenew.rst
```diff
@@ -0,0 +1,49 @@
...
+``acme_autorenew``
+******************
+
+``POST``
+========
+Generates SSL certificates and private keys for all :term:`Delivery Services` that have certificates expiring within the configured time. This uses:abbr:`ACME (Automatic Certificate Management Environment)` or Let's Encrypt depending on the certificate.
+
+:Auth. Required: Yes
+:Roles Required: "admin" or "operations"
+:Response Type:  Object
```
ocket8888, line 28: I think the response type is actually `undefined` here.

docs/source/api/v4/acme_autorenew.rst
```diff
@@ -0,0 +1,49 @@
...
+No parameters available
+
+
+Response Structure
+------------------
+
+.. code-block:: http
+	:caption: Response Example
+
+	HTTP/1.1 200 OK
```
ocket8888, line 41: From [the API guidelines](https://traffic-control-cdn.readthedocs.io/en/latest/development/api_guidelines.html#accepted): 

> _"`202 Accepted` MUST be used when the server is performing some task asynchronously (e.g. refreshing DNSSEC keys) but the status of that task cannot be ascertained at the current time._"

There's also:

> _"Endpoints that create asynchronous jobs SHOULD provide a URI to which the client may send GET requests to obtain a representation of the job’s current state in the Location HTTP header. They MAY also provide an info-level Alert that provides the same or similar information in a more human-friendly manner."_

and the section has more information on what that endpoint should look like, but that's totally optional, and can even be added at a later date in a different PR, if you want. Just wanted to put that on your radar.

docs/source/api/v1/letsencrypt_autorenew.rst
mattjackson220: i think these were just bad docs. the other way was the original plan but then it took too long so i switched it to be async and the other info was sent in an email instead of as the response. i dont think that version was ever released

docs/source/api/v4/acme_autorenew.rst
mattjackson220: oh thats good to know! ill update these to 202 and then get a different PR in for the status endpoint!

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -192,7 +191,8 @@ func renewAcmeCerts(cfg *config.Config, dsName string, ctx context.Context, curr
 	return nil, nil, http.StatusOK
 }
 
-func getAcmeAccountConfig(cfg *config.Config, acmeProvider string) *config.ConfigAcmeAccount {
+// GetAcmeAccountConfig returns the ACME account information from cdn.conf for a given provider
```
ocket8888, line 194: GoDoc comments should end with a `.`

traffic_ops/traffic_ops_golang/deliveryservice/autorenewcerts.go
```diff
@@ -91,7 +102,16 @@ func RenewCertificates(w http.ResponseWriter, r *http.Request) {
...
 	go RunAutorenewal(existingCerts, inf.Config, ctx, inf.User)
 
-	api.WriteRespAlert(w, r, tc.SuccessLevel, "Beginning async call to renew Let's Encrypt certificates.  This may take a few minutes.")
+	var alerts tc.Alerts
+	if deprecated {
+		alerts.AddAlerts(api.CreateDeprecationAlerts(deprecation))
+	}
+
+	alerts.AddAlert(tc.Alert{
+		Text:  "Beginning async call to renew certificates.  This may take a few minutes.",
```
ocket8888, line 111: Looks like an extra space here between "certificates." and "This"

docs/source/api/v1/letsencrypt_autorenew.rst
ocket8888: alright, in that case: good work fixing it

ocket8888 (approved):
This LGTM; I confirmed the async job takes off (though not successfully) in CDN-in-a-Box, and since it's plugging into existing, working ACME logic to do anything more I'm confident it actually works in an environment with a working ACME provider.

PR #5515, opened 2021-02-12
Fix migration due to different constraint naming across PGSQL versions

rawlinp:
In 13.1, the constraint appears to get named 'ip_address_interface_server_fkey' by default, but in 9.6 it appears to get named 'ip_address_interface_fkey' by default. This fixes the migration to change the name to the 9.6 version before altering the constraint in order for upgrades in 13.1 to work.

This issue was found when resetting the DB in 13.1.

- Traffic Ops

On postgres 9.6, run the DB migration (reset, upgrade, down, upgrade), ensure it is successful. Do the same with postgres 13.1.

- basically just master because 13.1 support isn't official yet

- [x] DB migration fix, covered by DB tests
- [x] DB migration fix, no docs necessary
- [x] not including changelog because this just the start of adding 13.1 support (which will be coming later)


mattjackson220 (approved):
Looks good! goose works now for both ip_address_interface_server_fkey and ip_address_interface_fkey

PR #5521, opened 2021-02-16
Fix TOclient topologies path

rob05c:
Fixes the TO cilent's topologies GetTopologiesWithHdr path, which got missed in the `/api/` removal.

Incidentally, this was discovered with the new ORT Integration Test Framework.

Includes a test for GetTopologiesWithHdr (one didn't exist before, which is how it got missed).
No docs, no interface change.
No changelog, no interface change and bug isn't in a release.

Fixes `traffic_ops/v3-client.GetTopologiesWithHdr`.

- Traffic Control Client Go

Run TO API Tests. Run ORT Integration tests.. Run ORT against a TO with Topologies. Write an app calling `traffic_ops/v3-client.GetTopologiesWithHdr`.

- master
- 


traffic_ops/testing/api/v3/topologies_test.go
```diff
@@ -61,6 +62,19 @@ func CreateTestTopologies(t *testing.T) {
 	}
 }
 
+func GetTestTopologies(t *testing.T) {
```
rawlinp, line 65: For posterity, we should also add this test to the v4 tests.

rawlinp (changes requested):

traffic_ops/testing/api/v3/topologies_test.go
rob05c: Done

rawlinp (approved):
Looks good, thanks!

Issue #5488, opened 2021-02-02
Fix pkg.go.dev


rob05c:
pkg.go.dev replaces the old godoc.org Go auto-documentation site. It requires license files in certain places, or won't display documentation for libraries, for legal reasons.

It looks like it's broken for all ATC libraries: https://pkg.go.dev/github.com/apache/trafficcontrol/lib/go-tc

We should fix it. It looks like it might just be a matter of putting license files in the right places for it to detect.

Labelling bug, even though it's not our fault, because having the canonical Go docs fail really is an issue we should fix.

## I'm submitting a ...
-  bug report
- 
## Traffic Control components affected ...
-  Documentation

## Current behavior:
pkg.go.dev fails for all Go libraries in ATC.

## Expected behavior:
pkg.go.dev succeeds for all Go libraries in ATC.

## Minimal reproduction of the problem with instructions:
Go to https://pkg.go.dev/github.com/apache/trafficcontrol/lib/go-tc or any other ATC library path, observe failure.

ocket8888:
This should be fixed by #5229 as per the discussion in #4380

zrhoffman:
#5229 does not fix this issue. Our Godocs not displaying is a licensing issue, not a Go modules issue.

The Apache instructions for `LICENSE` and `NOTICE` files has instructions for [bundling permissively-licensed dependencies](https://infra.apache.org/licensing-howto.html#permissive-deps):
> <h3 id="permissive-deps">Bundling permissively-licensed dependencies</h3>
> <p>Bundling a dependency which is issued under one of the following licenses is straightforward, assuming that license applies uniformly to all files within the dependency:</p>
> <ul><li>BSD (without advertising clause)</li><li>MIT/X11</li></ul>
> <p>In <code>LICENSE</code>, add a <a href="http://s.apache.org/Hqj" target="_blank">pointer</a> to the dependency's license within the distribution and a short note summarizing its licensing:</p>
> <pre><code class="hljs coffeescript">This product bundles SuperWidget <span class="hljs-number">1.2</span><span class="hljs-number">.3</span>, which <span class="hljs-keyword">is</span> available under a <span class="hljs-string">"3-clause BSD"</span> license. For details, see deps<span class="hljs-regexp">/superwidget/</span>.</code></pre>
> <p>Under normal circumstances, there is no need to modify <code>NOTICE</code> to mention a bundled dependency.</p>
[...]

PR #5527, opened 2021-02-16
LICENSE file cleanup

zrhoffman:
Our `LICENSE` file lists so many bundled dependencies that it no longer meets [the 75% coverage threshold](https://github.com/golang/pkgsite/blob/5867665b19/internal/licenses/licenses.go#L43-L50) required in order for pkg.go.dev to identify the `LICENSE` file as an Apache-2.0 license:
```go
const (
	// coverageThreshold is the minimum percentage of the file that must contain
	// license text.
	coverageThreshold = 75

	// unknownLicenseType is for text in a license file that's not recognized.
	unknownLicenseType = "UNKNOWN"
)
```

It was ~33% around [`09e25013fd~`](https://github.com/apache/trafficcontrol/commit/09e25013fd~), then #5229 brought it up to ~51% but we're still a ways from the required 75% coverage threshold. If we do not move that bundled depedencies license list to another file like [`NOTICE`](https://github.com/apache/trafficcontrol/blob/master/NOTICE), we would need a google/licensecheck License Regular Expression exception added to [/internal/licenses/exceptions](https://github.com/golang/pkgsite/tree/5867665b19/internal/licenses/exceptions) tree of the [golang/pkgsite](https://github.com/golang/pkgsite) project.

This PR cleans up our `LICENSE` file a bit before such an exception is added to golang/pkgsite. Once the LRE exception is added, we can additionally add a GitHub Action to ensure that future changes to `LICENSE` to not break pkg.go.dev compatibility.

- - [x] This PR is related to Issue #5488 but does not fix it.

- This PR also removes unused licenses in the /licenses directory.

- LICENSE

Check if `LICENSE` file looks good

- [x] I have explained why tests are unnecessary (tests can be added after an exception is added to golang/pkgsite)
- [x] An update to CHANGELOG.md not necessary


zrhoffman:
@alficles Would you please review?

alficles (approved):
This is good stuff, we should merge this. This should help make the license file a bit easier for heuristic parsers. Thanks for cleaning up the old license files in the license dir as well.

PR #5528, opened 2021-02-16
Reorder SQL migrations to fix some upgrade issues

hbeatty:

This moves some SQL migrations past the last one in 5.0.x that were not in the 4.1.x and 5.0.x branch.

- Traffic Ops

Do a DB upgrade from 5.0.0 to 5.1.0

- 5.0.0
- 5.1.0


rawlinp:
Ok, once this is merged it can be backported to 5.1.x, so we don't need https://github.com/apache/trafficcontrol/pull/5528.

ocket8888:
Yeah, I can do that for you. No need of a separate PR

hbeatty:
Just curious what the difference is?

rawlinp:
By cherrypicking, we include the original commit hash from master by using the `-x` option. It also helps ensure that the `master` version and the `5.1.x` version are actually the same changes (in case we request changes here which would cause it to diverge from the 5.1.x PR that was opened simultaneously).

rawlinp (approved):
LGTM, DB upgrades from `2020110200000000_add_service_category_deleted.sql` (first migration script in 5.0) to the latest and downgrades back successfully.

Issue #5427, opened 2021-01-12
Traffic Ops should auto-calculate the max origin connections per mid 


dneuman64:
## I'm submitting a ...

-  new feature / enhancement request

## Traffic Control components affected ...

-  Traffic Ops
-  Traffic Portal

## Current behavior:

Currently, if the max origin connections field is set in TO, it is only used to determine connections per server for DSs that do NOT use Mids. 

## New behavior:

If the Max Origin Connections field is set, Traffic Ops should automatically calculate the number of connections per Edge OR Mids for a DS.   This should be based on the amount of REPORTED caches.

dneuman64:
It looks like Traffic Portal might not make that field available for all DSs, so TP might need to be fixed as well.

ocket8888:
Also, if it currently does only apply to Edge-tier cache servers, then [the related documentation](https://traffic-control-cdn.readthedocs.io/en/latest/overview/delivery_services.html#max-origin-connections) is wrong, as it specifies that the setting applies to *Mid-tier* cache servers.

rob05c:
ORT should be applying it to Edges for direct DSes (HTTP_LIVE, DNS_LIVE) and to Mids for DSes that use the Mids: https://github.com/apache/trafficcontrol/blob/master/lib/go-atscfg/headerrewritemiddotconfig.go#L153

Unless there's a bug. 

IMO it's better for ORT to do the calculation than TO, just because it's more canaryable.

mitchell852:
> ORT should be calculating and applying it to Edges for direct DSes (HTTP_LIVE, DNS_LIVE) and to Mids for DSes that use the Mids: https://github.com/apache/trafficcontrol/blob/master/lib/go-atscfg/headerrewritemiddotconfig.go#L153

yeah, i thought it did this as well. I originally created this feature in TP/TO and @rob05c moved it into ORT which i believe is the right place to do the calculations.

rawlinp:
We also need to keep topologies in mind here -- the origin connections need to be calculated based upon the number of caches in the last (non-origin) tier of the DS's topology.

PR #5536, opened 2021-02-17
Fix atscfg comparing pointers not values

rob05c:
This causes the config gen to think CDNs are different when they aren't, causing it to think no servers on the same CDN are assigned, omitting a max_origin_connections entry.

Includes modifying the tests to have different pointers, so the tests will fail without this fix.
Includes changelog.
No docs, no interface change.

- Traffic Ops ORT

Generate config on a Mid with MaxOriginConnetions on a DS, verify max_origin_connections is inserted in header rewrite file.

- master
- 5.0
- 5.1

## Additional Information


jpappa200:
This looks good to me.

rawlinp:
@rob05c this seems to fix https://github.com/apache/trafficcontrol/issues/5427 -- does it?

rob05c:
I'm not entirely clear on the scope of that case. But yes, with this fix, ATC will auto-calculate the max origin connections per mid.

dsouza93 (approved):
This looks good to me. I was able to get an ATS mid to properly configure the max origin connections.

PR #5537, opened 2021-02-17
update service category header rewrite logic

jehunte:
- This PR correctly sets the ATS header when a service category is assigned to a delivery service
- Tests are not necessary for this pr because there are already tests in headerrewritedotconfig_test.go checking that this header gets added to the string

- Traffic Ops ORT
- Documentation is not required because it fixes a bug with the existing implementation. Documentation is already in place for the functionality of the feature.

- Check the traffic server hdr_wr config file contains the correct headers/rewrite string

- master 
- 5.x.x

### DIFF
--- a/lib/go-atscfg/headerrewritedotconfig.go
+++ b/lib/go-atscfg/headerrewritedotconfig.go
@@ -35,7 +35,7 @@ const HeaderRewritePrefix = "hdr_rw_"
 const ContentTypeHeaderRewriteDotConfig = ContentTypeTextASCII
 const LineCommentHeaderRewriteDotConfig = LineCommentHash
 
-const ServiceCategoryHeader = "X-CDN-SVC"
+const ServiceCategoryHeader = "CDN-SVC"
 
 const MaxOriginConnectionsNoMax = 0 // 0 indicates no limit on origin connections
 
--- a/lib/go-atscfg/headerrewritedotconfig.go
+++ b/lib/go-atscfg/headerrewritedotconfig.go
@@ -35,7 +35,7 @@ const HeaderRewritePrefix = "hdr_rw_"
 const ContentTypeHeaderRewriteDotConfig = ContentTypeTextASCII
 const LineCommentHeaderRewriteDotConfig = LineCommentHash
 
-const ServiceCategoryHeader = "CDN-SVC"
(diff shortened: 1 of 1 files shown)


lib/go-atscfg/headerrewritedotconfig.go
```diff
@@ -35,7 +35,7 @@ const HeaderRewritePrefix = "hdr_rw_"
 const ContentTypeHeaderRewriteDotConfig = ContentTypeTextASCII
 const LineCommentHeaderRewriteDotConfig = LineCommentHash
 
-const ServiceCategoryHeader = "CDN-SVC"
+const ServiceCategoryHeader = "X-CDN-SVC"
```
rob05c, line 38: Why `X-`? 

The IETF has asked people to stop prefixing headers with `X-` because it makes it hard to remove in a standard later when they become common: https://tools.ietf.org/html/rfc6648

rob05c:
So, I'm noticing the `set-header` logic doesn't add its own hook. Which means it's going to fall under whatever the last hook happens to be, which may be a response rather than request hook, or it could be an error if there's no MaxOriginConnections or EdgeHeaderRewrite.

I'm also noticing the CDN-SVC header isn't in `topologyheaderrewritedotconfig.go`, so it won't be added for Topology DSes.

This PR doesn't introduce either of those issues, so I wouldn't insist it fix them. If you want to, that'd be great, or if not, whoever merges this should make Github Issues so we don't forget.

rawlinp:
Should we add this to the 5.1.1 milestone?

jehunte:
> Should we add this to the 5.1.1 milestone?

That would be great! @rob05c and I chatted about his comments and that will need to be added for the release too. It should be pretty straight forward. He was going to take a look at this stuff but I'm not sure where that falls on his list of priorities. Rob do you think this will be ready for the 5.1.1 milestone?

rob05c (approved):
Looks good! This fixes an issue with the header rewrite, and tests pass.

rob05c:
@jehunte I'm not sure what the timeline on 5.1.1 is, but yeah, I think this plus #5641 should go in it, if possible. If it's more than a week or so, there are a list of ORT fixes and features that should go in. I'll add the `5.1.1` tag to them all, and people can remove if necessary

Issue #5377, opened 2020-12-15
Ansible playbooks still use Traffic Ops API v1


zrhoffman:
## I'm submitting a ...

-  improvement request (usability, performance, tech debt, etc.)

## Traffic Control components affected ...

- Ansible playbooks

## Current behavior:

The following references to the Traffic Ops API in our Ansible playbooks need to be updated:

| HTTP Method | Endpoint | Line of Code |
| ----------- | -------- |:------------ |
| `POST` | `api/1.3/user/login` | https://github.com/apache/trafficcontrol/blob/df92555ddfa90760320c64ec619ac1c34af7f0ca/infrastructure/ansible/influxdb_relay.yml#L34 |
| `GET` | `api/1.3/servers` | https://github.com/apache/trafficcontrol/blob/df92555ddfa90760320c64ec619ac1c34af7f0ca/infrastructure/ansible/influxdb_relay.yml#L49 |
| * | `1.3` | https://github.com/apache/trafficcontrol/blob/df92555ddfa90760320c64ec619ac1c34af7f0ca/infrastructure/ansible/roles/dataset_loader/defaults/main.yml#L25 |
| `GET` | `internal/api/1.3/federations.json` | https://github.com/apache/trafficcontrol/blob/df92555ddfa90760320c64ec619ac1c34af7f0ca/infrastructure/ansible/roles/dataset_loader/defaults/main.yml#L743 |
| `GET` | `api/1.3/profiles/name/{{name}}/parameters` | https://github.com/apache/trafficcontrol/blob/df92555ddfa90760320c64ec619ac1c34af7f0ca/infrastructure/ansible/roles/dataset_loader/profile.parameter.conversion.md#L23 |
| `POST` | `api/1.2/user/login` | https://github.com/apache/trafficcontrol/blob/df92555ddfa90760320c64ec619ac1c34af7f0ca/infrastructure/ansible/roles/to_api/tasks/login.yml#L17 |
| `GET` | `api/1.2/cdns` | https://github.com/apache/trafficcontrol/blob/df92555ddfa90760320c64ec619ac1c34af7f0ca/infrastructure/ansible/roles/to_api/tasks/queue_updates.yml#L20 |
| `POST` | `api/1.2/cdns/{{name}}/queue_update` | https://github.com/apache/trafficcontrol/blob/df92555ddfa90760320c64ec619ac1c34af7f0ca/infrastructure/ansible/roles/to_api/tasks/queue_updates.yml#L32 |
[...]

Issue #5339, opened 2020-12-01
Add/update SSL keys on a delivery service does not create change log entry


mitchell852:
## I'm submitting a ...
-  improvement request (usability, performance, tech debt, etc.)

## Traffic Control components affected ...
-  Traffic Ops

## Current behavior:
It appears that the following route does not create a change log entry:

```
{api.Version{3, 0}, http.MethodPost, `deliveryservices/sslkeys/add$`, deliveryservice.AddSSLKeys, auth.PrivLevelAdmin, Authenticated, nil, 28728785833, noPerlBypass},
```

## New behavior:
Exercise the following API route successfully:

```
{api.Version{3, 0}, http.MethodPost, `deliveryservices/sslkeys/add$`, deliveryservice.AddSSLKeys, auth.PrivLevelAdmin, Authenticated, nil, 28728785833, noPerlBypass},
```
and ensure an message like this is created:

```
DS: testqa, ID: 372, ACTION: Added/updated SSL keys
```

You should see the new entry here:

`GET api/3.0/logs`

## Minimal reproduction of the problem with instructions:
Hit the route. If it didn't create a change log entry, it needs one.

## Anything else:

Also, this endpoint:

```
{api.Version{3, 0}, http.MethodPost, `deliveryservices/sslkeys/generate/?$`, deliveryservice.GenerateSSLKeys, auth.PrivLevelOperations, Authenticated, nil, 2534390513, noPerlBypass},
```

should create an change log like:

```
DS: testqa, ID: 372, ACTION: Generated SSL keys
```

TaylorCFrey:
I got this one

mitchell852:
Actually, looks like generate is working:

```
DS: testqa, ID: 372, ACTION: Added SSL keys
```

Maybe that can simply be changed to "Generated SSL keys" instead of "Added"

mitchell852:
but 


```
deliveryservices/sslkeys/add
```

is definitely not working.

mitchell852:
and this works too

```
DS: testqa, ID: 372, ACTION: Deleted SSL keys
```

i'll update the issue

Issue #4461, opened 2020-03-03
API 1.x endpoint calls need to be updated to 2.0 


ericholguin:
## I'm submitting a ...

- [ ] bug report
- [ ] new feature / enhancement request
- [X] improvement request (usability, performance, tech debt, etc.)
- [ ] other 

## Traffic Control components affected ...

- [X] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [X] Traffic Ops
- [ ] Traffic Ops ORT
- [X] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

The following code needs to be updated to use 2.0:

https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/test/router/server/server.go#L59
https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/test/router/server/server.go#L67
https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/traffic_ops/install/bin/copy-pgdata.sh#L59
https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/traffic_ops/install/bin/copy-pgdata.sh#L82
https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/traffic_ops/install/bin/copy-pgdata.sh#L208
https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/traffic_ops/app/bin/checks/ToAutorenewCerts.pl#L83
https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/traffic_ops/traffic_ops_golang/trafficstats/util_test.go#L54
https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go#L365
https://github.com/apache/trafficcontrol/blob/4b3d01d7855b72762eda171bb2a7335a0d155d55/traffic_ops/traffic_ops_golang/invalidationjobs/userinvalidationjobs.go#L137

[...]

PR #5539, opened 2021-02-18
Remove api v1

ocket8888:

This PR removes API version 1 entirely, including documentation and tests.

- Documentation
- Traffic Ops Client (Go)
- Traffic Ops
- CI tests

verify existing API tests pass for all versions greater than 1, make a request to APIv1 and verify it doesn't succeed.

- [x] Tests are unnecessary


ocket8888:
This can't be done until #5541 is closed

ocket8888:
> The tests all pass now, but we should not merge it until #5377 is done.

I made this not a draft because someone had told me that was done... I'll investigate to see if #5377 can be closed

zrhoffman (changes requested):
When grepping the repo for `api/1`, there are still matches in several files:

* cache-config/t3c-request/getdata/getdata.go
* cache-config/t3cutil/getdata.go
* docs/source/development/debugging.rst
* grove/grovetccfg/grovetccfg.go
* infrastructure/cdn-in-a-box/ort/traffic_ops_ort/to_api.py
* infrastructure/docker/README.md
* infrastructure/docker/traffic_monitor/run.sh
* infrastructure/docker/traffic_ops/run.sh
* infrastructure/docker/traffic_router/run.sh
* infrastructure/docker/traffic_server_edge/run.sh
* infrastructure/docker/traffic_server_mid/run.sh
* infrastructure/docker/traffic_stats/run.sh
* infrastructure/docker/traffic_vault/run.sh
* lib/go-tc/federation.go
* test/router/js/load-test.jsx
* test/router/server/server.go
* test/traffic_ops_cfg/cfg_test.pl
* traffic_control/clients/python/trafficops/restapi.py
* traffic_control/clients/python/trafficops/tosession.py
* traffic_ops/install/bin/copy-pgdata.sh
* traffic_ops/testing/api/v2/cdnfederations_test.go
* traffic_ops/testing/api/v3/cdnfederations_test.go
* traffic_ops/testing/api/v4/cdnfederations_test.go
* traffic_ops/traffic_ops_golang/deliveryservice/consistenthash/consistenthash.go
* traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
* traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
* traffic_ops/traffic_ops_golang/login/logout_test.go
* traffic_ops/traffic_ops_golang/routing/routing_test.go
* traffic_ops/traffic_ops_golang/swaggerdocs/v13/docs.go
* traffic_ops/traffic_ops_golang/trafficstats/util_test.go
* traffic_portal/app/src/scripts/config.js

ocket8888:
Usages in `cache-config/` are in GoDoc comments - it looks safe to change them since the information they're conveying isn't actually specific to that API version. I can do that if you want, but it shouldn't affect anything.

Usage in docs example has been updated

Usage in the ORT.py package is irrelevant, since that is due for removal. Pointless to change now, since it can't possibly work now that `atstccfg` has been removed, and whatever I do will be deleted anyway.

Usage in Grove is in a comment that wouldn't even appear in package documentation. I can get rid of it if you want, but it's not affecting anything.

I don't tend to touch anything in `infrastructure/docker` (except in the `build` subdirectory) because those Dockerfiles are archaic, they're parsing that API output in a way that may not be trivial (e.g. getting/setting server IP addresses), and their use is nowhere recommended or even documented. I don't even know if those work. I can change it, but it might break things in weird and unexpected ways, which doesn't seem worth it for something we don't even really support.

Usage in `lib/go-tc` is in GoDoc comments, which I can change if you want.

Usage in `load-test.jsx` is not referring to a Traffic Ops host - you can tell because it's passing a hostname for a Traffic Ops instance as a query parameter. `test/router/server.go` is the server serving that API, so the only real change necessary should be in there. But on the other hand - that test suite doesn't work. All browsers I've tried (Firefox, Chrome, Opera, all at latest versions) will refuse to load the `.jsx` file on the grounds of CORS request failure. If those tests are intended to work, they'll need to be reworked anyway, so the changes could just take place then.

`cfg_test.pl` is a script that, I believe, has not worked for longer than your involvement in the project. Possibly longer than mine. It was replaced by the `compare` tool, which has since been removed because it no longer serves any purpose. This script should not exist.

Usage in restapi.py is for a generic API, an API base URL containing an `api/1` segment is perfectly valid in that context. In `tosession.py` it's in a comment on an example - which itself works afaik - and doesn't appear in any documentation. I can change it if you want, but it's not affecting anything.

Usage in `copy-pgdata.sh` has been updated. I thought that was for migrating from a mysql database to postgresql, but the comment header implies otherwise. Regardless, the usages are trivial enough to change easily.

Usage in `traffic_ops/testing/` is in non-documenting comments, which I can change if you want but it's not affecting anything.

Usage in `traffic_ops/traffic_ops_golang/` in GoDoc comments has been updated, as it was incorrect - except for one instance, which is still technically correct. One usage as a hard-coded value in a response header and one usage in a test have both been updated. The usages in the routing test are specifically checking for APIv1 not existing, and so should remain. The URL used in the DS Stats test is arbitrary, I can change it, but it won't be any more or less valid; only the headers on the request actually matter.

Swagger is not actually supported by Traffic Ops. The `traffic_ops/traffic_ops_golang/swaggerdocs/` directory and any and all sub-packages thereof should be deleted, in my opinion. Unless someone is going to make them work, in which case they ought to update them to be correct themselves. But I don't think that's actually possible, in general, without significant code changes and some way to integrate the output with our existing documentation system (which doesn't exist afaik).

Usage in TP was previously removed, but must have been re-introduced during rebase with the commit that added/updated the `stable` API version.

zrhoffman:
> Usages in cache-config/ are in GoDoc comments

> Usage in lib/go-tc is in GoDoc comments

> In tosession.py it's in a comment on an example - which itself works afaik - and doesn't appear in any documentation.

> Usage in traffic_ops/testing/ is in non-documenting comments

> The URL used in the DS Stats test is arbitrary

Updating all of these makes sense to me, IMO.

> cfg_test.pl is a script that, I believe, has not worked for longer than your involvement in the project. Possibly longer than mine. It was replaced by the compare tool, which has since been removed because it no longer serves any purpose. This script should not exist.

Sounds like it could have been removed in the Perlectomy. Removing it with the rest of API v1 is the next best thing, right?

---


> Usage in Grove is in a comment that wouldn't even appear in package documentation. I can get rid of it if you want, but it's not affecting anything.

https://github.com/apache/trafficcontrol/blob/1c72915665192d9ffa4d08e5ba202e044b38c2be/grove/grovetccfg/grovetccfg.go#L195

I'm seeing more API v1 stuff lower down:
https://github.com/apache/trafficcontrol/blob/1c72915665192d9ffa4d08e5ba202e044b38c2be/grove/grovetccfg/grovetccfg.go#L232-L265

Removing the Grove API v1 stuff would make more sense in a separate PR.

> Usage in the ORT.py package is irrelevant, since that is due for removal.

> I don't tend to touch anything in infrastructure/docker (except in the build subdirectory) because those Dockerfiles are archaic, they're parsing that API output in a way that may not be trivial (e.g. getting/setting server IP addresses), and their use is nowhere recommended or even documented.

> Usage in load-test.jsx is not referring to a Traffic Ops host - you can tell because it's passing a hostname for a Traffic Ops instance as a query parameter. test/router/server.go is the server serving that API, so the only real change necessary should be in there. But on the other hand - that test suite doesn't work. All browsers I've tried (Firefox, Chrome, Opera, all at latest versions) will refuse to load the .jsx file on the grounds of CORS request failure. If those tests are intended to work, they'll need to be reworked anyway, so the changes could just take place then.

> Swagger is not actually supported by Traffic Ops. The traffic_ops/traffic_ops_golang/swaggerdocs/ directory and any and all sub-packages thereof should be deleted, in my opinion. Unless someone is going to make them work, in which case they ought to update them to be correct themselves.

Agreed that taking care of these outside of #5339 makes sense.

zrhoffman:
Oops, missed a response.

> Usage in restapi.py is for a generic API, an API base URL containing an `api/1` segment is perfectly valid in that context. In `tosession.py` it's in a comment on an example - which itself works afaik - and doesn't appear in any documentation. I can change it if you want, but it's not affecting anything.

https://github.com/apache/trafficcontrol/blob/b64fcd5bf675c73ea1d085ee40245fcbd92e0df5/traffic_control/clients/python/trafficops/restapi.py#L141
https://github.com/apache/trafficcontrol/blob/b64fcd5bf675c73ea1d085ee40245fcbd92e0df5/traffic_control/clients/python/trafficops/restapi.py#L147

Updating these makes sense to me, too.

ocket8888:
The parameter documentation for `api_base_path` in `restapi.py` doesn't need to be updated. That is not referring to the Traffic Ops API.

zrhoffman:
> The parameter documentation for `api_base_path` in `restapi.py` doesn't need to be updated. That is not referring to the Traffic Ops API.

You are referring to

https://github.com/apache/trafficcontrol/blob/b64fcd5bf675c73ea1d085ee40245fcbd92e0df5/traffic_control/clients/python/trafficops/restapi.py#L141-L148

?

ocket8888:
Yes. `RestApiSession` is a class that abstracts making request to an arbitrary REST API, it is not specific to Traffic Ops in any way. TO-specific code is in `tosession.py`.

```python3
class RestApiSession(object):
	"""
	This class represents a login session with a generic REST API server. It provides base
	functionality inherited by :class:`TOSession`.
	"""
	...
```

ocket8888:
> I'm seeing more API v1 stuff lower down:

I don't know what I can do about that. The comment says to remove that block when the API is updated to 1.3, but we're moving past that here. Plus from what I can tell the information it's gathering is actually necessary for the utility to do its work, so idk how I can remove it without breaking everything. It seems like Grove may need to be refactored, only @rob05c could possibly know for sure.

zrhoffman:
> The parameter documentation for `api_base_path` in `restapi.py` doesn't need to be updated. That is not referring to the Traffic Ops API.

Changing `1.2` to a version that Traffic Ops has never used still seems like an improvement, even if that means changing it to a version Traffic Ops will never use.

ocket8888:
> Changing `1.2` to a version that we have never removed from Traffic Ops still seems like an improvement, even if that means changing it to a version Traffic Ops will never use.

If we do change it, it should specifically be to a version Traffic Ops would never use. I don't want this thing that has nothing to do with Traffic Ops to be a part of the upkeep of eliminating a Traffic Ops API version that happens to have path segments that look similar to it. So if we're going to change it, it should be to a version that isn't a valid TO API version (getting rid of it wouldn't be good, it's including an example that shows an API version in the base path segment in addition to an example that doesn't so that the reader understands that they may or may not include a version in the base path at their discretion) like, say, `/api/v1.2`?

zrhoffman:
> [I]f we're going to change it, it should be to a version that isn't a valid TO API version (getting rid of it wouldn't be good, it's including an example that shows an API version in the base path segment in addition to an example that doesn't so that the reader understands that they may or may not include a version in the base path at their discretion) like, say, `/api/v1.2`?

IMO it still looks similar to a TO API version. `/api/latest`?

zrhoffman (changes requested):
There are API v1 versions for versioned structs.

```shell
grep -rn ^$'\t''[A-Z][A-Za-z0-9_-]\+V[0-9]\{2\}' lib/go-tc | sort
```

Can the v1 versions of versioned structs be combined into the highest non-v1 version?

https://github.com/apache/trafficcontrol/blob/f07cf9211abfffc8c6c9df57ee69d3310ef7c256/lib/go-tc/cdns_dnssec.go#L74-L89

https://github.com/apache/trafficcontrol/blob/f07cf9211abfffc8c6c9df57ee69d3310ef7c256/lib/go-tc/cdns_dnssec.go#L95-L104

https://github.com/apache/trafficcontrol/blob/f07cf9211abfffc8c6c9df57ee69d3310ef7c256/lib/go-tc/deliveryservices.go#L106-L181

https://github.com/apache/trafficcontrol/blob/f07cf9211abfffc8c6c9df57ee69d3310ef7c256/lib/go-tc/deliveryservices.go#L383-L686

https://github.com/apache/trafficcontrol/blob/f07cf9211abfffc8c6c9df57ee69d3310ef7c256/lib/go-tc/roles.go#L41-L72

ocket8888:
I can't change those without making breaking changes to structures that are still in use after the removal of APIv1

zrhoffman (changes requested):
Can `V14` be updated to `V20` here? I see no versioned struct in use.
https://github.com/apache/trafficcontrol/blob/f07cf9211abfffc8c6c9df57ee69d3310ef7c256/traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go#L109-L110

The API v1 reference should be removed:
https://github.com/apache/trafficcontrol/blob/f07cf9211abfffc8c6c9df57ee69d3310ef7c256/traffic_control/clients/README.md#L11-L12

`tc.LegacyInterfaceDetails` is now only for API v2:
https://github.com/apache/trafficcontrol/blob/f07cf9211abfffc8c6c9df57ee69d3310ef7c256/lib/go-tc/servers.go#L176-L177

zrhoffman:
The [`traffic_router/core/src/test/resources/api/1.3`](https://github.com/apache/trafficcontrol/tree/f07cf9211a/traffic_router/core/src/test/resources/api/1.3) directory can be removed.

traffic_control/clients/README.md
```diff
@@ -8,8 +8,11 @@ There are two client libraries supported:
 * [Documentation](https://github.com/apache/trafficcontrol/tree/master/traffic_control/clients/python/trafficops)
 
 ## Golang
-### TO API v1.5 _(Deprecated)_
-* [Documentation](https://github.com/apache/trafficcontrol/tree/master/traffic_ops/v1-client)
+### TO API v2 _(Deprecated)_
+* [Documentation](https://github.com/apache/trafficcontrol/tree/master/traffic_ops/client)
```
zrhoffman, line 12: Should be `v2-client`

traffic_control/clients/README.md
```diff
@@ -8,8 +8,11 @@ There are two client libraries supported:
...
+
+### TO API v3 _(Deprecated)_
+* [Documentation](https://github.com/apache/trafficcontrol/tree/master/traffic_ops/client)
```
zrhoffman, line 15: Should be `v3-client`

traffic_control/clients/README.md
```diff
@@ -8,8 +8,11 @@ There are two client libraries supported:
 * [Documentation](https://github.com/apache/trafficcontrol/tree/master/traffic_control/clients/python/trafficops)
 
 ## Golang
-### TO API v1.5 _(Deprecated)_
-* [Documentation](https://github.com/apache/trafficcontrol/tree/master/traffic_ops/v1-client)
+### TO API v2 _(Deprecated)_
+* [Documentation](https://github.com/apache/trafficcontrol/tree/master/traffic_ops/client)
+
+### TO API v3 _(Deprecated)_
```
zrhoffman, line 14: TO API v3 is deprecated? I thought it's the "stable" version until API v2 is removed and API v5 is added.

zrhoffman (changes requested):
There is a remaining `api/1` line in restapi.py:
https://github.com/apache/trafficcontrol/blob/8a0b8553f5e22668170ddae1affe9971f0c38859/traffic_control/clients/python/trafficops/restapi.py#L147

traffic_control/clients/README.md
zrhoffman: > TO API v3 is deprecated? I thought it's the "stable" version until API v2 is removed and API v5 is added.

zrhoffman (changes requested):
I'm only seeing [1 unaddressed comment](https://github.com/apache/trafficcontrol/pull/5539/#discussion_r667094662), the rest looks good.

traffic_portal/app/src/scripts/config.js
```diff
@@ -23,4 +23,4 @@
 
 angular.module('config', [])
 
-.constant('ENV', { api: { root:'/api/4.0/', legacy: '/api/1.5/', stable: "/api/3.1/" } });
```
zrhoffman, line 26: Why remove the `legacy` property? Can it not be incremented to API v2?

traffic_control/clients/README.md
ocket8888: My understanding was that versions 2 and 3 are deprecated once 4 is released Then with ATCv7 we can remove them and deprecate 4 assuming 5 was under development after ATCv6's release and released with ATCv7 as the new stable API.

traffic_portal/app/src/scripts/config.js
ocket8888: It could, but it wouldn't matter because that's unused. It was added to allow TP to still show cache server configuration files after moving to APIv2 - since those routes no longer work we have no further use for "legacy" afaik.

traffic_control/clients/README.md
zrhoffman: That makes sense, but in order to remove them API v2 and API v3 in ATC 7, a separate PR will need to mark all of the endpoints as deprecated in the documentation for ATC 6.

zrhoffman (approved):
Looks good. Things to address outside of this PR:

*   Remaining API v1 references:
    - Grove API v1 references
    - Swagger API v1 references
    - `test/router` API v1 references
    - `infrastructure/docker` API v1 references

    #4461 captures most of these already.

* Deprecating API v2 and API v3. Ideally, API v2 and API v3 responses would show deprecation alerts, but at a minimum, we need to deprecate the endpoints in the documentation.

PR #5547, opened 2021-02-19
Set Traffic Router to only accept TLSv1.1 and TLSv1.2

hbeatty:
## What does this PR (Pull Request) do?

The reason I did not turn off TLSv1.1 is that I had some issues getting things to work correctly with it off. The reason I did not turn on TLSv1.3 is that it is not supported in CentOS 7.

TLSv1 and TLSv1.1 have been deprecated by all the major browsers since March of 2020. We might want to look at logging the negotiated protocol if that hasn't already been done.

- [x] This PR is not related to any Issue 

## Which Traffic Control components are affected by this PR?

- Traffic Router

## What is the best way to verify this PR?

The best way I know to test it is to run this prior to and after changing the server.xml.

`docker run -it drwetter/testssl.sh:latest -p <some_traffic_router>`

Sample output:

```
Testing protocols via sockets except NPN+ALPN 

 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      not offered
 TLS 1.1    offered (deprecated)
 TLS 1.2    offered (OK)
 TLS 1.3    not offered and downgraded to a weaker protocol
 NPN/SPDY   not offered
 ALPN/HTTP2 not offered
```

## If this is a bug fix, what versions of Traffic Control are affected?
not a bug fix.

## The following criteria are ALL met by this PR

- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #5550, opened 2021-02-19
TR: fix server.xml protocols list to use '+' instead of ','

rawlinp:
Follow-up to https://github.com/apache/trafficcontrol/pull/5547: TR seems to support commas but prints the following warning when using them:
```
  The protocol [TLSv1.2] was added to the list of protocols on the
  SSLHostConfig named [example.org]. Check if a +/- prefix is missing.
```
Also, TR ignores TLSv1.3 being in the list if the backend doesn't support it, so it is safe to include by default. On systems that don't support it, the following warning is printed in the tomcat logs:
```
org.apache.tomcat.util.net.SSLUtilBase.getEnabled Some of the specified [protocols] are not supported by the SSL engine and have been skipped: [[TLSv1.3]]
```

For reference: https://tomcat.apache.org/tomcat-8.5-doc/config/http.html#SSL_Support_-_SSLHostConfig

- Traffic Router

```
docker run --rm -it drwetter/testssl.sh:latest -p <some_traffic_router>
```
Sample output:
```

 SSLv2      not offered (OK)
 SSLv3      not offered (OK)
 TLS 1      not offered
 TLS 1.2    offered (OK)
 NPN/SPDY   not offered
 ALPN/HTTP2 not offered
```

- [x] default config change, tests unnecessary
- [x] default config change, docs unnecessary

### DIFF
--- a/traffic_router/core/src/main/conf/server.xml
+++ b/traffic_router/core/src/main/conf/server.xml
@@ -40,11 +40,11 @@
 		<Connector port="3333" protocol="com.comcast.cdn.traffic_control.traffic_router.protocol.LanguidNioProtocol" maxThreads="10000"
 				   connectionTimeout="10000" mbeanPath="traffic-router:name=languidState" readyAttribute="Ready" portAttribute="ApiPort"/>
 		<Connector port="3443" protocol="com.comcast.cdn.traffic_control.traffic_router.protocol.LanguidNioProtocol" maxThreads="10000"
-				   scheme="https" secure="true" SSLEnabled="true" clientAuth="false" sslProtocol="TLS" protocols="TLSv1.1,TLSv1.2" connectionTimeout="10000"
+				   scheme="https" secure="true" SSLEnabled="true" clientAuth="false" sslProtocol="TLS" protocols="+TLSv1.1+TLSv1.2+TLSv1.3" connectionTimeout="10000"
 				   mbeanPath="traffic-router:name=languidState" readyAttribute="Ready" portAttribute="SecureApiPort" sendReasonPhrase="true"
 				   sslImplementationName="com.comcast.cdn.traffic_control.traffic_router.protocol.RouterSslImplementation"> </Connector>
 		<Connector port="443" protocol="com.comcast.cdn.traffic_control.traffic_router.protocol.LanguidNioProtocol" maxThreads="10000"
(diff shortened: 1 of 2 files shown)


traffic_router/core/src/main/conf/server.xml
```diff
@@ -40,11 +40,11 @@
...
 		<Connector port="3443" protocol="com.comcast.cdn.traffic_control.traffic_router.protocol.LanguidNioProtocol" maxThreads="10000"
-				   scheme="https" secure="true" SSLEnabled="true" clientAuth="false" sslProtocol="TLS" protocols="TLSv1.1,TLSv1.2" connectionTimeout="10000"
+				   scheme="https" secure="true" SSLEnabled="true" clientAuth="false" sslProtocol="TLS" protocols="+TLSv1.1+TLSv1.2+TLSv1.3" connectionTimeout="10000"
 				   mbeanPath="traffic-router:name=languidState" readyAttribute="Ready" portAttribute="SecureApiPort" sendReasonPhrase="true"
 				   sslImplementationName="com.comcast.cdn.traffic_control.traffic_router.protocol.RouterSslImplementation"> </Connector>
 		<Connector port="443" protocol="com.comcast.cdn.traffic_control.traffic_router.protocol.LanguidNioProtocol" maxThreads="10000"
-				   scheme="https" secure="true" SSLEnabled="true" clientAuth="false" sslProtocol="TLS" protocols="TLSv1.1,TLSv1.2" connectionTimeout="10000"
+				   scheme="https" secure="true" SSLEnabled="true" clientAuth="false" sslProtocol="TLS" protocols="+TLSv1.1+TLSv1.2+TLSv1.3" connectionTimeout="10000"
```
zrhoffman, line 47: Having commas there like `+TLSv1.1,+TLSv1.2,+TLSv1.3` or `+TLSv1.1,+TLSv1.2,+TLSv1.3,` would still help people who don't know at a glance which SSL protocol version in the list `+` refers to.

zrhoffman (changes requested):

traffic_router/core/src/main/conf/server.xml
rawlinp: Indeed, that does appear to work the same too. Fixed.

zrhoffman (approved):

PR #5551, opened 2021-02-19
Update ORT/atstccfg old client to V3

rob05c:
This updates the old client to V3.

The new client is still V3, because the v4-client isn't done yet. It
adds a TODO comment, so when v4 is ready, ORT should be changed to
use it.

It also refactors the client usage a bit, so the toreq user
doesn't have to check and fall back itself, but rather the entire
client falls back. Making it easier to use and more efficient.

The ORT Integration Test framework tests this. This should have no interface change.
No docs, no interface change.

- Traffic Ops ORT

Run ORT Integration Tests. Run ORT against latest `master` and previous release TO, verify config is generated correctly.

Not a bug fix.

## Additional Information


jrushford (approved):
Looks good to me.  I see no problems in the code with the exception that I'll have to refactor PR #5614 prior to it being merged.  We should merge this one first and I'll fix PR #5614.

I built and ran all tests, no problems found.

PR #5564, opened 2021-02-23
Python 2.7 support for Python postinstall script

zrhoffman:

* This PR makes the Python postinstall script added in #4763 compatible with Python 2.7.
* This PR also makes the Python postinstall script compatible with `/usr/lib/platform-python` (Python 3.6) on CentOS 8, which does not have the `typing` module.

- Traffic Ops

* Run the postinstall end-to-end tests against Python 2.7 on CentOS 7 after installing the test dependencies:
  ```shell
  <<'COMMANDS' docker run --rm -i -v $(pwd):/trafficcontrol -w /trafficcontrol centos:7 bash
      yum -y install epel-release file openssl &&
      yum -y install jq &&
      time traffic_ops/install/bin/postinstall.test.sh -2 &&
      echo Tests succeeded for CentOS 7
  COMMANDS
  ```

* Run the postinstall end-to-end tests against Python 3.6 on CentOS 8 after installing the test dependencies:
  ```shell
  <<'COMMANDS' docker run --rm -i -v $(pwd):/trafficcontrol -w /trafficcontrol centos:8 bash
  yum -y install epel-release file openssl &&
  yum -y install jq &&
  time traffic_ops/install/bin/postinstall.test.sh -3 -s -b /usr/libexec/platform-python &&
  echo Tests succeeded for CentOS 8
  COMMANDS
  ```

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/install/bin/postinstall.py
+++ b/traffic_ops/install/bin/postinstall.py
@@ -744,9 +744,9 @@ def setup_maxmind(maxmind_answer, root): # type: (str, str) -> None
 			universal_newlines=True
 		)
 	except subprocess.CalledProcessError as e:
-		failed_download(6)
+		failed_download(e, 6)
 	except subprocess.SubprocessError as e:
-		failed_download(6)
+		failed_download(e, 6)
 
 def exec_openssl(description, *cmd_args): # type: (str, ...) -> bool
 	"""
--- a/traffic_ops/install/bin/postinstall.py
+++ b/traffic_ops/install/bin/postinstall.py
@@ -749,7 +749,9 @@ def setup_certificates(conf, root, ops_user, ops_group): # type: (SSLConfig, str
 		with open(cdn_conf_path) as conf_file:
 			cdn_conf = json.load(conf_file)
 	except (OSError, json.JSONDecodeError) as e:
-		raise OSError("reading {cdn_conf_path}: {e}".format(cdn_conf_path=cdn_conf_path, e=e)) from e
+		exception = OSError("reading {cdn_conf_path}: {e}".format(cdn_conf_path=cdn_conf_path, e=e))
+		exception.__cause__ = e
+		raise exception
 
 	if (
 		not isinstance(cdn_conf, dict) or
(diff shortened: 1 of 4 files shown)


traffic_ops/install/bin/postinstall.py
```diff
@@ -560,48 +703,52 @@ def unmarshal_config(dct: dict) -> Dict[str, List[Question]]:
...
-	if maxmind_answer.casefold() not in {'y', 'yes'}:
+	if maxmind_answer.lower() not in {'y', 'yes'}:
 		logging.info("Not downloading Maxmind data")
 		return
 
 	os.chdir(os.path.join(root, 'opt/traffic_ops/app/public/routing'))
 
+	def failed_download(ip_version):  # type: (int) -> None
+		logging.error("Failed to download MaxMind data")
+		logging.debug("(ipv%d) Exception: %s", ip_version, e)
```
ocket8888, line 719: `e` is not defined in the scope of this function

traffic_ops/install/bin/postinstall.py
```diff
@@ -1167,34 +1338,41 @@ def main(
...
+				"Use the script `/opt/traffic_ops/install/bin/todb_bootstrap.sh` "
 				"on the db server to create it and run `postinstall` again."
 			)
+
+		try:
+			setup_database_data(conn_str, admin_conf, paramconf, root_dir)
+		except (OSError, subprocess.CalledProcessError)as e:
+			db_connect_failed()
+			return -1
+		except subprocess.SubprocessError as e:
```
ocket8888, line 1354: Why split these `excepts` if they have the same handler?

zrhoffman: `subprocess.SubprocessError` is not defined in Python 2. Separating them like that prevent
```
AttributeError: 'module' object has no attribute 'SubprocessError'
```
in Python 2

traffic_ops/install/bin/postinstall.test.sh
```diff
@@ -15,18 +15,72 @@
...
+while getopts :23sb: opt; do
+	case "$opt" in
+		2)	python_version=2;;
+		3)	python_version=3;;
+		s)	skip_python2=true;;
+		b)	python_bin="$OPTARG";;
+		*)	echo "Invalid flag received: ${OPTARG}" >&2
+			exit 1;;
+	esac;
+done;
```
ocket8888, line 32: I just realized; this used to have no options. Now that there _are_ options, it'd be nice to have a `-h` and/or `-?` to pass in order to see them.

traffic_ops/install/bin/postinstall.py
ocket8888: really? I would expect that to still give an error. I suppose if the test passes that means it's fine.

traffic_ops/install/bin/postinstall.py
zrhoffman: Passing `e` to the function in d94b755af3

traffic_ops/install/bin/postinstall.py
```diff
@@ -560,48 +703,52 @@ def unmarshal_config(dct: dict) -> Dict[str, List[Question]]:
...
-		subprocess.run(
+		subprocess.check_call(
 			cmd,
 			stderr=subprocess.PIPE,
 			stdout=subprocess.PIPE,
-			check=True,
 			universal_newlines=True
 		)
+	except subprocess.CalledProcessError as e:
+		failed_download(6)
```
ocket8888, line 747: passing `6` as `Exception` - missing those type hints already? :P

traffic_ops/install/bin/postinstall.py
```diff
@@ -560,48 +703,52 @@ def unmarshal_config(dct: dict) -> Dict[str, List[Question]]:
...
 	except subprocess.SubprocessError as e:
-		logging.error("Failed to download MaxMind data")
-		logging.debug("(ipv6) Exception: %s", e)
+		failed_download(6)
```
ocket8888, line 749: passing `6` as `Exception`

traffic_ops/install/bin/postinstall.py
zrhoffman: PyCharm gave me the warnings, I just didn't scroll down.
Passing 6 as an `int` in ab6881f0fb

traffic_ops/install/bin/postinstall.py
zrhoffman: Fixed in ab6881f0fb

traffic_ops/install/bin/postinstall.test.sh
zrhoffman: Added help string in 1e7fd5752d

ocket8888 (approved):
Appears to still work, tests passing.

PR #5569, opened 2021-02-23
Replace outdated Traffic Control graphic

ocket8888:

This PR updates the diagram in the "Overview" documentation section, removing "Traffic Logs" and being a little more specific about the connections between components

- Documentation

Build the documentation (you could just look at the image(s), but how it looks on the rendered document is also important) and check out the new graphic on the "Overview" Introduction page. Make sure it makes sense, and is the least amount of confusing possible while still being accurate and helpful. That was not an easy balance to strike.

- [x] There are no documentation tests.


rawlinp:
Why remove the "HTTP 302 Redirect" arrow and "DNSSEC shield?"

ocket8888:
The DNSSEC shield felt like too much detail. It's just a feature TR supports, and it could also be in place between the client and their DNS server. It's just like an implementation detail at every DNS level.

The HTTP 302 Redirect was an accident; I'll put that back in.

TaylorCFrey (commented):
Seems to make sense with what I'm gathering and learning. I suppose my only observation is that the REST API and TO are one and the same, correct? This almost makes it seem like there is some sort of API gateway or something acting as a REST API between all the components.

ocket8888:
I guess I just really liked the gear, and it was hard to extract from the PNG so it felt like a waste to throw it away

TaylorCFrey (approved):

mitchell852:
> I guess I just really liked the gear, and it was hard to extract from the PNG so it felt like a waste to throw it away

i dunno. might be better to toss it rather than cause confusion. What if you change it from "rest api" to "traffic ops api" and leave the gear?

docs/source/overview/images/traffic.control.overview.svg
```diff
@@ -0,0 +1,2110 @@
...
+     inkscape:window-height="2073"
+     id="namedview2045"
+     showgrid="false"
+     inkscape:zoom="0.5026132"
+     inkscape:cx="2458.2048"
+     inkscape:cy="2212.0306"
+     inkscape:window-x="0"
+     inkscape:window-y="0"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg2043" />
```
zrhoffman, line 66: SVG will omit the `<sodipodi:namedview>` tag if you save it as a plain SVG instead of an Inkscape SVG

zrhoffman (changes requested):

mitchell852:
what this supposed to represent?

image

docs/source/overview/images/traffic.control.overview.svg
ocket8888: It was, originally, an "Optimized SVG" without any indentation or line breaks. I think I saved it on accident after exporting the PNG.

tl;dr: yes.

ocket8888:
I think it's supposed to be someone who contracts out the CDN owner to run a Delivery Service for them. The person who owns the origin.

mitchell852:
> I think it's supposed to be someone who contracts out the CDN owner to run a Delivery Service for them. The person who owns the origin.

right but the monitor icon seem unnecessary this seems sufficient

image

dneuman64 (commented):
A few things I noticed: 
- Traffic Monitor does health and stat polling, not just health
- Traffic Stats gets cache and DS statistics, not so much health 
- nitpick: why is the line so thick from TP to TO API?  
- TR gets config and health data from TM, might be worth noting
- Should we add an ORT/T3C thing on the mids/edges?  It is a TC component but is not called out anywhere on here

ocket8888:
> TR gets config and health data from TM, might be worth noting

That health data is just part of its configuration though. It's separate HTTP requests, but the upshot is that TR does what it does based on information from  TO and TM that dictate its behavior.

> Should we add an ORT/T3C thing on the mids/edges? It is a TC component but is not called out anywhere on here

I guess in this case ORT would be the lines? What if I change the box title from "Traffic Server" to "Traffic Server and ORT"?

mitchell852 (approved):
i like it!

zrhoffman (approved):
Graphic makes sense to me and `<sodipodi:namedview>` tag is gone

dneuman64 (approved):

PR #5570, opened 2021-02-24
Update the Python TO client to API 3.0

rawlinp:
Update the Python TO client to API 3.0. Also, fix up the topologies methods.

Write a script that uses the python TO client (or use it interactively). Make requests using the various methods, verify they work successfully.

- [x] python TO client doesn't have tests
- [x] no docs necessary


ocket8888 (approved):
LGTM

rawlinp:
Something seems up w/ the CiaB origin/ats possibly. The logs show successful 302 but then ats returns 502.

ocket8888:
Yeah, I was just running it locally to see if I can find the problem

ocket8888:
Found it. Line 49 in `infrastructure/cdn-in-a-box/ort/traffic_ops_ort/to_api.py`. It's setting the ORT API version to 2.0, but if the client only supports 3.0 then that needs to get updated.

rawlinp:
Ah, I think I know the issue. Python ORT is setting the version to 2.0, which is likely causing an exception since 2.0 is "not supported".

rawlinp:
JINX 1, 2, 3!

Issue #1624, opened 2017-12-07
TO: LUA Plugin Script Configuration Improvements


ryandurfey:
Right now, the LUA script deployment process is painful. It involves having config mgmt deploy the script to the (hopefully correct) edge nodes, and then adding a raw remap rule to the DS to enable the lua plugin and point to that script which has already been deployed.
We need to make this process easier so that LUA scripts can be added to services using the standard configuration process.

ashish-timilsina:
@ryandurfey 
@MattMills 
@ashish-timilsina

mitchell852:
@rob05c - any thoughts on this? how would this relate to t3c?

PR #5574, opened 2021-02-25
Fix ORT to reload traffic server when lua scripts are added or changed.  

jpappa200:

This PR will have ort touch remap.config and run traffic_ctl config reload when .lua scripts are added or changed. It helps with issue #1624, but doesn't completely address that issue. 
 

- Traffic Ops ORT

Install updated ORT on an edge, add/make changes to a lua script, and run ORT in synds mode.

This is not a bug fix.


dneuman64 (changes requested):
Can you please add a changelog entry?
I also think this should be added to t3c, do you mind making that change and adding it to this PR?

rob05c (approved):
Looks good! I tested, works as expected. Changing a lua file results in a server reload.

I'm approving, with the caveats Neuman mentioned: before merging a changelog entry needs added, and either t3c added or a separate Issue created to track that.

rob05c:
Ok, I tested `t3c` in `master`, and it's already doing this. I tested, changed only a lua file, saw `t3c` do an ORT reload.

So I'll merge as soon as the conflict is fixed.

dneuman64 (approved):

PR #5592, opened 2021-03-01
DNSSEC keys should not be created with ANY_MAP services 

### LINKED ISSUE #2272: Only create/update/modify DNSSEC keys for DS types that are routed on the CDN
Create DNSSEC keys only for DS types that are routed on the CDN, such as any HTTP or DNS type. ANY_MAP, or other types that are not routed on the CDN should not have DNSSEC keys created for them.

jpappa200:

- [X] This PR fixes #2272  is not related to any Issue 

- Traffic Ops

Since we do not have a lab environment with DNSSEC enabled. I used CDN in a Box, enabled DNSSEC, tried to create a regular routed service and the creation failed with DNSSEC key errors, but any_map services were created without error. 
Spun up CDN in a Box from master, with DNSSEC enabled, and creation of both routed and any_map services fail with DNSSEC key error. 

### DIFF
--- a/lib/go-tc/enum.go
+++ b/lib/go-tc/enum.go
@@ -597,11 +597,7 @@ func DSTypeFromString(s string) DSType {
 
 //IsAnyMap returns weather the DSType is an ANY_MAP category.
 func (t DSType) IsAnyMap() bool {
-	switch t {
-	case DSTypeAnyMap:
-		return true
-	}
-	return false
+	return t == DSTypeAnyMap
 }
 
 // IsHTTP returns whether the DSType is an HTTP category.
--- a/lib/go-tc/enum.go
+++ b/lib/go-tc/enum.go
@@ -595,7 +595,7 @@ func DSTypeFromString(s string) DSType {
 	}
 }
 
-// IsAnyMap returns weather the DSType is an ANY_MAP service.
+// IsAnyMap returns whether the DSType is an ANY_MAP service.
 func (t DSType) IsAnyMap() bool {
 	return t == DSTypeAnyMap
 }
--- a/lib/go-tc/enum.go
+++ b/lib/go-tc/enum.go
@@ -595,7 +595,7 @@ func DSTypeFromString(s string) DSType {
 	}
 }
 
-//IsAnyMap returns weather the DSType is an ANY_MAP category.
+//IsAnyMap returns weather the DSType is an ANY_MAP service.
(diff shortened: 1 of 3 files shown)


lib/go-tc/enum.go
```diff
@@ -595,6 +595,11 @@ func DSTypeFromString(s string) DSType {
 	}
 }
 
+//IsAnyMap returns weather the DSType is an ANY_MAP service.
```
rawlinp, line 598: sp: weather -> whether, and I think godoc likes a space after the `//`

rawlinp (changes requested):
Just some minor nitpicks but this looks good otherwise.

lib/go-tc/enum.go
jpappa200: Fixed spelling error and added white space after //

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -479,7 +479,7 @@ func createV40(w http.ResponseWriter, r *http.Request, inf *api.APIInfo, dsV40 t
 		return nil, http.StatusInternalServerError, nil, errors.New("ensuring ds parameters:: " + err.Error())
 	}
 
-	if dnssecEnabled {
+	if dnssecEnabled && !ds.Type.IsAnyMap() {
```
rawlinp, line 482: Actually, looking at the DNSSEC refresh code here: https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go#L195

I think we should refactor that condition into a new method 
```
func (t DSType) UsesDNSSECKeys() bool {
    return t.IsHTTP() || t.IsSteering() || t.IsDNS()
}
```
and use it in both places (i.e. `!t.UsesDNSSECKeys()` here and in dnssecrefresh.go).

The idea being that if we add a new delivery service type, we have to explicitly decide whether or not it uses dnssec keys, rather than getting dnssec keys by default.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
jpappa200: Changes have been made to the method and refresh code.

rawlinp (approved):
Looks good!

PR #5552, opened 2021-02-21
Add TO client v4 fallback

rob05c:
Also moves duplicate fallback and other duplicate client code
to a shared lib.

No new tests, existing TO API Tests cover these changes.
No docs, no interface change.
No changelog, client fallback already has a changelog entry.

- [x] This PR fixes #5518 OR is not related to any Issue

## Which Traffic Control components are affected by this PR?
- Traffic Control Client Go

## What is the best way to verify this PR?
Run tests. Make requests with the client, verify they work as expected.

## If this is a bug fix, what versions of Traffic Control are affected?
Not a bug fix.

## The following criteria are ALL met by this PR
- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

## Additional Information

PR #5603, opened 2021-03-03
Removing non existent route /tools/flash_and_close/{message}

### LINKED ISSUE #5597: GET /tools/write_crconfig/{cdn} redirects to removed /tools/flash_and_close/{message} endpoint
## I'm submitting a ...

-  bug report

-  Traffic Ops

## Current behavior:

`GET /tools/write_crconfig/{cdn}` (endpoint includes no API version) redirects to `/tools/flash_and_close/{message}` (also includes no API version).

## Expected behavior:

Because `/tools/flash_and_close/{message}` was only implemented in Traffic Ops Perl and #5209 removed Traffic Ops Perl, `GET /tools/write_crconfig/{cdn}` should instead redirect to `/api/1.1/cdns/{name}/snapshot`:
https://github.com/apache/trafficcontrol/blob/f21adfd6e76d7d359ac174e2fe168b69aae2b1ba/traffic_ops/traffic_ops_golang/routing/routes.go#L1616

srijeet0406:

- [x] This PR fixes #5597  

- Traffic Ops

Make sure all the unit tests and API tests pass

- master

- [x] Tests are unnecessary because I'm removing a non existent route
- [x] No documentation is necessary

### DIFF
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	client "github.com/apache/trafficcontrol/traffic_ops/v1-client"
 	"net/http"
 	"strconv"
 	"time"
@@ -35,6 +34,7 @@ import (
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/monitoring"
+	client "github.com/apache/trafficcontrol/traffic_ops/v1-client"
 )
 
 // Handler creates and serves the CRConfig from the raw SQL data.
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -256,9 +256,9 @@ func snapshotHandler(w http.ResponseWriter, r *http.Request, deprecated bool) {
 
 // SnapshotOldGUIHandler creates the CRConfig JSON and writes it to the snapshot table in the database. The response emulates the old Perl UI function. This should go away when the old Perl UI ceases to exist.
 func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
-	inf, userErr, sysErr, code := api.NewInfo(r, []string{"cdn"}, nil)
+	inf, userErr, sysErr, errCode := api.NewInfo(r, []string{"cdn"}, nil)
 	if userErr != nil || sysErr != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, code, userErr, errors.New(r.RemoteAddr+" unable to get info from request: "+sysErr.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, errors.New(r.RemoteAddr+" unable to get info from request: "+sysErr.Error()))
 		return
 	}
 	defer inf.Close()
@@ -273,7 +273,7 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	// We never store tm_path, even though low API versions show it in responses.
 	crConfig, err := Make(inf.Tx.Tx, cdn, inf.User.UserName, r.Host, inf.Config.Version, inf.Config.CRConfigUseRequestHost, false)
 	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, err, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
 		return
 	}
 
@@ -284,7 +284,7 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := Snapshot(inf.Tx.Tx, crConfig, tm); err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, err, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
 		return
 	}
 
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	client "github.com/apache/trafficcontrol/traffic_ops/v1-client"
 	"net/http"
 	"strconv"
 	"time"
@@ -293,5 +294,5 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig performed for "+cdn, inf.User, inf.Tx.Tx)
-	http.Redirect(w, r, "client.API_v13_CDNs"+cdn+"/snapshot", http.StatusFound)
+	http.Redirect(w, r, client.API_v13_CDNs+"/"+cdn+"/snapshot", http.StatusFound)
 }
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	client "github.com/apache/trafficcontrol/traffic_ops/v1-client"
 	"net/http"
 	"strconv"
 	"time"
@@ -293,5 +294,5 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig performed for "+cdn, inf.User, inf.Tx.Tx)
-	http.Redirect(w, r, "/api/1.1/cdns/"+cdn+"/snapshot", http.StatusFound)
+	http.Redirect(w, r, "/api/1.1/cdns/"+client.API_v13_CDNs+"/snapshot", http.StatusFound)
 }
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -270,10 +270,15 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	cdn := inf.Params["cdn"]
+	_, exists, _ := dbhelpers.GetCDNIDFromName(inf.Tx.Tx, tc.CDNName(cdn))
+	if !exists {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New("unable to find the CDN: "+cdn), nil)
+		return
+	}
 	// We never store tm_path, even though low API versions show it in responses.
 	crConfig, err := Make(inf.Tx.Tx, cdn, inf.User.UserName, r.Host, inf.Config.Version, inf.Config.CRConfigUseRequestHost, false)
 	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), nil)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
 		return
 	}
 
@@ -284,7 +289,7 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := Snapshot(inf.Tx.Tx, crConfig, tm); err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), nil)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
 		return
 	}
 
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -23,7 +23,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	client "github.com/apache/trafficcontrol/traffic_ops/v1-client"
 	"net/http"
 	"strconv"
 	"time"
@@ -294,5 +293,5 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig performed for "+cdn, inf.User, inf.Tx.Tx)
-	http.Redirect(w, r, "/api/1.1/cdns/"+client.API_v13_CDNs+"/snapshot", http.StatusFound)
+	http.Redirect(w, r, "client.API_v13_CDNs"+cdn+"/snapshot", http.StatusFound)
 }
--- a/traffic_ops/traffic_ops_golang/crconfig/handler.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/handler.go
@@ -273,7 +273,7 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	// We never store tm_path, even though low API versions show it in responses.
 	crConfig, err := Make(inf.Tx.Tx, cdn, inf.User.UserName, r.Host, inf.Config.Version, inf.Config.CRConfigUseRequestHost, false)
 	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), nil)
 		return
 	}
 
@@ -284,7 +284,7 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := Snapshot(inf.Tx.Tx, crConfig, tm); err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), nil)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -296,11 +303,11 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
...
-	http.Redirect(w, r, "/tools/flash_and_close/"+url.PathEscape("Successfully wrote the CRConfig.json!"), http.StatusFound)
+	http.Redirect(w, r, "/api/1.1/cdns/"+strconv.Itoa(cdnID)+"/snapshot", http.StatusFound)
 }
 
-func writePerlHTMLErr(w http.ResponseWriter, r *http.Request, tx *sql.Tx, logErr error, err error) {
+func writePerlHTMLErr(w http.ResponseWriter, r *http.Request, tx *sql.Tx, logErr error, err error, cdnId int) {
 	log.Errorln(logErr.Error())
 	tx.Rollback()
-	http.Redirect(w, r, "/tools/flash_and_close/"+url.PathEscape("Error: "+err.Error()), http.StatusFound)
+	http.Redirect(w, r, "/api/1.1/cdns/"+strconv.Itoa(cdnId)+"/snapshot", http.StatusFound)
```
zrhoffman, line 312: When there's an error that's safe for the user to see, we should show that error to the user as an alert, not redirect.

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -271,11 +270,19 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
...
-
+	cdnID, ok, err := dbhelpers.GetCDNIDFromName(inf.Tx.Tx, tc.CDNName(cdn))
+	if !ok {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusNotFound, fmt.Errorf("couldn't find CDN ID with name %s", cdn), nil)
+		return
+	}
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, err, nil)
+		return
+	}
```
zrhoffman, line 281: If no rows are returned, the error returned from `dbhelpers.GetCDNIDFromName()`, the error returned is `nil`:
https://github.com/apache/trafficcontrol/blob/ae2ced905d223789c818831ead1c6979c488eaf0/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go#L518-L520

So this should be a 500-level internal server error, not 400 bad request.

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -296,11 +303,11 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig performed for "+cdn, inf.User, inf.Tx.Tx)
-	http.Redirect(w, r, "/tools/flash_and_close/"+url.PathEscape("Successfully wrote the CRConfig.json!"), http.StatusFound)
+	http.Redirect(w, r, "/api/1.1/cdns/"+strconv.Itoa(cdnID)+"/snapshot", http.StatusFound)
```
zrhoffman, line 306: The path parameter is the CDN name, not the CDN ID

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -303,11 +293,5 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig performed for "+cdn, inf.User, inf.Tx.Tx)
-	http.Redirect(w, r, "/api/1.1/cdns/"+strconv.Itoa(cdnID)+"/snapshot", http.StatusFound)
-}
-
-func writePerlHTMLErr(w http.ResponseWriter, r *http.Request, tx *sql.Tx, logErr error, err error, cdnId int) {
-	log.Errorln(logErr.Error())
-	tx.Rollback()
-	http.Redirect(w, r, "/api/1.1/cdns/"+strconv.Itoa(cdnId)+"/snapshot", http.StatusFound)
+	http.Redirect(w, r, "/api/1.1/cdns/"+cdn+"/snapshot", http.StatusFound)
```
zrhoffman, line 296: This can use `client.API_v13_CDNs` to avoid hard-coding `"/api/1.1/cdns/"`

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	client "github.com/apache/trafficcontrol/traffic_ops/v1-client"
```
zrhoffman, line 26: This import should be grouped with the other ATC imports

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -296,11 +293,5 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig performed for "+cdn, inf.User, inf.Tx.Tx)
-	http.Redirect(w, r, "/tools/flash_and_close/"+url.PathEscape("Successfully wrote the CRConfig.json!"), http.StatusFound)
-}
-
-func writePerlHTMLErr(w http.ResponseWriter, r *http.Request, tx *sql.Tx, logErr error, err error) {
-	log.Errorln(logErr.Error())
-	tx.Rollback()
-	http.Redirect(w, r, "/tools/flash_and_close/"+url.PathEscape("Error: "+err.Error()), http.StatusFound)
+	http.Redirect(w, r, "client.API_v13_CDNs"+cdn+"/snapshot", http.StatusFound)
```
zrhoffman, line 296: Constants like `client.API_v13_CDNs` should be unquoted

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/crconfig/handler.go
srijeet0406: ugh, cant seem to stop making typos

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -20,12 +20,11 @@ package crconfig
  */
 
 import (
-	"database/sql"
 	"encoding/json"
 	"errors"
 	"fmt"
+	client "github.com/apache/trafficcontrol/traffic_ops/v1-client"
```
zrhoffman, line 26: This import should still be grouped with the other ATC imports

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -286,7 +284,7 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := Snapshot(inf.Tx.Tx, crConfig, tm); err != nil {
-		writePerlHTMLErr(w, r, inf.Tx.Tx, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), err)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, err, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
```
zrhoffman, line 287: The error returned from `crconfig.Snapshot()` is 500-level error and should not be displayed to the user.

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -271,11 +270,10 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	cdn := inf.Params["cdn"]
-
 	// We never store tm_path, even though low API versions show it in responses.
 	crConfig, err := Make(inf.Tx.Tx, cdn, inf.User.UserName, r.Host, inf.Config.Version, inf.Config.CRConfigUseRequestHost, false)
 	if err != nil {
-		writePerlHTMLErr(w, r, inf.Tx.Tx, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), err)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, err, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
```
zrhoffman, line 276: The error returned from `crconfig.Make()` is 500-level error and should not be displayed to the user. Right now, requesting `GET /tools/write_crconfig/nonexistent` yields:

```
HTTP/2 500
access-control-allow-credentials: true
access-control-allow-headers: Origin, X-Requested-With, Content-Type, Accept, Set-Cookie, Cookie
access-control-allow-methods: POST,GET,OPTIONS,PUT,DELETE
access-control-allow-origin: *
content-type: application/json
set-cookie: mojolicious=...; Path=/; Expires=Fri, 05 Mar 2021 22:48:36 GMT; Max-Age=3600; HttpOnly
vary: Accept-Encoding
whole-content-sha512: 5gzjE1ZKa484M1quoUVNTeNS4mMdzk0m/3uxQizzQ+hUmBaLWvmregVOjDa2I1cVeVRPX5Mw5KTym3iJxRUBhQ==
x-server-name: traffic_ops_golang/
content-length: 123
date: Fri, 05 Mar 2021 21:48:36 GMT

{"alerts":[{"text":"Error getting CDN info: Error querying CDN domain name: sql: no rows in result set","level":"error"}]}
```

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -257,9 +256,9 @@ func snapshotHandler(w http.ResponseWriter, r *http.Request, deprecated bool) {
 
 // SnapshotOldGUIHandler creates the CRConfig JSON and writes it to the snapshot table in the database. The response emulates the old Perl UI function. This should go away when the old Perl UI ceases to exist.
 func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
-	inf, userErr, sysErr, _ := api.NewInfo(r, []string{"cdn"}, nil)
+	inf, userErr, sysErr, code := api.NewInfo(r, []string{"cdn"}, nil)
```
TaylorCFrey, line 259: NIT: The other handlers label this variable with `errCode` instead of just `code`

traffic_ops/traffic_ops_golang/crconfig/handler.go
TaylorCFrey, line 276: The method signature for `api.HandleErr` is:
```
func HandleErr(w http.ResponseWriter, r *http.Request, tx *sql.Tx, statusCode int, userErr error, sysErr error)
```
So, we don't want to pass the `err` into the Handler because it will be assumed to be a userErr. Perhaps `api.HandleErr` is a bit too general handling all types of errors. I might argue for a `api.HandleUserErr` and a `api.HandleServerErr`. But in any case, for the time being we want to pass `nil` along instead.
```
api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
```

traffic_ops/traffic_ops_golang/crconfig/handler.go
TaylorCFrey, line 287: Same here, we want to pass `nil` instead of `err` since this is a sysErr not a userErr.

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -296,11 +294,5 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig performed for "+cdn, inf.User, inf.Tx.Tx)
-	http.Redirect(w, r, "/tools/flash_and_close/"+url.PathEscape("Successfully wrote the CRConfig.json!"), http.StatusFound)
-}
-
-func writePerlHTMLErr(w http.ResponseWriter, r *http.Request, tx *sql.Tx, logErr error, err error) {
-	log.Errorln(logErr.Error())
-	tx.Rollback()
-	http.Redirect(w, r, "/tools/flash_and_close/"+url.PathEscape("Error: "+err.Error()), http.StatusFound)
+	http.Redirect(w, r, client.API_v13_CDNs+"/"+cdn+"/snapshot", http.StatusFound)
```
TaylorCFrey, line 297: So this just feels weird to me. Including the _client_ which is based on the TO api seems almost cyclical? Don't we want to simply redirect akin to:
```
http.Redirect(w, r, "/api/1.1/cdns/"+cdn+"/snapshot", http.StatusSeeOther)
```
I know we're just trying to use a const version here, but I dunno. Should we make it a const in this package so we aren't relying on a v1 client? I do like the idea, once removing v1 api and client that this will automatically be picked up, but something just itches the wrong way for me here.

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -296,11 +294,5 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	api.CreateChangeLogRawTx(api.ApiChange, "Snapshot of CRConfig performed for "+cdn, inf.User, inf.Tx.Tx)
-	http.Redirect(w, r, "/tools/flash_and_close/"+url.PathEscape("Successfully wrote the CRConfig.json!"), http.StatusFound)
-}
-
-func writePerlHTMLErr(w http.ResponseWriter, r *http.Request, tx *sql.Tx, logErr error, err error) {
-	log.Errorln(logErr.Error())
-	tx.Rollback()
```
TaylorCFrey, line 304: Interesting. I haven't seen many transaction rollbacks. Is this something we still need to account for before redirecting?

TaylorCFrey (commented):
Some thoughts and observations

traffic_ops/traffic_ops_golang/crconfig/handler.go
zrhoffman: > Should we make it a const in this package so we aren't relying on a v1 client?

If we weren't getting rid of API v1, we might put it in [/lib/go-tc](https://github.com/apache/trafficcontrol/tree/master/lib/go-tc), since that's the canonical place to put shared constants.

Another place to put this kind of thing is [/traffic_ops/toclientlib](https://github.com/apache/trafficcontrol/tree/master/traffic_ops/toclientlib) (added in #5552), though IMO that should be somewhere in /lib, too.

traffic_ops/traffic_ops_golang/crconfig/handler.go
srijeet0406: Yeah makes sense. I was trying to follow what was being done for `writePerlHTMLErr`, where both the user error and the system error were being passed. Changed now.

traffic_ops/traffic_ops_golang/crconfig/handler.go
srijeet0406: Yeah referring the client seemed a bit weird to me as well, but as Zach mentioned, we are getting rid of API v1.

traffic_ops/traffic_ops_golang/crconfig/handler.go
srijeet0406: Normally the `api.HandleErr` function takes care of this for us.

traffic_ops/traffic_ops_golang/crconfig/handler.go
srijeet0406: Fixed now.

traffic_ops/traffic_ops_golang/crconfig/handler.go
srijeet0406: Fixed now.

TaylorCFrey (approved):
LGTM to me now. Addressed my thoughts and concerns and good to go once any other remaining reviews are resolved.

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -271,11 +270,10 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	cdn := inf.Params["cdn"]
-
 	// We never store tm_path, even though low API versions show it in responses.
 	crConfig, err := Make(inf.Tx.Tx, cdn, inf.User.UserName, r.Host, inf.Config.Version, inf.Config.CRConfigUseRequestHost, false)
 	if err != nil {
-		writePerlHTMLErr(w, r, inf.Tx.Tx, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), err)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
 		return
 	}
```
zrhoffman, line 278: If the user provides the name of a CDN that does not exist, I get a 500 error:

```
[user@host tmp]# curl -D- --cookie cookies https://${TO_HOST}.${INFRA_SUBDOMAIN}.${TLD_DOMAIN}/tools/write_crconfig/nonexistent
HTTP/2 500
access-control-allow-credentials: true
access-control-allow-headers: Origin, X-Requested-With, Content-Type, Accept, Set-Cookie, Cookie
access-control-allow-methods: POST,GET,OPTIONS,PUT,DELETE
access-control-allow-origin: *
content-type: application/json
set-cookie: mojolicious=...; Path=/; Expires=Tue, 09 Mar 2021 01:12:37 GMT; Max-Age=3600; HttpOnly
vary: Accept-Encoding
whole-content-sha512: /HT8WHLIsAIPgTPjZAa4GmpUwYtwph3fcU1//Cr7W4rkL1lKRW3kUEDTIz1DoA3/zUm3oa0elunc5VHcism6Qg==
x-server-name: traffic_ops_golang/
content-length: 62
date: Tue, 09 Mar 2021 00:12:37 GMT

{"alerts":[{"text":"Internal Server Error","level":"error"}]}
```

Errors that are the user's fault should be 400-level.

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -273,7 +273,7 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
 	// We never store tm_path, even though low API versions show it in responses.
 	crConfig, err := Make(inf.Tx.Tx, cdn, inf.User.UserName, r.Host, inf.Config.Version, inf.Config.CRConfigUseRequestHost, false)
 	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), nil)
 		return
 	}
```
zrhoffman: An error returned from `crconfig.Make()` *is* 500-level, not 400-level. The endpoint should check at an earlier point whether a CDN with the given name exists and return a 400 bad request at *that* point if it doesn't.

traffic_ops/traffic_ops_golang/crconfig/handler.go
```diff
@@ -284,7 +284,7 @@ func SnapshotOldGUIHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	if err := Snapshot(inf.Tx.Tx, crConfig, tm); err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, errors.New(r.RemoteAddr+" making CRConfig: "+err.Error()), nil)
 		return
 	}
```
zrhoffman: Same here: An error returned from `crconfig.Snapshot()` *is* 500-level, not 400-level. The endpoint should check at an earlier point whether a CDN with the given name exists and return a 400 bad request at *that* point if it doesn't.

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/crconfig/handler.go
srijeet0406: Ah sorry I misunderstood your comment.

zrhoffman (approved):
Looks good!

PR #5605, opened 2021-03-04
Added the ability to change a server capability name 

### LINKED ISSUE #5479: Add the ability to change a server capability name
## I'm submitting a ...

-  Traffic Ops
-  Traffic Portal

## Current behavior:
Currently, you cannot change the name of a server capability in TP and the TO API:

image

## New behavior:
Allow users to edit the name.

Navigate to https://tp.domain.tld/#!/server-capabilities and create a server capability and then try to change the name.

## Anything else:
Users make mistakes or change their mind. They should be able to make changes.

rimashah25:

- [ x This PR fixes #5479  

- Documentation
- Traffic Ops
- Traffic Portal

Method 1: curl PUT API end point `api/4.0/server_capabilities?name=<server_capability_name>
```
{"alerts":[{
    "text":"server capability was updated.",
    "level":"success"}],
  "response":{"RequestedName":"tar","name":"tar","lastUpdated":"2021-03-04 02:50:49+00"}
}
```
Method 2: Using TP's page for server capability to edit(update) the name and verify name was updated with write message

Method 3: Run TP's end to end test.

Method 4: Run TO's API-v4 test casse

- [x] This PR includes tests


traffic_ops/app/db/migrations/2021030300000000_server_capability_update_cascade.sql
```diff
@@ -0,0 +1,28 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+ALTER TABLE public.server_server_capability DROP CONSTRAINT fk_server_capability;
```
srijeet0406, line 15: Do you need the `public` prefix?

traffic_ops/app/db/migrations/2021030300000000_server_capability_update_cascade.sql
```diff
@@ -0,0 +1,28 @@
...
+ALTER TABLE public.deliveryservices_required_capability ADD CONSTRAINT fk_required_capability FOREIGN KEY (required_capability) REFERENCES server_capability(name) ON UPDATE CASCADE ON DELETE RESTRICT;
+
+-- +goose Down
+-- SQL section 'Down' is executed when this migration is rolled back
+ALTER TABLE public.server_server_capability DROP CONSTRAINT fk_server_capability;
+ALTER TABLE public.deliveryservices_required_capability DROP CONSTRAINT fk_required_capability;
+ALTER TABLE public.server_server_capability ADD CONSTRAINT fk_server_capability FOREIGN KEY (server_capability) REFERENCES server_capability(name) ON DELETE RESTRICT;
+ALTER TABLE public.deliveryservices_required_capability ADD CONSTRAINT fk_required_capability FOREIGN KEY (required_capability) REFERENCES server_capability(name) ON DELETE RESTRICT;
+
+
```
srijeet0406, line 27: There are 2 extra blank lines at the end.

traffic_ops/testing/api/v4/servercapabilities_test.go
```diff
@@ -89,6 +90,41 @@ func ValidationTestServerCapabilities(t *testing.T) {
 	}
 }
 
+func UpdateTestServerCapabilities(t *testing.T) {
+	var header http.Header
+
+	// Get server capability name and edit it to a new name
+	resp, _, err := TOSession.GetServerCapabilitiesWithHdr(header)
+	if err != nil {
```
srijeet0406, line 98: Could you add a check for `len(resp) != 1` here, so that we can avoid chances of a panic on line 101/ 103?

traffic_ops/testing/api/v4/servercapabilities_test.go
```diff
@@ -89,6 +90,41 @@ func ValidationTestServerCapabilities(t *testing.T) {
...
+
+	// Update server capability with new name
+	updateResponse, _, err := TOSession.UpdateServerCapabilityByName(origName, &resp[0])
+	if err != nil {
+		t.Errorf("cannot PUT server capability: %v - %v", err, updateResponse)
+	}
+
+	// Get updated name
+	getResp, _, err := TOSession.GetServerCapabilityWithHdr(newSCName, header)
+	if err != nil {
```
srijeet0406, line 113: Could you add a nil check for `getResp` here?

traffic_ops/testing/api/v4/serverservercapability_test.go
```diff
@@ -222,6 +223,43 @@ func GetTestServerServerCapabilities(t *testing.T) {
 	}
 }
 
+func UpdateTestServerServerCapabilities(t *testing.T) {
+	var header http.Header
+
+	// Get server capability name and edit it to a new name
+	resp, _, err := TOSession.GetServerCapabilitiesWithHdr(header)
+	if err != nil {
```
srijeet0406, line 231: Same comment about `len(resp)` here.

traffic_ops/testing/api/v4/serverservercapability_test.go
```diff
@@ -222,6 +223,43 @@ func GetTestServerServerCapabilities(t *testing.T) {
...
+		t.Errorf("cannot PUT server capability: %v - %v", err, updateResponse)
+	}
+
+	//To check whether the primary key change trickled down to server table
+	ssc, _, err := TOSession.GetServerServerCapabilitiesWithHdr(nil, nil, &newSCName, nil)
+	if err != nil {
+		t.Fatalf("cannot GET server capabilities assigned to servers by server capability name %v: %v", newSCName, err)
+	}
+	for i, s := range ssc {
+		if *s.ServerCapability != *ssc[i].ServerCapability {
```
srijeet0406, line 250: This will always be same. Basically, you're just comparing the element with what you get when you get the element by its index. IMO, the test should be like this:
1.) You already have a list of servers with assigned server capabilities; Keep a track of which servers are assigned to the capability in question
2.) Update the name of this server capability
3.) Get the list of servers associated with this (updated)capability again
4.) Make sure the list in 1 matches the list in 3; If not, throw error.
 
Also, missing nil checks before dereferencing pointers.

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
```diff
@@ -105,6 +119,29 @@ func (v *TOServerCapability) Read(h http.Header, useIMS bool) ([]interface{}, er
...
+		return userErr, sysErr, errCode
+	}
+	if len(sc) != 1 {
+		return fmt.Errorf("cannot find exactly one server capability with the query string provided"), nil, http.StatusBadRequest
+	}
+	rows, err := v.ReqInfo.Tx.Query(v.updateQuery(), v.RequestedName, v.Name)
+	if err != nil {
+		return nil, fmt.Errorf("server capability update: error setting the name for server capability %v: %v", v.Name, err.Error()), http.StatusInternalServerError
+	}
+	defer log.Close(rows, "unable to close DB connection")
```
srijeet0406, line 135: You need to add checks for if the capability was already modified by someone else before you updated it. Look at 
https://github.com/apache/trafficcontrol/blob/ae2ced905d223789c818831ead1c6979c488eaf0/traffic_ops/traffic_ops_golang/origin/origins.go#L303
to
https://github.com/apache/trafficcontrol/blob/ae2ced905d223789c818831ead1c6979c488eaf0/traffic_ops/traffic_ops_golang/origin/origins.go#L313

srijeet0406 (changes requested):

traffic_ops/app/db/migrations/2021030300000000_server_capability_update_cascade.sql
ocket8888: Yes, we do. The schema needs to be explicit or some people who have been running ATC since pre-1.4 will be unable to run the migration successfully.

traffic_ops/testing/api/v4/servercapabilities_test.go
rimashah25: In GetServerCapabilityWithHdr(), the check is already present and hence I didn't add the check.

traffic_ops/testing/api/v4/serverservercapability_test.go
rimashah25: Same comment as above, in GetServerCapabilityWithHdr(), the check is already present and hence I didn't add the check.

traffic_ops/testing/api/v4/servercapabilities_test.go
rimashah25: In GetServerCapabilityWithHdr(), the check is already present and hence I didn't add the check.

traffic_ops/app/db/migrations/2021030300000000_server_capability_update_cascade.sql
rimashah25: Done.

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
rimashah25: Done.

traffic_ops/testing/api/v4/servercapabilities_test.go
rimashah25: Done

traffic_ops/testing/api/v4/serverservercapability_test.go
rimashah25: Done

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
rimashah25: Done.

traffic_ops/testing/api/v4/serverservercapability_test.go
rimashah25: Updated test case to correctly check the trickle of server capability name change.

traffic_portal/app/src/common/api/ServerCapabilityService.js
```diff
@@ -68,6 +68,19 @@ var ServerCapabilityService = function($http, ENV, locationUtils, messageModel)
 		);
 	};
 
+	this.updateServerCapability = function(currentName, serverCapability) {
+		return $http.put(ENV.api['root'] + 'server_capabilities', serverCapability, {params: {"name": currentName}}).then(
+			function(result) {
+				// messageModel.setMessages(result.data.alerts, false);
```
mitchell852, line 74: u can delete that commented out line

rimashah25: ok

mitchell852:
I don't understand the `RequestedName` property of the PUT response. Is this something new we are doing?

rimashah25:
> I don't understand the `RequestedName` property of the PUT response. Is this something new we are doing?

Yes, so server_capability API uses CRUDer function. In CRUD Update, there is a function (SetKeys) that basically sets the name field (primary key) to that of the param key value. Due to which, we lose the new name (RequestedName) that the user is trying to change to. To ensure, we don't lose the new name and at the same time ensure CRUDer functionality says un-impacted, this new variable was introduced.

traffic_ops/testing/api/v4/serverservercapability_test.go
```diff
@@ -222,6 +223,61 @@ func GetTestServerServerCapabilities(t *testing.T) {
...
+	newSCName := "sc-test"
+	resp[0].Name = newSCName
+
+	// Get all servers related to original sever capability name
+	servResp, _, err := TOSession.GetServerServerCapabilitiesWithHdr(nil, nil, &origName, nil)
+	if err != nil {
+		t.Fatalf("cannot GET server capabilities assigned to servers by server capability name %v: %v", origName, err)
+	}
+	if len(servResp) == 0 {
+		t.Fatalf("no servers with associate server capability name: %s", origName)
```
srijeet0406, line 247: nit: with associate -> associated with

traffic_ops/testing/api/v4/serverservercapability_test.go
```diff
@@ -222,6 +223,61 @@ func GetTestServerServerCapabilities(t *testing.T) {
...
+		t.Fatalf("cannot GET server capabilities assigned to servers by server capability name %v: %v", newSCName, err)
+	}
+	if len(ssc) == 0 {
+		t.Fatalf("no server with associated server capability name:%s", newSCName)
+	}
+	for i, s := range ssc {
+		if newSCName != *ssc[i].ServerCapability {
+			t.Errorf("GET server server capabilities by server capability returned non-matching server capability: %s", *s.ServerCapability)
+		}
+		if *servResp[i].Server != *s.Server {
```
srijeet0406, line 268: Is it guaranteed that the sequence of the servers will be the same in both the responses? If they are not, this check will fail.

traffic_ops/testing/api/v4/serverservercapability_test.go
```diff
@@ -222,6 +223,61 @@ func GetTestServerServerCapabilities(t *testing.T) {
...
+	//To check whether the primary key change trickled down to server table
+	ssc, _, err := TOSession.GetServerServerCapabilitiesWithHdr(nil, nil, &newSCName, nil)
+	if err != nil {
+		t.Fatalf("cannot GET server capabilities assigned to servers by server capability name %v: %v", newSCName, err)
+	}
+	if len(ssc) == 0 {
+		t.Fatalf("no server with associated server capability name:%s", newSCName)
+	}
+	for i, s := range ssc {
+		if newSCName != *ssc[i].ServerCapability {
```
srijeet0406, line 265: nit: I would suggest using the same format in both the checks (this one and the one on line 268)
Basically, either use `ssc[i]` or `s` in both the checks, because they are both the same element. Makes it easier for the reader to see whats going on.

traffic_ops/testing/api/v4/serverservercapability_test.go
rimashah25: Yes because of the SortTestServerServerCapabilities() that is ran before UpdateTestServerServerCapabilities().

rimashah25:
> > Yes, so server_capability API uses CRUDer function. In CRUD Update, there is a function (SetKeys) that basically sets the name field (primary key) to that of the param key value. Due to which, we lose the new name (RequestedName) that the user is trying to change to. To ensure, we don't lose the new name and at the same time ensure CRUDer functionality says un-impacted, this new variable was introduced.
> 
> Not that I think you need to/should do so in this PR, but this is something that would motivate me to rewrite the endpoint to not use the "CRUDer"

IMO, we should get rid of all CRUDer functionality.

ocket8888:
> IMO, we should get rid of all CRUDer functionality. This PR is a good example of how we end up applying a bandaid to not impact exists functionality

I would love that. It's just hard to justify the time it would take to yank all of that stuff out.

PR #5606, opened 2021-03-04
Blueprint for phys_location change

pbivrell:

This PR contains the blueprint documenting a change to the phys location

- Documentation
- Traffic Ops
- Traffic Portal

No verification just discussion

### DIFF
--- a/blueprints/enhanced-physlocations.md
+++ b/blueprints/enhanced-physlocations.md
@@ -1,0 +1,352 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+# Enhanced physlocations 
+
+## Problem Description
+
+This blue print proposes two semi major alterations to the physical location model with ATC
+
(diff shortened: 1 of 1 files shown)


blueprints/enhanced-physlocations.md
```diff
@@ -0,0 +1,352 @@
...
+    }]
+}
+
+```
+
+For the POST request body it will be important to validate the longitude and latitude as sensible values.
+
+Some of the address lines can be optional.
+
+The type and method fields could probably be enumeration types. But I am not sure we understand/want to enforce all of the possible use cases so it may be important to limit the input text length.
```
ocket8888, line 146: I think we should address whether we'll be using an enumeration or not now, since it will affect behavior, i.e. whether or not new types/methods can be added at runtime.

Personally I like enums.

blueprints/enhanced-physlocations.md
```diff
@@ -0,0 +1,352 @@
...
+------------------+--------------------------+-----------+----------+---------
+id                | bigint                   |           | NOT NULL | 
+addressLine1      | text					 |           | NOT NULL | 
+addressLine2      | text					 |           | NOT NULL | 
+addressLine3      | text					 |           | NOT NULL | 
+addressLine4      | text					 |           | NOT NULL | 
+country           | text					 |           | NOT NULL | 
+administrativeArea| text					 |           | NOT NULL | 
+longitude         | bigfloat                 |           | NOT NULL | 
+latitude          | bigfloat                 |           | NOT NULL |
```
ocket8888, line 219: Why not let these be null? I don't think I've ever known the latitude and longitude for anywhere I've sent a letter.

pbivrell: I am on the fence about this. It is pretty easy to find latitude and longitude information for a location. And it minimizes ambiguity potential in addresses, as well as could be useful for visualizing locations. Those are all "nice to have" but maybe not something that should be enforced.

blueprints/enhanced-physlocations.md
pbivrell: Do you think we can understand enough of the use cases to make these enums?

I think for location `type` we could have an enum. However `method` for contact I think there are just to many communications methods to even attempt to enumerate them unless we want to be unhelpfully generic.

PR #5611, opened 2021-03-06
servers/details endpoint

tyagian:
- This PR is not related to any issue

This will add [servers/details](https://traffic-control-cdn.readthedocs.io/en/latest/api/v4/servers_details.html#to-api-servers-details) endpoint to Python client.

- call the added method, ensure it works


ocket8888:
What does this pull request do?

traffic_control/clients/python/trafficops/tosession.py
```diff
@@ -1645,6 +1645,16 @@ def get_server_delivery_services(self, server_id=None):
 		:raises: Union[LoginError, OperationError]
 		"""
 
+	@restapi.api_request('get', 'servers/details?hostName={name}', ('3.0','4.0',))
+    def get_server_details(self, name=None):
+        """
+		Retrieve server info in detail /api/4.0/servers/details?hostName=:name
+		:ref: `to-api-servers-details-hostName`
```
zrhoffman, line 1652: This should be a reStructuredText reference to a a label in the documentation somewhere.

Example:

https://github.com/apache/trafficcontrol/blob/038a439ebedd57ec6a4df1b2e65a390e04d6a00a/traffic_control/clients/python/trafficops/tosession.py#L270-L273

corresponds with

https://github.com/apache/trafficcontrol/blob/038a439ebedd57ec6a4df1b2e65a390e04d6a00a/docs/source/api/v4/asns.rst#L17-L21

zrhoffman (changes requested):

traffic_control/clients/python/trafficops/tosession.py
```diff
@@ -1645,6 +1645,15 @@ def get_server_delivery_services(self, server_id=None):
 		:raises: Union[LoginError, OperationError]
 		"""
 
+	@restapi.api_request('get', 'servers/details?hostName={name}', ('3.0','4.0',))
+	def get_server_details(self, name=None):
+		"""
+		Get servers/details
+		:ref: `to-api-servers-details-hostName`
```
ocket8888, line 1652: This label, `to-api-servers-details-hostName`, doesn't exist in the documentation. Closest is `to-api-servers-details`, which I think is what you want.

Also, there shouldn't be a space between the `ref` role and its argument in grave accents ("backticks").

tyagian: My mistake. I mentioned in reference to GET request: https://traffic-control-cdn.readthedocs.io/en/latest/api/v4/servers_details.html#to-api-servers-details
GET /api/4.0/servers/details?hostName but yes, endpoint is servers/details. Updated now.

ocket8888 (approved):
This works, but note that it _only_ works if you pass the `name` parameter; it can't call `servers/details`, it can only do it by passing the `hostName` query parameter. So getting details for all servers is an N+1 problem for this method.

Which is still better than nothing at all

zrhoffman (approved):

PR #5615, opened 2021-03-08
If-UnModified-Since/If-Match Improvements

rimashah25:

- [x] This PR is not related to any Issue but a good check to have when updating topology name 

- Traffic Ops

Run API tests.

- [x]  I have explained why tests are unnecessary
- [x] T I have explained why documentation is unnecessary
- [x] This PR does not include an update to CHANGELOG.md as such an update is not necessary

## Additional Information

The update logic or documentation hasn't changed.. This is an additional check that has been added to ensure no two users are updating the topology name at the same time.


traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -600,6 +600,20 @@ func (topology *TOTopology) Update(h http.Header) (error, error, int) {
...
+	var existingLastUpdated *tc.TimeNoMod
+	q := `SELECT last_updated FROM topology WHERE name = $1`
+	if err := topology.ReqInfo.Tx.QueryRow(q, topology.Name).Scan(&existingLastUpdated); err != nil {
+		if err == sql.ErrNoRows {
+			return errors.New("topology was not found"), nil, http.StatusNotFound
+		}
+		return nil, errors.New("topology update: querying: " + err.Error()), http.StatusInternalServerError
+	}
+	if !api.IsUnmodified(h, existingLastUpdated.Time) {
+		return errors.New("resource was modified"), nil, http.StatusPreconditionFailed
```
mitchell852, line 614: doesn't really sound like an error message.... sounds more like a success message :)

rimashah25: lol.. i'll update it to make it sound like an error message :)

srijeet0406 (commented):
The code and manual testing look good. could you pls add some tests around this functionality? You can refer to these lines as an exmaple:

https://github.com/apache/trafficcontrol/blob/004b52ed40f2f7d7e5bfa259e5cc222917ec869a/traffic_ops/testing/api/v4/cdns_test.go#L36

to

https://github.com/apache/trafficcontrol/blob/004b52ed40f2f7d7e5bfa259e5cc222917ec869a/traffic_ops/testing/api/v4/cdns_test.go#L43

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -600,6 +600,20 @@ func (topology *TOTopology) Update(h http.Header) (error, error, int) {
 	if len(topologies) != 1 {
 		return fmt.Errorf("cannot find exactly 1 topology with the query string provided"), nil, http.StatusBadRequest
 	}
+
+	// check if the name field was being updated by someone else
+	var existingLastUpdated *tc.TimeNoMod
```
rawlinp, line 605: This is pretty much the same code as https://github.com/apache/trafficcontrol/pull/5605/files#diff-cc0495e268314edbdfc50fba81e5ad6943ff1a3cd940fd94ef7de741862710f3R135

We have a shared function that already does this for `id`-based objects (https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/api/api.go#L1131), but we should add another similar function `CheckIfUnModifiedByName()` for `name`-based objects, which calls `GetLastUpdatedByName` instead of `GetLastUpdated`. Additionally, I'm thinking we should update the method to skip the `LastUpdated` IMS check if the request headers don't even contain `if-match`/`if-modified-since`. Otherwise, it's just a wasted query.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: Done.

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -600,6 +600,13 @@ func (topology *TOTopology) Update(h http.Header) (error, error, int) {
 	if len(topologies) != 1 {
 		return fmt.Errorf("cannot find exactly 1 topology with the query string provided"), nil, http.StatusBadRequest
 	}
+
+	// check if the name field was being updated by someone else
```
srijeet0406, line 604: This comment is misleading. This function just checks whether or not this entity (queried by its name) was modified or not. The modified field could be anything though. For ex, in case of topologies, the field could either be name or description.

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
```diff
@@ -131,17 +129,10 @@ func (v *TOServerCapability) Update(h http.Header) (error, error, int) {
...
-	q := `SELECT last_updated FROM server_capability WHERE name = $1`
-	if err := v.ReqInfo.Tx.QueryRow(q, v.Name).Scan(&existingLastUpdated); err != nil {
-		if err == sql.ErrNoRows {
-			return errors.New("server capability was not found"), nil, http.StatusNotFound
-		}
-		return nil, errors.New("server capability update: querying: " + err.Error()), http.StatusInternalServerError
-	}
-	if !api.IsUnmodified(h, existingLastUpdated.Time) {
-		return errors.New("the resource has been modified since the time specified by the request headers"), nil, http.StatusPreconditionFailed
+	// check if the name field was being updated by someone else
```
srijeet0406, line 132: As is the case with the next review comment, you can just change this comment to be `check if the entity was already updated by someone else`

traffic_ops/testing/api/v4/servercapabilities_test.go
```diff
@@ -83,6 +96,29 @@ func GetTestServerCapabilities(t *testing.T) {
...
+		t.Fatalf("Expected no error, but got %v", err.Error())
+	}
+	if len(resp) == 0 {
+		t.Fatal("no server capability in response, quitting")
+	}
+	originalName := resp[0].Name
+	newSCName := "sc-test"
+	resp[0].Name = newSCName
+
+	if (resp) != nil {
```
srijeet0406, line 111: if `len(resp) != 0`, can `resp` ever be `nil`?

traffic_ops/testing/api/v4/servercapabilities_test.go
```diff
@@ -83,6 +96,29 @@ func GetTestServerCapabilities(t *testing.T) {
...
+		_, reqInf, err := TOSession.UpdateServerCapabilityByNameWithHdr(originalName, &resp[0], header)
```
srijeet0406, line 112: Also, I noticed that in some functions, the header is passed as the second argument, and in some, it is passed in as the last argument. It would be nice to keep it consistent.

rimashah25: Header is passed as the last argument.. a function may have one, two or three (etc) variables being passed, and header is the last variable. 
UpdateTopologyWithHdr(name string, t tc.Topology, header http.Header)
UpdateServerCapabilityByNameWithHdr(name string, sc *tc.ServerCapability, header http.Header)
GetTopologyWithHdr(name string, header http.Header)
GetServerCapabilityWithHdr(name string, header http.Header)
GetServerCapabilitiesWithHdr(header http.Header)

traffic_ops/testing/api/v4/servercapabilities_test.go
rimashah25: nope. i'll remove the redundant check.

traffic_ops/traffic_ops_golang/servercapability/servercapability.go
rimashah25: Changed it to a general comment.

traffic_ops/traffic_ops_golang/topology/topologies.go
rimashah25: Same as below.

traffic_ops/v4-client/topology.go
```diff
@@ -64,11 +64,11 @@ func (to *Session) GetTopology(name string) (*tc.Topology, toclientlib.ReqInf, e
 	return to.GetTopologyWithHdr(name, nil)
 }
 
-// UpdateTopology updates a Topology by name.
-func (to *Session) UpdateTopology(name string, t tc.Topology) (*tc.TopologyResponse, toclientlib.ReqInf, error) {
+// UpdateTopologyWithHdr updates a Topology by name with header.
+func (to *Session) UpdateTopologyWithHdr(name string, t tc.Topology, header http.Header) (*tc.TopologyResponse, toclientlib.ReqInf, error) {
```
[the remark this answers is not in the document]
rimashah25: Done.

traffic_ops/v4-client/servercapability.go
```diff
@@ -69,11 +69,11 @@ func (to *Session) GetServerCapability(name string) (*tc.ServerCapability, tocli
 	return to.GetServerCapabilityWithHdr(name, nil)
 }
 
-// UpdateServerCapabilityByName updates a Server Capability by name.
-func (to *Session) UpdateServerCapabilityByName(name string, sc *tc.ServerCapability) (*tc.ServerCapability, toclientlib.ReqInf, error) {
+// UpdateServerCapabilityByNameWithHdr updates a Server Capability by name.
+func (to *Session) UpdateServerCapabilityByNameWithHdr(name string, sc *tc.ServerCapability, header http.Header) (*tc.ServerCapability, toclientlib.ReqInf, error) {
```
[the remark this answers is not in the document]
rimashah25: Based on our side conversation, I will change the function and remove `WithHdr` assuming all other function calls that have these defined in them, will be updated later in the future.

rawlinp (approved):
This looks solid

PR #5616, opened 2021-03-08
Fixed GET /servercheck filtering for unknown and extra query param

### LINKED ISSUE #5609: /servercheck does not filter when extra query parameters are given
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
GET requests to `/servercheck` can be filtered by ID and/or hostName, but if an extra query parameter is provided it breaks filtering. That is, `GET /servercheck?id=1` returns only the serverchecks for server 1, but `GET /servercheck?id=1&foo=bar` returns all serverchecks for all servers.

## Expected behavior:
The API should perform requested filtering and ignore all unrelated query parameters.

make a GET request to `/servercheck` with an `id` query string parameter set to the ID of some cache server, plus any other query parameter (except `hostName`, `limit` or `offset`) with any value.

rimashah25:

- [x] This PR fixes #5609 

- Traffic Ops

Run API tests or curl for GET /servercheck for different query (known and unknown) param (like id, hostName, foo, blah, cdn etc)

- master (038a439)
- 5.0.0

- [x] This PR includes tests

## Additional Information

No documentation change required since no new query params were added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added an endpoint for statuses on asynchronous jobs and applied it to the ACME renewal endpoint.
 
 ### Fixed
+- [#5609](https://github.com/apache/trafficcontrol/issues/5609) - Fixed GET /servercheck filter for an extra query param.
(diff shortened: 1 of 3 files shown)


ocket8888 (approved):
Works, tests cover the case and are passing.

PR #5619, opened 2021-03-09
Fix Ordering of PURGE Deny for atscfg

dsouza93:

    Adds rules to ip_allow in atscfg to allow for PURGE requests over
    localhost. Currently there is a parameter called purge_allow_ip
    that specifies specific IPs where purge requests can come from. This
    parameter DOES NOT WORK on Mids because PURGE/PUSH request denials
    are prepended to the very beginning of ip_allow.
    
    This does not fix that parameter, but opens up PURGE requests specifically
    over localhost.

- Traffic Ops ORT

Run the unit tests.
Generate the IP_allow config with ORT and ensure that on mids additional rules are prepended 
before the PUSH/PURGE deny all rules that allow for PURGE over localhost.

### DIFF
--- a/lib/go-atscfg/ipallowdotconfig.go
+++ b/lib/go-atscfg/ipallowdotconfig.go
@@ -271,8 +271,19 @@ func MakeIPAllowDotConfig(
 		sort.Sort(ipAllowDatas(ipAllowDat))
 
 		// start with a deny for PUSH and PURGE - TODO CDL: parameterize
+		// but leave purge open through localhost
 		if isMid { // Edges already deny PUSH and PURGE
 			ipAllowDat = append([]ipAllowData{
+				{
+					Src:    `127.0.0.1`,
+					Action: ActionAllow,
+					Method: `PURGE`,
+				},
+				{
+					Src:    `::1`,
+					Action: ActionAllow,
+					Method: `PURGE`,
+				},
 				{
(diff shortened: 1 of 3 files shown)


lib/go-atscfg/ipallowdotconfig.go
```diff
@@ -270,20 +270,32 @@ func MakeIPAllowDotConfig(
...
+					Method: `PUSH`,
 				},
 				{
 					Src:    `::-ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff`,
 					Action: ActionDeny,
-					Method: `PUSH|PURGE`,
+					Method: `PUSH`,
 				},
 			}, ipAllowDat...)
+			// PURGE gets placed at the end of the file
```
alficles, line 288: Unfortunately, this effectively opens PURGE up to every host that can request any data at all from the mid. This isn't likely to be what folks want. Ideally, we parameterize it, but it might work reasonably well to allow it from localhost in the simplest case so folks can use local system access controls to flexibly limit access.

alficles (changes requested):

alficles (approved):
I haven't tested this, but it's pretty straightforward and should do precisely what it says on the tin.

alficles (approved):
Still looks good.

rawlinp (approved):
LGTM

PR #5620, opened 2021-03-10
Convert the fakeOrigin tool to leverage pkg and standard ATC version

jhg03a:

This PR aligns the fakeOrigin testing tool in the repo with the norm for other ATC components both in terms of using pkg to build it and the version number following that of the main ATC project.

- fakeOrigin

`./pkg -o fakeorigin_build`
After build or install the binary version may be checked with `fakeOrigin --version`

- [x] This PR is only to adjust the build method and versioning and as such no tests are required
- [x] This PR does not include a changelog entry as it is an ancillary test tool


pkg
```diff
@@ -99,8 +99,8 @@ while getopts :?78abdf:lopqv opt; do
 			PROJECTS=`$SELF -l | sed "s/^/  - /"`
 			<<-HELP_TEXT cat
 			Usage: $SELF [options] [projects]
-			  -7           Build RPMs targeting CentOS 7 (default)
-			  -8           Build RPMs targeting CentOS 8
+			  -7           Build RPMs targeting CentOS 7
+			  -8           Build RPMs targeting CentOS 8 (default)
```
ocket8888, line 103: Is this true, builds use CentOS8 builds by default?

jhg03a: Yes

infrastructure/docker/build/docker-compose-opt.yml
```diff
@@ -26,3 +26,12 @@ services:
         RHEL_VERSION: ${RHEL_VERSION:-8}
     volumes:
       - ../../../traffic_server/tsb:/opt/tsb-ats:z
+  fakeorigin_build:
+    image: apache/fakeorigin_builder:master
```
zrhoffman, line 30: Can this be `apache/traffic_fakeorigin_builder:master` so that all of the ATC images start with `apache/traffic*_builder`?

infrastructure/docker/build/Dockerfile-fakeOrigin
```diff
@@ -0,0 +1,63 @@
...
+COPY GO_VERSION /
+RUN go_version=$(cat /GO_VERSION) && \
+	curl -Lo go.tar.gz https://dl.google.com/go/go${go_version}.linux-amd64.tar.gz && \
+	tar -C /usr/local -xvzf go.tar.gz && \
+	ln -s /usr/local/go/bin/go /usr/bin/go && \
+	rm go.tar.gz
+
+###
+
+CMD /trafficcontrol/build/clean_build.sh fakeOrigin
```
zrhoffman, line 61: Getting `fakeOrigin/build/build_rpm.sh not found`:
```shell
+ for project in "$@"
+ ./build/build.sh fakeOrigin
+ tee dist/build-fakeOrigin.log
/usr/bin/realpath
fakeOrigin/build/build_rpm.sh not found
The following subdirectories had errors:
   fakeOrigin
Error on line 1 of ./build/build.sh
```

Since `fakeOrigin` is in the `test` directory, this should be:
```dockerfile
CMD /trafficcontrol/build/clean_build.sh test/fakeOrigin
```

For an example of another component that requires this, see `grovetccfg`:
https://github.com/apache/trafficcontrol/blob/74c07997b905fbc6c5feace5f7c849d032bdbc7f/infrastructure/docker/build/Dockerfile-grovetccfg#L57

zrhoffman (changes requested):

zrhoffman (changes requested):

infrastructure/docker/build/docker-compose-opt.yml
jhg03a: That's not its name and would be confusing.

jhg03a: Also, that convention is already broken with grove.

zrhoffman: > Also, that convention is already broken with grove.

That convention is not already broken with grove.

https://github.com/apache/trafficcontrol/blob/41669a17927a1d11e8e7f5cbbf930bec2ac216d1/infrastructure/docker/build/docker-compose.yml#L100
https://github.com/apache/trafficcontrol/blob/41669a17927a1d11e8e7f5cbbf930bec2ac216d1/infrastructure/docker/build/docker-compose.yml#L110

infrastructure/docker/build/Dockerfile-fakeOrigin
jhg03a: Addressed with commit.  I'd already dealt with most of the issues, but I think I must have had a stale docker image or cache.

test/fakeOrigin/build/fakeOrigin.spec
```diff
@@ -35,16 +33,22 @@ A fake HTTP CDN Origin for testing
...
+	cd "$godir" && \
+	cp -r "$TC_DIR"/test/%{name}/* .
+) || { echo "Could not copy go program at $(pwd): $!"; exit 1; }
 
 %install
+cd src/github.com/apache/trafficcontrol/test/%{name}
 rm -rf %{buildroot}/opt/%{name}
 mkdir -p %{buildroot}/opt/%{name}/example
 cp -p %{name} %{buildroot}/opt/%{name}
+#cp -p LICENSE %{buildroot}/opt/LICENSE
```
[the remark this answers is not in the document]
jhg03a: Addressed with commit

infrastructure/docker/build/docker-compose-opt.yml
jhg03a: If a consistent naming convention for all ATC images is desired, it should be `apache/atc_.*_builder` so we also don't conflict with ATS accidentally.  Regardless, I think making that kind of a sweeping change is outside the scope of a single addition.

jhg03a: Ah, you're right.  I was looking in the wrong place.  I've updated it since everything else is consistent, but stand by that we should have used a different prefix for this.

zrhoffman (approved):

PR #5631, opened 2021-03-11
Add RHEL version to grove RPM filename

jhg03a:
This makes grove consistent with every other ATC rpm

- Grove

Build grove and grovetccfg and note that either `el8` or `el7` will be present.

- master

- [x] Grove's tests don't leverage the rpm filename
- [x] Grove is still an experimental component
- [x] Such an update is not necessary


zrhoffman (approved):
Confirmed `grove` and `grovetccfg` RPM filenames include the RHEL version now.

PR #5632, opened 2021-03-12
Allow sphinx to use multiple cpu cores

jhg03a:

- Documentation

Ensure that the documentation at least compiles still, and probably takes less time depending on hardware.

- master

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -19,7 +19,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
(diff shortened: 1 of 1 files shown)


jhg03a:
On a 64 core system, this took the running of `./pkg -v docs` from 5m26s to 43s.

ocket8888 (approved):
This seems to work. The only docs warnings are unrelated to the change. I broke a couple tables earlier in the week.

I've been meaning to do this for a while, but never got around to it; thanks.

PR #5635, opened 2021-03-12
Prevent leftover artifacts of TSB breaking builds

jhg03a:
Several projects like traffic_monitor will break presently if you build 
ATS with the optional build in pkg due to tsb leaving its source code 
behind which then gets picked up by the main go mod calls.

- ATS
- Build

1. `./pkg -v traffic_monitor_build`
2. `./pkg -v -o ats`
3. `./pkg -v traffic_monitor_build`

- master

### DIFF
--- a/traffic_server/tsb/go.mod
+++ b/traffic_server/tsb/go.mod
@@ -1,3 +1,20 @@
 module github.com/apache/trafficcontrol/traffic_server/tsb
 
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


zrhoffman (commented):
I am -1 on creating new Go modules for the sole purpose of ignoring their contents because
* We should limit ourselves to a single go.mod for the entire project.
* This is not the only directory that we would benefit from ignoring (for example, the tests in `traffic_ops_ort/testing/ort-tests` fail outside of the ORT testing environment), and having several `go.mod` files in various parts of the project would get confusing.
* For Go projects, the idiomatic way of having `go` ignore a file or directory is to [have it start with `.` or `_`](https://github.com/golang/go/blob/cdd08e615a/src/cmd/go/internal/help/helpdoc.go#L113-L114). From `go help packages`:
  > ```go
  > Directory and file names that begin with "." or "_" are ignored
  > by the go tool, as are directories named "testdata".
  > ```

PRed the directory rename in #5639, feel free to review.

zrhoffman:
@jhg03a TSB no longer breaks builds after #5639 was merged. Can #5635 be closed?

PR #5641, opened 2021-03-14
Fix atscfg header rewrite injection

rob05c:
Fixes header rewrite injection for:
- max origin connections to not [L], which can break custom rewrites
- max origin connections for new ATS 9 param
- service category to always include hook
- service category to sanitize user input
- fix header rewrite code duplication
- fix to ignore hdr_rw_ location params for Topology DSes

- Traffic Ops ORT

Run ORT unit tests. Run ORT integration tests. Run ORT manually, verify config generates correctly, particularly header rewrites.


lib/go-atscfg/headerrewritedotconfig.go
```diff
@@ -23,74 +23,300 @@ import (
...
+func InnerHeaderRewriteConfigFileName(dsName string) string {
+	return HeaderRewriteInnerPrefix + dsName + ConfigSuffix
+}
+
+func LastHeaderRewriteConfigFileName(dsName string) string {
+	return HeaderRewriteLastPrefix + dsName + ConfigSuffix
+}
+
+// MakeHeaderRewriteDotConfig makes the header rewrite file for
+// an Edge hdr_rw_ or Mid hdr_rw_mid_ or Topology hdr_rw_{first,inner,last} file,
```
alficles, line 60: Is this comment correct? It looks like it should be `{first,inner,last}_hdr_rw_`, right? The order presented here suggests a potential conflict with a ds named `inner_`, for example.

rob05c: Yes, the comment is correct: https://github.com/apache/trafficcontrol/blob/0170bba93b4e8c3e9079b1d7f930d2c878de29e1/lib/go-atscfg/topologyheaderrewritedotconfig.go#L38

It matches the existing pattern. Yes, a DS named `inner` could cause problems, but that's true today for a DS named `mid` or `hdr_rw`.

For what it's worth, it won't cause problems once DeliveryServiceServers go away, and all DSes are using Topologies.

The only real way to fix it before that, would be to refactor and rename the DSS Edge file to be `hdr_rw_edge_*`, which would incur a non-change on every single Edge server when people upgrade.

rawlinp:
Question: does this supersede https://github.com/apache/trafficcontrol/pull/5537?

rawlinp:
Also, please add this to the 5.1.1 milestone if you think it should go into the 5.1.1 release.

rob05c:
> Question: does this supersede #5537?

Yes, this included that fix. I should have actually included that changeset to give @jehunte credit, I was just being lazy.
I went ahead and merged #5537 and rebased this.

lib/go-atscfg/headerrewritedotconfig.go
```diff
@@ -23,74 +23,300 @@ import (
...
+	topologies := makeTopologyNameMap(topologiesArr)
+
+	// TODO verify prefix and suffix? Perl doesn't
+	dsName := fileName
+	dsName = strings.TrimSuffix(dsName, ConfigSuffix)
+	dsName = strings.TrimPrefix(dsName, HeaderRewriteFirstPrefix)
+	dsName = strings.TrimPrefix(dsName, HeaderRewriteInnerPrefix)
+	dsName = strings.TrimPrefix(dsName, HeaderRewriteLastPrefix)
+	dsName = strings.TrimPrefix(dsName, HeaderRewriteMidPrefix)
+	dsName = strings.TrimPrefix(dsName, HeaderRewritePrefix)
```
alficles, line 99: This code means that someone who can create or modify a DS, possibly with scope limited by tenancy can alter the config for a different DS by manipulating the name of the DS. This feels like a potential security issue in the making.

alficles (changes requested):
I think this is mostly fine, but I'm worried about the potential naming issue.

alficles (approved):
Looks like underscores in the DS name will cause all sorts of other issues and just need to be removed entirely. This issue isn't unique to this, so we can merge.

lib/go-atscfg/headerrewritedotconfig.go
rob05c: > Looks like underscores in the DS name will cause all sorts of other issues and just need to be removed entirely. This issue isn't unique to this, so we can merge.

Yep. I made https://github.com/apache/trafficcontrol/issues/5658 to add that validation to Traffic Ops

ocket8888:
This seems based on some larger bodies of work not slated for inclusion in 5.1.1. If a backport PR can be made and you wanna do that, let me know and I'll make sure an RC0 includes it, but the commit resulting from this PR cannot be cherry-picked in as-is due to conflicts.

PR #5643, opened 2021-03-14
Improves cdn notifications functionality

mitchell852:

Changes include:

1. provides the ability to create multiple notifications per CDN (instead of just one)
2. provides the ability to dismiss (or hide) a notification in TP and uses localstorage to remember the user's decision to dismiss the notification.
3. provides a drop down of all notifications for a cdn
4. provides 2 grids to help manage notifications (https://tp.domain.tld/#!/notifications and https://tp.domain.tld/#!/cdns/{{id}}/notifications)
5. Adds the ability to view/create notifications for the "ALL" cdn which is basically the same as a global notification.

- Documentation
- Traffic Ops
- Traffic Portal


traffic_portal/app/src/common/modules/header/header.tpl.html
```diff
@@ -80,9 +80,34 @@
                     <button type="button" class="btn btn-link" ng-if="hasCapability('cdns-snapshot')" ng-click="snapshot()"><i class="fa fa-camera"></i></button>
                 </div>
             </li>
-            <li ng-repeat="cdn in cdns">
-                <div class="btn-group" title="{{(cdn.notificationCreatedBy) ? 'Delete Global ' + cdn.name + ' Notification' : 'Create Global ' + cdn.name + ' Notification'}}">
-                    <button type="button" class="notification btn btn-link" ng-click="toggleNotification(cdn)">{{cdn.name}} <i class="fa" ng-class="{ 'fa-bell': cdn.notificationCreatedBy, 'fa-bell-o': !cdn.notificationCreatedBy }"></i> <i ng-show="cdn.notificationCreatedBy" class="fa fa-times"></i></button>
+            <li role="presentation" class="dropdown" ng-repeat="cdn in cdns | orderBy: '-name'">
+                <div class="btn-group" title="{{cdn.name}} Notifications" uib-dropdown is-open="{{cdn.name}}-notifications.isopen">
```
ocket8888, line 84: button should have `type=button` since it isn't part of a `form`

traffic_portal/app/src/common/modules/header/header.tpl.html
```diff
@@ -80,9 +80,34 @@
...
+                    <button id="{{cdn.name}}-notifications" type="button" class="btn btn-link" ng-click="getNotifications(cdn)" uib-dropdown-toggle>
+                        {{cdn.name}} <i class="fa" ng-class="{ 'fa-bell': cdn.hasNotifications, 'fa-bell-o': !cdn.hasNotifications }"></i>
+                    </button>
+                    <ul class="uib-dropdown-menu list-unstyled msg_list animated fadeInDown" role="menu">
+                        <li ng-show="loadingNotifications" style="text-align: center"><i class="fa fa-refresh fa-spin fa-1x fa-fw"></i></li>
```
ocket8888, line 89: I think this should use `ng-if`, since after the http request resolves the value will never change again, so there's no need to clutter the DOM with this element.

traffic_portal/app/src/common/modules/header/header.tpl.html
```diff
@@ -80,9 +80,34 @@
...
+                        <li ng-show="notifications.length === 0">
```
ocket8888, line 90: I think this check will cause it to say there are no notifications for every CDN if and only if there are no notifications for _any_ CDN. So if you have a notification for CDN A and none for CDN B, CDN B won't show this.

traffic_portal/app/src/common/modules/header/header.tpl.html
```diff
@@ -80,9 +80,34 @@
...
+                                    <span>{{::notification.user}}</span>
+                                    <span class="time">{{::getRelativeTime(notification.lastUpdated)}}</span>
+                                </span>
+                                <span class="message">{{::notification.notification}}</span>
+                            </a>
+                        </li>
+                        <li>
+                            <div class="text-center">
+                                <a>
+                                    <strong><a ng-click="navigateToPath('/cdns/' + cdn.id + '/notifications')">Manage {{cdn.name}} notifications</a></strong>
```
ocket8888, line 105: anchors should use `href` attributes, not click handlers

traffic_portal/app/src/common/modules/notifications/notifications.tpl.html
```diff
@@ -18,7 +18,8 @@
 -->
 
 <div id="notificationsContainer">
-    <div class="alert alert-info" ng-repeat="n in notifications">
+    <div ng-show="!dismissedNotifications.includes(n.id)" class="alert alert-dismissable alert-info" ng-repeat="n in notifications | orderBy: 'cdn'">
```
ocket8888, line 21: This should be using `ng-if`, since a notification can't be un-dismissed

traffic_ops/app/db/migrations/2021031400000000_cdn_notifications_multiple.sql
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+DROP TABLE IF EXISTS cdn_notification;
```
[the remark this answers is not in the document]
mitchell852: i think it's fine. we're between releases.

traffic_ops/app/db/migrations/2021031400000000_cdn_notifications_multiple.sql
```diff
@@ -0,0 +1,50 @@
...
+
+CREATE TABLE cdn_notification (
+    id BIGSERIAL PRIMARY KEY,
+    cdn text NOT NULL,
+    "user" text NOT NULL,
+    notification text,
```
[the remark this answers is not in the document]
mitchell852: good point

traffic_portal/app/src/common/modules/header/header.tpl.html
mitchell852: its seems to work perfect but i'll rethink my logic on this.

traffic_portal/app/src/common/modules/notifications/notifications.tpl.html
mitchell852: maybe i'll try ng-if but remember that notificaitons get refreshed so sometimes it needs to dynamically re-dismiss if you dismissed it earlier...

ocket8888: Do they? Then it's probably a wash in performance gains between the DOM insertion/deletion work and the creation/deletion of data binding event handlers and DOM slots.

I'd guess - without measuring - it'd save memory to use `ng-if` but save some processing time to use `ng-show`. Probably not a big deal either way, but generally I tend to prefer `ng-if` because any given DOM insertion work is typically far less disruptive on a large scale than the memory usage of child scopes.

mitchell852: right. i prefer ng-if as well. i'll use it if it works. tbh i only tried ng-show as i was thinking it was needed.

traffic_portal/app/src/common/modules/header/header.tpl.html
ocket8888: I could be wrong, it's just what it looks like to me, without testing

mitchell852: yeah, it works great. the `loadingNotifications` flag is set and reset in the getNotifications() call.

traffic_portal/app/src/common/modules/notifications/notifications.tpl.html
mitchell852: switched to ng-if

traffic_portal/app/src/common/modules/header/header.tpl.html
mitchell852: switched to ng-if

traffic_portal/app/src/common/modules/header/header.tpl.html
mitchell852: switched to ng-if

traffic_portal/app/src/common/modules/header/header.tpl.html
mitchell852: the button on line 85 has type=button already.

traffic_portal/app/src/common/modules/header/header.tpl.html
mitchell852: fixed

traffic_ops/testing/api/v4/cdnnotifications_test.go
```diff
@@ -54,9 +54,17 @@ func CreateTestCDNNotifications(t *testing.T) {
...
 	for _, cdn := range testData.CDNs {
-		_, _, err := TOSession.DeleteCDNNotification(cdn.Name)
+		// Retrieve the notifications for a cdn
+		resp, _, err := TOSession.GetCDNNotifications(cdn.Name, nil)
 		if err != nil {
-			t.Errorf("cannot DELETE CDN notification: '%s' %v", cdn.Name, err)
+			t.Errorf("cannot GET notifications for a CDN: %v - %v", cdn.Name, err)
+		}
+		if len(resp) > 0 {
+			respNotification := resp[0]
```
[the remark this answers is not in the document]
mitchell852: i think this is correct. i only created one notification per cdn and this deletes the first one for each cdn.

traffic_ops/app/db/migrations/2021031400000000_cdn_notifications_multiple.sql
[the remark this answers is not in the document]
mitchell852: fixed

traffic_portal/app/src/common/modules/header/header.tpl.html
ocket8888: My bad, missed it.

traffic_ops/app/db/migrations/2021031400000000_cdn_notifications_multiple.sql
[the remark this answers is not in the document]
ocket8888: That's technically correct, but generally I'd consider an inability to goose up/down through the migrations without destroying data bad practice.

traffic_ops/testing/api/v4/cdnnotifications_test.go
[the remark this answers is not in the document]
ocket8888: That's going to work for now, but any modifications to the notifications-related testing will need to keep this in mind for the forseeable future.

ocket8888 (approved):
This all appears to work fine. There are a few UI improvements that we discussed and @mitchell852 said he'd like to do it in a separate PR.

PR #5645, opened 2021-03-14
Add ORT dns local bind inference

rob05c:
Adds a flag to ORT to set the ATS dns local binding to the service address(s).
This is necessary when there are multiple addresses and interfaces on the host, ATS may arbitrarily choose an undesirable address to bind to, unless the config is explicit.

Includes tests.
Includes docs, ort flags are documented in the app itelf.
Includes changelog.

- Traffic Ops ORT

Run tests.
Run ORT with the new `dns_local_bind` flag, verify records.config has `dns.local_ipv4` and `dns.local_ipv6` matching the server's Interface Service Addresses.

Not a bug fix.

## Additional Information

### DIFF
--- a/traffic_ops_ort/atstccfg/config/config.go
+++ b/traffic_ops_ort/atstccfg/config/config.go
@@ -66,6 +66,7 @@ type Cfg struct {
 	TOUser          string
 	Dir             string
 	ViaRelease      bool
+	SetDNSLocalBind bool
 	ParentComments  bool
 }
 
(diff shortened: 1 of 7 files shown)


traffic_ops_ort/atstccfg/config/config.go
```diff
@@ -104,6 +105,7 @@ func GetCfg() (Cfg, error) {
 	disableProxyPtr := flag.BoolP("traffic-ops-disable-proxy", "p", false, "Whether to not use the Traffic Ops proxy specified in the GLOBAL Parameter tm.rev_proxy.url")
 	dirPtr := flag.StringP("dir", "D", "", "ATS config directory, used for config files without location parameters or with relative paths. May be blank. If blank and any required config file location parameter is missing or relative, will error.")
 	viaReleasePtr := flag.BoolP("via-string-release", "", false, "Whether to use the Release value from the RPM package as a replacement for the ATS version specified in the build that is returned in the Via and Server headers from ATS.")
+	dnsLocalBind := flag.BoolP("dns-local-bind", "", false, "Whether to use the server's Service Addresses to set the ATS DNS local bind address.")
```
serDrem, line 108: Bucking the trend to not name new new var dnsLocalBindPtr makes it look different from the others, when it's really not.

serDrem (approved):
did not know adding an option was that complicated.

traffic_ops_ort/atstccfg/config/config.go
rob05c: Eh, I originally did the `viaReleasePtr` thing. I think that was a mistake now, it adds complexity for no value. You're right, consistency is good, they should really all be changed

rob05c: Changed

PR #5646, opened 2021-03-16
List 1 path specification and license per bundled dependency

zrhoffman:

Now that golang/go#44968 is fixed, this PR breaks up larger bundled depependency pointers to instead list a single path specification and license each.

- LICENSE file

Run Weasel

- [x] LICENSE file changed only, running weasel is a sufficient test
- [x] LICENSE file changed only, documentation is unnecessary

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -332,7 +332,7 @@ Refer to the above license for the full text.
 
 This product bundles gopkg.in/tomb.v1, which is available under a BSD-3-Clause license.
 @vendor/gopkg.in/tomb.v1/*
-@vendor/gopkg.in/tomb.v1/LICENSE
+./vendor/gopkg.in/tomb.v1/LICENSE
 Refer to the above license for the full text.
 
 This product bundles yaml, which is available under an Apache-2.0 license.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -236,19 +236,35 @@ Refer to the above license for the full text.
 
 This product bundles JSON formatter, which is available under an Apache-2.0 license.
 @traffic_portal/app/src/assets/css/jsonformatter.min_0.6.0.css
```
alficles, line 238: We could put `@traffic_portal/app/src/assets/*/jsonformatter.min_0.6.0.*` and match both with one line.

LICENSE
```diff
@@ -332,12 +356,16 @@ Refer to the above license for the full text.
 
 This product bundles gopkg.in/tomb.v1, which is available under a BSD-3-Clause license.
 @vendor/gopkg.in/tomb.v1/*
-@vendor/gopkg.in/tomb.v1/LICENSE
+./vendor/gopkg.in/tomb.v1/LICENSE
 Refer to the above license for the full text.
 
 This product bundles yaml, which is available under an Apache-2.0 license.
 @vendor/gopkg.in/yaml.v2/*
 ./vendor/gopkg.in/yaml.v2/LICENSE
```
alficles, line 364: The license links are much shorter here. I think because both license files apply to the path there, it's potentially misleading to split them out. A reader might easily see one of these sections and not realize there are two license files there. I would put these back together.

LICENSE
```diff
@@ -349,6 +377,10 @@ Refer to the above license for the full text.
 This product bundles influxdb, which is available under an MIT license.
 @vendor/github.com/influxdata/influxdb/*
 ./vendor/github.com/influxdata/influxdb/LICENSE
+Refer to the above license for the full text.
+
+This product bundles influxdb, which is available under an MIT license.
+@vendor/github.com/influxdata/influxdb/*
```
alficles, line 383: Same here. Even if this winds up a bit too long, having both licenses listed where it's clear they both apply is worth it, imo.

alficles (changes requested):
This is pretty good and it'll help make it match what we offered on pkgsite. I think there are a few places we could slightly improve, though.

LICENSE
zrhoffman: Globbed in 5fb856767d

LICENSE
zrhoffman: Put them back together in 80ccbf0b2b

LICENSE
zrhoffman: Combined influxdb pointers in 80ccbf0b2b

alficles (approved):
Excellent. We should merge this.

PR #5651, opened 2021-03-17
Le fix 5.1.x

mattjackson220:

This PR fixes a bug where the LE code can cause TO to crash if no lines are deleted in the CleanUp function because err is nil

- Traffic Ops

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/letsencryptcert.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/letsencryptcert.go
@@ -125,8 +125,8 @@ func (d *DNSProviderTrafficRouter) CleanUp(domain, token, keyAuth string) error 
 			return errors.New("Determining rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
 		}
 		if rows == 0 {
-			log.Errorf("Zero rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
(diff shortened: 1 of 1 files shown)


traffic_ops/traffic_ops_golang/deliveryservice/letsencryptcert.go
```diff
@@ -125,8 +125,8 @@ func (d *DNSProviderTrafficRouter) CleanUp(domain, token, keyAuth string) error
 			return errors.New("Determining rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
 		}
 		if rows == 0 {
-			log.Errorf("Zero rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
-			return errors.New("Zero rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
+			log.Errorf("Zero rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value)
```
ocket8888, line 128: If you're using Errorf anyway, why not use formatting parameters instead of string concatenation?

ocket8888: Actually, nevermind. I just remembered this is a backport. That comment should've gone on the original, no point in whining about it here.

mattjackson220: good point! im working on updating this in master anyway for the other acme work so i can get that fixed to use formatting parameters

ocket8888 (approved):
Looks good, confirmed that the test failure is a problem with the test and not this PR

PR #5653, opened 2021-03-18
Remove CDN in a Box Makefile directory check

zrhoffman:

This PR includes various improvements to the CDN in a Box Makefile:
- Allows you to use the Makefile even if a parent directory name contains a space
- Allows you to run the Makefile without needing to be in the CDN in a Box directory

- CDN in a Box

* In `${GOPATH}/src/github.com/apache`, rename the `trafficcontrol` directory to `Traffic Control` with a space.
* In the `Traffic Control` directory, use the CDN in a Box Makefile to build the RPMs:
  ```shell
  make -f infrastructure/cdn-in-a-box/Makefile
  ```
* Test the RPMs in CDN in a Box

- master (8f6365fa6b)
- 5.1.0

- [x] Building CDN in a Box is a sufficient test, additional tests are unnecessary
- [x] The CDN in a Box documentation page does not mention not being able to use spaces, documentation is unnecessary

### DIFF
--- a/infrastructure/cdn-in-a-box/Makefile
+++ b/infrastructure/cdn-in-a-box/Makefile
@@ -50,10 +50,26 @@ TP_SOURCE := $(wildcard $(TC_DIR)/traffic_portal/**)
 TR_SOURCE := $(wildcard $(TC_DIR)/traffic_router/**)
 TS_SOURCE := $(wildcard $(TC_DIR)/traffic_stats/**)
 
+TM_RPM := $(CIAB_DIR)/traffic_monitor/traffic_monitor.rpm
+TO_RPM := $(CIAB_DIR)/traffic_ops/traffic_ops.rpm
+TP_RPM := $(CIAB_DIR)/traffic_portal/traffic_portal.rpm
+TR_RPM := $(CIAB_DIR)/traffic_router/traffic_router.rpm
+TOMCAT_RPM := $(CIAB_DIR)/traffic_router/tomcat.rpm
+TS_RPM := $(CIAB_DIR)/traffic_stats/traffic_stats.rpm
+ORT_RPM := $(CIAB_DIR)/cache/traffic_ops_ort.rpm
+
+TM_DIST_RPM := $(TC_DIR)/dist/traffic_monitor-$(SPECIAL_SAUCE)
+TO_DIST_RPM := $(TC_DIR)/dist/traffic_ops-$(SPECIAL_SAUCE)
+TP_DIST_RPM := $(TC_DIR)/dist/traffic_portal-$(SPECIAL_SAUCE)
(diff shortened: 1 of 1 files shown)


infrastructure/cdn-in-a-box/Makefile
```diff
@@ -105,43 +100,44 @@ endif
 
 # Actual output rpm recipies
 traffic_monitor/traffic_monitor.rpm: ../../dist/traffic_monitor-$(SPECIAL_SAUCE)
-	cp -f $? $@
+	cp -f "$(CIAB_DIR)/$?" "$(CIAB_DIR)/$@"
```
ocket8888, line 103: This recipe is a lie. Using this recipe does not produce the target file, unless PWD === CIAB_DIR.

You can get around that by making these recipes `.PHONY`, but that doesn't help with the problem this causes: `make` when run outside of CIAB_DIR will never conclude that the target doesn't need to be rebuilt - unless you happen to be doing it in a place where the target file actually _does_ coincidentally exist, in which case `make` will get confused and use the wrong information to arrive at its conclusions.

You could also make the targets variables and then write recipes with that like

```make
TM_RPM := $(CIAB_DIR)/traffic_monitor/traffic_monitor.rpm
TM_DIST_RPM := $(TC_DIR)/dist/traffic_monitor-$(SPECIAL_SAUCE)
$(TM_RPM): $(TM_DIST_RPM)
    cp -f $? $@

$(TM_DIST_RPM): $(TM_SOURCE)
    "$(PKG_COMMAND)" $(PKG_FLAGS) traffic_monitor$(BUILD_SUFFIX)
```

or you could still enforce the `make` command is run from the same directory as the `Makefile` but just fix the string manipulation to account for whitespace. I think that's perfectly acceptable
$(TM_RPM): $(TC

ocket8888 (changes requested):
`make` fails when run outside of the CDN-in-a-Box directory:

```
make: *** No rule to make target '../../dist/tomcat-9.0.43-11235.596a95d7.el8.x86_64.rpm', needed by 'traffic_router/tomcat.rpm'.  Stop.
```

zrhoffman:
> `make` fails when run outside of the CDN-in-a-Box directory:
> 
> ```
> make: *** No rule to make target '../../dist/tomcat-9.0.43-11235.596a95d7.el8.x86_64.rpm', needed by 'traffic_router/tomcat.rpm'.  Stop.
> ```

Fixed the Tomcat RPM target path in f7414d9c86.

infrastructure/cdn-in-a-box/Makefile
zrhoffman: Telling the truth in 2d9e90bac5

ocket8888 (approved):
This works now

PR #5657, opened 2021-03-19
Fix atscfg Service Category Header to be Internal

rob05c:
Fixes the TC Service Category header to be internal to ATS, and not sent by ATS in HTTP requests or responses.

Docs already exist for Service Category.
Tested by existing Service Category Header Rewrite tests. I also manually tested, verified the `@` header is not sent by ATS, and verified the header rewrite is created for both Edges and Mids.

- Traffic Ops ORT

Run unit tests.
Generate config, verify header rewrites are prefixed with `@` to make them internal to ATS. Run ATS, verify header writes are not passed upstream or downstream and are logged if an appropriate `svc=%<{@CDN_SVC}cqh>` log field exists.

`master`

## Additional Information

### DIFF
--- a/lib/go-atscfg/headerrewritedotconfig.go
+++ b/lib/go-atscfg/headerrewritedotconfig.go
@@ -36,7 +36,9 @@ const HeaderRewriteMidPrefix = "hdr_rw_mid_"
 const ContentTypeHeaderRewriteDotConfig = ContentTypeTextASCII
 const LineCommentHeaderRewriteDotConfig = LineCommentHash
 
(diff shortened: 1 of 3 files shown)


alficles (approved):
This looks good. It's a pretty simple change. I would note that this is a security improvement when origins are configured to respond to TRACE requests. (They generally shouldn't be.) Previously, these categories were leaked publically in that case. For most use cases, this probably isn't a huge risk, but this is an improvement in any case.

dsouza93 (approved):
I tested this as well for what its worth. It looks good to me

PR #5630, opened 2021-03-11
Update Traffic Ops database from PostgreSQL 9.6 to 13.2

rawlinp:
## What does this PR (Pull Request) do?
Change the supported Traffic Ops database version from PostgreSQL 9.6 to 13.2. Updates documentation, cdn-in-a-box, the Traffic Ops rpm, the traffic_ops_db container + tests, and the traffic_ops_ort integration tests to PostgreSQL 13.2.

Additionally, it makes one minor change to `db/admin.go` to print `psql` stderr output in case the `createdb` or `create_user` commands fail.

## Which Traffic Control components are affected by this PR?
- CDN in a Box
- Documentation
- CI tests

## What is the best way to verify this PR?
Run cdn-in-a-box and the TO API tests, ensure they still pass. Build and run the traffic_ops_db container. Build and run the traffic_ops_db docker tests. Build and run the traffic_ops_ort integration tests. 

## The following criteria are ALL met by this PR
- [x] Tests now use postgres 13.2
- [x] This PR includes documentation
- [x] This PR includes an update to CHANGELOG.md
- [x] This PR includes any and all required license headers
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #5672, opened 2021-03-23
Fix TODB tests when providing initial DB dump to test with

rawlinp:
Fixes a new issue w/ the TODB tests when providing an initial database dump file. This test feature was accidentally broken in the conversion to postgres 13.2: https://github.com/apache/trafficcontrol/pull/5630

- Traffic Ops DB tests

NOTE: the TODB postgres version that produces the TO DB dump should be 13.2. If it's 9.6, the test will fail because you can't restore a 9.6 DB dump into a 13.2 DB. If you use CIAB to get the TO dbdump, it should work, since CIAB now uses 13.2.
```
cd traffic_ops_db/test/docker
# get a DB dump from TO via /api/3.0/dbdump, put the dump file in ./initdb.d/
docker-compose build
docker-compose up --exit-code-from trafficops-db-admin
```

- master

- [x] This PR includes tests
- [x] test fix, no docs necessary
- [x] unreleased bug, no changelog necessary

### DIFF
--- a/traffic_ops_db/test/docker/initdb.d/90-load-dumps.sh
+++ b/traffic_ops_db/test/docker/initdb.d/90-load-dumps.sh
@@ -23,7 +23,7 @@ set -ex
 d=/docker-entrypoint-initdb.d
 for dump in "$d"/*dump; do
     [[ -f $dump ]] || break
-    t=$(mktemp XXX.sql)
+    t=$(mktemp -p /tmp XXX.sql)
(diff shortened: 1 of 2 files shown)


zrhoffman (approved):
Approving as db tests pass now using a 13.2 dump from CDN in a Box, but it would be nice to be able to test a PG 9.6 dump in PG 13.2.

PR #5678, opened 2021-03-25
Delivery Service Request "originals"

ocket8888:

TThis PR overhauls how DSRs work (in APIv4). Instead of a single `deliveryservice` property, DSRs will now have `requested` and `original`.

`requested` is the Delivery Service as it is requested to exist. For `delete`-type DSRs, this property will not exist in responses and will be ignored in requests. In most senses, it's pretty analogous to the old `deliveryservice` property.

`original` is the Delivery Service as it exists/existed before the DSR is/was/would have been applied. This property does not exist on `create`-type DSRs.  For non-`create` DSRs that are "open", this is either given in the request (for `delete` DSRs) or generated by Traffic Ops for responses from the current DS that matches the DSR's `requested`'s XMLID (for `update` DSRs). When a DSR is "closed", the original is saved from the current DS so that even if that DS changes in the future, the actual change enacted by the now-closed DSR is preserved.

Also, a couple more changes: Assignee ID, Author ID, and Last Edited By ID are no longer present on objects returned by the API.

For a (hopefully) complete explanation of the new model and interface, refer to the PR's docs changes.

Finally, this makes TP use the stable API 3.1 endpoints for DSR-related operations, to avoid shoving TP work into this PR.

- Documentation
- Traffic Ops
- Traffic Portal

Make sure the tests pass, make sure working with DSRs in Traffic Portal is unaffected.

- [x] This PR includes tests


mitchell852:
getting error on migration @ocket8888 

```
goose: migrating db environment 'development', current version: 2021031400000000, target: 2021032600000000
2021/03/26 13:06:34 FAIL 2021032600000000_dsrs_originals.sql (pq: check constraint "appropriate_requested_and_original_for_change_type" is violated by some row), quitting migration.
Can't run goose: exit status 1
```

traffic_portal/app/src/scripts/config.js
```diff
@@ -23,6 +23,4 @@
 
 angular.module('config', [])
 
-.constant('ENV', { api: { root:'/api/4.0/', legacy: '/api/1.5/' } })
-
-;
+.constant('ENV', { api: { root:'/api/4.0/', legacy: '/api/1.5/', stable: "/api/3.1" } });
```
mitchell852, line 26: stable needs a trailing /

mitchell852 (commented):
a couple of comments:

1. for changetype=delete, requested is not applicable so the requested property is omitted from the json response, correct? can you not omit it and just set its value to null instead. it's weird imo to have a different json structure for different types.

2. for a DSR, the xmlId is an interesting piece of information but i have to look in either the requested or original value to find it. should we just promote it to a first level property of all DSRs?

mitchell852:
> For update requests that change the XMLID, to which XMLID does the "top-level" XMLID property refer?

well, you can't ever change xmlId

ocket8888:
Oh yeah, you can't do that I forgot

mitchell852 (changes requested):
i think i found a bug: when i fulfill a DSR (change it's status to pending or complete) it looks like it's updating the "orginal" value to look like the "requested" value. original should NEVER change, right? if it does, the diff between requested and original is lost forever which is not good.

update: nm. it works fine. my error.

mitchell852 (changes requested):
This no longer works:

```
GET /deliveryservice_requests?xmlId=foo
```

mitchell852 (changes requested):
When performing `PUT api/4.0/deliveryservice_requests?id=id` on a DSR where changeType=update and simply changing a value in the `requested` field (regionalGeoBlocking from false to true) i get an internal server error and i see this in the TO logs:

```
ERROR: api.go:240: 2021-04-01T16:49:37.486221Z: 127.0.0.1:56495 pq: new row for relation "deliveryservice_request" violates check constraint "appropriate_requested_and_original_for_change_type"
```

so not sure what's going on there but it appears i can't update a DSR. this also happens when trying to update a 'delete' request. although i'm not sure why you'd ever want to do that. to summarize:

- updating an update DSR - fails
- updating a create DSR - success
- updating a delete DSR - fails

^^ the first 2 are important. not sure about the 3rd probably not but it should not return a 500.

ocket8888:
> This no longer works:

> `GET /deliveryservice_requests?xmlId=foo&status=submitted`

> It was a pretty handy way to see if a ds had "open" dsrs... and TP leverages that query param to "auto-complete" DSRs.

Which part of that no longer works? I can filter by both xmlId and status. I tested this just now by:

1. created an "update" DSR for `demo1` in "draft" status
2. requested `deliveryservice_requests?xmlId=demo1` - returns that DSR
3. requested `deliveryservice_requests?xmlId=demo2` - returns an empty array
4. requested `deliveryservice_requests?xmlId=demo1&status=submitted` - returns an empty array
5. update the status to "submitted" (which succeeds but returns a 500, for some reason I still have to look into)
6. requested `deliveryservice_requests?xmlId=demo1` - returns that DSR
7. requested `deliveryservice_requests?xmlId=demo2` - returns an empty array
8. requested `deliveryservice_requests?xmlId=demo1&status=submitted` - returns that DSR

So it's all working fine from what I can see. Plus I'm pretty confident that the API tests exercise at least the XMLID param.

ocket8888:
> updating an update DSR - fails

I believe I have fixed this, testing updating a "delete" DSR now

ocket8888:
Yeah, updating "delete" DSRs is working now, too.

mitchell852 (changes requested):
Pretty sure this is an issue. I created a 'create' DSR and then fetched it and requested.type is null but typeId=8.

image

update: my mistake. type is not populated on a 'create' dsr.  only typeId. this is correct.

mitchell852 (changes requested):
looking pretty good so far but i hit this potential bug:

i fulfilled a 'delete' request which
1. set the DSR to pending and 
2. deletes the DS

but then i tried to set the DSR to 'complete' and got this:

```
ERROR: api.go:240: 2021-04-09T16:32:32.924133Z: 127.0.0.1:56897 expected exactly one DS with XMLID 'dns-345346546', found: 0
```

ocket8888:
> looking pretty good so far but i hit this potential bug:
> 
> i fulfilled a 'delete' request which
> 
>     1. set the DSR to pending and
> 
>     2. deletes the DS
> 
> 
> but then i tried to set the DSR to 'complete' and got this:
> 
> ```
> ERROR: api.go:240: 2021-04-09T16:32:32.924133Z: 127.0.0.1:56897 expected exactly one DS with XMLID 'dns-345346546', found: 0
> ```

mitchell852 (approved):

PR #5681, opened 2021-03-26
Use the Python Postinstall implementation by default

### LINKED ISSUE #3437: TO: document postinstall input from json
`postinstall` has a feature to write all questions to a json file to answer all the questions that it would ask if invoked with no options,  and another option to use a json file as input:

```
sudo ./postinstall -defaults=/tmp/postinstall.json
sudo vim /tmp/postinstall.json
sudo ./postinstall -cfile=/tmp/postinstall.json
```

This needs to be documented in the Traffic Ops Administration documentation.

Also, the file `traffic_ops/install/bin/input.json` should be removed entirely -- it easily will get out of date and confuse users.

### LINKED ISSUE #5680: CDN in a Box has hard-coded configs instead of using Postinstall
## I'm submitting a ...

-  CDN in a Box
-  Traffic Ops (Postinstall)

## Current behavior:

CDN in a Box does not generate its configs using the Postinstall script.

## New behavior:

CDN in a Box should generate its configs using the Postinstall script.

https://github.com/apache/trafficcontrol/blob/95db4e8ef65c22f6af17aa6f63733ef80322a02f/infrastructure/cdn-in-a-box/traffic_ops/config.sh#L182-L186

https://github.com/apache/trafficcontrol/blob/95db4e8ef65c22f6af17aa6f63733ef80322a02f/infrastructure/cdn-in-a-box/traffic_ops/config.sh#L73-L77

## Anything else:

There is a Python rewrite of the Postinstall script (#4763). Instead of relying on Perl dependencies for the old Postinstall script, we should consider having CDN in a Box use the Python version of Postinstall.

zrhoffman:

- Makes Postinstall use the Python backend instead of the old Perl one
- Deprecates the Perl version of the Postinstall script
- CDN in  Box now configures Traffic Ops using the Python rewrite of the TO Postinstall script (fixes #5680).
- Adds a GitHub Actions workflow to run the Postinstall unit tests

- [x] This PR fixes #3437 by documenting JSON input to the Postinstall script. We do not need to remove `input.json` because if it ever were to become out-of-date, CDN in a Box would break.

- CDN in a Box
- Documentation
- Traffic Ops Postinstall script
- CI tests

- Run `postinstall.test.sh`
- Verify that the CDN in a Box Delivery Service can be reached

- [x] This PR includes tests


infrastructure/cdn-in-a-box/traffic_ops/config.sh
```diff
@@ -100,108 +92,103 @@ cat <<-EOF >/opt/traffic_ops/app/conf/cdn.conf
...
-    },
-    "acme_accounts": [
-        {
-            "acme_provider" : "",
-            "user_email" : "",
-            "acme_url" : "",
-            "kid" : "",
-            "hmac_encoded" : ""
-        }
-    ]
```
ocket8888, line 165: I don't think `postinstall` will add LE/ACME things.

traffic_ops/install/bin/_postinstall
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/perl
-
+#!/usr/bin/env bash
```
ocket8888, line 1: Did you change this Python code at all? I can't tell, the diff is so massive. I think if you use `git mv` to rename the files it'll make the diff cleaner, but I'm not sure since there was already a file with this name...

traffic_ops/install/bin/input.json
```diff
@@ -1,179 +1,151 @@
...
     {
-      "LDAP Search Base":"",
-      "config_var":"search_base"
+      "Traffic Ops database user": "dbuser",
+      "config_var": "user"
+    },
+    {
+      "Traffic Ops database password": "dbpass",
+      "config_var": "password",
+      "hidden": "1"
```
ocket8888, line 67: Can we add something to the deprecation? The Perl version of `_postinstall` only supports `hidden` properties as strings containing an integer encoding a boolean concept (I think?), but in Python it'll happily work with just booleans. And they really ought to be boolean. I've got other complaints about the format of these input JSON files, but that's the big one.

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
```diff
@@ -35,56 +35,72 @@ RUN set -o nounset -o errexit && \
...
 	dnf -y install "https://download.postgresql.org/pub/repos/yum/reporpms/EL-${RHEL_VERSION%%.*}-x86_64/pgdg-redhat-repo-latest.noarch.rpm"; \
-	# libicu required by postgresql13
-	dnf -y install libicu; \
-	dnf -y $use_repo -- install postgresql13; \
 	dnf -y install epel-release; \
-	dnf -y $enable_repo install      \
+	dnf -y install \
+		# libicu is required by postgresql13
+		libicu \
+		# libicu-devel, clang-devel, and llvm-devel are required by postgresql13-devel
```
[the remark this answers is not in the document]
zrhoffman: The spec requires it:

https://github.com/apache/trafficcontrol/blob/95db4e8ef65c22f6af17aa6f63733ef80322a02f/traffic_ops/build/traffic_ops.spec#L39

Maybe we can remove that dependency in a future PR.

traffic_ops/install/bin/_postinstall
zrhoffman: I did use `git mv`, 

> Did you change this Python code at all?

Not in 83d69c3778, no. Running

```shell
git mv _postinstall _postinstall.pl;
git mv postinstall.py _postinstall;
```

yields

```shell
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   _postinstall
        new file:   _postinstall.pl
        deleted:    postinstall.py
```
, so I can see not wanting to do both steps in the same commit.

Rebased to split that commit into 2 `git mv` commits in ab87a7815e and 3179f52e43 (though the overall diff will still be just as big).

infrastructure/cdn-in-a-box/traffic_ops/config.sh
```diff
@@ -100,108 +92,103 @@ cat <<-EOF >/opt/traffic_ops/app/conf/cdn.conf
...
+  "password": "$INFLUXDB_ADMIN_PASSWORD",
+  "secure": false,
+  "user": "$INFLUXDB_ADMIN_USER"
 }
-EOF
+INFLUX_CONF
+
+install_bin=/opt/traffic_ops/install/bin
+input_json="${install_bin}/input.json"
+echo "$(jq "$(<<JQ_FILTER cat
```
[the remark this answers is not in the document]
zrhoffman: Using envsubst in 9c3d3f9ecf

infrastructure/cdn-in-a-box/traffic_ops/config.sh
zrhoffman: Re-added LE/ACME things in baf2134eda

infrastructure/cdn-in-a-box/traffic_ops/config.sh
```diff
@@ -100,108 +92,103 @@ cat <<-EOF >/opt/traffic_ops/app/conf/cdn.conf
...
-        "backend_max_connections": {
-            "mojolicious": 4
-        },
-        "whitelisted_oauth_urls": [],
-        "oauth_client_secret": "",
-        "routing_blacklist": {
-            "ignore_unknown_routes": false,
-            "disabled_routes": []
-        },
-        "supported_ds_metrics": [ "kbps", "tps_total", "tps_2xx", "tps_3xx", "tps_4xx", "tps_5xx" ]
```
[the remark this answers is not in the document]
zrhoffman: The default `cdn.conf` includes all of this already.

infrastructure/cdn-in-a-box/traffic_ops/config.sh
zrhoffman: Removed it again. The default `cdn.conf` includes these exact LE/ACME settings.

infrastructure/cdn-in-a-box/traffic_ops/config.sh
```diff
@@ -100,108 +92,103 @@ cat <<-EOF >/opt/traffic_ops/app/conf/cdn.conf
...
-    },
-    "cors" : {
-        "access_control_allow_origin" : "*"
+        "db_query_timeout_seconds": ${DEBUGGING_TIMEOUT:-20}
     },
     "to" : {
-        "base_url" : "https://$TO_FQDN",
-        "email_from" : "no-reply@$INFRA_SUBDOMAIN.$TLD_DOMAIN",
-        "no_account_found_msg" : "A Traffic Ops user account is required for access. Please contact your Traffic Ops user administrator."
```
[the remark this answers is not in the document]
zrhoffman: I keep `to.email_from`, but `to.no_account_found_msg` is provided by `cdn.conf` and `to.base_url` is generated properly from the postinstall.

Here is what that part of the config looks like after `postinstall` has run in CiaB:

```json
        "to": {
                "base_url": "https://trafficops.infra.ciab.test:443",
                "email_from": "no-reply@infra.ciab.test",
                "no_account_found_msg": "A Traffic Ops user account is required for access. Please contact your Traffic Ops user administrator."
        },
```

traffic_ops/install/bin/input.json
zrhoffman: I'd rather not make breaking changes in this PR. This is supposed to be a drop-in-replacement right now, not a better-thought-out version with breaking changes.

infrastructure/cdn-in-a-box/traffic_ops/Dockerfile
```diff
@@ -99,6 +115,8 @@ COPY infrastructure/cdn-in-a-box/enroller/server_template.json \
 	infrastructure/cdn-in-a-box/traffic_ops/trafficops-init.sh \
 	infrastructure/cdn-in-a-box/variables.env \
 	/
+COPY infrastructure/cdn-in-a-box/traffic_ops_data /traffic_ops_data
+COPY traffic_router/core/src/test/resources/geo/GeoLite2-City.mmdb.gz /opt/traffic_ops/app/public/
```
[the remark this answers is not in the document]
zrhoffman: `COPY` instructions are fast and IMO should go at the end of the Dockerfile, but I don't have strong feelings about these 2 steps. Reverted those lines in 4bae66184d.

traffic_ops/install/bin/input.json
ocket8888: I wasn't suggesting a breaking change, just a deprecation.

traffic_ops/install/bin/_postinstall
ocket8888: The number of commits doesn't bother me. I just wish the diff was handled cleaner by git itself.

traffic_ops/install/bin/input.json
zrhoffman: Gotcha, deprecated in 71a84ca259.

ocket8888 (changes requested):
The script itself seems to work well enough as a drop-in replacement for Perl, but the renaming seems to have broken the tests: 
``` ocket8888@system76  ~/src/github.com/apache/trafficcontrol/traffic_ops/install/bin  ➦ 798af76a9  ./postinstall.test.sh 
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
ModuleNotFoundError: No module named '_postinstall'
```

I believe that importlib will explicitly check file extensions. So, in order to get that to work, I think we need to

* Rename it to use `.py` and just update `postinstall` itself to call with the `.py` extension. This is cleanest for the code and tests, but might be a bit more of a breaking change, as we're no longer dropping in our "drop-in" replacement
* Manipulate importlib and/or make use of `exec` in the tests. That's not ideal, feels hacky and complicated, but since it's just a test I'm not super leery of `exec` usage.
* Alter the tests to not need imports. That could potentially result in a much more lengthy, complex test script with possibly lots of duplicated code, definitely the most work of the three options - but it's cleanest for "dropping in" and doesn't do anything hacky or hard to understand.

Those are all the options I could think of

ocket8888:
As an aside, we could also add a GHA to run the tests when postinstall is updated. Doesn't need to go in this PR, but it might be helpful.

zrhoffman:
> The script itself seems to work well enough as a drop-in replacement for Perl, but the renaming seems to have broken the tests:

Fixed in 3c3c5b4e36

> As an aside, we could also add a GHA to run the tests when postinstall is updated. Doesn't need to go in this PR, but it might be helpful.

GHA added in 8b2740d54b

ocket8888 (approved):
Postinstall unit tests now passing, as proven by new action

PR #5692, opened 2021-03-30
Use `curl` to download the MaxMind database, not `wget`

zrhoffman:

- This PR avoids `AttributeError: 'module' object has no attribute 'SubprocessError'` in Python 2 if an `OSError` is thrown (e.g., if the `download_tool` path does not exit)
- This PR changes the MaxMind database download tool from wget to curl

- Traffic Ops (Python Postinstall script)

- Run the Postinstall unit tests
- Run CDN in a Box under CentOS 7 with the answer to *Download Maxmind Database?* in `input.json` changed to `"yes"`, make sure Traffic Ops starts

- master (2d566df2af)

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/install/bin/_postinstall
+++ b/traffic_ops/install/bin/_postinstall
@@ -711,10 +711,10 @@ def unmarshal_config(dct): # type: (dict) -> dict[str, list[Question]]
 
 	return ret
 
-def setup_maxmind(maxmind_answer, root, download_tool = '/usr/bin/curl'): # type: (str, str, str) -> None
+def setup_maxmind(maxmind_answer, root, download_tool = '/usr/bin/wget'): # type: (str, str, str) -> None
 	"""
 	If 'maxmind_answer' is a truthy response ('y' or 'yes' (case-insensitive), sets up a Maxmind
-	database using `curl`.
+	database using `wget`.
 	"""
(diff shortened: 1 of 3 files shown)


traffic_ops/install/bin/_postinstall
```diff
@@ -711,10 +711,10 @@ def unmarshal_config(dct): # type: (dict) -> dict[str, list[Question]]
 
 	return ret
 
-def setup_maxmind(maxmind_answer, root): # type: (str, str) -> None
+def setup_maxmind(maxmind_answer, root, download_tool = '/usr/bin/curl'): # type: (str, str, str) -> None
```
ocket8888, line 714: Changing the tool to use `curl` instead of `wget` is technically a breaking change in the otherwise "drop-in" script. The problem *should* be solved, as I understand it, by just `except`ing the proper type.

zrhoffman: The OSes that this project officially supports are CentOS 7 and CentOS 8. Neither OS includes `wget` by default and both include `curl` by default, so I consider the old tool using `wget` instead of `curl` a bug by itself.

ocket8888 (approved):
Works, no breaking changes

PR #5696, opened 2021-03-31
Change lib/go-log to escape newlines

rob05c:
Changes lib/go-log to escape newlines in the message. 

This is intended to make log parsing easier. It's currently difficult or impossible for loggers to parse messages across multiple lines. Traffic Control logs should generally not be multiple lines, but this guarantees it for log parsers.

Includes tests.
No docs, the exact log format isn't documented.

This is a library. So theoretically everything. But nothing should be logging newlines anyway.

Run unit tests.

Not a bug fix.

### DIFF
--- a/lib/go-log/log.go
+++ b/lib/go-log/log.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"log"
 	"os"
+	"strings"
 	"time"
 )
 
@@ -76,14 +77,14 @@ func logf(logger *log.Logger, format string, v ...interface{}) {
 	if logger == nil {
 		return
 	}
-	logger.Output(stackFrame, time.Now().UTC().Format(timeFormat)+": "+fmt.Sprintf(format, v...))
+	logger.Output(stackFrame, time.Now().UTC().Format(timeFormat)+": "+strings.Replace(strings.TrimSpace(fmt.Sprintf(format, v...)), "\n", `\n`, -1)+"\n")
 }
 
 func logln(logger *log.Logger, v ...interface{}) {
 	if logger == nil {
 		return
 	}
-	logger.Output(stackFrame, time.Now().UTC().Format(timeFormat)+": "+fmt.Sprintln(v...))
+	logger.Output(stackFrame, time.Now().UTC().Format(timeFormat)+": "+strings.Replace(strings.TrimSpace(fmt.Sprintln(v...)), "\n", `\n`, -1)+"\n")
 }
 
 const timeFormat = time.RFC3339Nano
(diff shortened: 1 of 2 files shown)


lib/go-log/log_test.go
```diff
@@ -42,3 +42,15 @@ func TestUTC(t *testing.T) {
 		t.Errorf("expected UTC time, actual '" + actual + "'")
 	}
 }
+
+func TestNewlinesSingleLogLine(t *testing.T) {
+	buf := &bytes.Buffer{}
+	Init(nil, writeCloser{buf}, nil, nil, nil)
+	// Have to concatenate, because Go doesn't allow you to test multiple trailing newline behavior.
+	Errorln("foo\n\nbar\n" + "\n" + "\n")
```
ocket8888, line 50: What does this mean? Surely Go doesn't just blatantly use different strings than the ones the programmer creates?

ocket8888: That's truly crazy to me. I should be able to print as many blank lines as I want. And I'd bet that the compiler actually stores a single string constant rather than doing the concatenation at runtime as an optimization, so Go's basically refusing to allow you to do something it does itself.

lib/go-log/log.go
```diff
@@ -76,14 +77,14 @@ func logf(logger *log.Logger, format string, v ...interface{}) {
...
-	logger.Output(stackFrame, time.Now().UTC().Format(timeFormat)+": "+fmt.Sprintf(format, v...))
+	logger.Output(stackFrame, time.Now().UTC().Format(timeFormat)+": "+strings.Replace(strings.TrimSpace(fmt.Sprintf(format, v...)), "\n", `\n`, -1)+"\n")
 }
 
 func logln(logger *log.Logger, v ...interface{}) {
 	if logger == nil {
 		return
 	}
-	logger.Output(stackFrame, time.Now().UTC().Format(timeFormat)+": "+fmt.Sprintln(v...))
+	logger.Output(stackFrame, time.Now().UTC().Format(timeFormat)+": "+strings.Replace(strings.TrimSpace(fmt.Sprintln(v...)), "\n", `\n`, -1)+"\n")
```
ocket8888, line 87: nit: these `+"\n"`s are unnecessary, because in a call to `log.Logger.Output`:

> A newline is appended if the last character of s is not already a newline.

Further nit: using `fmt.Sprintln` instead of `fmt.Sprintf` just gives `strings.TrimSpace` one extra character to delete.

ocket8888 (approved):
This works just fine as-is. If you don't address the nitpicks by Friday I'll assume you don't want to and just merge it.

PR #5703, opened 2021-04-01
Separate Makefile targets for relative and absolute RPM paths

zrhoffman:

Before #5653, you could build a single RPM by running `make cache/traffic_ops_ort.rpm` or `make traffic_monitor/traffic_monitor.rpm`, etc. and these Makefile targets were available to shell completion.
As of #5653, no RPM targets are completable in the shell and the only completable targets are the phony ones: `all`, `build-builders`, `clean`, `debug`, `native`, `nearly-all`, `pull-builders`, and `very-clean`.
This PR re-adds completion for those relative RPM targets but does *not* add them to the `all` target.

- CDN in a Box

In the CDN in a Box directory:
* Run `make` to verify that the `all` target still works
* Run `make traffic_ops/traffic_ops.rpm` and verify that the RPM ends up at `infrastructure/cdn-in-a-box/traffic_ops/traffic_ops.rpm`

In the main project directory:
* Run `make -f infrastructure/cdn-in-a-box/Makefile traffic_ops/traffic_ops.rpm` and verify that the RPM ends up at `infrastructure/cdn-in-a-box/traffic_ops/traffic_ops.rpm` and not `traffic_ops/traffic_ops.rpm`

 - master (c928be0441)

- [x] Building RPMs for CDN in a Box and building/running CDN in a Box itself is a sufficient test for the CiaB Makefile
- [x] An update to CHANGELOG.md is unnecessary

### DIFF
--- a/infrastructure/cdn-in-a-box/Makefile
+++ b/infrastructure/cdn-in-a-box/Makefile
@@ -27,8 +27,9 @@ else
 	RHEL_VERSION := $(shell echo $(RHEL_VERSION) | cut -f1 -d.)
 endif
 
-CIAB_DIR := $(realpath $(dir $(MAKEFILE_LIST)))
-TC_DIR := $(shell cd "$(CIAB_DIR)/../.." && pwd -P)
+CIAB_DIR_RELATIVE := $(dir $(MAKEFILE_LIST))
+CIAB_DIR_ABSOLUTE := $(shell cd $(CIAB_DIR_RELATIVE) && pwd)
+TC_DIR := $(CIAB_DIR_RELATIVE)../..
 
 PKG_COMMAND := $(TC_DIR)/pkg
 PKG_FLAGS := -v -$(RHEL_VERSION)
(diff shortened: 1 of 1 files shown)


ocket8888:
For the record, after #5653 you can still `make` individual RPMs, even if shell completions have a hard time with it

zrhoffman:
> For the record, after #5653 you can still `make` individual RPMs, even if shell completions have a hard time with it

Not with the same command, I can't. This is what I get:

```shell
[user@computer cdn-in-a-box]$ make traffic_ops/traffic_ops.rpm
make: *** No rule to make target 'traffic_ops/traffic_ops.rpm'.  Stop.
```
`make $(realpath traffic_ops/traffic_ops.rpm)` works, but that takes awhile to type because there's no shell completion.

ocket8888:
What's your `make` version? Works fine for me:
```
$ make traffic_ops/traffic_ops.rpm
make: Nothing to be done for 'traffic_ops/traffic_ops.rpm'.
$
```

zrhoffman:
> What's your `make` version? Works fine for me:
> 
> ```
> $ make traffic_ops/traffic_ops.rpm
> make: Nothing to be done for 'traffic_ops/traffic_ops.rpm'.
> $
> ```

Yes, but remove `traffic_ops/traffic_ops.rpm` and try again.

I'm using GNU Make 4.3

ocket8888:
oh, you're right. Man, that's weird...

infrastructure/cdn-in-a-box/Makefile
```diff
@@ -130,6 +130,15 @@ $(TS_RPM): $(TS_DIST_RPM)
...
 	cp -f "$?" "$@"
 
+# RPM target aliases
+traffic_monitor/traffic_monitor.rpm: $(TM_RPM)
+traffic_ops/traffic_ops.rpm: $(TO_RPM)
+traffic_portal/traffic_portal.rpm: $(TP_RPM)
+traffic_router/traffic_router.rpm: $(TR_RPM)
+traffic_router/tomcat.rpm: $(TOMCAT_RPM)
+traffic_stats/traffic_stats.rpm: $(TS_RPM)
+cache/traffic_ops_ort.rpm: $(ORT_RPM)
```
ocket8888, line 140: These are, unfortunately, lies. These recipes are only the truth if you're running `make` from within the CDN-in-a-Box directory. If you run it anywhere else, it doesn't produce the target, which is something we now support.

ocket8888: So again, we could make these `.PHONY` if we're okay with `make` not being able to tell when it needs to rebuild them, or we could again restrict it to running within the CDN-in-a-Box directory if we're okay with the that restriction. Of course, it'd still need to support running in directories with spaces in them, since a user should be able to clone wherever they want and we don't have control over that. In that case, I'd be totally fine taking out the checks at the start of the makefile that it used to have to error out if you're not calling it from the right place, if those are too much of a pain to do correctly. I don't think people did that, in general, and I don't think `cd` as a dependency is asking too much.

zrhoffman: Gave the Makefile separate targets for relative paths and absolute paths in fc200a5946. Now it tab-completes both the relative-path RPM filename and the full-path RPM filename.

If you build a relative-path RPM target, Make treats the corresponding full-path RPM target as built, too.

zrhoffman:
Force-pushed to reword the commit message

ocket8888 (approved):
This works fine, and if there are any negative consequences to having multiple (identical) recipes for the same file, I can't think what they are.

PR #5705, opened 2021-04-01
Fix TO client urlsig path

rob05c:
Fixes the TO client urlsig path being wrong.

This is critical, because ORT uses this client function now, causing ORT url_sig keys to break.

- Traffic Ops Client
- Traffic Ops ORT

Run tests. Run ORT, and verify url_sig files are generated correctly. Write an app to use the TO client urlsig function, verify it returns the keys successfully.

ORT is critically broken in `master and 5.1. Client v2-client and v3-client functions have been broken since inception (but nothing inside ATS used the function, as far as I know).

## Additional Information

### DIFF
--- a/traffic_ops/testing/api/v2/deliveryservices_test.go
+++ b/traffic_ops/testing/api/v2/deliveryservices_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


rob05c:
Don't merge yet, this is effectively a draft, but I want to see the checks run, and drafts don't run them apparently.

traffic_ops/testing/api/v2/deliveryservices_test.go
```diff
@@ -32,6 +32,7 @@ import (
 func TestDeliveryServices(t *testing.T) {
 	WithObjs(t, []TCObj{CDNs, Types, Tenants, Users, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, DeliveryServices}, func() {
 		GetAccessibleToTest(t)
+		GetTestDeliveryServicesURLSigKeys(t)
```
rawlinp, line 35: Since this new test depends on Riak, it should only be run `if includeSystemTests`.

rawlinp (changes requested):

traffic_ops/testing/api/v2/deliveryservices_test.go
rob05c: Done

rawlinp (approved):

PR #5711, opened 2021-04-02
Traffic Portal integration tests strict compilation mode

ocket8888:

This PR enables certain strict compilation options for the Traffic Portal integration test. Specifically, it sets:
```json
{
    "strict": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitThis": true
}
```

It does *not*, however, enable the `noImplicityAny` compiler option, because the test data is not currently type-safe, and changing that would be a multi-thousand-line change in its own right.

- Traffic Portal

Ensure the TP integration tests compile, run, and pass.

- [x] This PR includes tests

### DIFF
--- a/traffic_portal/test/integration/CommonUtils/API.ts
+++ b/traffic_portal/test/integration/CommonUtils/API.ts
@@ -21,6 +21,7 @@
 import { Agent } from "https";
 
 import axios from 'axios';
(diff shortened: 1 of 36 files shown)


shamrickus (approved):
LGTM - really like the changes to no longer using try/catch

PR #5713, opened 2021-04-05
Fix ORT adding remap lines for DSes not on Mids

rob05c:
Bug was ORT Mid remap gen not using the new lib/go-tc DSType
UsesMidCache func, but having old code checking live and natnl,
which ended up with HTTP_NO_CACHE generating remap lines on Mids.

This could particularly cause issues because the hdr_rw_mid_ file
generation did use ds.Type.UsesMidCache, potentially resulting
in a remap line with a file that didn't exist, causing ATS to fail
to load.

This could be mitigated by creating a location Param for the file,
to force it to be created.

Note this did not affect Topologies, which use their own criteria
for both header rewrite file and remap line generation.

Includes tests.
No changelog, no interface change, and bug is not in a release.
No docs, no interface change.

- Traffic Ops ORT

Run tests. Create a non-topology DS of type HTTP_NO_CACHE with a Mid Header Rewrite, run ORT, verify no remap line exists for the DS.

- master

## Additional Information

### DIFF
--- a/lib/go-atscfg/remapdotconfig.go
+++ b/lib/go-atscfg/remapdotconfig.go
@@ -138,7 +138,8 @@ func getServerConfigRemapDotConfigForMid(
 				continue
 			}
 		}
-		if ds.Type.IsLive() && !ds.Type.IsNational() && !hasTopology {
+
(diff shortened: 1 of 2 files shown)


dsouza93 (approved):
I built this and tested this. It does indeed fix the bug.

PR #5720, opened 2021-04-07
Convert Traffic Portal integration tests data to be type-safe

ocket8888:

This PR, fundamentally, adds the `noImplicitAny` Typescript compiler option, and makes the code compliant with that. What that entails is mostly rewriting all of the JSON test data as Typescript. With the exception of a few case changes and collapsing directories into files per tested object, there should be no structural changes to the testing data.

There were also some untyped function parameters (which default to an implicit `any`) that necessitated adding interface types to page objects, and occasionally some of these required logic changes to handle strict `null` and `undefined` checking.

Also, three of the dependencies imported were untyped and do not have `@types` type definitions, so `.d.ts` files declaring their typing were added. Specifically:

* `random-ipv6` - This library is relatively small, and the typing provided is expected to be complete.
* `jasmine-reporters` - This library is kinda big, but the typing provided is expected to be complete.
* `protractor-beautiful-reporter` - This library is massive, and the typing provided is only enough to cover current usage. If currently unused options are added the typing definition *will* need to be updated.

Finally, this PR replaces usages of the `jasmine-data-provider` module with a simple loop over array contents, which not only provides type safety (as `jasmine-data-provider` is untyped) but eliminates the need for that dependency, which has therefore been removed from the package files.

- Traffic Portal - tests

Ensure the integration tests compile and pass, and that no test cases have become unused or have been removed.

- [x] This PR includes tests


dpham692:
Test pass. This PR also fix the webdriver problem. More abstract for the config file.  Save time by just run the npm test so user doesn't have to type down tsc command down and just run the test.
Everything work fine for me.

dpham692 (approved):
LGTM

PR #5723, opened 2021-04-08
GHA DB tests now pass if you modify the migration after creation in the same PR

### LINKED ISSUE #5690: GHA DB tests fail if you modify the migration after creation in the same PR
## I'm submitting a ...
-  bug report

-  CI tests

## Current behavior:
If a migration is edited in the same PR that creates it, the CI action for the database migrations will report that it is out of order, even when it isn't. 

## Expected behavior:
The tests should account for the possibility that a migration may be edited before being merged, and test accordingly.

My PR #5678 displays this behavior. This is because - I think - the test script assumes that `git --no-pager log --format=%ct 2021032600000000_dsrs_originals.sql` is going to output one line, the time of creation of the migration. It then adds this to an array, and later compares it against the latest modification time of any migration using `git log -1 --name-status --format="%ct" . | head -n 1` with `!=`. So essentially it's comparing
```bash
mtime_array+=( "$(git --no-pager log --format=%ct 2021032600000000_dsrs_originals.sql)" )
mtime_length=${#mtime_array[@]}
echo ${mtime_array[$mtime_length-1]}
# Outputs: 1617049981 1616777950
```
to
```bash
git log -1 --name-status --format="%ct" . | head -n 1
# Outputs: 1617049981
```
Which you can see is lining up with _part_ of the `mtime_array` entry, but will never be exactly equal because the entry has two numbers.

rimashah25:

- [x] This PR fixes #5690 

- CI tests

1. Create a draft PR and add a db migration file.
2. Update the newly added migration file.
If the file added/modified is the latest file, then GH actions checks will pass. 
If it is out of order (wrt name), then it will fails the GH actions.

- master (4e8e3ba)
- 5.1.1

## Additional Information

Since logic hasn't changed, tests and documentation weren't added.

### DIFF
--- a/.github/actions/todb-tests/entrypoint.sh
+++ b/.github/actions/todb-tests/entrypoint.sh
@@ -59,7 +59,7 @@ LATEST_FILE_TIME="$(git log -1 --name-status --format="%ct" . | head -n 1 )"
 mtime_array=()
 arr=($(ls))
 for file in "${arr[@]}"; do
-  mtime_array+=( "$(git log --format=%ct  $file)" )
+  mtime_array+=( "$(git log -1 --format=%ct  $file)" )
 done
 mtime_length=${#mtime_array[@]}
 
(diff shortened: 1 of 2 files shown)


ocket8888 (approved):
Tested by rebasing my PR #5678 on these changes and running the actions. The TO RPM failed to build so the action actually failed, in case anyone's checking my work on this for whatever reason. But that happened after the migration names tests passed.

PR #5725, opened 2021-04-08
Set server XMPPID to hostname if empty

### LINKED ISSUE #5724: PUT /api/3.x/servers with no xmppId (hashId) results in an internal server error
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
PUT /api/3.x/servers with no xmppId (hashId) results in an internal server error

```
ERROR: api.go:209: 2021-04-08T18:39:19.577714586Z: [ip] original server had no XMPPID
```

## Expected behavior:
No ISE. You can't change xmppId after server create so i'm not even sure why the api is checking for one.

Create a server, in the db, delete the xmppId.

PUT /api/3.x/servers/your-new-server-id and you should get a 500 Internal Server Error.

## Anything else:
Workaround: add an xmppId to the db for that server and the error goes away

Also, i see lots of instances of xmppId defined like this in the documentation which is not accurate:

```
xmppId
An identifier to be used in XMPP communications with the server - in nearly all cases this will be the same as hostName
```

zrhoffman:

- [x] This PR fixes #5724 by allowing a Server to be updated when its XMPPID field is empty.

This PR also adds a migration to set the XMPPID to the hostname for each server in the database for which the `xmpp_id` field is empty.

- Traffic Ops

- Run the TO API tests
- Running an SQL query directly against the database, set the XMPPID for a server to NULL or an empty string, then try updating it using Traffic Portal

- master (4e8e3bab98)

- [x] This PR includes tests
- [x] This PR contains documentation


traffic_ops/app/db/migrations/2021040819481469_generate_uuid_for_xmpp_id_where_empty.sql
```diff
@@ -0,0 +1,23 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+-- +goose Up
+-- SQL in section 'Up' is executed when this migration is applied
+CREATE EXTENSION pgcrypto; /* Loads pgcrypto if gen_random_uuid() does not already exist */
+UPDATE "server" s set xmpp_id = gen_random_uuid() WHERE s.xmpp_id IS NULL OR s.xmpp_id = '';
```
rawlinp, line 19: We can't update existing xmpp_ids to UUIDs because they're currently getting defaulted to their hostnames. So the migration would have to update null xmpp_ids to the server's host_name; otherwise, TR will consistent-hash the server differently, potentially cold-caching all of them.

rawlinp (changes requested):

traffic_ops/app/db/migrations/2021040819481469_generate_uuid_for_xmpp_id_where_empty.sql
zrhoffman: Setting XMPPID to hostname for the migration in 0cb43fea0c

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1638,7 +1640,11 @@ func Update(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if server.XMPPID != nil && *server.XMPPID != originalXMPPID {
+	if server.XMPPID == nil || *server.XMPPID == "" {
+		log.Infof("Found empty XMPPID for server %s, setting random XMPPID", *server.HostName)
+		server.XMPPID = newUUID()
```
rawlinp, line 1645: We can't really do this either -- we'd have to set it to the hostname instead. Or rather, we shouldn't set it at all in an update, but just assume that nil/empty means hostname.

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/server/servers.go
zrhoffman: Not setting XMPPID in an update in cb1b9976cc.

mitchell852:
there was some incorrect documentation regarding XMPPID. want to fix that too? lots of this:

```
xmppId
An identifier to be used in XMPP communications with the server - in nearly all cases this will be the same as hostName
```

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1459,10 +1460,8 @@ func Update(w http.ResponseWriter, r *http.Request) {
 	}
 
 	original := originals[0]
-	if original.XMPPID == nil {
-		sysErr = errors.New("original server had no XMPPID")
-		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, sysErr)
-		return
+	if original.XMPPID == nil || *original.XMPPID == "" {
+		log.Warnln("original server had no XMPPID")
```
rawlinp, line 1464: we should include more context, including the server hostname, in this log message

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -443,6 +443,7 @@ UPDATE server SET
 	tcp_port=:tcp_port,
 	type=:server_type_id,
 	upd_pending=:upd_pending,
+	xmpp_id=:xmpp_id,
```
rawlinp, line 446: we should remove this line to ensure that it's impossible to update (assuming any future bugs in validation logic)

rawlinp (changes requested):

rawlinp:
> A poorly named property of a server used by Traffic Router for consistent hashing. On create, this is set to a UUID, however, legacy behaviour would use server hostname if this value was undefined.

This is pretty good, but I would add that it's an _immutable_ property.

rimashah25:
> there is some incorrect documentation regarding XMPPID. want to fix that too? lots of this:
> 
> ```
> xmppId
> An identifier to be used in XMPP communications with the server - in nearly all cases this will be the same as hostName
> ```

Isn't that part of v1 docs.. soon to go away? I checked the codebase otherwise and couldn't find the above mention in any other documentation.

traffic_ops/traffic_ops_golang/server/servers.go
rimashah25: I agree, that was the reason I had removed it in #4881

mitchell852:
> > there is some incorrect documentation regarding XMPPID. want to fix that too? lots of this:
> > ```
> > xmppId
> > An identifier to be used in XMPP communications with the server - in nearly all cases this will be the same as hostName
> > ```
> 
> Isn't that part of v1 docs.. soon to go away? I checked the codebase and couldn't find the above mention in any other documentation.

you are right @rimashah25 and it looks like it's better defined elsewhere:

```
xmppId
A system-generated UUID used to generate a server hashId for use in Traffic Router’s consistent hashing algorithm. This value is set when a server is created and cannot be changed afterwards.
```

traffic_ops/traffic_ops_golang/server/servers.go
zrhoffman: Removed `xmpp_id` from the update query in 68c51e362d.

traffic_ops/traffic_ops_golang/server/servers.go
zrhoffman: Including the hostname in the warning in 034ad01440.

zrhoffman:
> > > there is some incorrect documentation regarding XMPPID. want to fix that too? lots of this:
> > > ```
> > > xmppId
> > > An identifier to be used in XMPP communications with the server - in nearly all cases this will be the same as hostName
> > > ```
> > 
> > 
> > Isn't that part of v1 docs.. soon to go away? I checked the codebase and couldn't find the above mention in any other documentation.
> 
> you are right @rimashah25 and it looks like it's better defined elsewhere:
> 
> ```
> xmppId
> A system-generated UUID used to generate a server hashId for use in Traffic Router’s consistent hashing algorithm. This value is set when a server is created and cannot be changed afterwards.
> ```

Using the newer XMPPID definition for API v1 in a20cdf0def.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - [#5405](https://github.com/apache/trafficcontrol/issues/5405) - Prevent Tenant update from choosing child as new parent
 - [#5384](https://github.com/apache/trafficcontrol/issues/5384) - New grids will now properly remember the current page number.
 - [#5548](https://github.com/apache/trafficcontrol/issues/5548) - Don't return a `403 Forbidden` when the user tries to get servers of a non-existent DS using `GET /servers?dsId={{nonexistent DS ID}}`
+- [#5724](https://github.com/apache/trafficcontrol/issues/5724) - Generate XMPPID on update if the server had none
```
rawlinp, line 50: Since this PR no longer generates UUIDs for XMPPID, we should probably say `Set XMPPID to hostname if the server had none, don't error on server update when XMPPID is empty`.

Also, would the title would make more sense now as "Set server XMPPID to hostname if empty"?

rawlinp (approved):
This looks good now, just nitpicking the changelog and title :)

CHANGELOG.md
zrhoffman: Updated PR title and updated changelog entry in 04f137b767.

rawlinp:
Also if this PR is considered for backporting to a release branch, the backport should _not_ include the DB migration. The code fix alone should be sufficient for fixing the bug.

PR #5726, opened 2021-04-09
Add a Federation to CDN in a Box

### LINKED ISSUE #3697: CIAB: only works with one deliveryservice
adding another deliveryservice by creating a new file in `traffic_ops_data/deliveryservices/020-dns1.json:

```
{
    "xmlId": "dns1",
    "displayName": "DNS1",
    "tenant": "root",
    "protocol": 2,
    "orgServerFqdn": "http://origin.infra.ciab.test",
    "cdnName": "$CDN_NAME",
...
```

I get the following message in the `trafficops` logs:
```
return
trafficops_1       | ++ curl -k -LfsS --cookie /tmp/tmp.yffSZJpgz2 -X GET 'https://trafficops.infra.ciab.test:443/api/1.4/deliveryservices/xmlId/demo1
trafficops_1       | dns1/sslkeys?decode=true'
trafficops_1       | curl: (3) Illegal characters found in URL
```

This should be handled appropriately....,

zrhoffman:

- [x] This PR fixes #3697 by adding a `xmlId` to the query parameter used by `run-go.sh` when verifying the `demo1` Delivery Service is created to prevent more than 1 DS in the response 
- Adds a `demo2` DNS Delivery Service to CDN in a Box
- Adds a Federation check to the `readiness` CDN in a Box test

- CDN in a Box
- Documentation
- Traffic Control Go Client Library

Wait for CDN in a Box to start, then confirm that Traffic Router returns a CNAME record for the federation:
```shell
docker-compose exec trafficrouter dig +short @trafficrouter.infra.ciab.test -t CNAME video.demo2.mycdn.ciab.test
```

Expected response:
```shell
foo.kabletown.net.
```

 - master (ae5866ebcf)
 - 5.1.1
 - 5.1.2 (RC2)

- [x] This PR includes tests


TaylorCFrey (approved):
Can confirm verification steps pass. Code looks good too. Launched CiaB and verified both delivery services (one HTTP demo1 and the other DNS demo2) through proxied TP as well.

LGTM. Approved.

docs/source/admin/quick_howto/ciab.rst
```diff
@@ -102,10 +102,19 @@ In a typical scenario, if the steps in `Building`_ have been followed, all that'
...
 
-	sudo docker-compose exec enroller /usr/bin/curl -L "http://video.demo1.mycdn.ciab.test"
+	sudo docker-compose exec enroller curl -L "http://video.demo1.mycdn.ciab.test"
+
+To test the ``foo.kabletown.net.`` Federation:
+
+	[user@computer cdn-in-a-box]$ sudo docker-compose exec trafficrouter dig +short @trafficrouter.infra.ciab.test -t CNAME video.demo2.mycdn.ciab.test
+
+	# Expected response:
+	foo.kabletown.net.
```
ocket8888, line 117: Code blocks should be typeset using a `code-block` directive

infrastructure/cdn-in-a-box/enroller/enroller.go
```diff
@@ -713,6 +713,125 @@ func enrollServerCapability(toSession *session, r io.Reader) error {
...
+// and IPv6 resolvers to that Federation.
+func enrollFederation(toSession *session, r io.Reader) error {
+	dec := json.NewDecoder(r)
+	var federation tc.AllDeliveryServiceFederationsMapping
+	err := dec.Decode(&federation)
+	if err != nil {
+		log.Infof("error decoding Server Capability: %s\n", err)
+		return err
+	}
+	mapping := federation.Mappings[0]
```
ocket8888, line 727: This will segfault and crash the enroller if the federation has empty mappings

infrastructure/cdn-in-a-box/enroller/enroller.go
```diff
@@ -713,6 +713,125 @@ func enrollServerCapability(toSession *session, r io.Reader) error {
...
+			CName: mapping.CName,
+			TTL:   mapping.TTL,
+		}
+		resp, _, err := toSession.CreateCDNFederation(cdnFederation, cdnName)
+		if err != nil {
+			log.Infof("creating CDN Federation: %s", err.Error())
+			return err
+		}
+		cdnFederation = resp.Response
+		if _, err = toSession.CreateFederationDeliveryServices(*resp.Response.ID, []int{*deliveryService.ID}, true); err != nil {
```
ocket8888, line 753: there are a lot of places where dereferencing pointer values like this could be causing segfaults to crash the enroller. On one hand, other possible segfaults (like if `deliveryServices` has zero length above) that _are_ handled, so it seems like not crashing if TO (or the client) is broken is the intended behavior. But on the other hand it's difficult to say how the enroller could do anything useful if TO (or the client) is broken, which is the only way these could wind up null.

You can these problems or not, but I thought I'd just point it out.

docs/source/admin/quick_howto/ciab.rst
zrhoffman: Oops, `code-block` directive added in ed4e624bd2.

lib/go-tc/federation_resolver.go
```diff
@@ -25,6 +25,11 @@ import "net"
 
 import "github.com/go-ozzo/ozzo-validation"
 
+type FederationResolverResponse struct {
```
ocket8888, line 28: GoDoc?

Also, to avoid breaking changes to old client call signatures, we've started explicitly versioning library response structures with a rolling alias for the latest minor version. So in this case it'd look like

```go
type FederationResolverResponse struct {
    Alerts
    Response FederationResolver `json:"response"`
}

type FederationResolverResponseV4 = FederationResolverResponseV40
```
and then the client uses the V4 version

infrastructure/cdn-in-a-box/readiness/run.sh
```diff
@@ -30,20 +30,20 @@ while ! to-ping; do
...
 done
 
+xmlID="$(<<<"$DS_HOSTS" sed 's/ .*//g')" # Only get the first xmlID
 while true; do
     sleep 3
-    exampleURLs=($(to-get /api/${TO_API_VERSION}/deliveryservices | jq -r '.response[].exampleURLs[]' || true))
-    if ! exampeURLs=($(to-get "/api/${TO_API_VERSION}/deliveryservices" | jq -r '.response[].exampleURLs[]')) \
-            || [[ "${#exampleURLs[@]}" -eq 0 ]]; then
-        echo waiting for delivery service example URLs
+    if ! exampleURLs=($(to-get "/api/${TO_API_VERSION}/deliveryservices?xmlId=${xmlID}" | jq -r '.response[].exampleURLs[]')) ||
```
ocket8888, line 36: I think the reason you split this up into first and second xmlids was because it wasn't properly checking all example URLs for all Delivery Services, right? I think the intention was to support an arbitrary number of Delivery Services (>0) - can we just loop over the expected

infrastructure/cdn-in-a-box/enroller/enroller.go
ocket8888: Generally as I'm going through the client methods in #5747 if any of them entail enroller changes I'm adding the necessary guards for API calls I have to modify, if that affects your decision at all. Which is to say, if you don't do it now I'll probably do it in a rebase at some point.

infrastructure/cdn-in-a-box/enroller/enroller.go
zrhoffman: Enrolling as many Federation mappings as are given in 286336f3f1.

lib/go-tc/federation_resolver.go
zrhoffman: Versioning the `FederationResolverResponse` struct in 2532a74f0a and added godocs.

infrastructure/cdn-in-a-box/enroller/enroller.go
zrhoffman: I have have added a nil check where `enrollFederation()` dereferences the `UserName` field in e1b2105aeb.

As for the other dereferences, IMO, a nil check on on a Traffic Ops response struct field whose corresponding database column is part of that table's primary key or has a `NOT NULL` constraint is unnecessary.

zrhoffman: That being said, all the more power to you if you choose to add checks for them in #5747.

infrastructure/cdn-in-a-box/readiness/run.sh
zrhoffman: > I think the reason you split this up into first and second xmlids was because it wasn't properly checking all example URLs for all Delivery Services, right?

I split them up because the checks for the first and second DSes are different.

For the `demo1` DS, the check is to GET the example URLs, follow redirects, and check status codes.

For the `demo2` DS, the check is to verify the Federation CNAME is returned when querying the Delivery Service hostname.

> can we just loop over the xmlids to support as many as exist?

If we do that, we'd need to remove `demo3` from the `DS_HOSTS` environment variable so we can expect all DSes listed to exist:

https://github.com/apache/trafficcontrol/blob/473d9c242247618f17ac64a2b4e53e7e14fdc7ed/infrastructure/cdn-in-a-box/variables.env#L21

I suppose we could loop over the Delivery Service hosts given in DS_HOSTS and
- GET example URLs if it's an HTTP-routed Delivery Service
- Find the Federation using the `FEDERATION_CNAME` environment variable, get its Delivery Service's `xmlID` from that, and for that `xmlID`, expect the value `FEDERATION_CNAME` as the response
- `dig` the Delivery Service hostname if it's a DNS-routed Delivery Service that is not used by the Federation specified by `FEDERATION_CNAME`

zrhoffman: That approach would also involve exiting with an error if at least 1 Delivery Services of each type we expect to check is not found.

ocket8888: Yes that sounds very convoluted - never mind

infrastructure/cdn-in-a-box/enroller/enroller.go
ocket8888: It's certainly not as important as if this were a test - where we specifically shouldn't trust the software being tested - but for things like this it's really more important to just check user input, to be sure. But:

> As for the other dereferences, IMO, a nil check on on a Traffic Ops response struct field whose corresponding database column is part of that table's primary key or has a `NOT NULL` constraint is unnecessary.

In *my* opinion, a "nullable" field on a struct property that is `NOT NULL` or a primary key is unnecessary. That would solve this problem too.

Of course, that's not the place of this PR to change. Just food for thought.

As far as me changing them, it just offends my delicate sensibilities to have programs crash with memory access violations instead of printing an error and exiting non-zero normally. It's not really a security concern or anything, so I probably don't actually need to do anything, but I also probably still will.

ocket8888 (approved):

Issue #5558, opened 2021-02-22
TM UI and /api/cache-statuses reports aggregate bandwidth_kbps as 0 unless interfaces are monitored


rawlinp:
## I'm submitting a ...
-  bug report

## Traffic Control components affected ...
-  Traffic Monitor

## Current behavior:
In the TM UI, bandwidth is reported as `0` unless the server has interfaces with `monitor = true`.
![Screen Shot 2021-02-22 at 2 08 30 PM](https://user-images.githubusercontent.com/6724053/108770910-2546e480-7518-11eb-997d-8d459686340d.png)

The `/api/cache-statuses` API also has the same issue:
```
  "odol-atsmid-oshob-01": {
    "type": "MID",
    "load_average": 1.55,
    "query_time_ms": 1939,
    "health_time_ms": 1,
    "stat_time_ms": 1,
    "stat_span_ms": 5819,
    "health_span_ms": 1940,
    "status": "REPORTED - available",
    "status_poller": "health",
    "bandwidth_kbps": 0,
    "bandwidth_capacity_kbps": 1000000,
    "connection_count": 16,
    "ipv4_available": true,
    "ipv6_available": true,
    "combined_available": true,
    "interfaces": {
      "ens3": {
        "status": "available",
        "status_poller": "health",
        "bandwidth_kbps": 266,
        "available": true
      }
    }
  },
```
As you can see, this host is doing 266 kbps, but it's not showing in the UI or in the `bandwidth_kbps` field (the aggregate of all interfaces) in this API.

## Expected behavior:
The UI and /api/cache-statuses API should report the aggregate bandwidth of a server's interfaces, even if `monitor = false`.

## Minimal reproduction of the problem with instructions:
Open the TM UI and observe what I've described for servers with no monitored interfaces. Then make one or more of those interfaces monitored and observe that the bandwidth starts reporting correctly.

srijeet0406:
@rawlinp I can take this one

PR #5730, opened 2021-04-09
Fixes an issue where multiple interfaces reported by a cache are included in the vitals (bandwidth)

### LINKED ISSUE #5695: Incorrect bandwidth calculations from dual homed servers
## I'm submitting a ...
-  bug report

-  Traffic Monitor

## Current behavior:
Traffic Monitor indiscrimantly sums up bandwidth calculations from all reported interfaces, whether the interface is marked as "Monitor this interface" or not. When more than 1 interface is returned for a cache server it is assumed to be part of the calculations. This results in inaccurate numbers as some interfaces are "nested" / children interfaces or loopback which result in 2x calculations at a minimum.

## Expected behavior:
Only the interfaces marked with "Monitor this interface" should be included in the calculations.

Query a Traffic Server cache where the `/_stats` endpoint returns multiple interfaces, including those that may not necessarily be noted in the TP/TO. These additional interfaces will be included in the bandwidth calculation that can then be seen in the Traffic Monitor endpoint `/CacheStats` or `/CacheStatsNew`. The bandwidth should be based only on the interfaces marked as "Monitor this interface" as can be seen in the Traffic Portal web application.

TaylorCFrey:

- [X] This PR fixes #5695

- Traffic Monitor

Unit tests have been added to the Traffic Monitor project to ensure this logic will work.

`go test apache/trafficcontrol/traffic_monitor/health` must pass

Additionally, verify Traffic Monitor still continues to report Bandwidth as reported in:
https://github.com/apache/trafficcontrol/issues/5558

- master (f8c82a7)
- 5.1.+


CHANGELOG.md
```diff
@@ -49,8 +49,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - [#5405](https://github.com/apache/trafficcontrol/issues/5405) - Prevent Tenant update from choosing child as new parent
 - [#5384](https://github.com/apache/trafficcontrol/issues/5384) - New grids will now properly remember the current page number.
 - [#5548](https://github.com/apache/trafficcontrol/issues/5548) - Don't return a `403 Forbidden` when the user tries to get servers of a non-existent DS using `GET /servers?dsId={{nonexistent DS ID}}`
+- [#5695](https://github.com/apache/trafficcontrol/issues/5695) - Ensure vitals are calculated only against monitored interfaces
 - [#5724](https://github.com/apache/trafficcontrol/issues/5724) - Set XMPPID to hostname if the server had none, don't error on server update when XMPPID is empty
 
+
```
srijeet0406, line 55: nit: Extra blank line

traffic_monitor/health/cache.go
```diff
@@ -53,14 +53,44 @@ func GetVitals(newResult *cache.Result, prevResult *cache.Result, mc *tc.Traffic
 		return
 	}
 
+	if mc == nil {
+		log.Errorf("mc TrafficMonitorConfigMap cannot be nil")
```
srijeet0406, line 57: The word `mc` might not make much sense in the logs.

traffic_monitor/health/cache.go
```diff
@@ -53,14 +53,44 @@ func GetVitals(newResult *cache.Result, prevResult *cache.Result, mc *tc.Traffic
...
+
 	if newResult.InterfaceVitals == nil {
 		newResult.InterfaceVitals = map[string]cache.Vitals{}
 	}
 
 	// proc.loadavg -- we're using the 1 minute average (!?)
 	newResult.Vitals.LoadAvg = newResult.Statistics.Loadavg.One
 
-	for ifaceName, iface := range newResult.Interfaces() {
+	if mc.TrafficServer[newResult.ID].Interfaces == nil {
```
srijeet0406, line 68: Could you add some checks here to see that the length of `mc.TrafficServer` is not 0, and also to make sure that `mc.TrafficServer` has the key `newResult.ID`?

traffic_monitor/health/cache.go
```diff
@@ -53,14 +53,44 @@ func GetVitals(newResult *cache.Result, prevResult *cache.Result, mc *tc.Traffic
...
+	}
+
+	if len(monitoredInterfaces) == 0 {
+		log.Debugf("no interfaces selected to be monitored for %v", newResult.ID)
+		return
+	}
+
+	for _, monitoredInterface := range monitoredInterfaces {
+		ifaceName := monitoredInterface.Name
+		iface, exists := newResult.Interfaces()[ifaceName]
```
srijeet0406, line 87: Maybe add some similar checks here like my last comment^^

traffic_monitor/health/cache.go
```diff
@@ -53,14 +53,44 @@ func GetVitals(newResult *cache.Result, prevResult *cache.Result, mc *tc.Traffic
...
+		if !exists {
+			// monitored interface doesn't exist in Result interfaces, skip
+			log.Warnf("monitored interface %v does not exist in %v cache.Result.Interfaces()", ifaceName, newResult.ID)
```
srijeet0406, line 90: Do you think this log line would read better as:
`log.Warnf("monitored interface %v does not exist in cache %v", ifaceName, newResult.ID)`?

traffic_monitor/health/cache_test.go
```diff
@@ -33,9 +33,364 @@ import (
...
+	}
+
+	// No interfaces were selected to be monitored so no vitals
+	// should have been calculated
+	if firstResult.Vitals != zeroValueVitals {
+		t.Errorf("Vitals should have zero values. expected: %v actual: %v:", zeroValueVitals, firstResult.Vitals)
+	}
+
+	// multiple interfaces, plus extras
+	secondResult := cache.Result{
```
srijeet0406, line 119: Instead of rewriting the whole struct here, can we just take `firstResult`, copy it into `secondResult` and modify just the required fields?

traffic_monitor/health/cache_test.go
```diff
@@ -33,9 +33,364 @@ import (
...
+		MaxKbpsOut: 200000000,
+	}
+	// Only two interfaces were selected to be monitored so vitals
+	// should have been calculated based on those two (bond0 and bond1)
+	if firstResult.Vitals != expectedFirstVitals {
+		t.Errorf("Vitals do not match expected output. expected: %v actual: %v:", expectedFirstVitals, firstResult.Vitals)
+	}
+
+	// multiple interfaces, plus extras
+	secondResult := cache.Result{
```
srijeet0406, line 283: Same comment here.

traffic_monitor/health/cache.go
```diff
@@ -53,14 +53,44 @@ func GetVitals(newResult *cache.Result, prevResult *cache.Result, mc *tc.Traffic
...
 	// proc.loadavg -- we're using the 1 minute average (!?)
 	newResult.Vitals.LoadAvg = newResult.Statistics.Loadavg.One
 
-	for ifaceName, iface := range newResult.Interfaces() {
+	if mc.TrafficServer[newResult.ID].Interfaces == nil {
+		log.Debugf("no interfaces reported in config map to be monitored")
+		return
+	}
+
+	var monitoredInterfaces []tc.ServerInterfaceInfo
```
srijeet0406, line 73: Just a general question with respect to the `lo` interface. Could we ever have a case where the `lo` interface is marked as `Monitor this interface`? And in that case, should we be including the `lo` interface stats?

TaylorCFrey: Excellent question. Ideally with DH servers, this would be the single interface being collected, rather than marking both (or more) of the other interfaces with "Monitor this interface". As such, we don't want to put any restrictions on this value for two reasons:
1) We can't guarantee the naming of the interfaces. Thus any restrictions would have to be done just on conventions around names (such as `lo`)
2) Should we want to monitor the `lo` interface in the future, any coded restrictions now would require a codebase change and a deployment, rather than a toggle of a checkbox in Traffic Portal. This does mean there is a potential for User-Error by checking or unchecking the various interfaces in TP to be monitored, but there isn't an elegant approach beyond that for now. I will look into adding a tooltip to the "Monitor this interface" checkbox label in TP to see if that is an appropriate fix to help prevent accidental User-Error

traffic_monitor/health/cache_test.go
TaylorCFrey: Will do.

traffic_monitor/health/cache_test.go
TaylorCFrey: Will do.

traffic_monitor/health/cache.go
TaylorCFrey: Appreciate the clarity and simplicity. Will do.

traffic_monitor/health/cache.go
TaylorCFrey: Since this is a map, the `exists` value will tell us if a match was found and we can log the information thusly. Otherwise the map returns an empty interface, rather than `nil`. This should be satisfactory as written.

traffic_monitor/health/cache.go
TaylorCFrey: This won't necessarily panic or cause any issues if the length is 0 or the cache isn't found. However, this is actually silently failing for those reasons. I will put checks in so that we can at least log the information to uncover the issues in the future should the cache map be empty or if the cache we're attempting to monitor is missing from said map.

traffic_monitor/health/cache.go
TaylorCFrey: Will remove.

CHANGELOG.md
TaylorCFrey: Will remove.

TaylorCFrey: Done

traffic_monitor/health/cache.go
TaylorCFrey: Done

traffic_monitor/health/cache.go
TaylorCFrey: Done

traffic_monitor/health/cache.go
TaylorCFrey: Done

traffic_monitor/health/cache_test.go
TaylorCFrey: Done

traffic_monitor/health/cache_test.go
TaylorCFrey: Done

srijeet0406 (approved):
Code looks great! All api/ manual tests pass. 
Manually tested by linking up my local TM to the staging TO and made sure that the bandwidth shows the value calculated by only those interfaces that are monitored. LGTM!

PR #5733, opened 2021-04-10
Updates to the ORT plugin verifier

jrushford:

Updates the ORT plugin verifier.

- Adds a check to ensure that plugin config files are not empty
- Adds info logging of successful plugin and plugin config file
  verification.
- Changes to the pattern matching for plugin config files in
  plugin.config.
- Adds searching for plugin YAML files


rob05c (approved):
Looks good! Code looks good, tests pass, manually testing worked as expected.

PR #5738, opened 2021-04-12
Acme async

mattjackson220:

This PR updates the ACME generation endpoints to include async status updates.

- Documentation
- Traffic Ops
- Traffic Portal

Run ACME cert generation with any cert provider and verify that it returns an async status endpoint and that the updates to that endpoint are accurate.

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/acme.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/acme.go
@@ -29,6 +29,7 @@ import (
 	"database/sql"
 	"encoding/pem"
 	"errors"
+	"fmt"
 	"net/http"
 	"strconv"
 	"strings"
@@ -104,12 +105,12 @@ func (d *DNSProviderTrafficRouter) Present(domain, token, keyAuth string) error 
 	tx.Commit()
 	if err != nil {
 		log.Errorf("Inserting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
-		return errors.New("Inserting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
+		return fmt.Errorf("Inserting dns txt record for fqdn '"+fqdn+"' record '"+value+"': %v", err)
 	} else {
 		rows, err := response.RowsAffected()
 		if err != nil {
 			log.Errorf("Determining rows affected dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
-			return errors.New("Determining rows affected dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
+			return fmt.Errorf("Determining rows affected dns txt record for fqdn '"+fqdn+"' record '"+value+"': %v", err)
 		}
 		if rows == 0 {
 			log.Errorf("Zero rows affected when inserting dns txt record for fqdn '" + fqdn + "' record '" + value)
@@ -130,12 +131,12 @@ func (d *DNSProviderTrafficRouter) CleanUp(domain, token, keyAuth string) error 
 	tx.Commit()
 	if err != nil {
 		log.Errorf("Deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
-		return errors.New("Deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
+		return fmt.Errorf("Deleting dns txt record for fqdn '"+fqdn+"' record '"+value+"': %v", err)
 	} else {
 		rows, err := response.RowsAffected()
 		if err != nil {
 			log.Errorf("Determining rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
-			return errors.New("Determining rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value + "': " + err.Error())
+			return fmt.Errorf("Determining rows affected when deleting dns txt record for fqdn '"+fqdn+"' record '"+value+"': %v", err)
 		}
 		if rows == 0 {
 			log.Errorf("Zero rows affected when deleting dns txt record for fqdn '" + fqdn + "' record '" + value)
@@ -162,7 +163,7 @@ func GenerateAcmeCertificates(w http.ResponseWriter, r *http.Request) {
 
 	req := tc.DeliveryServiceAcmeSSLKeysReq{}
 	if err := api.Parse(r.Body, nil, &req); err != nil {
-		api.HandleErr(w, r, nil, http.StatusBadRequest, errors.New("parsing request: "+err.Error()), nil)
+		api.HandleErr(w, r, nil, http.StatusBadRequest, fmt.Errorf("parsing request: %v", err), nil)
 		return
 	}
 	if *req.DeliveryService == "" {
@@ -171,7 +172,7 @@ func GenerateAcmeCertificates(w http.ResponseWriter, r *http.Request) {
 
 	dsID, cdnName, ok, err := dbhelpers.GetDSIDAndCDNFromName(inf.Tx.Tx, *req.DeliveryService)
 	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("deliveryservice.GenerateLetsEncryptCertificates: getting DS ID from name "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, fmt.Errorf("deliveryservice.GenerateLetsEncryptCertificates: getting DS ID from name: %v", err))
 		return
 	} else if !ok {
 		api.HandleErr(w, r, inf.Tx.Tx, http.StatusNotFound, errors.New("no DS with name "+*req.DeliveryService), nil)
@@ -186,7 +187,7 @@ func GenerateAcmeCertificates(w http.ResponseWriter, r *http.Request) {
 
 	_, ok, err = dbhelpers.GetCDNIDFromName(inf.Tx.Tx, tc.CDNName(*req.CDN))
 	if err != nil {
-		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("checking CDN existence: "+err.Error()))
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, fmt.Errorf("checking CDN existence: %v", err))
 		return
 	} else if !ok {
 		api.HandleErr(w, r, inf.Tx.Tx, http.StatusNotFound, errors.New("cdn not found with name "+*req.CDN), nil)
@@ -238,7 +239,7 @@ func GenerateLetsEncryptCertificates(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := api.Parse(r.Body, nil, &req); err != nil {
-		api.HandleErr(w, r, nil, http.StatusBadRequest, errors.New("parsing request: "+err.Error()), nil)
+		api.HandleErr(w, r, nil, http.StatusBadRequest, fmt.Errorf("parsing request: %v", err), nil)
 		return
 	}
 	if *req.DeliveryService == "" {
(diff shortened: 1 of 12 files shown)


traffic_portal/app/src/common/api/DeliveryServiceSslKeysService.js
```diff
@@ -47,7 +43,11 @@ var DeliveryServiceSslKeysService = function($http, locationUtils, messageModel,
 
         return $http.post(ENV.api['root'] + endpoint, generateSslKeyForm).then(
             function(result) {
-            	messageModel.setMessages([{level: 'success', text: message + deliveryService.xmlId}], true);
+            	if (message === null) {
+            		messageModel.setMessages(result.data.alerts, true);
+				} else {
+					messageModel.setMessages([{level: 'success', text: message + deliveryService.xmlId}], true);
+				}
```
ocket8888, line 50: Indentation here mixes spaces and tabs - unfortunately the file currently appears to use spaces, so we should stick with that inferior choice.

ocket8888 (commented):
Code looks fine, need to test

ocket8888 (changes requested):
Looks like the handler can segfault, crashing Traffic Ops:
```
trafficops_1      | panic: runtime error: invalid memory address or nil pointer dereference
trafficops_1      | [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa520ea]
trafficops_1      | 
trafficops_1      | goroutine 33601 [running]:
trafficops_1      | github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice.GetAcmeCertificates(0xc000359400, 0x0, 0xc001c3c330, 0xc001c3c340, 0x0, 0x0, 0x0, 0xc001c3c350, 0x0, 0x0, ...)
trafficops_1      |     /tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice/acme.go:331 +0xbea
trafficops_1      | created by github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice.GenerateLetsEncryptCertificates
trafficops_1      |     /tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice/acme.go:277 +0x645
```
I'm not sure if that's introduced by this PR, but it should be impossible to crash Traffic Ops with a GET request, so we need to fix that immediately

ocket8888:
Yeah, looks probably not caused by this PR. Problem is that `authType` isn't required, but is a "nullable" field that is dereferenced:
```go
provider := *req.AuthType
```

ocket8888:
You might consider wrapping the goroutine with a `recover` call that updates the async job status to mark it as failed and bails safely.

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -219,7 +249,11 @@ func GenerateLetsEncryptCertificates(w http.ResponseWriter, r *http.Request) {
 
 	ctx, _ := context.WithTimeout(r.Context(), AcmeTimeout)
 
-	req := tc.DeliveryServiceLetsEncryptSSLKeysReq{}
+	req := tc.DeliveryServiceAcmeSSLKeysReq{}
+	if req.AuthType == nil {
+		*req.AuthType = tc.LetsEncryptAuthType
```
ocket8888, line 254: This line will segfault every single time it's run. If `req.AuthType` is `nil`, then dereferencing it like this can only result in a panic. If you want to default to the "Let's Encrypt" auth type, you can do
```go
req.AuthType = new(string)
*req.AuthType = tc.LetsEncryptAuthType
```
(because you can't take the address of a constant, and probably don't want that to be mutable)

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -257,43 +291,76 @@ func GenerateLetsEncryptCertificates(w http.ResponseWriter, r *http.Request) {
...
+			}
 
-	alerts.AddAlerts(api.CreateDeprecationAlerts(util.StrPtr(API_ACME_GENERATE_LE)))
+			if asyncErr := api.UpdateAsyncStatus(db, api.AsyncFailed, "ACME renewal failed.", asyncStatusId, true); asyncErr != nil {
+				log.Errorf("updating async status for id %v: %v", asyncStatusId, asyncErr)
+			}
+			api.HandleErr(w, r, nil, http.StatusInternalServerError, nil, fmt.Errorf("panic: (err: %v) stacktrace:\n%s\n", err, util.Stacktrace()))
+			return
+		}
+	}()
```
ocket8888, line 314: To test this recovery, I added a `panic` at the top of `GetAcmeCertificates` on line 331. Traffic Ops still crashed, and I never saw any of these log messages. So it appears that you must handle the panic from within the same goroutine that started it.

ocket8888 (changes requested):

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -306,17 +355,26 @@ func GetAcmeCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEncryptSS
 	if err != nil {
 		log.Errorf("deliveryservice.GenerateSSLKeys: getting DS ID from name " + err.Error())
 		api.CreateChangeLogRawTx(api.ApiChange, "DS: "+*req.DeliveryService+", ID: "+strconv.Itoa(dsID)+", ACTION: FAILED to add SSL keys with "+provider, currentUser, logTx)
+		if err = api.UpdateAsyncStatus(db, api.AsyncFailed, "ACME renewal failed.", asyncStatusId, true); err != nil {
+			log.Errorf("updating async status for id %v: %v", asyncStatusId, err)
+		}
 		return errors.New("deliveryservice.GenerateSSLKeys: getting DS ID from name " + err.Error())
```
ocket8888, line 361: This situation is one of the reasons I tend to prefer `fmt.Errorf` to `errors.New`.

If getting the Delivery Service ID from its XMLID fails with some error, this will log the error, enter a changelog entry, update the async job status, and then segfault - _unless_ updating the async job status also fails. Because you're re-assigning `err` based on the output of `api.UpdateAsyncStatus`, if that succeeds `err` will be `nil` and `err.Error` is an invalid dereference

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -347,13 +408,19 @@ func GetAcmeCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEncryptSS
 	if err != nil {
 		log.Errorf("acme: getting acme client for provider %s", provider)
 		api.CreateChangeLogRawTx(api.ApiChange, "DS: "+*req.DeliveryService+", ID: "+strconv.Itoa(dsID)+", ACTION: FAILED to add SSL keys with "+provider, currentUser, logTx)
+		if err = api.UpdateAsyncStatus(db, api.AsyncFailed, "ACME renewal failed.", asyncStatusId, true); err != nil {
+			log.Errorf("updating async status for id %v: %v", asyncStatusId, err)
+		}
 		return errors.New("getting acme client: " + err.Error())
```
ocket8888, line 414: same problem here

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -398,23 +471,34 @@ func GetAcmeCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEncryptSS
 	if err := tv.PutDeliveryServiceSSLKeys(dsSSLKeys, tx); err != nil {
 		log.Errorf("Error putting ACME certificate in Traffic Vault: %s", err.Error())
 		api.CreateChangeLogRawTx(api.ApiChange, "DS: "+*req.DeliveryService+", ID: "+strconv.Itoa(dsID)+", ACTION: FAILED to add SSL keys with "+provider, currentUser, logTx)
+		if err = api.UpdateAsyncStatus(db, api.AsyncFailed, "ACME renewal failed.", asyncStatusId, true); err != nil {
+			log.Errorf("updating async status for id %v: %v", asyncStatusId, err)
+		}
 		return errors.New(deliveryService + ": putting keys in Traffic Vault: " + err.Error())
```
ocket8888, line 477: same problem here

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -398,23 +471,34 @@ func GetAcmeCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEncryptSS
...
 	}
 
 	tx2, err := db.Begin()
 	if err != nil {
 		log.Errorf("starting sql transaction for delivery service " + *req.DeliveryService + ": " + err.Error())
+		if err = api.UpdateAsyncStatus(db, api.AsyncFailed, "ACME renewal failed.", asyncStatusId, true); err != nil {
+			log.Errorf("updating async status for id %v: %v", asyncStatusId, err)
+		}
 		return errors.New("starting sql transaction for delivery service " + *req.DeliveryService + ": " + err.Error())
```
ocket8888, line 486: same problem here

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -398,23 +471,34 @@ func GetAcmeCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEncryptSS
...
 	}
 
 	if err := updateSSLKeyVersion(*req.DeliveryService, req.Version.ToInt64(), tx2); err != nil {
 		log.Errorf("updating SSL key version for delivery service '" + *req.DeliveryService + "': " + err.Error())
+		if err = api.UpdateAsyncStatus(db, api.AsyncFailed, "ACME renewal failed.", asyncStatusId, true); err != nil {
+			log.Errorf("updating async status for id %v: %v", asyncStatusId, err)
+		}
 		return errors.New("updating SSL key version for delivery service '" + *req.DeliveryService + "': " + err.Error())
```
ocket8888, line 494: same problem here

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -398,23 +471,34 @@ func GetAcmeCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEncryptSS
...
 	}
 	tx2.Commit()
 
 	api.CreateChangeLogRawTx(api.ApiChange, "DS: "+*req.DeliveryService+", ID: "+strconv.Itoa(dsID)+", ACTION: Added SSL keys with "+provider, currentUser, logTx)
-
+	if err = api.UpdateAsyncStatus(db, api.AsyncSucceeded, "ACME renewal complete.", asyncStatusId, true); err != nil {
+		log.Errorf("updating async status for id %v: %v", asyncStatusId, err)
+	}
 	return nil
```
ocket8888, line 502: Should you be returning this error if it's not `nil`?

mattjackson220: my thinking on these was that we should just log any errors to the async updates so that it is just a status and doesnt impact the actual workflow. I'm very open to changing it though if we think it would be better to have it return an error if the async part fails

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -338,6 +397,9 @@ func GetAcmeCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEncryptSS
 		if acmeAccount == nil {
 			log.Errorf("acme: no account information found for %s" + provider)
```
ocket8888, line 398: This prints, e.g.

```
acme: no account information found for %!s(MISSING)LetsEncrypt
```
when provider is `"LetsEncrypt"`. Should be <kbd>,</kbd> not <kbd>+</kbd>.

I know that's not this PR's fault, it's just something I noticed

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
```diff
@@ -347,13 +409,19 @@ func GetAcmeCertificates(cfg *config.Config, req tc.DeliveryServiceLetsEncryptSS
 	if err != nil {
 		log.Errorf("acme: getting acme client for provider %s", provider)
```
ocket8888, line 410: Because this is being called as a goroutine, the returned error is discarded. Because of this, and because this print statement does not include the error message, if this error occurs operators will have no idea what caused it or what to do to rectify the issue.

At a glance this _looks_ like the only place where this kind of error swallowing occurs, but I'd appreciate if you'd look it over and make sure I didn't miss anything.

Alternatively, if you wanted, you could cut down on the log statements by making `GetAcmeCertificates` a wrapper with a panic handler that gets an error from something unexported like e.g. `getACMECertificates` that just _returns_ an error, and then the wrapper could log it.

But that's a pretty big structural change and afaict if you just append the error message here it won't be an actual problem, so you don't have to.

traffic_ops/traffic_ops_golang/deliveryservice/acme.go
ocket8888: no, that makes sense to me

ocket8888 (approved):

PR #5746, opened 2021-04-14
Prevent looping when a user's login attempt fails

### LINKED ISSUE #5739: Login failure loop in v5.1.0 go v3-client
## I'm submitting a ...

-  bug report

-  Traffic Control Client
-  Traffic Ops

## Current behavior:
When building go client applications with the TC v5.1.0 v3-client library, we have observed that a login failure (such as resulting from passing incorrect login credentials) results in a looping condition, where a steady stream of login attempts are made against TO API.

## Expected behavior:
When invalid login credentials are used, the login attempt should fail immediately, without a apparent retry loop.

Build a client app using `github.com/apache/trafficcontrol v5.1.0+incompatible`.
Pass invalid login credentials to that app.

## Anything else:
Confirmed that building with `github.com/apache/trafficcontrol v5.0.0+incompatible` does not reproduce the problem.

srijeet0406:

- [x] This PR fixes #5739  

- Traffic Ops

Manual testing:
Write a small app/ test that uses the client's `LoginWithagent` method to login to the TO system. 
Make sure you use an invalid username/ password combination.
Tail the access logs in TO, and make sure that the login endpoint is not hit repeatedly after the first login fails.

- master

- [x] Just changing one of the methods to not loop anymore, tests not reqd.


shamrickus (approved):
LGTM
* No longer creates infinite loops trying to login

PR #5748, opened 2021-04-16
TO client api tests for Regions automation

dmohan001c:

- CDN in a Box
- Traffic Ops
- CI tests


traffic_ops/testing/api/v4/regions_test.go
```diff
@@ -142,6 +159,36 @@ func SortTestRegions(t *testing.T) {
 	}
 }
 
+func SortTestRegionsDesc(t *testing.T) {
+
+	var header http.Header
```
ocket8888, line 164: just so you know, `nil` is a valid `http.Header` value, so you can just pass `nil` if you don't need the header

traffic_ops/testing/api/v4/regions_test.go
```diff
@@ -142,6 +159,36 @@ func SortTestRegions(t *testing.T) {
...
+	if len(respAsc) > 0 && len(respDesc) > 0 {
+		// reverse the descending-sorted response and compare it to the ascending-sorted one
+		for start, end := 0, len(respDesc)-1; start < end; start, end = start+1, end-1 {
+			respDesc[start], respDesc[end] = respDesc[end], respDesc[start]
+		}
+		if respDesc[0].Name != "" && respAsc[0].Name != "" {
+			if !reflect.DeepEqual(respDesc[0].Name, respAsc[0].Name) {
+				t.Errorf("Regions responses are not equal after reversal: %s - %s", *&respDesc[0].Name, *&respAsc[0].Name)
+			}
+		}
```
ocket8888, line 186: This test should `Fatal` before this point if the lengths of `respAsc` and `respDesc` are not exactly equal, otherwise code in here will segfault

traffic_ops/testing/api/v4/regions_test.go
```diff
@@ -142,6 +159,36 @@ func SortTestRegions(t *testing.T) {
...
+	}
+
+	if len(respAsc) > 0 && len(respDesc) > 0 {
+		// reverse the descending-sorted response and compare it to the ascending-sorted one
+		for start, end := 0, len(respDesc)-1; start < end; start, end = start+1, end-1 {
+			respDesc[start], respDesc[end] = respDesc[end], respDesc[start]
+		}
+		if respDesc[0].Name != "" && respAsc[0].Name != "" {
+			if !reflect.DeepEqual(respDesc[0].Name, respAsc[0].Name) {
+				t.Errorf("Regions responses are not equal after reversal: %s - %s", *&respDesc[0].Name, *&respAsc[0].Name)
```
ocket8888, line 184: nit but `*&` is a no-op; `*&respDesc.Name` is the same as just `respDesc.Name`

traffic_ops/testing/api/v4/regions_test.go
```diff
@@ -175,15 +222,69 @@ func UpdateTestRegions(t *testing.T) {
...
-func GetTestRegions(t *testing.T) {
-	for _, region := range testData.Regions {
-		resp, _, err := TOSession.GetRegionByName(region.Name, nil)
-		if err != nil {
-			t.Errorf("cannot GET Region by region: %v - %v", err, resp)
-		}
+func VerifyPaginationSupportRegion(t *testing.T) {
+	qparams := url.Values{}
+	qparams.Set("orderby", "id")
+	regions, _, err := TOSession.GetRegions(qparams, nil)
```
ocket8888, line 230: This test contains many lines that will segfault if the length of `regions` is not at least 3 - which it should check for

traffic_ops/testing/api/v4/regions_test.go
```diff
@@ -243,3 +343,39 @@ func DeleteTestRegions(t *testing.T) {
...
+	}
+}
+
+func GetRegionByInvalidName(t *testing.T) {
+	regionResp, _, err := TOSession.GetRegionByName("abcd", nil)
+	if err != nil {
+		t.Errorf("Error!! Getting Region by Invalid Name %v", err)
+	}
+	if len(regionResp) >= 1 {
+		t.Errorf("Error!! Invalid Name shouldn't have any response %v Error %v", regionResp, err)
```
ocket8888, line 379: You don't need to prefix test errors with "Error!! ", the output's gonna put a big "ERROR" in front of it for you

traffic_ops/testing/api/v4/regions_test.go
dmohan001c: Fixed.

traffic_ops/testing/api/v4/regions_test.go
dmohan001c: updated as per review comments.

traffic_ops/testing/api/v4/regions_test.go
dmohan001c: Fixed.,

traffic_ops/testing/api/v4/regions_test.go
dmohan001c: Fixed.

traffic_ops/testing/api/v4/regions_test.go
dmohan001c: Fixed

traffic_ops/testing/api/v4/regions_test.go
```diff
@@ -175,15 +226,90 @@ func UpdateTestRegions(t *testing.T) {
...
+		regionsWithLimit, _, err := TOSession.GetRegions(qparams, nil)
+		if err == nil {
+			if !reflect.DeepEqual(regions[:1], regionsWithLimit) {
+				t.Error("expected GET Regions with limit = 1 to return first result")
+			}
+		} else {
+			t.Error("Error in getting regions by limit")
 		}
+
+		if len(regions) > 1 {
```
ocket8888, line 253: The `regions` slice in here is sliced to `1:2` which will segfault if it can't get index `1` (or, equivalently, `:2`) - which would mean they need to have a length of at least 2

dmohan001c: yes, if the condition falls `if len(regions) > 1` it should be greater than or equal to 2. It won't segfault.

ocket8888: oh,. you're right, of course. My mistake

traffic_ops/testing/api/v4/regions_test.go
```diff
@@ -142,6 +161,38 @@ func SortTestRegions(t *testing.T) {
...
+			if respDesc[0].Name != "" && respAsc[0].Name != "" {
+				if !reflect.DeepEqual(respDesc[0].Name, respAsc[0].Name) {
+					t.Errorf("Regions responses are not equal after reversal: %s - %s", respDesc[0].Name, respAsc[0].Name)
+				}
+			}
+		} else {
+			t.Errorf("No Response returned from GET Regions using SortOrder")
+		}
+	} else {
+		t.Fatalf("Region response length are not equal %v %v", respAsc, respAsc)
```
ocket8888, line 192: I think you wanted to print `respAsc` and `descAsc` here, not `respAsc` twice - but also maybe you only wanted to print the lengths?

ocket8888 (approved):
LGTM

PR #5753, opened 2021-04-19
Deprecate TO API route `/vault/bucket/:bucket/key/:key/values`

rawlinp:
Remove the TO API route `/vault/bucket/:bucket/key/:key/values` from API v4.

- Traffic Ops

`GET /api/4.0/vault/bucket/foo/key/bar/values` should return 404 not found.

- [x] Route removal, no tests necessary


zrhoffman (approved):
Looks good, the proposal for removing this endpoint already reached consensus on dev@trafficcontrol.apache.org: https://lists.apache.org/thread.html/r7263c53603a8d93014accac4e59f316eb6a30a5ddceda9905ab647a7%40%3Cdev.trafficcontrol.apache.org%3E

zrhoffman (commented):
When building the docs:
```rst
undefined label: to-api-vault-bucket-bucket-key-key-values
```

So on our *Migrating from API V1* page,
https://github.com/apache/trafficcontrol/blob/22a6e2a1a7b731af11d38d190c236467ad14c3bc/docs/source/api/migrating-from-v1.rst#L500-L502
should be changed to either reference `to-api-v3-vault-bucket-bucket-key-key-values` or be `|none|`, meaning no newer TO API equivalent.

ocket8888:
I've seen that, and was just going to simply ignore it until #5539 is merged, since that PR would get rid of the table entirely.

PR #5755, opened 2021-04-19
Return legacy Monitoring Config from legacy Monitoring Config handler

### LINKED ISSUE #5754: API v1 monitoring.json Profile Parameters no longer contain `health.threshold.*`
## I'm submitting a ...

-  bug report

-  Traffic Monitor
-  Traffic Ops

## Current behavior:

The `/api/1.3/cdns/{name}/configs/monitoring.json` and `/api/2.0/cdns/{name}/configs/monitoring.json` return Profiles with Health Threshold parameters structured differently than ATC 4.1.1 and before had them in. What they look like at 473d9c2422:

```json
          "health_threshold": {
            "availableBandwidthInKbps": {
              "Comparator": ">",
              "Val": 1750000
            },
            "loadavg": {
              "Comparator": "<",
              "Val": 25
            },
            "queryTime": {
              "Comparator": "<",
              "Val": 1000
            }
```

## Expected behavior:

The `/api/1.3/cdns/{name}/configs/monitoring.json` and `/api/2.0/cdns/{name}/configs/monitoring.json` should use the same format as previous releases used:
   https://github.com/apache/trafficcontrol/blob/473d9c242247618f17ac64a2b4e53e7e14fdc7ed/docs/source/api/v2/cdns_name_configs_monitoring.rst#L197-L199

1. `GET /api/1.3/cdns/{name}/configs/monitoring.json`
2. Notice the Health Threshold Paramters are not in the expected format given above.

## Anything else:

- This issue first appeared at 44d79af799 (#4774).

- Attached:
    * [A monitoring.json from API v1 at RELEASE-4.1.1](https://github.com/apache/trafficcontrol/files/6337471/monitoring-4.1.1.json.txt)
    * [A monitoring.json from API v1 at master](https://github.com/apache/trafficcontrol/files/6337472/monitoring-master.json.txt) at 473d9c2422

zrhoffman:

- [x] This PR fixes #5754 

- Traffic Monitor
- Traffic Ops

Run the TO v1 and TO v2 API tests

- master (473d9c2422)
- 5.1.2 (RC2)
- 5.1.1 (and all 5.x.x through 5.0.0)

- [x] Bugfix, documentation is unnecessary, though the release notes for the next release should mention that it is the first release for which Traffic Ops can be used with a ATC 4.1.1 Traffic Monitor.


ocket8888:
It's curious that all the CI tests pass; I'm getting this locally:

```
/trafficops-init.sh: line 119: /shared/enroller/federations/010-foo.kabletown.net.json: No such file or directory
```

which I believe based on context probably is because `/shared/enroller/federations` doesn't exist. Shouldn't that be handled automagically? It's just a volume, so I don't think I should have to re-build the enroller container, right?

zrhoffman:
> It's curious that all the CI tests pass; I'm getting this locally:
> 
> ```
> /trafficops-init.sh: line 119: /shared/enroller/federations/010-foo.kabletown.net.json: No such file or directory
> ```
> 
> which I believe based on context probably is because `/shared/enroller/federations` doesn't exist. Shouldn't that be handled automagically? It's just a volume, so I don't think I should have to re-build the enroller container, right?

`/traffic_ops_data/` (copied to `/shared/enroller/`) is a volume between all services, but it is not mounted, the `trafficops` Dockerfile copies it in.

ocket8888:
It looks like for whatever reason rebuilding the enroller was the key to getting the federation thing to work. Odd.

Anyway, is the intention of this PR to only change the behavior for the "legacy" monitoring API endpoint? Because in 4.0 I'm still seeing:

```json
[
        {
                "name": "CCR_CIAB",
                "type": "CCR",
                "parameters": null
        },
        {
                "name": "ATS_EDGE_TIER_CACHE",
                "type": "EDGE",
                "parameters": {
                        "health.connection.timeout": 2000,
                        "health.polling.url": "http://${hostname}/_astats?application=&inf.name=${interface_name}",
                        "health.threshold.availableBandwidthInKbps": ">1750000",
                        "health.threshold.loadavg": "25.0",
                        "health.threshold.queryTime": 1000,
                        "history.count": 30
                }
        },
        {
                "name": "ATS_MID_TIER_CACHE",
                "type": "MID",
                "parameters": {
                        "health.connection.timeout": 2000,
                        "health.polling.url": "http://${hostname}/_astats?application=&inf.name=${interface_name}",
                        "health.threshold.availableBandwidthInKbps": ">1750000",
                        "health.threshold.loadavg": "25.0",
                        "health.threshold.queryTime": 1000,
                        "history.count": 30
                }
        }
]
```

(from `toget -kpa 4.0 cdns/CDN-in-a-Box/configs/monitoring | jq '.response.profiles'`)

rawlinp:
> Anyway, is the intention of this PR to only change the behavior for the "legacy" monitoring API endpoint? Because in 4.0 I'm still seeing:

I'm not sure why the format change was even necessary in the first place, but this seems like it needs to be fixed in both TO _and_ TM. Basically, TO API 2.x needs to return the original format, and 3.x needs to continue returning the _new_ format. On the TM side, we need to properly "upgrade" the 2.x format into the 3.x format. Currently, it only looks like the "servers" field is upgraded properly from 2.x to 3.x.

Essentially, if we were upgrading from 4.1 to 5.0, we would need to be able to do both:
1. upgrade TO to 5.0 first, leaving TM at 4.1
2. upgrade TM to 5.0 first, leaving TO at 4.1

If we only fix TO to return the old format at the old API version, that only fixes the 1st scenario.

ocket8888:
Ah, so the behavior of `2,0/cdns/{{name}}/monitoring` wasn't matching documentation/expectations?

Then I think you're good

rawlinp:
> ... so I think the 2nd scenario is already covered.

Oh, I was confused. @ocket8888's comment made it sound like the format was changed as part of v3, but it wasn't. So yeah, the 2nd scenario is definitely covered. Carry on 😄

zrhoffman:
Hmm I'm noticing that API v1/v2/v3/v4 responses for `cdns/{name}/configs/monitoring` are now identical. Removed the legacy handler in 749ba38b60.

ocket8888:
> Hmm I'm noticing that API v1/v2/v3/v4 responses for `cdns/{name}/configs/monitoring` are now identical. Removed the legacy handler in [749ba38](https://github.com/apache/trafficcontrol/commit/749ba38b60825c601cd39cdeb47b8fd7d4d83ac8).

Well, the `config` section may be the same, but the `trafficServer` section is different, because as of APIv3 the servers will have interfaces

zrhoffman:
> Well, the `config` section may be the same, but the `trafficServer` section is different, because as of APIv3 the servers will have interfaces

From what I'm seeing, the Monitoring Config at the latest TO API version has been totally unchanged from 4.1.1 through master at f84ae27eda. It's identical, line for line. The `"trafficServer"` array using TO API 2.0 at RELEASE-4.1.1:
```json
    "trafficServers": [
      {
        "cachegroup": "CDN_in_a_Box_Edge",
        "fqdn": "edge.infra.ciab.test",
        "hashid": "7825f580-26e7-476f-af8a-8d5adfe7f741",
        "hostname": "edge",
        "interfaces": [
          {
            "ipAddresses": [
              {
                "address": "172.16.239.3",
                "gateway": "172.16.239.1",
                "serviceAddress": true
              },
              {
                "address": "fc01:9400:1000:8::3",
                "gateway": "fc01:9400:1000:8::1",
                "serviceAddress": true
              }
            ],
            "maxBandwidth": null,
            "monitor": true,
            "mtu": 1500,
            "name": "eth0"
          }
        ],
        "port": 80,
        "profile": "ATS_EDGE_TIER_CACHE",
        "status": "REPORTED",
        "type": "EDGE"
      },
```

Let me know if you can find any differences, but from my testing, my `diff` is completely empty when diffing the whole Monitoring Config between different TO versions.

zrhoffman:
Never mind, forgot to snap. Re-adding the legacy handler...

zrhoffman:
Rebased to fix a merge conflict in CHANGELOG.md.

zrhoffman:
Setting the Health Threshold Parameters when converting to the legacy-format Traffic Monitor Config in 662cbc95ec, this is ready for review again.

traffic_ops/testing/api/v1/crconfig_test.go
```diff
@@ -153,6 +155,89 @@ func UpdateTestCRConfigSnapshot(t *testing.T) {
...
+	const expectedThresholdParameters = 3
+	if len(parameterMap) != expectedThresholdParameters {
+		t.Fatalf("expected Profile '%s' to contain %d Parameters with names starting with '%s' but %d such Parameters were found", profileName, expectedThresholdParameters, tc.ThresholdPrefix, len(parameterMap))
+	}
+	tmConfig, _, err := TOSession.GetTrafficMonitorConfig(cdnName)
+	if err != nil {
+		t.Fatalf("getting Traffic Monitor Config: %s", err.Error())
+	}
+	profileFound := false
+	var profile tc.TMProfile
```
ocket8888, line 211: `profile` does not appear to be used in this scope

lib/go-tc/traffic_monitor.go
```diff
@@ -424,7 +445,14 @@ type TMParameters struct {
 	HealthPollingType       string `json:"health.polling.type"`
 	HistoryCount            int    `json:"history.count"`
 	MinFreeKbps             int64
-	Thresholds              map[string]HealthThreshold `json:"health_threshold"`
+	Thresholds              map[string]HealthThreshold `json:"health_threshold,omitempty"`
+	HealthThresholdJSONParameters
+}
+
+type HealthThresholdJSONParameters struct {
```
ocket8888, line 452: GoDoc?

traffic_ops/testing/api/v2/crconfig_test.go
```diff
@@ -153,6 +155,89 @@ func UpdateTestCRConfigSnapshot(t *testing.T) {
...
+	profiles, _, err := TOSession.GetProfileByName(profileName)
+	if err != nil {
+		t.Fatalf("getting Profiles with name '%s': %s", profileName, err.Error())
+	}
+	if len(cdns) < 1 {
+		t.Fatalf("expected to find a Profile named %s", profileName)
+	}
+	if len(cdns) > 1 {
+		t.Fatalf("expected exactly 1 Profiles named %s but found %d Profiles", profileName, len(profiles))
+	}
```
ocket8888, line 183: same as above RE: checking `cdns` instead of `profiles`

traffic_ops/testing/api/v2/crconfig_test.go
```diff
@@ -153,6 +155,89 @@ func UpdateTestCRConfigSnapshot(t *testing.T) {
...
+	const expectedThresholdParameters = 3
+	if len(parameterMap) != expectedThresholdParameters {
+		t.Fatalf("expected Profile '%s' to contain %d Parameters with names starting with '%s' but %d such Parameters were found", profileName, expectedThresholdParameters, tc.ThresholdPrefix, len(parameterMap))
+	}
+	tmConfig, _, err := TOSession.GetTrafficMonitorConfig(cdnName)
+	if err != nil {
+		t.Fatalf("getting Traffic Monitor Config: %s", err.Error())
+	}
+	profileFound := false
+	var profile tc.TMProfile
```
ocket8888, line 211: same as above RE: variable not used in the scope of its declaration

traffic_ops/testing/api/v3/crconfig_test.go
```diff
@@ -154,6 +156,89 @@ func UpdateTestCRConfigSnapshot(t *testing.T) {
...
+	}
+	profiles, _, err := TOSession.GetProfileByNameWithHdr(profileName, nil)
+	if err != nil {
+		t.Fatalf("getting Profiles with name '%s': %s", profileName, err.Error())
+	}
+	if len(cdns) < 1 {
+		t.Fatalf("expected to find a Profile named %s", profileName)
+	}
+	if len(cdns) > 1 {
+		t.Fatalf("expected exactly 1 Profiles named %s but found %d Profiles", profileName, len(profiles))
```
ocket8888, line 183: same as above RE: checking `cdns` instead of `profiles`

traffic_ops/testing/api/v4/crconfig_test.go
```diff
@@ -203,6 +205,89 @@ func UpdateTestCRConfigSnapshot(t *testing.T) {
...
+	}
+	profiles, _, err := TOSession.GetProfileByName(profileName, nil)
+	if err != nil {
+		t.Fatalf("getting Profiles with name '%s': %s", profileName, err.Error())
+	}
+	if len(cdns) < 1 {
+		t.Fatalf("expected to find a Profile named %s", profileName)
+	}
+	if len(cdns) > 1 {
+		t.Fatalf("expected exactly 1 Profiles named %s but found %d Profiles", profileName, len(profiles))
```
ocket8888, line 232: same as above RE: checking `cdns` instead of `profiles`

traffic_ops/testing/api/v3/crconfig_test.go
```diff
@@ -154,6 +156,89 @@ func UpdateTestCRConfigSnapshot(t *testing.T) {
...
+	const expectedThresholdParameters = 3
+	if len(parameterMap) != expectedThresholdParameters {
+		t.Fatalf("expected Profile '%s' to contain %d Parameters with names starting with '%s' but %d such Parameters were found", profileName, expectedThresholdParameters, tc.ThresholdPrefix, len(parameterMap))
+	}
+	tmConfig, _, err := TOSession.GetTrafficMonitorConfig(cdnName)
+	if err != nil {
+		t.Fatalf("getting Traffic Monitor Config: %s", err.Error())
+	}
+	profileFound := false
+	var profile tc.TMProfile
```
ocket8888, line 212: same as above RE: variable not used in the scope of its declaration

traffic_ops/testing/api/v4/crconfig_test.go
```diff
@@ -203,6 +205,89 @@ func UpdateTestCRConfigSnapshot(t *testing.T) {
...
+	const expectedThresholdParameters = 3
+	if len(parameterMap) != expectedThresholdParameters {
+		t.Fatalf("expected Profile '%s' to contain %d Parameters with names starting with '%s' but %d such Parameters were found", profileName, expectedThresholdParameters, tc.ThresholdPrefix, len(parameterMap))
+	}
+	tmConfig, _, err := TOSession.GetTrafficMonitorConfig(cdnName)
+	if err != nil {
+		t.Fatalf("getting Traffic Monitor Config: %s", err.Error())
+	}
+	profileFound := false
+	var profile tc.TMProfile
```
ocket8888, line 261: same as above RE: variable not used in the scope of its declaration

traffic_ops/testing/api/v1/crconfig_test.go
```diff
@@ -153,6 +155,89 @@ func UpdateTestCRConfigSnapshot(t *testing.T) {
...
+	}
+	profiles, _, err := TOSession.GetProfileByName(profileName)
+	if err != nil {
+		t.Fatalf("getting Profiles with name '%s': %s", profileName, err.Error())
+	}
+	if len(cdns) < 1 {
+		t.Fatalf("expected to find a Profile named %s", profileName)
+	}
+	if len(cdns) > 1 {
+		t.Fatalf("expected exactly 1 Profiles named %s but found %d Profiles", profileName, len(profiles))
```
ocket8888, line 182: I'm not sure why this didn't make it into my review, but it was supposed to:

This seems like it should be checking the length of `profiles`, not `cdns`.

Also, I think if you remove the `len(thing) < 1` blocks and change the `len(thing) > 1` blocks' conditions to `len(thing) != 1` the error message therein will suffice to cover both cases. It's too late to save some typing now, but I thought I'd point it out in case you wanna change it anyway.

lib/go-tc/traffic_monitor.go
zrhoffman: Godocs added in 4f368bdc30.

traffic_ops/testing/api/v1/crconfig_test.go
zrhoffman: Yes it is. It's used down here: https://github.com/apache/trafficcontrol/blob/4f368bdc30c847d6b091b1da5114776bd75fa59c/traffic_ops/testing/api/v1/crconfig_test.go#L221

If it was actually unused in that scope, Go would throw a compile error.

traffic_ops/testing/api/v2/crconfig_test.go
zrhoffman: It's used down here: https://github.com/apache/trafficcontrol/blob/4f368bdc30c847d6b091b1da5114776bd75fa59c/traffic_ops/testing/api/v2/crconfig_test.go#L221

traffic_ops/testing/api/v3/crconfig_test.go
zrhoffman: Here is where it is used: https://github.com/apache/trafficcontrol/blob/4f368bdc30c847d6b091b1da5114776bd75fa59c/traffic_ops/testing/api/v3/crconfig_test.go#L222

traffic_ops/testing/api/v4/crconfig_test.go
zrhoffman: Variable used in the scope of its declaration: https://github.com/apache/trafficcontrol/blob/4f368bdc30c847d6b091b1da5114776bd75fa59c/traffic_ops/testing/api/v4/crconfig_test.go#L222

traffic_ops/testing/api/v1/crconfig_test.go
zrhoffman: Checking the length of `profiles` in 3963215ba4.

Removed the redundant condition in d5387d1139.

traffic_ops/testing/api/v2/crconfig_test.go
zrhoffman: Checking the length of `profiles` in 3963215ba4

traffic_ops/testing/api/v3/crconfig_test.go
zrhoffman: Checking the length of `profiles` in 3963215ba4

traffic_ops/testing/api/v4/crconfig_test.go
zrhoffman: Checking the length of `profiles` in 3963215ba4

traffic_ops/testing/api/v1/crconfig_test.go
ocket8888: [No it wouldn't](https://play.golang.org/p/dj8TDpb8Ii5), but my mistake. I checked like 5 times, can't believe I didn't see that.

ocket8888 (approved):

PR #5778, opened 2021-04-23
Fix APIv4 /servers/{{host name}}/update_status to use 'response' object

ocket8888:

This PR fixes the response type for the `/servers/{{host name}}/update_status` API endpoint (version 4.0) to use the same consistent `response` object as all the others.

- Documentation
- Traffic Control Client (Go)
- Traffic Ops

Make sure the API tests still pass, make sure the output reflects the documented behavior, the documentation builds with no errors or warnings, and that the documented/observed behavior is, in fact, consistent with the rest of the API

- [x] This PR includes tests


mattjackson220 (approved):
Looks good to me! Tests pass, code looks good, endpoint uses 'response' object as expected.

PR #5780, opened 2021-04-23
Opt out of Google FLoC

ocket8888:

This PR adds an HTTP header (`Permissions-Policy: interest-cohort()`) to the responses of the Traffic Ops API, Traffic Portal's proxy and static file handlers, and Traffic Monitor's API and static file handlers. It would also add it to Traffic Router if I could figure out how.

This header opts the servers out of Google's new browser-baked-in fingerprinting and tracking system called FLoC. FLoC tracks users browsing activity and reports it to Google to be sold to third parties for the purpose of targeted advertising. For a more accurate and in-depth description, refer to [The Electronic Frontier Foundation's article/open letter to Google on the subject](https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea).

The behavior is present in Chrome browsers at version 89 or later (and possibly other Chromium-based browsers, I haven't looked into that).

- Traffic Monitor
- Traffic Ops
- Traffic Portal

Make requests to the TO API, verify the presence of the new header, do the same with Traffic Monitor, load the TM UI and verify the presence of the new header, load Traffic Portal and verify that both static file requests and proxied API requests have the new header.

- [x] Tests are unnecessary

### DIFF
--- a/lib/go-rfc/http.go
+++ b/lib/go-rfc/http.go
@@ -36,6 +36,8 @@ const (
 	ContentDisposition = "Content-Disposition" // RFC6266
 	ContentEncoding    = "Content-Encoding"    // RFC7231§3.1.2.2
 	ContentType        = "Content-Type"        // RFC7231§3.1.1.5
+	PermissionsPolicy  = "Permissions-Policy"  // W3C "Permissions Policy"
+	Server             = "Server"              // RFC7231§7.4.2
 	Vary               = "Vary"                // RFC7231§7.1.4
 )
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added integration to use ACME to generate new SSL certificates.
 - Add a Federation to the Ansible Dataset Loader
 - Added asynchronous status to ACME certificate generation.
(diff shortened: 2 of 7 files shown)


lib/go-rfc/http.go
```diff
@@ -36,6 +36,8 @@ const (
 	ContentDisposition = "Content-Disposition" // RFC6266
 	ContentEncoding    = "Content-Encoding"    // RFC7231§3.1.2.2
 	ContentType        = "Content-Type"        // RFC7231§3.1.1.5
+	PermissionsPolicy  = "Permissions-Policy"  // W3C "Permissions Policy"
+	Server             = "Server"              // RFC7231§7.4.2
```
zrhoffman, line 40: Were you planning to do something with `rfc.Server`?

ocket8888: I was, but then I decided I should do it in a different PR. But I didn't want to look up the RFC number and section again, and it's not hurting anything so I left it in.

ocket8888: Do you want me to take it out? I don't really feel strongly about it. It's an artifact of laziness.

zrhoffman: Your reason for leaving it in made sense to me, leaving it sounds fine

zrhoffman (approved):

PR #5791, opened 2021-04-27
Ats 9 logging support

ezelkow1:
This PR adds a top level `logging` scope to the logging.yaml. This is needed by ATS 9

- Traffic Ops ORT

Test with a cache using ATS 8, verify that logging.yaml is generated correctly with proper indentation. Change the cache to use an ATS 9 RPM file via its profile, then update its files again checking the logging.yaml for consistency and that ATS can start and log in both scenarios.


traeak (approved):
You could separate indent level and spaces per indent, but that's not something critical to do.

PR #5801, opened 2021-04-30
URL Sig Keys in Postgres and DELETE methods

mattjackson220:

This PR adds support for URL Sig Keys using Postgres as the Traffic Vault backend.  It also adds DELETE functionality for URL Sig Keys for both Postgres and Riak backends.

- Traffic Ops
- Traffic Vault
- CI tests

Make sure you have Traffic Vault set up to use Postgres in your test environment.
Update cdn.conf to use Postgres
Verify that all of these endpoints work as expected:
POST `deliveryservices/xmlId/{name}/urlkeys/copyFromXmlId/{copy-name}/?$`,
POST `deliveryservices/xmlId/{name}/urlkeys/generate/?$`
GET `deliveryservices/xmlId/{name}/urlkeys/?$`
DELETE 	`deliveryservices/xmlId/{name}/urlkeys/?$`
GET `deliveryservices/{id}/urlkeys/?$`
DELETE `deliveryservices/{id}/urlkeys/`

Verify that the data from those endpoints is successfully going into your TV in Postgres.
Run it again without the Postgres backend to point back to Riak.
Verify that all of those endpoints work with Riak still and that the 2 new DELETE endpoints work.
Verify that the API tests run successfully (Postgres TV in ciab is not set up yet so verify that the tests work like normal with Riak and that the new tests work locally or a non-ciab environment with Postgres)

No doc updates since its using the same endpoints / format as before just a different backend.

### DIFF
--- a/docs/source/api/v4/deliveryservices_xmlid_xmlid_urlkeys.rst
+++ b/docs/source/api/v4/deliveryservices_xmlid_xmlid_urlkeys.rst
@@ -66,3 +66,36 @@ Response Structure
 		"key14":"DtXsu8nsw04YhT0kNoKBhu2G3P9WRpQJ",
 		"key7":"cmKoIIxXGAxUMdCsWvnGLoIMGmNiuT5I"
 	}}
+
+
+``DELETE``
+==========
+.. seealso:: :ref:`to-api-deliveryservices-id-urlkeys`
+
+Deletes URL signing keys for a :term:`Delivery Service`.
+
+:Auth. Required: Yes
+:Roles Required: "admin" or "operations"
+:Response Type:  Object
+
+Request Structure
+-----------------
+.. table:: Request Path Parameters
+
+	+-------+------------------------------------------------------+
+	|  Name |              Description                             |
+	+=======+======================================================+
+	| xmlid | The 'xml_id' of the desired :term:`Delivery Service` |
+	+-------+------------------------------------------------------+
+
+Response Structure
+------------------
+.. code-block:: json
+	:caption: Response Example
+
+	{
+		"alerts": [{
+			"level": "success",
+			"text": "Successfully deleted URL Sig keys from Traffic Vault"
+		}]
+	}
--- a/traffic_ops/testing/api/v4/deliveryservices_test.go
+++ b/traffic_ops/testing/api/v4/deliveryservices_test.go
@@ -1390,12 +1390,30 @@ func CreateTestDeliveryServicesURLSigKeys(t *testing.T) {
 		t.Error("failed to create url sig keys: " + err.Error())
 	}
 
-	newKeys, _, err := TOSession.GetDeliveryServiceURLSigKeys(*firstDS.XMLID, nil)
+	firstKeys, _, err := TOSession.GetDeliveryServiceURLSigKeys(*firstDS.XMLID, nil)
 	if err != nil {
 		t.Error("failed to get url sig keys: " + err.Error())
 	}
-	if len(newKeys) == 0 {
+	if len(firstKeys) == 0 {
 		t.Errorf("failed to create url sig keys")
+	}
+
+	// Create new keys again and check that they are different
+	_, _, err = TOSession.CreateDeliveryServiceURLSigKeys(*firstDS.XMLID, nil)
+	if err != nil {
+		t.Error("failed to create url sig keys: " + err.Error())
+	}
+
+	secondKeys, _, err := TOSession.GetDeliveryServiceURLSigKeys(*firstDS.XMLID, nil)
+	if err != nil {
+		t.Error("failed to get url sig keys: " + err.Error())
+	}
+	if len(secondKeys) == 0 {
+		t.Errorf("failed to create url sig keys")
+	}
+
+	if secondKeys["key0"] == firstKeys["key0"] {
+		t.Errorf("second create did not generate new url sig keys")
 	}
 }
 
--- a/traffic_ops/testing/api/v4/deliveryservices_test.go
+++ b/traffic_ops/testing/api/v4/deliveryservices_test.go
@@ -44,7 +44,9 @@ func TestDeliveryServices(t *testing.T) {
 
 		if includeSystemTests {
 			SSLDeliveryServiceCDNUpdateTest(t)
+			CreateTestDeliveryServicesURLSigKeys(t)
 			GetTestDeliveryServicesURLSigKeys(t)
+			DeleteTestDeliveryServicesURLSigKeys(t)
 		}
 
 		GetTestDeliveryServicesIMS(t)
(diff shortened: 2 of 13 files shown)


traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/url_sig_keys.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/apache/trafficcontrol/lib/go-tc"
+
+	"github.com/jmoiron/sqlx"
+)
+
+func getURLSigKeys(xmlID string, tvTx *sqlx.Tx) (tc.URLSigKeys, bool, error) {
+	var jsonUrlKeys string
+	if err := tvTx.QueryRow("SELECT data FROM url_sig_key WHERE deliveryservice = $1", xmlID).Scan(&jsonUrlKeys); err != nil {
+		if err == sql.ErrNoRows {
+			return tc.URLSigKeys{}, true, nil
```
rawlinp, line 36: I think this should return `false` because the row didn't exist and was hence "not found"

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -133,11 +133,32 @@ func (p *Postgres) DeleteDNSSECKeys(cdnName string, tx *sql.Tx, ctx context.Cont
 }
 
 func (p *Postgres) GetURLSigKeys(xmlID string, tx *sql.Tx, ctx context.Context) (tc.URLSigKeys, bool, error) {
-	return tc.URLSigKeys{}, false, notImplementedErr
+	tvTx, dbCtx, cancelFunc, err := p.beginTransaction(ctx)
+	if err != nil {
+		return tc.URLSigKeys{}, false, err
+	}
+	defer p.commitTransaction(tvTx, dbCtx, cancelFunc)
+	return getURLSigKeys(xmlID, tvTx)
```
rawlinp, line 141: I think we may want to pass `ctx` into these function as well in order to use `checkErrWithContext` after attempting the DB query. That function is basically there to give us better context in the error message in case the client closed the connection or the query timed out -- both of which _cancel_ the context, cancelling the DB query.

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/url_sig_keys.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		return tc.URLSigKeys{}, false, err
```
rawlinp, line 38: Before this return is where we should call `checkErrWithContext` in case the context was cancelled

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -133,11 +133,32 @@ func (p *Postgres) DeleteDNSSECKeys(cdnName string, tx *sql.Tx, ctx context.Cont
...
 
 func (p *Postgres) PutURLSigKeys(xmlID string, keys tc.URLSigKeys, tx *sql.Tx, ctx context.Context) error {
-	return notImplementedErr
+	tvTx, dbCtx, cancelFunc, err := p.beginTransaction(ctx)
+	if err != nil {
+		return err
+	}
+	defer p.commitTransaction(tvTx, dbCtx, cancelFunc)
+
+	return putURLSigKeys(xmlID, tvTx, keys)
```
rawlinp, line 151: ditto prior comment

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -133,11 +133,32 @@ func (p *Postgres) DeleteDNSSECKeys(cdnName string, tx *sql.Tx, ctx context.Cont
...
+}
+
+func (p *Postgres) DeleteURLSigKeys(xmlID string, tx *sql.Tx, ctx context.Context) error {
+	tvTx, dbCtx, cancelFunc, err := p.beginTransaction(ctx)
+	if err != nil {
+		return err
+	}
+	defer p.commitTransaction(tvTx, dbCtx, cancelFunc)
+
+	return deleteURLSigKeys(xmlID, tvTx)
```
rawlinp, line 161: ditto prior comment

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/url_sig_keys.go
```diff
@@ -0,0 +1,73 @@
...
+
+func putURLSigKeys(xmlID string, tvTx *sqlx.Tx, keys tc.URLSigKeys) error {
+	keyJSON, err := json.Marshal(&keys)
+	if err != nil {
+		return errors.New("marshalling keys: " + err.Error())
+	}
+
+	res, err := tvTx.Exec("INSERT INTO url_sig_key (deliveryservice, data) VALUES ($1, $2)", xmlID, keyJSON)
+	if err != nil {
+		return err
```
rawlinp, line 58: Before this return is where we should call `checkErrWithContext` in case the context was cancelled

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/url_sig_keys.go
```diff
@@ -0,0 +1,73 @@
...
+	} else if rowsAffected == 0 {
+		return errors.New("URL Sign Keys: no keys were inserted")
+	}
+	return nil
+}
+
+func deleteURLSigKeys(xmlID string, tvTx *sqlx.Tx) error {
+	if _, err := tvTx.Exec("DELETE FROM url_sig_key WHERE deliveryservice = $1", xmlID); err != nil {
+		return err
```
rawlinp, line 70: Before this return is where we should call `checkErrWithContext` in case the context was cancelled

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/url_sig_keys.go
```diff
@@ -0,0 +1,73 @@
...
+	return urlSignKey, true, nil
+}
+
+func putURLSigKeys(xmlID string, tvTx *sqlx.Tx, keys tc.URLSigKeys) error {
+	keyJSON, err := json.Marshal(&keys)
+	if err != nil {
+		return errors.New("marshalling keys: " + err.Error())
+	}
+
+	res, err := tvTx.Exec("INSERT INTO url_sig_key (deliveryservice, data) VALUES ($1, $2)", xmlID, keyJSON)
```
rawlinp, line 56: This query needs to be idempotent, so we may need to `DELETE FROM url_sig_key WHERE deliveryservice = $1` before running the `INSERT`. The way this is currently, the request will fail if someone tries to generate URL sig keys more than once on the same DS without deleting them first.

traffic_ops/v4-client/deliveryservice.go
```diff
@@ -329,6 +335,25 @@ func (to *Session) GetDeliveryServiceURLSigKeys(dsName string, header http.Heade
 	return data.Response, reqInf, nil
 }
 
+// CreateDeliveryServiceURLSigKeys creates new URL-signing keys used by the Delivery Service
+// identified by the XMLID 'dsName'
+func (to *Session) CreateDeliveryServiceURLSigKeys(dsName string, header http.Header) (tc.Alerts, toclientlib.ReqInf, error) {
+	var alerts tc.Alerts
+	reqInf, err := to.post(fmt.Sprintf(APIDeliveryServicesURLSigKeysGenerate, dsName), nil, header, &alerts)
+	if err != nil {
```
rawlinp, line 343: This `err != nil` check is redundant, we can just return `err` regardless

rawlinp (changes requested):

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -1252,3 +1254,42 @@ func GetTestDeliveryServicesURLSigKeys(t *testing.T) {
...
+		t.Error("failed to create url sig keys: " + err.Error())
+	}
+
+	newKeys, _, err := TOSession.GetDeliveryServiceURLSigKeys(*firstDS.XMLID, nil)
+	if err != nil {
+		t.Error("failed to get url sig keys: " + err.Error())
+	}
+	if len(newKeys) == 0 {
+		t.Errorf("failed to create url sig keys")
+	}
```
rawlinp, line 1278: For this test we should call `Create...` again and `Get...` again and make sure the two generated keys are different. This will also ensure that `Create...` is idempotent.

rawlinp (commented):
This all looks good, but I think we need an API doc for the new `DELETE` method and there is also a merge conflict to resolve.

rawlinp (commented):
I pulled this down and tested it out, and it all worked beautifully. However, I noticed the new `DELETE` methods don't create a changelog entry stating that URL sig keys were deleted for the given DS. Once that is added, I think this is good to go!

rawlinp (approved):

ocket8888:
Whoops, it seems I waited too long to hit "Submit Review"

docs/source/api/v4/deliveryservices_xmlid_xmlid_urlkeys.rst
```diff
@@ -66,3 +66,36 @@ Response Structure
...
+:Response Type:  Object
+
+Request Structure
+-----------------
+.. table:: Request Path Parameters
+
+	+-------+------------------------------------------------------+
+	|  Name |              Description                             |
+	+=======+======================================================+
+	| xmlid | The 'xml_id' of the desired :term:`Delivery Service` |
```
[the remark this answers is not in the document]
mattjackson220: oh good call. thats copy / paste leftover from the GET docs above but ill open a new PR to fix both!

PR #5807, opened 2021-05-03
t3c/torequest: syncds only checks if packages exist, doesn't try to fix them

traeak:
…ix them

- Traffic Ops ORT

With a server fully configure in TC:  Remove ATS and manually install different ATS packages.
Run syncds.  Update will print out warnings about package mismatch but install will continue.

- Traffic Ops ORT
- 

### DIFF
--- a/traffic_ops_ort/t3c/torequest/torequest.go
+++ b/traffic_ops_ort/t3c/torequest/torequest.go
@@ -1194,7 +1194,7 @@ func (r *TrafficOpsReq) ProcessConfigFiles() (UpdateStatus, error) {
 		if strings.Contains(cfg.Path, "/opt/trafficserver/") || strings.Contains(cfg.Dir, "udev") {
 			cfg.Service = "trafficserver"
 			if r.Cfg.RunMode == config.SyncDS && !r.IsPackageInstalled("trafficserver") {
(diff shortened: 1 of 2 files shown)


traffic_ops_ort/t3c/torequest/torequest.go
```diff
@@ -1281,83 +1281,103 @@ func (r *TrafficOpsReq) ProcessPackages() error {
...
+			}
+		} else if r.Cfg.RunMode == config.SyncDS {
+			// Only check if packages exist and complain if they are wrong.
+			if instpkg == fullPackage {
+				log.Infof("%s Currently installed.\n", reqpkg)
+				r.pkgs[fullPackage] = true
+				continue
+			} else if instpkg != "" { // the installed package needs upgrading.
+				log.Infof("%s Wrong version currently installed.\n", instpkg)
+				log.Errorf("%s Wrong version currently installed.\n", instpkg)
```
rob05c, line 1321: Nitpick: if this is an Error, doesn't need to also be an Info.

rob05c (approved):
Looks good! Code looks good, tested, works as expected, syncds didn't install a new ats version, badass did.

traffic_ops_ort/t3c/torequest/torequest.go
traeak: i'll remove those.

PR #5810, opened 2021-05-04
Fix CDN DNSSEC key generation timeout

### LINKED ISSUE #5732: TO API POST /cdns/dnsseckeys/generate times out with large numbers of delivery services
## I'm submitting a ...
-  bug report

-  Traffic Ops
-  Traffic Vault

## Current behavior:
With a large number of delivery services (e.g. 266), the `POST /cdns/dnsseckeys/generate` route times out and returns a 503 because the operation exceeds 1 minute.

## Expected behavior:
The operation should succeed in less than 1 minute instead of timing out.

In a CDN with a large number of delivery services (e.g. 266), call this API, and observe that it times out after 1 minute and returns a 503.

## Anything else:
This was observed on a VM w/ 8 CPUs at 2596.986 MHz, so the results might be different on baremetal or more powerful VMs. However, we should be able to run TO on low-performance VMs. I would propose that we parallelize this CPU-intensive operation so that we can take advantage of multicore servers.

rawlinp:
With large numbers of delivery services, serially generating DNSSEC keys
for a CDN can take longer than the default 20 second timeout, causing
the DNSSEC key generation operation to time out and fail. Instead,
generate DNSSEC keys in parallel (by default using 66% of the available
CPUs) in order to complete DNSSEC key generation more quickly. Also, use
a separate DB transaction timeout for writing the changelog.

- [x] This PR fixes #5732 

- Documentation
- Traffic Control Client (Go)
- Traffic Ops
- Traffic Vault

In an environment with a large number of delivery services (e.g. > 500), generate DNSSEC keys for the CDN. The operation should succeed.

- [x] This PR includes tests

### DIFF
--- a/docs/source/api/v2/cdns_dnsseckeys_generate.rst
+++ b/docs/source/api/v2/cdns_dnsseckeys_generate.rst
@@ -32,7 +32,6 @@ Request Structure
 :effectiveDate:         An optional string containing the date and time at which the newly-generated :abbr:`ZSK (Zone-Signing Key)` and :abbr:`KSK (Key-Signing Key)` become effective, in :RFC:`3339` format. Defaults to the current time if not specified.
 :key:                   Name of the CDN
 :kskExpirationDays:     Expiration (in days) for the :abbr:`KSKs (Key-Signing Keys)`
-:name:                  Domain name used by the CDN
 :ttl:                   Time, in seconds, for which the keypairs shall remain valid
 :zskExpirationDays:     Expiration (in days) for the :abbr:`ZSKs (Zone-Signing Keys)`
 
@@ -50,7 +49,6 @@ Request Structure
 	{
 		"key": "CDN-in-a-Box",
 		"kskExpirationDays": 1095,
-		"name": "mycdn.ciab.test",
 		"ttl": 3600,
 		"zskExpirationDays": 1095
 	}
--- a/lib/go-tc/cdns_dnssec.go
+++ b/lib/go-tc/cdns_dnssec.go
@@ -18,10 +18,12 @@ import (
 	"database/sql"
 	"errors"
 	"strconv"
-	"strings"
 	"time"
 
+	"github.com/apache/trafficcontrol/lib/go-tc/tovalidate"
 	"github.com/apache/trafficcontrol/lib/go-util"
+
+	validation "github.com/go-ozzo/ozzo-validation"
 )
 
 const (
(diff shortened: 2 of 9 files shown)


docs/source/api/v2/cdns_dnsseckeys_generate.rst
```diff
@@ -32,7 +32,6 @@ Request Structure
 :effectiveDate:         An optional string containing the date and time at which the newly-generated :abbr:`ZSK (Zone-Signing Key)` and :abbr:`KSK (Key-Signing Key)` become effective, in :RFC:`3339` format. Defaults to the current time if not specified.
 :key:                   Name of the CDN
 :kskExpirationDays:     Expiration (in days) for the :abbr:`KSKs (Key-Signing Keys)`
-:name:                  Domain name used by the CDN
```
ocket8888, line 35: I don't think we can make this breaking change to the v2 API, as it's already been released

rawlinp: It's not really a breaking change, it's a completely unused field that has no effect on behavior/data whatsoever. API users can continue to send it, it will just be totally ignored (as it basically was before, except without the unnecessary validation).

traffic_ops/traffic_ops_golang/cdn/dnssec.go
```diff
@@ -86,7 +94,26 @@ func CreateDNSSECKeys(w http.ResponseWriter, r *http.Request) {
...
 	}
-	api.CreateChangeLogRawTx(api.ApiChange, "CDN: "+string(cdnName)+", ID: "+strconv.Itoa(cdnID)+", ACTION: Generated DNSSEC keys", inf.User, inf.Tx.Tx)
+	// NOTE: using a separate transaction (with its own timeout) for the changelog because the main
+	// transaction can time out if DNSSEC generation takes too long
+	db, err := api.GetDB(r.Context())
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("generating CDN DNSSEC keys: getting DB from request context for changelog: "+err.Error()))
+		return
+	}
+	logCtx, logCancel := context.WithTimeout(r.Context(), time.Duration(30)*time.Second)
```
zrhoffman, line 104: This can be just
```go
	logCtx, logCancel := context.WithTimeout(r.Context(), 30*time.Second)
```

docs/source/api/v2/cdns_dnsseckeys_generate.rst
ocket8888: Oh, this endpoint only has a POST method. I wouldn't have removed it even if it's no longer used, personally, but I suppose it's not a big deal.

lib/go-tc/cdns_dnssec.go
```diff
@@ -0,0 +1,163 @@
...
+	if r.KSKExpirationDays == nil {
+		errs = append(errs, "kskExpirationDays must be set")
+	}
+	if r.ZSKExpirationDays == nil {
+		errs = append(errs, "zskExpirationDays must be set")
+	}
+	// effective date is optional
+	if len(errs) > 0 {
+		return errors.New("missing fields: " + strings.Join(errs, "; "))
+	}
```
zrhoffman, line 161: Any reason not to use `tovalidate.ToErrors(validation.Errors{})` to validate required fields?

zrhoffman (changes requested):

docs/source/api/v2/cdns_dnsseckeys_generate.rst
rawlinp: I know, but I think we should reserve the right to remove things like this when it makes sense.

lib/go-tc/cdns_dnssec.go
rawlinp: I didn't add this code, I just moved it into a more targeted file and removed the `r.Name` check. But I could change it to use that function instead.

zrhoffman (approved):
Generating DNSSEC keys for a CDN with 502 DSes takes 2 minutes before this PR (after extending timeouts) and ~28 seconds as of this PR.

zrhoffman (approved):

PR #5812, opened 2021-05-04
Implement SSLKeys methods for PostgreSQL Traffic Vault backend

srijeet0406:

- Traffic Ops
- Traffic Vault

Setup TO to be configured to use `postgres` as the traffic vault backend.
Make sure you have the `tv_development` and `tv_test` databases and their corresponding tables setup correctly.
Try generating SSL keys for deliveryservices, and GETting them and make sure they work finr.
Check the same in the database.
Repeat for CDN SSL keys.

Make sure the unit and API tests pass.

- master

- [x] This PR includes tests.
- [x] Documentation already done in the initial PRs by Rawlin.
- [x] This PR does not include an update to CHANGELOG.md.

### DIFF
--- a/lib/go-tc/deliveryservices.go
+++ b/lib/go-tc/deliveryservices.go
@@ -30,13 +30,6 @@ const DefaultMaxRequestHeaderBytes = 0
 const MinRangeSliceBlockSize = 262144   // 265Kib
 const MaxRangeSliceBlockSize = 33554432 // 32Mib
 
-// SSLKeysAddResponse is a struct to store the response of addition of ssl keys for a DS,
-// along with any alert messages
-type SSLKeysAddResponse struct {
-	Response string `json:"response"`
-	Alerts
-}
-
 // GetDeliveryServiceResponse is deprecated use DeliveryServicesResponse...
 type GetDeliveryServiceResponse struct {
 	Response []DeliveryService `json:"response"`
--- a/lib/go-tc/deliveryservices.go
+++ b/lib/go-tc/deliveryservices.go
@@ -30,6 +30,13 @@ const DefaultMaxRequestHeaderBytes = 0
 const MinRangeSliceBlockSize = 262144   // 265Kib
 const MaxRangeSliceBlockSize = 33554432 // 32Mib
 
+// SSLKeysAddResponse is a struct to store the response of addition of ssl keys for a DS,
+// along with any alert messages
+type SSLKeysAddResponse struct {
+	Response string `json:"response"`
+	Alerts
+}
+
 // GetDeliveryServiceResponse is deprecated use DeliveryServicesResponse...
 type GetDeliveryServiceResponse struct {
 	Response []DeliveryService `json:"response"`
--- a/traffic_ops/testing/api/v4/deliveryservices_test.go
+++ b/traffic_ops/testing/api/v4/deliveryservices_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-rfc"
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/lib/go-util"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice"
 	toclient "github.com/apache/trafficcontrol/traffic_ops/v4-client"
 )
 
@@ -41,12 +42,13 @@ func TestDeliveryServices(t *testing.T) {
 		header = make(map[string][]string)
 		header.Set(rfc.IfModifiedSince, ti)
 		header.Set(rfc.IfUnmodifiedSince, ti)
-
 		if includeSystemTests {
 			SSLDeliveryServiceCDNUpdateTest(t)
 			CreateTestDeliveryServicesURLSigKeys(t)
 			GetTestDeliveryServicesURLSigKeys(t)
 			DeleteTestDeliveryServicesURLSigKeys(t)
+			DeleteCDNOldSSLKeys(t)
+			DeliveryServiceSSLKeys(t)
 		}
 
 		GetTestDeliveryServicesIMS(t)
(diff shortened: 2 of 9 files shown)


traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -171,13 +173,278 @@ func cleanUp(t *testing.T, ds tc.DeliveryServiceV4, oldCDNID int, newCDNID int)
...
+	customDS2.MultiSiteOrigin = util.BoolPtr(false)
+	customDS2.OrgServerFQDN = util.StrPtr("https://test2.com")
+	customDS2.Protocol = util.IntPtr(2)
+	customDS2.QStringIgnore = util.IntPtr(0)
+	customDS2.RangeRequestHandling = util.IntPtr(0)
+	customDS2.RegionalGeoBlocking = util.BoolPtr(false)
+	customDS2.TenantID = util.IntPtr(1)
+	customDS2.TypeID = util.IntPtr(types[0].ID)
+	customDS2.XMLID = util.StrPtr("dsID2")
+	customDS2.MaxRequestHeaderBytes = nil
```
mattjackson220, line 266: maybe pull these and the ones below into a getCustomDs func or something to cut down on repeated lines?

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -100,24 +101,149 @@ func (p *Postgres) commitTransaction(tx *sqlx.Tx, ctx context.Context, cancelFun
...
+	}
+	defer p.commitTransaction(tvTx, dbCtx, cancelFunc)
+
+	keyJSON, err := json.Marshal(&key)
+	if err != nil {
+		return errors.New("marshalling keys: " + err.Error())
+	}
+
+	// delete the old ssl keys first
+	_, err = tvTx.Exec("DELETE FROM sslkey WHERE deliveryservice=$1", key.DeliveryService)
```
mattjackson220, line 150: i dont think we want to delete the old keys. we just want to up the version and save that out. at least i believe thats how Riak worked

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -100,24 +101,149 @@ func (p *Postgres) commitTransaction(tx *sqlx.Tx, ctx context.Context, cancelFun
...
 func (p *Postgres) GetDeliveryServiceSSLKeys(xmlID string, version string, tx *sql.Tx, ctx context.Context) (tc.DeliveryServiceSSLKeysV15, bool, error) {
-	return tc.DeliveryServiceSSLKeysV15{}, false, notImplementedErr
+	tvTx, dbCtx, cancelFunc, err := p.beginTransaction(ctx)
+	if err != nil {
+		return tc.DeliveryServiceSSLKeysV15{}, false, err
+	}
+	defer p.commitTransaction(tvTx, dbCtx, cancelFunc)
+	var jsonKeys string
+	query := "SELECT data FROM sslkey WHERE deliveryservice=$1"
+	if version != "" {
```
mattjackson220, line 115: we should require version to be set otherwise we will get a list of old SSL keys

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -100,24 +101,149 @@ func (p *Postgres) commitTransaction(tx *sqlx.Tx, ctx context.Context, cancelFun
...
+// PutDeliveryServiceSSLKeys stores the given SSL keys for a delivery service.
 func (p *Postgres) PutDeliveryServiceSSLKeys(key tc.DeliveryServiceSSLKeys, tx *sql.Tx, ctx context.Context) error {
-	return notImplementedErr
+	tvTx, dbCtx, cancelFunc, err := p.beginTransaction(ctx)
+	if err != nil {
+		return err
+	}
+	defer p.commitTransaction(tvTx, dbCtx, cancelFunc)
+
+	keyJSON, err := json.Marshal(&key)
```
mattjackson220, line 144: we should check to make sure it has a version provided and if it doesnt then read the db and increase from the latest

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -100,24 +101,149 @@ func (p *Postgres) commitTransaction(tx *sqlx.Tx, ctx context.Context, cancelFun
...
+// if version is empty) for the delivery service identified by the given xmlID.
 func (p *Postgres) DeleteDeliveryServiceSSLKeys(xmlID string, version string, tx *sql.Tx, ctx context.Context) error {
-	return notImplementedErr
+	tvTx, dbCtx, cancelFunc, err := p.beginTransaction(ctx)
+	if err != nil {
+		return err
+	}
+	defer p.commitTransaction(tvTx, dbCtx, cancelFunc)
+	query := "DELETE FROM sslkey WHERE deliveryservice=$1"
+	if version != "" {
```
mattjackson220, line 176: we should require version in this query otherwise delete the latest. if version isnt provided as it is now then it will delete all ssl keys for that ds

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -100,24 +101,149 @@ func (p *Postgres) commitTransaction(tx *sqlx.Tx, ctx context.Context, cancelFun
...
+	var keys []tc.CDNSSLKey
+	var key tc.CDNSSLKey
+
+	tvTx, dbCtx, cancelFunc, err := p.beginTransaction(ctx)
+	if err != nil {
+		return keys, err
+	}
+	defer p.commitTransaction(tvTx, dbCtx, cancelFunc)
+
+	rows, err := tvTx.Query("SELECT data from sslkey WHERE cdn=$1", cdnName)
```
mattjackson220, line 228: we will also want to filter this query to only get the latest keys for each ds

mattjackson220 (changes requested):

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -171,13 +173,278 @@ func cleanUp(t *testing.T, ds tc.DeliveryServiceV4, oldCDNID int, newCDNID int)
...
+		Country:      util.StrPtr("CO"),
+		State:        util.StrPtr("ST"),
+	})
+	if err != nil {
+		t.Fatalf("unable to generate sslkeys for DS %v: %v", *customDS2.XMLID, err)
+	}
+
+	tries := 0
+	var cdnKeys []tc.CDNSSLKeys
+	for tries < 5 {
```
rawlinp, line 287: This loop could be more traditional like `for tries := 0; tries < 5; tries++`

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -171,13 +173,278 @@ func cleanUp(t *testing.T, ds tc.DeliveryServiceV4, oldCDNID int, newCDNID int)
...
+	if err != nil {
+		t.Errorf("could not delete delivery service %v", *ds.XMLID)
+	}
+	_, _, err = TOSession.SnapshotCRConfigByID(cdn.ID)
+	if err != nil {
+		t.Fatalf("couldn't snap CDN: %v", err)
+	}
+	tries = 0
+	var newCdnKeys []tc.CDNSSLKeys
+	for tries < 5 {
```
rawlinp, line 312: This loop could be more traditional like `for tries := 0; tries < 5; tries++`

traffic_ops/testing/api/v4/deliveryservices_test.go
rawlinp: +1 on this

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -171,13 +173,278 @@ func cleanUp(t *testing.T, ds tc.DeliveryServiceV4, oldCDNID int, newCDNID int)
...
+	}
+	defer cleanUp(t, ds, cdn.ID, -1)
+
+	if ds.XMLID == nil {
+		t.Fatalf("got a DS with an invalid xml ID")
+	}
+
+	tries := 0
+	var dsSSLKey *tc.DeliveryServiceSSLKeys
+	for tries < 5 {
```
rawlinp, line 384: This loop could be more traditional like `for tries := 0; tries < 5; tries++`

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -171,13 +173,278 @@ func cleanUp(t *testing.T, ds tc.DeliveryServiceV4, oldCDNID int, newCDNID int)
...
+	customDS.TypeID = util.IntPtr(types[0].ID)
+	customDS.XMLID = util.StrPtr("dsID2")
+	customDS.MaxRequestHeaderBytes = nil
+
+	ds, _, err := TOSession.CreateDeliveryService(customDS)
+	if err != nil {
+		t.Fatal(err)
+	}
+	ds.CDNName = &cdn.Name
+	_, _, err = TOSession.GenerateSSLKeysForDS(*ds.XMLID, *ds.CDNName, tc.SSLKeyRequestFields{
```
rawlinp, line 365: nothing to do here, just an observation: I have no idea why this method would need to take a CDN name. The CDN name can be inferred by the provided DS. I'll open a ticket to follow up on this.

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -171,13 +173,278 @@ func cleanUp(t *testing.T, ds tc.DeliveryServiceV4, oldCDNID int, newCDNID int)
...
+		Country:         &dsSSLKey.Country,
+		State:           &dsSSLKey.State,
+		Key:             &dsSSLKey.Key,
+		Version:         &dsSSLKey.Version,
+		Certificate:     &dsSSLKey.Certificate,
+	}
+	_, _, err = TOSession.AddSSLKeysForDS(tc.DeliveryServiceAddSSLKeysReq{DeliveryServiceSSLKeysReq: dsSSLKeyReq})
+
+	tries = 0
+	for tries < 5 {
```
rawlinp, line 428: This loop could be more traditional like `for tries := 0; tries < 5; tries++`

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
rawlinp: Yeah, `if version == ""` we should set it to `"latest"` and get that from the DB, but callers of this method should still be free to leave `version` empty.

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
rawlinp: I think callers of this method are still free to leave `version` empty, in which case we would write it to the DB w/ `version = "latest"`.

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
rawlinp: So I think we will need to delete two rows possibly. If the given version is `""`, it is assumed to be `"latest"`. So we will always need to delete the key w/ version `"latest"` first. If version is _not empty_, then we'll also have to delete the key (if any) at _that given version_.

Then, we will `INSERT` the key with version `"latest"` _and_ at the given version (if non-empty).

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
rawlinp: Callers of this method are free to leave `version` empty, but it should then be set to `"latest"` by default in this method.

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
rawlinp: Yes, this query should only select keys `WHERE cdn=$1 AND version = 'latest'`

traffic_ops/v4-client/deliveryservice.go
```diff
@@ -289,6 +292,18 @@ func (to *Session) GenerateSSLKeysForDS(XMLID string, CDNName string, sslFields
 	return response.Response, reqInf, nil
 }
 
+// AddSSLKeysForDS adds SSL Keys for the given DS
+func (to *Session) AddSSLKeysForDS(request tc.DeliveryServiceAddSSLKeysReq) (string, toclientlib.ReqInf, error) {
```
rawlinp, line 296: I think @ocket8888 may have to come in and clean this up once this is merged, but it might help him if this returned something like
```
struct {
    Response string `json:"response"`
    Alerts
}
```
instead of just a string. That way, alerts (if any) are actually unmarshalled and checkable. I think there is already a type like that that could be aliased if need be.

rawlinp (changes requested):

traffic_ops/v4-client/deliveryservice.go
ocket8888: I would appreciate that, because if you don't do it I most certainly will

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
srijeet0406: So the `version` cannot be empty in this request, the validators make sure of that before passing the request body into this method. But I've made the change so that the methos actually insert two rows, one with the specified version, and one with `latest`, after deleting the `latest` and the key with the given version (if provided)

lib/go-tc/deliveryservices.go
```diff
@@ -30,6 +30,13 @@ const DefaultMaxRequestHeaderBytes = 0
 const MinRangeSliceBlockSize = 262144   // 265Kib
 const MaxRangeSliceBlockSize = 33554432 // 32Mib
 
+// SSLKeysAddResponse is a struct to store the response of addition of ssl keys for a DS,
+// along with any alert messages
+type SSLKeysAddResponse struct {
```
rawlinp, line 35: This should probably go in `deliveryservice_ssl_keys.go` instead and it looks like it can just use this type alias:
```
type SSLKeysAddResponse DeliveryServiceSSLKeysGenerationResponse
```
since `DeliveryServiceSSLKeysGenerationResponse` is the same.

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -140,18 +140,31 @@ func (p *Postgres) PutDeliveryServiceSSLKeys(key tc.DeliveryServiceSSLKeys, tx *
...
+	if err != nil {
+		e := checkErrWithContext("Traffic Vault PostgreSQL: executing INSERT SSL Key query", err, ctx.Err())
+		return e
+	}
+	if rowsAffected, err := res.RowsAffected(); err != nil {
+		return err
+	} else if rowsAffected == 0 {
+		return errors.New("SSL Key: no keys were inserted")
+	}
+	res, err = tvTx.Exec("INSERT INTO sslkey (deliveryservice, data, cdn, version) VALUES ($1, $2, $3, $4)", key.DeliveryService, keyJSON, key.CDN, latestVersion)
```
rawlinp, line 167: we should combine these `INSERT`s into a single query

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -140,18 +140,31 @@ func (p *Postgres) PutDeliveryServiceSSLKeys(key tc.DeliveryServiceSSLKeys, tx *
...
-	_, err = tvTx.Exec("DELETE FROM sslkey WHERE deliveryservice=$1", key.DeliveryService)
+	_, err = tvTx.Exec("DELETE FROM sslkey WHERE deliveryservice=$1 and version=$2", key.DeliveryService, strconv.FormatInt(int64(key.Version), 10))
 	if err != nil && err != sql.ErrNoRows {
 		return err
 	}
-	res, err := tvTx.Exec("INSERT INTO sslkey (deliveryservice, data, cdn, version) VALUES ($1, $2, $3, $4)", key.DeliveryService, keyJSON, key.CDN, key.Version)
+	_, err = tvTx.Exec("DELETE FROM sslkey WHERE deliveryservice=$1 and version=$2", key.DeliveryService, latestVersion)
+	if err != nil && err != sql.ErrNoRows {
+		return err
+	}
```
rawlinp: we should combine these `DELETE`s into a single query

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -100,24 +103,147 @@ func (p *Postgres) commitTransaction(tx *sqlx.Tx, ctx context.Context, cancelFun
...
+	keyJSON, err := json.Marshal(&key)
+	if err != nil {
+		return errors.New("marshalling keys: " + err.Error())
+	}
+
+	// delete the old ssl keys first
+	oldVersions := []string{strconv.FormatInt(int64(key.Version), 10), latestVersion}
+	_, err = tvTx.Exec("DELETE FROM sslkey WHERE deliveryservice=$1 and version=ANY($2)", key.DeliveryService, pq.Array(oldVersions))
+	if err != nil && err != sql.ErrNoRows {
+		return err
```
rawlinp, line 152: before returning we should use `checkErrWithContext` here in case the error is due to context cancellation

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -100,24 +103,147 @@ func (p *Postgres) commitTransaction(tx *sqlx.Tx, ctx context.Context, cancelFun
...
+	rows, err := tvTx.Query("SELECT data from sslkey WHERE cdn=$1 AND version=$2", cdnName, latestVersion)
+	if err != nil {
+		e := checkErrWithContext("Traffic Vault PostgreSQL: executing GET SSL Keys for CDN query", err, ctx.Err())
+		return keys, e
+	}
+	defer rows.Close()
+	for rows.Next() {
+		jsonKey := ""
+		if err := rows.Scan(&jsonKey); err != nil {
+			return keys, errors.New("scanning cdn ssl keys: " + err.Error())
```
rawlinp, line 237: before returning we should use `checkErrWithContext` here in case the error is due to context cancellation

rawlinp (commented):
This is looking great, just a couple places where I think we're missing calls to `checkErrWithContext`.

rawlinp (changes requested):
Everything looks good, but after running the `TestDeliveryServices` test there is an sslkey still left in the TV DB:
```
{"cdn": "sslkeytransfer", "key": "dsID", "version": 1, "authType": "Self Signed", "hostname": "*.test2.com",
```
Looking at your test code, it looks like all the certs should be getting cleaned up, but there are a couple certs created using `"dsID"` as the key. So I'm not sure which one isn't getting cleaned up.

PR #5828, opened 2021-05-10
Add TO Client API for Profiles Automation

dmohan001c:

- CDN in a Box
- Traffic Control Client 
- Traffic Ops
- CI tests

Execute all the Integration tests and make sure the tests passed.


traffic_ops/testing/api/v4/profiles_test.go
```diff
@@ -238,9 +239,22 @@ func UpdateTestProfiles(t *testing.T) {
 	if err != nil {
 		t.Errorf("cannot GET Profile by name: %v - %v", firstProfile.Name, err)
 	}
+	cdns, _, err := TOSession.GetCDNByName("cdn2", nil)
 	remoteProfile := resp[0]
+	oldName := remoteProfile.Name
+
 	expectedProfileDesc := "UPDATED"
+	expectedCDNId := &cdns[0].ID
```
ocket8888, line 247: This line will segfault if Traffic Ops returns an improper response

traffic_ops/testing/api/v4/profiles_test.go
```diff
@@ -238,9 +239,22 @@ func UpdateTestProfiles(t *testing.T) {
...
+	expectedName := "testing"
+	expectedRoutingDisabled := true
+	expectedType := "TR_PROFILE"
+
 	remoteProfile.Description = expectedProfileDesc
+	remoteProfile.Type = expectedType
+	remoteProfile.CDNID = *expectedCDNId
```
ocket8888, line 254: It looks like you don't need to bother making a pointer to this, as you only use the dereferenced value, not the reference itself.

traffic_ops/testing/api/v4/profiles_test.go
```diff
@@ -256,7 +270,23 @@ func UpdateTestProfiles(t *testing.T) {
 	if respProfile.Description != expectedProfileDesc {
 		t.Errorf("results do not match actual: %s, expected: %s", respProfile.Description, expectedProfileDesc)
 	}
-
+	if respProfile.Type != expectedType {
+		t.Errorf("results do not match actual: %s, expected: %s", respProfile.Type, expectedType)
+	}
+	if respProfile.CDNID != *expectedCDNId {
+		t.Errorf("results do not match actual: %d, expected: %d", respProfile.CDNID, expectedCDNId)
```
ocket8888, line 277: This will print the pointer `expectedCDNId`, not the value to which it points. But if you just don't use a pointer, since you don't need to, this problem will go away.

traffic_ops/v4-client/profile.go
```diff
@@ -77,9 +77,13 @@ func (to *Session) GetParametersByProfileName(profileName string, header http.He
 }
 
 // GetProfiles returns all Profiles stored in Traffic Ops.
-func (to *Session) GetProfiles(header http.Header) ([]tc.Profile, toclientlib.ReqInf, error) {
+func (to *Session) GetProfiles(params url.Values, header http.Header) ([]tc.Profile, toclientlib.ReqInf, error) {
+	uri := APIProfiles
+	if params != nil {
```
ocket8888, line 82: Instead of checking for `nil`, we should just check that the length is greater than 1, because right now if you call it like `session.GetProfiles(url.Values{}, nil)` it'll use the path `profiles?`. That shouldn't pose a functional problem, but could potentially be confusing in log messages - people might think the actual query string wasn't printed properly, or mistake the <kbd>?</kbd> for punctuation.

traffic_ops/testing/api/v4/profiles_test.go
dmohan001c: Fixed it.

traffic_ops/v4-client/profile.go
dmohan001c: Fixed it.

ocket8888 (approved):

PR #5834, opened 2021-05-11
CDN Locks Blueprint

srijeet0406:

This is the blueprint to propose CDN locks as a safety mechanism for the CDN's snap and queue operations.

- Documentation

Go through the blueprint and make sure it looks correct .

- master

- [x] Tests are unnecessary, this is just a blueprint PR
- [x] This PR doe not include an update to CHANGELOG.md 

### DIFF
--- a/blueprints/to-locks.md
+++ b/blueprints/to-locks.md
@@ -20,11 +20,18 @@ under the License.
 
 ## Problem Description
 
-Currently, there is no way to guarantee that only your changes will make their way into a snapshot or queue updates 
-for the servers of a CDN. If somebody else makes changes while you're in the middle of modifying your CDN components
-or properties, their changes will "dirty" your snapshot and change the integrity of the data that you'd expect in your 
-snapshot. This could result in data mismatches and inconsistencies.
+1. Currently, there is no way to guarantee that only your changes will make their way into a snapshot of a CDN. 
+If somebody else makes changes while you're in the middle of modifying your CDN components or properties, 
+their changes will "dirty" your snapshot and change the integrity of the data that you'd expect in your 
+snapshot.
 
+2. There is no way to prevent another user from performing a snapshot and prematurely propagating the (incomplete) changes 
+   to Traffic Router and Traffic Monitor. 
+
+3. There is no way to prevent other users from queueing updates on servers, and prematurely propagating the (incomplete) changes 
+   to the Traffic Servers.
+
+All the above three cases could result in data mismatches and inconsistencies.
 CDN locks will serve as a way to avoid data corruption in snapshots and queue update activities by ensuring
 that only the intended user(s) is able to pass their changes to the snapshot/ queue updates. By placing locks 
 on a CDN, a user can essentially block out other users from modifying data that would dirty their view of the 
@@ -84,6 +91,11 @@ The following endpoints will need to handle the locks logic:
 - `/servers/{{hostname}}/queue_update`
 - `/snapshot`
 - `/topologies/{{name}}/queue_update`
+
+In addition to these, all `PUT`, `POST`, `DELETE` endpoints that scope to a CDN will have to add in the logic to check if there 
+is a "hard" lock by some other user, before a user can modify data.
+Basically, every endpoint that affects data related to a CDN directly (for example, `profiles`), and every endpoint that affects data 
+related to a CDN indirectly (for example, `parameters`, `cachegroups`) will need to be updated to perform the above mentioned check.
 
 #### REST API Impact
 
@@ -246,8 +258,9 @@ in the database before performing a snap or queue operation.
 We do not anticipate any impact on security due to the introduction of `cdn_locks`.
 
 ### Upgrade Impact
-There will be two database migrations, one to create the `cdn_locks` table, and the other to add
-the new capability. This does not depend on any existing data, so nothing should ideally cause this migration to fail.
+There will be one database migrations, to create the `cdn_locks` table. 
+The new capability can just be added to the `seeds.sql` file. 
+This does not depend on any existing data, so nothing should ideally cause this migration to fail.
 
 ### Operations Impact
 Operations will have to learn how to create, delete and work with CDN locks, if they wish to make use of this feature.
--- a/blueprints/to-locks.md
+++ b/blueprints/to-locks.md
@@ -68,7 +68,7 @@ the CDN on behalf of user `A`.
     - A way for the `admin` user to be able to unlock CDNs on other users' behalf
     
 - Snapshot/ Queue Updates Page
-    - A way to display an error on a snap/ queue operation by one user, if another user has the lock on that CDN
+    - Disable the snap/ queue button if another user has the lock on that CDN.
 
 ### Traffic Ops Impact
 
@@ -78,18 +78,24 @@ the CDN on behalf of user `A`.
 
 Traffic Ops will need to add the logic to check for locks before snapping/ queueing a CDN. It'll also need to account for
 `shared` vs `exclusive` locks, and forbid a user from snapping/ queueing a CDN if another user possesses a lock on that CDN.
+The following endpoints will need to handle the locks logic:
+- `/cachegroups/{{ID}}/queue_update`
+- `/cdns/{{ID}}/queue_update`
+- `/servers/{{hostname}}/queue_update`
+- `/snapshot`
+- `/topologies/{{name}}/queue_update`
 
 #### REST API Impact
 
 The following is the JSON representation of a `CDN_Lock` object:
 ```JSON
 {
-  "userName": "foo",
-  "cdnName": "cdn1",
+  "username": "foo",
+  "cdn": "cdn1",
   "message": "snapping cdn",
   "shared": false,
   "creator": true,
-  "lastUpdated": "2021-05-10 16:03:34-06"
+  "lastUpdated": "2021-05-10T16:03:34-06:00"
 }
 ```
 
@@ -97,22 +103,23 @@ The following table describes the top level `CDN_Lock` object:
 
 | field       | type                        | optionality | description                                                                             |
 | ----------- | --------------------------- | ----------- | ----------------------------------------------------------------------------------------|
-| userName    | string                      | required    | the user name of the user that wants to acquire a lock on the CDN                       |
-| cdnName     | string                      | required    | the name of the CDN on which the lcok needs to be acquired                              |
+| username    | string                      | required    | the user name of the user that wants to acquire a lock on the CDN                       |
+| cdn         | string                      | required    | the name of the CDN on which the lcok needs to be acquired                              |
 | message     | string                      | optional    | the message stating a reason behind locking the CDN                                     |
 | shared      | boolean                     | required    | whether or not this is a shared lock                                                    |
-| creator     | boolean                     | optional    | whether or not the requesting `userName` is the first one to acquire a lock on `cdnName`| 
-| lastUpdated | time                        | optional    | the last time this lock was updated                                                     |
+| creator     | boolean                     | required    | whether or not the requesting `username` is the first one to acquire a lock on `cdn`| 
+| lastUpdated | time                        | required    | the last time this lock was updated                                                     |
 
 **API constraints:**
-- a `userName` and `cdnName` combination must be unique id `shared` is set to `false`. In other words, there can be only one `exclusive`
+- a `username` and `cdn` combination must be unique if `shared` is set to `false`. In other words, there can be only one `exclusive`
 lock for a CDN
 - a CDN can have multiple `shared` locks
-- the `creator` will be set to `true` if the user is the first one acquiring a lock for the specified `cdnName`
-- a user can snap/ queue a CDN only if the `creator` field corresponding to the `userName` and `cdnName` combination is `true`
+- the `creator` will be set to `true` if the user is the first one acquiring a lock for the specified `cdn`
+- a user can snap/ queue a CDN only if the `creator` field corresponding to the `username` and `cdn` combination is `true`
 - a user can delete their `exclusive` lock whenever, since the user is the creator in this case
 - a user who is the creator cannot delete their `shared` lock until all other non-creator users have releases their shared locks for that CDN
-- an `admin` user can delete the locks of other users on any CDN
+- a user with at least `all-read` and `all-write` capabilities can delete the locks of other users on any CDN. The `all-write` capability will be modified
+to include a new capability `delete-all-locks`.
 
 Three new endpoints will be added for `GET`, `POST` and `DELETE` functionality with respect to CDN locks.
 ##### GET `cdn_locks`
@@ -122,20 +129,20 @@ response JSON:
 {
   "response": [
     {
-      "userName": "foo",
-      "cdnName": "cdn1",
+      "username": "foo",
+      "cdn": "cdn1",
       "message": "snapping cdn",
       "shared": true,
       "creator": true,
-      "lastUpdated": "2021-05-10 16:03:34-06"
+      "lastUpdated": "2021-05-10T16:03:34-06:00"
     },
     {
-      "userName": "bar",
-      "cdnName": "cdn2",
+      "username": "bar",
+      "cdn": "cdn2",
       "message": "queue cdn",
       "shared": false,
       "creator": true,
-      "lastUpdated": "2021-05-10 17:04:34-06"
+      "lastUpdated": "2021-05-10T17:04:34-06:00"
     }
   ]
 }
@@ -146,7 +153,7 @@ response JSON:
 request JSON:
 ```JSON
 {
-  "cdnName": "bar",
+  "cdn": "bar",
   "message": "snapping cdn",
   "shared": false
 }
@@ -162,12 +169,12 @@ response JSON:
     }
   ],
   "response": {
-    "userName": "foo",
-    "cdnName": "bar",
+    "username": "foo",
+    "cdn": "bar",
     "message": "snapping cdn",
     "shared": false,
     "creator": true,
-    "lastUpdated": "2021-05-10 17:05:30-06"
+    "lastUpdated": "2021-05-10T17:05:30-06:00"
   }
 }
 ```
@@ -182,7 +189,15 @@ response JSON:
       "text": "Cdn lock deleted",
       "level": "success"
     }
-  ]
+  ],
+  "response": {
+    "username": "foo",
+    "cdn": "bar",
+    "message": "snapping cdn",
+    "shared": false,
+    "creator": true,
+    "lastUpdated": "2021-05-10T17:05:30-06:00"
+  }
 }
 ```
 
@@ -191,16 +206,6 @@ response JSON:
 New Go client methods will be added for the `/cdn_locks` endpoints in order to write TO API tests for the new endpoints.
 
 #### Data Model / Database Impact
-<!--
-*How* will this impact the Traffic Ops data model?
-*How* will this impact the Traffic Ops database schema?
-
-What changes to the lib/go-tc structs will be required?
-What new tables and columns will be required?
-How will existing tables and columns be changed?
-What are the column data types and modifiers?
-What are the FK references and constraints?
--->
 A new database table for `cdn_lock`, as described below, will be created.
 ```text
             Table "traffic_ops.cdn_lock"
@@ -214,12 +219,14 @@ A new database table for `cdn_lock`, as described below, will be created.
  creator       | boolean                  |           | not null | false
  last_updated  | timestamp with time zone |           | not null | now() 
 Indexes:
-    "pk_cdn_lock" PRIMARY KEY(id)
+    "pk_cdn_lock" PRIMARY KEY(username, cdn)
 Foreign-key constraints:
-    "fk_lock_cdn" FOREIGN KEY (cdn_name) REFERENCES cdn(name)
-    "fk_lock_username" FOREIGN KEY ("username") REFERENCES tm_user(username)
+    "fk_lock_cdn" FOREIGN KEY (cdn) REFERENCES cdn(name)
+    "fk_lock_username" FOREIGN KEY (username) REFERENCES tm_user(username)
 ```
 
+The `capability` table will need to add another capability by the name of `delete-all-locks` and the `all-write` capability 
+will include this new capability. 
 ### ORT Impact
 No impact
 
@@ -250,8 +257,8 @@ in the database before performing a snap or queue operation.
 We do not anticipate any impact on security due to the introduction of `cdn_locks`.
 
 ### Upgrade Impact
-There will be one database migration, to create the `cdn_locks` table. This does not depend on any
-existing data, so nothing should ideally cause this migration to fail.
+There will be two database migrations, one to create the `cdn_locks` table, and the other to add
+the new capability. This does not depend on any existing data, so nothing should ideally cause this migration to fail.
 
 ### Operations Impact
 Operations will have to learn how to create, delete and work with CDN locks, if they wish to make use of this feature.
--- a/blueprints/to-locks.md
+++ b/blueprints/to-locks.md
@@ -69,7 +69,7 @@ the CDN on behalf of user `A`.
     - New "lock" icon to give the user the ability to lock a particular CDN
     - A dropdown list that appears when you click the above mentioned lock icon, which lists the list of CDNs that you can lock
     - A `Message` field that appears under the dropdown list that the user can populate with a custom message stating the reason behind locking the CDN
-    - A `Soft Lock` field that can be set to `true` or `false` based on user requirement
+    - A set of radio buttons to show the type of lock that the user wants, that is, `Soft Lock` and `Hard Lock`. A tooltip explaining what each lock denotes will also be provided.
     - A CDN notification displaying which CDN is locked by which user
     - An "unlock" option that appears next to the notification, only for the user who has locked the CDN
     - A way for the `admin` user to be able to unlock CDNs on other users' behalf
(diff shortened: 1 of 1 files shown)


blueprints/to-locks.md
```diff
@@ -0,0 +1,272 @@
...
+
+We propose a locking mechanism by which you can guarantee the exclusivity of your changes.
+With CDN locks, you can ensure one or both of the following conditions:
+- Only you can make changes and snap/ queue your CDN (exclusive locks)
+- Anybody can make changes but only you can snap/ queue your CDN (shared locks)
+
+`Exclusive` locks will help in cases where you are not collaborating with other users on the changes that you are
+making to a CDN. We anticipate all automated scripts and cron jobs to make use of these exclusive locks.
+
+The second category of locks is `shared` locks. This will be useful in a scenario when a group of users is working
```
mitchell852, line 46: with a shared lock, one person owns it (and can snap/queue the cdn) but many others can make changes right? do we need to specify who these others are? and if so, how?

blueprints/to-locks.md
```diff
@@ -0,0 +1,272 @@
...
+#### REST API Impact
+
+The following is the JSON representation of a `CDN_Lock` object:
+```JSON
+{
+  "userName": "foo",
+  "cdnName": "cdn1",
+  "message": "snapping cdn",
+  "shared": false,
+  "creator": true,
```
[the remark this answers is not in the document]
srijeet0406: Yeah, so the `creator` field is for the following case:
Say user A acquires a shared lock on a CDN. This creates a shared lock in the database with the creator set to `user A`
After that, user B also acquires a shared lock on the same CDN. This time the creator does not get set to `user B`, but is set as `user A`, since `A` was the first one to create this shared lock.
Now when user B tries to snap the CDN, the system should reject the request saying that since the "creator" (or first creator) of the shared lock was user A, only A can snap/ queue the CDN. Does that make sense?

blueprints/to-locks.md
```diff
@@ -0,0 +1,272 @@
...
+    - A `Message` field that appears under the dropdown list that the user can populate with a custom message stating the reason behind locking the CDN
+    - A `Shared` field that can be set to `true` or `false` based on user requirement
+    - A CDN notification displaying which CDN is locked by which user
+    - An "unlock" option that appears next to the notification, only for the user who has locked the CDN
+    - A way for the `admin` user to be able to unlock CDNs on other users' behalf
+    
+- Snapshot/ Queue Updates Page
+    - A way to display an error on a snap/ queue operation by one user, if another user has the lock on that CDN
+
+### Traffic Ops Impact
```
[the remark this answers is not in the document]
srijeet0406: Yeah I mentioned that the snap/ queue endpoints will have to add logic to handle the locks, but I will explicitly mention the endpoints to make it clearer.

ocket8888:
okay. It seems I've fundamentally misunderstood the nature of a lock. It doesn't actually prevent anything other than snap/queue whether it's shared or not, correct? I think I get it now.

srijeet0406:
> okay. It seems I've fundamentally misunderstood the nature of a lock. It doesn't actually prevent anything other than snap/queue whether it's shared or not, correct? I think I get it now.

Exactly, the main goal is to prevent those users from snapping/ queueing who either do not have the lock on a CDN/ did not create the lock on a CDN.

blueprints/to-locks.md
```diff
@@ -0,0 +1,279 @@
...
+grab the shared lock and make changes on the CDN. However, only the first user who grabbed the shared lock can actually
+snap/ queue updates on the CDN. This way, the first user to grab the lock still has the chance to review the changes, and
+also ensure that nobody else snaps/ queues the CDN before they are done with their intended modifications.
+
+There is also the third option of not using locks at all, in which case, the software will behave exactly the way it does
+today, that is, no safety that your changes will not be corrupted by somebody else before you snap/ queue.
+
+For unlocking a CDN under normal circumstances, only the user who has acquired the lock can unlock it (both shared 
+and exclusive). However, in the rare case that a user `A` forgets to unlock a CDN, an `admin` role user can unlock 
+the CDN on behalf of user `A`.
```
limited, line 57: Can any user lock a CDN?

blueprints/to-locks.md
```diff
@@ -0,0 +1,279 @@
...
+| username    | string                      | required    | the user name of the user that wants to acquire a lock on the CDN                       |
+| cdn         | string                      | required    | the name of the CDN on which the lcok needs to be acquired                              |
+| message     | string                      | optional    | the message stating a reason behind locking the CDN                                     |
+| shared      | boolean                     | required    | whether or not this is a shared lock                                                    |
+| creator     | boolean                     | required    | whether or not the requesting `username` is the first one to acquire a lock on `cdn`| 
+| lastUpdated | time                        | required    | the last time this lock was updated                                                     |
+
+**API constraints:**
+- a `username` and `cdn` combination must be unique if `shared` is set to `false`. In other words, there can be only one `exclusive`
```
limited, line 114: `cdn` by itself must be unique if the lock is `exclusive`, right?
(`user1`, `cdn`) and (`user2`, `cdn`) should not be allowed for an exclusive lock

For a shared lock, (`username`, `cdn`) must be unique.

blueprints/to-locks.md
```diff
@@ -0,0 +1,279 @@
...
+lock for a CDN
+- a CDN can have multiple `shared` locks
+- the `creator` will be set to `true` if the user is the first one acquiring a lock for the specified `cdn`
```
limited, line 117: `owner` of the lock might be a more common term

blueprints/to-locks.md
```diff
@@ -0,0 +1,279 @@
...
+- a user can snap/ queue a CDN only if the `creator` field corresponding to the `username` and `cdn` combination is `true`
```
limited, line 118: a user can snap/queue a CDN only if their username matches the `creator` username?

blueprints/to-locks.md
```diff
@@ -0,0 +1,279 @@
...
+- a user can delete their `exclusive` lock whenever, since the user is the creator in this case
+- a user who is the creator cannot delete their `shared` lock until all other non-creator users have releases their shared locks for that CDN
```
limited, line 120: what use cases/concerns is this addressing?

blueprints/to-locks.md
```diff
@@ -0,0 +1,279 @@
...
+    "lastUpdated": "2021-05-10T17:05:30-06:00"
+  }
+}
+```
+
+#### Client Impact
+
+New Go client methods will be added for the `/cdn_locks` endpoints in order to write TO API tests for the new endpoints.
+
+#### Data Model / Database Impact
```
limited, line 208: I didn't see any discussion of the actual implementation, so I'm curious. 

How will the lock be implemented? Will you be using Postgres Advisory locks? If not, are there particular atomic operations you'll make use of to avoid race conditions in acquiring, evaluating, and releasing locks?

srijeet0406: So the lock object is just another database table. So when a snap/ queue request comes in, we just do a look up in the `cdn_lock` table to see if the current user has the lock, and if so, proceed with the operation. In the current scenarion, this is just like checking for the `If-Modified-Since` or `If-Unmodified-Since` header.

blueprints/to-locks.md
srijeet0406: yes, you're correct. Because in exclusive locks, no one else has the lock, and in shared locks, only the creator/ owner of the lock is allowed to snap/ queue the cdn.

blueprints/to-locks.md
srijeet0406: I'll make that change.

blueprints/to-locks.md
srijeet0406: Yes, for an exclusive lock, `cdn` must be unique.
For a shared lock, (`username`, `cdn`) must be unique.

blueprints/to-locks.md
srijeet0406: Any user who is capable of snapping or queueing the CDN can lock it.

blueprints/to-locks.md
limited: What prevents two users from obtaining a lock on the same resource?

srijeet0406: When a user sends a request to snap/ queue a CDN, the backend code for API endpoint does a database check to see if the current user has a lock on the CDN. If not, then a `403 Forbidden` is returned to the user.
The API uniqueness constraint (of `cdn` in case of an exclusive lock and `username`+`cdn` in case of a shared lock) will prevent two users obtaining the lock on the same resource. We also suggest the clients make use of the `If-Unmodified-Since` header to make sure the locks data doesn't change on them before they complete their request.

limited: Is either of these examples possible?

```
User1 API request received to acquire exclusive lock. API checks database, no existing lock
User2 API request received to acquire exclusive lock. API checks database, no existing lock
User1 writes lock to DB. Returns lock via API
User2 writes lock to DB. Returns lock via API
```

```
(User1, cdn) lock in db
User1 sends API request to snap. API checks database, lock exists
Before snap completes, user1 releases lock, removing entry from db table
User2 modifies CDN
User1 snap resumes and includes User 2 modifications (bad?)
```

blueprints/to-locks.md
mitchell852: yeah, creating/maintaining a list of shared lock users would be unmanageable imo. so it seems there are just 2 types of locks:

exclusive lock: only the owner of the lock can snap/queue OR make cdn changes (that affect the snap/queue)
shared lock: only the owner of the lock can snap/queue but ANYBODY can make cdn changes (that affect the snap/queue)

blueprints/to-locks.md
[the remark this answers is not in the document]
mitchell852: lock.creator still feels redundant with lock.username imo. if lock.username != me, i know that i'm not the "creator" :)

mitchell852:
> Well I was thinking that you could just do
> 
> 1. User `A` grabs a shared lock on CDN `foo`.
> 2. Users `B` and `C` can still make changes, because the lock is "shared", but cannot snap or queue
> 3. User `A` deletes the shared lock. Now anyone can snap or queue

this is how i understood a "shared lock" to work.

blueprints/to-locks.md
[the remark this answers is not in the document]
srijeet0406: I'm going to change it to `owner`. The username will be what the user of the current lock request is. If we dont have an `owner` field, we wont know who created the lock in the first place. This is information that we'll need in case of shared locks.

srijeet0406:
> > Well I was thinking that you could just do
> > 
> > 1. User `A` grabs a shared lock on CDN `foo`.
> > 2. Users `B` and `C` can still make changes, because the lock is "shared", but cannot snap or queue (only user A can)
> > 3. User `A` deletes the shared lock. Now anyone can snap or queue
> 
> ^^ this is how i understood a "shared lock" to work and i understood an "exclusive lock" to work like:
> 
> 1. User `A` grabs an exclusive lock on CDN `foo`.
> 2. Users `B` and `C` CANNOT make CDN changes or  snap/queue because the lock is "exclusive" to User `A`
> 3. User `A` deletes the exclusive lock. Now anyone can make changes/snap/queue
> 
> Only one person ever "owns" a lock (shared or exclusive) at a time.

Right, thats the case for exclusive locks. But in case of shared locks, if the person who had originally created the shared lock deletes the lock, who has the right to snap/ queue the CDN? If we leave it open to anyone, then we are back to square one IMO.

blueprints/to-locks.md
srijeet0406: In the first scenario, if user 2 uses `If-UnModified-Since` in their request (which we can make a requirement in the backend).
For the second scenario, we can add a mutex or some kind of sync mechanism to make sure that a lock cannot be released by the "owner" when a snap is in progress. Frankly, I think thats a very rare edge case, because the snap usually takes a very small amount of time, and you would really have to try hard to release the lock between the time you press the `snap` button and the snap completes.
Note that when `user1` has a lock on `cdn` (assuming `user1` is the owner of the lock), only `user1` can snap/ queue, so it won't really matter if another user that has the shared lock on `cdn` releases the lock during this time or not. Hope that makes sense.

mitchell852:
> if the person who had originally created the shared lock deletes the lock, who has the right to snap/ queue the CDN? If we leave it open to anyone, then we are back to square one IMO.

yep, if no lock (shared or exclusive) then anybody can do anything which i think is fine.

srijeet0406:
> > if the person who had originally created the shared lock deletes the lock, who has the right to snap/ queue the CDN? If we leave it open to anyone, then we are back to square one IMO.
> 
> yep, if no lock (shared or exclusive) then anybody can do anything which i think is fine. if you don't like that then don't delete the lock :) you don't pass locks from one person to another. you just take them when _you_ want one and delete it when _you_ don't need it anymore.

I think when you open a shared lock, you are agreeing to the fact that others might use it, and if you created the lock, you take on the responsibility of not corrupting the snapshot. If we were to follow what you're saying, then there will be a period when a shared lock doesn't have an owner, and in that time frame, anyone who has access to the shared lock can queue/ snap, inspite of not being the owner. Also, this would create problems in the following case:

user `A` gets the shared lock on `cdn1`
users `B` and `C` get the shared lock on `cdn1`
at this point, user `A` releases the lock, so now `B` and `C`. an snap/ queue `cdn1`(which opens us up to the same issue again)
user `D` comes in and gets the shared lock on `cdn1`, making `D` the `owner` of the lock
now again, `B` and `C` are locked out of snapping/ queueing `cdn1`

Form `B` and `C`'s perspective, the system will let them allow snapping/ queueing once, followed by denying them, even though they haven't changed anything from their end. I think this might create confusion for the users.

mitchell852:
> user `A` gets the shared lock on `cdn1`
> users `B` and `C` get the shared lock on `cdn1`
> at this point, user `A` releases the lock, so now `B` and `C`. an snap/ queue `cdn1`(which opens us up to the same issue again)
> user `D` comes in and gets the shared lock on `cdn1`, making `D` the `owner` of the lock
> now again, `B` and `C` are locked out of snapping/ queueing `cdn1`
> 
> Form `B` and `C`'s perspective, the system will let them allow snapping/ queueing once, followed by denying them, even though they haven't changed anything from their end. I think this might create confusion for the users.

i think you've overcomplicated it. i think this is all that is needed:

```
exclusive lock: only owner can make cdn changes/snap/queue
shared lock: only owner can snap/queue; owner and non-owners can make cdn changes
no lock: anyone can make cdn changes/snap/queue
```

the user can decide what level of protection they are looking for.

srijeet0406:
> > user `A` gets the shared lock on `cdn1`
> > users `B` and `C` get the shared lock on `cdn1`
> > at this point, user `A` releases the lock, so now `B` and `C`. an snap/ queue `cdn1`(which opens us up to the same issue again)
> > user `D` comes in and gets the shared lock on `cdn1`, making `D` the `owner` of the lock
> > now again, `B` and `C` are locked out of snapping/ queueing `cdn1`
> > Form `B` and `C`'s perspective, the system will let them allow snapping/ queueing once, followed by denying them, even though they haven't changed anything from their end. I think this might create confusion for the users.
> 
> i think you've overcomplicated it. i think this is all that is needed:
> 
> ```
> exclusive lock: only owner can make cdn changes/snap/queue
> shared lock: only owner can snap/queue; owner and non-owners can make cdn changes
> no lock: anyone can make cdn changes/snap/queue
> ```
> 
> the user can decide what level of protection they are looking for. i'm pretty sure this is what @rawlinp was thinking when he proposed the shared lock.

Then I suggest deleteing all non-owner shared locks for a CDN when the owner deletes their shared lock, because its as good as not having a lock at that point. We could use the notifications feature to show this in TP.

rawlinp:
> i think you've overcomplicated it. i think this is all that is needed:
> ```
> exclusive lock: only owner can make cdn changes/snap/queue
> shared lock: only owner can snap/queue; owner and non-owners can make cdn changes
> no lock: anyone can make cdn changes/snap/queue
> ```
> the user can decide what level of protection they are looking for. i'm pretty sure this is what @rawlinp was thinking when he proposed the shared lock.

Yeah, that is what I was thinking. I thought there would only be a single exclusive lock and a single shared lock. Each lock can only have one owner at a time, and only one type of lock can be held at a time.

I think it helps to start with the use cases behind each type of lock. For example, if I wanted to update a DS certificate, I could grab the exclusive lock since I don't want anyone else's changes to be propagated until I've had a chance to test the new cert on a single cache. Once I've verified the cert, I will queue the entire CDN then release the exclusive lock.

As another example, say there are a lot of certs about to expire and we want to divide the work among multiple operators. One operator would grab the shared lock (preventing anyone else from queuing updates). The other operators are still free to update their certs anytime (since only a shared lock is held), and when ready, the lock holder will queue updates on single cache. Then all the operators can test their certs before the lock holder queues the entire CDN and releases the shared lock.

We shouldn't make this solution overly complicated, since I see this more as a temporary solution until we have the time to tackle the full solution (delivery service snapshots and server snapshots). Right?

srijeet0406:
> > i think you've overcomplicated it. i think this is all that is needed:
> > ```
> > exclusive lock: only owner can make cdn changes/snap/queue
> > shared lock: only owner can snap/queue; owner and non-owners can make cdn changes
> > no lock: anyone can make cdn changes/snap/queue
> > ```
> > 
> > 
> >   
> >     
> >     
> > 
> >     
> >     
> > 
> >   
> > 
> > 
> > the user can decide what level of protection they are looking for. i'm pretty sure this is what @rawlinp was thinking when he proposed the shared lock.
> 
> Yeah, that is what I was thinking. I thought there would only be a single exclusive lock and a single shared lock (per-CDN). Each lock can only have one owner at a time, and only one type of lock can be held at a time (per-CDN).
> 
> I think it helps to start with the use cases behind each type of lock. For example, if I wanted to update a DS certificate, I could grab the exclusive lock since I don't want anyone else's changes to be propagated until I've had a chance to test the new cert on a single cache. Once I've verified the cert, I will queue the entire CDN then release the exclusive lock.
> 
> As another example, say there are a lot of certs about to expire and we want to divide the work among multiple operators. One operator would grab the shared lock (preventing anyone else from queuing updates). The other operators are still free to update their certs anytime (since only a shared lock is held), and when ready, the lock holder will queue updates on single cache. Then all the operators can test their certs before the lock holder queues the entire CDN and releases the shared lock.
> 
> We shouldn't make this solution overly complicated, since I see this more as a temporary solution until we have the time to tackle the full solution (delivery service snapshots and server snapshots). Right?

Alright, I think I may have misunderstood your explanation of "shared" locks earlier. This will simplify the solution a lot. I'll make the changes to the blueprint.

mitchell852:
just a thought. what if instead of naming it exclusive/shared locks (as shared is a bit confusing), we call it hard/soft locks? just spit balling or maybe that's confusing as well. :)

```
hard lock: only the owner can make cdn changes/snap/queue
soft lock: only the owner can snap/queue (others are free to make cdn changes)
no lock: anyone (with the proper permissions of course) can make cdn changes/snap/queue (current behavior)
```

srijeet0406:
> just a thought. what if instead of naming it exclusive/shared locks (as shared is a bit confusing), we call it hard/soft locks? just spit balling or maybe that's confusing as well. :)
> 
> ```
> hard lock: only the owner can make cdn changes/snap/queue
> soft lock: only the owner can snap/queue (others are free to make cdn changes)
> no lock: anyone (with the proper permissions of course) can make cdn changes/snap/queue (current behavior)
> ```

Agreed, in my head, `shared` == `a lock that can be used by multiple users`, but since thats not the case, renaming them might be a good idea.

rawlinp:
"Hard" and "soft" locks sound ok to me.

srijeet0406:
Alright, all changes should be done now.

blueprints/to-locks.md
```diff
@@ -77,42 +77,43 @@ the CDN on behalf of user `A`.
...
 
 Traffic Ops will need to add the logic to check for locks before snapping/ queueing a CDN. It'll also need to account for
-`shared` vs `exclusive` locks, and forbid a user from snapping/ queueing a CDN if another user possesses a lock on that CDN.
+`soft` vs `hard` locks, and forbid a user from snapping/ queueing a CDN if another user possesses a lock on that CDN.
+The following endpoints will need to handle the locks logic:
+- `/cachegroups/{{ID}}/queue_update`
+- `/cdns/{{ID}}/queue_update`
+- `/servers/{{hostname}}/queue_update`
+- `/snapshot`
+- `/topologies/{{name}}/queue_update`
```
ocket8888, line 86: So, let me try to make sure I understand this one last time.

If I have a Soft lock on a CDN, anyone can still make changes, but these endpoints need to be aware that only a certain user is allowed to use them.

But if I have a Hard lock, these endpoints still need to be aware of that, but also nobody else is permitted to make any changes within the scope of the CDN, right? That means that very nearly every endpoint is affected; it might be a shorter list to itemize the endpoints that *won't* need to be changed as a result of this.

Basically, every endpoint that controls data scoped to a CDN (e.g. `/profiles`), every endpoint that controls data owned by objects scoped to a CDN (e.g. `/parameters`), and every endpoint that controls data that defines interactions between objects scoped to a CDN even if the data itself is not scoped to a CDN (e.g. `/cachegroups`) will need to be updated to be aware of the locking mechanism.

blueprints/to-locks.md
```diff
@@ -0,0 +1,268 @@
...
+### Performance Impact
+We do not anticipate significant performance impact, as this process will just entail one additional check
+in the database before performing a snap or queue operation.
+
+### Security Impact
+We do not anticipate any impact on security due to the introduction of `cdn_locks`.
+
+### Upgrade Impact
+There will be two database migrations, one to create the `cdn_locks` table, and the other to add
+the new capability. This does not depend on any existing data, so nothing should ideally cause this migration to fail.
```
ocket8888, line 250: Oh, also, you don't need a migration to add the new capabilities. Those can just go in `seeds.sql`.

blueprints/to-locks.md
srijeet0406: You are correct. So basically all `PUT/POST/DELETE` endpoints that have any relation to a CDN should be locked down (to other users) if a user has a Hard lock on a particular CDN. I'll add this statement to the blueprint.

blueprints/to-locks.md
```diff
@@ -0,0 +1,268 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+# CDN Traffic Ops Locks
+
+## Problem Description
+
+Currently, there is no way to guarantee that only your changes will make their way into a snapshot or queue updates
```
mitchell852, line 23: there are really 3 problems:

1. there is no way to prevent others from "dirtying" your snapshot with their own changes (you address this)
2. there is no way to prevent others from performing a snapshot and prematurely propagating changes to TR/TM
3. there is no way to prevent others from queuing (server) updates and prematurely propagating changes to caches

i would suggest that you mention all 3.

blueprints/to-locks.md
```diff
@@ -0,0 +1,281 @@
...
+and hard). However, in the rare case that a user `A` forgets to unlock a CDN, an `admin` role user can unlock 
+the CDN on behalf of user `A`.
+
+### Traffic Portal Impact
+
+- Landing Page
+    - New "lock" icon to give the user the ability to lock a particular CDN
+    - A dropdown list that appears when you click the above mentioned lock icon, which lists the list of CDNs that you can lock
+    - A `Message` field that appears under the dropdown list that the user can populate with a custom message stating the reason behind locking the CDN
+    - A `Soft Lock` field that can be set to `true` or `false` based on user requirement
```
mitchell852, line 72: the UI probably needs 2 radio buttons. 1 for hard lock and one for soft lock (with a tooltip of what each means or something) so the user can see both options and decide which makes sense for them.

blueprints/to-locks.md
```diff
@@ -0,0 +1,281 @@
...
+    - A CDN notification displaying which CDN is locked by which user
+    - An "unlock" option that appears next to the notification, only for the user who has locked the CDN
+    - A way for the `admin` user to be able to unlock CDNs on other users' behalf
+    
+- Snapshot/ Queue Updates Page
+    - Disable the snap/ queue button if another user has the lock on that CDN.
```
mitchell852, line 78: this is a nice to have imo but it adds complexity to the UI. typically nothing is disabled in TP and the TO API simply catches actions the user is not authorized to perform and returns a 403.

blueprints/to-locks.md
```diff
@@ -0,0 +1,281 @@
...
+| message     | string                      | optional    | the message stating a reason behind locking the CDN                                                                          |
+| soft        | boolean                     | required    | whether or not this is a shared lock, meaning if a user has the lock, whether or not other users can make changes to the CDN |
+| lastUpdated | time                        | required    | the last time this lock was updated                                                                                          |
+
+**API constraints:**
+- a CDN can have only one `hard` or `soft` lock at a time
+- a user can snap/ queue a CDN only if they are the one holding the lock (of either kind) on the `cdn`. Alternatively, if no one has the lock on the `cdn`, anyone can snap/ queue/ make changes to the CDN
+- a user can delete their `hard` or `soft` lock whenever they want
+- a user with at least `all-read` and `all-write` capabilities can delete the locks of other users on any CDN. The `all-write` capability will be modified
```
mitchell852, line 127: this is fine but permissions don't work that way currently and maybe not for some time so you need to pin each action to a role.

GET locks/lock = read/only role
POST locks = operations+ role?
DELETE lock = operations+ role and only your own lock unless admin role?

blueprints/to-locks.md
ocket8888: or a checkbox (or "switch") for whether or not it's a "hard" lock - down with `select`s containing only the options "true" and "false".

blueprints/to-locks.md
ocket8888: disabling an option you know will fail is a good UX IMO - and there are precedents in TP, actually. For example, right click on a server that isn't a cache server in the servers table, notice that "Queue Updates" and "Clear Updates" are disabled

blueprints/to-locks.md
ocket8888: Yeah, sorry, I should've been more clear. It should still specify the privilege levels, since that's what we actually use today

blueprints/to-locks.md
srijeet0406: I'm not a UI expert, so I'll let you two figure this one out.

mitchell852: yes, you are correct @ocket8888. if the info is readily available, i do disable certain features in the UI for good UX reasons. but in this case a cdn does not know it's locked unless it makes an extra api call to GET cdn_locks and that's fine if we want to build in the api call to determine if the button is enabled/disabled.

mitchell852 (approved):

ocket8888 (approved):

PR #5841, opened 2021-05-12
Add sleep to make DS test pass

srijeet0406:

- Traffic Ops
- CI tests

Run all the api tests against CIAB and make sure everything passes.
After all the tests run, go into the traffic vault database and make sure there is nothing left in there.
NOTE: The `includeSystemTests` flag needs to be set to `true` to verify this behavior.

- master

- [x] This PR does not include tests, it just fixes the existing tests
- [x] No documentation is necessary
- [x] No update to CHANGELOG.md is necessary

### DIFF
--- a/traffic_ops/testing/api/v4/deliveryservices_test.go
+++ b/traffic_ops/testing/api/v4/deliveryservices_test.go
@@ -296,12 +296,11 @@ func DeleteCDNOldSSLKeys(t *testing.T) {
 	if err != nil {
 		t.Fatalf("couldn't snap CDN: %v", err)
 	}
-	time.Sleep(5 * time.Second)
 	var newCdnKeys []tc.CDNSSLKeys
 	for tries := 0; tries < 5; tries++ {
 		time.Sleep(time.Second)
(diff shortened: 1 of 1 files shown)


traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -296,6 +296,7 @@ func DeleteCDNOldSSLKeys(t *testing.T) {
 	if err != nil {
 		t.Fatalf("couldn't snap CDN: %v", err)
 	}
+	time.Sleep(5 * time.Second)
```
rawlinp, line 299: instead of a simple sleep, can we use a shorter sleep in the retry loop and check each time? When using Postgres, the update could complete much earlier than Riak, so the tests wouldn't have to wait as long

rawlinp (changes requested):

traffic_ops/testing/api/v4/deliveryservices_test.go
srijeet0406: Should be fixed now

rawlinp (approved):

PR #5843, opened 2021-05-12
Updated TV Postgres to use AES encryption for URL Sig Keys

mattjackson220:

- Documentation
- Traffic Ops
- Traffic Vault

Create a key locally and put the location into the cdn.conf field
Run TO with TV setup to use Postgres
Verify that GET, POST, PUT, DELETE all work as expected
Verify that the data in Postgres is encrypted successfully

No test updates, changelog, or license header updates needed since this is a backend change and is covered by a previous changelog entry

### DIFF
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -33,6 +33,7 @@ In order to use the PostgreSQL backend for Traffic Vault, you will need to set t
 :password:                  The password to use when connecting to the database
 :port:                      The port number that the database listens for new connections on (NOTE: the PostgreSQL default is 5432)
 :user:                      The username to use when connecting to the database
+:encryption_key_location:   The location on-disk for the key used to encrypt secrets before they are stored.
 :conn_max_lifetime_seconds: Optional. The maximum amount of time (in seconds) a connection may be reused. If negative, connections are not closed due to a connection's age. If 0 or unset, the default of 60 is used.
 :max_connections:           Optional. The maximum number of open connections to the database. Default: 0 (unlimited)
 :max_idle_connections:      Optional. The maximum number of connections in the idle connection pool. If negative, no idle connections are retained. If 0 or unset, the default of 30 is used.
@@ -57,7 +58,8 @@ Example cdn.conf snippet:
 				"conn_max_lifetime_seconds": 60,
 				"max_connections": 500,
 				"max_idle_connections": 30,
-				"query_timeout_seconds": 10
+				"query_timeout_seconds": 10,
+				"encryption_key_location": "/opt/traffic_ops/app/conf/tv.key"
 			}
 		}
 	}
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -33,7 +33,7 @@ In order to use the PostgreSQL backend for Traffic Vault, you will need to set t
 :password:                  The password to use when connecting to the database
 :port:                      The port number that the database listens for new connections on (NOTE: the PostgreSQL default is 5432)
 :user:                      The username to use when connecting to the database
-:aes_key_location:          The location on-disk for an AES key used to encrypt secrets before they are stored.
+:aes_key_location:          The location on-disk for an AES, base64 encoded key used to encrypt secrets before they are stored.
 :conn_max_lifetime_seconds: Optional. The maximum amount of time (in seconds) a connection may be reused. If negative, connections are not closed due to a connection's age. If 0 or unset, the default of 60 is used.
 :max_connections:           Optional. The maximum number of open connections to the database. Default: 0 (unlimited)
 :max_idle_connections:      Optional. The maximum number of connections in the idle connection pool. If negative, no idle connections are retained. If 0 or unset, the default of 30 is used.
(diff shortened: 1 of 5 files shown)


docs/source/admin/traffic_vault.rst
```diff
@@ -33,6 +33,7 @@ In order to use the PostgreSQL backend for Traffic Vault, you will need to set t
 :password:                  The password to use when connecting to the database
 :port:                      The port number that the database listens for new connections on (NOTE: the PostgreSQL default is 5432)
 :user:                      The username to use when connecting to the database
+:encryption_key_location:   The location on-disk for the key used to encrypt secrets before they are stored.
```
rawlinp, line 36: instead of calling this `encryption_key_location`, which might be confused w/ a public key, we should call it `aes_key_location` and note that it must be an AES key which provides both encryption and decryption

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/encrypt.go
```diff
@@ -0,0 +1,92 @@
...
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"errors"
+	"io"
+	"io/ioutil"
+)
+
+func encryptString(stringToEncrypt string, encryptionKeyLocation string) (string, error) {
+	key, err := readKeyFromFile(encryptionKeyLocation)
```
rawlinp, line 32: Thinking about fetching this key file from HashiCorp Vault, we probably wouldn't want to fetch it every time we needed to encrypt/decrypt. I think the same would apply to on-disk storage. We probably just want to read it into memory during startup (in `postgresLoad()`) and save it as a private field in the `Postgres` struct in order to reuse it.

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/encrypt.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	nonce := make([]byte, gcm.NonceSize())
+	if _, err = io.ReadFull(rand.Reader, nonce); err != nil {
+		return "", err
+	}
+
+	return string(gcm.Seal(nonce, nonce, []byte(stringToEncrypt), nil)), nil
```
rawlinp, line 52: Just for my understanding: in the [example](https://golang.org/pkg/crypto/cipher/#example_NewGCM_encrypt), `dst` is `nil`, but here it is `nonce`. I assume that's just to prepend the nonce to the resulting ciphertext?

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/url_sig_keys.go
```diff
@@ -30,26 +30,31 @@ import (
 	"github.com/jmoiron/sqlx"
 )
 
-func getURLSigKeys(xmlID string, tvTx *sqlx.Tx, ctx context.Context) (tc.URLSigKeys, bool, error) {
-	var jsonUrlKeys string
-	if err := tvTx.QueryRow("SELECT data FROM url_sig_key WHERE deliveryservice = $1", xmlID).Scan(&jsonUrlKeys); err != nil {
+func getURLSigKeys(xmlID string, tvTx *sqlx.Tx, ctx context.Context, encryptionKeyLocation string) (tc.URLSigKeys, bool, error) {
+	var encryptedUrlSigKey string
```
rawlinp, line 34: Would it make sense to have this be `[]byte` instead of a `string`? Since the encryption/decryption works w/ `[]byte` and the DB type is `bytea`?

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/encrypt.go
mattjackson220: yea just to prepend it there since we dont want to save it elsewhere

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -433,7 +435,12 @@ func postgresLoad(b json.RawMessage) (trafficvault.TrafficVault, error) {
 		log.Infoln("successfully pinged the Traffic Vault database")
 	}
 
-	return &Postgres{cfg: pgCfg, db: db}, nil
+	aesKey, err := readKeyFromFile(pgCfg.AesKeyLocation)
+	if err != nil {
+		return nil, err
+	}
+
```
rawlinp, line 442: We should call `aes.NewCipher` here with the key contents just to ensure it's a valid key. Also, I was thinking we might want to store the key data in base64 encoding rather than raw bytes. That would make it a little easier to pass around (for instance in HashiCorp Vault we might want to store it that way as well).

rawlinp (changes requested):

docs/source/admin/traffic_vault.rst
```diff
@@ -33,6 +33,7 @@ In order to use the PostgreSQL backend for Traffic Vault, you will need to set t
 :password:                  The password to use when connecting to the database
 :port:                      The port number that the database listens for new connections on (NOTE: the PostgreSQL default is 5432)
 :user:                      The username to use when connecting to the database
+:aes_key_location:          The location on-disk for an AES key used to encrypt secrets before they are stored.
```
rawlinp, line 36: here we should also state that the key must be stored in base64 encoding

rawlinp (changes requested):
this looks good, just one last docs issue. I'll pull this down and retest

rawlinp (approved):
Works great!

PR #5844, opened 2021-05-13
Only copy Transport exported fields

### LINKED ISSUE #5758: `go vet`: Assignment copies lock value
## I'm submitting a ...

-  bug report

-  Grove

## Current behavior:

`go vet -copylocks ./...` finds 1 issue:

<table><thead><th>Message</th><th>Line of code</th></thead><tbody>

<tr><td><code>assignment copies lock value to newTransport: net/http.Transport contains sync.Mutex</code></td><td>

https://github.com/apache/trafficcontrol/blob/61e423f33b8f4b61a6228c9e3d0fbac646d231f2/grove/remap/remap.go#L500
</td></tr>
</tbody></table>

## Expected behavior:

`go vet -copylocks ./...` should find 0 issues.

```shell
go vet -copylocks ./...
```

zrhoffman:

- [x] <strike>This PR fixes #5758 by copying a struct containing locks using unsafe Go.</strike>
- [x] This PR fixes #5758 only copying the exported fields of `http.Transport`.

- Grove

Run `go vet -copylocks ./...`

- master (5b9e199244)
- 5.1.1
- 5.1.2 (RC3)

- [x] Effect of struct copy is unchanged, tests are unnecessary
- [x] Behavior is unchanged, documentation is unnecessary

### DIFF
--- a/grove/remap/remap.go
+++ b/grove/remap/remap.go
@@ -30,6 +30,7 @@ import (
 	"github.com/apache/trafficcontrol/grove/plugin"
 	"github.com/apache/trafficcontrol/grove/remapdata"
 	"github.com/apache/trafficcontrol/grove/web"
+
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-rfc"
 )
--- a/grove/remap/remap.go
+++ b/grove/remap/remap.go
@@ -30,7 +30,6 @@ import (
 	"github.com/apache/trafficcontrol/grove/plugin"
 	"github.com/apache/trafficcontrol/grove/remapdata"
 	"github.com/apache/trafficcontrol/grove/web"
-
 	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-rfc"
 )
@@ -497,7 +496,8 @@ func makeTo(tosJSON []RemapRuleToJSON, rule remapdata.RemapRule, baseTransport *
 				return nil, fmt.Errorf("error parsing to %v proxy_url: %v", to.URL, toJSON.ProxyURL)
 			}
 			to.ProxyURL = proxyURL
-			newTransport := *baseTransport
+			// See b7953cc239 for the reasoning behind copying the Transport.
+			newTransport := *baseTransport.Clone()
 			if proxyURL != nil && proxyURL.Host != "" {
 				newTransport.Proxy = http.ProxyURL(proxyURL)
 			}
--- a/.github/workflows/go.vet.yml
+++ b/.github/workflows/go.vet.yml
@@ -48,8 +48,4 @@ jobs:
     - name: Install modules
       run: go mod vendor -v
     - name: go vet
-      run: |
-        skip_analyzers=(
-          -copylocks=false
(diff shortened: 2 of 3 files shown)


grove/remap/remap.go
```diff
@@ -497,7 +498,11 @@ func makeTo(tosJSON []RemapRuleToJSON, rule remapdata.RemapRule, baseTransport *
 				return nil, fmt.Errorf("error parsing to %v proxy_url: %v", to.URL, toJSON.ProxyURL)
 			}
 			to.ProxyURL = proxyURL
-			newTransport := *baseTransport
+			newTransport := http.Transport{}
+			const transportSize = unsafe.Sizeof(*baseTransport)
+			// Copy baseTransport to newTransport without failing the copylocks `go vet` check. See
+			// b7953cc239 for the reasoning behind copying the Transport.
+			copy((*[transportSize]byte)(unsafe.Pointer(&newTransport))[:], (*[transportSize]byte)(unsafe.Pointer(baseTransport))[:])
```
rob05c, line 505: I assume this is the only way to make `go vet` happy with the copy?

ocket8888: I, too, was hoping there was a way to do this without `unsafe`

zrhoffman: We could do

```go
newTransport := *baseTransport.Clone()
```
but it would skip unexported fields like locks. We could use `reflect` to copy and set the unexported fields, but that would also involve using `unsafe`.

As for silencing the check on this line, Go offers no such option. From [go/issues#17058](https://github.com/golang/go/issues/17058#issuecomment-259012764):

> I am adamant that source code not be annotated to silence vet. That will reduce the options available but that is my position.

rob05c: We don't _need_ to copy the lock, we just need to copy the other fields we set from config, and we need the lock to not be the same as the other object. Default-initialization of the lock is fine.
 
We could also manually record and set all the transport fields we care about. I don't like that option, it's error-prone, it would silently fail if someone added a new option and forgot to add it to the "copy record."

I think `Clone` will work. All the fields we're setting in config are exported, and unless some function is added to Transport to set a private field, that shouldn't cause subtle bugs in the future.

zrhoffman: Using `Clone` in da61fe67d0.

zrhoffman:
Rebased to remove the commit message about unsafe Go.

TaylorCFrey (approved):
LGTM. Verified `go vet` is satisfied. Built and ran Grove locally to ensure the transport is copied.

zrhoffman:
Rebased to resolve merge conflicts in `.github/workflows/go.vet.yml` and `.golangci.yml` now that #5854 is merged.

rob05c:
I'll try to find time to review. But whoever reviews, just running `go vet -copylocks` isn't enough, it really needs verified that running Grove, all remaps use the correct Proxy in their upstream request. I'm _pretty sure_ they will from the code, but it's a pretty critical thing, it really needs tested to be sure.

rob05c (approved):
Looks good! Code looks good, and I tested, set up an environment with Grove with multiple remaps with different parents, verified requests went to their proper parents.

PR #5845, opened 2021-05-13
Migrate cache group test from E2E to integration test 

dpham692:

- CDN in a Box
- Traffic Portal
- CI tests

In traffic_portal/test/integration directory run command below and make sure all test pass.
`npm test -- --params.baseUrl https://localhost --capabilities.shardTestFiles true`
Make sure the new cache group test in Data file is good and reasonable.

### DIFF
--- a/traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
+++ b/traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
@@ -116,15 +116,15 @@ export class CacheGroupPage extends BasePage {
         if (cachegroup.Type == "EDGE_LOC") {
             const name = cachegroup.FailoverCG + this.randomize;
             await this.txtFailoverCG.click();
-            if (await browser.isElementPresent(element(by.xpath(`//select[@name="fallbackOptions"]//option[@label="` + name + `"]`)))) {
-                await element(by.xpath(`//select[@name="fallbackOptions"]//option[@label="` + name + `"]`)).click();
+            if (await browser.isElementPresent(element(by.css(`select[name="fallbackOptions"] > option[label="${name}"]`)))) {
+                await element(by.css(`select[name="fallbackOptions"] > option[label="${name}"]`)).click();
             } else {
                 result = undefined;
             }
         }
         await this.txtType.sendKeys(cachegroup.Type);
         await snp.ClickUpdate();
-        if (result != undefined) {
+        if (result !== undefined) {
             await basePage.GetOutputMessage().then(function (value) {
                 if (outputMessage === value) {
                     result = true;
@@ -157,12 +157,7 @@ export class CacheGroupPage extends BasePage {
         return result;
     }
     public async CheckCSV(name: string): Promise<boolean> {
-        let linkName = name;
-        let result = false;
-        if (await browser.isElementPresent(element(by.xpath("//span[text()='" + linkName + "']"))) == true) {
-            result = true;
-        }
-        return result;
+        return element(by.cssContainingText("span", name)).isPresent();
     }
     public async ToggleTableColumn(name: string): Promise<boolean> {
         let result = false;
--- a/traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
+++ b/traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
@@ -160,23 +160,10 @@ export class CacheGroupPage extends BasePage {
         return element(by.cssContainingText("span", name)).isPresent();
     }
     public async ToggleTableColumn(name: string): Promise<boolean> {
-        let result = false;
         await this.btnTableColumn.click();
-        //if it's already on, turn it off and return false
-        if (await browser.isElementPresent(element(by.xpath("//th[text()='" + name + "']"))) === true){
-            await element(by.xpath("//label[text()=' " + name + "']")).click();
-            result = false;
-        }else{
-            //of it already off, turn it on return true
-            await element(by.xpath("//label[text()=' " + name + "']")).click();
-            result = true;
-        }
+        const result = await element(by.cssContainingText("th", name)).isPresent();
+        await element(by.cssContainingText("label", name)).click();
         await this.btnTableColumn.click();
-        return result;
+        return !result;
     }
-
-
-
-
-
 }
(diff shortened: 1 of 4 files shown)


traffic_portal/test/integration/specs/CacheGroup.spec.ts
```diff
@@ -41,20 +41,38 @@ cachegroups.tests.forEach(cacheGroupData => {
                 await cacheGroupPage.OpenTopologyMenu();
                 await cacheGroupPage.OpenCacheGroupsPage();
             })
+            cacheGroupData.check.forEach(check => {
+                it(check.description, async () => {
+                    expect(await cacheGroupPage.CheckCSV(check.Name)).toBeTruthy();
```
ocket8888, line 46: `CheckCSV` returns a `Promise<boolean>`, so this can check specifically for `.toBeTrue()`, not merely `.toBeTruthy()`. Or you could use `.toBe(true)`, the difference between that and `.toBeTrue()` is convoluted and involves using `Boolean` as an ES6 constructor - which you should never do anyway.

traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
```diff
@@ -117,16 +116,15 @@ export class CacheGroupPage extends BasePage {
         if (cachegroup.Type == "EDGE_LOC") {
             const name = cachegroup.FailoverCG + this.randomize;
             await this.txtFailoverCG.click();
-            if(await browser.isElementPresent(element(by.xpath(`//select[@name="fallbackOptions"]//option[@label="`+ name + `"]`)))){
-                await element(by.xpath(`//select[@name="fallbackOptions"]//option[@label="`+ name + `"]`)).click();
-            }else{
+            if (await browser.isElementPresent(element(by.xpath(`//select[@name="fallbackOptions"]//option[@label="` + name + `"]`)))) {
```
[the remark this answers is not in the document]
ocket8888: Also, I wouldn't ask you to change it in this PR, but I think most people are more familiar with CSS selectors than XPath, so this could be written as:
```typescript
by.css(`select[name="fallbackOptions"] > option[label="${name}"]`)
```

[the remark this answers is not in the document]
dpham692: Fixed.

traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
```diff
@@ -117,16 +116,15 @@ export class CacheGroupPage extends BasePage {
...
+                await element(by.xpath(`//select[@name="fallbackOptions"]//option[@label="` + name + `"]`)).click();
+            } else {
                 result = undefined;
             }
         }
         await this.txtType.sendKeys(cachegroup.Type);
         await snp.ClickUpdate();
-        if(result != undefined)
-        {
+        if (result != undefined) {
```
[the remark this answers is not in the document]
dpham692: Fixed.

traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
```diff
@@ -158,6 +156,31 @@ export class CacheGroupPage extends BasePage {
...
         return result;
     }
+    public async CheckCSV(name: string): Promise<boolean> {
+        let linkName = name;
+        let result = false;
+        if (await browser.isElementPresent(element(by.xpath("//span[text()='" + linkName + "']"))) == true) {
+            result = true;
+        }
+        return result;
+    }
```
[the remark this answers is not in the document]
dpham692: Fixed.

traffic_portal/test/integration/specs/CacheGroup.spec.ts
dpham692: Fixed.

traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
```diff
@@ -158,6 +156,26 @@ export class CacheGroupPage extends BasePage {
...
         return result;
     }
+    public async CheckCSV(name: string): Promise<boolean> {
+        return element(by.cssContainingText("span", name)).isPresent();
+    }
+    public async ToggleTableColumn(name: string): Promise<boolean> {
+        let result = false;
+        await this.btnTableColumn.click();
+        //if it's already on, turn it off and return false
+        if (await browser.isElementPresent(element(by.xpath("//th[text()='" + name + "']"))) === true){
```
ocket8888, line 166: You don't need to compare a `boolean` to `true`, you can just use it as the condition. Technically, `browser.isElementPresent` returns a `Promise<any>`, but that's confusing because it's actually just an alias for `element(locator).isPresent` which returns a `Promise<boolean>`. So really this line could just be:

```typescript
if (await element(by.cssContainingText("th", name)).isPresent()) {
```

ocket8888, line 166: XPath is unfamiliar and confusing - especially because HTML is not an XML dialect, so for some valid HTML XPath queries would be fundamentally incapable of selecting certain tags. Those situations are a bit contrived, but in general web developers are much, much more likely to be familiar with CSS selectors than XPath.

This `by` directive can be written more simply and shorter as just `by.cssContainingText("th", name)`.

traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
```diff
@@ -158,6 +156,26 @@ export class CacheGroupPage extends BasePage {
...
+            await element(by.xpath("//label[text()=' " + name + "']")).click();
```
ocket8888, line 167: same as above RE: XPath

traffic_portal/test/integration/PageObjects/CacheGroup.po.ts
```diff
@@ -158,6 +156,26 @@ export class CacheGroupPage extends BasePage {
...
+            result = false;
+        }else{
+            //of it already off, turn it on return true
+            await element(by.xpath("//label[text()=' " + name + "']")).click();
```
ocket8888, line 171: same as above RE: XPath

ocket8888, line 171: It seems like you do this click on the same element regardless of the `if` condition, could it be moved out of the `if`/`else` entirely? If so, this method can be reduced to simply:

```typescript
public async ToggleTableColumn(name: string): Promise<boolean> {
    await this.btnTableColumn.click();
    const result = await element(by.cssContainingText("th", name).isPresent();
    await element(by.cssContainingText("label", name)).click();
    await this.btnTableColumn.click();
    return !result;
}
```
no `if` required

ocket8888 (approved):

PR #5847, opened 2021-05-13
Add maxmind database updating to t3c

ezelkow1:

This adds cmdline options to t3c for passing in a url for updating a maxmind database. If the URL is set then t3c will check for an existing file on disk. If it does not exist the database will be downloaded and unpacked, if it does exist then its modify time is used to issue an IMS on the provided URL to check if it needs to be updated, if it does then it is downloaded and unpacked.

- Traffic Ops ORT

Run t3c-apply with and without the -M url option pointing to a location of a maxmind gzipped database, verify that it gets unpacked properly into the ATS config dir. If no url provided verify that things are still updated or not as needed

### DIFF
--- a/cache-config/t3c-apply/config/config.go
+++ b/cache-config/t3c-apply/config/config.go
@@ -100,6 +100,11 @@ type Cfg struct {
 	DisableParentConfigComments bool
 	DefaultClientEnableH2       *bool
 	DefaultClientTLSVersions    *string
+	// MaxMindLocation is a URL string for a download location for a maxmind database
+	// for use with either HeaderRewrite or Maxmind_ACL plugins
+	MaxMindLocation string
+	TsHome          string
+	TsConfigDir     string
 }
 
 type UseGitFlag string
@@ -206,6 +211,7 @@ func GetCfg() (Cfg, error) {
 	disableParentConfigCommentsPtr := getopt.BoolLong("disable-parent-config-comments", 'c', "Whether to disable verbose parent.config comments. Default false.")
 	defaultEnableH2 := getopt.BoolLong("default-client-enable-h2", '2', "Whether to enable HTTP/2 on Delivery Services by default, if they have no explicit Parameter. This is irrelevant if ATS records.config is not serving H2. If omitted, H2 is disabled.")
 	defaultClientTLSVersions := getopt.StringLong("default-client-tls-versions", 'V', "", "Comma-delimited list of default TLS versions for Delivery Services with no Parameter, e.g. --default-tls-versions='1.1,1.2,1.3'. If omitted, all versions are enabled.")
+	maxmindLocationPtr := getopt.StringLong("maxmind-location", 'M', "", "URL of a maxmind gzipped database file, to be installed into the trafficserver etc directory.")
 
 	getopt.Parse()
 
@@ -244,6 +250,7 @@ func GetCfg() (Cfg, error) {
 	waitForParents := *waitForParentsPtr
 	dnsLocalBind := *dnsLocalBindPtr
 	help := *helpPtr
+	maxmindLocation := *maxmindLocationPtr
 
 	if help {
 		Usage()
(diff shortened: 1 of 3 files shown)


cache-config/t3c-apply/t3c-apply.go
```diff
@@ -226,3 +231,26 @@ func GitCommitAndExit(exitCode int, cfg config.Cfg) {
 	}
 	os.Exit(exitCode)
 }
+
+// CheckMaxmindUpdate will (if a url is set) check for a db on disk
```
rob05c, line 235: Nitpick: GoDoc should be complete sentences, punctuation etc.

cache-config/t3c-apply/config/config.go
```diff
@@ -100,6 +100,11 @@ type Cfg struct {
 	DisableParentConfigComments bool
 	DefaultClientEnableH2       *bool
 	DefaultClientTLSVersions    *string
+	// MaxMindLocation is a URL string for a download location for a maxmind database
+	// for use with either HeaderRewrite or Maxmind_ACL plugins
+	MaxMindLocation string
+	TsHome          string
```
rob05c, line 106: Why is this PR adding `TsHome` and `TsConfigDir`? Merge error?

rob05c (changes requested):
Looks good, just a few potential panics and the date race, if you don't mind fixing those

cache-config/t3c-apply/config/config.go
ezelkow1: We need the configdir to know where to store the mmdb, home I just brought a long for completeness in the data structure if it was ever needed

cache-config/t3c-apply/util/util.go
```diff
@@ -432,3 +433,81 @@ func Touch(fn string) error {
...
+	filePath := cfg.TsConfigDir + "/" + fileName
+	stdOut, _, code := t3cutil.Do(`date`,
+		"+%a, %d %b %Y %r %Z",
+		"-u",
+		"-r",
+		filePath)
+
+	if code == 0 {
+		// The file exists so run an IMS check
+		ims := fmt.Sprintf("If-Modified-Since: %s", string(stdOut[:len(stdOut)-1]))
```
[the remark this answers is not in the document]
ezelkow1: yea this is just trimming the newline, didnt realize trimspace would also handle that

cache-config/t3c-apply/util/util.go
```diff
@@ -432,3 +433,81 @@ func Touch(fn string) error {
...
+	if err != nil {
+		log.Errorf("error running gunzip: %v\n", err)
+		return false
+	}
+	if code == 0 {
+		log.Errorf("Maxmind DB at %s successfully updated from %s", filePath, cfg.MaxMindLocation)
+		return true
+	}
+
+	log.Errorf("failure gunzip maxmind db")
```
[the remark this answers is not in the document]
ezelkow1: Yea I can just reorganize this area, i think some of it was leftover from attempts at using .Do, which seemed to have issues with using the redirection commands that were necessary and so exec.Command was the method I got working first.

Either way these blocks aren't necessary any more and the gunzip err if there is one gets logged inside the err check anyway and these other checks can be removed since they were already checked (and exited on failure if needed) earlier

cache-config/t3c-apply/config/config.go
rob05c: Ah, I missed the usage, I see it now

cache-config/t3c-apply/util/util.go
```diff
@@ -432,3 +436,100 @@ func Touch(fn string) error {
...
+	urlpath := url.Path
+	fileName := path.Base(urlpath)
+
+	if fileName == "." || fileName == "/" {
+		log.Errorf("filename for maxmind from url invalid: %s", fileName)
+		return false
+	}
+
+	// Check if filename exists in ats etc
+	filePath := filepath.Join(cfg.TsConfigDir, "/", fileName)
```
rob05c, line 466: Nitpick: Join adds delimiters, the `/` is unnecessary. This will become `configdir//name`. Which works on Mac and Linux, POSIX specifies that duplicated slashes are collapsed. But it'll fail on Windows or other non `/` delimited systems.

rob05c (approved):
LGTM, code looks good; tested, works as expected, pulls the db, does an IMS/304 if it exists, works as expected with no `-M` argument.

PR #5851, opened 2021-05-14
Delivery Service Supported TLS Versions Blueprint

ocket8888:

This PR adds a blueprint for setting the supported TLS versions for content requests on a per-Delivery Service basis.

None

Read the blueprint.

- [x] Tests are unnecessary


blueprints/delivery.service.tls.versions.md
```diff
@@ -0,0 +1,351 @@
...
+protocol versions, to facilitate the ability of system administrators to
+quickly upgrade to a new protocol as it's released.
+
+A warning should also be issued if any Delivery Service is found to be a target
+of the Delivery Service being edited that does not support all of the TLS
+versions that the Delivery Service being edited itself does support, or if the
+Delivery Service being edited does not support all of the TLS versions supported
+by a Delivery Service of which it is a target, or if the Delivery Service being
+edited is the target of a Delivery Service which has other targets that support
+TLS versions not supported by the Delivery Service being edited.
```
rawlinp, line 112: TLS restrictions shouldn't apply to `CLIENT_STEERING/STEERING` since those won't affect TR anyways. So really it's just a matter of issuing a warning if target A and target B of the same client_steering/steering DS have different TLS restrictions.

blueprints/delivery.service.tls.versions.md
```diff
@@ -0,0 +1,351 @@
...
+
+- `/deliveryservices`
+- `/deliveryservices/{{ID}}`
+- `/deliveryservice_requests`
+
+When a Delivery Service or Delivery Service Request is created or modified
+using one of these endpoints, the API MUST verify that the `tlsVersions`
+property is either `null` or an array of strings that match the regular
+expression `\d+\.\d+` (an empty array may be treated implicitly as `null` in
+request bodies, but MUST ALWAYS be emitted as `null` in responses). In the
```
rawlinp, line 168: since it would never be valid to disable all TLS versions, why not just make `null` and `[]` both mean _no restrictions_?

blueprints/delivery.service.tls.versions.md
```diff
@@ -0,0 +1,351 @@
...
+
+Additionally, a migration will be needed to set the value of `tlsVersions` on
+existing Delivery Service Requests to `null`.
+
+## Documentation Impact
+The new field will need to be added to both the Delivery Services overview
+section as well as all of the API documentation for endpoints in which it will
+appear. Note also that the current documentation for the `PUT` method of
+`/deliveryservices/{{ID}}` incorrectly states that the response is empty,
+whereas it actually returns an array of Delivery Service representationns. The
```
rawlinp, line 225: sp: representationns

blueprints/delivery.service.tls.versions.md
ocket8888: So should it be possible to set `tlsVersions` to not `null` on STEERING/CLIENT_STEERING, or should the API just disallow it?

blueprints/delivery.service.tls.versions.md
ocket8888: In requests they do. It's just more obvious in responses to use `null` IMO, because then you don't ever have to think about what it means for the collection of allowed TLS versions to be empty. `null` seems like a less "special" case than `[]`, if that makes sense.

blueprints/delivery.service.tls.versions.md
rawlinp: I'm thinking the API should disallow it because it won't actually affect the client steering DS or its targets.

blueprints/delivery.service.tls.versions.md
rawlinp: Oh ok. That sounds fine

rawlinp (approved):

blueprints/delivery.service.tls.versions.md
```diff
@@ -0,0 +1,358 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Delivery Service TLS Versions
+
+## Problem Description
+Currently, supprted TLS versions cannot be configured. This means that a
```
srijeet0406, line 23: `supported`

blueprints/delivery.service.tls.versions.md
```diff
@@ -0,0 +1,358 @@
...
+`tlsVersions` property that can be either null, or an array of N strings,
+where N is *at least one*. In other words, this can be null, but cannot be an
+empty array.
+
+The reason for this is best understood by first understanding what `null` means
+for this field. If `tlsVersions` is `null`, it means that no TLS version
+restrictions in place; the behavior is the same as today without the field, all
+TLS versions are implicitly supported.
+
+An empty array would seem to signify that no TLS versions are supported, which
```
srijeet0406, line 60: Could we build this behavior into TP by default? Meaning the `TLSVersions` field is set to an empty array by default for `HTTP` delivery services, and is unmodifiable.

ocket8888: We can hide the controls, but because all Delivery Service "Types" are still the same "Type" of object we should not overwrite existing tlsVersions. For example, Traffic Router has issues with Delivery Services that don't have SSL keys but do support HTTPS, but you can't generate keys for a DS that doesn't exist. So, you could reasonably create a DS as HTTP-only with everything configured the way you want it, then generate the keys and switch it to use HTTPS

srijeet0406: Got you, makes sense.

srijeet0406 (approved):

blueprints/delivery.service.tls.versions.md
rob05c: I'm not strictly a -1, but UI hacks like this are poor design. It shouldn't be possible for users to assign TLS versions to an HTTP service. If the system can't handle that, we should fix the system.

rob05c: I realize that's not trivial, which is why I'm not -1'ing this. But I do wish we would prioritize fixing those footguns, instead of adding more hacks around them.

ocket8888: Breaking DSes out into separate types per Type (or at least routing Type) has been on my to-do list for well over a year now. It's a big ask.

PR #5852, opened 2021-05-15
Fix ort references

rob05c:
Fixes ort references in ansible.

Also fixes renamed verify references.

No tests, ansible code doesn't have a test framework. I manually tested with our build system, it fixes the issues.
No docs, no interface change.
No changelog, no interface change, and bug is not in a release.

- other, automation

Run Ansible, verify build works.

- master

### DIFF
--- a/cache-config/t3c/README.md
+++ b/cache-config/t3c/README.md
@@ -28,7 +28,7 @@ The `t3c` app is divided into commands, each of which can be run as an argument 
 |               command                  | description |
 | -------------------------------------: | :---------- |
 | [t3c-apply](./t3c-apply/README.md)     | Generate and apply cache configuration |
+| [t3c-check](./t3c-check/README.md)    | Check that new config can be applied |
 | [t3c-diff](./t3c-diff/README.md)       | Diff config files, like diff or git-diff but with config-specific logic |
 | [t3c-request](./t3c-request/README.md) | Request data from Traffic Ops |
 | [t3c-update](./t3c-update/README.md)   | Update a server's queue and reval status in Traffic Ops |
-| [t3c-verify](./t3c-verify/README.md)   | Verify config can be applied without error |
--- a/cache-config/t3c/README.md
+++ b/cache-config/t3c/README.md
@@ -25,10 +25,10 @@ The `t3c` app generates and applies cache configuration for Traffic Control.
 
 The `t3c` app is divided into commands, each of which can be run as an argument to `t3c` or by themselves, e.g. `t3c apply ...` or `t3c-apply ...`.
 
--------------------------------------- | --------------------------------------------------------------------
-[t3c-apply](./t3c-apply/README.md)     | Generate and apply cache configuration
-[t3c-update](./t3c-update/README.md)   | Update Traffic Ops, set or un-set update and reval flags for a cache
-[t3c-diff](./t3c-diff/README.md)       | Diff config files, like diff or git-diff but with config-specific logic
-[t3c-request](./t3c-request/README.md) | Request data from Traffic Ops
-[t3c-update](./t3c-update/README.md)   | Update a server's queue and reval status in Traffic Ops
-[t3c-verify](./t3c-verify/README.md)   | Verify config can be applied without error
+|               command                  | description |
+| -------------------------------------: | :---------- |
+| [t3c-apply](./t3c-apply/README.md)     | Generate and apply cache configuration |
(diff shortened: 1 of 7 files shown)


infrastructure/ansible/roles/ats/tasks/ats.yml
```diff
@@ -53,12 +53,22 @@
...
+    path: /tmp/trafficcontrol-cache-config
+    state: directory
+    recurse: yes
+    owner: root
+    group: root
+    mode: 0755
+
 - name: Run ORT Badass
-  shell: >-
+  shell:
```
jhg03a, line 66: You're missing the executable flag to force it to use bash.  This is needed because `PIPESTATUS` is a bash-ism.

jhg03a (changes requested):

infrastructure/ansible/roles/ats/tasks/ats.yml
rob05c: Fixed

cache-config/t3c/README.md
```diff
@@ -25,10 +25,10 @@ The `t3c` app generates and applies cache configuration for Traffic Control.
...
-[t3c-apply](./t3c-apply/README.md)     | Generate and apply cache configuration
-[t3c-update](./t3c-update/README.md)   | Update Traffic Ops, set or un-set update and reval flags for a cache
-[t3c-diff](./t3c-diff/README.md)       | Diff config files, like diff or git-diff but with config-specific logic
-[t3c-request](./t3c-request/README.md) | Request data from Traffic Ops
-[t3c-update](./t3c-update/README.md)   | Update a server's queue and reval status in Traffic Ops
-[t3c-verify](./t3c-verify/README.md)   | Verify config can be applied without error
+|               command                  | description |
+| -------------------------------------: | :---------- |
+| [t3c-apply](./t3c-apply/README.md)     | Generate and apply cache configuration |
+| [t3c-check](./t3c-verify/README.md)    | Check that new config can be applied |
```
zrhoffman, line 31: Still links to `./t3c-verify/README.md`

zrhoffman (changes requested):

infrastructure/ansible/roles/ats/tasks/ats.yml
jhg03a: This isn't at the place.  It's part of the shell module, not the task.

cache-config/t3c/README.md
rob05c: Fixed

jhg03a (changes requested):

zrhoffman (approved):
Confirmed that references to `t3c-verify` are removed (@jhg03a is still reviewing the main Ansible portion of the PR)

infrastructure/ansible/roles/ats/tasks/ats.yml
rob05c: Ah, moved

infrastructure/ansible/roles/ats/tasks/ats.yml
```diff
@@ -66,7 +66,7 @@
   environment:
     PATH: "{{ ansible_env.PATH }}:/bin:/usr/bin"
   shell:
-    executable: "{{ lookup('pipe','which bash') }}"
+    executable: "{{ lookup('pipe','/usr/bin/which bash') }}"
```
zrhoffman, line 69: I opted for using `type -p` instead of `which` in our build scripts since CentOS 7 and CentOS 8 do not include `which` by default.

rob05c: I'm fine with that. I thought `which` was part of LSB, but apparently not. But `type` is, so +1.

Not that the CentOS project cares much about adhering to the LSB standard, but IMO we should, and we should by default install CentOS packages necessary to fulfill it, and be ok using LSB things, IMO.

Saying "We adhere to and require the Linux Standard Base" is better than saying "we require whatever random stuff the CentOS devs decide to add or remove this version."

rob05c: Though, I don't fully understand the Ansible here (@jhg03a probably does), but I don't think `type` is what we need here, `type -t` just returns "alias." I think `which` here is needed to get the full path, not just whether it exists. It looks like `command -v` might be what we need, and it's part of POSIX.

zrhoffman: `command -v` sounds solid, too

PR #5855, opened 2021-05-17
t3c-apply: ensure files and directories are created with ats uid/gid

### LINKED ISSUE #5035: traffic_ops_ort/t3c etc/<directory> ownership and permissions settings need tweaking
## I'm submitting a ...
-  bug report

-  Traffic Ops ORT

## Current behavior:

Observed in traffic_ops_ort-4.1.0, t3c built from d809e9abc892b2dc690cd65b795e1fd1d256d151

For testing purposes I will remove trafficserver and trafficserver-debuginfo and blow away /opt/trafficserver, then run ORT badass as root.
With traffic_ops_ort.pl sub directories in /opt/trafficserver/etc/ such as "ssl" and "dscp" end up as root.root permission set to 0700.  This is consistent and part of my clean install involves chowning these subdirs.
I tested with t3c and it will create these same directories with permissions 0755 but still root.root which allows ATS to start with no complaints.

## Expected / new behavior:
Likely these subdirs should be created as ats.ats so they can be picked up when ats starts.
It might be that we want to keep the ssl directory as 700 owned by ats for security purposes, but possibly keep others like dscp at 755.

rpm remove trafficserver and trafficserver-debuginfo.  rm -rf /opt/trafficserver
set up server (for these cases I'll use the TO API to inject my test machine, then TP to clone delivery services).
run traffic_ops_ort.pl, wipe again, run t3c.
observe ownership and permissions of /opt/trafficserver/etc/  dscp, ssl, etc.

## Anything else:

Directories like  etc/trafficserver/body_factory are created by the ats install package.  For those directories either they should be ownership changed OR at least created to ensure they and their child directories are readable by the ats user.

traeak:

- [x] This PR fixes #5035 

This fixes a new issue where the new t3c-apply does not set ats uid/gid for hdr_rw_*, url_sig_*, etc files.
This won't fix directories created by an ats install (like body_factory)

- Traffic Ops ORT

Run badass and inspect permissions assigned to files and directoris in /opt/ats/etc/
Only the body_factory subdirectory will still be root as that is created by the ats rpm.

- master

### DIFF
--- a/cache-config/t3c-apply/util/util.go
+++ b/cache-config/t3c-apply/util/util.go
@@ -193,7 +193,7 @@ func ServiceStart(service string, cmd string) (bool, error) {
 	return false, nil
 }
 
-func WriteFileWithOwner(fn string, data []byte, uid int, gid int, perm os.FileMode) (int, error) {
+func WriteFileWithOwner(fn string, data []byte, uid *int, gid *int, perm os.FileMode) (int, error) {
 	fd, err := os.OpenFile(fn, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
 	if err != nil {
 		return 0, errors.New("unable to open '" + fn + "' for writing: " + err.Error())
@@ -205,8 +205,8 @@ func WriteFileWithOwner(fn string, data []byte, uid int, gid int, perm os.FileMo
 	}
 	fd.Close()
 
-	if uid > -1 && gid > -1 {
-		err = os.Chown(fn, uid, gid)
+	if uid != nil && gid != nil {
+		err = os.Chown(fn, *uid, *gid)
 		if err != nil {
 			return 0, errors.New("error changing ownership on '" + fn + "': " + err.Error())
 		}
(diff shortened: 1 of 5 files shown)


cache-config/t3c-apply/util/util.go
```diff
@@ -408,6 +408,14 @@ func doMkDir(name string, cfg config.Cfg, all bool) bool {
 					log.Errorf("unable to create the directory '%s': %v", name, err)
 					return false
 				}
+
+				if uid > -1 && gid > -1 {
```
rob05c, line 412: Negative uid and gids are valid on some *nix systems. Would you mind changing this to indicate with booleans or pointers? Maybe just change them to int*? ```if uid != nil && gid != nil { err = os.Chown(name, *uid, *gid)`

rob05c (changes requested):
Looks good! Code looks good, tested, works as expected, new files are created with `ats` owner.

Only thing is the negative uids, if you don't mind fixing that, I'll merge

cache-config/t3c-apply/util/util.go
traeak: changed as instructed

rob05c (approved):
Looks good! My concern was addressed. Looks like `go vet` is failing, looks like a test is failing to build. This can be merged when the checks pass.

PR #5856, opened 2021-05-17
Fix CDN in a Box vnc dependencies in Dockerfile

zrhoffman:

This PR gets the CDN in a Box `vnc` image to build by installing `tigervnc` specifically instead of installing the `vnc` alias and installing Google Noto fonts instead of `webcore-fonts`.

- CDN in a Box

Verify that the dependencies issue mentioned by @ARMmaster17 on the ATC slack no longer occurs when building the `vnc` image:

```
#6 13.52 Error: Unable to find a match: webcore-fonts vnc
------
executor failed running [/bin/sh -c dnf -y install <https://download1.rpmfusion.org/free/el/rpmfusion-free-release-7.noarch.rpm> epel-release &&     dnf -y install xterm firefox git tigervnc-server sudo bind-utils net-tools which passwd which                    fluxbox webcore-fonts terminus-fonts vnc mplayer wget openssl curl nc &&     dnf -y clean all && rm -rf /var/cache/dnf]: exit code: 1
ERROR: Service 'vnc' failed to build : Build failed
```

- [x] CDN in a Box changed only, tests are unnecessary


ARMmaster17:
Thank you for this PR, that was fast! I tried pulling this branch and running it locally. The container builds fine now, but the VNC service never comes up. Poking around some more I found that the old `vnc` package used a really old version of TigerVNC. Newer versions use `systemd` instead of the `vncserver` script, and the container as it is right now errors out with:

```
vncserver has been replaced by a systemd unit.
Please read /usr/share/doc/tigervnc/HOWTO.md for more information.
```

I've spent all morning on this without much luck. I can work around this for the time being using SOCKS. I'll keep poking at it to see if I find anything new, but if anyone else wants to give a shot at this PR, here's all the things I tried that don't work:
- Revert to `tigervnc-server` v1.7.1 (package not in any CentOS repos, and manual RPM install was having odd dependency issues).
- Install `systemd` packages into Docker container (requires a multi-step build process)
- Use `centos/systemd` as base image (running into permission errors)

zrhoffman:
Red Hat [re-added the `vncserver` script in CentOS Stream](https://github.com/TigerVNC/tigervnc/issues/1161#issuecomment-743221598) (and it still exists in CentOS 7), so this is a CentOS 8-only issue. Manually re-added the `vncserver` script in c6b0195b43.

ARMmaster17 (approved):
Tested locally and everything works! Thank you!

ocket8888 (approved):

PR #5858, opened 2021-05-17
Fix t3c diff for empty files

rob05c:
Fixes t3c-diff to report a diff for a new semantically empty file, which causes t3c-apply to create it.

This was a bug, because if another config (like remap.config plugins) references the file, it may be legal to be empty, but ATS will fail to load if it doesn't exist.

No docs, no changelog, no interface change and bug isn't in a release.
Includes tests.

- Traffic Ops ORT

Run ORT integration tests. Create an empy file in Traffic Ops such as with a location param to a custom file with no other parameters, ensure t3c-apply creates the file.

- master

### DIFF
--- a/cache-config/t3c-diff/t3c-diff.go
+++ b/cache-config/t3c-diff/t3c-diff.go
@@ -53,12 +53,12 @@ func main() {
 		os.Exit(4)
 	}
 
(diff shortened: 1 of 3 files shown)


traeak (approved):
looks fine and tested successfully on a DS with empty header rewrite and with and without service categories.

PR #5865, opened 2021-05-18
Provide ability to fetch Traffic Vault secret key from HashiCorp Vault

rawlinp:
Add new `hashicorp_vault` options to the Postgres Traffic Vault backend in order to provide the ability to fetch the TV secret AES key from a HashiCorp Vault server.

- Documentation
- Traffic Ops
- Traffic Vault

Install Vault locally (if on macOS, you can just `brew install vault`) and follow the instructions below to set up AppRole and the AES key secret:
```
# start the dev server in its own terminal:

    vault server -dev

# in a separate terminal, run:

    export VAULT_ADDR='http://127.0.0.1:8200'
    export VAULT_TOKEN=<insert Root Token from 1st terminal output here>
    vault status

    # create a secret (AES key base64-encoded)
    vault kv put secret/trafficvault traffic_vault_key="$(openssl rand -base64 32)"
    vault kv get secret/trafficvault

    # enable approle authentication
    vault auth enable approle

    # create a policy for the approle to use
    vault policy write my-policy - << EOF
# Dev servers have version 2 of KV secrets engine mounted by default, so will
# need these paths to grant permissions:
path "secret/data/*" {
  capabilities = ["create", "update"]
}
path "secret/data/trafficvault" {
  capabilities = ["read"]
}
EOF

    # create the approle
    vault write auth/approle/role/my-role \
        secret_id_ttl=10m \
        token_num_uses=10 \
        token_ttl=20m \
        token_max_ttl=30m \
        secret_id_num_uses=40 \
        token_policies=my-policy

    # get the approle's role_id and secret_id for it to authenticate with
    vault read -field=role_id auth/approle/role/my-role/role-id
    vault write -f -field=secret_id auth/approle/role/my-role/secret-id
```
Then follow the documentation to configure the Postgres TV backend to use Vault instead of `aes_key_location`. Once TO starts up successfully, run the TO API tests. Then, restart TO using `aes_key_location` (but use the same key), and ensure that still works.

- [x] This PR doesn't add new tests, but it might be better to augment CiaB to optionally use a HashiCorp Vault container in order to test the actual integration
- [x] No changelog necessary since this is already covered in the changelog

### DIFF
--- a/traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
+++ b/traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
@@ -36,6 +36,7 @@ import (
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/trafficvault"
 
 	validation "github.com/go-ozzo/ozzo-validation"
+	"github.com/go-ozzo/ozzo-validation/is"
 	"github.com/jmoiron/sqlx"
 	"github.com/lib/pq"
 )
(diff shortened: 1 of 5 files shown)


traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -482,8 +505,22 @@ func validateConfig(cfg Config) error {
 		"port":                  validation.Validate(cfg.Port, validation.By(tovalidate.IsValidPortNumber)),
 		"max_connections":       validation.Validate(cfg.MaxConnections, validation.Min(0)),
 		"query_timeout_seconds": validation.Validate(cfg.QueryTimeoutSeconds, validation.Min(0)),
-		"aes_key_location":      validation.Validate(cfg.AesKeyLocation, validation.Required),
 	})
+	aesKeyLocSet := cfg.AesKeyLocation != ""
+	hashiCorpVaultSet := cfg.HashiCorpVault != nil
```
mattjackson220, line 510: maybe add something like this `&& *cfg.HashiCorpVault != HashiCorpVault{}` so people can leave an empty field in the cdn.conf? not important to me if you think thats not worth doing, just an idea cause i tried testing it that way

rawlinp: sure, I can live with that!

mattjackson220 (approved):
Looks great! code looks good, tested locally and works as expected!

PR #5868, opened 2021-05-19
Fix log out function in TP integration test

dpham692:

So sometimes protractor run the process so fast it does not wait for the page to be load. So to make sure the page is completely checkout, I put the wait in log out function instead of log in.

- CDN in a Box
- Traffic Portal
- CI tests

Run `npm test -- --params.baseUrl https://localhost --capabilities.shardTestFiles true` in traffic_portal/test/integration directory and make sure the test pass in your local.
Check CI TP test and make sure none fail.

### DIFF
--- a/traffic_portal/test/integration/PageObjects/LoginPage.po.ts
+++ b/traffic_portal/test/integration/PageObjects/LoginPage.po.ts
@@ -16,7 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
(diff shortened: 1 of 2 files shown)


dpham692:
I removed the check in log in because when I check failed cases picture (below) and I noticed user was never log out so problem is not come from the log in function but in log out function.
TestFailed

ocket8888 (approved):
GHAs have passed 5 times in a row.

PR #5870, opened 2021-05-19
Migrate user tests from E2E to integration

rimashah25:

- CDN in a Box
- Traffic Portal
- CI tests

Required: Local TO or CIAB TO running

Go to traffic_portal/test/integration 
Update specs, apiUrl and baseUrl in config.json as follows:
```
"specs": [
    "specs/Users.spec.js"
  ],

"params": {
    "apiUrl": " https://localhost:9090/api/4.0",
    "baseUrl": "https://localhost:9090/",
....
}
```
and then run the below two commands in a separate terminal window:
```
1. For me chrome v90 doesn't work, so, I ran the following cmd: webdriver-manager start --versions.chrome 86.0.4240.22
2. npm test -- --params.baseUrl https://localhost:9090 --capabilities.shardTestFiles true
```
```localhost port may differ for every user```

All 10 integration test cases should pass.

- [x] This PR includes tests

## Additional Information

There is no logic change to require an update in documentation

### CODE REVIEW

File: traffic_portal/test/integration/PageObjects/UsersPage.po.ts [hunk 1]
```diff
@@ -46,13 +68,39 @@ export class UsersPage extends BasePage {
...
+    private txtSearch = element(by.id('usersTable_filter')).element(by.css('label input'));
+    private btnTableColumn = element(by.css('[title="Select Table Columns"]'));
     private randomize = randomize;
 
-    async OpenUserPage(){
-      let snp = new SideNavigationPage();
-      await snp.ClickUserAdminMenu();
-      await snp.NavigateToUsersPage();
-     }
+    async OpenUserPage() {
```
ocket8888, line 75: Missing return type on call signature

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -46,13 +68,39 @@ export class UsersPage extends BasePage {
...
+        let snp = new SideNavigationPage();
+        await snp.NavigateToUsersPage();
+    }
+
+    async OpenUserMenu() {
```
ocket8888, line 80: Missing return type on call signature

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -46,13 +68,39 @@ export class UsersPage extends BasePage {
...
+    private btnTableColumn = element(by.css('[title="Select Table Columns"]'));
     private randomize = randomize;
 
-    async OpenUserPage(){
-      let snp = new SideNavigationPage();
-      await snp.ClickUserAdminMenu();
-      await snp.NavigateToUsersPage();
-     }
+    async OpenUserPage() {
+        let snp = new SideNavigationPage();
```
ocket8888, line 76: nit: `snp` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -46,13 +68,39 @@ export class UsersPage extends BasePage {
...
+        await snp.NavigateToUsersPage();
+    }
+
+    async OpenUserMenu() {
+        let snp = new SideNavigationPage();
```
ocket8888, line 81: nit: `snp` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -46,13 +68,39 @@ export class UsersPage extends BasePage {
...
     private randomize = randomize;
 
-    async OpenUserPage(){
-      let snp = new SideNavigationPage();
-      await snp.ClickUserAdminMenu();
-      await snp.NavigateToUsersPage();
-     }
+    async OpenUserPage() {
+        let snp = new SideNavigationPage();
+        await snp.NavigateToUsersPage();
```
ocket8888, line 77: this works fine as-is, but if it's more intuitive for you, you should knowyou can just directly return this Promise to run asynchronously and resolve the Promise returned by this method when it resolves. It'll work the same way, essentially, since this method is `async`, but might make more sense since I asked for a return type for the function to actually `return` something explicitly. (To be clear, it is returning something implicitly already).

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -46,13 +68,39 @@ export class UsersPage extends BasePage {
...
+
+    public async CheckCSV(name: string): Promise<boolean> {
+        return element(by.cssContainingText("span", name)).isPresent();
+    }
+
+    public async CheckToggle(name: string): Promise<boolean> {
+        let result = false;
+        await this.btnTableColumn.click();
+        //if the box is already checked, uncheck it and return false
+        if (await browser.isElementPresent(element(by.xpath("//th[text()='" + name + "']"))) === true) {
```
ocket8888, line 93: nit: instead of using an XPath here, you could just do like you did above with the CheckCSV span:

```typescript
if (await element(by.cssContainingText("th", name)).isPresent()) {
    // ...
}
```

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -46,13 +68,39 @@ export class UsersPage extends BasePage {
...
+            await element(by.xpath("//label[text()=' " + name + "']")).click();
```
ocket8888, line 94: same as above RE: `by.xpath` vs `by.cssContainingText`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -46,13 +68,39 @@ export class UsersPage extends BasePage {
...
+            result = false;
+        } else {
+            //if the box is unchecked, then check it and return true
+            await element(by.xpath("//label[text()=' " + name + "']")).click();
```
ocket8888, line 98: same as above RE: `by.xpath` vs `by.cssContainingText`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
ocket8888, line 93: Don't need to compare boolean values to `true`, you can just use them as the condition.

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
-      }else if(await basePage.GetOutputMessage() == user.validationMessage){
+      }else if(await basePage.GetOutputMessage() === user.validationMessage){
         result = true;
       }else{
         result = false;
       }
       return result;
     }
 
+    public async SearchUser(nameUser: string) {
```
ocket8888, line 130: Missing return type on call signature

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+        await this.txtSearch.clear();
+        await this.txtSearch.sendKeys(name);
+        await element.all(by.repeater('u in ::users')).filter(function (row) {
+            return row.element(by.name('username')).getText().then(function (val) {
+                return val === name;
+            });
+        }).first().click();
+    }
+
+    public async SearchEmailUser(nameEmail: string) {
```
ocket8888, line 143: Missing return type on call signature

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+      }else if(await basePage.GetOutputMessage() === user.validationMessage){
         result = true;
       }else{
         result = false;
       }
       return result;
     }
 
+    public async SearchUser(nameUser: string) {
+        let snp = new SideNavigationPage();
```
ocket8888, line 131: nit: `snp` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+        await this.txtSearch.sendKeys(name);
+        await element.all(by.repeater('u in ::users')).filter(function (row) {
+            return row.element(by.name('username')).getText().then(function (val) {
+                return val === name;
+            });
+        }).first().click();
+    }
+
+    public async SearchEmailUser(nameEmail: string) {
+        let snp = new SideNavigationPage();
```
ocket8888, line 144: nit: `snp` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+            default:
+                return undefined;
+        }
+        return await basePage.GetOutputMessage().then(value => user.validationMessage === value);
+    }
+
+    public async RegisterUser(user: RegisterUser): Promise<boolean> {
+        let result = false;
+        let basePage = new BasePage();
+        let snp = new SideNavigationPage();
```
ocket8888, line 173: nit: `snp` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
         result = true;
       }else{
         result = false;
       }
       return result;
     }
 
+    public async SearchUser(nameUser: string) {
+        let snp = new SideNavigationPage();
+        let name = nameUser + this.randomize;
```
ocket8888, line 132: nit: `name` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+        await element.all(by.repeater('u in ::users')).filter(function (row) {
+            return row.element(by.name('username')).getText().then(function (val) {
+                return val === name;
+            });
+        }).first().click();
+    }
+
+    public async SearchEmailUser(nameEmail: string) {
+        let snp = new SideNavigationPage();
+        let name = nameEmail + this.randomize;
```
ocket8888, line 145: nit: `name` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+        await this.txtSearch.sendKeys(name);
+        await element.all(by.repeater('u in ::users')).filter(function (row) {
+            return row.element(by.name('email')).getText().then(function (val) {
+                return val === name;
+            });
+        }).first().click();
+    }
+
+    public async UpdateUser(user: UpdateUser): Promise<boolean  | undefined> {
+        let basePage = new BasePage();
```
ocket8888, line 157: nit: `basePage` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+                break;
+            default:
+                return undefined;
+        }
+        return await basePage.GetOutputMessage().then(value => user.validationMessage === value);
+    }
+
+    public async RegisterUser(user: RegisterUser): Promise<boolean> {
+        let result = false;
+        let basePage = new BasePage();
```
ocket8888, line 172: nit: `basePage` is never reassigned, could be `const`

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+        await this.txtSearch.clear();
+        await this.txtSearch.sendKeys(name);
+        await element.all(by.repeater('u in ::users')).filter(function (row) {
+            return row.element(by.name('email')).getText().then(function (val) {
+                return val === name;
+            });
+        }).first().click();
+    }
+
+    public async UpdateUser(user: UpdateUser): Promise<boolean  | undefined> {
```
ocket8888, line 156: It looks like the tests actually only check that the resolution of this is "truthy" or literally `true`, so why not just return a `Promise<boolean>` instead of allowing it to be `undefined`? Allowing `undefined` could be confusing, for example, if `UpdateUser` returns a `Promise<boolean>`, these are the same check:

```typescript
expect(await usersPage.UpdateUser(u)).toBeTrue(); // or .toBe(true)
expect(await usersPage.UpdateUser(u)).not.toBeFalse(); // or .toBe(false)
```

... but if it returns a `Promise<boolean | undefined>`, they are **not** the same. These pairs of checks also have the same problem where they're identical for `Promise<boolean>` but not `Promise<boolean>`:

```typescript
expect(await usersPage.UpdateUser(u)).toBeFalse();
expect(await usersPage.UpdateUser(u)).toBe(false);
expect(await usersPage.UpdateUser(u)).toBeFalsey()
```

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
```diff
@@ -61,22 +109,94 @@ export class UsersPage extends BasePage {
...
+        let basePage = new BasePage();
+        switch (user.description) {
+            case "update registered user's fullname":
+                await this.txtFullName.sendKeys(user.NewFullName);
+                await basePage.ClickUpdate();
+                break;
+            default:
+                return undefined;
+        }
+        return await basePage.GetOutputMessage().then(value => user.validationMessage === value);
```
ocket8888, line 200: nit: you don't need to `await` a Promise you're returning.

traffic_portal/test/integration/specs/Users.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+                    expect(await usersPage.UpdateRegisterUser(updateRegisterUser)).toBe(true);
+                    await usersPage.OpenUserPage();
+                });
+            });
+            it('can logout', async () => {
+                expect(await topNavigation.Logout()).toBeTruthy();
+            });
+        });
+    });
+});
```
ocket8888, line 91: Missing newline at EOF

traffic_portal/test/integration/Data/users.ts
```diff
@@ -0,0 +1,89 @@
...
+                    validationMessage: "User created"
+                },
+            ],
+            register: [
+                {
+                    description: "create a registered User",
+                    Email: "test2@cdn.tc.com",
+                    Role: "admin",
+                    Tenant: "- tenantSame",
+                    validationMessage: "Sent user registration to {{ test2@cdn.tc.com"+randomize+"}} with the following permissions [ role: admin | tenant: tenantSame"+randomize+" ]"
```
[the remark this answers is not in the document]
rimashah25: I am not sure I understand what you are trying to imply.

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
rimashah25: Why would I need a return type when the function isn't returning anything?

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
rimashah25: Same question as the above

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
rimashah25: Again, not returning anything, so why a return type?

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
rimashah25: Same question as above

traffic_portal/test/integration/Data/users.ts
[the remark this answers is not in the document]
ocket8888: What I mean is that
```typescript
"part of a string "+something+" the rest of the string";
```
is the same as
```typescript
`part of a string ${something} the rest of the string`;
```

traffic_portal/test/integration/PageObjects/UsersPage.po.ts
ocket8888: That's why, because it actually does return something, and that's not immediately obvious:
```javascript
// A Promise that just resolves to 5
const p = new Promise(resolve=>resolve(5));
async function sideNavMethod() {
	await p; // discarding return value; just wanted to 'await' something
}

async function userPageMethod() {
	await sideNavMethod();
}

// outputs: object
console.log(typeof(userPageMethod()));
```

traffic_portal/test/integration/Data/users.ts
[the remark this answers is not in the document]
rimashah25: Ok, Thanks for the explanation. I am neutral on the easier read wrt template string vs string manipulation. But I have seen template string usage in TO/TM/TS go code and can maintain the same in typescript too.

ocket8888 (approved):

PR #5872, opened 2021-05-19
Fix enroller build error due to out-of-scope ARG

ocket8888:

This PR fixes a bug in the CDN-in-a-Box build system where the enroller's Dockerfile uses an `ARG` that was defined in a previous build stage. From the Dockerfile reference:

> _"An `ARG` instruction goes out of scope at the end of the build stage where it was defined. To use an arg in multiple stages, each stage must include the `ARG` instruction."_

- CDN in a Box

Verify that the enroller image can build (may need to clean up Docker cache).

- master

- [x] Tests are unnecessary

### DIFF
--- a/infrastructure/cdn-in-a-box/enroller/Dockerfile
+++ b/infrastructure/cdn-in-a-box/enroller/Dockerfile
@@ -16,7 +16,6 @@
 # under the License.
 
 FROM debian:buster AS enroller-dependencies
-ARG ENROLLER_DEBUG_BUILD=false
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
--- a/infrastructure/cdn-in-a-box/enroller/Dockerfile
+++ b/infrastructure/cdn-in-a-box/enroller/Dockerfile
@@ -34,6 +34,8 @@ RUN go_version=$(cat /GO_VERSION) && \
 ENV GOPATH=/go
 
 FROM enroller-dependencies AS enroller-builder
+ARG ENROLLER_DEBUG_BUILD=false
+
 # enroller source and dependencies
 COPY ./lib/ /go/src/github.com/apache/trafficcontrol/lib/
 COPY ./go.mod ./go.sum /go/src/github.com/apache/trafficcontrol/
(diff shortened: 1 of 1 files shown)


infrastructure/cdn-in-a-box/enroller/Dockerfile
```diff
@@ -43,7 +43,7 @@ COPY ./traffic_ops/v4-client/ /go/src/github.com/apache/trafficcontrol/traffic_o
 COPY ./infrastructure/cdn-in-a-box/ /go/src/github.com/apache/trafficcontrol/infrastructure/cdn-in-a-box/
 
 WORKDIR /go/src/github.com/apache/trafficcontrol/infrastructure/cdn-in-a-box/enroller
-RUN set -o errexit -o nounset; \
+RUN set -o errexit; \
```
zrhoffman, line 46: We shouldn't remove `-o nounset`, though, since it found an error. We just need to move the definition for the build arg that is not properly defined.

zrhoffman (changes requested):

infrastructure/cdn-in-a-box/enroller/Dockerfile
ocket8888: I did that on purpose instead of copying the arg down, because I was thinking "well it's only unset because I'm not debugging". That's not how that works.

zrhoffman (approved):

zrhoffman (changes requested):
Now the `ENROLLER_DEBUG_BUILD` arg is defined for the `enroller-builder` stage, but it is still unnecessarily-defined in the `enroller-dependencies` build stage.

https://github.com/apache/trafficcontrol/blob/56c9d148d52ca77f65e0caaaf1a611ff38d0584d/infrastructure/cdn-in-a-box/enroller/Dockerfile#L18-L19

Having the build arg defined for the `enroller-dependencies` build stage *does* change build behavior, it busts the cache for the `enroller-dependencies` layers every time the `ENROLLER_DEBUG_BUILD` arg changes.

zrhoffman (approved):

zrhoffman:
Passed on @ocket8888's fork
https://github.com/ocket8888/trafficcontrol/actions/runs/861243318

PR #5874, opened 2021-05-20
Define the ENROLLER_DEBUG_BUILD arg in the enroller-builder stage

zrhoffman:

This PR fixes a bug introduced by #5866 that affects older Docker versions.

#5866 defined a new build stage in the `enroller` Dockerfile, and without moving the definition for the `ENROLLER_DEBUG_BUILD` build arg, `ENROLLER_DEBUG_BUILD` is no longer set within the `enroller-builder` build stage. <strike>On newer versions, like Docker `20.10.6`, this is not a build issue, it just results in the Dockerfile never making use of `ENROLLER_DEBUG_BUILD` when it is truthy.

On older Docker versions, however, like `18.06.0-ce, build 0ffa825`,</strike> This causes a build failure:

```dockerfile
/bin/sh: 1: ENROLLER_DEBUG_BUILD: parameter not set
ERROR: Service 'enroller' failed to build : The command '/bin/sh -c set -o errexit -o nounset;     go clean;     go mod vendor -v;     gcflags= ldflags=;     tags='usergo netgo';     if [ "$ENROLLER_DEBUG_BUILD" = true ]; then         apt-get install -y --no-install-recommends gcc libstdc++-8-dev;         echo 'Building Enroller without optimization or inlining';         gcflags='all=-N -l';     else         echo 'Optimizing Enroller build';         ldflags='-s -w';     fi;     go build -ldflags "$ldflags" -gcflags "$gcflags" -tags "$tags"' returned a non-zero code: 2
```

This PR moves the `ENROLLER_DEBUG_BUILD` build arg definition to the `enroller-builder` build stage.

- CDN in a Box

In the CDN in a Box directory, <strike>using Docker version `18.06.0-ce, build 0ffa825`</strike> build the `enroller` image, make sure building it suceeds.

- master (8c32ae28ee)

- [x] Building CDN in a Box is sufficient to test, no further tests necessary
- [x] Bugix, documentation is unnecessary

### DIFF
--- a/infrastructure/cdn-in-a-box/enroller/Dockerfile
+++ b/infrastructure/cdn-in-a-box/enroller/Dockerfile
@@ -16,7 +16,6 @@
 # under the License.
 
 FROM debian:buster AS enroller-dependencies
-ARG ENROLLER_DEBUG_BUILD=false
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
@@ -34,6 +33,7 @@ RUN go_version=$(cat /GO_VERSION) && \
 ENV GOPATH=/go
 
 FROM enroller-dependencies AS enroller-builder
+ARG ENROLLER_DEBUG_BUILD=false
 # enroller source and dependencies
 COPY ./lib/ /go/src/github.com/apache/trafficcontrol/lib/
 COPY ./go.mod ./go.sum /go/src/github.com/apache/trafficcontrol/


ocket8888:
Related to #5872 ?

zrhoffman:
> also, I hit this issue using Docker version 20.10.6, build 370c28948e

Oh right, I forgot that I always build CDN in a Box using a function instead of `docker-compose` that adds all of the extra compose files that I use.

> Related to #5872 ?

Yes. We shouldn't remove `-o nounset`, though, since it found an error.

zrhoffman:
Fixed in #5872

PR #5876, opened 2021-05-21
Use `jq` to set the trafficserver package version for t3c integration tests

zrhoffman:

- This PR uses `jq` to manipulate the JSON of `tc-fixtures.json` to set the ATS version instead of using `sed`

- Added validation to ensure that the ATS RPM version was set in the right place.

- Includes what the unexpected version format was in the error message citing unexpected version format

Manipulating the JSON to set the ATS RPM version in this way also has the benefit of not relying on the file containing a specific string in order for replacement to occur.

- Traffic Control Cache Config integration tests

Run the `t3c` integration tests

- [x] Intended behavior is unchanged, documentation is unnecessary

## Additional Information

It would also be nice to use `jq` to reformat `tc-fixtures.json` for easy comparison once `jq` sets the ATS RPM version, but since that would create a huge diff, I would only include the reformatting commit if the other changes in the PR are approved.

### DIFF
--- a/.github/workflows/cache-config-tests.yml
+++ b/.github/workflows/cache-config-tests.yml
@@ -25,6 +25,7 @@ on:
       - go.mod
       - go.sum
       - GO_VERSION
+      - lib/go-atscfg/**.go
       - traffic_ops/*client/**.go
       - traffic_ops/traffic_ops_golang/**.go
       - cache-config/**.go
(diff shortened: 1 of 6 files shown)


jrushford (approved):
Looks good.

zrhoffman:
> It would also be nice to use `jq` to reformat `tc-fixtures.json` for easy comparison once `jq` sets the ATS RPM version, but since that would create a huge diff, I would only include the reformatting commit if the other changes in the PR are approved.

Rebased to avoid a merge conflict now that the PR includes reformatting `tc-fixtures.json` in d4d57b279f.

PR #5879, opened 2021-05-24
Add TO Client API for CDN Automation

dmohan001c:

- CDN in a Box
- Traffic Control Client 
- Traffic Ops
- CI tests

Execute all the Integration tests and make sure the tests passed.


traffic_ops/testing/api/v4/cdns_test.go
```diff
@@ -348,6 +373,46 @@ func GetTestCDNs(t *testing.T) {
 	}
 }
 
+func GetTestCDNsbyDomainName(t *testing.T) {
+	if len(testData.CDNs) < 1 {
```
ocket8888, line 377: line 382 dereferences `testData.CDNs[1]`, so this needs to check that the length is at least 2 to prevent segfaulting (or it could instead use the `0` element on line 382)

dmohan001c: updated the code to get the first value of CDN.

traffic_ops/testing/api/v4/cdns_test.go
```diff
@@ -376,3 +441,166 @@ func DeleteTestCDNs(t *testing.T) {
...
+	firstData := testData.CDNs[0]
+	firstData.DomainName = ""
+	resp, reqInf, err := TOSession.CreateCDN(firstData, client.RequestOptions{})
+	if err == nil {
+		t.Errorf("Expected 'domainName' cannot be blank  but got - alerts: %+v", resp.Alerts)
+	}
+	if reqInf.StatusCode != http.StatusBadRequest {
+		t.Errorf("Expected 400 status code, got %v", reqInf.StatusCode)
+	}
+}
```
ocket8888, line 486: I don't think these tests are comprehensive enough to test what they claim to test; CDN names and domain names must be unique, so if there was a bug in the API that let you create a CDN with a blank name or a blank domain name, it would still give you back a non-nil error and a 400 status code. Because the CDN defined by `testData.CDNs[0]` was already created prior to this test - as long as creation is working properly - a CDN with the same domain name and name will exist when you attempt to create it in the first test and the second test, respectively.

ocket8888 (approved):

PR #5880, opened 2021-05-24
Api info reorganization

ocket8888:

This PR restructures the familiar `api.APIInfo` struct a bit. It retains a reference to an http response writer, allowing many `api` package functions to be aliased with a much shorter argument list.

For example,

```go
api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
```

can be called as simply
```go
inf.HandleErr(errCode, userErr, sysErr)
```

... allowing you to pass only the information that has actually changed since the handler function's execution began.

New methods include aliases for response writing, some IMS/IUS/IM niceties, parsing/parse-and-validating, and a handler for parsing and writing db errors in the same step, which converts

```go
if err := dbOperation(); err != nil {
    userErr, sysErr, errCode := api.ParseDBError(err)
    api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
    return
}
```
into
```go
if err := dbOperation(); err != nil {
    inf.HandleDBErr(err)
    return
}
```

This also changes the info struct's name from `api.APIInfo` to `api.Info` (because the former "stutters"), moves it and its methods to its own file, and changes the `Version` member from a reference to an `api.Version` to just an `api.Version` value. Apart from naming, call signature (now requires the response writer to construct) and that one field change, most code is untouched. However, the `/acme_accounts` handlers have been retooled to showcase the new Info methods, and part of the DSRs handler was also done to show off an IMS-related method (because `/acme_accounts` does not support IMS requests).

Also, this PR removed the deprecated `/lib/go-tc.GetHandleErrorsFunc` and the last three places it was used.

- Traffic Ops

Make sure unit and client/TO integration tests still pass.

- [x] This PR includes tests
- [x] Documentation is unnecessary, because the changes only affect internal development.
- [x] An update to CHANGELOG.md is not necessary because client-facing behavior is unaltered.

### DIFF
--- a/lib/go-rfc/http.go
+++ b/lib/go-rfc/http.go
@@ -39,6 +39,7 @@ const (
 	PermissionsPolicy  = "Permissions-Policy"  // W3C "Permissions Policy"
 	Server             = "Server"              // RFC7231§7.4.2
 	UserAgent          = "User-Agent"          // RFC7231§5.5.3
+	Location           = "Location"            // RFC7231§6.3.2
 	Vary               = "Vary"                // RFC7231§7.1.4
 )
 
--- a/lib/go-tc/alerts.go
+++ b/lib/go-tc/alerts.go
@@ -39,6 +39,15 @@ type Alert struct {
 	Level string `json:"level"`
 }
 
+// NewAlert constructs and returns an Alert of the given level with the given
+// text.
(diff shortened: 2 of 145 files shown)


ocket8888:
Superseded by #7902

PR #5884, opened 2021-05-24
Fix Traffic Vault postgres backend to match Riak backend for empty URI signing keys

rawlinp:

Make the Traffic Vault Postgres backend return the same JSON response as the Riak
backend when the delivery service has no URI signing keys.

- Traffic Ops
- Traffic Vault

Run the added TO API test with the Postgres Traffic Vault backend configured, followed by the Riak Traffic Vault backend, and verify both pass.

- master

- [x] This PR includes tests
- [x] Bugfix, no docs necessary
- [x] No changelog necessary, unreleased bug

## Additional Information
I'm not 100% sure if this disparity between the responses provided by the Postgres and Riak Traffic Vault backends in this case even causes any real world problems, but I'd rather be safe than sorry, especially since it is a simple fix.


mattjackson220 (approved):
Looks great! Tests pass for Riak and Postgres, local testing works as expected

PR #5887, opened 2021-05-25
Fix t3c reval to reload

rob05c:
Reval reload was failing because reval doesn't get and check packages
from Traffic Ops. Which it doesn't need to, but the func to check
if a package was installed was only checking the cache populated by
the TO fetch.

This fixes it to check rpm if the requested package isn't in the
cache, which then fixes the call to know trafficserver is installed,
which then fixes the thing that calls that to reload ATS correctly.

Includes an integration test to verify a reval triggers a reload properly from now on.
No changelog, no docs, no interface change, and bug isn't in a release.

- Traffic Ops ORT

Run t3c/ORT Integration tests. Create a Reval and set the Reval Flag on a server, then run `t3c apply --run-mode=revalidate`, verify it gets the new file and Reloads ATS.

- master

## Additional Information

### DIFF
--- a/cache-config/t3c-apply/t3c-apply.go
+++ b/cache-config/t3c-apply/t3c-apply.go
@@ -120,7 +120,9 @@ func main() {
 		syncdsUpdate, err = trops.CheckRevalidateState(false)
 		if err != nil || syncdsUpdate == torequest.UpdateTropsNotNeeded {
 			if err != nil {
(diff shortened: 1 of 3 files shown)


jrushford (approved):
Code looks good to me.  Ran cache-config integration tests locally and all tests pass.

PR #5895, opened 2021-05-27
Add the ability to CRD CDN Locks

srijeet0406:

- Documentation
- Traffic Control Go Client 
- Traffic Ops
- CI tests

Make sure the migrations run successfully.
Make sure all the API and unit tests pass.
Send requests to hit the POST, GET, DELETE and DELETE ADMIN endpoints for `cdn_locks` and make sure you see the expected results.

- master

- [x] This PR includes tests 


lib/go-tc/cdn_lock.go
```diff
@@ -0,0 +1,57 @@
...
+	"database/sql"
+
+	validation "github.com/go-ozzo/ozzo-validation"
+
+	"github.com/apache/trafficcontrol/lib/go-tc/tovalidate"
+	"github.com/apache/trafficcontrol/lib/go-util"
+)
+
+// CdnLock is a struct to store the details of a lock that a user wishes to acquire on a CDN.
+type CdnLock struct {
```
ocket8888, line 32: nit but `CdnLock` should be `CDNLock` since "CDN" is an initialism.

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+	WithObjs(t, []TCObj{Tenants, Roles, Users, CDNs}, func() {
+		CRDCdnLocks(t)
+		AdminCdnLocks(t)
+	})
+}
+
+func getCDNName(t *testing.T) string {
+	cdnResp, _, err := TOSession.GetCDNs(client.RequestOptions{})
+	if err != nil {
+		t.Fatalf("couldn't get CDNs: %v", err.Error())
```
ocket8888, line 39: nit: don't need `.Error()` when formatting with `%v`

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+	cdn := getCDNName(t)
+	// CREATE
+	var cdnLock tc.CdnLock
+	cdnLock.Cdn = cdn
+	cdnLock.UserName = TOSession.UserName
+	cdnLock.Message = util.StrPtr("snapping cdn")
+	cdnLock.Soft = util.BoolPtr(true)
+	cdnLockResp, _, err := TOSession.CreateCdnLock(cdnLock, client.RequestOptions{})
+	if err != nil {
+		t.Fatalf("couldn't create cdn lock: %v", err.Error())
```
ocket8888, line 57: same as above RE: `.Error()` with `%v`

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+		t.Errorf("expected Message %v, got %v", *cdnLock.Message, *cdnLockResp.Response.Message)
+	}
+	if cdnLockResp.Response.Soft != nil && *cdnLockResp.Response.Soft != *cdnLock.Soft {
+		t.Errorf("expected 'Soft' to be %v, got %v", *cdnLock.Soft, *cdnLockResp.Response.Soft)
+	}
+
+	// READ
+	cdnLocksReadResp, _, err := TOSession.GetCdnLocks(client.RequestOptions{})
+	if err != nil {
+		t.Fatalf("could not get CDN Locks: %v", err.Error())
```
ocket8888, line 78: same as above RE: `.Error()` with `%v`

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+		t.Errorf("expected Message %v, got %v", *cdnLock.Message, *cdnLocksReadResp.Response[0].Message)
+	}
+	if cdnLocksReadResp.Response[0].Soft != nil && *cdnLocksReadResp.Response[0].Soft != *cdnLock.Soft {
+		t.Errorf("expected 'Soft' to be %v, got %v", *cdnLock.Soft, *cdnLocksReadResp.Response[0].Soft)
+	}
+
+	// DELETE
+	_, reqInf, err := TOSession.DeleteCdnLocks(client.RequestOptions{QueryParameters: url.Values{"cdn": []string{cdnLock.Cdn}}})
+	if err != nil {
+		t.Fatalf("couldn't delete cdn lock, err: %v", err.Error())
```
ocket8888, line 102: same as above RE: `.Error()` with `%v`

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+	if reqInf.StatusCode != http.StatusOK {
+		t.Errorf("expected status code of 200, but got %d instead", reqInf.StatusCode)
+	}
+
+}
+
+func AdminCdnLocks(t *testing.T) {
+	resp, _, err := TOSession.GetTenants(client.RequestOptions{})
+	if err != nil {
+		t.Fatalf("could not GET tenants: %v", err.Error())
```
ocket8888, line 113: same as above RE: `.Error()` with `%v`

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	cdn := getCDNName(t)
+	// Create a lock for this user
+	_, _, err = userSession.CreateCdnLock(tc.CdnLock{
+		Cdn:     cdn,
+		Message: util.StrPtr("test lock"),
+		Soft:    util.BoolPtr(true),
+	}, client.RequestOptions{})
+	if err != nil {
+		t.Fatalf("couldn't create cdn lock: %v", err.Error())
```
ocket8888, line 149: same as above RE: `.Error()` with `%v`

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+		t.Fatalf("expected an error while deleting other user's lock, but got nothing")
+	}
+	if reqInf.StatusCode != http.StatusNotFound {
+		t.Fatalf("expected a 404 status code, but got %d instead", reqInf.StatusCode)
+	}
+
+	// Now try to delete another user's lock by hitting the admin DELETE endpoint for cdn_locks -> this should pass
+	_, reqInf, err = TOSession.AdminDeleteCdnLocks(client.RequestOptions{QueryParameters: url.Values{"cdn": []string{cdn}}})
+	if err != nil {
+		t.Fatalf("expected no error while deleting other user's lock using admin endpoint, but got %v", err.Error())
```
ocket8888, line 173: same as above RE: `.Error()` with `%v`

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if cdnLockResp.Response.Message == nil {
+		t.Errorf("expected a valid message, but got nothing")
+	}
+	if cdnLockResp.Response.Message != nil && *cdnLockResp.Response.Message != *cdnLock.Message {
+		t.Errorf("expected Message %v, got %v", *cdnLock.Message, *cdnLockResp.Response.Message)
+	}
+	if cdnLockResp.Response.Soft != nil && *cdnLockResp.Response.Soft != *cdnLock.Soft {
+		t.Errorf("expected 'Soft' to be %v, got %v", *cdnLock.Soft, *cdnLockResp.Response.Soft)
+	}
```
ocket8888, line 73: Shouldn't this also error if `cdnLockResp.Response.Soft == nil`

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+	if cdnLocksReadResp.Response[0].Message == nil {
+		t.Errorf("expected a valid message, but got nothing")
+	}
+	if cdnLocksReadResp.Response[0].Message != nil && *cdnLocksReadResp.Response[0].Message != *cdnLock.Message {
+		t.Errorf("expected Message %v, got %v", *cdnLock.Message, *cdnLocksReadResp.Response[0].Message)
+	}
+	if cdnLocksReadResp.Response[0].Soft != nil && *cdnLocksReadResp.Response[0].Soft != *cdnLock.Soft {
+		t.Errorf("expected 'Soft' to be %v, got %v", *cdnLock.Soft, *cdnLocksReadResp.Response[0].Soft)
+	}
```
ocket8888, line 97: same as above RE: soft being `nil`

lib/go-tc/cdn_lock.go
```diff
@@ -0,0 +1,57 @@
...
+	UserName    string    `json:"userName" db:"username"`
+	Cdn         string    `json:"cdn" db:"cdn"`
```
ocket8888, line 34: nit but `Cdn` should be `CDN` since "CDN" is an initialism

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,178 @@
...
+	_, _, err = TOSession.CreateUser(user1, client.RequestOptions{})
+	if err != nil {
+		t.Fatalf("could not create test user with username: %s", *user1.Username)
+	}
+	defer ForceDeleteTestUsersByUsernames(t, []string{"lock_user1"})
+
+	// Establish a session with the newly created non admin level user
+	userSession, _, err := client.LoginWithAgent(Config.TrafficOps.URL, *user1.Username, *user1.LocalPassword, true, "to-api-v4-client-tests", false, toReqTimeout)
+	if err != nil {
+		t.Fatalf("could not login with user lock_user1: %v", err.Error())
```
ocket8888, line 139: same as above RE: `.Error()` with `%v`

lib/go-tc/cdn_lock.go
```diff
@@ -0,0 +1,47 @@
...
+type CdnLocksGetResponse struct {
+	Response []CdnLock `json:"response"`
+	Alerts
+}
+
+type CdnLockDeleteResponse CdnLockCreateResponse
+
+func (c CdnLock) Validate(tx *sql.Tx) error {
+	return nil
+}
```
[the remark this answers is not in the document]
ocket8888: The file changed during my review, so I sort of said the same thing twice. Marking this as resolved so the discussion stays in one place.

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -133,6 +134,12 @@ func Routes(d ServerData) ([]Route, []RawRoute, http.Handler, error) {
 		 * 4.x API
 		 */
 
+		// CDN lock
+		{api.Version{Major: 4, Minor: 0}, http.MethodGet, `cdn_locks/?$`, cdn_lock.Read, auth.PrivLevelReadOnly, Authenticated, nil, 4134390561},
+		{api.Version{Major: 4, Minor: 0}, http.MethodPost, `cdn_locks/?$`, cdn_lock.Create, auth.PrivLevelOperations, Authenticated, nil, 4134390562},
+		{api.Version{Major: 4, Minor: 0}, http.MethodDelete, `cdn_locks/?$`, cdn_lock.Delete, auth.PrivLevelOperations, Authenticated, nil, 4134390564},
+		{api.Version{Major: 4, Minor: 0}, http.MethodDelete, `cdn_locks/admin/?$`, cdn_lock.AdminDelete, auth.PrivLevelAdmin, Authenticated, nil, 4134390565},
```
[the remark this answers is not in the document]
srijeet0406: So here's what I was thinking: If an `admin` level user hits the `DELETE` endpoint, it might be because the user wants to delete their own lock, or some other user's lock. That why I wanted to split them out into two different endpoints, one for deleting the user's own locks, and one for other deleting other people's locks (for an admin). However, I could add another query parameter to the DELETE endpoint, saying something like `self=true` to imply that the user wants to delete their own locks. Thoughts?

traffic_ops/v4-client/cdn_lock.go
```diff
@@ -0,0 +1,60 @@
...
+
+// apiAdminCDNLocks is the API version-relative path for the /cdn_locks/admin API endpoint.
+const apiAdminCDNLocks = "/cdn_locks/admin"
+
+// CreateCdnLock creates a CDN Lock.
+func (to *Session) CreateCdnLock(cdnLock tc.CdnLock, opts RequestOptions) (tc.CdnLockCreateResponse, toclientlib.ReqInf, error) {
+	var response tc.CdnLockCreateResponse
+	var alerts tc.Alerts
+	reqInf, err := to.post(apiCDNLocks, opts, cdnLock, &alerts)
+	response.Response = cdnLock
```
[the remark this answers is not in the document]
srijeet0406: Sorry that was a bad paste. Fixed now.

traffic_ops/traffic_ops_golang/routing/routes.go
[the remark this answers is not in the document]
ocket8888: The primary key for a CDN lock is the CDN which it locks, so it's impossible to for more than one user to have a lock on a CDN at a time. the `DELETE` method requires the name of the CDN to be passed in the `cdn` query string parameter according to the documentation. So the request already fully and uniquely defines the lock to be deleted, so passing `self=true` or `self=false` doesn't make much sense to me. The lock either belongs to you or it doesn't, you can't delete only "your" lock on a CDN because a CDN only has one lock. The only thing that needs to be decided is whether or not the requesting user has permission to perform the operation.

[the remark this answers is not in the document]
srijeet0406: Agreed, but that does leave us to the chance of an admin accidentally or unintentionally deleting someone else's lock. I know we could pass it off as user error, but I was just thinking of adding some more safety into it. What do you think?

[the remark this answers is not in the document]
ocket8888: Well that's what I mean, the admin user will say "I'm going to unlock this CDN" and then... that's exactly what happens. I don't think it's super likely to accidentally delete someone else's lock, because if someone else has it locked then you couldn't have locked it in the first place. So if you have no memory of locking the CDN, you should be able to immediately suppose that the lock was placed by someone else.

If we're worried about users doing legal things accidentally, then I'd say the proper way to handle that is to place a warning and/or confirmation dialog in the UI. But IMO we don't need a whole separate endpoint to describe a second way of checking for permissions to do a single operation. 1 operation <-> 1 method of one endpoint makes the most sense to me.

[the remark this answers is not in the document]
srijeet0406: Sounds good, just wanted to put it out there. 👍

docs/source/api/v4/cdn_locks.rst
```diff
@@ -0,0 +1,197 @@
...
+:Response Type: Object
+
+Request Structure
+-----------------
+.. table:: Request Query Parameters
+
+	+---------------+----------+-----------------------------------------------------------------------------------+
+	| Parameter     | Required | Description                                                                       |
+	+===============+==========+===================================================================================+
+	| cdn           | yes      | Delete the CDN lock for the CDN that has the name ``cdn``                    |
```
ocket8888, line 152: Table is malformed

ocket8888 (changes requested):

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
+)
+
+const readQuery = `SELECT username, cdn, message, soft, last_updated FROM cdn_lock`
+const insertQuery = `INSERT INTO cdn_lock (username, cdn, message, soft) VALUES (:username, :cdn, :message, :soft) RETURNING username, cdn, message, soft, last_updated`
+const deleteQuery = `DELETE FROM cdn_lock WHERE cdn=$1 AND username=$2 RETURNING username, cdn, message, soft, last_updated`
+const deleteAdminQuery = `DELETE FROM cdn_lock WHERE cdn=$1 RETURNING username, cdn, message, soft, last_updated`
+
+func Read(w http.ResponseWriter, r *http.Request) {
```
ocket8888, line 41: Should this endpoint support IMS?

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
```diff
@@ -0,0 +1,168 @@
...
+	if err := json.NewDecoder(r.Body).Decode(&cdnLock); err != nil {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, err, nil)
+		return
+	}
+	if cdnLock.CDN == "" {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, errors.New("field 'cdn' must be present"), nil)
+		return
+	}
+	// by default, always create soft (or shared) locks
+	if cdnLock.Soft == nil {
```
ocket8888, line 105: The blueprint listed the `soft` property as required and the database migration restricts it to `NOT NULL` - is there a reason to allow it to be null or undefined?

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
```diff
@@ -0,0 +1,168 @@
...
+		err = inf.Tx.Tx.QueryRow(deleteAdminQuery, cdn).Scan(&result.UserName, &result.CDN, &result.Message, &result.Soft, &result.LastUpdated)
+	} else {
+		err = inf.Tx.Tx.QueryRow(deleteQuery, cdn, inf.User.UserName).Scan(&result.UserName, &result.CDN, &result.Message, &result.Soft, &result.LastUpdated)
+	}
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			api.HandleErr(w, r, tx, http.StatusNotFound, fmt.Errorf("deleting cdn lock with cdn name %s: lock not found", cdn), nil)
+			return
+		}
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("deleting cdn lock with cdn name %s : %v", cdn, err.Error()))
```
ocket8888, line 161: formatting parameters for errors should look like `fmt.Errorf("%w", err)` rather than `fmt.Errorf("%v", err.Error())` - `%w` is a special formatting directive for errors that only `Errorf` provides, which "wraps" the error it formats. Wrapping errors is depended upon by things like `errors.Is`.

traffic_ops/v4-client/cdn_lock.go
```diff
@@ -0,0 +1,48 @@
...
+	reqInf, err := to.get(apiCDNLocks, opts, &data)
+	return data, reqInf, err
+}
+
+// DeleteCdnLocks deletes the CDN lock of a particular(requesting) user.
+func (to *Session) DeleteCdnLocks(opts RequestOptions) (tc.CdnLockDeleteResponse, toclientlib.ReqInf, error) {
+	var data tc.CdnLockDeleteResponse
+	reqInf, err := to.del(apiCDNLocks, opts, &data)
+	return data, reqInf, err
+}
```
ocket8888, line 48: nit but CDN should be capitalized in these exported symbols, since it's an initialism.

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		t.Fatalf("couldn't create cdn lock: %v", err)
+	}
+
+	// Non admin user trying to delete another user's lock -> this should fail
+	_, reqInf, err := userSession2.DeleteCdnLocks(client.RequestOptions{QueryParameters: url.Values{"cdn": []string{cdn}}})
+	if err == nil {
+		t.Fatalf("expected error when a non admin user tries to delete another user's lock, but got nothing")
+	}
+	if reqInf.StatusCode != http.StatusNotFound {
```
ocket8888, line 189: Shouldn't this be a `403 Forbidden` if the user doesn't have permission to delete the lock?

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
srijeet0406: So that required/ not null field was intended for the database storage. While creating a lock, a user is free to leave out the `soft` property from the request structure, and by default we would always create `soft` locks. Its better of the user has to specify lesser details. Does that sound ok to you?

traffic_ops/testing/api/v4/cdn_locks_test.go
srijeet0406: Well, for any non admin user, it is assumed that the lock that the user is trying to delete will be their own lock. And if a non admin user(A) is trying to delete someone else's(B) lock, the DB query that gets constructed will have the username set to `A`. Since this lock doesn't exist, the user gets a 404.

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
srijeet0406: Ideally, it could, but I dont think the benefit would be too much. The number of locks can only be less than or equal to the number of CDNs (worst case). Depending on how often we make the call to `GET` all the locks (which I dont think will be a lot), the bandwidth advantage isn't very high. Additionally, we'd need to add another database migration for it.
I can add it if you want.

traffic_ops/testing/api/v4/cdn_locks_test.go
ocket8888: That sounds like it could be confusing. I can `GET` something with 200 OK but then try to `DELETE` it and the API tells me it doesn't exist. But I can still `GET` it. A client-side error should help to inform the client how to rectify the situation.

A 404 indicates, it seems to me, that you can't request the path you're requesting, so the solution is to change the path. But that won't help in this case, because the actual problem is that since the lock doesn't belong to me and I'm not an administrator, I don't have permission to do the thing I'm trying to do, which is what a 403 indicates IMO.

srijeet0406: Ok I was just thinking of 404 in another way, that the lock doesn't exist. But what you said makes sense to me. I'll change it to return a 403.

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
ocket8888: [The database storage section lists it as `not null`](https://github.com/apache/trafficcontrol/blob/master/blueprints/to-locks.md#data-model--database-impact), but also [the REST API section labels it as "required"](https://github.com/apache/trafficcontrol/blob/master/blueprints/to-locks.md#rest-api-impact). Personally, I think it's better to be explicit about typing rather than implicit; it's a lot harder to make a mistake if you **have** to specify the value yourself.

Honestly, if it were up to me entirely, I would eliminate optional properties entirely. But that would be a major overhaul since Go is incapable of distinguishing between `null` and `undefined` unless you jump through some `json.RawMessage` hoops.

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
ocket8888: idk how useful it could be, I just thought we wanted everything to do that from now on. Idc about adding it, it's fine.

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
srijeet0406: Fixed now.

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
```diff
@@ -0,0 +1,172 @@
+package cdn_lock
```
ocket8888, line 1: Package should have a GoDoc

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
+)
+
+const readQuery = `SELECT username, cdn, message, soft, last_updated FROM cdn_lock`
+const insertQuery = `INSERT INTO cdn_lock (username, cdn, message, soft) VALUES (:username, :cdn, :message, :soft) RETURNING username, cdn, message, soft, last_updated`
+const deleteQuery = `DELETE FROM cdn_lock WHERE cdn=$1 AND username=$2 RETURNING username, cdn, message, soft, last_updated`
+const deleteAdminQuery = `DELETE FROM cdn_lock WHERE cdn=$1 RETURNING username, cdn, message, soft, last_updated`
+
+func Read(w http.ResponseWriter, r *http.Request) {
```
ocket8888, line 41: GoDoc?

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
```diff
@@ -0,0 +1,172 @@
...
+		return
+	}
+	alerts := tc.CreateAlerts(tc.SuccessLevel, "CDN lock acquired!")
+	api.WriteAlertsObj(w, r, http.StatusCreated, alerts, cdnLock)
+
+	changeLogMsg := fmt.Sprintf("USER: %s, CDN: %s, ACTION: Lock Acquired", inf.User.UserName, cdnLock.CDN)
+	api.CreateChangeLogRawTx(api.ApiChange, changeLogMsg, inf.User, tx)
+}
+
+func Delete(w http.ResponseWriter, r *http.Request) {
```
ocket8888, line 136: GoDoc?

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
```diff
@@ -1,3 +1,4 @@
+// cdn_lock package contains the CRD methods related to CDNLocks
```
ocket8888, line 1: Package GoDocs should be complete sentences that start with `Package {{package name}}`.

ocket8888 (approved):

PR #5897, opened 2021-05-28
Fix t3c rpm to create var/lib dir

rob05c:
Fixes the t3c rpm to create the var/lib dir that t3c-apply needs when running.

This is a minor issue, it's only used to write the 'status' file. When that fails, nothing else fails, everything else seems to work fine. I'm not sure that file is even needed for anything, it's something ort.pl did so t3c does as well.

Includes tests.
No docs, no changelog, no interface change and issue isn't in a release.

- Traffic Ops ORT

Run tests. Install trafficcontrol-cache-config RPM, verify /var/lib/trafficcontrol-cache-config dir is created, run t3c-apply, verify /var/lib/trafficcontrol-cache-config/status is created.

- master

## Additional Information


ezelkow1 (approved):
Tested with older version vs. newer version, saw dir created with newer version while the older did not

PR #5904, opened 2021-06-02
Add change log count to user response payload and username as query param in logs API

### LINKED ISSUE #5451: Add change log count to user response payload
## I'm submitting a ...

-  Traffic Ops
-  Traffic Portal

## Current behavior:
I don't believe there is a way currently via TP and/or the TO API to see how many change log entries have been created by a particular user. IMO this information would be helpful when evaluating user activity along with #5412 

## New behavior:
Do one of 2 things or maybe both:
1. Add changeLogCount to the response payload of the user TO api endpoints (this would require a join to the log table)
2. add query param support such as GET /logs?username=

## Anything else:
If this is added to the api, then it needs to be added to TP as well. I.e. 
- https://tp.domain.tld/#!/users should show the changeLogCount in the grid
- https://tp.domain.tld/#!/users/{{id}} should have a new button called 'view change logs' or something.

rimashah25:

- [x] This PR fixes #5451 

- Documentation
- Traffic Ops
- Traffic Portal
-  CDN in a Box

1. Use GET curl calls for **log** api for any permutation of limit, days, and username
```
https://<TO-API host:port>/api/4.0/logs?days=10&limit=3
OR 
https://<TO-API host:port>/api/4.0/logs?username=admin&days=10
```

2. Login to TP, and ensure that change log section on dashboard and change log widget image shows summary of logs

3. Use GET curl call to **user** api (https://<TO-API host:port>/api/4.0/users), and one should see a new field **chaneLogCount** associated with each user.
image

4. Login to TP, go to Users section, click on select table column image, and you should see the number of change logs associated with each user.
image

image

5. Login to TP, go to Users section, and then click on a user. `View Change Log` button should exists. Click the button to see the logs associated with the user (user should be re-routed to change log page)

6. Run API/v4 tests for logs

- [x] This PR includes tests


docs/source/api/v4/logs.rst
```diff
@@ -32,18 +32,20 @@ Request Structure
...
-	| limit | no       | The number of records to which to limit the response |
-	+-------+----------+------------------------------------------------------+
+	+----------+----------+------------------------------------------------------+
+	| Name     | Required | Description                                          |
+	+==========+==========+======================================================+
+	| days     | no       | An integer number of days of change logs to return   |
+	+----------+----------+------------------------------------------------------+
+	| limit    | no       | The number of records to which to limit the response |
+	+----------+----------+------------------------------------------------------+
+	| username | no       | A name to which to limit the response too 					 |
```
ocket8888, line 42: Malformed table

docs/source/api/v4/logs.rst
```diff
@@ -92,5 +95,8 @@ Response Structure
 			"user": "admin",
 			"id": 443,
 			"message": "1 delivery services were assigned to test"
-		}
-	]}
+		}],
+		"summary": {
+      "count": 20
+    }
+	}
```
ocket8888, line 102: You're using space-based indentation here - documentation should use tabs

docs/source/api/v4/logs.rst
```diff
@@ -57,6 +59,7 @@ Response Structure
 :message:     Log detail about what occurred
 :ticketNum:   Optional field to cross reference with any bug tracking systems
 :user:        Name of the user who made the change
+:summary.count: Associated changelog entries for a given query param
```
ocket8888, line 62: `summary.count` is not a property of the response object. Check out the `/servers` endpoint documentation to see an example of how that can be unambiguously handled.

docs/source/api/v4/users.rst
```diff
@@ -90,6 +90,7 @@ Response Structure
 :tenantId:         The integral, unique identifier of the tenant to which this user belongs
 :uid:              A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX user ID of the user - now it is always ``null``
 :username:         The user's username
+:changeLogCount:   The number of change log entries created by a particular user
```
ocket8888, line 93: I think this would be clearer if you replace "a particular" with "the"

traffic_ops/testing/api/v4/logs_test.go
```diff
@@ -46,3 +47,15 @@ func GetTestLogsByLimit(t *testing.T) {
...
+	opts := client.NewRequestOptions()
+	opts.QueryParameters.Set("username", "admin")
+	toLogs, _, err := TOSession.GetLogs(opts)
+	if err != nil {
+		t.Errorf("error getting logs: %v - alerts: %+v", err, toLogs.Alerts)
+	}
+	if len(toLogs.Response) <= 0 {
+		t.Fatalf("Get logs by username: incorrect number of logs returned (%d)", len(toLogs.Response))
+	}
+}
```
ocket8888, line 61: Should this verify that only change logs for the requested user are returned?

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -121,26 +122,50 @@ func getLastSeenCookie(r *http.Request) (time.Time, bool) {
...
+	}
+
+	timeInterval := fmt.Sprintf("l.last_updated > now() - INTERVAL '%v' DAY", days)
+	if _, ok := inf.Params["username"]; ok {
+		where = where + " AND " + timeInterval
+	} else {
+		where = "\nWHERE " + timeInterval
+	}
+	if orderBy == "" {
+		orderBy = orderBy + "\nORDER BY l.last_updated DESC"
```
ocket8888, line 148: This should be done by setting the `inf.Params` value at `"orderby"` if it isn't already set. That way, `dbhelpers.BuildWhereAndOrderByAndPagination` will build an `ORDER BY` clause for you, if needed.

Also, since this endpoint uses `dbhelpers.BuildWhereAndOrderByAndPagination` it supports all sorting pagination controls that other endpoints support (e.g. `sortOrder`), but that isn't documented.

docs/source/api/v4/logs.rst
rimashah25: I was using tabs but somehow it doesn't like it. Anyways I updated the docs by copying what was seen in /servers.rst

traffic_ops/testing/api/v4/logs_test.go
rimashah25: Yes

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -121,26 +122,51 @@ func getLastSeenCookie(r *http.Request) (time.Time, bool) {
...
 		l := tc.Log{}
 		if err = rows.Scan(&l.ID, &l.Level, &l.Message, &l.User, &l.TicketNum, &l.LastUpdated); err != nil {
-			return nil, errors.New("scanning logs: " + err.Error())
+			return nil, count, errors.New("scanning logs: " + err.Error())
 		}
+		count += 1
 		ls = append(ls, l)
 	}
-	return ls, nil
+	return ls, count, nil
```
ocket8888, line 169: So regarding `summary`'s `count` property, the docs define it to mean:

> _"`count` contains an unsigned integer that defines the total number of results that could possibly be returned given the non-pagination query parameters supplied by the client."_

... which means you can't just count the things you're returning to the client. You have to make a separate query that doesn't include any pagination. Typically, you can build this query by doing

```go
const countQuery = `
SELECT COUNT(*)
FROM logs
`

func get(args...) {
	// normal API stuff, getting the actual data etc.

	// this 'where' is the one that comes from the dbhelpers function from above
	query = countQuery + where
	var count uint64
	ns, err := inf.Tx.PrepareNamed(query)
	if err != nil {
		// handle the error
	}
	// 'queryValues' are, again, as returned by the dbhelpers function
	err = inf.Tx.NamedStmt(ns).QueryRow(queryValues).Scan(&count)
	if err != nil {
		// handle the error
	}
	// 'count' now holds the appropriate log count
	// do whatever else you need to do, write the response etc.
}
```


... but if that sounds like too much work to you, then don't worry about it, you don't need to add a `summary` to the endpoint, that's a totally optional extension for API endpoints.

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -121,26 +122,51 @@ func getLastSeenCookie(r *http.Request) (time.Time, bool) {
...
+	if _, ok := inf.Params["orderby"]; !ok {
+		inf.Params["orderby"] = "last_updated"
+	}
+
+	queryParamsToQueryCols := map[string]dbhelpers.WhereColumnInfo{
+		"username":     dbhelpers.WhereColumnInfo{Column: "u.username", Checker: nil},
+		"last_updated": dbhelpers.WhereColumnInfo{Column: "l.last_updated", Checker: nil},
+	}
+	where, orderBy, pagination, queryValues, errs :=
+		dbhelpers.BuildWhereAndOrderByAndPagination(inf.Params, queryParamsToQueryCols)
```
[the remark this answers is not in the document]
ocket8888: One advantage of doing the second thing is that you can make some breaking API changes, like removing the annoying default values for `days` and `limit`.

traffic_portal/app/src/common/modules/form/user/FormUserController.js
```diff
@@ -37,6 +37,16 @@ var FormUserController = function(user, $scope, $location, formUtils, stringUtil
             });
     };
 
+    $scope.changeLogs = [];
+
+    $scope.getChangeLogs = function(userName) {
+        changeLogService.getChangeLogs({ username: userName })
+            .then(function(response) {
+                $scope.changeLogs = response;
+                locationUtils.navigateToPath('/change-logs?username='+userName);
```
ocket8888, line 46: `userName` should be sanitized with [`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent).

traffic_portal/app/src/common/modules/form/user/form.user.tpl.html
```diff
@@ -25,6 +25,7 @@
             <li class="active">{{userName}}</li>
         </ol>
         <div class="pull-right" role="group" ng-show="!settings.isNew">
+            <button name="viewChangelog" class="btn btn-primary" title="Change Log" ng-click="getChangeLogs(user.username)">View Change Log</button>
```
ocket8888, line 28: can't this just be a link to `/change-logs?username={{username()}}` instead of a button that does navigation? (`$scope.username` will need to be a function that returns the sanitized username with `encodeURIComponent` for that example to work)

traffic_ops/testing/api/v4/logs_test.go
```diff
@@ -46,3 +47,20 @@ func GetTestLogsByLimit(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("error getting logs: %v - alerts: %+v", err, toLogs.Alerts)
+	}
+	for _, user := range toLogs.Response {
+		if *user.User != "admin" {
+			t.Fatalf("incorrect username seen in logs, expected: `admin`, got: %v", *user.User)
+		}
+	}
+	if len(toLogs.Response) <= 0 {
+		t.Fatalf("Get logs by username: incorrect number of logs returned (%d)", len(toLogs.Response))
```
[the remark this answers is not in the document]
rimashah25: So, I am running the `go test` for logs only and without doing anything additional, I am able to see 10 logs for the user since the first statement of the TestLogs() is:
```
WithObjs(t, []TCObj{Roles, Tenants, Users}, func() { // Objs added to create logs when this test is run alone
```

So, the probability that no logs will be present is 0.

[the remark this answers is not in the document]
ocket8888: :+1:

traffic_portal/app/src/common/modules/form/user/form.user.tpl.html
rimashah25: I can but what's wrong in how it is right now?

traffic_portal/app/src/common/modules/form/user/FormUserController.js
```diff
@@ -37,6 +37,16 @@ var FormUserController = function(user, $scope, $location, formUtils, stringUtil
             });
     };
 
+    $scope.changeLogs = [];
+
+    $scope.getChangeLogs = function(userName) {
+        changeLogService.getChangeLogs({ username: userName })
```
shamrickus, line 43: I believe this call to `getChangeLogs` is un-needed. The first thing the [change-logs page does](https://github.com/apache/trafficcontrol/blob/5b6db34a8a2c930aa56ec73249161f26cd876328/traffic_portal/app/src/modules/private/changeLogs/list/index.js#L35) is run a `getChangeLogs`.

shamrickus (changes requested):

traffic_portal/app/src/common/modules/form/user/FormUserController.js
rimashah25: ok but why?

traffic_ops/traffic_ops_golang/logs/log.go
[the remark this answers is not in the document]
rimashah25: I am choosing the first option and thanks for the helpful suggestion

traffic_portal/app/src/common/modules/form/user/FormUserController.js
rimashah25: oh.. so change the html button to a link instead of a call?

traffic_ops/traffic_ops_golang/logs/log.go
rimashah25: Given this; https://github.com/apache/trafficcontrol/issues/5451#issuecomment-764983956, above suggested changed doesn't make sense.

traffic_portal/app/src/common/modules/form/user/FormUserController.js
shamrickus: That's what @ocket8888 is requesting. If it were to stay a call  the function would just need to be `locationUtils.navigateToPath('/change-logs?username='+userName)`.

traffic_portal/app/src/common/modules/form/user/FormUserController.js
ocket8888: Because if the username contains <kbd>;</kbd>, <kbd>,</kbd>, <kbd>/</kbd>, <kbd>?</kbd>, <kbd>:</kbd>, <kbd>@</kbd>, <kbd>&</kbd>, <kbd>=</kbd>, <kbd>+</kbd>, <kbd>$</kbd>, or <kbd> </kbd> the URL will be parsed incorrectly - or if the username contains any UTF8 codepoints that begin or end with bytes that could be misinterpreted as those characters. Reserved characters and non-ascii code-points *must* be properly encoded in URLs, **especially** when they come from user input, which usernames do. It's a possible attack vector.

traffic_portal/app/src/common/modules/form/user/form.user.tpl.html
ocket8888: Because click handlers should never perform navigation if possible. I can't middle-click a button to open it in a new tab, or right-click -> copy link URL, or any of the other things I expect from links. It also messes with assistive technologies; if you rely on a screen reader to find hyperlinks between pages, this can be annoying to dig through.

traffic_ops/traffic_ops_golang/logs/log.go
ocket8888: that comment was about a possible alternative to this, but apparently there's enough value in it that we're doing it, so what I said on #5451 no longer matters.

traffic_portal/app/src/common/modules/form/user/form.user.tpl.html
rimashah25: Understood.

traffic_portal/app/src/common/modules/form/user/FormUserController.js
rimashah25: Got it..since I changed the button to a link, we no longer need this function and its sanitization.

traffic_portal/app/src/common/modules/form/user/form.user.tpl.html
```diff
@@ -25,6 +25,7 @@
             <li class="active">{{userName}}</li>
         </ol>
         <div class="pull-right" role="group" ng-show="!settings.isNew">
+            <a class="btn btn-primary" ng-href="/#!/change-logs?username={{userName}}">View Change Log</a>
```
ocket8888, line 28: unfortunately, for whatever reason the ngHref directive won't sanitize for you, so the value here should be inserted via a function that produces a sanitized output from the username using `encodeURIComponent`. Alternatively, you could create a simple AngularJS filter for query string parameters named something like `encodeURIComponent`, for example, then apply it here. That'd look like this in a file in `traffic_portal/app/src/common/filters/`:
```javascript
/**
 * This is a factory function for an AngularJS filter meant to allow easy
 * sanitization of interpolated input into URL query strings.
 * 
 * @example
 * 
 * <a ng-href="https://example.test/something?foo={{someString | encodeURIComponent}}">
 *   A safe link
 * </a>
 * 
 * @returns {(input: string) => string} A filter that sanitizes input text for safe insertion into URL query strings.
 */
 function EncodeURIComponentFilter() {
	return function(input) {
		return encodeURIComponent(input);
	};
};

module.exports = EncodeURIComponentFilter;
```
Then, just add a line to `traffic_portal/app/src/common/filters/index.js` like the other filters have, something like
```javascript
// in the @example above this is the name I used, but it can be anything
.filter("encodeURIComponent", require("./EncodeURIComponentFilter.js")),
```
(assuming the filter was written to `EncodeURIComponentFilter.js` - but it can be any file name).
Finally, just like the example, pass the data through the pipe in the `ng-href`:
```html
<a class="btn btn-primary" ng-href="/#!/change-logs?username={{userName | encodeURIComponent}}">View Change Log</a>
```

This is roughly one more line of work than just encoding it in a function in the users form controller, but it would allow us to re-use it everywhere it's needed without any extra work. Doesn't matter which you choose, though.

rimashah25: Copy!

traffic_portal/app/src/common/filters/EncodeURIComponentFilter.js
```diff
@@ -0,0 +1,39 @@
...
+ * @example
+ *
+ * <a ng-href="https://example.test/something?foo={{someString | encodeURIComponent}}">
+ *   A safe link
+ * </a>
+ *
+ * @returns {(input: string) => string} A filter that sanitizes input text for safe insertion into URL query strings.
+ */
+
+var EncodeURIComponentFilter = function() {
```
ocket8888, line 33: There shouldn't be a blank line between a JSDoc and the thing it documents.

I would feel worse about asking for such a small change if our CI was working, but fortunately (?) it isn't.

Also nit but I don't think you need to do `var Thing = function ...`, I'd be shocked if `function Thing ...` didn't work.

rimashah25: Ok.. And other filters have `var thing = function`, so was keeping up with codebase. If we are changing in one, we should follow suit with other 3 filters too.

ocket8888: That's fine. It doesn't matter, functionally.

rimashah25: kewl.. all changes pushed.

ocket8888 (commented):
This all works as expected, and the only failing tests are the ones that'll be fixed by #5957 .

However, this appears to be adding a field to `/users` endpoint across _all_ API versions. The behavior should be as-documented: restricted to APIv4 only

mitchell852:
@rimashah25 - very nice feature. i think this will help a lot with user cleanup. couple of very minor comments:

image

image

Probably want to change that to "Change Log Count" because it's a label.

lib/go-tc/users.go
```diff
@@ -108,6 +107,12 @@ type User struct {
 	commonUserFields
 }
 
+// UserV40 contains ChangeLogCount field
```
ocket8888, line 110: Nit: GoDoc missing punctuation

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -204,7 +206,17 @@ func (this *TOUser) Read(h http.Header, useIMS bool) ([]interface{}, error, erro
 
 	groupBy := "\n" + `GROUP BY u.id, r.name, t.name`
 	orderBy = groupBy + orderBy
-	query := this.SelectQuery() + where + orderBy + pagination
+
+	version := inf.Version
+	if version == nil {
+		return nil, nil, fmt.Errorf("TOUsers.Read called with invalid API version: %d.%d", version.Major, version.Minor), http.StatusInternalServerError, nil
```
ocket8888, line 212: Since `version` is nil in here, this line will always segfault trying to dereference `version.Major` and `version.Minor`

rimashah25: oops

ocket8888 (approved):

PR #5908, opened 2021-06-02
Fix t3c parent.config gen for EDGE/MID+ Types

rob05c:
Fixed t3c to generate topology parents correctly for parents with the Type MID+ or EDGE+ versus just the literal. Naming cache types to not be exactly 'EDGE' or 'MID' is still discouraged and not guaranteed to work, but it's unfortutely somewhat common, so this fixes it in one particular case.

Includes tests.
Includes changelog.
No docs, no interface change.

Labelling "bug" even though the behavior it "fixes" isn't guaranteed, because it's a non-uncommon user behavior that it does fix. 

Long-term, cache config generation for Topologies needs to entirely remove "Edge" and "Mid" terminology, though it will still need some way to distinguish Caches from non-cache Servers (unless TO is made more Normal).

- Traffic Ops ORT

Run lib/go-atscfg unit tests. Create a Topology with a parent server of type "MID_FOO," assign to a DS, generate config for an child server, verify the MID_FOO type parent is in the child's parent.config.

- master
- 

## Additional Information


ezelkow1 (approved):
Tested before and after this change while having topology servers set to MID+ naming. Previously no caches were found. With this change and having at least 1 cache set to reported in the parent group then the topology was correctly set in the parent.config with the appropriate parent cache and topology comment set

Issue #5825, opened 2021-05-07
Add regex_revalidate support for MISS


traeak:
## I'm submitting a ...

-  new feature / enhancement request

## Traffic Control components affected ...

-  Traffic Ops
-  Traffic Ops ORT
-  Traffic Portal
-  CI tests

## Current behavior:

No ability to force asset refetch, only refresh.

## New behavior:

A new feature will be added to the ATS regex_revalidate plugin to allow refetch in addition to just refresh.
This is based on ATS PR: https://github.com/apache/trafficserver/pull/6455

This new addition is supported to the regex_revalidate.conf file:

```
http://origin\.tld/images/foo\.jpg 1447483647 MISS
```

the keyword "STALE" may be also be used or left off for the normal regex_revalidate behavior (the next 2 are equivalent):

```
http://origin\.tld/images/foo\.jpg 1447483647
http://origin\.tld/images/foo\.jpg 1447483647 STALE
```

## Anything else:

If possible there should be a safety around this.  Worst case behavior of this could result in the refetch of an entire library and not just refreshes.  Perhaps rules can be made as stale and then "promoted" to MISS by an operator.  Any rule that is switched between MISS and STALE is reset and treated as fresh by the regex_revalidate plugin.

TaylorCFrey:
I'll take this one!

TaylorCFrey:
This work was completed as part of PR: https://github.com/apache/trafficcontrol/pull/6118

Good to close!

PR #5910, opened 2021-06-03
Blueprint/refetch

TaylorCFrey:

This is a blueprint proposal to solve a few issues, however the primary one is:
- [X] Is related to #5825 

This blueprint has the potential to impact:

- CDN in a Box
- Documentation
- Traffic Control Clients (Go)
- Traffic Ops
- Traffic Ops ORT
- Traffic Portal
- CI tests

No verification necessary for the blueprint. Just review the file `blueprints/refetch_invalidation.md`

### DIFF
--- a/blueprints/refetch-invalidation.md
+++ b/blueprints/refetch-invalidation.md
@@ -20,7 +20,9 @@ under the License.
 
 ## Problem Description
 
-Currently, within ATC, there is a concept of Invalidation Jobs. These Invalidation Jobs give a user the ability to queue an invalidation for a resource, primarily based on regular expressions. The invalidation is gathered and treated as though there was a cache **STALE**, allowing the CDN to query the origin server to **REFRESH** the resource. However, should the cache policy still be incorrect or misconfigured, the resource could be updated on the origin server, but the cache will still receive a 304 - Not Modified HTTP status response.
+Currently, within ATC, there is a concept of Invalidation Jobs. These Invalidation Jobs give a user the ability to queue an invalidation for a resource, primarily based on regular expressions. The invalidation is gathered and treated as though there was a cache **STALE**, allowing the CDN to query the origin server to **REFRESH** the resource. However, should the cache policy still be incorrect or misconfigured, the resource could be updated on the origin server, but the CDN will still receive a 304 - Not Modified HTTP status response.
+
+It should be noted that this problem originally arose from a misconfigured origin, however this applies to any parent cache within the topology.
 
 ## Proposed Change
 
@@ -46,7 +48,9 @@ Both the API and the database schema will likely be updated, which in turn will 
 
 #### REST API Impact
 
-No new endpoints will be required. However the current invalidation job will now include an optional field during `Create`. Invalidation jobs are added by submitting a POST to the jobs endpoint. 
+No new endpoints will be required. However the current invalidation job will now include an optional field during `Create`. Invalidation jobs are added by submitting a POST to the jobs endpoint.
+
+Globally, a new parameter will be added (recommend something akin to `refetch_enabled`, which defaults to **false**), that will be used to validate any **POSTS** calls to the _/jobs_ endpoint. This will be an initial check done to perform that the CDN is configured to process and return `refetch` jobs. If a `refetch` is submitted and the value of the parameter is **true**, the **POST** will succeed. If the value is `refetch` and the parameter is set to **false**, the **POST** will fail with a 400 - Bad Request. If the value is `refresh`, the parameter need not be checked at all and the **POST** will succeed. 
 
 **POST** /api/4.0/jobs
 
@@ -89,16 +93,15 @@ Body:
 }
 ```
 
-This struct now contains the `InvalidationType *string` field.
+This struct now contains the `InvalidationType *string` field. Additionally, the `DeliveryService` and `TTL` are no longer empty interfaces. Also `DeliveryService` and `Regex` are no longer optional fields. 
+
 ```go
 type InvalidationJobInput struct {
-	DeliveryService  *interface{} `json:"deliveryService"`
-	Regex            *string      `json:"regex"`
-	InvalidationType *string      `json:"invalidationType,omitempty"`
-	StartTime        *Time        `json:"startTime"`
-	TTL              *interface{} `json:"ttl"`
-	dsid             *uint
-	ttl              *time.Duration
+	DeliveryService  string  `json:"deliveryService"`
+	Regex            string  `json:"regex"`
+	InvalidationType *string `json:"invalidationType,omitempty"`
+	StartTime        *Time   `json:"startTime"`
+	TTL              *uint   `json:"ttl"`
 }
 ```
 
--- a/blueprints/refetch-invalidation.md
+++ b/blueprints/refetch-invalidation.md
@@ -99,7 +99,7 @@ This struct now contains the `InvalidationType *string` field. Additionally, the
 type InvalidationJobInput struct {
 	DeliveryService  string  `json:"deliveryService"`
 	Regex            string  `json:"regex"`
-	InvalidationType *string `json:"invalidationType,omitempty"`
+	InvalidationType *string `json:"invalidationType"`
 	StartTime        *Time   `json:"startTime"`
 	TTL              *uint   `json:"ttl"`
 }
--- a/blueprints/refetch-invalidation.md
+++ b/blueprints/refetch-invalidation.md
@@ -64,11 +64,11 @@ Which is mapped to a go `struct` in the `go-tc` lib.
 ```go
 type InvalidationJobInput struct {
 	DeliveryService *interface{} `json:"deliveryService"`
-	Regex *string `json:"regex"`
-	StartTime *Time `json:"startTime"`
-	TTL *interface{} `json:"ttl"`
-	dsid *uint
-	ttl  *time.Duration
+	Regex           *string      `json:"regex"`
+	StartTime       *Time        `json:"startTime"`
+	TTL             *interface{} `json:"ttl"`
+	dsid            *uint
+	ttl             *time.Duration
 }
 ```
 
@@ -90,13 +90,13 @@ Body:
 This struct now contains the `InvalidationType *string` field.
 ```go
 type InvalidationJobInput struct {
-	DeliveryService *interface{} `json:"deliveryService"`
-	Regex *string `json:"regex"`
-	InvalidationType *string `json:"invalidationType,omitempty"`
-	StartTime *Time `json:"startTime"`
-	TTL *interface{} `json:"ttl"`
-	dsid *uint
-	ttl  *time.Duration
+	DeliveryService  *interface{} `json:"deliveryService"`
+	Regex            *string      `json:"regex"`
+	InvalidationType *string      `json:"invalidationType,omitempty"`
+	StartTime        *Time        `json:"startTime"`
+	TTL              *interface{} `json:"ttl"`
+	dsid             *uint
+	ttl              *time.Duration
 }
 ```
 
--- a/blueprints/refetch-invalidation.md
+++ b/blueprints/refetch-invalidation.md
@@ -1,0 +1,240 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+# Add REFETCH capability option for Content Invalidation
+
+## Problem Description
+
+Currently, within ATC, there is a concept of Invalidation Jobs. These Invalidation Jobs give a user the ability to queue an invalidation for a resource, primarily based on regular expressions. The invalidation is gathered and treated as though there was a cache **STALE**, allowing the cache to query the origin server to **REFRESH** the resource. However, should the cache policy still be incorrect or misconfigured, the resource could be updated on the origin server, but the cache will still receive a 304 - Not Modified HTTP status response.
+
+## Proposed Change
+
+ To address this potential conflict, a proposal to add **REFETCH** as an option for Invalidation Jobs. This will then be treated by caches as a **MISS** (rather than a **STALE**), thusly allowing the cache to retrieve the resource regardless of cache policies. The original **REFRESH**/**STALE** will be the default option, where **REFETCH**/**MISS** will be the addition.
+
+### Traffic Portal Impact
+
+##### Create and Update
+Traffic Portal will need to update the Invalidation Job to account for the different options. When creating an Invalidation Job both options will need to be present (Perhaps a radio button? Default will be the original **REFRESH**).
+
+Tooltips should be added to ensure an understanding of this feature at a high level.
+
+##### Read
+When displaying the information, the **Invalidation Requests** table current shows the `Parameters` field, so the display will be straight forward with no manipulation.
+
+However, we derived and calculate the expiration field based on the TTL. This code will need to be modified to account for the additional information contained in the `Parameters` field.
+
+### Traffic Ops Impact
+
+Both the API and the database schema will likely be updated, which in turn will result in changes downstream (such as T3C/ORT, clients) as well.
+
+#### REST API Impact
+
+No new endpoints will be required. However the current invalidation job will now include an optional field during `Create`. Invalidation jobs are added by submitting a POST to the jobs endpoint. 
+
+**POST** /api/4.0/jobs
+
+##### Current Request
+
+Body:
+```
+{"startTime":"2021-06-02T15:23:21.348Z","deliveryService":11,"regex":"/path/.*\\.jpeg","ttl":24}
+```
+
+Which is mapped to a go `struct` in the `go-tc` lib.
+```go
+type InvalidationJobInput struct {
+	DeliveryService *interface{} `json:"deliveryService"`
+	Regex *string `json:"regex"`
+	StartTime *Time `json:"startTime"`
+	TTL *interface{} `json:"ttl"`
+	dsid *uint
+	ttl  *time.Duration
+}
+```
+
+##### Proposed
+
+Add an "InvalidationType" to signify a specific type of invalidation request. The InvalidationType is an optional field and will not break backwards compatibility with previous API versions. If the field is included, it _must_ be either "refetch" or "refresh".
+
+Body:
+```
+{"startTime":"2021-06-02T15:23:21.348Z","deliveryService":11,"regex":"/path/.*\\.jpeg","ttl":24,"invalidationType":"refresh"}
+```
+
+This struct now contains the `InvalidationType *string` field.
+```go
+type InvalidationJobInput struct {
+	DeliveryService *interface{} `json:"deliveryService"`
+	Regex *string `json:"regex"`
+	InvalidationType *string `json:"invalidationType,omitempty"`
+	StartTime *Time `json:"startTime"`
+	TTL *interface{} `json:"ttl"`
+	dsid *uint
+	ttl  *time.Duration
+}
+```
+
+##### Parsing the value
+
+Since the field is optional and existing functionality only signifies a **REFRESH**/**STALE** capability, if the field is omitted, empty, malformed, or in any way _not_ `refetch` then it will be treated as `refresh`.
+
+##### Response
+
+The response will be modified, then, to return this new value as well.
+
+Sample current response:
+```
+{"alerts":[{"text":"Invalidation request created for http://amc-linear-origin.local.tld/path/.*\\.jpeg, start:2021-06-02 15:23:21.348 +0000 UTC end 2021-06-03 15:23:21.348 +0000 UTC","level":"success"}],"response":{"assetUrl":"http://amc-linear-origin.local.tld/path/.*\\.jpeg","createdBy":"admin","deliveryService":"amc-live","id":1,"keyword":"PURGE","parameters":"TTL:24h""startTime":"2021-06-02 09:23:21-06"}}
+```
+
+Sample new response (includes the `invalidationType` on parameters field):
+```
+{"alerts":[{"text":"Invalidation request created for http://amc-linear-origin.local.tld/path/.*\\.jpeg, start:2021-06-02 15:23:21.348 +0000 UTC end 2021-06-03 15:23:21.348 +0000 UTC","level":"success"}],"response":{"assetUrl":"http://amc-linear-origin.local.tld/path/.*\\.jpeg","createdBy":"admin","deliveryService":"amc-live","id":1,"keyword":"PURGE","parameters":"TTL:24h,invalidationType:refresh","startTime":"2021-06-02 09:23:21-06"}}
+```
+
+___
+
+> Note: There are still 1.x routes that reference `UserInvalidationJob`, such as 
+		`user/current/jobs(/|\.json/?)?$`
+		`user/current/jobs(/|\.json/?)?$`
+		These routes are currently deprecated and the corresponding `structs` will be removed in a future release as well.
+
+#### Client Impact
+
+Likewise with Traffic Portal, the `go` clients will need to be updated to provide this additional functionality. Since an additional field has been added to `InvalidationJobInput` in `go-tc` lib, this can be set by the client as well. If left unset, it will default to "refresh".
+
+#### Data Model / Database Impact
+
+
+The current column `parameters` will now contain a cskv (comma separated key value) string. Currently it only stores the `TTL` for the invalidation request:
+```
+TTL:48h
+```
+
+Moving forward, this column will also contain the type of cache invalidation. For instance, the string may read:
+```
+TTL:48h,invalidationType:refetch
+```
+
+If there is no `invalidationType` in the cskv, it is assumed to be **REFRESH**/**STALE** as it's default value keeping with the current implementation. Otherwise the `invalidationType` will be either `refetch` or `refresh`, defaulting to `refresh` during validation.
+
+*OPTIONAL, BUT RECOMMENDED*: As part of this effort, the _Boy Scout Rule_ will be applied ("Always leave the campground cleaner than you found it."). The `agent`, `status`, `asset_type`, `object_type`, and `object_name` columns will be removed. `agent` and `status` are currently hardcoded to the value of 1. Similarly, `asset_type` is never a anything besides "file". `object_type` and `object_name` are not used at all. This would require a DB migration with Goose as well. If this were not implemented, no migration is necessary.
+
+> The removal of these columns will impact the `UserInvalidationJob` struct. Even though the only endpoint utilizing this struct (v1.1) is deprecated, it is still in use.
+
+### ORT/T3C Impact
+
+The `regexrevalidatedotconfig.go` in _lib/tc-atscfg_ currently has a function called `filterJobs` that is responsible for parsing the _parameters_ information from the job. Right now, that is only parsing the **TTL**, however it would then also need to parse the optional, extra field of **invalidationType**. 
+
+```go
+type revalJob struct {
+	AssetURL string
+	PurgeEnd time.Time
+	Type     string // MISS or STALE (default)
+}
+```
+
+As we can see above, if the value **invalidationType** is missing or equal to **refresh**, then the `revalJob` struct's `Type` field will be set to **STALE**. Otherwise, if the **invalidationType** is **refetch** then the `Type` field will be set to **MISS**. The struct is already prepared for this information, it needs only to be parsed by during the `filterJobs` function call.
+
+
+### Traffic Monitor Impact
+
+N/A - No changes
+
+### Traffic Router Impact
+
+N/A - No changes
+
+### Traffic Stats Impact
+
+N/A - No changes
+
+### Traffic Vault Impact
+
+N/A - No changes
+
+### Documentation Impact
+
+This information will need to be added to the current **Forcing Content Invalidation** section under the **_Quick How To Guides_** section under the **_Administrator's Guide_**
+[Content Invalidation](https://traffic-control-cdn.readthedocs.io/en/latest/admin/quick_howto/content_invalidation.html)
+
+Additionally, the **_Traffic Ops API_** `jobs` routes will need to be updated with the changes, API V1-V4.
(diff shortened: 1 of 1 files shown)


blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,240 @@
...
+##### Current Request
+
+Body:
+```
+{"startTime":"2021-06-02T15:23:21.348Z","deliveryService":11,"regex":"/path/.*\\.jpeg","ttl":24}
+```
+
+Which is mapped to a go `struct` in the `go-tc` lib.
+```go
+type InvalidationJobInput struct {
```
srijeet0406, line 60: The json tags dont seem to be aligned properly.

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,240 @@
...
+
+No new endpoints will be required. However the current invalidation job will now include an optional field during `Create`. Invalidation jobs are added by submitting a POST to the jobs endpoint. 
+
+**POST** /api/4.0/jobs
+
+##### Current Request
+
+Body:
+```
+{"startTime":"2021-06-02T15:23:21.348Z","deliveryService":11,"regex":"/path/.*\\.jpeg","ttl":24}
```
srijeet0406, line 55: Could this be displayed on multiple lines? Makes reading it a bit easier.

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,240 @@
...
+
+##### Create and Update
+Traffic Portal will need to update the Invalidation Job to account for the different options. When creating an Invalidation Job both options will need to be present (Perhaps a radio button? Default will be the original **REFRESH**).
+
+Tooltips should be added to ensure an understanding of this feature at a high level.
+
+##### Read
+When displaying the information, the **Invalidation Requests** table current shows the `Parameters` field, so the display will be straight forward with no manipulation.
+
+However, we derived and calculate the expiration field based on the TTL. This code will need to be modified to account for the additional information contained in the `Parameters` field.
```
srijeet0406, line 39: nit: `derive`

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,240 @@
...
+Add an "InvalidationType" to signify a specific type of invalidation request. The InvalidationType is an optional field and will not break backwards compatibility with previous API versions. If the field is included, it _must_ be either "refetch" or "refresh".
+
+Body:
+```
+{"startTime":"2021-06-02T15:23:21.348Z","deliveryService":11,"regex":"/path/.*\\.jpeg","ttl":24,"invalidationType":"refresh"}
+```
+
+This struct now contains the `InvalidationType *string` field.
+```go
+type InvalidationJobInput struct {
```
srijeet0406, line 81: same comment about json tags here

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,240 @@
...
+}
+```
+
+##### Proposed
+
+Add an "InvalidationType" to signify a specific type of invalidation request. The InvalidationType is an optional field and will not break backwards compatibility with previous API versions. If the field is included, it _must_ be either "refetch" or "refresh".
+
+Body:
+```
+{"startTime":"2021-06-02T15:23:21.348Z","deliveryService":11,"regex":"/path/.*\\.jpeg","ttl":24,"invalidationType":"refresh"}
```
srijeet0406, line 76: same comment about multiple lines here

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,240 @@
...
+
+Sample current response:
+```
+{"alerts":[{"text":"Invalidation request created for http://amc-linear-origin.local.tld/path/.*\\.jpeg, start:2021-06-02 15:23:21.348 +0000 UTC end 2021-06-03 15:23:21.348 +0000 UTC","level":"success"}],"response":{"assetUrl":"http://amc-linear-origin.local.tld/path/.*\\.jpeg","createdBy":"admin","deliveryService":"amc-live","id":1,"keyword":"PURGE","parameters":"TTL:24h""startTime":"2021-06-02 09:23:21-06"}}
+```
+
+Sample new response (includes the `invalidationType` on parameters field):
+```
+{"alerts":[{"text":"Invalidation request created for http://amc-linear-origin.local.tld/path/.*\\.jpeg, start:2021-06-02 15:23:21.348 +0000 UTC end 2021-06-03 15:23:21.348 +0000 UTC","level":"success"}],"response":{"assetUrl":"http://amc-linear-origin.local.tld/path/.*\\.jpeg","createdBy":"admin","deliveryService":"amc-live","id":1,"keyword":"PURGE","parameters":"TTL:24h,invalidationType:refresh","startTime":"2021-06-02 09:23:21-06"}}
+```
```
srijeet0406, line 108: same comment about multiple lines in both the above responses.

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,240 @@
...
+
+### Performance Impact
+
+There will be no performance impact for Traffic Control.
+
+> Note: This functionality may create a performance impact on caches that implement a REFETCH/MISS manual override based on a regex.
+
+### Security Impact
+
+The validation in Traffic Ops of the `invalidationType` field will be such that it can only be explicitly set to **refetch**. Any other value (missing, malformed, wrong data type, etc.) will result in either a 400 level error or a default to **refresh**. No other permissions are modified.
```
srijeet0406, line 214: Just to be clear, if the value of `invalidationType` is malformed or of the wrong data type, it will result in a `400`, else, if it is missing, or any value other than `refetch` or `refresh` will always default to `refresh`, correct?

blueprints/refetch-invalidation.md
ocket8888: should also syntax highlight as JSON

blueprints/refetch-invalidation.md
TaylorCFrey: Hm. How do you mean?

blueprints/refetch-invalidation.md
TaylorCFrey: Good call!

traffic_portal/test/integration/README.md
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
TaylorCFrey: I believe this was addressed in a subsequent push. Please let me know if you're still seeing this change.

go.mod
```diff
@@ -54,7 +54,6 @@ require (
 	github.com/miekg/dns v1.0.6-0.20180406150955-01d59357d468
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
-	github.com/ogier/pflag v0.0.2-0.20201025181535-73e519546fc0
```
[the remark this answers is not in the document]
TaylorCFrey: I believe this was addressed in a subsequent push. Please let me know if you're still seeing this change.

blueprints/refetch-invalidation.md
TaylorCFrey: Fixed.

blueprints/refetch-invalidation.md
TaylorCFrey: Fixed

blueprints/refetch-invalidation.md
TaylorCFrey: Fixed.

blueprints/refetch-invalidation.md
```diff
@@ -98,13 +109,44 @@ Since the field is optional and existing functionality only signifies a **REFRES
...
+			"level":"success"
+		}
+	],
+	"response":{
+		"assetUrl":"http://amc-linear-origin.local.tld/path/.*\\.jpeg",
+		"createdBy":"admin",
+		"deliveryService":"amc-live",
+		"id":1,
+		"keyword":"PURGE",
+		"parameters":"TTL:24h""startTime":"2021-06-02 09:23:21-06"
```
ocket8888, line 126: Invalid JSON here - also, can the start times be RFC3339 compliant?

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+Moving forward, this column will also contain the type of cache invalidation. For instance, the string may read:
+```
+TTL:48h,invalidationType:refetch
+```
+
+If there is no `invalidationType` in the cskv, it is assumed to be **REFRESH**/**STALE** as it's default value keeping with the current implementation. Otherwise the `invalidationType` will be either `refetch` or `refresh`, defaulting to `refresh` during validation.
+
+*OPTIONAL, BUT RECOMMENDED*: As part of this effort, the _Boy Scout Rule_ will be applied ("Always leave the campground cleaner than you found it."). The `agent`, `status`, `asset_type`, `object_type`, and `object_name` columns will be removed. `agent` and `status` are currently hardcoded to the value of 1. Similarly, `asset_type` is never a anything besides "file". `object_type` and `object_name` are not used at all. This would require a DB migration with Goose as well. If this were not implemented, no migration is necessary.
```
zrhoffman, line 178: > is never a anything

can be reworded

blueprints/refetch-invalidation.md
srijeet0406: I was thinking maybe we should have all the `json` tags indented at the same level. Something like this:
`
	AssetURL        *string `json:"assetUrl"`
	CreatedBy       *string `json:"createdBy"`
	DeliveryService *string `json:"deliveryService"`
	ID              *uint64 `json:"id"`
	Keyword         *string `json:"keyword"`
	Parameters      *string `json:"parameters"`
`

blueprints/refetch-invalidation.md
srijeet0406: Yeah I think if the user includes the `invalidationType` key and gives a blank string, that could probably be treated as `refresh` as well, but I'm new to invalidation jobs so maybe @ocket8888 would know better.
The following four types should be accepted IMO:
1.) `refresh`
2.) `refetch`
3.) blank string -> defaults to `refresh`
4.) `invalidationType` key absent -> defaults to `refresh`
Anything else is treated as a 400.

ocket8888: You absolutely _can_ do that - I would suggest instead that the field not be optional, and the client creates objects exactly as they desire them to exist with no room for surprises.

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+		These routes are currently deprecated and the corresponding `structs` will be removed in a future release as well.
+
+#### Client Impact
+
+Likewise with Traffic Portal, the `go` clients will need to be updated to provide this additional functionality. Since an additional field has been added to `InvalidationJobInput` in `go-tc` lib, this can be set by the client as well. If left unset, it will default to "refresh".
+
+#### Data Model / Database Impact
+
+
+The current column `parameters` will now contain a cskv (comma separated key value) string. Currently it only stores the `TTL` for the invalidation request:
```
limited, line 166: Which table is this in? 

Why not add a new column rather than storing multiple values in the same column?

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+TTL:48h,invalidationType:refetch
+```
+
+If there is no `invalidationType` in the cskv, it is assumed to be **REFRESH**/**STALE** as it's default value keeping with the current implementation. Otherwise the `invalidationType` will be either `refetch` or `refresh`, defaulting to `refresh` during validation.
+
+*OPTIONAL, BUT RECOMMENDED*: As part of this effort, the _Boy Scout Rule_ will be applied ("Always leave the campground cleaner than you found it."). The `agent`, `status`, `asset_type`, `object_type`, and `object_name` columns will be removed. `agent` and `status` are currently hardcoded to the value of 1. Similarly, `asset_type` is never a anything besides "file". `object_type` and `object_name` are not used at all. This would require a DB migration with Goose as well. If this were not implemented, no migration is necessary.
+
+> The removal of these columns will impact the `UserInvalidationJob` struct. Even though the only endpoint utilizing this struct (v1.1) is deprecated, it is still in use.
+
+### ORT/T3C Impact
```
limited, line 182: Can you please describe how the construction of the `regexrevalidate.config` file will change given that this an external interface of Traffic Control?

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+
+The `regexrevalidatedotconfig.go` in _lib/tc-atscfg_ currently has a function called `filterJobs` that is responsible for parsing the _parameters_ information from the job. Right now, that is only parsing the **TTL**, however it would then also need to parse the optional, extra field of **invalidationType**.
```
limited, line 184: What considerations are made for TrafficServer version dependencies? 

Is `MISS` supported by all ATS versions? If not, what happens when an operator sets a `REFETCH` job on a set of caches that do not support that invalidation type?

blueprints/refetch-invalidation.md
mitchell852: yeah, surprises are not good imo. so i guess 400 if invalidationType != refresh or refetch. and because you are adding this to a new major version of the api (4.0), you have the luxury of adding the new required field to the request.

blueprints/refetch-invalidation.md
mitchell852: agree. stuffing lots of values in a column seems problematic going forward. can't you just reuse the `keyword` column of the job table? currently, keyword is PURGE for EVERY job. maybe keyword should be REFRESH and REFETCH instead and do a database migration to change all existing PURGE values to REFRESH?

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+# Add REFETCH capability option for Content Invalidation
+
+## Problem Description
+
+Currently, within ATC, there is a concept of Invalidation Jobs. These Invalidation Jobs give a user the ability to queue an invalidation for a resource, primarily based on regular expressions. The invalidation is gathered and treated as though there was a cache **STALE**, allowing the cache to query the origin server to **REFRESH** the resource. However, should the cache policy still be incorrect or misconfigured, the resource could be updated on the origin server, but the cache will still receive a 304 - Not Modified HTTP status response.
```
mitchell852, line 23: so i don't think the cache queries the "origin server" per se. instead it queries the "parent server" which in some cases will be a parent mid (if the cache is an edge) and the origin server (if the cache is a mid) and is simply the "parent server" when a topology is imployed by a delivery service.

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+
+##### Create and Update
+Traffic Portal will need to update the Invalidation Job to account for the different options. When creating an Invalidation Job both options will need to be present (Perhaps a radio button? Default will be the original **REFRESH**).
+
+Tooltips should be added to ensure an understanding of this feature at a high level.
+
+##### Read
+When displaying the information, the **Invalidation Requests** table current shows the `Parameters` field, so the display will be straight forward with no manipulation.
+
+However, we derive and calculate the expiration field based on the TTL. This code will need to be modified to account for the additional information contained in the `Parameters` field.
```
mitchell852, line 39: tbh i think the parameters column was a poor design (there should have simply been a TTL column) and we should probably not follow that approach.

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+##### Response
+
+The response will be modified, then, to return this new value as well.
+
+Sample current response:
+```json
+{
+	"alerts":[
+		{
+			"text":"Invalidation request created for http://amc-linear-origin.local.tld/path/.*\\.jpeg, start:2021-06-02 15:23:21.348 +0000 UTC end 2021-06-03 15:23:21.348 +0000 UTC",
```
mitchell852, line 116: How about changing the response text to "Invalidation request (refetch) created for..."

blueprints/refetch-invalidation.md
ocket8888: Seems like there are two options when `MISS` is unavailable based on ATS version: ignore it or use refresh. Can you suggest which would be better?

blueprints/refetch-invalidation.md
ocket8888: I agree, any jobs that don't have a numeric TTL parameter will fail to operate properly; I think a migration could reasonably parse it and change it to a numeric TTL column, since if the migration fails it means ORT is failing in some regard.

blueprints/refetch-invalidation.md
ocket8888: Like if you ran `go fmt` on a file that included that code, it would change it to:
```go
type InvalidationJobInput {
	DeliveryService *interface{} `json:"deliveryService"`
	Regex           *string      `json:"regex"`
	StartTime       *Time        `json:"startTime"`
	TTL             *interface{} `json:"ttl"`
	dsid            *uint
	ttl             *time.Duration
}
```

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+	"regex":"/path/.*\\.jpeg",
+	"ttl":24,
+	"invalidationType":"refresh"
+}
+```
+
+This struct now contains the `InvalidationType *string` field.
+```go
+type InvalidationJobInput struct {
+	DeliveryService *interface{} `json:"deliveryService"`
```
ocket8888, line 93: This was a `*interface{}` in APIv3 to give a transition period between using Delivery Service ID (APIv2 and earlier) to XMLID (hopefully all versions moving forward) - so this could be a `string`. It's not really related to the changes we're making, but it's something that could be cleaned up since we're making structure changes anyway.

That would also eliminate the need for the unexported `dsid` and the hoops you need to jump through for that.

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+	Regex *string `json:"regex"`
```
ocket8888, line 94: `regex` isn't actually optional, right? Can this just be a `string` instead of a `*string`?

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+	InvalidationType *string `json:"invalidationType,omitempty"`
```
ocket8888, line 95: I don't really like optional properties, but also why `omitempty`? That won't actually be allowed to be empty in the database, right?

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+	StartTime *Time `json:"startTime"`
+	TTL *interface{} `json:"ttl"`
```
ocket8888, line 97: Does anyone really like being able to put in "duration strings"? I don't think that's as useful as I thought it was. JSON and Javascript numbers are IEEE double-precision floating point numbers, which means that as long as we don't need jobs to last longer than 290 years this could just be a `time.Duration` which is a duration in nanoseconds.

That would also eliminate the need for the unexported `ttl` and the hoops you need to jump through for that.

blueprints/refetch-invalidation.md
```diff
@@ -0,0 +1,282 @@
...
+	dsid *uint
+	ttl  *time.Duration
+}
+```
+
+##### Parsing the value
+
+Since the field is optional and existing functionality only signifies a **REFRESH**/**STALE** capability, if the field is omitted, empty, malformed, or in any way _not_ `refetch` then it will be treated as `refresh`.
```
ocket8888, line 105: So if I pass in a type that's `demo1`, for example, that'll be treated as `refresh`? That worries me a bit.

blueprints/refetch-invalidation.md
TaylorCFrey: Ahh, interestingly, I believe it is spaced correclty in the source code, but I deleted the comments between each of the properties for brevity, which resulted in misaligned tags. I can/will fix this.

blueprints/refetch-invalidation.md
limited: I'm not an ATS expert, so I can't make a solid suggestion here. 

I think whatever happens probably needs to be bubbled up to the operator. You don't want to silently ignore or silently do a stale-refresh when you expect a Miss-Refetch.

ocket8888: Well, when I need an ATS and/or ORT/t3c/atstccfg expert I ask @rob05c - lets see if he has an opinion

rob05c: It's definitely not supported on older versions. It's a new feature of the regex_revalidate plugin.

1. Unfortunately, there's not an easy way to bubble it up. We don't really have a mechanism to send data from caches to TO, and even if we did, it'd be pretty unwieldy to collect that from every single cache.

We do have the ATS version in a Parameter in TO, we could hard-code it. But someone might backport the plugin changeset to an older version, and then ATC just won't allow using it. So that's not a great solution.

2. the "refetch" feature is **only** necessary or useful if the Origin server is fundamentally broken and completely violating the HTTP Spec. Without "refetch," ATS does an IMS request. So, the only reason it's ever needed is when a Tenant's Origin is responding with a 304 when the asset actually is modified. In complete violation of RFC 7234 and every HTTP Caching specification. But of course, we deal with horribly broken Origins every day. But I do think we should make that clear in the docs, and maybe the UI.

3. To work correctly, every single Cache has to support it. Which is mostly fine, since CDN operators don't _usually_ operate with widely varying ATS versions. Some do, but I think the majority have a single version and maybe some newer Canaries.

With those 3 things in mind, I think the best solution might be to make it a config Feature Flag in Traffic Ops/+Portal, with Documentation that they must have version X of regex_revalidate. As an additional safety, we should also make `t3c` check and log an error (which operators can then Alarm on).

It's not as good as bubbling it up, but considering all those things, that's the best solution I'm seeing at the moment.

blueprints/refetch-invalidation.md
rawlinp: Yeah, I think nil = refresh, empty/malformed = invalid.

blueprints/refetch-invalidation.md
rawlinp: Just for awareness @TaylorCFrey, a "feature flag" in TO usually means a parameter in the `GLOBAL` profile to enable the new feature -- similar to what we've done in the past w/ `use_tenancy` and what we currently do w/ `use_reval_pending`. That sounds like a good solution here.

blueprints/refetch-invalidation.md
rawlinp: Just because we have a new major version we can add this to doesn't mean we should just make every new change a breaking change. I don't think making this a new required field adds enough value to warrant that since we can easily provide a default instead.

blueprints/refetch-invalidation.md
ocket8888: Could also be in the TO configuration file

rawlinp: No, because really the feature flag is for `t3c` since it's generating the ATS config. I guess it could be a `t3c` option, but I'm not sure that's better than our tried and true method of global profile params.

ocket8888: Maybe I'm misunderstanding how this is supposed to work. I thought it'd be like: you make a new job, if it's using "refetch" then TO checks if the feature flag is allowed. If it isn't, you can't make that job, and any existing jobs are filtered out of GET requests for the /jobs endpoint.

It sounds like you're saying you set the flag in TO, then ORT/t3c/atstccfg looks at it and determines based on that what to do with jobs that use "refetch" - which makes sense, because otherwise I suppose I don't know how `t3c` could "check and log an error"

rawlinp: > It sounds like you're saying you set the flag in TO, then ORT/t3c/atstccfg looks at it and determines based on that what to do with jobs that use "refetch" - which makes sense, because otherwise I suppose I don't know how t3c could "check and log an error"

Yes, that is how `use_reval_pending` works currently and is how I imagine a feature flag for this feature would work as well.

rob05c: When I said "feature flag" I was thinking the TO configuration file; but a global Parameter also works.

> I thought it'd be like: you make a new job, if it's using "refetch" then TO checks if the feature flag is allowed. If it isn't, you can't make that job

That's what I was thinking, but also having a check in `t3c-apply` as a fail-safe, if it ever does get a job and the installed regex_revalidate is too old to understand it, it would log an error (but probably still do the reval, because an IMS reval is better than nothing and doesn't hurt if they wanted a no-IMS reval).

rob05c: We need to not allow creating those jobs at all, if the feature-flag isn't set. Because by the time it gets to the cache, it's too late: the user put in jobs, that they expected to work, and now we have an error we can't reasonably propagate up to them that the job won't work. 

But we already knew it wasn't going to work, we have the flag in TO. We should be able to return an error from the /jobs POST.

rob05c: Bearing in mind, these won't be intermittent errors. The endpoint will always return an error if the flag isn't set.

That's a good reason to put the flag in the Parameter instead of the config file: the Param is already available in the API. Which lets UIs like the Portal fetch it, to decide whether to hide that extra button from users, so they never even see the option that will always error.

blueprints/refetch-invalidation.md
TaylorCFrey: I've added a bit more of a description. There isn't much more to the construction of T3C's `regexrevalidate.config` file. This is already implemented, we are allowing it to be explicitly controlled via ATC's APIs, interfaces, and clients.

Essentially, it remains the same format, but with an optional third field for each non-comment entry.

blueprints/refetch-invalidation.md
TaylorCFrey: I've modified this to be explicitly either `refresh` or `refetch`. All other submissions, including missing/omitted, will be reject as a client error (400 - Bad Request).

blueprints/refetch-invalidation.md
TaylorCFrey: Good catch, fixed. Also added information stating any Time fields will be RFC3339 compliant. However, there is a `jobs.go` file in `lib/go-tc` that is already deprecated that is using it's own time format.

blueprints/refetch-invalidation.md
TaylorCFrey: Nice one! Fixed, and hopefully clearer (more clear?).

blueprints/refetch-invalidation.md
TaylorCFrey: I've pivoted a bit on this original idea. My first attempt was to change as little as possible, reusing the `parameters` column. However, I believe we all agree that explicity and purposefully is a better approach. I've taken the approach of explicit columns (and datatypes). I also added a recommendation which I think we should implement that performs a great deal of clean up for the _jobs_ table. We are carrying a decent amount of future-proofing along with us that doesn't appear to be utilized anywhere.

Definitely let me know if this newer proposal is lacking in clarity.

blueprints/refetch-invalidation.md
TaylorCFrey: I've modified the proposal to be more explicit, rather than the first attempt (which was to make the most minimal changes). Relish the feedback.

blueprints/refetch-invalidation.md
TaylorCFrey: Hm, this is a really good point and observation. My understanding is that this all revolves entirely around misconfigured _origins_, and even more specifically around _origins_ we can't "fix". If it was an issue within the CDN topology, would we be able to manually intervene in the chain?

blueprints/refetch-invalidation.md
TaylorCFrey: I added some clarity to the alert message.

blueprints/refetch-invalidation.md
TaylorCFrey: Removed `omitempty`. I originally thought we could aim for backwards compatibility. Also, whether it was an explicit column in the DB I figured it would default to `NULL`. However, with the newer changes this will be explicit and no longer optional.

blueprints/refetch-invalidation.md
TaylorCFrey: I went ahead and made it explicit, but we can always opt for it to be optional if we decide to pivot going forward.

blueprints/refetch-invalidation.md
TaylorCFrey: This is a good catch, and I believe it should be documented at the very least. Same with the `TTL` below as well. I'm happy to add this as part of the refactor effort, especially since I expect more `-` than `+` :)

blueprints/refetch-invalidation.md
TaylorCFrey: Another good observation. What is actually optional? Any of the fields? We can make them all explicitly required at this time as well if you believe that's a benefit at this time.

blueprints/refetch-invalidation.md
TaylorCFrey: The API call implementations (whether go clients or TP), from what I can gather, are sending up JS numbers. We can certainly enforce a numeric data type here, and even if we stick with the hours format, it would still be clearer than empty interface.

blueprints/refetch-invalidation.md
ocket8888: That is the purpose for introducing the feature, but as I understand it the implementation will have the same effects on any requests made to it's parent for the affected content - regardless of whether or not it's an origin.

blueprints/refetch-invalidation.md
ocket8888: I don't think any of the properties were optional, but it's easier to tell for strings than other things if a value was actually given. For example `"regex": ""` is just as invalid as `"regex": null` or a missing `"regex"` property, and Go's parser can't tell the difference between those cases for a `string` property. However, for something numeric, Go can't tell the difference between `0`, `null` and undefined values for an `int` property, but sometimes `0` is a valid value, so you have to use `*int`. It's a very annoying and pedantic balancing act you have to perform with the number of things you have to check for `nil` all the time versus what the Go parser can and can't distinguish between.

Historically, though, the reason these fields are all pointers is because originally this was going to be the first endpoint to allow PATCH requests, where every property is optional by definition. That didn't pan out because there was prerequisite work that needed to be done to support PATCH safely, but I guess I never changed the fields back from pointers.

blueprints/refetch-invalidation.md
ocket8888: I'm going to resolve this then, since the principal concern of the original comment has been addressed; there's already a discussion about whether or not it should be optional elsewhere.

blueprints/refetch-invalidation.md
TaylorCFrey: Cool info on the PATCH. That makes sense. I've gone ahead and made both `DeliveryService` and `Regex` nonoptional (required for brevity?).

blueprints/refetch-invalidation.md
TaylorCFrey: I modified the original description to reference a CDN rather than a cache. Also added clarity that this could be a parent along the chain.

blueprints/refetch-invalidation.md
TaylorCFrey: I've added some language specifically stating the addition of a global parameter, defaulting to rejecting `refetch` POSTS until specfically set to allow.

PR #5913, opened 2021-06-03
Re-encrypt Tool for TV Postgres

mattjackson220:

This PR adds a tool to re-encrypt the data in TV Postgres to use a new key.

- Documentation
- Traffic Ops
- Traffic Vault

Run CIAB with TV setup to use Postgres and put the Postgres TV information into the `reencrypt.cfg` file.
Log into TO and verify that the reencrypt binary is at `app/db/reencrypt/reencrypt` and the conf file is in the same directory
Verify that you can run the binary in there with the previousKey and newKey inputs in order to re-encrypt your SSL Keys, URL Sign Keys, URI Signing Keys, and DNSSEC Keys in Postgres TV
Verify that the old key is saved with the date / time of the re-encrypt.
Restart TO and verify that you can successfully use the API endpoints / TP pages for those keys

No tests since it is a stand-alone tool.

### DIFF
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -112,7 +112,7 @@ Options and Arguments
 .. code-block:: bash
 	:caption: Example Usage
 
-	 ./reencrypt --new-key ~/exampleNewKey.txt --previous-key ~/exampleOldKey.txt
+	./reencrypt --new-key ~/exampleNewKey.txt --previous-key ~/exampleOldKey.txt
 
 reencrypt.conf
 """"""""""""""
@@ -170,7 +170,7 @@ Configuring Riak
 Follow these steps to configure Riak in a production environment.
 
 Self Signed Certificate configuration
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+"""""""""""""""""""""""""""""""""""""
 .. note:: Self-signed certificates are not recommended for production use. Intended for development or learning purposes only. Modify subject as necessary.
 
 .. code-block:: shell
@@ -192,7 +192,7 @@ Self Signed Certificate configuration
 
 
 Riak Configuration File
-^^^^^^^^^^^^^^^^^^^^^^^
+"""""""""""""""""""""""
 The following steps need to be performed on each Riak server in the cluster:
 
 #. Log into Riak server as root
@@ -212,14 +212,14 @@ The following steps need to be performed on each Riak server in the cluster:
 .. _tv-admin-enable-tlsv1.1:
 
 Enabling TLS 1.1 (required)
-"""""""""""""""""""""""""""
+'''''''''''''''''''''''''''
 
 #. Add a line at the bottom of the :file:`riak.conf` for TLSv1.1 by setting ``tls_protocols.tlsv1.1 = on``
 #. Once the configuration file has been updated restart Riak
 #. Consult the `Riak documentation <https://docs.riak.com/riak/kv/latest/setup/installing/verify/>`_ for instructions on how to verify the installed service
 
 ``riak-admin`` Configuration
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+""""""""""""""""""""""""""""
 ``riak-admin`` is a command line utility used to configure Riak that needs to be run as root on a server in the Riak cluster.
 
 .. seealso:: `The riak-admin documentation <https://docs.riak.com/riak/kv/latest/using/admin/riak-admin/>`_
@@ -257,7 +257,7 @@ Enabling TLS 1.1 (required)
 
 
 Traffic Ops Configuration
-^^^^^^^^^^^^^^^^^^^^^^^^^
+"""""""""""""""""""""""""
 Before a fully set-up Riak instance may be used as the Traffic Vault backend, it must be added as a server to Traffic Ops. The easiest way to accomplish this is via Traffic Portal at :menuselection:`Configure --> Servers`, though :ref:`to-api-servers` may also be used by low-level tools and/or scripts. The Traffic Ops configuration file :file:`/opt/traffic_ops/app/conf/cdn.conf` must be updated to set ``traffic_vault_backend`` to ``"riak"`` and the ``traffic_vault_config`` to include the correct username and password for accessing the Riak database.
 
 Configuring Riak Search
@@ -265,7 +265,7 @@ Configuring Riak Search
 In order to more effectively support retrieval of SSL certificates by Traffic Router and :term:`ORT`, the Riak backend for Traffic Vault uses `Riak search <https://docs.riak.com/riak/kv/latest/using/reference/search/>`_. Riak Search uses `Apache Solr <https://lucene.apache.org/solr>`_ for indexing and searching of records. This section explains how to enable, configure, and validate Riak Search.
 
 Riak Configuration
-^^^^^^^^^^^^^^^^^^
+""""""""""""""""""
 On each Traffic Vault server follow these steps.
 
 #. If Java (JDKv1.8+) is not already installed on your Riak server, install Java
@@ -299,7 +299,7 @@ On each Traffic Vault server follow these steps.
 		systemctl restart riak
 
 One-time Configuration
-""""""""""""""""""""""
+''''''''''''''''''''''
 After Riak has been configured to use Riak Search, permissions still need need to be updated to allow users to utilize this feature. Unlike actually setting up Riak Search, the permissions step need only be done on any *one* of the Riak servers in the cluster.
 
 #. Use ``riak-admin`` to grant ``search.admin`` permissions to the "admin" user and ``search.query`` permissions to **both** the "admin" user and the "riakuser" user. The "admin" user will also require ``search.admin`` permissions on the ``schema`` (in addition to ``index``) and ``riak_core.set_bucket`` permissions on ``any``.
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -114,6 +114,18 @@ Options and Arguments
 
 	 ./reencrypt --new-key ~/exampleNewKey.txt --previous-key ~/exampleOldKey.txt
 
+reencrypt.conf
+""""""""""""""
+This file deals with configuration of the Traffic Vault Database to be used with the :program:`reencrypt` tool.
+
+:dbname: The name of the PostgreSQL database used.
+:hostname: The hostname (:abbr:`FQDN (Fully Qualified Domain Name)`) of the server that runs the Traffic Vault Database.
+:password: The password to use when authenticating with the Traffic Vault database.
+:port: The port number (as a string) on which the Traffic Vault Database is listening for incoming connections (NOTE: the PostgreSQL default is 5432).
+:ssl: A boolean that sets whether or not the Traffic Vault Database encrypts its connections with SSL.
+:user: The name of the user as whom to connect to the database.
+
+
 .. _traffic_vault_riak_backend:
 
 Riak (deprecated)
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -79,6 +79,36 @@ Administration of the PostgreSQL database for Traffic Vault
 
 Similar to administering the Traffic Ops database, the :ref:`admin <database-management>` tool should be used for administering the PostgreSQL Traffic Vault backend.
 
+AES Re-Encryption Tool:
+-----------------------
+
+.. program:: reencrypt
+
+app/db/reencrypt/reencrypt
+=====================================================================
+The :program:`app/db/reencrypt/reencrypt` binary is for use to re-encrypt all keys in the Postgres Traffic Vault.
+
+.. note:: For proper resolution of configuration files, it's recommended that this binary be run from the ``app/db/reencrypt`` directory
+
+Usage
+-----
+``./reencrypt [options]``
+
+Options and Arguments
+---------------------
+.. option:: -newKey
+
+	The file path for the new base64-encoded AES key.
+
+.. option:: -previousKey
+
+	The file path for the previous base64-encoded AES key.
+
+.. code-block:: bash
+	:caption: Example Usage
+
+	 ./reencrypt -newKey ~/exampleNewKey.txt -previousKey ~/exampleOldKey.txt
+
 .. _traffic_vault_riak_backend:
 
 Riak (deprecated)
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -83,7 +83,7 @@ Similar to administering the Traffic Ops database, the :ref:`admin <database-man
 
 app/db/reencrypt/reencrypt
 --------------------------
-The :program:`app/db/reencrypt/reencrypt` binary is used to re-encrypt all data in the Postgres Traffic Vault with a new base64-encoded AES key.
+The :program:`reencrypt` binary is used to re-encrypt all data in the Postgres Traffic Vault with a new base64-encoded AES key.
 
 .. note:: For proper resolution of configuration files, it's recommended that this binary be run from the ``app/db/reencrypt`` directory.
 
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -96,18 +96,22 @@ Usage
 
 Options and Arguments
 ---------------------
-.. option:: -newKey
+.. option:: --new-key
 
 	The file path for the new base64-encoded AES key.
 
-.. option:: -previousKey
+.. option:: --previous-key
 
 	The file path for the previous base64-encoded AES key.
+
+.. option:: --cfg
+
+	The path for the configuration file. Default is ``./reencrypt.conf``.
 
 .. code-block:: bash
 	:caption: Example Usage
 
-	 ./reencrypt -newKey ~/exampleNewKey.txt -previousKey ~/exampleOldKey.txt
+	 ./reencrypt --new-key ~/exampleNewKey.txt --previous-key ~/exampleOldKey.txt
 
 .. _traffic_vault_riak_backend:
 
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -121,7 +121,7 @@ This file deals with configuration of the Traffic Vault Database to be used with
 :dbname: The name of the PostgreSQL database used.
 :hostname: The hostname (:abbr:`FQDN (Fully Qualified Domain Name)`) of the server that runs the Traffic Vault Database.
 :password: The password to use when authenticating with the Traffic Vault database.
-:port: The port number (as a string) on which the Traffic Vault Database is listening for incoming connections (NOTE: the PostgreSQL default is 5432).
+:port: The port number on which the Traffic Vault Database is listening for incoming connections (NOTE: the PostgreSQL default is 5432).
 :ssl: A boolean that sets whether or not the Traffic Vault Database encrypts its connections with SSL.
 :user: The name of the user as whom to connect to the database.
 
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -79,16 +79,13 @@ Administration of the PostgreSQL database for Traffic Vault
 
 Similar to administering the Traffic Ops database, the :ref:`admin <database-management>` tool should be used for administering the PostgreSQL Traffic Vault backend.
 
-AES Re-Encryption Tool:
------------------------
-
 .. program:: reencrypt
 
 app/db/reencrypt/reencrypt
-=====================================================================
-The :program:`app/db/reencrypt/reencrypt` binary is for use to re-encrypt all keys in the Postgres Traffic Vault.
+--------------------------
+The :program:`app/db/reencrypt/reencrypt` binary is used to re-encrypt all data in the Postgres Traffic Vault with a new base64-encoded AES key.
 
-.. note:: For proper resolution of configuration files, it's recommended that this binary be run from the ``app/db/reencrypt`` directory
+.. note:: For proper resolution of configuration files, it's recommended that this binary be run from the ``app/db/reencrypt`` directory.
 
 Usage
 -----
@@ -96,17 +93,21 @@ Usage
 
 Options and Arguments
 ---------------------
-.. option:: --new-key
+.. option:: --new-key NEW_KEY
 
-	The file path for the new base64-encoded AES key.
+	(Optional) The file path for the new base64-encoded AES key. Default is ``/opt/traffic_ops/app/conf/new.key``.
 
-.. option:: --previous-key
+.. option:: --previous-key PREVIOUS_KEY
 
-	The file path for the previous base64-encoded AES key.
+	(Optional) The file path for the previous base64-encoded AES key. Default is ``/opt/traffic_ops/app/conf/aes.key``.
 
-.. option:: --cfg
+.. option:: --cfg CONFIG_FILE
 
-	The path for the configuration file. Default is ``./reencrypt.conf``.
+	(Optional) The path for the configuration file. Default is ``./reencrypt.conf``.
+
+.. option:: --help
+
+	(Optional) Print usage information and exit.
 
 .. code-block:: bash
 	:caption: Example Usage
--- a/docs/source/admin/traffic_vault.rst
+++ b/docs/source/admin/traffic_vault.rst
@@ -88,11 +88,11 @@ The :program:`app/db/reencrypt/reencrypt` binary is used to re-encrypt all data 
 .. note:: For proper resolution of configuration files, it's recommended that this binary be run from the ``app/db/reencrypt`` directory.
 
 Usage
------
+"""""
 ``./reencrypt [options]``
 
 Options and Arguments
----------------------
+"""""""""""""""""""""
 .. option:: --new-key NEW_KEY
 
 	(Optional) The file path for the new base64-encoded AES key. Default is ``/opt/traffic_ops/app/conf/new.key``.
--- a/traffic_ops/app/db/reencrypt/reencrypt.conf
+++ b/traffic_ops/app/db/reencrypt/reencrypt.conf
@@ -1,8 +1,8 @@
 {
 	"dbname": "",
-    "hostname": "",
-    "user": "",
-    "password": "",
+	"hostname": "",
+	"user": "",
+	"password": "",
 	"port": 5432,
 	"ssl": false
 }
--- a/traffic_ops/app/db/reencrypt/reencrypt.conf
+++ b/traffic_ops/app/db/reencrypt/reencrypt.conf
@@ -1,0 +1,8 @@
+{
+	"dbname": "",
+    "hostname": "",
+    "user": "",
+    "password": "",
+	"port": 5432,
+	"ssl": false
+}
--- a/traffic_ops/app/db/reencrypt/reencrypt.go
+++ b/traffic_ops/app/db/reencrypt/reencrypt.go
@@ -55,7 +55,7 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/trafficvault/backends/postgres"
+	"github.com/apache/trafficcontrol/lib/go-util"
 
 	"github.com/jmoiron/sqlx"
 	_ "github.com/lib/pq"
@@ -187,12 +187,12 @@ func reEncryptSslKeys(db *sqlx.DB, previousKey []byte, newKey []byte) error {
 		if err = rows.Scan(&id, &encryptedSslKeys); err != nil {
 			return errors.New(fmt.Sprintf("getting SSL Keys: %v", err))
 		}
-		jsonKeys, err := postgres.AESDecrypt(encryptedSslKeys, previousKey)
+		jsonKeys, err := util.AESDecrypt(encryptedSslKeys, previousKey)
 		if err != nil {
 			return errors.New(fmt.Sprintf("reading SSL Keys: %v", err))
 		}
 
-		reencryptedKeys, err := postgres.AESEncrypt(jsonKeys, newKey)
+		reencryptedKeys, err := util.AESEncrypt(jsonKeys, newKey)
 		if err != nil {
 			return errors.New(fmt.Sprintf("encrypting SSL Keys with new key: %v", err))
 		}
@@ -239,12 +239,12 @@ func reEncryptUrlSigKeys(db *sqlx.DB, previousKey []byte, newKey []byte) error {
 		if err = rows.Scan(&ds, &encryptedSslKeys); err != nil {
 			return errors.New(fmt.Sprintf("getting URL Sig Keys: %v", err))
 		}
-		jsonKeys, err := postgres.AESDecrypt(encryptedSslKeys, previousKey)
+		jsonKeys, err := util.AESDecrypt(encryptedSslKeys, previousKey)
 		if err != nil {
 			return errors.New(fmt.Sprintf("reading URL Sig Keys: %v", err))
 		}
 
-		reencryptedKeys, err := postgres.AESEncrypt(jsonKeys, newKey)
+		reencryptedKeys, err := util.AESEncrypt(jsonKeys, newKey)
 		if err != nil {
 			return errors.New(fmt.Sprintf("encrypting URL Sig Keys with new key: %v", err))
 		}
@@ -291,12 +291,12 @@ func reEncryptUriSigningKeys(db *sqlx.DB, previousKey []byte, newKey []byte) err
 		if err = rows.Scan(&ds, &encryptedSslKeys); err != nil {
 			return errors.New(fmt.Sprintf("getting URI Signing Keys: %v", err))
 		}
-		jsonKeys, err := postgres.AESDecrypt(encryptedSslKeys, previousKey)
+		jsonKeys, err := util.AESDecrypt(encryptedSslKeys, previousKey)
 		if err != nil {
 			return errors.New(fmt.Sprintf("reading URI Signing Keys: %v", err))
 		}
 
-		reencryptedKeys, err := postgres.AESEncrypt(jsonKeys, newKey)
+		reencryptedKeys, err := util.AESEncrypt(jsonKeys, newKey)
 		if err != nil {
 			return errors.New(fmt.Sprintf("encrypting URI Signing Keys with new key: %v", err))
 		}
@@ -343,12 +343,12 @@ func reEncryptDNSSECKeys(db *sqlx.DB, previousKey []byte, newKey []byte) error {
 		if err = rows.Scan(&ds, &encryptedSslKeys); err != nil {
 			return errors.New(fmt.Sprintf("getting DNSSEC Keys: %v", err))
 		}
-		jsonKeys, err := postgres.AESDecrypt(encryptedSslKeys, previousKey)
+		jsonKeys, err := util.AESDecrypt(encryptedSslKeys, previousKey)
 		if err != nil {
 			return errors.New(fmt.Sprintf("reading DNSSEC Keys: %v", err))
 		}
 
-		reencryptedKeys, err := postgres.AESEncrypt(jsonKeys, newKey)
+		reencryptedKeys, err := util.AESEncrypt(jsonKeys, newKey)
 		if err != nil {
 			return errors.New(fmt.Sprintf("encrypting DNSSEC Keys with new key: %v", err))
 		}
--- a/traffic_ops/app/db/reencrypt/reencrypt.go
+++ b/traffic_ops/app/db/reencrypt/reencrypt.go
@@ -174,7 +174,7 @@ func reEncryptSslKeys(tx *sql.Tx, previousKey []byte, newKey []byte) error {
 	}
 	defer rows.Close()
 
-	sslKeyMap := map[int]string{}
+	sslKeyMap := map[int][]byte{}
 
 	for rows.Next() {
 		id := 0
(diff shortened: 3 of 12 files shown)


traffic_ops/app/db/reencrypt/reencrypt.conf
```diff
@@ -0,0 +1,8 @@
+{
+	"dbname": "",
+    "hostname": "",
+    "user": "",
+    "password": "",
+	"port": 5432,
+	"ssl": false
```
shamrickus, line 7: Inconsistent white space.

shamrickus (changes requested):
Can't leave comments on it, but you should ignore & remove the reencrypt binary

docs/source/admin/traffic_vault.rst
```diff
@@ -79,6 +79,36 @@ Administration of the PostgreSQL database for Traffic Vault
...
+
+.. note:: For proper resolution of configuration files, it's recommended that this binary be run from the ``app/db/reencrypt`` directory
+
+Usage
+-----
+``./reencrypt [options]``
+
+Options and Arguments
+---------------------
+.. option:: -newKey
```
ocket8888, line 99: If an option takes an option-argument it should be shown in the `.. option::` directive's argument, and as long options are a GNU extension it should be shown in GNU standard long form (two `-`s) - according to POSIX guidelines, this single option is equivalent to `-n -e -w -K ey` (ambiguity caused by two `e`s). GNU also recommends that long options be hyphen separated (kebab-case rather than camelCase)

e.g.
```rst
.. option:: --new-key PATH

	The file path for the new base64-encoded AES key.
```

I think switching to kebab-case would be an improvement, personally, but that would be a code change and it doesn't really matter much. What's important is that everywhere it's documented it uses the proper `--` to avoid confusion, and indicates that it requires an argument.


Excerpt from [the GNU libc manual, Section 25.1.1](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html):

> POSIX recommends these conventions for command line arguments. getopt (see Getopt) and argp_parse (see Argp) make it easy to implement them.
>
>>   * Arguments are options if they begin with a hyphen delimiter (‘-’).
>>   * Multiple options may follow a hyphen delimiter in a single token if the options do not take arguments. Thus, ‘-abc’ is equivalent to ‘-a -b -c’.
>>   * Option names are single alphanumeric characters (as for isalnum; see Classification of Characters).
>>   * Certain options require an argument. For example, the ‘-o’ command of the ld command requires an argument—an output file name.
>>   * An option and its argument may or may not appear as separate tokens. (In other words, the whitespace separating them is optional.) Thus, ‘-o foo’ and ‘-ofoo’ are equivalent.
>>   * Options typically precede other non-option arguments.
>>
>>    The implementations of getopt and argp_parse in the GNU C Library normally make it appear as if all the option arguments were specified before all the non-option arguments for the purposes of parsing, even if the user of your program intermixed option and non-option arguments. They do this by reordering the elements of the argv array. This behavior is nonstandard; if you want to suppress it, define the _POSIX_OPTION_ORDER environment variable. See Standard Environment.
>>   * The argument ‘--’ terminates all options; any following arguments are treated as non-option arguments, even if they begin with a hyphen.
>>   * A token consisting of a single hyphen character is interpreted as an ordinary non-option argument. By convention, it is used to specify input from or output to the standard input and output streams.
>>   * Options may be supplied in any order, or appear multiple times. The interpretation is left up to the particular application program. 
>
> GNU adds long options to these conventions. Long options consist of ‘--’ followed by a name made of alphanumeric characters and dashes. Option names are typically one to three words long, with hyphens to separate words. Users can abbreviate the option names as long as the abbreviations are unique.

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/encrypt.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/hashicorpvault"
 )
 
-func aesEncrypt(bytesToEncrypt []byte, aesKey []byte) (string, error) {
+func AesEncrypt(bytesToEncrypt []byte, aesKey []byte) (string, error) {
```
ocket8888, line 35: exported symbols should have GoDocs (e.g. is that error safe to show to a client, or only for logging server-side?) and nit but as an initialism AES should probably be capitalized

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import (
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/trafficvault/backends/postgres"
```
ocket8888, line 23: Instead of importing Traffic Ops internals, should the function you're using be pulled out into e.g. `lib/go-tc/`? Or some new library for crypto things like `lib/crypto/` maybe?

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	"crypto/aes"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
```
ocket8888, line 37: Standard imports should be first, followed by project-internal, then third party

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
+package main
```
ocket8888, line 1: commands should document their usage in the `main` package's GoDoc, for an example, check out [the GoDoc for `godoc` itself](https://github.com/golang/tools/blob/v0.1.2/cmd/godoc/doc.go).

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	flag.Parse()
+
+	if previousKeyLocation == nil || *previousKeyLocation == "" {
+		fmt.Println("previousKey flag is required.")
+		os.Exit(0)
+	}
+	if newKeyLocation == nil || *newKeyLocation == "" {
+		fmt.Println("newKey flag is required.")
+		os.Exit(0)
+	}
```
ocket8888, line 54: If these are required options, should they be positional instead of options? Normally options are optional.

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+const PROPERTIES_FILE = "./reencrypt.conf"
+
+func main() {
+	previousKeyLocation := flag.String("previousKey", "", "The file path for the previous base64 encoded AES key.")
+	newKeyLocation := flag.String("newKey", "", "The file path for the new base64 encoded AES key.")
+	flag.Parse()
+
+	if previousKeyLocation == nil || *previousKeyLocation == "" {
+		fmt.Println("previousKey flag is required.")
+		os.Exit(0)
```
ocket8888, line 49: Improper invocation should exit with a failure code, not success

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	if newKeyLocation == nil || *newKeyLocation == "" {
+		fmt.Println("newKey flag is required.")
+		os.Exit(0)
```
ocket8888, line 53: Same as above RE: failure exit code

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	newKey, err := readKey(*newKeyLocation)
+	if err != nil {
+		fmt.Println("reading newKey: ", err.Error())
+		os.Exit(0)
```
ocket8888, line 59: A failure should be indicated by the exit code, `0` is success.

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	previousKey, err := readKey(*previousKeyLocation)
+	if err != nil {
+		fmt.Println("reading previousKey: ", err.Error())
+		os.Exit(0)
```
ocket8888, line 65: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	dbConfBytes, err := ioutil.ReadFile(PROPERTIES_FILE)
+	if err != nil {
+		fmt.Println("reading db conf '", PROPERTIES_FILE, "': ", err.Error())
+		os.Exit(0)
```
ocket8888, line 71: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	pgCfg := Config{}
+	err = json.Unmarshal(dbConfBytes, &pgCfg)
+	if err != nil {
+		fmt.Println("unmarshalling '", PROPERTIES_FILE, "': ", err.Error())
+		os.Exit(0)
```
ocket8888, line 78: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	sslStr := "require"
+	if !pgCfg.SSL {
+		sslStr = "disable"
+	}
+	db, err := sqlx.Open("postgres", fmt.Sprintf("postgres://%s:%s@%s:%d/%s?sslmode=%s&fallback_application_name=trafficvault", pgCfg.User, pgCfg.Password, pgCfg.Hostname, pgCfg.Port, pgCfg.DBName, sslStr))
+	if err != nil {
+		fmt.Println("opening database: " + err.Error())
+		os.Exit(0)
```
ocket8888, line 88: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	if err = reEncryptSslKeys(db, previousKey, newKey); err != nil {
+		fmt.Println("re-encrypting SSL Keys: ", err.Error())
+		os.Exit(0)
```
ocket8888, line 93: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	if err = reEncryptUrlSigKeys(db, previousKey, newKey); err != nil {
+		fmt.Println("re-encrypting URL Sig Keys: ", err.Error())
+		os.Exit(0)
```
ocket8888, line 97: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	if err = reEncryptUriSigningKeys(db, previousKey, newKey); err != nil {
+		fmt.Println("re-encrypting URI Signing Keys: ", err.Error())
+		os.Exit(0)
```
ocket8888, line 101: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	if err = reEncryptDNSSECKeys(db, previousKey, newKey); err != nil {
+		fmt.Println("re-encrypting DNSSEC Keys: ", err.Error())
+		os.Exit(0)
```
ocket8888, line 105: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	if err = updateKeyFile(previousKey, *previousKeyLocation, newKey); err != nil {
+		fmt.Println("updating the key file: ", err.Error())
+		os.Exit(0)
```
ocket8888, line 110: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	return key, nil
+}
+
+func reEncryptSslKeys(db *sqlx.DB, previousKey []byte, newKey []byte) error {
+	tx, err := db.Begin()
+	if err != nil {
+		fmt.Println("transaction begin failed ", err, " ", tx)
+		os.Exit(0)
```
ocket8888, line 151: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	defer tx.Commit()
+
+	rows, err := tx.Query("SELECT id, data FROM sslkey")
+	if err != nil {
+		fmt.Println("querying: ", err)
+		os.Exit(0)
```
ocket8888, line 158: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		updateTx, err := db.Begin()
+		if err != nil {
+			fmt.Println("transaction begin failed ", err, " ", updateTx)
+			os.Exit(0)
```
ocket8888, line 166: Same as above RE: exit with failure code on failure

traffic_ops/app/db/reencrypt/reencrypt.go
ocket8888: Also, sometimes this function returns an error, and sometimes it forcibly exits the entire program - why not just always return the error and have the caller handle exiting or not?

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	return nil
+}
+
+func reEncryptUrlSigKeys(db *sqlx.DB, previousKey []byte, newKey []byte) error {
+	tx, err := db.Begin()
+	if err != nil {
+		fmt.Println("transaction begin failed ", err, " ", tx)
+		os.Exit(0)
```
ocket8888, line 212: Same as above RE: exit with failure code on failure and also returning error vs forcibly quitting.

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	return nil
+}
+
+func reEncryptUriSigningKeys(db *sqlx.DB, previousKey []byte, newKey []byte) error {
+	tx, err := db.Begin()
+	if err != nil {
+		fmt.Println("transaction begin failed ", err, " ", tx)
+		os.Exit(0)
```
ocket8888, line 273: Same as above RE: exit with failure code on failure and also returning an error versus forcibly quitting

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	return nil
+}
+
+func reEncryptDNSSECKeys(db *sqlx.DB, previousKey []byte, newKey []byte) error {
+	tx, err := db.Begin()
+	if err != nil {
+		fmt.Println("transaction begin failed ", err, " ", tx)
+		os.Exit(0)
```
ocket8888, line 334: Same as above RE: exit with failure code on failure and also returning an error versus forcibly quitting

traffic_ops/app/db/reencrypt/reencrypt.go
ocket8888: Alternatively, you can also just make them optional by providing default values. For example, most GNU/POSIX/Unix utilities, when they can read input from a file and output to a file, will, by default, read from `stdin` and output to `stdout`. `grep` is one such utility that will read from `stdin` if not given a file argument.

This allows for commands to make use of powerful Unix piping, so I could do something like `sudo cat /path/to/aes/key | reencrypt > /ouput/location` to read input from a file that is read-protected so only root can read it, or something like `reencript --previous-key=/path/to/aes/key | tee /output/path | wc -c` to write the key out and report the number of bytes written.

Of course, that stuff won't work if you print success messages and error messages to `stdout`. Normally, command output goes on `stdout` and error/informational/debugging output goes on `stderr`.

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	previousKey, err := readKey(*previousKeyLocation)
+	if err != nil {
+		fmt.Println("reading previousKey: ", err.Error())
+		os.Exit(0)
+	}
+
+	dbConfBytes, err := ioutil.ReadFile(PROPERTIES_FILE)
+	if err != nil {
```
ocket8888, line 69: I think we should consider making this location configurable via a command line argument

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,406 @@
...
+	var keyBase64 string
+	keyBase64Bytes, err := ioutil.ReadFile(keyLocation)
+	if err != nil {
+		return []byte{}, errors.New("reading file '" + keyLocation + "':" + err.Error())
+	}
+	keyBase64 = string(keyBase64Bytes)
+
+	key, err := base64.StdEncoding.DecodeString(keyBase64)
+	if err != nil {
+		return []byte{}, errors.New("AES key cannot be decoded from base64")
```
ocket8888, line 135: This should return the error as well, I think, to help diagnose the problem

traffic_ops/app/db/reencrypt/reencrypt.go
mattjackson220: ill update this next week. Steve also has a PR that will move them so ill work with him to make sure we get them in the same place to minimize rebasing headache

mattjackson220: done!

docs/source/admin/traffic_vault.rst
```diff
@@ -79,6 +79,40 @@ Administration of the PostgreSQL database for Traffic Vault
 
 Similar to administering the Traffic Ops database, the :ref:`admin <database-management>` tool should be used for administering the PostgreSQL Traffic Vault backend.
 
+AES Re-Encryption Tool:
```
ocket8888, line 82: Association of a heading with its content is done by virtue of it being a heading - you don't need the <kbd>:</kbd>

docs/source/admin/traffic_vault.rst
```diff
@@ -79,6 +79,40 @@ Administration of the PostgreSQL database for Traffic Vault
...
+-----------------------
+
+.. program:: reencrypt
+
+app/db/reencrypt/reencrypt
+=====================================================================
```
ocket8888, line 88: This heading level (`====`) places this heading above the previous section "AES Re-Encryption Tool:". In HTML terms, this is something like
```html
<section>
  <h1>Traffic Vault Administration</h1>
  <!-- some content... -->
   <section>
    <h2>PostgreSQL</h2>
    <!-- some content... -->
    <section>
      <h3>AES Re-Encryption Tool:</h3>
    </section>
  </section>
  <section>
    <h2>app/db/reencrypt/reencrypt</h2>
    <!-- some content... -->
  </section>
  <!-- some content -->
</section>
```
when I think what you actually want is something more like
```html
<section>
  <h1>Traffic Vault Administration</h1>
  <!-- some content... -->
  <section>
    <h2>PostgreSQL</h2>
    <!-- some content... -->
    <section>
      <h3>AES Re-Encryption Tool:</h3>
	  <section>
		<h4>app/db/reencrypt/reencrypt</h4>
		<!-- some content... -->
	  </section>
    </section>
	<!-- some content... -->
  </section>
  <!-- some content -->
</section>
```
although, the "AES Re-Encryption Tool:" section contains no content besides the content contained in its sub-section in that case, so really you don't need it, so this could just simply be structured:
```html
<section>
  <h1>Traffic Vault Administration</h1>
  <!-- some content... -->
  <section>
    <h2>PostgreSQL</h2>
    <!-- some content... -->
    <section>
	  <h3>app/db/reencrypt/reencrypt</h3>
	  <!-- some content... -->
    </section>
	<!-- some content... -->
  </section>
  <!-- some content -->
</section>
```

docs/source/admin/traffic_vault.rst
```diff
@@ -79,6 +79,40 @@ Administration of the PostgreSQL database for Traffic Vault
...
+
+	The file path for the new base64-encoded AES key.
+
+.. option:: --previous-key
+
+	The file path for the previous base64-encoded AES key.
+
+.. option:: --cfg
+
+	The path for the configuration file. Default is ``./reencrypt.conf``.
```
ocket8888, line 109: Options that take option-arguments should indicate so in their `.. option::` directive, e.g. `.. option:: --new-key NEW_KEY` - as opposed to options without option-arguments like `.. option:: -h, --help` (which, by the way, is an option I'd encourage all utilities to implement).

docs/source/admin/traffic_vault.rst
ocket8888, line 88: I don't think it's actually a problem for rendering, but this underline is much, much longer than the title it underlines

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+	var keyBase64 string
+	keyBase64Bytes, err := ioutil.ReadFile(keyLocation)
+	if err != nil {
+		return []byte{}, errors.New("reading file '" + keyLocation + "':" + err.Error())
+	}
+	keyBase64 = string(keyBase64Bytes)
+
+	key, err := base64.StdEncoding.DecodeString(keyBase64)
+	if err != nil {
+		return []byte{}, errors.New(fmt.Sprintf("AES key cannot be decoded from base64: %v", err))
```
ocket8888, line 153: instead of `errors.New(fmt.Sprintf(args...))` just use `fmt.Errorf(args...)` (and in that case formatting an error should use `%w` instead of `%v` to wrap the error)

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+		return []byte{}, err
+	}
+
+	return key, nil
+}
+
+func reEncryptSslKeys(db *sqlx.DB, previousKey []byte, newKey []byte) error {
+	tx, err := db.Begin()
+	if err != nil {
+		return errors.New(fmt.Sprintf("transaction begin failed %v %v ", err, tx))
```
ocket8888, line 168: same as above RE: `fmt.Errorf(args...)` instead of `errors.New(fmt.Sprintf(args...))`

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	defer tx.Commit()
+
+	rows, err := tx.Query("SELECT id, data FROM sslkey")
+	if err != nil {
+		return errors.New(fmt.Sprintf("querying: %v", err))
```
ocket8888, line 174: same as above RE: `fmt.Errorf(args...)` instead of `errors.New(fmt.Sprintf(args...))`

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		updateTx, err := db.Begin()
+		if err != nil {
+			return errors.New(fmt.Sprintf("transaction begin failed %v %v", err, updateTx))
```
ocket8888, line 181: same as above RE: `fmt.Errorf(args...)` instead of `errors.New(fmt.Sprintf(args...))`

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+		defer updateTx.Commit()
+
+		id := 0
+		var encryptedSslKeys []byte
+		if err = rows.Scan(&id, &encryptedSslKeys); err != nil {
+			return errors.New(fmt.Sprintf("getting SSL Keys: %v", err))
```
ocket8888, line 188: same as above RE: `fmt.Errorf(args...)` instead of `errors.New(fmt.Sprintf(args...))`

... seems there's a few of these, you get the idea

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+	savedKeyLocation := fmt.Sprintf("%s/savedEncryptionKey-%s.txt", filepath.Dir(previousKeyLocation), time.Now().Format(time.RFC3339))
+	if err := ioutil.WriteFile(savedKeyLocation, []byte(previousKeyStr), 0644); err != nil {
+		return errors.New("writing key file " + savedKeyLocation + ": " + err.Error())
+	}
+
+	return nil
+}
+
+func die(message string) {
+	fmt.Println(message)
```
ocket8888, line 391: Should this print to stderr instead of stdout?

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+		flag.Usage()
+		os.Exit(1)
+	}
+
+	if previousKeyLocation == nil || *previousKeyLocation == "" {
+		die("previous-key flag is required.")
+	}
+	if newKeyLocation == nil || *newKeyLocation == "" {
+		die("new-key flag is required.")
+	}
```
ocket8888, line 82: If these options aren't optional, should they be positional arguments?

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+
+	return nil
+}
+
+// updateKeyFile saves the new key in the same location as the old one so TO can use the new one.
+// It also saves the old one as a new file with a date to indicate when the re-encrypt was performed.
+func updateKeyFile(previousKey []byte, previousKeyLocation string, newKey []byte) error {
+	newKeyStr := base64.StdEncoding.EncodeToString(newKey)
+	if err := ioutil.WriteFile(previousKeyLocation, []byte(newKeyStr), 0644); err != nil {
```
rawlinp, line 371: I think the file permissions should be `0600` (ditto on the line below)

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+
+	sslStr := "require"
+	if !pgCfg.SSL {
+		sslStr = "disable"
+	}
+	db, err := sqlx.Open("postgres", fmt.Sprintf("postgres://%s:%s@%s:%d/%s?sslmode=%s&fallback_application_name=trafficvault", pgCfg.User, pgCfg.Password, pgCfg.Hostname, pgCfg.Port, pgCfg.DBName, sslStr))
+	if err != nil {
+		die("opening database: " + err.Error())
+	}
+
```
rawlinp, line 107: Rather than creating a transaction per table, we should create a single transaction here and use it for everything. That way, if any errors occur, the _entire_ change is rolled back, and we should never end up in a state where the TV DB was partially re-encrypted. That would be a bit of a mess 😃 .

Also, if an error occurs, we should rollback the transaction instead of committing, and if the transaction commit/rollback is deferred in this function, any paths that lead to calling `die` will skip any deferred functions.

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,393 @@
...
+	defer tx.Commit()
+
+	rows, err := tx.Query("SELECT id, data FROM sslkey")
+	if err != nil {
+		return errors.New(fmt.Sprintf("querying: %v", err))
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		updateTx, err := db.Begin()
```
rawlinp, line 173: Rather than creating two transactions, you can probably just read all of the rows into map. Then once all the rows have been read into the map, you should be able to use the same transaction to do the updates.

This comment applies to the other functions as well.

rawlinp (changes requested):

traffic_ops/app/db/reencrypt/reencrypt.go
mattjackson220: that was my thinking too especially since the inputs will be so similar i like having the flags to indicate which is which. but i added in defaults so we can still have the flags and also have them really be optional

docs/source/admin/traffic_vault.rst
```diff
@@ -79,6 +79,41 @@ Administration of the PostgreSQL database for Traffic Vault
...
+The :program:`app/db/reencrypt/reencrypt` binary is used to re-encrypt all data in the Postgres Traffic Vault with a new base64-encoded AES key.
+
+.. note:: For proper resolution of configuration files, it's recommended that this binary be run from the ``app/db/reencrypt`` directory.
+
+Usage
+-----
+``./reencrypt [options]``
+
+Options and Arguments
+---------------------
```
ocket8888, line 95: These headings all use the same heading level - I think what you want is for "Usage" and "Options and Arguments" to be subsections of "app/db/reencrypt/reencrypt". You might want to review the [docs' sectioning guide](https://traffic-control-cdn.readthedocs.io/en/latest/development/documentation_guidelines.html#section-headings).

docs/source/admin/traffic_vault.rst
```diff
@@ -79,6 +79,41 @@ Administration of the PostgreSQL database for Traffic Vault
 
 Similar to administering the Traffic Ops database, the :ref:`admin <database-management>` tool should be used for administering the PostgreSQL Traffic Vault backend.
 
+.. program:: reencrypt
+
+app/db/reencrypt/reencrypt
+--------------------------
+The :program:`app/db/reencrypt/reencrypt` binary is used to re-encrypt all data in the Postgres Traffic Vault with a new base64-encoded AES key.
```
ocket8888, line 86: the argument to `:program:` should be the same as the argument to the `.. program::` directive; in this case: `reencryt`

docs/source/admin/traffic_vault.rst
```diff
@@ -79,6 +79,53 @@ Administration of the PostgreSQL database for Traffic Vault
...
+This file deals with configuration of the Traffic Vault Database to be used with the :program:`reencrypt` tool.
+
+:dbname: The name of the PostgreSQL database used.
+:hostname: The hostname (:abbr:`FQDN (Fully Qualified Domain Name)`) of the server that runs the Traffic Vault Database.
+:password: The password to use when authenticating with the Traffic Vault database.
+:port: The port number (as a string) on which the Traffic Vault Database is listening for incoming connections (NOTE: the PostgreSQL default is 5432).
+:ssl: A boolean that sets whether or not the Traffic Vault Database encrypts its connections with SSL.
+:user: The name of the user as whom to connect to the database.
+
+
```
ocket8888, line 128: This all looks correct. Unfortunately, due to things you didn't change, the docs are having some issues:

```
 docs/source/admin/traffic_vault.rst#L172

Title level inconsistent:
```

etc.

The problem is that the existing structure is using:

```rst
**************
Document Title
**************

Section Header
==============

Subsection Header
-----------------

Sub-subsection Header
^^^^^^^^^^^^^^^^^^^^^

Aside
"""""
```

which isn't following the guidelines I linked in my last review. You can either change your <kbd>"</kbd> heading underlines to <kbd>^</kbd>, or fix the document to conform to the guidelines. The document should be updated, but it's not related to the work your doing, so you don't have to.

ocket8888 (approved):
My concerns were all addressed

docs/source/admin/traffic_vault.rst
```diff
@@ -79,6 +79,53 @@ Administration of the PostgreSQL database for Traffic Vault
...
+	./reencrypt --new-key ~/exampleNewKey.txt --previous-key ~/exampleOldKey.txt
+
+reencrypt.conf
+""""""""""""""
+This file deals with configuration of the Traffic Vault Database to be used with the :program:`reencrypt` tool.
+
+:dbname: The name of the PostgreSQL database used.
+:hostname: The hostname (:abbr:`FQDN (Fully Qualified Domain Name)`) of the server that runs the Traffic Vault Database.
+:password: The password to use when authenticating with the Traffic Vault database.
+:port: The port number (as a string) on which the Traffic Vault Database is listening for incoming connections (NOTE: the PostgreSQL default is 5432).
```
rawlinp, line 124: My bad, maybe this wasn't a total copy/paste of database.conf. This re-encrypt tool expects a string, not an int. For convenience/consistency, should we just make it parse as a string instead of an int? Then we can just reuse database.conf to run this tool.

rawlinp: Oh, nevermind, we won't be able to reuse database.conf since that's TODB not TVDB. It should parse as an int, so this line should not have `(as a string)`.

mattjackson220: oh dang it. ill get that note out of there. i think we should leave it as a separate conf file. we couldnt use database.conf (since thats for TO DB) but I thought about using the traffic vault info from cdn.conf and decided it would be better as a completely stand alone tool. this way we could reencrpyt a different db than is in use (like for testing or something) without impacting TO at all

mattjackson220: i should have refreshed the page before commenting back haha. we're on the same page

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,370 @@
...
+	if err = reEncryptUriSigningKeys(tx, previousKey, newKey); err != nil {
+		tx.Rollback()
+		die("re-encrypting URI Signing Keys: " + err.Error())
+	}
+	if err = reEncryptDNSSECKeys(tx, previousKey, newKey); err != nil {
+		tx.Rollback()
+		die("re-encrypting DNSSEC Keys: " + err.Error())
+	}
+
+	if err = updateKeyFile(previousKey, *previousKeyLocation, newKey); err != nil {
```
rawlinp, line 132: So after trying this out, I think this behavior might be a little surprising to the user. If there was only one copy of the original key in existence (which would be crazy I know), running this tool would erase it, making it impossible to roll back. Should we remove this step? It's easy enough to manually `mv` the files after.

traffic_ops/app/db/reencrypt/reencrypt.go
```diff
@@ -0,0 +1,370 @@
...
+
+	sslKeyMap := map[int][]byte{}
+
+	for rows.Next() {
+		id := 0
+		var encryptedSslKeys []byte
+		if err = rows.Scan(&id, &encryptedSslKeys); err != nil {
+			return fmt.Errorf("getting SSL Keys: %w", err)
+		}
+		jsonKeys, err := util.AESDecrypt(encryptedSslKeys, previousKey)
```
rawlinp, line 185: So as I was using this tool, I had a thought: what if we accidentally mixed up the keys -- e.g. passed the new key as `previous-key` and the old key as `new-key`? I think this might happily decrypt the data using the wrong key, which would result in garbage data, then re-encrypt the garbage data with the new key. I think we could reverse ourselves out of that situation if it happened, but we should probably include sanity-checks here between decrypting and re-encrypting. For example, making sure the decrypted data is actually JSON. Maybe we should try unmarshalling it into its struct and checking that the resulting struct isn't all empty?

rawlinp (changes requested):

traffic_ops/app/db/reencrypt/reencrypt.go
mattjackson220: It saves it out in the same location as the original key but with the name `savedEncryptionKey-` and then the date and time of re-encryption. so if you do only have one copy then it will still save it

traffic_ops/app/db/reencrypt/reencrypt.go
mattjackson220: if you pass the wrong key in as the `new-key` then it gets this error:
`re-encrypting SSL Keys: reading SSL Keys: cipher: message authentication failed` when trying to decrypt

mattjackson220: probably worth putting those checks in anyway though. its too important of data to risk

traffic_ops/app/db/reencrypt/reencrypt.go
mattjackson220: on second thought i also dont really care if it doesnt save it out / replace the old key. i thought it would be nice so you can stop TO, re-encrypt, and restart TO without much thought and save the annoyance of restarting TO with the wrong key location but as you said its not much effort to just `mv` it later. so i dont have a strong opinion either way if we want to change it back

rawlinp: Oh, I remember reading that in the code on my first review but forgot about that. I think doing an `mv` after is ok, then we don't also have to remove old backup key files later. If we were to automate this, we'd probably wrap it in a script that generates a new key and handles key backup and other stuff. Then re-encryption would basically be a single button.

traffic_ops/app/db/reencrypt/reencrypt.go
rawlinp: Oh, well that's good! I think when I tried this out I must've re-encrypted w/ the same key because I saved it off and reused it but it was overwritten. I'm like 0/3 today 🤕

rawlinp (approved):

PR #5923, opened 2021-06-08
Fix codeql t3c false positives

rob05c:
"Fixes" codeql false positives.

None of these are actual issues. Checking for overflows is deceptive and not useful. For example, the Parent Rank being `2147483649` is no more invalid than `2147483645`, but we don't have a hard limit on Parent Rank, so there's no sane max we can impose.

But this is the path of least resistance. Much as it frustrates me to make code worse for bad tools, this will make tools stop bothering us every few months to fix things they don't have the context to understand aren't issues.

No new tests, code already has tests, and any specific tests around the overflow would be deceptive, misleading, and fallacious.
No docs, no interface change.
No changelog, no interface change, and these aren't real bugs.

- Traffic Ops ORT

Run tests. Observe code is obviously identical in behavior, except for overflows astronomically larger than valid values.

Not a bug fix.

### DIFF
--- a/lib/go-atscfg/atscfg.go
+++ b/lib/go-atscfg/atscfg.go
@@ -205,11 +205,11 @@ func getATSMajorVersionFromATSVersion(atsVersion string) (int, error) {
 	}
 	majorVerStr := atsVersion[:dotPos]
 
-	majorVer, err := strconv.ParseUint(majorVerStr, 10, 64)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


ocket8888:
Alternatively, CodeQL scanning messages can just be dismissed as false positives by any committer

rob05c:
I almost did that. But I figured it'd just pop up again with some other "vet" tool or person in a few months

PR #5924, opened 2021-06-08
Add tool to migrate data between TV backends

shamrickus:

It adds a new go tool in `tools/traffic_vault_migrate` which can be used to migrate TV data from different backends. It is intended to connect directly to the backends such that a running TO/TV service is not needed.

It also moves the Postgres encryption functions to `lib/go-util` as well as adding godocs and tests.

The PR also removes the `id` column from `sslkey` and replaces the primary key with  the composite key: `(deliveryservice, cdn, version)`

- Documentation
- Traffic Ops
- Traffic Vault

Run the tool in both PSQL -> Riak and Riak -> PSQL mode. Verify that all the command line options work and return as intended. Also make sure the keys are actually migrated appropriately. Also make sure that this tool operates in a reasonable time frame when migrating a large amount of keys.

Read the documentation and confirm it makes sense.

Run the tests (for both the tool and `go-util/encrypt`) and confirm they pass.

## Additional Information
If testing PSQL -> Riak, I suggest using CiaB as both are already setup. The AES key can be got by running `docker exec -it cdn-in-a-box_trafficops_1 /bin/cat /opt/traffic_ops/app/conf/aes.key`
Riak -> PSQL, is a bit more work as it requires standing up your own psql db with the `app/db/admin` tool.


docs/source/tools/traffic_vault_migrate.rst
```diff
@@ -0,0 +1,107 @@
...
+
+ :password: The password for the user to log into the PG server.
+
+ :database: The database to connect to.
+
+ :port: The port on which the PG server is listening.
+
+ :host: The hostname of the PG server.
+
+ :sslmode: The ssl settings for the client connection, `explanation here <https://www.postgresql.org/docs/9.1/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS>`_. Options are 'disable', 'allow', 'prefer', 'require', 'verify-ca' and 'verify-full'
```
ocket8888, line 101: this links to the docs for an unsupported version of PostgreSQL, supported versions include  Current (13) / 12 / 11 / 10 / 9.6 - I'd suggest just linking to 13 for now, kicks the can as far down the road as possible

docs/source/tools/traffic_vault_migrate.rst
```diff
@@ -0,0 +1,107 @@
...
+
+		From server types (Riak|PG) (default "Riak")
+
+.. option:: -to_cfg
+
+		To server config file (default "pg.json")
+
+.. option:: -to_type
+
+		From server types (Riak|PG) (default "PG")
```
ocket8888, line 63: GNU-style long options should use `--` - according to POSIX `-dump` is equivalent to `-d -u -m -p` (Go's default command line argument parser is neither POSIX nor GNU-compliant, which is why generally I advise people use [pborman's `getopt` v2](https://pkg.go.dev/github.com/pborman/getopt/v2))

Also, this is just my personal opinion but I think kebab-case is superior to snake_case for command line utilities and their arguments. It's a GNU standard, and since we're not GNU we don't need to adhere to it, but it is just easier to type when you don't have to hold shift.

docs/source/tools/traffic_vault_migrate.rst
```diff
@@ -0,0 +1,107 @@
...
+
+ :host: The hostname of the PG server.
+
+ :sslmode: The ssl settings for the client connection, `explanation here <https://www.postgresql.org/docs/9.1/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS>`_. Options are 'disable', 'allow', 'prefer', 'require', 'verify-ca' and 'verify-full'
+
+ :aesKey: The base64 encoding of a 16, 24, or 32 bit AES key.
+
+Development
+=============
+To add a plugin, implement the traffic_vault_migrate.go:TVBackend interface and add the backend to the returned values in supportedBackends
```
ocket8888, line 107: To easily create a formatted link to a public Go symbol, you can use the [``:atc-godoc`` text role](https://traffic-control-cdn.readthedocs.io/en/latest/development/documentation_guidelines.html?highlight=atc-godoc#extension)

Also missing punctuation

docs/source/tools/traffic_vault_migrate.rst
ocket8888: Also options that accept option-arguments should indicate that in the argument to the `.. option` directive e.g. `.. option:: --from-cfg CFG`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+type PGBackend struct {
+	sslKey PGSSLKeyTable
+	dnssec PGDNSSecTable
+	uri    PGURISignKeyTable
+	url    PGURLSigKeyTable
+	cfg    PGConfig
+	db     *sql.DB
+}
+
+func (pg *PGBackend) String() string {
```
ocket8888, line 56: GoDoc? Also, does this receiver need to be a pointer?

(Typically for `func (this Thing) String() string` methods I just write "String implements the fmt.Stringer interface.")

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	data += fmt.Sprintf("\tDNSSec Keys: %v\n", len(pg.dnssec.Records))
+	data += fmt.Sprintf("\tURI Keys: %v\n", len(pg.uri.Records))
+	data += fmt.Sprintf("\tURL Keys: %v\n", len(pg.url.Records))
+	return data
+}
+func (pg *PGBackend) Name() string {
+	return "PG"
+}
+func (pg *PGBackend) ReadConfig(s string) error {
+	genericCfg, err := UnmarshalConfig(s, reflect.TypeOf(pg.cfg))
```
ocket8888, line 68: The type is known; we don't need to use reflection here, do we?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+
+func (pg *PGBackend) String() string {
+	data := fmt.Sprintf("PG server %v@%v:%v\n", pg.cfg.User, pg.cfg.Host, pg.cfg.Port)
+	data += fmt.Sprintf("\tSSL Keys: %v\n", len(pg.sslKey.Records))
+	data += fmt.Sprintf("\tDNSSec Keys: %v\n", len(pg.dnssec.Records))
+	data += fmt.Sprintf("\tURI Keys: %v\n", len(pg.uri.Records))
+	data += fmt.Sprintf("\tURL Keys: %v\n", len(pg.url.Records))
+	return data
+}
+func (pg *PGBackend) Name() string {
```
ocket8888, line 64: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	return "PG"
+}
+func (pg *PGBackend) ReadConfig(s string) error {
```
ocket8888, line 67: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+
+	pg.cfg = *genericCfg.Interface().(*PGConfig)
+
+	pg.cfg.AESKey, err = base64.StdEncoding.DecodeString(pg.cfg.Key)
+	if err != nil {
+		return fmt.Errorf("unable to decode PG AESKey: %v", err)
+	}
+	return nil
+}
+func (pg *PGBackend) Insert() error {
```
ocket8888, line 81: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	if err := pg.uri.insertKeys(pg.db); err != nil {
+		return err
+	}
+	return nil
+}
+func (pg *PGBackend) Start() error {
+	sqlStr := fmt.Sprintf("postgres://%s:%s@%s:%s/%s?sslmode=%s", pg.cfg.User, pg.cfg.Password, pg.cfg.Host, pg.cfg.Port, pg.cfg.Database, pg.cfg.SSLMode)
+	db, err := sql.Open("postgres", sqlStr)
+	if err != nil {
+		return fmt.Errorf("Unable to start PG client: %v", err)
```
ocket8888, line 100: `fmt.Errorf` format arg for errors should use `%w` to wrap errors

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	}
+	if err := pg.url.insertKeys(pg.db); err != nil {
+		return err
+	}
+	if err := pg.uri.insertKeys(pg.db); err != nil {
+		return err
+	}
+	return nil
+}
+func (pg *PGBackend) Start() error {
```
ocket8888, line 96: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+
+	pg.db = db
+	pg.sslKey = PGSSLKeyTable{}
+	pg.dnssec = PGDNSSecTable{}
+	pg.url = PGURLSigKeyTable{}
+	pg.uri = PGURISignKeyTable{}
+
+	return nil
+}
+func (pg *PGBackend) ValidateKey() []string {
```
ocket8888, line 111: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	}
+	if errs := pg.uri.validate(); errs != nil {
+		errors = append(errors, errs...)
+	}
+	if errs := pg.url.validate(); errs != nil {
+		errors = append(errors, errs...)
+	}
+	return errors
+}
+func (pg *PGBackend) Stop() error {
```
ocket8888, line 127: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	return pg.db.Close()
+}
+func (pg *PGBackend) Ping() error {
```
ocket8888, line 130: GoDoc?

tools/traffic_vault_migrate/postgres.go
ocket8888, line 127: Do you think this would benefit from being an [`io.Closer`](https://pkg.go.dev/io#Closer) interface instead of using `Stop`?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	return pg.db.Ping()
+}
+func (pg *PGBackend) Fetch() error {
```
ocket8888, line 133: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+}
+
+type PGDNSSecRecord struct {
+	Key tc.DNSSECKeysTrafficVault
+	CDN string
+	PGCommonRecord
+}
+type PGDNSSecTable struct {
+	Records []PGDNSSecRecord
+}
```
ocket8888, line 205: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	PGCommonRecord
+
+	// These records are stored on the table but are duplicated
+	DeliveryService string
+	Version         string
+	CDN             string
+}
+type PGSSLKeyTable struct {
+	Records []PGSSLKeyRecord
+}
```
ocket8888, line 314: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+
+type PGURLSigKeyRecord struct {
+	Keys            tc.URLSigKeys
+	DeliveryService string
+	PGCommonRecord
+}
+type PGURLSigKeyTable struct {
+	Records []PGURLSigKeyRecord
+}
```
ocket8888, line 431: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+
+type PGURISignKeyRecord struct {
+	Keys            map[string]tc.URISignerKeyset
+	DeliveryService string
+	PGCommonRecord
+}
+type PGURISignKeyTable struct {
+	Records []PGURISignKeyRecord
+}
```
ocket8888, line 537: GoDoc?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	PGCommonRecord
+}
+type PGDNSSecTable struct {
+	Records []PGDNSSecRecord
+}
+
+func (tbl *PGDNSSecTable) gatherKeys(db *sql.DB) error {
+	sz, err := getSize(db, "dnssec")
+	if err != nil {
+		log.Println("PGDNSSec gatherKeys: unable to determine size of dnssec table")
```
ocket8888, line 210: Should this log using `log.Errorln`?

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	}
+	tbl.Records = make([]PGDNSSecRecord, sz)
+
+	rows, err := db.Query("SELECT cdn, data from dnssec")
+	if err != nil {
+		return fmt.Errorf("PGDNSSec gatherKeys: unable to query: %v", err)
+	}
+	defer rows.Close()
```
ocket8888, line 218: nit: this is done commonly in our codebase, but it is throwing away an error return value. There's a convenience method provided by `github.com/apache/trafficcontrol/lib/go-log/` to log the error using `log.Error` if an error occurs - [`log.Close`](https://pkg.go.dev/github.com/apache/trafficcontrol/lib/go-log#Close) (or `log.Closef` to use format strings). Some static analysis tools won't detect that as deferring a Close on a Closer resource, though.

e.g.
```go
defer log.Close(rows, "Closing PGDNSSEC rows from gathering keys")
```

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+func (tbl *PGDNSSecTable) gatherKeys(db *sql.DB) error {
+	sz, err := getSize(db, "dnssec")
+	if err != nil {
+		log.Println("PGDNSSec gatherKeys: unable to determine size of dnssec table")
+	}
+	tbl.Records = make([]PGDNSSecRecord, sz)
+
+	rows, err := db.Query("SELECT cdn, data from dnssec")
+	if err != nil {
+		return fmt.Errorf("PGDNSSec gatherKeys: unable to query: %v", err)
```
ocket8888, line 216: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	}
+	defer rows.Close()
+	i := 0
+	for rows.Next() {
+		if i > len(tbl.Records)-1 {
+			return fmt.Errorf("PGDNSSec gatherKeys got more results than expected %v", len(tbl.Records))
+		}
+		err := rows.Scan(&tbl.Records[i].CDN, &tbl.Records[i].DataEncrypted)
+		if err != nil {
+			return fmt.Errorf("PGDNSSec gatherKeys unable to scan row: %v", err)
```
ocket8888, line 226: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+		i += 1
+	}
+	return nil
+}
+func (tbl *PGDNSSecTable) decrypt(aesKey []byte) error {
+	for i, _ := range tbl.Records {
+		data, err := decryptInto(aesKey, tbl.Records[i].DataEncrypted, reflect.TypeOf(tbl.Records[i].Key))
+		if err != nil {
+			return fmt.Errorf("unable to decrypt into keys: %v", err)
```
ocket8888, line 236: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+		tbl.Records[i].Key = *data.Interface().(*tc.DNSSECKeysTrafficVault)
+	}
+	return nil
+}
+func (tbl *PGDNSSecTable) encrypt(aesKey []byte) error {
+	for i, dns := range tbl.Records {
+		data, err := json.Marshal(&dns.Key)
+		if err != nil {
+			return fmt.Errorf("encrypt issue marshalling keys: %v", err)
```
ocket8888, line 246: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+		dat, err := encrypt(data, aesKey)
+		if err != nil {
+			return fmt.Errorf("encrypt error: %v", err)
```
ocket8888, line 250: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		queryArgs[j+1] = record.DataEncrypted
+		queryArgs[j+2] = record.CDN
+		queryArgs[j+3] = record.Version
+	}
+	return insertIntoTable(db, queryFmt, 4, queryArgs)
+}
+func (tbl *PGSSLKeyTable) gatherKeys(db *sql.DB) error {
+	sz, err := getSize(db, "sslkey")
+	if err != nil {
+		return fmt.Errorf("PGSSLKey gatherKeys unable to determine size of sslkey table: %v", err)
```
ocket8888, line 332: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	}
+	tbl.Records = make([]PGSSLKeyRecord, sz)
+
+	rows, err := db.Query("SELECT data, deliveryservice, cdn, version from sslkey")
+	if err != nil {
+		return fmt.Errorf("PGSSLKey gatherKeys unable to query: %v", err)
```
ocket8888, line 338: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	}
+	defer rows.Close()
+	i := 0
+	for rows.Next() {
+		if i > len(tbl.Records)-1 {
+			return errors.New("PGSSLKey gatherKeys: got more results than expected")
+		}
+		err := rows.Scan(&tbl.Records[i].DataEncrypted, &tbl.Records[i].DeliveryService, &tbl.Records[i].CDN, &tbl.Records[i].Version)
+		if err != nil {
+			return fmt.Errorf("PGSSLKey gatherKeys unable to scan row: %v", err)
```
ocket8888, line 348: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+		i += 1
+	}
+	return nil
+}
+func (tbl *PGSSLKeyTable) decrypt(aesKey []byte) error {
+	for i, dns := range tbl.Records {
+		data, err := decryptInto(aesKey, dns.DataEncrypted, reflect.TypeOf(dns.Keys))
+		if err != nil {
+			return fmt.Errorf("unable to decrypt into keys: %v", err)
```
ocket8888, line 358: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+		tbl.Records[i].Keys = *data.Interface().(*tc.DeliveryServiceSSLKeys)
+	}
+	return nil
+}
+func (tbl *PGSSLKeyTable) encrypt(aesKey []byte) error {
+	for i, dns := range tbl.Records {
+		data, err := json.Marshal(dns.Keys)
+		if err != nil {
+			return fmt.Errorf("encrypt issue marshalling keys: %v", err)
```
ocket8888, line 368: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+		dat, err := encrypt(data, aesKey)
+		if err != nil {
+			return fmt.Errorf("encrypt error: %v", err)
```
ocket8888, line 372: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+func (tbl *PGURLSigKeyTable) gatherKeys(db *sql.DB) error {
+	sz, err := getSize(db, "url_sig_key")
+	if err != nil {
+		log.Println("PGURLSigKey gatherKeys: unable to determine url_sig_key table size")
+	}
+	tbl.Records = make([]PGURLSigKeyRecord, sz)
+
+	rows, err := db.Query("SELECT deliveryservice, data from url_sig_key")
+	if err != nil {
+		return fmt.Errorf("PGURLSigKey gatherKeys error creating query: %v", err)
```
ocket8888, line 453: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	}
+	defer rows.Close()
+	i := 0
+	for rows.Next() {
+		if i > len(tbl.Records)-1 {
+			return fmt.Errorf("PGURLSigKey gatherKeys: got more results than expected %v", len(tbl.Records))
+		}
+		err := rows.Scan(&tbl.Records[i].DeliveryService, &tbl.Records[i].DataEncrypted)
+		if err != nil {
+			return fmt.Errorf("PGURLSigKey gatherKeys: unable to scan row: %v", err)
```
ocket8888, line 463: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+		i += 1
+	}
+	return nil
+}
+func (tbl *PGURLSigKeyTable) decrypt(aesKey []byte) error {
+	for i, sig := range tbl.Records {
+		data, err := decryptInto(aesKey, sig.DataEncrypted, reflect.TypeOf(sig.Keys))
+		if err != nil {
+			return fmt.Errorf("unable to decrypt into keys: %v", err)
```
ocket8888, line 473: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+
+		tbl.Records[i].Keys = *data.Interface().(*tc.URLSigKeys)
+	}
+	return nil
+}
+func (tbl *PGURLSigKeyTable) encrypt(aesKey []byte) error {
+	for i, sig := range tbl.Records {
+		data, err := json.Marshal(&sig.Keys)
+		if err != nil {
+			return fmt.Errorf("encrypt issue marshalling keys: %v", err)
```
ocket8888, line 484: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+
+		dat, err := encrypt(data, aesKey)
+		if err != nil {
+			return fmt.Errorf("encrypt error: %v", err)
```
ocket8888, line 489: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	}
+	defer rows.Close()
+	i := 0
+	for rows.Next() {
+		if i > len(tbl.Records)-1 {
+			return fmt.Errorf("PGURISignKey gatherKeys: got more results than expected %v", len(tbl.Records))
+		}
+		err := rows.Scan(&tbl.Records[i].DeliveryService, &tbl.Records[i].DataEncrypted)
+		if err != nil {
+			return fmt.Errorf("PGURISignKey gatherKeys: unable to scan row: %v", err)
```
ocket8888, line 569: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+func (tbl *PGURISignKeyTable) gatherKeys(db *sql.DB) error {
+	sz, err := getSize(db, "uri_signing_key")
+	if err != nil {
+		log.Println("PGURISignKey gatherKeys: unable to determine size of uri_signing_key table")
+	}
+	tbl.Records = make([]PGURISignKeyRecord, sz)
+
+	rows, err := db.Query("SELECT deliveryservice, data from uri_signing_key")
+	if err != nil {
+		return fmt.Errorf("PGURISignKey gatherKeys error while query: %v", err)
```
ocket8888, line 559: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+		}
+		i += 1
+	}
+	return nil
+}
+func (tbl *PGURISignKeyTable) decrypt(aesKey []byte) error {
+	for i, sign := range tbl.Records {
+		data, err := decryptInto(aesKey, sign.DataEncrypted, reflect.TypeOf(sign.Keys))
+		if err != nil {
+			return fmt.Errorf("unable to decrypt into keys: %v", err)
```
ocket8888, line 579: same as above RE: `%w` vs `%v`

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,716 @@
...
+	"errors"
+	"fmt"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	util "github.com/apache/trafficcontrol/lib/go-util"
+	_ "github.com/lib/pq"
+	"log"
+	"reflect"
+	"strconv"
+	"strings"
+)
```
ocket8888, line 35: import order should be

1. Standard Library
2. Project-internal (those under `github.com/apache/trafficcontrol/`)
3. Third-party libraries

tools/traffic_vault_migrate/postgres.go
ocket8888: I just realized you're using the standard `log` package, not `github.com/apache/trafficcontrol/lib/go-log` - so what I said doesn't apply unless you wanna switch libraries.

tools/traffic_vault_migrate/postgres.go
ocket8888: I just realized you're not using a `log` package that provides `Errorln` - what I mean, then, is should this be logging to stderr?

tools/traffic_vault_migrate/traffic_vault_migrate.go
```diff
@@ -0,0 +1,453 @@
...
+	// SetURISignKeys takes in keys and converts & encrypts the data into the backends internal format
+	SetURISignKeys([]URISignKey) error
+
+	// GetURLSigKeys converts the backends internal key representation into the common representation (URLSigKey)
+	GetURLSigKeys() ([]URLSigKey, error)
+	// SetURLSigKeys takes in keys and converts & encrypts the data into the backends internal format
+	SetURLSigKeys([]URLSigKey) error
+}
+
+type CommonRecord struct{}
```
ocket8888, line 402: What's the point of this type?

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/url_sig_keys.go
```diff
@@ -24,6 +24,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"github.com/apache/trafficcontrol/lib/go-util"
```
ocket8888, line 27: this should be grouped below with the other `github.com/apache/trafficcontrol/` import

traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/dsutil.go
```diff
@@ -23,6 +23,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"github.com/apache/trafficcontrol/lib/go-rfc"
```
ocket8888, line 26: This should be grouped below with the other `github.com/apache/trafficcontrol/` import

tools/traffic_vault_migrate/traffic_vault_migrate.go
shamrickus: Used to hold Last Updated, forgot to remove it.

tools/traffic_vault_migrate/postgres.go
shamrickus: For code reuse, and consistency. Figured loading configs is something that should be done the same for each backend (Riak also calls this).

ocket8888: It looks like the Riak backend calls it's own `ReadConfig` method that doesn't seem to be re-using this code - or vice-versa

shamrickus: ahhh...I misread your earlier comment, nope there is no reason.

tools/traffic_vault_migrate/postgres.go
ocket8888: I don't care about the pointer receiver thing, I meant to take that out of the comment. For a second I was like "well what if it's null?" - if that's true, then `nil.String()` is a segfault - that receiver type should never matter unless values are written.

tools/traffic_vault_migrate/traffic_vault_migrate.go
```diff
@@ -0,0 +1,448 @@
...
+		if err := toSrv.ReadConfig(toCfg); err != nil {
+			log.Fatalf("Unable to read toSrv cfg: %v", err)
+		}
+		toTime := time.Now().Sub(toTimer).Seconds()
+		log.Printf("Done [%v seconds]\n\tto: [%v seconds]\n\tfrom: [%v seconds]\n", toTime+fromTime, toTime, fromTime)
+	} else {
+		log.Printf("Done [%v seconds]\n", fromTime)
+	}
+
+	log.Println("Starting servers...")
```
rawlinp, line 111: Should this be "starting server connections..." instead?

tools/traffic_vault_migrate/traffic_vault_migrate.go
```diff
@@ -0,0 +1,448 @@
...
+		log.Fatalf("Unable to read fromSrv cfg: %v", err)
+	}
+	fromTime = time.Now().Sub(fromTimer).Seconds()
+
+	if toSrvUsed {
+		toTimer = time.Now()
+		if err := toSrv.ReadConfig(toCfg); err != nil {
+			log.Fatalf("Unable to read toSrv cfg: %v", err)
+		}
+		toTime := time.Now().Sub(toTimer).Seconds()
```
rawlinp, line 105: All this timing code is a bit distracting -- is it really necessary to know how long each step takes? Also, if we really needed to know, we should be able to use the log timestamps from "starting ..." to the next "starting..." message, right?

tools/traffic_vault_migrate/traffic_vault_migrate.go
```diff
@@ -0,0 +1,448 @@
...
+		return errors.New(fmt.Sprintf("Unable to set %v url keys: %v", be.Name(), err))
+	}
+	if err := be.SetURISignKeys(uriKeys); err != nil {
+		return errors.New(fmt.Sprintf("Unable to set %v uri keys: %v", be.Name(), err))
+	}
+	return nil
+}
+
+// GetKeys will get all of the keys for a backend
+func GetKeys(be TVBackend) ([]SSLKey, []DNSSecKey, []URISignKey, []URLSigKey, error) {
```
rawlinp, line 312: this function is a bit unwieldy with 5 outputs -- could we lump the keys altogether into a `type Secrets struct {SSLKeys, ...}`?

tools/traffic_vault_migrate/traffic_vault_migrate.go
```diff
@@ -0,0 +1,448 @@
...
+
+	if err := Validate(fromSrv); err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("Done [%v seconds]\n", time.Now().Sub(fromTimer).Seconds())
+
+	if dump {
+		log.Printf("Dumping data from %v...\n", fromSrv.Name())
+		fromTimer = time.Now()
+		sslKeysBytes, err := json.Marshal(&fromSSLKeys)
```
rawlinp, line 170: There's a pattern here -- we're doing the same thing 4 times in a row: marshalling keys and writing to a file. Can we condense this into a loop that iterates over a slice of structs to marshal the json and write to disk?

rawlinp (changes requested):
partial review since I ran out of time -- I'll be reviewing the rest tomorrow

docs/source/tools/traffic_vault_migrate.rst
shamrickus: Changed

tools/traffic_vault_migrate/postgres.go
shamrickus: Don't see the harm, renamed it.

tools/traffic_vault_migrate/traffic_vault_migrate.go
shamrickus: You are correct, I removed them.

tools/traffic_vault_migrate/riak.go
```diff
@@ -0,0 +1,746 @@
...
+	if err != nil {
+		return errors.New("failed to build riak ping command: " + err.Error())
+	}
+
+	if err = cluster.Execute(cmd); err != nil {
+		return err
+	}
+	response, ok := cmd.(*riak.PingCommand)
+	if !ok {
+		return errors.New(fmt.Sprintf("unexpected riak command type for ping: %v", cmd))
```
rawlinp, line 703: nit: `fmt.Errorf`

rawlinp (changes requested):
Sorry for the large number of comments -- this is a large PR and they are mostly the same nits. The type-specific formatting comments are due to the fact that reliance on `%v` for everything tends to lead to things like accidentally logging pointer addresses instead of the dereferenced values, so it's good practice to use the right formatter when you know the type.

I also had a bunch of comments about enhancing error messages, because we will want issues to be easy to troubleshoot if they arise.

Also, we should probably compile and include this tool in the TO RPM, similar to what Matt is doing in [his PR](https://github.com/apache/trafficcontrol/pull/5913/files#diff-183480071dfecfff7d87f61764581afb39c82cba7c9fcb7e35ffc756fd7d90f4R78).

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,743 @@
...
+
+		queryArgs[j+4] = record.DeliveryService
+		queryArgs[j+5] = record.DataEncrypted
+		queryArgs[j+6] = record.CDN
+		queryArgs[j+7] = "latest"
+	}
+	return insertIntoTable(db, queryBase, 4, queryArgs)
+}
+func (tbl *pgSSLKeyTable) gatherKeys(db *sql.DB) error {
+	sz, err := getSize(db, "sslkey WHERE version='latest'")
```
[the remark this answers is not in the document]
shamrickus: The other counterpart to each latest record is identical, it is inserted twice for searching/indexing.The reason I chose to only get latest records was that otherwise there are duplicate `SSLKey` records.

tools/traffic_vault_migrate/riak.go
```diff
@@ -0,0 +1,746 @@
...
+	if err != nil {
+		return nil, err
+	}
+	for _, key := range keys {
+		objects, err := getObject(cluster, bucket, key)
+		if err != nil {
+			return nil, err
+		}
+		if len(objects) > 1 {
+			return nil, fmt.Errorf("Unexpected number of objects %v, ignoring\n", len(objects))
```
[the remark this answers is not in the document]
shamrickus: Good catch, used to be a log with a continue.

tools/traffic_vault_migrate/postgres.go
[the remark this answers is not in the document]
rawlinp: Right, if we only care to get the latest version of an sslkey, then getting `latest` is fine. But I think we might want to migrate all _prior_ versions too.

For instance, let's say a DS is on ssl_key_version 5. So it currently has all these sslkey versions stored:
- 0
- 1
- 2
- 3
- 4
- 5
- latest

Versions `5` and `latest` are the same sslkey, but we might also want to migrate versions `0` through `4`.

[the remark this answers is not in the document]
shamrickus: Hmmm...I was misreading how we inserted ssl keys, you are correct. I'll remove the filter.

tools/traffic_vault_migrate/postgres.go
shamrickus: Switched to `go-log`

tools/traffic_vault_migrate/traffic_vault_migrate.go
```diff
@@ -423,6 +422,17 @@ type URLSigKey struct {
 	tc.URLSigKeys
 }
 
+func writeKeys(filename string, data interface{}) error {
+	bytes, err := json.Marshal(&data)
+	if err != nil {
+		return err
+	}
+	if err = ioutil.WriteFile(filename, bytes, 0640); err != nil {
```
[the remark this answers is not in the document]
shamrickus: Removed all group perms

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,743 @@
...
+	data += fmt.Sprintf("\tURL Keys: %v\n", len(pg.url.Records))
+	return data
+}
+
+// Name returns the name for this backend
+func (pg *PGBackend) Name() string {
+	return "PG"
+}
+
+// ReadConfig takes in a filename and will read it into the backends config
```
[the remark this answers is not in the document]
shamrickus: It does take in a file name (both backends call `UnmarshalConfig` which calls `ioutil.ReadFile`. I've renamed this fn to `ReadConfigFile` and I'll also change `toCfg/fromCfg` to `toCfgPath/fromCfgPath` so that it's more clear what they contain.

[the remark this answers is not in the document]
rawlinp: My bad -- I was reading `UnmarshalConfig` as `UnmarshalJSON`, so I thought `s` was actually the file's contents. Please disregard 😄

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,743 @@
...
+	Keys            map[string]tc.URISignerKeyset
+	DeliveryService string
+	pgCommonRecord
+}
+type pgURISignKeyTable struct {
+	Records []pgURISignKeyRecord
+}
+
+func (tbl *pgURISignKeyTable) insertKeys(db *sql.DB) error {
+	queryBase := "INSERT INTO uri_signing_key (deliveryservice, data) VALUES "
```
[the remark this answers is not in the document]
rawlinp: Expensive in terms of dev time or runtime? And do you mean skip it by `ON CONFLICT IGNORE` or something like that?

If we're talking expensive as in runtime, doing a migration should be pretty rare. If we're successful, we might only use this tool once. If we aren't successful at first, maybe we'll run it 3 times if we decide to un-migrate then re-migrate once the issue is fixed. So if it takes a little longer to run, that's probably not a big deal.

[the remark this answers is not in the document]
shamrickus: `ON CONFLICT` doesn't work because the only constrains on the sslkey table are on the `id` column. Regardless, I'll handle it for sslkeys as well.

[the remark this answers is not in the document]
rawlinp: Oh, I see what you're saying. We could currently have multiple sslkeys at the same version in the DB, because `version` is not part of the primary key?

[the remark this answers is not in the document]
shamrickus: Yep, at least from the DBs perspective; I don't think we can actually do that through the TO API.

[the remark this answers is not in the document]
rawlinp: Well, we can still change the schema if you'd like -- since it's unreleased we don't necessarily need a migration and can just update the create_schema.sql file.

docs/source/tools/traffic_vault_migrate.rst
```diff
@@ -0,0 +1,112 @@
...
+
+.. option:: -m, --noConfirm
+
+		Do not require confirmation before inserting records
+
+.. option:: -r, --dry
+
+		Do not perform writes. Will do a basic output of the keys on the 'from' backend.
+
+.. option:: -t, --fromType=TYPE
```
ocket8888, line 65: options that take option-arguments and have long and short forms should specify it for both forms e.g.
```rst
.. option:: -t TYPE, --from-type TYPE
```

The <kbd>=</kbd> can't be used with the short form, but using a space in the short form and an <kbd>=</kbd> in the long form doesn't hurt anything afaik.

tools/traffic_vault_migrate/postgres.go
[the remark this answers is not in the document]
shamrickus: That would be preferable, makes it so I don't have to use a CTE. I'll add it to this PR (otherwise testing the psql backend would be impossible).

lib/go-tc/deliveryservice_ssl_keys.go
```diff
@@ -73,6 +73,7 @@ type SSLKeyRequestFields struct {
 	HostName     *string `json:"hostname,omitempty"`
 	Country      *string `json:"country,omitempty"`
 	State        *string `json:"state,omitempty"`
+	Version      *int    `json:"version,omitempty"`
```
ocket8888, line 76: I'm not sure we can add this without making a new-version structure, because this is used in the APIv3 client method `GenerateSSLKeysForDS` which has been published. That's what @rob05c would say, I think. @rawlinp will disagree, saying that it won't break existing clients because they couldn't have been using it anyway - which is true.

Personally I don't care either way, but would find it easier to just make an `SSLKeyRequestV40` and `SSLKeyRequestV4` alias as it's a small enough change and won't hurt anything even if it's unnecessary. But if rob doesn't care I don't think anyone else will, so in that case don't worry about it.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+	"strings"
+
+	"github.com/apache/trafficcontrol/lib/go-log"
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	util "github.com/apache/trafficcontrol/lib/go-util"
+
+	_ "github.com/lib/pq"
+)
+
+// PGConfig represents the configuration options available to the PG backend
```
ocket8888, line 38: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+	Port     string `json:"port"`
+	User     string `json:"user"`
+	Password string `json:"password"`
+	SSLMode  string `json:"sslmode"`
+	Database string `json:"database"`
+	Key      string `json:"aesKey"`
+	AESKey   []byte
+}
+
+// PGBackend is the Postgres implementation of TVBackend
```
ocket8888, line 50: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+type PGBackend struct {
+	sslKey pgSSLKeyTable
+	dnssec pgDNSSecTable
+	uri    pgURISignKeyTable
+	url    pgURLSigKeyTable
+	cfg    PGConfig
+	db     *sql.DB
+}
+
+// String returns a high level overview of the backend and its keys
```
ocket8888, line 60: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+func (pg *PGBackend) String() string {
+	data := fmt.Sprintf("PG server %s@%s:%s\n", pg.cfg.User, pg.cfg.Host, pg.cfg.Port)
+	data += fmt.Sprintf("\tSSL Keys: %d\n", len(pg.sslKey.Records))
+	data += fmt.Sprintf("\tDNSSec Keys: %d\n", len(pg.dnssec.Records))
+	data += fmt.Sprintf("\tURI Keys: %d\n", len(pg.uri.Records))
+	data += fmt.Sprintf("\tURL Keys: %d\n", len(pg.url.Records))
+	return data
+}
+
+// Name returns the name for this backend
```
ocket8888, line 70: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+func (pg *PGBackend) Name() string {
+	return "PG"
+}
+
+// ReadConfigFile takes in a filename and will read it into the backends config
```
ocket8888, line 75: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+		return err
+	}
+
+	if pg.cfg.AESKey, err = base64.StdEncoding.DecodeString(pg.cfg.Key); err != nil {
+		return fmt.Errorf("unable to decode PG AESKey: %w", err)
+	}
+	return nil
+}
+
+// Insert takes the current keys and inserts them into the backend DB
```
ocket8888, line 88: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+	if err := pg.url.insertKeys(pg.db); err != nil {
+		return err
+	}
+	if err := pg.uri.insertKeys(pg.db); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Start initiates the connection to the backend DB
```
ocket8888, line 105: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+	pg.db = db
+	pg.sslKey = pgSSLKeyTable{}
+	pg.dnssec = pgDNSSecTable{}
+	pg.url = pgURLSigKeyTable{}
+	pg.uri = pgURISignKeyTable{}
+
+	return nil
+}
+
+// ValidateKey validates that the keys are valid (in most cases, certain fields are not null)
```
ocket8888, line 122: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+	if errs := pg.uri.validate(); errs != nil {
+		errs = append(errs, errs...)
+	}
+	if errs := pg.url.validate(); errs != nil {
+		errs = append(errs, errs...)
+	}
+	return errs
+}
+
+// Close terminates the connection to the backend DB
```
ocket8888, line 140: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+func (pg *PGBackend) Close() error {
+	return pg.db.Close()
+}
+
+// Ping checks the connection to the backend DB
```
ocket8888, line 145: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+func (pg *PGBackend) Ping() error {
+	return pg.db.Ping()
+}
+
+// Fetch gets all of the keys from the backend DB
```
ocket8888, line 150: nit: GoDoc missing punctuation.

tools/traffic_vault_migrate/postgres.go
```diff
@@ -0,0 +1,742 @@
...
+
+// GetURLSigKeys converts the backends internal key representation into the common representation (URLSigKey)
+func (pg *PGBackend) GetURLSigKeys() ([]URLSigKey, error) {
+	if err := pg.url.decrypt(pg.cfg.AESKey); err != nil {
+		return nil, err
+	}
+	return pg.url.toGeneric(), nil
+}
+
+// SetURLSigKeys takes in keys and converts & encrypts the data into the backends internal format
```
ocket8888, line 221: nit: GoDoc missing punctuation.

lib/go-tc/deliveryservice_ssl_keys.go
rawlinp: I believe we are free to add new fields to existing structs -- that is a non-breaking change.

ocket8888: > I believe we are free to add new fields to existing structs -- that is a non-breaking change.

Not only that, but it looks like the "new" field was used in older API versions already, the only difference is that the Go client used to hard-code it. So this is barely a **non**-breaking change.

rob05c: +1 - as long as it has a sane default, ideally as close to the prior behavior as possible.

It technically can break code, e.g. `foo := Foo{a,b,c}` will break. But it's reasonable to add fields in minor versions, generally speaking, and any of the alternatives in Go aren't reasonable.

So, as the person who's usually one of the biggest votes against breaking clients, IMO we should consider adding fields as non-breaking, and discourage clients from using that syntax (and the few others which break, which all have alternatives that don't).

tools/traffic_vault_migrate/traffic_vault_migrate.go
```diff
@@ -400,6 +356,49 @@ type TVBackend interface {
...
+		return s.dnssecKeys[a].CDN < s.dnssecKeys[b].CDN
+	})
+	sort.Slice(s.uriKeys[:], func(a, b int) bool {
+		return s.uriKeys[a].DeliveryService < s.uriKeys[b].DeliveryService
+	})
+	sort.Slice(s.urlKeys[:], func(a, b int) bool {
+		return s.urlKeys[a].DeliveryService < s.urlKeys[b].DeliveryService
+	})
+}
+func (s *Secrets) dump(directory string) {
```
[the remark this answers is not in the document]
shamrickus: Added an option `fill` to import data.

tools/traffic_vault_migrate/postgres.go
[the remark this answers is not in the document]
shamrickus: I also ended up removing the `id` column as it appears that the only use for it was for the unique constraint.

tools/traffic_vault_migrate/riak.go
```diff
@@ -0,0 +1,746 @@
...
+		return nil, fmt.Errorf("errors executing riak fetch value command; key not found: %v:%v", bucket, key)
+	}
+
+	return rsp.Values, nil
+}
+func setObject(cluster *riak.Cluster, obj *riak.Object, bucket string) error {
+	cmd, err := riak.NewStoreValueCommandBuilder().
+		WithBucket(bucket).
+		WithContent(obj).
+		WithTimeout(time.Second * 5).
```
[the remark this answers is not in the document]
shamrickus: Now done via the config file.

tools/traffic_vault_migrate/riak.go
```diff
@@ -0,0 +1,746 @@
...
+		if start == 0 {
+			if cmd.Response == nil || cmd.Response.NumFound == 0 {
+				return nil, nil
+			}
+			if cmd.Response.NumFound <= numRows {
+				return cmd.Response.Docs, nil
+			} else if numRows < cmd.Response.NumFound {
+				searchDocs = make([]*riak.SearchDoc, cmd.Response.NumFound)
+			}
+			if cmd.Response.NumFound > 10000 {
```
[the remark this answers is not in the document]
shamrickus: Set it to `numRows*10`

traffic_ops/app/db/traffic_vault_migrate/riak.go
```diff
@@ -0,0 +1,749 @@
...
+		if record.Version == "" {
+			errs = append(errs, fmt.Sprintf("SSL Key '%s': Version is blank!", record.Key))
+		}
+	}
+	return errs
+}
+
+type riakDNSSecKeyRecord struct {
+	CDN string
+	Key tc.DNSSECKeysRiak
```
rawlinp, line 341: nit: `tc.DNSSECKeysRiak` is deprecated, `tc.DNSSECKeysTrafficVault` is the same thing and could be used instead

rawlinp (approved):
This looks great -- all I could find were nitpicks 👍

PR #5926, opened 2021-06-09
Add t3c-apply flag to wait for parents in syncds mode

rob05c:
Adds a flag to `t3c-apply` to wait for parents in syncds mode. Currently, `--wait-for-parents=true` is the default, but doesn't actually wait in syncds mode, only reval mode.

This changes the flag to a ternary: true, false, and reval. The `reval` option is the new default, and behaves like the old default, so most users should see no change: it waits for parents in reval mode, but not syncds. And `true` now always waits.

The default `reval` is still the recommended option. Config deployments other than revalidations _probably_ don't need to wait for parents in most cases. But config deployment is so complex, it's impossible to say for sure that there are no cases that require waiting for parents. Which is why this flag is being added: so if anyone does encounter an case where it's required (potentially outage-inducing every time there's a change), it's possible to wait for parents in syncds with a flag change, rather than requiring a code change and a new deploy. And it's very little code for that safety, just a new flag and slight modification to the conditional.

Includes extensive tests for all 3 modes (previously wait-for-parents had no tests).
Includes docs.
Includes changelog.

- Traffic Ops ORT

Run t3c integration tests. Run t3c-apply with various wait-for-parents arguments, with a cache with parents pending, verify flags do what they say.

Not a bug fix.

## Additional Information


traeak (approved):
These changes all look fine, unit test looks good.  Tested with both syncds and revalidations on a running ats configuration.

PR #5930, opened 2021-06-10
Locking down snap/ queue/ update status endpoints for a CDN

srijeet0406:

This PR adds the cdn locking capability to the snap/ queue/ update status endpoints.

- Traffic Ops
- CI tests

Run TO and TP locally.
As user `A`, send an API request to grab the lock for a particular CDN `foo`.
Now, try to snap `foo` as user `A` -> this should succeed
After this, log in to TO as another user `B`, and try to snap `foo` as user `B` -> this should fail because `A` has the lock for `foo`
Try to snap any other CDN as user `B` or `A` -> both operations should pass
Repeat the same tests for queueing and updating status of he servers on a CDN

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v4/cdn_locks_test.go
+++ b/traffic_ops/testing/api/v4/cdn_locks_test.go
@@ -65,7 +65,7 @@ func getCDNNameAndServerID(t *testing.T) (string, int) {
 		if err != nil {
 			t.Errorf("could not get servers for cdn %s: %v", cdn.Name, err)
 		}
-		if len(serversResp.Response) != 0 {
+		if len(serversResp.Response) != 0 && serversResp.Response[0].ID != nil {
 			serverID = *serversResp.Response[0].ID
 			return cdn.Name, serverID
 		}
@@ -103,6 +103,7 @@ func getCDNDetailsAndTopologyName(t *testing.T) (int, string, string) {
 			}
 		}
 	}
+	t.Error("couldn't find a valid CDN and associated Topology name")
 	return -1, "", ""
 }
 
@@ -325,13 +326,13 @@ func SnapshotWithLock(t *testing.T) {
 		t.Errorf("expected error while snapping cdn %s by user admin, but got nothing", cdn)
 	}
 	if reqInf.StatusCode != http.StatusForbidden {
-		t.Fatalf("expected a 403 status code, but got %d instead", reqInf.StatusCode)
+		t.Errorf("expected a 403 status code, but got %d instead", reqInf.StatusCode)
 	}
 
 	// Delete the lock
 	_, _, err = userSession.DeleteCDNLocks(client.RequestOptions{QueryParameters: url.Values{"cdn": []string{cdn}}})
 	if err != nil {
-		t.Fatalf("expected no error while deleting other user's lock using admin endpoint, but got %v", err)
+		t.Errorf("expected no error while deleting other user's lock using admin endpoint, but got %v", err)
 	}
 }
 
--- a/traffic_ops/testing/api/v4/cdn_locks_test.go
+++ b/traffic_ops/testing/api/v4/cdn_locks_test.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strconv"
 	"testing"
 	"time"
 
@@ -28,9 +29,12 @@ import (
 )
 
 func TestCDNLocks(t *testing.T) {
-	WithObjs(t, []TCObj{Tenants, Roles, Users, CDNs}, func() {
+	WithObjs(t, []TCObj{Types, CacheGroups, CDNs, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, Servers, ServerCapabilities, ServerServerCapabilitiesForTopologies, Topologies, Tenants, DeliveryServices, TopologyBasedDeliveryServiceRequiredCapabilities, Roles, Users}, func() {
 		CRDCdnLocks(t)
 		AdminCdnLocks(t)
+		SnapshotWithLock(t)
+		QueueUpdatesWithLock(t)
+		QueueUpdatesFromTopologiesWithLock(t)
 	})
 }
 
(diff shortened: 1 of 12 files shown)


traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -45,6 +49,63 @@ func getCDNName(t *testing.T) string {
...
+	}
+	for _, cdn := range cdnResp.Response {
+		opts := client.NewRequestOptions()
+		opts.QueryParameters.Set("cdn", strconv.Itoa(cdn.ID))
+		serversResp, _, err := TOSession.GetServers(opts)
+		if err != nil {
+			t.Errorf("could not get servers for cdn %s: %v", cdn.Name, err)
+		}
+		if len(serversResp.Response) != 0 {
+			serverID = *serversResp.Response[0].ID
```
ocket8888, line 69: This will segfault if Traffic Ops returns a malformed response that has a `null` or undefined server ID

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -45,6 +49,63 @@ func getCDNName(t *testing.T) string {
...
+				if err != nil {
+					t.Errorf("couldn't get servers: %v", err)
+				}
+				if len(serversResp.Response) != 0 && serversResp.Response[0].CDNName != nil && serversResp.Response[0].CDNID != nil {
+					return *serversResp.Response[0].CDNID, *serversResp.Response[0].CDNName, top.Name
+				}
+			}
+		}
+	}
+	return -1, "", ""
```
ocket8888, line 106: should this mark the test as failed?

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -199,3 +260,227 @@ func AdminCdnLocks(t *testing.T) {
...
+		t.Errorf("expected no error while snapping cdn %s by user %s, but got %v", cdn, *user1.Username, err)
+	}
+
+	// Admin user, which doesn't have the lock on the CDN "bar", is trying to snap it -> this should fail
+	_, reqInf, err := TOSession.SnapshotCRConfig(opts)
+	if err == nil {
+		t.Errorf("expected error while snapping cdn %s by user admin, but got nothing", cdn)
+	}
+	if reqInf.StatusCode != http.StatusForbidden {
+		t.Fatalf("expected a 403 status code, but got %d instead", reqInf.StatusCode)
```
ocket8888, line 328: Is this error fatal? I think the rest of the test can proceed if this condition isn't met

traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -199,3 +260,227 @@ func AdminCdnLocks(t *testing.T) {
...
+	}
+
+	// Delete the lock
+	_, _, err = userSession.DeleteCDNLocks(client.RequestOptions{QueryParameters: url.Values{"cdn": []string{cdn}}})
+	if err != nil {
+		t.Fatalf("expected no error while deleting other user's lock using admin endpoint, but got %v", err)
```
ocket8888, line 334: I don't think this needs to be fatal; it isn't doing anything after this anyway

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -101,6 +101,31 @@ const getUserByEmailQuery = getUserBaseQuery + `
...
 `
 
+// CheckIfCurrentUserHasCdnLock checks if the current user has the lock on the cdn that the requested operation is to be performed on.
+// This will succeed if the either there is no lock by any user on the CDN, or if the current user has the lock on the CDN.
+func CheckIfCurrentUserHasCdnLock(tx *sql.Tx, cdn, user string) (error, error, int) {
+	query := `SELECT username FROM cdn_lock WHERE cdn=$1`
+	var userName string
+	rows, err := tx.Query(query, cdn)
+	if err != nil {
+		if err == sql.ErrNoRows {
```
ocket8888, line 111: checking for specific errors should use `errors.Is` to unwrap things

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -534,6 +559,15 @@ func GetCDNNameFromID(tx *sql.Tx, id int64) (tc.CDNName, bool, error) {
 	return tc.CDNName(name), true, nil
 }
 
+// GetCDNNameFromServerID gets the CDN name for the server with the given ID
+func GetCDNNameFromServerID(tx *sql.Tx, serverId int64) (tc.CDNName, error) {
+	name := ""
+	if err := tx.QueryRow(`SELECT name FROM cdn WHERE id = (SELECT cdn_id FROM server WHERE id=$1)`, serverId).Scan(&name); err != nil {
+		return "", errors.New("querying CDN name from server ID : " + err.Error())
```
ocket8888, line 566: if you use `fmt.Errorf("querying CDN name from server ID: %w", err)` instead will wrap the actual error, that way you can still add context, but if the caller wants to they can check if the returned error is a specific kind of error for specific handling e.g.
```go
cdn, err := dbhelpers.GetCDNNameFromServerID(tx, serverID)
if err != nil {
	if errors.Is(err, sql.ErrNoRows) {
		// handle this with maybe a 404 response or something
	} else {
		// more generic error handling
	}
}
```

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -534,6 +559,15 @@ func GetCDNNameFromID(tx *sql.Tx, id int64) (tc.CDNName, bool, error) {
 	return tc.CDNName(name), true, nil
 }
 
+// GetCDNNameFromServerID gets the CDN name for the server with the given ID
```
ocket8888, line 562: nit: GoDoc missing punctuation

ocket8888 (approved):
Functionality works as described, the only failing tests are the ones that are failing currently on master due to known locale issues.

PR #5940, opened 2021-06-15
Migrate division test from E2E to Integration

rimashah25:

**Note**: Most of the tests were already migrated. There was one remaining and this PR fixes that.

- Traffic Portal
- CI tests

Required: Local TO or CIAB TO running

Go to traffic_portal/test/integration 
Update specs, apiUrl and baseUrl in config.json as follows:
```
"specs": [
    "specs/Divisions.spec.js"
  ],

"params": {
    "apiUrl": " https://localhost:9090/api/4.0",
    "baseUrl": "https://localhost:9090/",
....
}
```
and then run the below two commands in a separate terminal window:
```
1. For me chrome v90 doesn't work, so, I ran the following cmd: webdriver-manager start --versions.chrome 86.0.4240.22
2. npm test -- --params.baseUrl https://localhost:9090 --capabilities.shardTestFiles true
```
```localhost port may differ for every user```

All division integration test cases should pass.

- [x] This PR includes tests

## Additional Information

There is no logic change to require an update in documentation

### DIFF
--- a/traffic_portal/test/integration/Data/divisions.ts
+++ b/traffic_portal/test/integration/Data/divisions.ts
@@ -63,6 +63,12 @@ export const divisions = {
 					password: "pa$$word"
 				}
 			],
+			check: [
+				{
+					description: "check CSV link from Division page",
+					Name: "Export as CSV"
+				}
+			],
 			add: [
 				{
 					description: "create a Divisions",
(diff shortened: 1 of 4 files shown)


dpham692:
Code looks good to me. Same like Cachegroup check CSV test.

dpham692 (approved):
Code looks good!

PR #5941, opened 2021-06-15
Show delivery service ID in TR log output

ARMmaster17:
These changes add a `svc=...` tag to request logging for both DNS and HTTP requests. The value of this field is the `XmlId` of the delivery service that routed the request.

- Traffic Router

Pull code and build CiaB. Make a request against a DNS or HTTP delivery service and observe that the log output contains lines that contain one of the following:

- `svc="mydeliveryservice1"`
- `svc="-"`

The second output should only be there if a request was made where no delivery service was used (e.g. an error occurred).

- [x] This PR includes tests


traffic_router/core/src/main/java/com/comcast/cdn/traffic_control/traffic_router/core/router/HTTPRouteResult.java
```diff
@@ -60,6 +61,10 @@ public void setUrl(final URL url) {
 		return deliveryServices;
 	}
 
+	public String getDeliveryServicesLogString() {
+		return this.getDeliveryServices().stream().map(DeliveryService::getId).collect(Collectors.joining("|"));
```
rawlinp, line 65: Is this due to `CLIENT_STEERING` delivery services? We want to log the target DS XML IDs joined with `|` instead of logging the XML ID of the `CLIENT_STEERING` DS?

ARMmaster17: Just tested with a `CLIENT_STEERING` DS and as of right now it only logs the target. L65 was originally added only because internally the `xml_id`s of an HTTP request are stored internally as a `List<>` object. Would you prefer only having the `CLIENT_STEERING` DS, or a combination of the two?

ARMmaster17:
I apologize for the delay. After some discussion it was determined that displaying all of the delivery services, delimited with a `|`, was the best way forward for our use case. This has been implemented in the latest commit and has been verified and tested locally.

smalenfant:
I'm ok with merging this. Running in production on 5.1.2.

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/http/RouterFilter.java
```diff
@@ -112,11 +113,17 @@ private void writeHttpResponse(final HttpServletResponse response, final HttpSer
...
+			if (routeResult != null && routeResult.getDeliveryServices().size() > 0) {
+				deliveryServiceIds = routeResult.getDeliveryServicesLogString();
+			}
+
 			final HTTPAccessRecord access = httpAccessRecordBuilder.resultType(track.getResult())
 				.resultDetails(track.getResultDetails())
 				.resultLocation(track.getResultLocation())
 				.requestHeaders(accessRequestHeaders)
 				.regionalGeoResult(track.getRegionalGeoResult())
+					.deliveryServiceIds(deliveryServiceIds)
```
rawlinp, line 126: nit: too much indentation here

rawlinp (approved):
Looks good, just one minor nit about indentation that I won't hold up this PR for 👍

PR #5955, opened 2021-06-17
Fix failing CDNs integration tests in APIv4

### LINKED ISSUE #5951: TO API test TestCDNs fails: Expected only one cdn response
## I'm submitting a ...

-  bug report

-  Traffic Ops

## Current behavior:

TestCDNs fails for API v4:

```go
INFO: traffic_ops_test.go:71: 2021-06-17T00:00:33.129707062Z: Using Config values:
                           TO Config File:       ../conf/traffic-ops-test.conf
                           TO Fixtures:          tc-fixtures.json
                           TO URL:               https://localhost:6443
                           TO Session Timeout In Secs:  60
                           DB Server:            localhost
                           DB User:              traffic_ops
                           DB Name:              traffic_ops
                           DB Ssl:               false
               UseIMS:               true
=== RUN   TestCDNs
    cdns_test.go:386: Expected only one cdn response {[] {[]}}
--- FAIL: TestCDNs (1.62s)
FAIL
FAIL    github.com/apache/trafficcontrol/traffic_ops/testing/api/v4     2.130s
FAIL
```

## Expected behavior:

TestCDNs should pass.

```shell
go test ./v4 -cfg=../conf/traffic-ops-test.conf -test.v -test.run '^TestCDNs$'
```

ocket8888:
- [x] This PR fixes #5951 

This PR fixes the TO/Go client integration tests. There were two failures - one from checking the incorrect response code, and one that occurred because TO data was being altered by an earlier test that then never restored original state, and the test assumed a specific state of the data in TO. The fix I chose was to cause the first test to restore state before exiting.

- Traffic Ops Client (Go) (tests only)

Make sure the TO API/Go client integration tests pass.

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v4/cdns_test.go
+++ b/traffic_ops/testing/api/v4/cdns_test.go
@@ -48,7 +48,7 @@ func TestCDNs(t *testing.T) {
 		header.Set(rfc.IfMatch, etag)
 		UpdateTestCDNsWithHeaders(t, header)
 		GetTestCDNs(t)
-		GetTestCDNsbyDomainName(t)
+		t.Run("get CDNs filtered by Domain Name", GetTestCDNsbyDomainName)
 		GetTestCDNsbyDnssec(t)
 		GetTestCDNsIMSAfterChange(t, header)
 		CreateTestCDNEmptyName(t)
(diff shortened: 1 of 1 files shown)


zrhoffman (approved):
TestCDNs passes :+1:
TestDeliveryServices fails, which is outside the scope of this PR.

PR #5956, opened 2021-06-17
t3c-apply/replaceCfgFile fix logic for reloads/restarts

traeak:

- [ ] This PR is not related to any Issue 

- Traffic Ops ORT

The very last call to replaceCfgFile ends up determining what services should be restarted and how ATS configs need to be reloaded.

In this case the order of config file evaluation determines whether the bug manifests itself or not.

- master


rob05c (changes requested):
LGTM. I tested, and it seems to work as expected. Hard to be certain with a race condition, but I changed multiple files including a header rewrite and _not_ remap.config, and saw the message when remap.config is touched.

But, I don't see any tests. Would you mind writing an Integration Test for this? I think you could make the race deterministic, if you made sure the config files list is sorted before t3c-apply iterates over it, and then made the test modify a file that comes alphabetically after `hdr_rw`.

PR #5957, opened 2021-06-17
Fix integration test

ocket8888:

This PR fixes integration tests known to be broken on master. Per [mailing list discussion](https://lists.apache.org/thread.html/r9e216e1f6ff378801c6a4f55e3cea554804171397076e84665d6f457%40%3Cdev.trafficcontrol.apache.org%3E) this removes the TO Go client/API integration test for default `/deliveryservice` response sort order and adds a unit test to the `github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice` package that verifies that an `ORDER BY xml_id` clause is generated for the relevant query when no query string parameters were provided in the client request.

This also upgrades the postgresql service used by the integration tests' GHA to version 13, to be the same as that used by CDN-in-a-Box, and adds a deprecation notice to [`github.com/apache/trafficcontrol/lib/go-tc.DBError`](https://pkg.go.dev/github.com/apache/trafficcontrol@v5.1.2+incompatible/lib/go-tc#pkg-constants) since it's too vague to ever be useful, and changes one place where it was previously used to instead return an error that previously was logged immediately.

### Update
I also found a problem with some `/cachegroupparameters` tests that weren't resetting query parameters that I'm going to tentatively declare fixed even before the actions are done running.

- Traffic Ops Client (Go) (tests only)
- Traffic Ops
- CI tests

Make sure all of the tests pass - for real this time.

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v4/deliveryservices_test.go
+++ b/traffic_ops/testing/api/v4/deliveryservices_test.go
@@ -43,15 +43,15 @@ func TestDeliveryServices(t *testing.T) {
 		header.Set(rfc.IfModifiedSince, ti)
 		header.Set(rfc.IfUnmodifiedSince, ti)
 		if includeSystemTests {
-			SSLDeliveryServiceCDNUpdateTest(t)
-			CreateTestDeliveryServicesURLSignatureKeys(t)
-			GetTestDeliveryServicesURLSignatureKeys(t)
-			DeleteTestDeliveryServicesURLSignatureKeys(t)
-			CreateTestDeliveryServicesURISigningKeys(t)
-			GetTestDeliveryServicesURISigningKeys(t)
-			DeleteTestDeliveryServicesURISigningKeys(t)
-			DeleteCDNOldSSLKeys(t)
-			DeliveryServiceSSLKeys(t)
+			t.Run("Update CDN for a Delivery Service with SSL keys", SSLDeliveryServiceCDNUpdateTest)
+			t.Run("Create URL Signature keys for a Delivery Service", CreateTestDeliveryServicesURLSignatureKeys)
+			t.Run("Retrieve URL Signature keys for a Delivery Service", GetTestDeliveryServicesURLSignatureKeys)
+			t.Run("Delete URL Signature keys for a Delivery Service", DeleteTestDeliveryServicesURLSignatureKeys)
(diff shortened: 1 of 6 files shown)


ocket8888:
Note that CiaB is currently failing in master - not my fault I swear

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -43,15 +42,15 @@ func TestDeliveryServices(t *testing.T) {
...
-			SSLDeliveryServiceCDNUpdateTest(t)
-			CreateTestDeliveryServicesURLSignatureKeys(t)
-			GetTestDeliveryServicesURLSignatureKeys(t)
-			DeleteTestDeliveryServicesURLSignatureKeys(t)
-			CreateTestDeliveryServicesURISigningKeys(t)
-			GetTestDeliveryServicesURISigningKeys(t)
-			DeleteTestDeliveryServicesURISigningKeys(t)
-			DeleteCDNOldSSLKeys(t)
-			DeliveryServiceSSLKeys(t)
+			t.Run("Update CDN for a Delivery Service with SSL keys", SSLDeliveryServiceCDNUpdateTest)
```
mattjackson220, line 45: should we make these changes to the lines below outside of the if statement too? just for consistency?

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices_test.go
```diff
@@ -205,3 +207,177 @@ func TestMakeExampleURLs(t *testing.T) {
 		t.Errorf("MakeExampleURLs expected %v actual %v", expected, actual)
 	}
 }
+
+func TestReadGetDeliveryServices(t *testing.T) {
+	mockDB, mock, err := sqlmock.New()
+	if err != nil {
+		t.Fatalf("an error '%s' was not expected when opening a stub database connection", err)
```
mattjackson220, line 214: should be `%v`

mattjackson220 (changes requested):

traffic_ops/testing/api/v4/deliveryservices_test.go
ocket8888: I can, sure

mattjackson220 (approved):
Code looks good, tests pass! Ill merge it after a rebase

PR #5963, opened 2021-06-21
Fix t3c docs missing commands

rob05c:
Fixes the t3c docs missing a few newer commands.

Is docs.
No tests, is docs.
No changelog, is docs, and not in a release.

- Traffic Ops ORT

Check t3c manpage, check t3c help flag, verify all sub-commands are present.

Not  a bug fix.

### DIFF
--- a/cache-config/t3c/README.md
+++ b/cache-config/t3c/README.md
@@ -51,7 +51,7 @@ The latest version and documentation can be found at https://github.com/apache/t
 
 # OPTIONS
 
---help
+\-\-help
     Prints the synopsis and usage information.
 
 # COMMANDS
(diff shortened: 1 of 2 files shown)


traeak:
Does the cache-config/t3c/README.md file also need t3c-check-refs and t3c-check-reload as well?

rob05c:
> Does the cache-config/t3c/README.md file also need t3c-check-refs and t3c-check-reload as well?

I don't think it should. `t3c-check` has that, IMO better to avoid duplication, and just have `t3c-check` in `t3c`, and then users should do `man t3c-check` to see its options (`t3c-check-refs` and `t3c-check-reload`)

traeak (approved):
new md file looks good.  comment regarding the 2 t3c-check-* should be okay.

PR #5974, opened 2021-06-24
Every endpoint that modifies a CDN, or components related to a CDN, needs to be "lock safe"

srijeet0406:

- Traffic Ops
- CI tests

Make sure all the unit tests and api tests pass.
Manual testing:
Run TO and TP locally.
Grab a hard lock for any cdn for a particular user.
Now, login to TP as another user, and try to modify the CDN, or CDN related components. These operations should be forbidden.
Log out and log back in to TP as the first user and try the same operations. This time they should all work correctly.

- master

- [x] This PR includes tests 


traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
```diff
@@ -611,12 +611,23 @@ func (cg *TOCacheGroup) Update(h http.Header) (error, error, int) {
 		return userErr, sysErr, errCode
 	}
 
+	cdns, err := dbhelpers.GetCDNsForCachegroup(cg.ReqInfo.Tx.Tx, cg.ID)
+	if err != nil {
+		return nil, err, http.StatusInternalServerError
+	}
+	for cdn, _ := range cdns {
+		userErr, sysErr, statusCode := dbhelpers.CheckIfCurrentUserCanModifyCDN(cg.ReqInfo.Tx.Tx, cdn, cg.ReqInfo.User.UserName)
```
rawlinp, line 619: Just perusing here, but we generally shouldn't make DB calls in a loop like this. It's better to make a _single_ DB query that gets all the stuff we need to check in memory and check everything in memory at once, as opposed to making a DB query per CDN or per delivery service. Especially for things as numerous as delivery services, that can really slow things down and increase load on the DB.

srijeet0406: Should be fixed now.

traffic_ops/traffic_ops_golang/servercheck/servercheck.go
```diff
@@ -131,6 +131,16 @@ func CreateUpdateServercheck(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	cdnName, err := dbhelpers.GetCDNNameFromServerID(inf.Tx.Tx, int64(id))
```
rawlinp, line 134: `servercheck` probably doesn't need to be lock-safe since it's primarily just for monitoring and doesn't really affect CDN operation

traffic_ops/traffic_ops_golang/topology/topologies.go
```diff
@@ -806,3 +821,22 @@ func selectMaxLastUpdatedQuery(where string) string {
...
+	serverIDs, err := dbhelpers.GetServerIDsFromCachegroupNames(topology.ReqInfo.Tx.Tx, cachegroups)
+	if err != nil {
+		return nil, err, http.StatusInternalServerError
+	}
+	cdns, err := dbhelpers.GetCDNNamesFromServerIds(topology.ReqInfo.Tx.Tx, serverIDs)
+	if err != nil {
+		return nil, err, http.StatusInternalServerError
+	}
+	for _, cdn := range cdns {
+		userErr, sysErr, statusCode := dbhelpers.CheckIfCurrentUserCanModifyCDN(topology.ReqInfo.Tx.Tx, cdn, topology.ReqInfo.User.UserName)
```
rawlinp, line 836: ditto earlier comments about making DB queries in a loop

rawlinp (changes requested):

ocket8888:
Do you have a list anywhere of the endpoints this affects? The blueprint didn't include one because it was just a bit too pedantic for a blueprint, but I think it'd be nice to know which endpoints are "lock-able" - would also maybe help reviewers verify that everything that _should_ be locked _is_.

srijeet0406:
> Do you have a list anywhere of the endpoints this affects? The blueprint didn't include one iirc because it was just a bit too pedantic for a blueprint, but I think it'd be nice to know which endpoints are "lock-able" - would also maybe help reviewers verify that everything that _should_ be locked _is_.

Yeah, you're right. I didn't include a list of endpoints in the blueprint, but I'll list them out here:

POST/ PUT/ DELETE endpoints for the following:
CDNs (not Post, because the CDN needs to be present for a user to be able to lock it)
Servers
Cachegroups (while assigning and unassigning servers)
Cachegroup parameters
Topologies (while adding or deleting cachegroups)
DeliveryServices
Delivery Service related cert or key renewal/ creation/ deletion
Delivery Service Servers
Delivery Service Regexes
Federations
Origins
Profiles
Profile Parameters

zrhoffman (changes requested):
> Yeah, you're right. I didn't include a list of endpoints in the blueprint, but I'll list them out here

Since several routes need to perform the same check (even though the CDN is found in different ways, depending on the endpoint), `CheckIfCurrentUserCanModifyCDNs()` would be a good candidate for a route middleware. The `Route` struct lets you pass in an array of middlewares:
https://github.com/apache/trafficcontrol/blob/bd0b5dca7e1b57494edb19c4caec44a51f466401/traffic_ops/traffic_ops_golang/routing/routing.go#L58

The default middlewares would also need to be included, since `middlewares` would not be nil:
https://github.com/apache/trafficcontrol/blob/bd0b5dca7e1b57494edb19c4caec44a51f466401/traffic_ops/traffic_ops_golang/routing/routing.go#L195-L197

But doing this would make it clear, just by looking at the routes in `routes.go`, which routes use CDN locks, and would hopefully reduce code duplication.

ocket8888:
Yeah, I was gonna suggest maybe adding a field to a Route structure, but it's much the same as a middleware. Would be many more lines of change than a middleware, too.

rawlinp:
So, I agree that we should try to reduce duplication as much as possible, but I'm not sure that changing the entire direction of this PR to use middleware is really appropriate for this. When I think of middleware, I think of things that generally apply to all routes (authorization, setting response headers, gzipping responses, etc.). These locking checks are somewhat specialized per route and depend on what is being modified. Somehow the middleware would have to know which CDNs were being modified, then check those against the current CDN locks. That means the route would have to set the CDN list in the request context (probably), and the middleware would have to get those out of the request context (after the request has been "handled successfully" as far as the handler knows) to run the check (which might fail). So where we're getting the CDNs and where we're checking the CDNs would be two totally different areas of the code (which isn't great IMO).

Maybe there is another way to do it better with middleware, but that is just the idea that came to mind. It just doesn't seem like a good fit. On the plus side, CDN locking should be unnecessary once we complete the plan for DS and server snapshots, so we should be able to deprecate and remove CDN locks once we get to that point. But in the meantime, they will tide us over.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -242,52 +289,59 @@ func CheckIfCurrentUserCanModifyCDNWithID(tx *sql.Tx, cdnID int64, user string)
...
 	defer rows.Close()
 	for rows.Next() {
-		err = rows.Scan(&userName, &soft)
+		err = rows.Scan(&userName, &cdn, &soft)
 		if err != nil {
-			return nil, errors.New("scanning cdn_lock for user " + user + " and cdn " + string(cdnName) + ": " + err.Error()), http.StatusInternalServerError
+			return nil, errors.New("scanning cdn_lock for user " + user + " and cachegroup ID " + strconv.Itoa(cachegroupID) + ": " + err.Error()), http.StatusInternalServerError
 		}
 	}
 	if userName != "" && user != userName && !soft {
```
rawlinp, line 316: Is this check supposed to be in the loop above?

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -242,52 +289,59 @@ func CheckIfCurrentUserCanModifyCDNWithID(tx *sql.Tx, cdnID int64, user string)
...
 	return nil, nil, http.StatusOK
 }
 
-// CheckIfCurrentUserCanModifyCachegroup checks if the current user has the lock on the cdns that are associated with the provided cachegroup ID.
-// This will succeed if no other user has a hard lock on any of the CDNs that relate to the cachegroup in question.
-func CheckIfCurrentUserCanModifyCachegroup(tx *sql.Tx, cachegroupID int, user string) (error, error, int) {
-	query := `SELECT username, cdn, soft FROM cdn_lock WHERE cdn IN (SELECT name FROM cdn WHERE id IN (SELECT cdn_id FROM server WHERE cachegroup = ($1)))`
+// CheckIfCurrentUserCanModifyCachegroups checks if the current user has the lock on the cdns that are associated with the provided cachegroup IDs.
+// This will succeed if no other user has a hard lock on any of the CDNs that relate to the cachegroups in question.
+func CheckIfCurrentUserCanModifyCachegroups(tx *sql.Tx, cachegroupID []int, user string) (error, error, int) {
```
rawlinp, line 324: nit: `cachegroupID` -> `cachegroupIDs` since it's a slice

rawlinp (changes requested):

zrhoffman:
> So where we're getting the CDNs and where we're checking the CDNs would be two totally different areas of the code (which isn't great IMO).

Since all "lock-safe" endpoints are either POST, PUT, or DELETE, the parameter identifying the resource will be almost always be unique to that resource. There could be an additional table like `locks_by_route` with 3 columns:
* `resource_key`
* `route`
* `cdn_is_locked`

and a CDN Locks middleware would be able to check if a given Route is already locked with respect to a given resource key without needing to derive which CDN that resource uses.

This almost always means needing to know in advance which parameters to use for `resource_key` (as far as I know, this would be primary key in all cases except endpoints implementing `OptionsDeleter.DeleteKeyOptions()`), but there could be an additional table indicating which DB tables correspond to which TO Routes.

In the case of resources implementing `OptionsDeleter.DeleteKeyOptions()` (Topologies and Regions only), there would need to be a row for each option returned by ``DeleteKeyOptions()` for that resource type.

Since this approach would involve setting up infrastructure in Traffic Ops that we don't currently have, it wouldn't necessarily immediately save implementation time, though maybe it would be more maintainable long-term than calling `CheckIfCurrentUserCanModifyCDN()` from each affected handler.

> The middleware would have to get those out of the request context (after the request has been "handled successfully" as far as the handler knows) to run the check (which might fail).

If there is a CDN Locks middleware on some Routes, it would go at the end of the middlewares list (after `middleware.WrapPanicRecover()`. Since it is at the  end of the middleware list, and since we'd be passing a custom middlewares list for each such route, calling `h(w, r)` from the CDN Locks middleware would always call the Route handler, meaning that if the CDN Locks check fails, calling `h(w, r)` could just be skipped.

This PR aside, middlewares are very underused in Traffic Ops (we actually have 0 endpoints that use non-default middlewares, `Route.Middlewares[]` is nil or empty for all Routes listed in `routes.go`), so hopefully someone does make use of them at some point.

rawlinp:
> maybe it would be more maintainable long-term than calling CheckIfCurrentUserCanModifyCDN() from each affected handler

I will refer back to my previous comment:
> On the plus side, CDN locking should be unnecessary once we complete the plan for DS and server snapshots, so we should be able to deprecate and remove CDN locks once we get to that point. But in the meantime, they will tide us over.

Long-term, I see CDN locking going away in favor of DS and server snapshots. Those are actually on our roadmap for Q4, so I'd be very hesitant to make this a perfect long-term solution when we may not need it for long anyways.

srijeet0406:
I agree that code duplication should be minimized, but as Rawlin specified, this was never meant to be the long term solution, so holding up this last PR(and the project and the 6.0 release) for a major refactor isn't super beneficial IMO. I'm also trying to keep the CDN locks simple without adding a bunch more database tables and such, in accordance with the blueprint. Additionally, while speaking with Dave and the rest of the team, this was supposed to be the "first pass" at CDN locks. We can keep improving it in the next iterations (until it is replaced) and I'd be happy to add a ticket to take care of this in a future HIP sprint. Does that sound good to everyone?

ocket8888:
I don't ultimately care how it gets done, just a suggestion. Makes sense that you don't want to overhaul the entire PR at this point.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
srijeet0406: Yes, because if any of the scans for the locks fails, we'd want to exit.

zrhoffman:
> I'd be very hesitant to make this a perfect long-term solution when we may not need it for long anyways.

> [T]his was never meant to be the long term solution

> Makes sense that you don't want to overhaul the entire PR at this point.

Good points, agreed that this PR is not the place to add a middleware.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rawlinp: Ok, this was marked resolved, but this check wasn't moved into the loop above

srijeet0406: Whoops, I thought you were referring to the previous error check. The other check should be inside the loop. Fixed now.

rawlinp (approved):
This looks good to me, but there are some merge conflicts now. Also, I re-ran the GH checks since there appeared to be an unrelated TO API test failure, so I'd like to make sure it's really unrelated.

zrhoffman (approved):

rawlinp:
Seems like this TO API v4 test might have spurious failures:
```
--- FAIL: TestCacheGroupParameters (1.99s)
```
However, the test output doesn't seem to indicate why it actually failed. FWIW I'm removing that v4 test in https://github.com/apache/trafficcontrol/pull/5993, so I'm not going to worry about it.

PR #5980, opened 2021-06-28
Handle pageination in Riak search

### LINKED ISSUE #5902: TO riak client is unable to get more than 1000 records
## I'm submitting a ...

-  bug report

-  Traffic Ops
-  Traffic Vault

## Current behavior:
When using the [search](https://github.com/apache/trafficcontrol/blob/60a9ffdcabdeb011773a900739ec2615da47ab33/traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/riak_services.go#L397) feature of Riak, the client has to pass a maximum number of rows that can be turned in a single query. This is [hardcoded to 1000 entries](https://github.com/apache/trafficcontrol/blob/60a9ffdcabdeb011773a900739ec2615da47ab33/traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/dsutil.go#L342).

The endpoints using this search query are
`GET cdns/name/{name}/sslkeys` -> Gets all sslkeys for DS's on a CDN
`PUT snapshot/` -> Deletes old sslkeys on a DS for a CDN

In effect, this means that if you have a CDN with over 1000 DS's with sslkeys, the client will be unable to get all of the sslkeys.
In terms of deleting, if _any_ change is made to a CDN with over 1000 keys, then only the first 1000 sslkey records for that CDN will be correctly cleaned up.

## Expected behavior:
CDN works as expected regardless of the number of sslkeys. Specifically `GET cdns/name/{name}/sslkeys` should be able to return more than 1000 records and `DELETE snapshot/` should cleanup all sslkeys, regardless of total count.

Create a CDN with over a thousand DS' that have sslkeys.

shamrickus:

- [x] This PR fixes #5902 

I have written the tests for this change, but given that it takes ~8 minutes I've decided to omit it.

- Traffic Ops
- Traffic Vault

Create >1000 DS with sslkeys on a single cdn, call `cdns/name/{name}/sslkeys` and verify it returns more than 1000 records.
Create >1000 SSLKeys on a single CDN (doesn't matter if it's one or many DS'). Then make a change (such as moving a DS to another CDN) that modifies more than 1000 ssl certs the cdn. Finally, perform a snapshot. Only 1000 of the SSLKey records will have been removed/updated. 
The order in which these are returned is transient as far as I can tell, so which 1000 are changed differs each run.

- every version

### DIFF
--- a/traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/riak_services.go
+++ b/traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/riak_services.go
@@ -401,7 +401,7 @@ func search(cluster StorageCluster, index string, query string, filterQuery stri
 			WithIndexName(index).
 			WithQuery(query).
 			WithNumRows(numRows).
-			WithStart(numRows)
+			WithStart(start)
 		if len(filterQuery) > 0 {
 			riakCmd = riakCmd.WithFilterQuery(filterQuery)
 		}
--- a/traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/riak_services.go
+++ b/traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/riak_services.go
@@ -396,7 +396,7 @@ func withCluster(tx *sql.Tx, authOpts *riak.AuthOptions, riakPort *uint, f func(
 // If fields is empty, all fields will be returned.
 func search(cluster StorageCluster, index string, query string, filterQuery string, numRows uint32, fields []string) ([]*riak.SearchDoc, error) {
 	var searchDocs []*riak.SearchDoc
-	for start := uint32(0); ; {
+	for start := uint32(0); ; start += numRows {
 		riakCmd := riak.NewSearchCommandBuilder().
 			WithIndexName(index).
 			WithQuery(query).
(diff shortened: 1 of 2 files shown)


traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/riak_services.go
```diff
@@ -394,31 +394,54 @@ func withCluster(tx *sql.Tx, authOpts *riak.AuthOptions, riakPort *uint, f func(
...
+		}
+		if err = cluster.Execute(iCmd); err != nil {
+			return nil, errors.New("executing Riak command index '" + index + "' query '" + query + "': " + err.Error())
+		}
+		cmd, ok := iCmd.(*riak.SearchCommand)
+		if !ok {
+			return nil, fmt.Errorf("riak command unexpected type %T", iCmd)
+		}
+		if cmd.Response == nil {
+			return nil, fmt.Errorf("riak recieved nil responce")
```
mattjackson220, line 424: spelling x2 ("received" and "response" )

mattjackson220 (changes requested):

traffic_ops/traffic_ops_golang/trafficvault/backends/riaksvc/riak_services.go
shamrickus: What, you've never seen it spelled "responce"!? :laughing:

mattjackson220 (approved):
Looks good to me! code looks good, tested locally in CiaB and works as expected.

ocket8888:
TO Go client/API integration tests will pass if you rebase

PR #5983, opened 2021-06-29
Make Traffic Stats fail over to other Traffic Monitors

### LINKED ISSUE #5942: Traffic Stats does not failover to another TM when one stops responding
## I'm submitting a ...
-  bug report

-  Traffic Stats

## Current behavior:
Traffic Stats does not gracefully handle the situation when a Traffic Monitor it is using stops responding.  What happens currently is if the Traffic Monitor that TS is getting data from goes away then TS just will stop collecting stats and will not write out any stats/data/metrics to an output platform (e.g. InfluxDB).

## Expected behavior:
Traffic Stats ideally should have a list of Online TMs that it can get data from and if one does not respond simply move on and use another one and continue to collect stats and output stats to whatever the desired output platform is.

1) Have TS running and have two Online TMs
2) Stop TM1 and if TS does not log an error then start that one up and stop TM2.
3) Once you stop the TM that TS is connecting to you will cease to see messages like this in the traffic_stats.log:
  [INFO] 2021-06-15 21:03:40 Collected XXX deliveryservice stats values for cdn @ 1623791020
  [INFO] 2021-06-15 21:03:40 Collected XXXX cache stats values for cdn @ 1623791020
4) You won't see those messages because TS is not collecting stats because the TM it was using is not responding and it does not try the other TM, it appears to be stuck trying to use the one that is no longer responding.

Versions in the test:
TS: 3.0
TM: 5.2

rawlinp:
- [x] This PR fixes #5942

This PR also improves much of the error logging in Traffic Stats to make it easier to troubleshoot and contains some minor cleanup of unused parameters/functions.

- Traffic Stats

1. Verify the included unit test passes.
2. In an environment with 2 or more TMs for TS to poll, shut down each of the TMs one at a time, view the logs, and verify that TS retries the available TM. Note: depending on which TM that TS tries first, shutting down the _other_ TM won't have any effect on TS. You can view the logs to see which TM that TS is querying.

4.x, 5.x, master

- [x] This PR includes tests
- [x] bugfix, no docs needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - [#5407](https://github.com/apache/trafficcontrol/issues/5407) - Make sure that you cannot add two servers with identical content
 - [#2881](https://github.com/apache/trafficcontrol/issues/2881) - Some API endpoints have incorrect Content-Types
 - [#5863](https://github.com/apache/trafficcontrol/issues/5863) - Traffic Monitor logs warnings to `log_location_info` instead of `log_location_warning`
+- [#5492](https://github.com/apache/trafficcontrol/issues/5942) - Traffic Stats does not failover to another Traffic Monitor when one stops responding
 - [#5363](https://github.com/apache/trafficcontrol/issues/5363) - Postgresql version changeable by env variable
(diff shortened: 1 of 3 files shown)


srijeet0406 (approved):
All unit/ api/ manual tests pass.
Code looks good. 
Thanks for the cleanup :)

PR #5984, opened 2021-06-29
Traffic Ops: log a warning instead of ignoring write errors

### LINKED ISSUE #5954: Traffic Ops HTTP response write errors are ignored
## I'm submitting a ...
-  bug report

-  Traffic Ops

## Current behavior:
Most of the functions that write HTTP responses in TO ignore write errors, [like this one for example](https://github.com/apache/trafficcontrol/blob/49349857d78a340269d6db2a7e4a3ef5bf1f2d6b/traffic_ops/traffic_ops_golang/api/api.go#L137). This means that if there are any errors writing HTTP responses (due to e.g. hitting timeouts), we are mostly blind to them currently.

## Expected behavior:
TO should log warnings if any errors occur when writing HTTP responses.

Since the current behavior is invisible, you're just going to have to look at the code and trust me on this.

rawlinp:
- [x] This PR fixes #5954

- Traffic Ops

Ensure the TO API tests still pass. Since this logging only happens on write errors, one way to test the logging would be to set the `traffic_ops_golang.request_timeout` in cdn.conf to 1 and make a request to a certain endpoint which is known to take longer than 1 second to return a response. Alternatively, you could add a `time.sleep(5*time.Second)` into some handler and call that instead. You should see a log like the following:
```
WARNING: api.go:120: 2021-06-29T21:44:57.123243Z: failed to write response (method = GET, URL = /api/3.1/ping, request ID = 1, remote addr = [::1]:62136, bytes written = 0): http: Handler timeout
```

4.x, 5.x, master

- [x] minor logging fixes, tests didn't seem necessary
- [x] logging fixes, no docs needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - [#5690](https://github.com/apache/trafficcontrol/issues/5690) - Fixed github action for added/modified db migration file.
 - [#2471](https://github.com/apache/trafficcontrol/issues/2471) - A PR check to ensure added db migration file is the latest.
 - [#5609](https://github.com/apache/trafficcontrol/issues/5609) - Fixed GET /servercheck filter for an extra query param.
+- [#5954](https://github.com/apache/trafficcontrol/issues/5954) - Traffic Ops HTTP response write errors are ignored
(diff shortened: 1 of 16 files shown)


srijeet0406 (approved):
Changes look good.
Unit + API tests pass.
Manual testing resulted in the log line being generated.
LGTM!

PR #5985, opened 2021-06-29
Add t3c If-Modified-Since Support

rob05c:
Adds t3c using If-Modified-Since requests to avoid making unnecessary requests to Traffic Ops.

Includes a feature flag `t3c-apply --no-cache` to disable.

Includes tests.
Includes docs.
Includes changelog.

- Traffic Ops ORT

Run tests. Run t3c twice, verify requests are not made for most endpoints (Traffic Ops doesn't support IMS on Riak "Vault" endpoints, and some are currently broken, but most should work).

Not a bug fix.


traffic_ops/toclientlib/toclientlib.go
```diff
@@ -808,6 +809,7 @@ type ReqInf struct {
 	CacheHitStatus CacheHitStatus
 	RemoteAddr     net.Addr
 	StatusCode     int
+	Headers        http.Header
```
rawlinp, line 812: Should these be called `ResponseHeaders` as opposed to just `Headers`? I think these could be confused as request headers, since the type name is short for "request info".

rob05c: Changed

alficles (approved):
Sorry for the delay. This is good to go, I think. Caching these requests will help considerably and this handles the caching correctly.

PR #5989, opened 2021-07-01
Removing LD2 and LD3 fields from TP

srijeet0406:

- Traffic Portal

Run TP and make sure the long desc 2 and long desc 3 fields don't show up.
Also make sure that the long desc 1 field now appears as `long desc`

- master

- [x] Removing non existent fields, tests didnt seem necessary


rimashah25 (approved):
LGTM
The fields `Long Description 1` and `Long Description 2` are no longer seen in Create DS page for HTTP, any_map, DNS or steering. I also don't see them as an option to select in the DS table either. Integration tests pass on local.

rawlinp (approved):
Looks good to me

PR #5998, opened 2021-07-06
Remove unused deliveryservice.multi_site_origin_algorithm column

### LINKED ISSUE #1546: deliveryservice.multi_site_origin_algorithm column isn't accessible via any UI
The question is is the deliveryservice.multi_site_origin_algorithm  even used anymore or was it abandoned when delivery service profiles were introduced? 

If it is fact used, it needs to be added to the TO UI and the TP UI. and i'm guessing it only pertains to DNS* and HTTP* delivery services.

If it is not used, the column should be deleted from the database in my opinion.

rawlinp:
- [x] This PR fixes #1546

- Traffic Ops

Run `db/admin upgrade`, `db/admin down`, then `db/admin upgrade` again to verify that the migration is successful.

- [x] removes unused DB column, no tests necessary
- [x] removes unused DB column, no docs necessary
- [x] not important enough to be in the changelog

### DIFF
--- a/traffic_ops/app/db/migrations/2021070600000000_remove_ds_mso_alg.sql
+++ b/traffic_ops/app/db/migrations/2021070600000000_remove_ds_mso_alg.sql
@@ -1,0 +1,17 @@
+/*
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


zrhoffman (approved):
Confirmed that up/down work as expected, and that the column is unused.

PR #6002, opened 2021-07-07
Remove test that isn't working properly

### LINKED ISSUE #5793: Test is checking for field that doesn't truly exist
## I'm submitting a ...
-  bug report

-  Traffic Ops Client (Go) (tests)

## Current behavior:
The tests for the `/profileparameter` endpoint methods of the client are checking that responses are by default sorted by Parameter Name. Which is impossible, because the endpoint doesn't return Parameter names - it succeeds only because a list of empty strings is trivially sorted both ascending and descending.

The specific test content is:
```go
func SortTestProfileParameters(t *testing.T) {
	var header http.Header
	var sortedList []string
	resp, _, err := TOSession.GetProfileParameters(nil, header)
	if err != nil {
		t.Fatalf("Expected no error, but got %v", err.Error())
	}
	for i := range resp {
		sortedList = append(sortedList, resp[i].Parameter)
	}

	res := sort.SliceIsSorted(sortedList, func(p, q int) bool {
		return sortedList[p] < sortedList[q]
	})
	if res != true {
		t.Errorf("list is not sorted by their names: %v", sortedList)
	}
}
```

Now, the endpoint _does_ return objects with a `parameter` property, but that is actually the Parameter's ID, not its name. The objects returned from the client have their ProfileName and ParameterID set (but not LastUpdated) by a coercion process in the client itself that unmarshals the response as an array of `ProfileParameterNullable` and then broadens the data to `ProfileParameter`, which contains many extraneous fields.

## Expected behavior:
Tests should only test expected behavior, and client methods should use appropriate data types in their call signatures.

Add a `t.Logf` to that test to print out the list, then run it and observe that the list contains only empty strings.

## Anything else:
I'm mostly just opening this issue to justify deleting the test from the v4 tests in #5747 and so I don't forget that justification come review time. But v3 and earlier presumably also suffer from it.

ocket8888:
- [x] This PR fixes #5793

This removes a `/profileparameters` Go client/API integration test that doesn't actually test anything. For details, refer to #5793.

- Traffic Control Client (Go) (API integration tests only)

Make sure the Go client/API integration tests pass (v3 is the version of interest, but they should *all* pass once #5957 is merged)

- master

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/testing/api/v3/profile_parameters_test.go
+++ b/traffic_ops/testing/api/v3/profile_parameters_test.go
@@ -18,7 +18,6 @@ package v3
 import (
 	"fmt"
 	"net/http"
-	"sort"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


shamrickus (approved):
LGTM

PR #6003, opened 2021-07-08
Fix Internal server error in Get Profiles by invalid Params

### LINKED ISSUE #5510: Unverified query parameter causes Internal Server Error
## I'm submitting a ...
-  bug report

-  Traffic Ops
-  Traffic Ops Client (Go)

## Current behavior:
The documentation mentions a query parameter, `param`, which is the ID of a Parameter, and it fetches all Profiles to which the thereby identified Parameter is assigned. This Parameter's type is unverified by the API (no `Checker` specified in the `WhereColumnInfo`), and if you use it with an unsupported value:

```http
GET /api/2.0/profiles?param=test HTTP/1.1
User-Agent: python-requests/2.24.0
Accept: */*
Connection: keep-alive
Cookie: mojolicious=...
```
it causes an Internal Server Error:
```http
Content-Encoding: gzip
Set-Cookie: mojolicious=...
Vary: Accept-Encoding/zUm3oa0elunc5VHcism6Qg==
Date: Wed, 10 Feb 2021 18:56:29 GMT
Content-Length: 83

{
        "alerts": [
                {
                        "level": "error"
                }
        ]
}
```

What's strange is that the client also has methods that utilize this Parameter and request input as a _string_, not an integral type, and there are tests that are passing that use those methods.

The `Profile` structure in `lib/go-tc` has a `Parameter` field and the tests appear to be using this field to provide the input for the query parameter. I can only assume that the tests are passing merely by virtue of the request returning a `200 OK` response, and that `Profile.Parameter` property is always an empty string because it has no purpose or meaning and is never populated.

## Expected behavior:
The client should test that query parameters behave in their intended way, data structures should not have fields that serve no purpose (requires `ProfileV4`?), and passing a bad query parameter should result not in a `500 Internal Server Error` response, but a `400 Bad Request` response with an Alert that explains why the request failed.

Make the request as shown above.

dmohan001c:

- [x] This PR fixes #5510  

This updates the Error message in GET profiles?param=test and profiles?cdn=test

- CDN in a Box
- Traffic Control Client 
- Traffic Ops
- CI tests

* Master

### DIFF
--- a/traffic_ops/testing/api/v2/profiles_test.go
+++ b/traffic_ops/testing/api/v2/profiles_test.go
@@ -213,7 +213,7 @@ func GetTestProfiles(t *testing.T) {
 		if len(pr.Parameters) > 0 {
 			parameter := pr.Parameters[0]
 			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
-			if err != nil{
+			if err != nil {
 				t.Errorf("Cannot GET Parameter by name: %v - %v", err, resp)
 			}
 			if len(respParameter) > 0 {
--- a/traffic_ops/testing/api/v2/profiles_test.go
+++ b/traffic_ops/testing/api/v2/profiles_test.go
@@ -212,7 +212,10 @@ func GetTestProfiles(t *testing.T) {
 		profileID := resp[0].ID
 		if len(pr.Parameters) > 0 {
 			parameter := pr.Parameters[0]
-			respParameter, _, _ := TOSession.GetParameterByName(*parameter.Name)
+			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
+			if err != nil{
+				t.Errorf("Cannot GET Parameter by name: %v - %v", err, resp)
+			}
 			if len(respParameter) > 0 {
 				parameterID := respParameter[0].ID
 				if parameterID > 0 {
--- a/traffic_ops/testing/api/v2/profiles_test.go
+++ b/traffic_ops/testing/api/v2/profiles_test.go
@@ -214,14 +214,14 @@ func GetTestProfiles(t *testing.T) {
 			parameter := pr.Parameters[0]
 			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
 			if err != nil {
-				t.Errorf("Cannot GET Parameter by name: %v - %v", err, resp)
+				t.Errorf("Cannot GET Parameter by name: %v", err)
 			}
 			if len(respParameter) > 0 {
 				parameterID := respParameter[0].ID
 				if parameterID > 0 {
 					resp, _, err = TOSession.GetProfileByParameterId(parameterID)
 					if err != nil {
-						t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+						t.Errorf("cannot GET Profile by param: %v", err)
 					}
 					if len(resp) < 1 {
 						t.Errorf("Expected atleast one response for Get Profile by Parameters, but found %d", len(resp))
--- a/traffic_ops/testing/api/v2/profiles_test.go
+++ b/traffic_ops/testing/api/v2/profiles_test.go
@@ -16,6 +16,7 @@
 package v2
 
 import (
+	"strconv"
 	"strings"
 	"testing"
 
@@ -210,12 +211,26 @@ func GetTestProfiles(t *testing.T) {
 			t.Errorf("cannot GET Profile by name: %v - %v", err, resp)
 		}
 		profileID := resp[0].ID
-
-		resp, _, err = TOSession.GetProfileByParameter(pr.Parameter)
-		if err != nil {
-			t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, _ := TOSession.GetParameterByName(*parameter.Name)
+			if len(respParameter) > 0 {
+				parameterID := respParameter[0].ID
+				if parameterID > 0 {
+					resp, _, err = TOSession.GetProfileByParameter(strconv.Itoa(parameterID))
+					if err != nil {
+						t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+					}
+					if len(resp) < 1 {
+						t.Errorf("Expected atleast one response for Get Profile by Parameters, but found %d", len(resp))
+					}
+				} else {
+					t.Errorf("Invalid parameter ID %d", parameterID)
+				}
+			} else {
+				t.Errorf("No response found for GET Parameters by name")
+			}
 		}
-
 		resp, _, err = TOSession.GetProfileByCDNID(pr.CDNID)
 		if err != nil {
 			t.Errorf("cannot GET Profile by cdn: %v - %v", err, resp)
--- a/traffic_ops/testing/api/v2/profiles_test.go
+++ b/traffic_ops/testing/api/v2/profiles_test.go
@@ -16,7 +16,6 @@
 package v2
 
 import (
-	"strconv"
 	"strings"
 	"testing"
 
@@ -217,7 +216,7 @@ func GetTestProfiles(t *testing.T) {
 			if len(respParameter) > 0 {
 				parameterID := respParameter[0].ID
 				if parameterID > 0 {
-					resp, _, err = TOSession.GetProfileByParameter(strconv.Itoa(parameterID))
+					resp, _, err = TOSession.GetProfileByParameterId(parameterID)
 					if err != nil {
 						t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
 					}
--- a/traffic_ops/testing/api/v3/profiles_test.go
+++ b/traffic_ops/testing/api/v3/profiles_test.go
@@ -86,7 +86,7 @@ func GetTestProfilesIMS(t *testing.T) {
 		if len(pr.Parameters) > 0 {
 			parameter := pr.Parameters[0]
 			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
-			if err != nil{
+			if err != nil {
 				t.Errorf("Cannot GET Parameter by name: %v - %v", err, respParameter)
 			}
 			if len(respParameter) > 0 {
@@ -291,7 +291,7 @@ func GetTestProfiles(t *testing.T) {
 		if len(pr.Parameters) > 0 {
 			parameter := pr.Parameters[0]
 			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
-			if err != nil{
+			if err != nil {
 				t.Errorf("Cannot GET Parameter by name: %v - %v", err, resp)
 			}
 			if len(respParameter) > 0 {
--- a/traffic_ops/testing/api/v3/profiles_test.go
+++ b/traffic_ops/testing/api/v3/profiles_test.go
@@ -87,7 +87,7 @@ func GetTestProfilesIMS(t *testing.T) {
 			parameter := pr.Parameters[0]
 			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
 			if err != nil {
-				t.Errorf("Cannot GET Parameter by name: %v - %v", err, respParameter)
+				t.Errorf("Cannot GET Parameter by name: %v", err)
 			}
 			if len(respParameter) > 0 {
 				parameterID := respParameter[0].ID
@@ -292,7 +292,7 @@ func GetTestProfiles(t *testing.T) {
 			parameter := pr.Parameters[0]
 			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
 			if err != nil {
-				t.Errorf("Cannot GET Parameter by name: %v - %v", err, resp)
+				t.Errorf("Cannot GET Parameter by name: %v", err)
 			}
 			if len(respParameter) > 0 {
 				parameterID := respParameter[0].ID
--- a/traffic_ops/testing/api/v3/profiles_test.go
+++ b/traffic_ops/testing/api/v3/profiles_test.go
@@ -17,6 +17,7 @@ package v3
 
 import (
 	"net/http"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -89,7 +90,7 @@ func GetTestProfilesIMS(t *testing.T) {
 			if len(respParameter) > 0 {
 				parameterID := respParameter[0].ID
 				if parameterID > 0 {
-					resp, _, err := TOSession.GetProfileByParameterWithHdr(parameterID, nil)
+					resp, _, err := TOSession.GetProfileByParameterWithHdr(strconv.Itoa(parameterID), nil)
 					if err != nil {
 						t.Fatalf("Expected no error, but got %v", err.Error())
 					}
@@ -291,7 +292,7 @@ func GetTestProfiles(t *testing.T) {
 			if len(respParameter) > 0 {
 				parameterID := respParameter[0].ID
 				if parameterID > 0 {
-					resp, _, err = TOSession.GetProfileByParameter(parameterID)
+					resp, _, err = TOSession.GetProfileByParameter(strconv.Itoa(parameterID))
 					if err != nil {
 						t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
 					}
--- a/traffic_ops/testing/api/v3/profiles_test.go
+++ b/traffic_ops/testing/api/v3/profiles_test.go
@@ -17,6 +17,7 @@ package v3
 
 import (
 	"net/http"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -83,12 +84,29 @@ func GetTestProfilesIMS(t *testing.T) {
 		if reqInf.StatusCode != http.StatusNotModified {
 			t.Fatalf("Expected 304 status code, got %v", reqInf.StatusCode)
 		}
-		_, reqInf, err = TOSession.GetProfileByParameterWithHdr(pr.Parameter, header)
-		if err != nil {
-			t.Fatalf("Expected no error, but got %v", err.Error())
-		}
-		if reqInf.StatusCode != http.StatusNotModified {
-			t.Fatalf("Expected 304 status code, got %v", reqInf.StatusCode)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, _ := TOSession.GetParameterByName(*parameter.Name)
+			if len(respParameter) > 0 {
+				parameterID := respParameter[0].ID
+				if parameterID > 0 {
+					t.Errorf("Parameter ID %d", parameterID)
+					resp, _, err := TOSession.GetProfileByParameterWithHdr(strconv.Itoa(parameterID), header)
+					if err != nil {
+						t.Fatalf("Expected no error, but got %v", err.Error())
+					}
+					if reqInf.StatusCode != http.StatusNotModified {
+						t.Fatalf("Expected 304 status code, got %v", reqInf.StatusCode)
+					}
+					if len(resp) < 1 {
+						t.Errorf("Expected atleast one response for Get Profile by Parameters, but found %d", len(resp))
+					}
+				} else {
+					t.Errorf("Invalid parameter ID %d", parameterID)
+				}
+			} else {
+				t.Errorf("No response found for GET Parameters by name")
+			}
 		}
 	}
 }
@@ -269,10 +287,25 @@ func GetTestProfiles(t *testing.T) {
 			t.Errorf("cannot GET Profile by name: %v - %v", err, resp)
 		}
 		profileID := resp[0].ID
-
-		resp, _, err = TOSession.GetProfileByParameter(pr.Parameter)
-		if err != nil {
-			t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, _ := TOSession.GetParameterByName(*parameter.Name)
+			if len(respParameter) > 0 {
+				parameterID := respParameter[0].ID
+				if parameterID > 0 {
+					resp, _, err = TOSession.GetProfileByParameter(strconv.Itoa(parameterID))
+					if err != nil {
+						t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+					}
+					if len(resp) < 1 {
+						t.Errorf("Expected atleast one response for Get Profile by Parameters, but found %d", len(resp))
+					}
+				} else {
+					t.Errorf("Invalid parameter ID %d", parameterID)
+				}
+			} else {
+				t.Errorf("No response found for GET Parameters by name")
+			}
 		}
 
 		resp, _, err = TOSession.GetProfileByCDNID(pr.CDNID)
--- a/traffic_ops/testing/api/v3/profiles_test.go
+++ b/traffic_ops/testing/api/v3/profiles_test.go
@@ -17,7 +17,6 @@ package v3
 
 import (
 	"net/http"
-	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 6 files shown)


traffic_ops/traffic_ops_golang/profile/profiles.go
```diff
@@ -21,11 +21,12 @@ package profile
 
 import (
 	"errors"
-	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/util/ims"
 	"net/http"
 	"strconv"
 	"time"
 
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/util/ims"
+
```
ocket8888, line 29: extraneous space

traffic_ops/traffic_ops_golang/profile/profiles.go
```diff
@@ -149,7 +150,7 @@ func (prof *TOProfile) Read(h http.Header, useIMS bool) ([]interface{}, error, e
 
 	rows, err := prof.ReqInfo.Tx.NamedQuery(query, queryValues)
 	if err != nil {
-		return nil, nil, errors.New("profile read querying: " + err.Error()), http.StatusInternalServerError, nil
+		return nil, errors.New("bad request found"), errors.New("profile read querying: " + err.Error()), http.StatusBadRequest, nil
```
ocket8888, line 153: So if you use `github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api.ParseDBError` you can get possibly some more information that will help the client correct their request, rather than just telling them it was bad.

dmohan001c: I tried using ParseDBError function getting the same 500 Internal error. Instead, I updated the code to print the exact errors. It returns 400 Bad Request.

traffic_ops/traffic_ops_golang/profile/profiles.go
```diff
@@ -149,7 +149,7 @@ func (prof *TOProfile) Read(h http.Header, useIMS bool) ([]interface{}, error, e
 
 	rows, err := prof.ReqInfo.Tx.NamedQuery(query, queryValues)
 	if err != nil {
-		return nil, nil, errors.New("profile read querying: " + err.Error()), http.StatusInternalServerError, nil
+		return nil, err, errors.New("profile read querying: " + err.Error()), http.StatusBadRequest, nil
```
srijeet0406, line 152: This is an internal database error and should probably remain as an internal server error.
Instead, what you can do is add another validation for the `param` query parameter to make sure it's an integer.
Also, you can add the `JOIN` clause for the `profile_parameter` to the main select query (just like the `cdn` join part), instead of checking for it separately.

Additionally, as mentioned in the GH issue, there are a lot of client methods that use the profile param as a string(name), instead of an int(id). We should probably fix those in this PR as well.

srijeet0406 (changes requested):

traffic_ops/traffic_ops_golang/profile/profiles.go
dmohan001c: removed the extraneous space.

traffic_ops/traffic_ops_golang/profile/profiles.go
```diff
@@ -149,7 +149,7 @@ func (prof *TOProfile) Read(h http.Header, useIMS bool) ([]interface{}, error, e
 
 	rows, err := prof.ReqInfo.Tx.NamedQuery(query, queryValues)
 	if err != nil {
-		return nil, nil, errors.New("profile read querying: " + err.Error()), http.StatusInternalServerError, nil
+		return nil, err, errors.New("profile read querying: " + err.Error()), http.StatusInternalServerError, nil
```
srijeet0406, line 152: As I mentioned in the previous comment, this should just stay the same as it was before.
You could add another validation for the param being an int.

srijeet0406 (changes requested):

traffic_ops/traffic_ops_golang/profile/profiles.go
dmohan001c: added another validation for the param and cdn to make sure it's an integer.

traffic_ops/traffic_ops_golang/profile/profiles.go
dmohan001c: done.

traffic_ops/traffic_ops_golang/profile/profiles.go
```diff
@@ -123,6 +124,10 @@ func (prof *TOProfile) Read(h http.Header, useIMS bool) ([]interface{}, error, e
 
 	// TODO add generic where clause to api.GenericRead
 	if paramValue, ok := prof.APIInfo().Params[ParamQueryParam]; ok {
+		//Check for interger value
```
srijeet0406, line 127: Instead of checking for individual query parameters like this, we should just add them as part of the checkers on lines 116-120.
You should also add the checker for the `ParamQueryParam` in there, something like this:
```	
queryParamsToQueryCols := map[string]dbhelpers.WhereColumnInfo{		
CDNQueryParam:  dbhelpers.WhereColumnInfo{Column: "c.id", Checker: api.IsInt},
NameQueryParam: dbhelpers.WhereColumnInfo{Column: "prof.name"},
IDQueryParam:   dbhelpers.WhereColumnInfo{Column: "prof.id", Checker: api.IsInt},
ParamQueryParam: dbhelpers.WhereColumnInfo{Column: "pp.parameter", Checker: api.IsInt},
}
```

Then, on line 126, if `ParamQueryParam` is present, you add the `Left Join...` subquery to the main select query.
This way, we can have one single place where we check for all the query params.

The API v2 and v3 tests currently fail because they are passing in a param name as an ID to the client methods. These will need to be fixed. Also, there is this section of code in the v4 tests, which can be uncommented after the fix:
```		// TODO: figure out what the 'Parameter' field of a Profile is and how
		// passing it to this is supposed to work.
		// resp, _, err = TOSession.GetProfileByParameter(pr.Parameter)
		// if err != nil {
		// 	t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
		// }```

**Review by srijeet0406 — CHANGES_REQUESTED** [review]

File: traffic_ops/traffic_ops_golang/profile/profiles.go (hunk 5 above)
Comment (dmohan001c) [reply traffic_ops/traffic_ops_golang/profile/profiles.go]: updated the DB Query to support params in GET profiles.

File: traffic_ops/traffic_ops_golang/profile/profiles.go [hunk 6]
```diff
@@ -113,19 +113,19 @@ func (prof *TOProfile) Read(h http.Header, useIMS bool) ([]interface{}, error, e
...
 	where, orderBy, pagination, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(prof.APIInfo().Params, queryParamsToQueryCols)
 
 	// Narrow down if the query parameter is 'param'
-
 	// TODO add generic where clause to api.GenericRead
 	if paramValue, ok := prof.APIInfo().Params[ParamQueryParam]; ok {
 		queryValues["parameter_id"] = paramValue
 		if len(paramValue) > 0 {
-			where += " LEFT JOIN profile_parameter pp ON prof.id  = pp.profile where pp.parameter=:parameter_id"
+			where = " LEFT JOIN profile_parameter pp ON prof.id = pp.profile " + where + " AND pp.parameter=:parameter_id"
```
srijeet0406, line 128: Once you add the `ParamQueryParam` to the `queryParamsToQueryCols`, the `BuildWhereAndOrderByAndPagination` function will automatically add the `WHERE` clause to the query, you don't need to modify the `where` variable above. Instead, what you can do is add just the ` LEFT JOIN profile_parameter pp ON prof.id  = pp.profile` part if the `parameter_id` param is present.

With the current changes, the query comes out to be this:
``
SELECT
prof.description,
prof.id,
prof.last_updated,
prof.name,
prof.routing_disabled,
prof.type,
c.id as cdn,
c.name as cdn_name
FROM profile prof
LEFT JOIN cdn c ON prof.cdn = c.id LEFT JOIN profile_parameter pp ON prof.id = pp.profile
WHERE pp.parameter=:param AND pp.parameter=:parameter_id
``
We don't need the two `WHERE` clauses in our query. Also, I think the tests are still failing for API v2 and v3.

srijeet0406 (changes requested):

traffic_ops/traffic_ops_golang/profile/profiles.go
dmohan001c: with the current changes the query comes out to be like this: 
`SELECT prof.description, prof.id, prof.last_updated, prof.name, prof.routing_disabled, prof.type, c.id as cdn, c.name as cdn_name FROM profile prof 
LEFT JOIN cdn c ON prof.cdn = c.id LEFT JOIN profile_parameter pp ON prof.id = pp.profile 
WHERE pp.parameter=:param AND pp.parameter=:parameter_id`

dmohan001c: If I didn't change the where as per above statement, the query looks incorrect. 
SELECT prof.description, prof.id, prof.last_updated, prof.name, prof.routing_disabled, prof.type, c.id as cdn, c.name as cdn_name FROM profile prof
LEFT JOIN cdn c ON prof.cdn = c.id
WHERE pp.parameter=:param 
LEFT JOIN profile_parameter pp ON prof.id  = pp.profile.

dmohan001c: V2 AND V3 profiles tests are failing. I am taking a look on it.

srijeet0406: No, the `WHERE` clause will be created by the `BuildWhereAndOrderByAndPagination` function and you dont need to account for it, if you're passing in all the right query params into it.

dmohan001c: Fixed.

traffic_ops/testing/api/v3/profiles_test.go
```diff
@@ -90,8 +89,7 @@ func GetTestProfilesIMS(t *testing.T) {
 			if len(respParameter) > 0 {
 				parameterID := respParameter[0].ID
 				if parameterID > 0 {
-					t.Errorf("Parameter ID %d", parameterID)
-					resp, _, err := TOSession.GetProfileByParameterWithHdr(strconv.Itoa(parameterID), header)
+					resp, _, err := TOSession.GetProfileByParameterWithHdr(parameterID, nil)
```
srijeet0406, line 92: You can't actually change the existing client signatures because that breaks the clients that are using them. I suggest creating a new method that takes in the parameterID, instead of the parameter name, and then using that method in the tests. You can leave the `GetProfileByParameterWithHdr` method as it is.

srijeet0406 (changes requested):

traffic_ops/v3-client/profile.go
```diff
@@ -126,16 +126,16 @@ func (to *Session) GetProfileByName(name string) ([]tc.Profile, toclientlib.ReqI
 	return to.GetProfileByNameWithHdr(name, nil)
 }
 
-func (to *Session) GetProfileByParameterWithHdr(param string, header http.Header) ([]tc.Profile, toclientlib.ReqInf, error) {
-	URI := fmt.Sprintf("%s?param=%s", APIProfiles, url.QueryEscape(param))
+func (to *Session) GetProfileByParameterWithHdr(param int, header http.Header) ([]tc.Profile, toclientlib.ReqInf, error) {
```
srijeet0406, line 129: You can't change the signatures of these methods, if this code is not a part of a major release. There might be other client scripts using these methods, which will break if you change the signature. Instead, I would suggest doing something like you did in the `v2` client method above. That way, the method signature stays the same.

srijeet0406 (changes requested):

traffic_ops/testing/api/v3/profiles_test.go
dmohan001c: reverted the method signature

traffic_ops/v2-client/profile.go
```diff
@@ -164,7 +164,8 @@ func (to *Session) GetProfileByName(name string) ([]tc.Profile, ReqInf, error) {
 
 // GetProfileByParameter GETs a Profile by the Profile "param".
 func (to *Session) GetProfileByParameter(param string) ([]tc.Profile, ReqInf, error) {
-	URI := fmt.Sprintf("%s?param=%s", API_PROFILES, url.QueryEscape(param))
+	paramId, _ := strconv.Atoi(url.QueryEscape(param))
```
srijeet0406, line 167: We should probably check for the error here.

srijeet0406: Also, as I suggested earlier, instead of breaking the way pre existing client methods work, why not create a new client method that takes in a param ID instead of a param name, and then make the `GET` call? Something like this:

```
func (to *Session) GetProfileByParameterID(paramID string, header http.Header) ([]tc.Profile, ReqInf, error) {
         uri := fmt.Sprintf("%s?param=%d", APIProfiles, paramID)
	var data tc.ProfilesResponse
	reqInf, err := to.get(uri, header, &data)
	return data.Response, reqInf, err
}
```

traffic_ops/v3-client/profile.go
```diff
@@ -127,7 +127,7 @@ func (to *Session) GetProfileByName(name string) ([]tc.Profile, toclientlib.ReqI
 }
 
 func (to *Session) GetProfileByParameterWithHdr(param string, header http.Header) ([]tc.Profile, toclientlib.ReqInf, error) {
-	URI := fmt.Sprintf("%s?param=%s", APIProfiles, url.QueryEscape(param))
+	URI := fmt.Sprintf("%s?param=%s", APIProfiles, param)
```
srijeet0406, line 130: Same comment as above here.

srijeet0406 (changes requested):

traffic_ops/v2-client/profile.go
dmohan001c: I don't think a new client is required, and I verified all the existing call to this method are passed.

traffic_ops/v3-client/profile.go
dmohan001c: Fixed.

traffic_ops/v2-client/profile.go
dmohan001c: I don't want to create a new method, since the bug fixes requires the existing method to be changed to accept int parameters, and I verified the existing methods call are updated as per new fix.

dmohan001c: If we keep the existing methods as per your advice, if someone accidentally calls the existing method, it will throw an error, since that API is not anymore supported string.

dmohan001c: Created a new client method

traffic_ops/testing/api/v4/profiles_test.go
```diff
@@ -517,12 +517,31 @@ func GetTestProfiles(t *testing.T) {
 		}
 		profileID := resp.Response[0].ID
 
-		// TODO: figure out what the 'Parameter' field of a Profile is and how
```
srijeet0406, line 520: You can probably remove these commented lines now.

dmohan001c: Fixed.

traffic_ops/testing/api/v4/profiles_test.go
```diff
@@ -518,11 +518,31 @@ func GetTestProfiles(t *testing.T) {
 		profileID := resp.Response[0].ID
 
 		// TODO: figure out what the 'Parameter' field of a Profile is and how
```
srijeet0406, line 520: Any reason to keep that first line in there still?

traffic_ops/testing/api/v2/profiles_test.go
```diff
@@ -210,12 +210,26 @@ func GetTestProfiles(t *testing.T) {
 			t.Errorf("cannot GET Profile by name: %v - %v", err, resp)
 		}
 		profileID := resp[0].ID
-
-		resp, _, err = TOSession.GetProfileByParameter(pr.Parameter)
-		if err != nil {
-			t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, _ := TOSession.GetParameterByName(*parameter.Name)
```
srijeet0406, line 215: We should probably check for the returned error here.

traffic_ops/testing/api/v3/profiles_test.go
```diff
@@ -83,12 +83,28 @@ func GetTestProfilesIMS(t *testing.T) {
...
 		}
-		_, reqInf, err = TOSession.GetProfileByParameterWithHdr(pr.Parameter, header)
-		if err != nil {
-			t.Fatalf("Expected no error, but got %v", err.Error())
-		}
-		if reqInf.StatusCode != http.StatusNotModified {
-			t.Fatalf("Expected 304 status code, got %v", reqInf.StatusCode)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, _ := TOSession.GetParameterByName(*parameter.Name)
```
srijeet0406, line 88: We should probably check for the returned error here.

traffic_ops/testing/api/v3/profiles_test.go
```diff
@@ -269,10 +285,25 @@ func GetTestProfiles(t *testing.T) {
 			t.Errorf("cannot GET Profile by name: %v - %v", err, resp)
 		}
 		profileID := resp[0].ID
-
-		resp, _, err = TOSession.GetProfileByParameter(pr.Parameter)
-		if err != nil {
-			t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, _ := TOSession.GetParameterByName(*parameter.Name)
```
srijeet0406, line 290: We should probably check for the returned error here.

traffic_ops/testing/api/v4/profiles_test.go
dmohan001c: I removed this line.

traffic_ops/testing/api/v2/profiles_test.go
dmohan001c: added error checks

traffic_ops/testing/api/v3/profiles_test.go
dmohan001c: added error checks.

traffic_ops/testing/api/v3/profiles_test.go
dmohan001c: added error checks

srijeet0406:
@dmohan001c It looks like the `go format` checks are failing. Once that is fixed, I can test this and approve.

traffic_ops/testing/api/v2/profiles_test.go
```diff
@@ -210,12 +210,29 @@ func GetTestProfiles(t *testing.T) {
...
 		profileID := resp[0].ID
-
-		resp, _, err = TOSession.GetProfileByParameter(pr.Parameter)
-		if err != nil {
-			t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
+			if err != nil {
+				t.Errorf("Cannot GET Parameter by name: %v - %v", err, resp)
```
srijeet0406, line 217: This `resp` is the wrong response here. Also, I don't think you need to print the response here, just the error is fine.

traffic_ops/testing/api/v2/profiles_test.go
```diff
@@ -210,12 +210,29 @@ func GetTestProfiles(t *testing.T) {
...
+			}
+			if len(respParameter) > 0 {
+				parameterID := respParameter[0].ID
+				if parameterID > 0 {
+					resp, _, err = TOSession.GetProfileByParameterId(parameterID)
+					if err != nil {
+						t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
```
srijeet0406, line 224: Same comment here, no need to print the response.

traffic_ops/testing/api/v3/profiles_test.go
```diff
@@ -83,12 +83,31 @@ func GetTestProfilesIMS(t *testing.T) {
...
-		if err != nil {
-			t.Fatalf("Expected no error, but got %v", err.Error())
-		}
-		if reqInf.StatusCode != http.StatusNotModified {
-			t.Fatalf("Expected 304 status code, got %v", reqInf.StatusCode)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
+			if err != nil {
+				t.Errorf("Cannot GET Parameter by name: %v - %v", err, respParameter)
```
srijeet0406, line 90: Same.

traffic_ops/testing/api/v3/profiles_test.go
```diff
@@ -269,10 +288,28 @@ func GetTestProfiles(t *testing.T) {
...
 		}
 		profileID := resp[0].ID
-
-		resp, _, err = TOSession.GetProfileByParameter(pr.Parameter)
-		if err != nil {
-			t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			respParameter, _, err := TOSession.GetParameterByName(*parameter.Name)
+			if err != nil {
```
srijeet0406, line 294: same

traffic_ops/testing/api/v4/profiles_test.go
```diff
@@ -517,12 +517,31 @@ func GetTestProfiles(t *testing.T) {
...
-		// TODO: figure out what the 'Parameter' field of a Profile is and how
-		// passing it to this is supposed to work.
-		// resp, _, err = TOSession.GetProfileByParameter(pr.Parameter)
-		// if err != nil {
-		// 	t.Errorf("cannot GET Profile by param: %v - %v", err, resp)
-		// }
+		if len(pr.Parameters) > 0 {
+			parameter := pr.Parameters[0]
+			opts.QueryParameters.Set("name", *parameter.Name)
+			respParameter, _, _ := TOSession.GetParameters(opts)
```
srijeet0406, line 523: You should be checking the error here, as you are doing in the v2 and v3 tests.

traffic_ops/v3-client/profile.go
```diff
@@ -139,6 +139,19 @@ func (to *Session) GetProfileByParameter(param string) ([]tc.Profile, toclientli
 	return to.GetProfileByParameterWithHdr(param, nil)
 }
 
+func (to *Session) GetProfileByParameterIdWithHdr(param int, header http.Header) ([]tc.Profile, toclientlib.ReqInf, error) {
```
srijeet0406, line 142: Godoc is missing.

traffic_ops/v2-client/profile.go
```diff
@@ -178,6 +178,21 @@ func (to *Session) GetProfileByParameter(param string) ([]tc.Profile, ReqInf, er
 	return data.Response, reqInf, err
 }
 
+func (to *Session) GetProfileByParameterId(param int) ([]tc.Profile, ReqInf, error) {
```
srijeet0406, line 181: Godoc is missing.

traffic_ops/testing/api/v2/profiles_test.go
dmohan001c: removed the response in error statement

traffic_ops/testing/api/v2/profiles_test.go
dmohan001c: removed the response from print statement

traffic_ops/testing/api/v3/profiles_test.go
dmohan001c: Removed.

traffic_ops/testing/api/v4/profiles_test.go
dmohan001c: added error checks

traffic_ops/testing/api/v3/profiles_test.go
dmohan001c: removed

traffic_ops/v2-client/profile.go
dmohan001c: Added godoc,.

traffic_ops/v3-client/profile.go
dmohan001c: Added godoc.

srijeet0406 (approved):
Code changes look good.
All tests pass.
Manual testing looks good as well.
LGTM!

ocket8888 (approved):

PR #6005, opened 2021-07-08
TP: Create AG-Grid Component and Convert DS Server Assignment & Cache Checks

### LINKED ISSUE #5530: Create TP Component for AG Grids
## I'm submitting a ...

-  Traffic Portal

## Current behavior:
Currently every place where ag-grid is used, most of the code is copy and pasted from the other existing pages. As a result, improvements to a single grid requires updating all of the grids. Currently this is 5 distinct locations (DS Requests, DS, Servers, ChangeLog, Jobs) but will continue to grow and get more unmanageable.

## New behavior:
Create a component that each page can call into that provides a single location for ag-grid.

Look at the code

shamrickus:

- [x] This PR fixes #5530 

It also converts two data-tables to ag-grid:
DS Server Assignment
Cache Checks

- Traffic Portal

Ensure the DS Server Assignment table works (able to assign/unassign servers, already assigned servers are checked, design and flow of new table works).
Ensure the Cache Checks table works (design and flow make sense, revalPending/updPending show a clock/check, the extension columns show either an x or a check, able to drill into the table on click).

Run CiaB off of master, switch to this branch and run TP locally, this makes comparing & contrasting the tables easy.

## Additional Information
There is functionality on the ag-grid component that is not used by the two converted tables, this is because I've done some conversions of other tables (not ready for PR yet) that needed those features.


traffic_portal/app/src/common/modules/table/agGrid/CommonGridController.d.ts
```diff
@@ -0,0 +1,65 @@
...
+        onClick(row: any): void;
+        isDisabled(row: any): boolean;
+        shown(row: any): void;
+
+        // If href is undefined, getHref is called
+        href: string;
+        getHref(row: any): string;
+
+        // If text is undefined, getText is called
+        text: string;
```
ocket8888, line 50: According to this typing, `href` and `text` can't be undefined

traffic_portal/app/src/common/modules/table/agGrid/CommonGridController.d.ts
```diff
@@ -0,0 +1,65 @@
...
+        getText(): string;
+    }
+
+    export interface DropDownOption extends CommonOption {
+    }
+
+    export interface ContextMenuOption extends CommonOption{
+    }
+
+}
```
ocket8888, line 65: Can we do any better than `any` for the rows passed around in these typings?

traffic_portal/app/src/common/modules/table/agGrid/CommonGridController.d.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export namespace CGC {
+    export enum OptionType {
+        Seperator = 0,
+        Button = 1,
+        Ancor = 2
+    }
+
+    export interface GridSettings {
+        rowClassRules?: any;
```
ocket8888, line 28: I think this is a separate issue from rows being `any` - can this be typed more strictly?

traffic_portal/app/src/common/modules/table/agGrid/CommonGridController.d.ts
shamrickus: If I add the typings for ag-grid, then a couple of the `any`s could be transformed, I was just trying to avoid installing any typings since this is the only place they're used (and we're transitioning away from this anyways)

traffic_portal/app/src/common/modules/table/agGrid/CommonGridController.d.ts
shamrickus: See above comment, this is "shadowing" an ag-grid type.

traffic_portal/app/src/common/modules/table/agGrid/CommonGridController.d.ts
ocket8888: Shouldn't you just be able to import the AG-grid typings in this file? They're included in the package, which is already a dependency.

shamrickus: You are correct

shamrickus: For the `CommonOption` methods, not really. It's literally the raw row data, which ag-grid also types as any.

ocket8888: Well, alright, they'd know I suppose.

traffic_portal/app/src/common/modules/table/deliveryServiceServers/TableAssignDSServersController.js
```diff
@@ -17,87 +17,54 @@
...
-		}
-	};
-
-	$scope.onChange = function() {
-		updateSelectedCount();
-	};
-
 	$scope.submit = function() {
-		var selectedServerIds = _.pluck(selectedServers, 'id');
+		const selectedServerIds = this.selectedServers.map(s => s["id"]);
```
ocket8888, line 55: Since when does our build system allow fat-arrow functions? I had to try it myself to believe it worked - that used to break in half.

ocket8888 (approved):

traffic_portal/app/src/common/modules/table/deliveryServiceServers/TableAssignDSServersController.js
shamrickus: /shrug

PR #6009, opened 2021-07-09
Logic to capture/display the last time a user successfully authenticated

### LINKED ISSUE #5412: Capture/display the last time a user successfully authenticated
## I'm submitting a ...

-  Traffic Ops
-  Traffic Portal

## Current behavior:
TP and the TO API will not provide you with any information regarding the last time/date a user successfully authenticated. This information can be helpful when disabling users that no longer use the system. i.e. data cleanup.

## New behavior:
When a user successfully authenticates (POST /user/login), capture the time/date and provide this information via the TO api and TP similar to the way user.lastUpdated is captured. This info should be added to:

TO API:

- GET /user/current
- GET /users
- GET /user?id=

TP:

- https://tp.domain.tld/#!/users
- https://tp.domain.tld/#!/users/{id}
- https://tp.domain.tld/#!/user

rimashah25:

- [x] This PR fixes #5412  

- Documentation
- Traffic Ops
- Traffic Portal
- CI tests

1. Use GET curl calls for users/user api
```
https://<TO-API host:port>/api/4.0/user/current
OR 
https://<TO-API host:port>/api/4.0/users
OR 
https://<TO-API host:port>/api/4.0/users?id=<id>
```

One should see **last_authenticated** field in the response struct

2. Login to TP, go to Users section, click on select table column image, and you should see the `Last Authenticated` time if a user was successfully authenticated.
image

3. Run API/v4 tests for users


rimashah25:
Ready for review.

mitchell852:
image

@rimashah25 is that UTC time? if so, want to put that in the column label? I think the jobs table does something similar.

lib/go-tc/users.go
```diff
@@ -107,20 +109,27 @@ type User struct {
...
 	UserName  *string `json:"username"`
 	LocalUser *bool   `json:"localUser"`
 	RoleName  *string `json:"roleName"`
 	commonUserFields
 }
 
+// UserCurrentV40 contains LastAuthenticated field.
+type UserCurrentV40 struct {
+	UserCurrent
+	LastAuthenticated *TimeNoMod `json:"lastAuthenticated" db:"last_authenticated"`
```
ocket8888, line 130: Can we just make this a `time.Time`? That way it comes out as RFC3339-formatted instead of our proprietary format.

rimashah25: ok.

mitchell852:
> @rimashah25 is that UTC time? if so, want to put that in the column label and value? I think the jobs table does something similar.

hmm. even better imo would be "relative time" . i.e. "one month ago", "30 minutes ago". there are example in other tables i think. search for getRelativeTime

rimashah25:
> getRelativeTime

Ok, I'll check it out.

docs/source/api/v4/user_current.rst
```diff
@@ -35,28 +35,29 @@ No parameters available.
...
+:addressLine1:      The user's address - including street name and number
+:addressLine2:      An additional address field for e.g. apartment number
+:city:              The name of the city wherein the user resides
+:company:           The name of the company for which the user works
+:country:           The name of the country wherein the user resides
+:email:             The user's email address
+:fullName:          The user's full name, e.g. "John Quincy Adams"
+:gid:               A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user
+:id:                An integral, unique identifier for this user
+:lastAuthenticated: The date and time at which the user was last authenticated, in :ref:`RFC3339`
```
ocket8888, line 47: > `undefined label: rfc3339`

I think you want:

```rst
:rfc:3339 or :rfc:`3339`
```

either of those work fine. A lot of people prefer to always use the <kbd>`</kbd>s because they are required in some cases, but in this case they aren't.

docs/source/api/v4/users.rst
```diff
@@ -68,29 +68,30 @@ Request Structure
...
+:addressLine2:      An additional address field for e.g. apartment number
+:changeLogCount:    The number of change log entries created by the user
+:city:              The name of the city wherein the user resides
+:company:           The name of the company for which the user works
+:country:           The name of the country wherein the user resides
+:email:             The user's email address
+:fullName:          The user's full name, e.g. "John Quincy Adams"
+:gid:               A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user - now it is always ``null``
+:id:                An integral, unique identifier for this user
+:lastAuthenticated: The date and time at which the user was last authenticated, in :ref:`RFC3339`
```
ocket8888, line 81: same as above RE: undefined label

docs/source/api/v4/users_id.rst
```diff
@@ -48,28 +48,29 @@ Request Structure
...
+:addressLine1:      The user's address - including street name and number
+:addressLine2:      An additional address field for e.g. apartment number
+:city:              The name of the city wherein the user resides
+:company:           The name of the company for which the user works
+:country:           The name of the country wherein the user resides
+:email:             The user's email address
+:fullName:          The user's full name, e.g. "John Quincy Adams"
+:gid:               A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user - now it is always ``null``
+:id:                An integral, unique identifier for this user
+:lastAuthenticated: The date and time at which the user was last authenticated, in :ref:`RFC3339`
```
ocket8888, line 60: same as above RE: undefined label

traffic_ops/testing/api/v4/user_test.go
```diff
@@ -454,6 +454,12 @@ func GetTestUsers(t *testing.T) {
 	if err != nil {
 		t.Errorf("cannot get users: %v - alerts: %+v", err, resp.Alerts)
 	}
+	if resp.Response == nil {
+		t.Fatalf("expected a users list, got nothing")
+	}
+	if resp.Response[0].LastAuthenticated == nil {
```
ocket8888, line 460: This line will still segfault if Traffic Ops returns an empty list - if you just check for `len(resp.Response) < 1` or similar it will also handle `nil` properly since `len([]string(nil)) == 0`

rimashah25: Ugghh.. was thinking len() but then did nil..

docs/source/api/v4/user_current.rst
```diff
@@ -44,7 +44,7 @@ Response Structure
 :fullName:          The user's full name, e.g. "John Quincy Adams"
 :gid:               A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user
 :id:                An integral, unique identifier for this user
-:lastAuthenticated: The date and time at which the user was last authenticated, in :ref:`RFC3339`
+:lastAuthenticated: The date and time at which the user was last authenticated, in :rfc:`RFC3339`
```
ocket8888, line 47: > `invalid RFC number RFC3339`

You don't need to put "RFC" in the argument to `:rfc`, that's implicit. The argument is just the number, like:
```rst
:rfc:`3339`
```

docs/source/api/v4/users.rst
```diff
@@ -78,7 +78,7 @@ Response Structure
 :fullName:          The user's full name, e.g. "John Quincy Adams"
 :gid:               A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user - now it is always ``null``
 :id:                An integral, unique identifier for this user
-:lastAuthenticated: The date and time at which the user was last authenticated, in :ref:`RFC3339`
+:lastAuthenticated: The date and time at which the user was last authenticated, in :rfc:`RFC3339`
```
ocket8888, line 81: same as above

docs/source/api/v4/users_id.rst
```diff
@@ -57,7 +57,7 @@ Response Structure
 :fullName:          The user's full name, e.g. "John Quincy Adams"
 :gid:               A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX group ID of the user - now it is always ``null``
 :id:                An integral, unique identifier for this user
-:lastAuthenticated: The date and time at which the user was last authenticated, in :ref:`RFC3339`
+:lastAuthenticated: The date and time at which the user was last authenticated, in :rfc:`RFC3339`
```
ocket8888, line 60: same as above

traffic_portal/app/src/common/modules/table/roleUsers/table.roleUsers.tpl.html
```diff
@@ -64,6 +65,7 @@
                 <td data-search="^{{::u.tenant}}$">{{::u.tenant}}</td>
                 <td data-search="^{{::u.rolename}}$">{{::u.rolename}}</td>
                 <td data-search="^{{::u.registrationSent}}$">{{::u.registrationSent}}</td>
+                <td data-search="^{{::getRelativeTime(u.lastAuthenticated)}}$">{{::getRelativeTime(u.lastAuthenticated)}}</td>
```
ocket8888, line 68: `getRelativeTime` doesn't handle null values, so if a user has never logged in the table will display "Invalid date" as their last authentication time. It should say "never" or similar.

This will happen whenever a new user is created or registered until they then log in.

ocket8888 (commented):
Everything seems to work fine with the exception of my one comment about handling an edge case, but I did have one last concern: since we're saving the authentication time to the database, login is now a modifying operation on a user. Tangibly this just means that every time you log in it updates `lastUpdated`:

```
ocket8888 $ toget -kpa 4.0 'users?username=admin' | jq '.response[0].lastUpdated'
"2021-07-15 18:45:27+00"
ocket8888 $ toget -kpa 4.0 'users?username=admin' | jq '.response[0].lastUpdated'
"2021-07-15 18:45:30+00"
```

(`toget` re-authenticates on every call)

I don't necessarily think this is a problem, I just want to make sure we knew this was going to happen and that it was determined to not be a problem. I can't think of any kind of big caching implications or anything.

traffic_portal/app/src/common/modules/table/roleUsers/table.roleUsers.tpl.html
rimashah25: Done

rimashah25:
> Everything seems to work fine with the exception of my one comment about handling an edge case, but I did have one last concern: since we're saving the authentication time to the database, login is now a modifying operation on a user. Tangibly this just means that every time you log in it updates `lastUpdated`:
> 
> ```
> ocket8888 $ toget -kpa 4.0 'users?username=admin' | jq '.response[0].lastUpdated'
> "2021-07-15 18:45:27+00"
> ocket8888 $ toget -kpa 4.0 'users?username=admin' | jq '.response[0].lastUpdated'
> "2021-07-15 18:45:30+00"
> ```
> 
> (`toget` re-authenticates on every call)
> 
> I don't necessarily think this is a problem, I just want to make sure we knew this was going to happen and that it was determined to not be a problem and this is intentional. I can't think of any kind of big caching implications or anything.

Yes, that is an expected behaviour where the field `lastUpdated` will be updated for two reasons: 1. when user logins and 2.  if a user makes any changes to the user profile.  
The field `lastAuthenticated` will be updated only when a user logs in.

ocket8888:
Okay, as long as it was considered.

ocket8888 (approved):

mitchell852:
@rimashah25 @ocket8888 sorry for the late comment on this but i'm a little concerned about the text that shows up in the UI table if last_authenticated = null ("Never Logged In"). this isn't exactly true, right? wondering if leaving the UI table cell blank would be a better option as "Never Logged In" can be a bit misleading.

ocket8888:
In what scenario will that not be true? The only way I can think of that field being `NULL` - apart from database manipulation that skips the API - is if the user has never logged in.

mitchell852:
> In what scenario will that not be true? The only way I can think of that field being `NULL` - apart from database manipulation that skips the API - is if the user has never logged in.

"Never Logged In" really means Never Logged In since this feature was added, right?

ocket8888:
That is true, but I don't think it's a big deal. If the users are active it'll fix itself pretty quickly. If they aren't, then this tells you that in a roundabout way, so it's not useless.

PR #6020, opened 2021-07-13
Update requests version required by the TO Python client

### LINKED ISSUE #6011: Migrate to the next version of Python requests when released
Hello Maintainers, 

I am a PMC member of Apache Airflow, and I wanted to give you a bit of heads-up with rather important migration to the upcoming version of `requests` library in your Python release. 

Since you are using `requests` library in your project, you are affected.

As discussed at length in https://issues.apache.org/jira/browse/LEGAL-572  we found out that the `chardet` library used by `requests` library was a mandatory dependency to requests and since it has LGPL licence, we should not release any Apache Software with it. 

Since then (and since in Airflow we rely on requests heavily) we have been working with the requests maintainers and "charset-normalizer" maintainer to make it possible to replace `chardet` with MIT-licensed `charset-normalizer` instead so that requests library can be used in Python releases by Apache projects.

This was a bumpy road but finally the PR by @ashb has been merged: https://github.com/psf/requests/pull/5797 and we hope soon a new version of requests library will be released. 

This is just a heads-up. I will let you know when it is released, but I have a kind requests as well - I might ask the maintainers to release a release candidate of requests and maybe you could help to test it before it is released, that would be some re-assurance for the maintainers of requests who are very concerned about stability of their releases.

Let me know if you need any more information and whether you would like to help in testing the candidate when it is out.

rawlinp:
- [x] This PR fixes #6011

Install the python TO client which should also install the latest version of the `requests` library, and manually test out some of the methods.

- [x] TO Python client doesn't have tests
- [x] Dependency upgrade, no docs needed
- [x] minor issue, no changelog necessary

### DIFF
--- a/traffic_control/clients/python/setup.py
+++ b/traffic_control/clients/python/setup.py
@@ -72,7 +72,7 @@ with open(os.path.join(HERE, "README.rst")) as fd:
 	       packages=find_packages(exclude=["contrib", "docs", "tests"]),
 	       install_requires=[
 	           "future>=0.16.0",
-	           "requests>=2.13.0",
+	           "requests>=2.26.0",
 	           "munch>=2.1.1",
 	       ],
 	       entry_points={


traffic_control/clients/python/setup.py
```diff
@@ -72,7 +72,7 @@
 	       packages=find_packages(exclude=["contrib", "docs", "tests"]),
 	       install_requires=[
 	           "future>=0.16.0",
-	           "requests>=2.13.0",
+	           "requests>=2.26.0",
```
zrhoffman, line 75: Should this be `requests[charset_normalizer]` instead of `requests`?

rawlinp: No, this newer version of requests will automatically install charset_normalizer and use that if chardet isn't already installed (which it wouldn't be on fresh installs).

zrhoffman (approved):

PR #6023, opened 2021-07-14
Fix Internal server error in update cachegroups by already exist cg name

### LINKED ISSUE #4442: Internal Server Error for PUT Request on cachegroups/{{ID}} endpoint
## I'm submitting a ...

- [x] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When user do a PUT Request to cachegroups/id endpoint with the `name` of another cachegroup, It will return 500 Internal Error.

Login to TO. Do a PUT Request `https://{{TO_BASE_URL}}/api/2.0/cachegroups/08` With this replacement json payload: 
```json
{
        "latitude": 0.0,
        "longitude": 0.0,
        "name": "test",
        "fallbacks": [],
        "fallbackToClosest": true,
        "shortName": "test",
        "typeId": 23,
        "localizationMethods": ["GEO"]
}
```

It will return status 500 Internal Server Error 
```json
{
    "alerts": [
        {
            "level": "error"
        }
    ]
}
```
## Expected / new behavior:

It should return bad request.

dmohan001c:

- [x] This PR fixes #4442  

This fixes the internal server error causing in update cache group with the already existing cache group name

- CDN in a Box
- Traffic Control Client 
- Traffic Ops
- CI tests

* Master


traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
```diff
@@ -688,7 +686,7 @@ func (cg *TOCacheGroup) handleCoordinateUpdate() (*int, error, error, int) {
 	}
 
 	if err = cg.updateCoordinate(); err != nil {
-		return nil, nil, tc.DBError, http.StatusInternalServerError
+		return nil, err, tc.DBError, http.StatusBadRequest
```
srijeet0406, line 689: So since this might actually be a database error, this should still return an Internal Server Error in my opinion. 
What you could do is add a check at the start of the `Update` method checking to see if the cachegroup with the given name(from the request body) has the same ID as that passed in the request. If true, then proceed, else, return a 400 or 409 status code with the correct error (something like `cachegroup name should be unique`).

Also, could you pls add some tests for this?

srijeet0406: Additionally, the `tc.DBError` is now deprecated. You can just return `err` in its place, and `nil` in place of the user error.

srijeet0406:
This PR shows 0 files changed, was this a merge issue?

srijeet0406:
@dmohan001c Is there another PR that I need to review for this issue, since this one is closed?

Issue #2429, opened 2018-06-19
Autocreate self-signed SSL certificates


jhg03a:
When creating a new Delivery service or modifying the type to support HTTPS, a default self-signed certificate should be created.  This could be triggered by a miss when looking up the DS in riak.  This specifically prevents situations where a CDN could be queued by a second user unaware of a DS change in flight.  This is not prevented by DS Requests since you must fulfill the DS before you could create SSL certs.  If this situation happens and ORT runs an update before SSL Keys are in place, a syncds will fail.

jhg03a:
With the version of TR in the repo today, this can also cause TrafficRouter to also stop processing CRConfigs.

dneuman64:
TR not processing the CrConfig due to missing certificates has been in TR since 1.7 when we added HTTPS support.

mitchell852:
also, see #2159 to figure out if implementing Let's encrypt would nullify the need for this

dneuman64:
I don't think this is the way to solve this problem.  We should be better with making changes more atomic.  Therefore, I don't think this is something we will plan on fixing.

PR #6024, opened 2021-07-14
Generate a self signed cert on ds creation/update

### LINKED ISSUE #5893: Snapshotting a CDN that has an HTTPS delivery service w/ no cert causes TR crconfig reload failure
## I'm submitting a ...
-  bug report

-  Traffic Ops
-  Traffic Router

## Current behavior:
TR will not load a snapshot that has an HTTPs delivery service with a missing cert, therefore, TR will be stuck with an old snapshot until the problem is resolved and a cert is created for the HTTPS ds or the HTTPS ds is switched to HTTP.

## Expected behavior:
Prevent the creation of an invalid snapshot or TR should handle invalid snapshots more gracefully. 

- Create a delivery service with protocol=HTTPS. Do NOT add a cert to the delivery service.
- snapshot the cdn for the delivery service
- TR will not reload the new snapshot due to the missing cert

## Anything else:
Possible solutions:
1. when snapshotting a cdn via the api, return 400 if any of the cdn's https delivery services are missing a cert.
2. [auto-generate a self-signed cert](https://github.com/apache/trafficcontrol/issues/2429) when creating an https ds or updating an http ds to an https ds.
3. TR could somehow filter out the bad ds's and consume the rest of the snapshot?

mattjackson220:

- [x] This PR fixes #5893  

This PR updates TO so that when a new HTTPS delivery service is created or an HTTP delivery service is updated to HTTPS, a self signed certificate is created as a placeholder.

- Traffic Ops

Create a new HTTPS DS and verify that a self signed certificate is generated and stored
Create an HTTP DS and verify that no cert is generated
Update that HTTP DS to use HTTPS and verify that a self signed cert is generated and stored
Update an existing HTTPS DS that already has a cert and verify that the previous cert remains unchanged
Verify that all other DS creation / update functionality is unchanged

This test does not include docs or tests since it is a minor change that will not impact these

### DIFF
--- a/traffic_ops/traffic_ops_golang/config/config.go
+++ b/traffic_ops/traffic_ops_golang/config/config.go
@@ -60,6 +60,7 @@ type Config struct {
 	Version                string
 	UseIMS                 bool `json:"use_ims"`
 	RoleBasedPermissions   bool `json:"role_based_permissions"`
+	DefaultCertificateInfo *DefaultCertificateInfo `json:"default_certificate_info"`
 }
 
 // ConfigHypnotoad carries http setting for hypnotoad (mojolicious) server
@@ -171,6 +172,14 @@ type ConfigAcmeAccount struct {
 	AcmeUrl      string `json:"acme_url"`
 	Kid          string `json:"kid"`
 	HmacEncoded  string `json:"hmac_encoded"`
+}
+
+type DefaultCertificateInfo struct {
+	BusinessUnit string `json:"business_unit"`
+	City         string `json:"city"`
+	Organization string `json:"organization"`
+	Country      string `json:"country"`
+	State        string `json:"state"`
 }
 
 // ConfigDatabase reflects the structure of the database.conf file
--- a/traffic_ops/traffic_ops_golang/config/config.go
+++ b/traffic_ops/traffic_ops_golang/config/config.go
@@ -182,6 +182,30 @@ type DefaultCertificateInfo struct {
 	State        string `json:"state"`
 }
 
+func (d *DefaultCertificateInfo) Validate() (error, bool) {
+	missingList := []string{}
+	if d.BusinessUnit == "" {
+		missingList = append(missingList, "BusinessUnit")
+	}
+	if d.City == "" {
+		missingList = append(missingList, "City")
+	}
+	if d.Organization == "" {
+		missingList = append(missingList, "Organization")
+	}
+	if d.Country == "" {
+		missingList = append(missingList, "Country")
+	}
+	if d.State == "" {
+		missingList = append(missingList, "State")
+	}
+
+	if len(missingList) != 0 {
+		return fmt.Errorf("default certificate information is missing: %s", missingList), false
+	}
+	return nil, true
+}
+
 // ConfigDatabase reflects the structure of the database.conf file
 type ConfigDatabase struct {
 	Description string `json:"description"`
--- a/traffic_ops/traffic_ops_golang/config/config.go
+++ b/traffic_ops/traffic_ops_golang/config/config.go
@@ -58,8 +58,8 @@ type Config struct {
 	InfluxEnabled          bool
 	InfluxDBConfPath       string `json:"influxdb_conf_path"`
 	Version                string
-	UseIMS                 bool `json:"use_ims"`
-	RoleBasedPermissions   bool `json:"role_based_permissions"`
+	UseIMS                 bool                    `json:"use_ims"`
+	RoleBasedPermissions   bool                    `json:"role_based_permissions"`
 	DefaultCertificateInfo *DefaultCertificateInfo `json:"default_certificate_info"`
 }
 
--- a/traffic_ops/traffic_ops_golang/deliveryservice/sslkeys.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/sslkeys.go
@@ -23,7 +23,6 @@ import (
 	"context"
 	"database/sql"
 	"errors"
-	"fmt"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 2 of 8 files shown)


traffic_ops/traffic_ops_golang/deliveryservice/sslkeys.go
```diff
@@ -115,3 +118,70 @@ func generatePutRiakKeys(req tc.DeliveryServiceGenSSLKeysReq, tx *sql.Tx, tv tra
...
+		DeliveryServiceSSLKeysReq: tc.DeliveryServiceSSLKeysReq{
+			CDN:             &cdnNameStr,
+			DeliveryService: ds.XMLID,
+			HostName:        &hostname,
+			Key:             ds.XMLID,
+			Version:         &version,
+			BusinessUnit:    util.StrPtr("Placeholder"),
+			City:            util.StrPtr("Placeholder"),
+			Organization:    util.StrPtr("Placeholder"),
+			Country:         util.StrPtr("United States (US)"),
```
rawlinp, line 165: Does "Placeholder" not work for country and state? If not, disregard :)

traffic_ops/traffic_ops_golang/deliveryservice/sslkeys.go
```diff
@@ -115,3 +118,70 @@ func generatePutRiakKeys(req tc.DeliveryServiceGenSSLKeysReq, tx *sql.Tx, tv tra
...
+	}
+
+	cdnNameStr := string(cdnName)
+
+	if ds.ExampleURLs == nil {
+		ds.ExampleURLs = MakeExampleURLs(ds.Protocol, *ds.Type, *ds.RoutingName, *ds.MatchList, cdnDomain)
+	}
+
+	hostname := strings.Split(ds.ExampleURLs[0], "://")[1]
+	if strings.Contains(ds.Type.String(), "HTTP") {
```
rawlinp, line 149: I think we may be able to change this to `if ds.Type.IsHTTP()`

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/deliveryservice/sslkeys.go
```diff
@@ -87,9 +90,9 @@ func GenerateSSLKeys(w http.ResponseWriter, r *http.Request) {
 	api.WriteResp(w, r, "Successfully created ssl keys for "+*req.DeliveryService)
 }
 
-// generatePutRiakKeys generates a certificate, csr, and key from the given request, and insert it into the Riak key database.
+// GeneratePutRiakKeys generates a certificate, csr, and key from the given request, and insert it into the Riak key database.
 // The req MUST be validated, ensuring required fields exist.
-func generatePutRiakKeys(req tc.DeliveryServiceGenSSLKeysReq, tx *sql.Tx, tv trafficvault.TrafficVault, ctx context.Context) error {
+func GeneratePutRiakKeys(req tc.DeliveryServiceGenSSLKeysReq, tx *sql.Tx, tv trafficvault.TrafficVault, ctx context.Context) error {
```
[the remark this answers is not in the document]
mattjackson220: my bad. leftover from when i had the `GeneratePlaceHolderSelfSignedCert` in a different file. fixed

traffic_ops/traffic_ops_golang/deliveryservice/sslkeys.go
```diff
@@ -115,3 +118,70 @@ func generatePutRiakKeys(req tc.DeliveryServiceGenSSLKeysReq, tx *sql.Tx, tv tra
...
+// GeneratePlaceholderSelfSignedCert generates a self-signed SSL certificate as a placeholder when a new HTTPS
+// delivery service is created or an HTTP delivery service is updated to use HTTPS.
+func GeneratePlaceholderSelfSignedCert(ds tc.DeliveryServiceV40, inf *api.APIInfo, context context.Context) (error, int) {
+	version := util.JSONIntStr(1)
+
+	db, err := api.GetDB(context)
+	if err != nil {
+		return err, http.StatusInternalServerError
+	}
+	tx, err := db.Begin()
```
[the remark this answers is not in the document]
mattjackson220: i think it will after the switch to using `createV40` and `updateV40` so ill get rid of this

traffic_ops/traffic_ops_golang/deliveryservice/sslkeys.go
mattjackson220: haha i think it will work but it felt weird. updated now

rawlinp (commented):
Code looks good, and manual testing confirms the certs are created as expected. Should we add TO API tests to check for the cert creation? It would probably need to be wrapped in `if includeSystemTests`.

Also, I can't tell if the TO API tests and TP tests are failing due to this PR. I've rerun them a few times, but they seem to be consistent.

smalenfant:
Would there be a way to configure these so the CSR is valid to use after being auto-generated? That would mean new configuration.

```
			BusinessUnit:    util.StrPtr("Placeholder"),
			City:            util.StrPtr("Placeholder"),
			Organization:    util.StrPtr("Placeholder"),
			Country:         util.StrPtr("Placeholder"),
			State:           util.StrPtr("Placeholder"),
```

rawlinp:
I like that idea. Just make the new config optional and use `placeholder` by default 👍

jhg03a:
Closes #2429

traffic_ops/traffic_ops_golang/config/config.go
```diff
@@ -173,6 +174,14 @@ type ConfigAcmeAccount struct {
 	HmacEncoded  string `json:"hmac_encoded"`
 }
 
+type DefaultCertificateInfo struct {
```
rawlinp, line 177: Can we add some config validation to ensure that if the config contains the `default_certificate_info` section, that all the fields within it are non-empty? I'm not sure if that would cause the certificate generation to fail or not, but I'd think if there were any empty fields that it would be unintentional.

traffic_ops/traffic_ops_golang/deliveryservice/sslkeys.go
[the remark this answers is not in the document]
rawlinp: So I think this went back and forth, but are we not able to use `inf.Tx.Tx` here? It seems like maybe there was a missing nil pointer check that ended up fixing the tests?

rawlinp (changes requested):

rawlinp (approved):
Code looks good, and TO generated a cert on DS create (with https enabled) and update (from http to https) as expected.

PR #6028, opened 2021-07-15
Remove types used only by removed APIv1 endpoints/client

ocket8888:

This PR removes a type alias (`RiakPingResp`) from `lib/go-tc` that was only used in endpoints now removed from Traffic Ops. Internal references lingering in Riak-specific Traffic Vault code in Traffic Ops internals was changed to use `TrafficVaultPing`as directed by the deprecation notice on the removed alias.

As of 2a0faa5a1085e01d459e9a9568daf2bb7bad36d8 this also removes "Cache Group Fallback" structures:
- `CacheGroupFallback`
- `CacheGroupFallbacksResponse`
- `CacheGroupFallbackNullable`
- `CacheGroupFallbackNullableResponse`

As those were only used by the API route `/cachegroup_fallbacks` which was removed with APIv1.

As of 127f714ba3f29b10658d2f9a3117c2ff070437a8 this also removes some Delivery Service-related structures that are unused:
- `GetDeliveryServiceResponse` - last references to this were removed in the TO Go client restructuring that happened between 4.1.1 and 5.0.0
- `DeliveryServicesResponse`, `CreateDeliveryServiceResponse`, and `UpdateDeliveryServiceResponse` - these were all only used by the now-removed APIv1 client
- `DeliveryServiceStateResponse`, `DeliveryServiceFailover`, `DeliveryServiceDestination`, and `DeliveryServiceState` - types used to define the structure of the response of the removed `/deliveryservices/{{ID}}/state` APIv1 endpoint
- `DeliveryServiceMatchesResp` - To my knowledge this was never used by anything - at least not in any official ATC release
- `DeliveryServicePatterns` and `DeliveryServiceMatchesResponse` - types used to define the structure of the response of the removed `/deliveryservice_matches` APIv1 endpoint.
- `DeliveryServiceRoutingResponse` and `DeliveryServiceRouting` - types used to define the structure of the response of the removed `/deliveryservices/{{ID}}/routing` APIv1 endpoint.
- `UserDeliveryServicesNullableResponse` - type used to define the structure of the response of the removed `/users/{{ID}/deliveryservices` endpoint.

There are more structures in the lib that refer to the now-removed `/users/{{ID}}/deliveryservices` and/or `/deliveryservice_user` API endpoints, but we can't remove those because they are linked to by the v2 and v3 clients, even though those methods can't possibly work (without a plugin, I suppose) because that could cause a compile-time breakage in users of the client.

As of 751dd276c0259f7a58f3ca9e3bd5eca33c2f691e this PR also removes logic for handling the now-removed `/cachegroups/{{ID}}/unassigned_parameters` endpoint that was only in APIv1. CGP in general needs to stick around until we get rid of APIv3, but this is now unused.

- Traffic Ops

# What is the best way to verify this PR?
Make sure all the tests still pass.

- [x] Tests are unnecessary


zrhoffman (approved):
Removed structs and functions are not referenced anywhere else in the repo.

Opened #6036 for all of the API endpoints that we need to remove from the `api_capability` table now that API v1 is removed.

PR #6030, opened 2021-07-15
Remove file containing only logic for APIv1

ocket8888:

This PR removes logic for handling the now-removed `/cachegroups/{{ID}}/unassigned_parameters` endpoint that was only in APIv1. CGP in general needs to stick around until we get rid of APIv3, but this is now unused.

- Traffic Ops

Make sure all the tests still pass.

- [x] Tests are unnecessary


zrhoffman:
#6030 seems related enough to #6028 for them to be combined. Any reason not to combine them?

ocket8888:
Supposedly the difference is that this is removing things from Traffic Ops while #6028 is removing things from `lib/go-tc`, but it is true that #6028 still does touch Traffic Ops to change some type names from `RiakPing` to `TrafficVaultPing`. They can probably be combined if you want - but then you need to promise to review the result!

zrhoffman:
Deal

ocket8888:
Closed in favor of #6028

PR #6031, opened 2021-07-15
TP convert ag-grid tables to component

shamrickus:

It converts several ag-grid tables to the `common-grid-component`

- Traffic Portal

Ensure the new pages still work, subtle functionality might be different since different versions of the ag-grid code exist.
The following pages have been changed:
- Jobs (Invalidation Requests)
  - DS Jobs
- Servers
  - Status >Servers
  - Type >Servers
  - CDN >Servers
  - Cache Group >Servers
  - Phys Loc >Servers
  - Topology >Servers
  - DS >Servers
- Notifications (on a CDN)
- Change logs

## Additional Information
The only remaining unconverted tables are deliveryservice and it's derivatives.


traffic_portal/app/src/common/modules/table/agGrid/CommonGridController.js
```diff
@@ -217,6 +230,9 @@ let CommonGridController = function ($scope, $document, $state, userModel, dateU
                 $scope.$apply();
             },
             onRowClicked: function(params) {
+                if (params.event.target.classList.contains('link')) {
+                    return;
+                }
```
ocket8888, line 235: shouldn't this just check `params.event.target instanceof HTMLAnchorElement` ? Classes can be improperly applied or overloaded for styling reasons, seems more reliable to just directly check if the target is actually a link.

traffic_portal/app/src/common/modules/table/agGrid/grid.tpl.html
```diff
@@ -18,14 +18,22 @@
...
-    <ol class="breadcrumb pull-left">
-        <li class="active" ng-if="$ctrl.title">
-            {{ $ctrl.title }}
-            <button ng-if="$ctrl.titleButton !== undefined" type="button" class="btn btn-link"
-                    ng-click="$ctrl.titleButton.onClick()">{{ $ctrl.titleButton.getText() }}</button>
-        </li>
+    <div class="pull-left">
+        <ol class="breadcrumb pull-left" ng-if="$ctrl.breadCrumbs !== undefined">
+            <li ng-repeat="bc in $ctrl.breadCrumbs track by $index" ng-class="{'active': ($index+1 === $ctrl.breadCrumbs.length)}">
+                <a ng-if="$ctrl.bcHasHref(bc)" href="{{ $ctrl.bcGetHref(bc) }}">{{ $ctrl.bcGetText(bc) }}</a>
```
ocket8888, line 24: interpolated content in link targets should use `ng-href` to guarantee binding

traffic_portal/app/src/common/modules/table/cdnNotifications/TableCDNNotificationsController.js
```diff
@@ -16,12 +16,42 @@
...
+	    	return "#!/cdns/" + $scope.cdn.id;
+		}
+	},
+	{
+		text: "Notifications"
+	}];
+
+	this.$onInit = function() {
+		let i;
+		for(i = 0; i < $scope.dropDownOptions.length; ++i) {
```
ocket8888, line 45: does webpack not let you do `for (let i=0; i < thing; ++i;){...}`? Or better yet, `for (const ddo of $scope.dropDownOptions) {...}`?

ocket8888 (commented):
Haven't finished looking through this, but about to be done for the day

traffic_portal/app/src/common/modules/table/agGrid/CommonGridController.js
shamrickus: Probably, I just copied it from the servers page

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -262,204 +220,171 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
...
+			    $scope.queueServerUpdates(entry, null);
+			},
+			text: "Queue Server Updates",
+		},
+		{
+			type: 1,
+			isDisabled: function(entry){
+				return !$scope.isCache(entry) || !entry.updPending;
+			},
+            onClick: function (entry){
```
ocket8888, line 313: inconsistent indentation here

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -262,204 +220,171 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
...
-				Object.assign(filterState, filter);
-				$scope.gridOptions.api.setFilterModel(filterState);
-			} catch (e) {
-				console.error("Failure to load stored filter state:", e);
+		{
+			type: 1,
+			onClick: function (entry) {
+				$scope.confirmDelete(entry, null);
+			},
+            text: "Delete"
```
ocket8888, line 286: inconsistent indentation here

traffic_portal/test/integration/PageObjects/ServersPage.po.ts
```diff
@@ -141,8 +141,7 @@ export class ServersPage extends BasePage {
     let name = nameServer+this.randomize;
     await this.txtQuickSearch.clear();
     await this.txtQuickSearch.sendKeys(name);
-    await browser.actions().mouseMove(element(by.xpath("//span[text()='"+name+"']"))).perform();
-    await browser.actions().doubleClick(element(by.xpath("//span[text()='"+name+"']"))).perform();
+    await browser.actions().click(element(by.xpath("//span[text()='"+name+"']"))).perform();
```
ocket8888, line 144: We shouldn't be using XPath, CSS is much more familiar and easier to understand. The equivalent selector in this case would be `by.cssContainingText("span", name)`, which is much simpler.

traffic_portal/app/src/common/modules/table/jobs/TableJobsController.js
```diff
@@ -78,25 +66,30 @@ var TableJobsController = function(tableName, jobs, $document, $scope, $state, $
...
+			$scope.createJob();
+		},
+		text: "Create Invalidation Request",
+		type: 1
+	}];
+
+	/** @type CGC.ContextMenuOption[] */
+	$scope.contextMenuOptions = [{
+		onClick: function (job, $event) {
+			$scope.confirmRemoveJob(jobs,  $event);
```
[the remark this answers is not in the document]
shamrickus: You are correct.

traffic_portal/app/src/common/modules/table/changeLogs/TableChangeLogsController.js
```diff
@@ -16,43 +16,33 @@
...
-			valueFormatter: dateCellFormatterRelative
+			relative: true,
 		},
 		{
 			headerName: "Created (UTC)",
 			field: "lastUpdated",
 			hide: false,
 			filter: "agDateColumnFilter",
-			tooltipValueGetter: dateCellFormatter,
-			valueFormatter: dateCellFormatter
```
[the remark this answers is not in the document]
ocket8888: Where is it doing that? ~~I only see moment being called for the relative~~ nvm, it looks like any column definition that uses the agDateColumnFilter is getting its tooltipValueGetter and valueFormatter set to the same UTC thing as before. It's not using moment, but it's not as much of a problem as I thought. Although, the UTC cell formatter is also outputting timezone information, so "UTC" is still unnecessary. But not possibly incorrect.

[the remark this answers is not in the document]
shamrickus: I was looking at a thing specifically for jobs. The functionality is the same as it was before, so if it was working before it should be fine now.

traffic_portal/app/src/common/modules/table/deliveryServiceJobs/TableDeliveryServiceJobsController.js
```diff
@@ -35,7 +35,7 @@ var TableDeliveryServiceJobsController = function(deliveryService, jobs, $contro
 			return $scope.deliveryService.xmlId;
 		},
 		getHref: function () {
-			return "#!/delivery-services/" + $scope.deliveryService.id + "?type=" + $scope.deliveryService.type;
+			return encodeURI("#!/delivery-services/" + $scope.deliveryService.id + "?type=" + $scope.deliveryService.type);
```
ocket8888: That won't work, `encodeURI` expects a full URI as input, and because we are putting query string parameters in x-www-form-urlencoded format inside a document fragment, it assumes nothing needs to be escaped and leaves the input unchanged:
image

Using `encodeURIComponent` expects the entirety of its input to be a single component of a URI, so it will encode too much:

image


You need to use `encodeURIComponent` only on the part that needs to be encoded as a single URI component, like:

```javascript
"#!/delivery-services/" + $scope.deliveryService.id + "?type=" + encodeURIComponent($scope.deliveryService.type);
```

shamrickus: Already changed

ocket8888 (changes requested):
Breadcrumb links appear broken.

1. Go to Topologies -> Cache Groups
2. Click on a Cache Group
3. Click on More -> View Servers
4. Click on "Cache Groups" in the breadcrumbs
5. It tries to take you to `https://localhost/$ctrl.bcGetHref(bc)` which can't load, the page shows a single

The entire page content is just
```html
<!DOCTYPE html>
<html lang="en"><head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot GET /$ctrl.bcGetHref(bc)</pre>
</body>
</html>
```

shamrickus:
All bread crumbs that had a custom href were broken, fixed now.

ocket8888 (approved):

PR #6038, opened 2021-07-19
Remove structs/funcs only used by the removed TO API version 1

ocket8888:

This PR removes a struct that was only used to represent the response to the now-removed `/profiles/trimmed` Traffic Ops APIv1 route.

As of ea4105d49be0f3de008466c34c2277e6fcb6b931 this also removes the `ServerV1` and `ServerV1Response` structures and their methods, which were only used by the APIv1 Traffic Ops Go client.

As of 9b350757a360d232795f52b2c91e1467e908c90b this also removes `LastUpdated` which was a struct that was only used in the now-removed APIv1.3-specific Go client (last reference disappeared between 4.1.1 and 5.0.0).

As of 25534c2c17de006ed338946b1b1f5833bab88a97 this also removes two constants (`TRAFFIC_STATS_VERSION` and `TRAFFIC_STATS_SOURCE`) as they were only used in the APIv1 handler for `/deliveryservice_stats` where they already served no purpose.

As of 06a39697119b8a73f913daf8b09cc28e137aafcf this also removes `UserV13` and `UsersResponseV13`, which are named as though they were used in API version 1.3 to represent its response representations of Users, but in actuality I failed to find any use of these types in the repository's history.

As of 44e70471053c5f1339ee7c31ea12cacf0f8c795d this also removes a legacy handler for Snapshots that was used in a Perl-compatibility un-versioned API route that has now been removed with API v1.

As of f7594c63473b5ae1775867695358ac667e581b8c this also removes a legacy handler for fetching monitoring configuration that had no known purpose.

As of d4211db00ac2b0c3f73c8b625e1c245ed9974936 this also removes a handler for the now-removed API v1 route `regions/name/{{name}}`, which was never even documented afaik.

As of 605712e60914caa1c6015bada18ce592ed7ad1ef this also removes `notImplementedHandler` which was used for at least one API route in Go that was not rewritten from Perl for APIv1 - which has now been removed.

As of 9171b8e5720c932ce15a1408b9f526970b8a8ee4 this also removes the `getCDNIDByDomainname` helper used internally by Traffic Ops - only in code which has already been removed.

as of 7687cb4cdc1374f621d0418e5ed301932707b052 this also removes definitions for types relating to the now-removed `/hwinfo` API endpoint (although apparently they weren't used even then)

None, the struct was unused.

Make sure all existing tests still pass.

- [x] Tests are unnecessary


rawlinp:
While you're at it, my IDE found that these unused functions which may or may not be remnants of v1:
```
handler.go: SnapshotOldGetHandler
monitoring.go: Get
name.go: GetName
routes.go: notImplementedHandler
keys.go: getCDNIDByDomainName
```

rawlinp (approved):
LGTM

PR #6044, opened 2021-07-20
Remove unmaintained, unused, legacy ATC website

ocket8888:

This PR removes an old version of the ATC website which is no longer used. The actual source for the website is now managed by committers at https://github.com/apache/trafficcontrol-website instead.

None.

Shouldn't really be much to do. This affects literally nothing. Files were only deleted, so ~~there should be no licensing concerns~~, nothing was documented or under test, and nothing needs to be.

Spoke too soon. Make sure Weasel passes.

- [x] Tests are unnecessary


zrhoffman (approved):
Confirmed that unmaintained, unused, legacy ATC website is removed.

PR #6046, opened 2021-07-20
Improve the DNSSEC refresh API

### LINKED ISSUE #3054: Improve the DNSSEC keys refresh API
The current `/api/1.4/cdns/dnsseckeys/refresh` endpoint should be modified/enhanced:
- Return a `202 Accepted` to indicate that the request was received successfully but further processing is occurring in the background.
- The `202` response should also include a link to a separate API endpoint that shows the current status of the `refresh` job that was kicked off (this means also implementing that separate API endpoint)
- The `refresh` endpoint should also be changed to a `PUT` because mutation should not be done in a `GET`.

rawlinp:
- [x] This PR fixes #3054

- Documentation
- Traffic Control Client (Go)
- Traffic Ops

Ensure the added TO API test passes.
Build and view the DNSSEC refresh v4 API doc, make sure it renders properly without warnings.
`PUT /api/4.0/cdns/dnsseckeys/refresh` and verify that a `Location` response header is returned, and use the value to `GET /async_status/{some_id}` and verify the status message.

- [x] This PR includes tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
+++ b/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
@@ -23,6 +23,7 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"fmt"
 	"net/http"
 	"strconv"
 	"sync/atomic"
@@ -103,7 +104,7 @@ func refreshDNSSECKeys(ctx context.Context) (int, bool, int, error, error) {
 		asyncTx, err := db.Begin()
 		if err != nil {
 			unsetInDNSSECKeyRefresh()
-			return 0, false, http.StatusInternalServerError, nil, errors.New("RefreshDNSSECKeys beginning tx: " + err.Error())
+			return 0, false, http.StatusInternalServerError, nil, errors.New("RefreshDNSSECKeys beginning asyncTx: " + err.Error())
 		}
 		jobID, status, usrErr, sysErr := api.InsertAsyncStatus(asyncTx, "DNSSEC refresh started")
 		if usrErr != nil || sysErr != nil {
@@ -137,8 +138,6 @@ func doDNSSECKeyRefresh(tx *sql.Tx, asyncDB *sqlx.DB, tv trafficvault.TrafficVau
 		}
 	}()
 	defer unsetInDNSSECKeyRefresh()
-
-	updatedAny := false
 
 	cdnDNSSECKeyParams, err := getDNSSECKeyRefreshParams(tx)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
```diff
@@ -274,9 +327,20 @@ func doDNSSECKeyRefresh(tx *sql.Tx, tv trafficvault.TrafficVault) {
 		if updatedAny {
 			if err := tv.PutDNSSECKeys(string(cdnInf.CDNName), keys, tx, context.Background()); err != nil {
 				log.Errorln("refreshing DNSSEC Keys: putting keys into Traffic Vault for cdn '" + string(cdnInf.CDNName) + "': " + err.Error())
+				anyErrors = true
 			}
 		}
 	}
+	api.CreateChangeLogRawTx(api.ApiChange, "Refreshed DNSSEC keys", user, tx)
```
srijeet0406, line 334: Noticing a couple of things here:

1.) If the `anyErrors` is true, then we should probably not create a changelog with the success message right away
2.) If some of the key refresh operations failed, maybe we could add some kind of a counter (or maybe DS names/ CDN names) that is transmitted back to the user (using changelog and the alerts in the response) that tells us which keys failed to refresh.

traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
```diff
@@ -33,51 +32,90 @@ import (
...
+		cfg, err := api.GetConfig(ctx)
 		if err != nil {
-			api.HandleErr(w, r, noTx, http.StatusInternalServerError, nil, errors.New("RefresHDNSSECKeys getting config from context: "+err.Error()))
 			unsetInDNSSECKeyRefresh()
-			return
+			return 0, false, http.StatusInternalServerError, nil, errors.New("RefreshDNSSECKeys getting config from context: " + err.Error())
+		}
+		user, err := auth.GetCurrentUser(ctx)
+		if err != nil {
+			unsetInDNSSECKeyRefresh()
```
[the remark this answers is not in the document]
rawlinp: I think it helps to not hide the call to `unsetInDNSSECKeyRefresh()` by putting these two lines in their own function. This makes it clear that it needs to be called if this function ever encounters an error and needs to return early before calling `doDNSSECKeyRefresh`.

traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
rawlinp: I'll go with counters since failures to refresh should be rare and would require digging into the logs which will contain the failed DS or CDN names.

srijeet0406 (approved):
Code changes look good.
Unit/ API tests pass.
Manually hit the `PUT api/4.0/cdns/dnsseckeys/refresh` endpoint and got a `202` status code, with a link to `GET` the status from.
LGTM!

PR #6049, opened 2021-07-21
Add t3c arg to not set outgoing ip

rob05c:
Adds a t3c flag to not add a records.config outgoing_ip entry of the Server's IP addresses.

Setting the outgoing_ip is almost always the correct and desired behavior. But this can be necessary if the Server's IP(s) are NAT or otherwise unusual and can't reach parents.

Includes tests.
Includes docs.
Includes changelog.

- Traffic Ops ORT

Run tests. Run `t3c apply --no-outgoing-ip=true` and verify the generated records.config does not have an outgoing_ip entry.

Not a bug fix.

## Additional Information


traeak (approved):
Changes are simple.  Ran test with and without options args.

PR #6052, opened 2021-07-22
update print statements in t3c to use go logger library

jpappa200:

This PR updates print statements to use go logger library for the maple forwarder. 

- Traffic Ops ORT

Build, install on a cache, queue updates, and run t3c apply

Not a bug fix.


cache-config/t3c-generate/plugin/plugin.go
```diff
@@ -59,7 +58,7 @@ type Plugins interface {
 func AddPlugin(priority uint64, funcs Funcs) {
 	_, filename, _, ok := runtime.Caller(1)
 	if !ok {
-		fmt.Println(time.Now().Format(time.RFC3339Nano) + " Error plugin.AddPlugin: runtime.Caller failed, can't get plugin names") // print, because this is called in init, loggers don't exist yet
+		log.Errorln(time.Now().Format(time.RFC3339Nano) + " Error plugin.AddPlugin: runtime.Caller failed, can't get plugin names")
```
rob05c, line 61: This needs to still be a Print, because like the comment said, the loggers haven't been initialized yet

cache-config/t3c-apply/config/config.go
```diff
@@ -329,25 +328,25 @@ func GetCfg() (Cfg, error) {
...
-				fmt.Printf("no override for TSHome was found, using the configured default: '%s'\n", TSHome)
+				log.Infof("no override for TSHome was found, using the configured default: '%s'\n", TSHome)
 			}
 		}
 	}
 	if tsHome != "" {
 		TSHome = tsHome
 		TSConfigDir = tsHome + "/etc/trafficserver"
-		fmt.Printf("TSHome: %s, TSConfigDir: %s\n", TSHome, TSConfigDir)
+		log.Infof("TSHome: %s, TSConfigDir: %s\n", TSHome, TSConfigDir)
```
rob05c, line 349: All of these logs happen before the `log.InitCfg(cfg)` call below. Until that's called, the loggers don't exist. So these logs will never be written.

The easiest solution here is probably to put the strings in an array, i.e.
```
toInfoLog := []string{}
```
And then e.g.
```
	if *tsHomePtr != "" {
		tsHome = *tsHomePtr
		toInfoLog = append(toInfoLog, fmt.Sprintf("set TSHome from command line: '%s'", TSHome)
	}
```

And then, after the loggers are initialized:

```
	if err = log.InitCfg(cfg); err != nil {
		return Cfg{}, errors.New("Initializing loggers: " + err.Error() + "\n")
	}

	for msg := range toInfoLog {
		log.Infoln(msg)
	}
```

cache-config/t3c-apply/config/config.go
```diff
@@ -329,25 +328,25 @@ func GetCfg() (Cfg, error) {
 	var tsHome = ""
 	if *tsHomePtr != "" {
 		tsHome = *tsHomePtr
-		fmt.Printf("set TSHome from command line: '%s'\n\n", TSHome)
+		log.Infof("set TSHome from command line: '%s'\n\n", TSHome)
```
rob05c, line 331: Everything with multiple newlines needs changed to only have 1. E.g. this needs changed to 

`log.Infof("set TSHome from command line: '%s'\n", TSHome)`

Multiple newlines will result in log lines without the required prefix, which breaks people's log parsers.

I made a PR to make that impossible by always prefixing every newline in the log function. But people didn't like that.

cache-config/t3c-apply/t3c-apply.go
```diff
@@ -61,7 +60,7 @@ func main() {
 	var lock util.FileLock
 	cfg, err := config.GetCfg()
 	if err != nil {
-		fmt.Println(err)
+		log.Errorln(err)
```
rob05c, line 63: This error happened because the config was malformed, so the loggers won't exist.
I think it needs to be a `fmt.Println`, there's just no way around it here. 

And we need the error message, otherwise it's going to be impossible to figure out why it's failing.

cache-config/t3c-check/t3c-check.go
```diff
@@ -43,18 +43,18 @@ func main() {
 	flagHelp := getopt.BoolLong("help", 'h', "Print usage information and exit")
 	getopt.Parse()
 	if *flagHelp {
-		fmt.Println(usageStr())
+		log.Errorln(usageStr())
```
rob05c, line 46: `t3c-check` hasn't initialized the loggers, so these log lines will never be printed. 

Most apps use an interface, but that's probably overkill here. You can just initialize them all to stderr here:

```
log.Init(os.Stderr, os.Stderr, os.Stderr, os.Stderr, os.Stderr)
```

cache-config/t3c-check/t3c-check.go
```diff
@@ -43,18 +43,18 @@ func main() {
...
 		os.Exit(ExitCodeSuccess)
 	}
 
 	if len(os.Args) < 2 {
-		fmt.Fprintf(os.Stderr, "no command\n\n"+usageStr())
+		log.Errorf("no command\n\n" + usageStr())
```
rob05c, line 51: Needs changed to a single newline.

cache-config/t3c-diff/t3c-diff.go
```diff
@@ -36,31 +36,31 @@ func main() {
 	help := getopt.BoolLong("help", 'h', "Print usage info and exit")
 	getopt.ParseV2()
 	if *help {
-		fmt.Println(usageStr)
+		log.Errorln(usageStr)
```
rob05c, line 39: `t3c-diff` hasn't initialized the loggers, needs a `log.Init(os.Stderr, os.Stderr, os.Stderr, os.Stderr, os.Stderr)`

cache-config/t3c-generate/config/config.go
```diff
@@ -84,7 +83,7 @@ func GetCfg() (Cfg, error) {
 	getopt.Parse()
 
 	if *version {
-		fmt.Println(AppName + " v" + Version)
+		log.Errorln(AppName + " v" + Version)
```
rob05c, line 86: `t3c-generate` hasn't initialized the loggers yet here.
You could either call `log.Init(os.Stderr, os.Stderr, os.Stderr, os.Stderr, os.Stderr)` immediately before this `log.Errorln` (since it exits anyway)

Or, this one is probably ok to leave as a print. The goal of the loggers here is to make it possible for people to have log parsers, and not have them ever break. Since this message results from a call to `--version`, it should never be part of a production system that needs the result to run thru a log parser.

cache-config/t3c-generate/t3c-generate.go
```diff
@@ -38,12 +37,12 @@ import (
 func main() {
 	cfg, err := config.GetCfg()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Getting config: "+err.Error()+"\n")
+		log.Errorf("Getting config: %s\n", err.Error())
```
rob05c, line 40: This is a result of failing to initialize the config, so the loggers won't exist and this message will never be displayed.

So it needs to `fmt.Print`, just no way around it.

cache-config/t3c-generate/update-to-client/update-to-client.go
```diff
@@ -32,26 +33,26 @@ import (
 
 func main() {
 	if len(os.Args) < 3 {
-		fmt.Fprintf(os.Stdout, usageStr())
+		log.Errorln(usageStr())
```
rob05c, line 36: This script hasn't initialized the loggers. You can either add a call to `log.Init(os.Stderr, os.Stderr, os.Stderr, os.Stderr, os.Stderr)`, or this one's probably fine not using the loggers, either way.

This is a one-off development script, it never makes sense to run in a production system. It would be very unusual for anyone to want or need to run this thru log parsers.

cache-config/t3c-request/config/config.go
```diff
@@ -83,7 +82,7 @@ func InitConfig() (Cfg, error) {
 		Usage()
 	}
 	if *versionPtr == true {
-		fmt.Println(AppName + " v" + Version)
+		log.Errorln(AppName + " v" + Version)
```
rob05c, line 85: Like with the other app, the loggers haven't been initialized here yet. You can either call `log.Init(os.Stderr, os.Stderr, os.Stderr, os.Stderr, os.Stderr)`, or just print since it's highly unlikely`--version` calls will ever need run thru a log parser.

cache-config/t3c-update/config/config.go
```diff
@@ -86,7 +85,7 @@ func InitConfig() (Cfg, error) {
 		Usage()
 	}
 	if *versionPtr == true {
-		fmt.Println(AppName + " v" + Version)
+		log.Errorln(AppName + " v" + Version)
```
rob05c, line 88: Like with the other app, the loggers haven't been initialized here yet. You can either call log.Init(os.Stderr, os.Stderr, os.Stderr, os.Stderr, os.Stderr), or just print since it's highly unlikely--version calls will ever need run thru a log parser.

cache-config/t3c/t3c.go
```diff
@@ -48,18 +48,18 @@ func main() {
 	flagHelp := getopt.BoolLong("help", 'h', "Print usage information and exit")
 	getopt.Parse()
 	if *flagHelp {
-		fmt.Println(usageStr())
+		log.Errorln(usageStr())
```
rob05c, line 51: Loggers were never initialized, needs a call to `log.Init(os.Stderr, os.Stderr, os.Stderr, os.Stderr, os.Stderr)`

cache-config/t3cutil/t3cutil.go
```diff
@@ -81,7 +81,7 @@ func NewLineFilter(str string) string {
 func ReadFile(f string) []byte {
 	data, err := ioutil.ReadFile(f)
 	if err != nil {
-		fmt.Println("Error reading file ", f)
+		log.Errorln("Error reading file ", f)
```
rob05c, line 84: Uhh...libraries should never log. We can't say whether loggers have been initialized, but if they haven't, someone is going to end up with their app exiting/crashing and it's going to be extremely confusing to figure out why.

This function isn't used anywhere. Please just delete the function, if you don't mind.

rob05c (changes requested):
Looks good, thanks! Definitely something we need, to make sure log parsers don't fail on the output.

Just a few requests, mostly around making sure the loggers are initialized before logging. Let me know if you have any questions

cache-config/t3c-generate/plugin/plugin.go
jpappa200: Changed back to print and also changed other log statement to print. Moved comment to top of func for readability.

cache-config/t3cutil/t3cutil.go
jpappa200: removed unused func

cache-config/t3c-request/config/config.go
jpappa200: Put back to print.

cache-config/t3c-update/config/config.go
jpappa200: Put back to print.

cache-config/t3c-generate/config/config.go
jpappa200: Put back to print.

cache-config/t3c-apply/config/config.go
jpappa200: Changed to single new line.

cache-config/t3c-check/t3c-check.go
jpappa200: changed to single new line

cache-config/t3c-apply/t3c-apply.go
jpappa200: Put back to fmt.Println

cache-config/t3c-generate/update-to-client/update-to-client.go
jpappa200: Added log.Init()

cache-config/t3c/t3c.go
jpappa200: Added log.Init()

cache-config/t3c-diff/t3c-diff.go
jpappa200: Added log.Init()

cache-config/t3c-generate/t3c-generate.go
jpappa200: Changed back to Fprintf

cache-config/t3c-check/t3c-check.go
jpappa200: added log.Init()

cache-config/t3c-apply/config/config.go
jpappa200: Added array for info messages and for loop to print them out once logging is initialized.

rob05c (approved):
LGTM. Ran t3c-apply, log lines are prefixed as-expected.

PR #6071, opened 2021-07-29
Migrate Profiles test from E2E to integrate test suite TP

dpham692:

- [x] This PR fixes  is not related to any Issue 

- CDN in a Box
- Traffic Portal
- CI tests

Make sure all tests are migrate in integration/Data/profiles.ts directory.
https://github.com/apache/trafficcontrol/blob/master/traffic_portal/test/e2e_deprecated/profiles/profiles-spec.js
Go to traffic_portal/test/integration and run
`npm test -- --params.baseUrl https://localhost --capabilities.shardTestFiles true` and make sure test all pass.
Make sure CI tests pass also.

### DIFF
--- a/traffic_portal/test/integration/specs/Profiles.spec.ts
+++ b/traffic_portal/test/integration/specs/Profiles.spec.ts
@@ -46,6 +46,24 @@ profiles.tests.forEach(async profilesData => {
                 await profilesPage.OpenConfigureMenu();
                 await profilesPage.OpenProfilesPage();
             });
+            profilesData.check.forEach(check => {
+                it(check.description, async () => {
+                    expect(await profilesPage.CheckCSV(check.Name)).toBe(true);
+                    await profilesPage.OpenProfilesPage();
+                });
+            });
+            profilesData.toggle.forEach(toggle => {
+                it(toggle.description, async () => {
+                    if(toggle.description.includes('hide')){
+                        expect(await profilesPage.ToggleTableColumn(toggle.Name)).toBe(false);
+                        await profilesPage.OpenProfilesPage();
+                    }else{
+                        expect(await profilesPage.ToggleTableColumn(toggle.Name)).toBe(true);
+                        await profilesPage.OpenProfilesPage();
+                    }
+
+                });
+            })
             profilesData.add.forEach(add => {
                 it(add.description, async () => {
                     expect(await profilesPage.CreateProfile(add)).toBeTruthy();
--- a/traffic_portal/test/integration/Data/profiles.ts
+++ b/traffic_portal/test/integration/Data/profiles.ts
@@ -107,6 +107,22 @@ export const profiles = {
 					password: "pa$$word"
 				}
 			],
+			check: [
+				{
+					description: "check CSV link from Profiles page",
+					Name: "Export as CSV"
+				}
+			],
+			toggle:[
+				{
+					description: "hide first table column",
+					Name: "CDN"
+				},
+				{
+					description: "redisplay first table column",
+					Name: "CDN"
+				}
+			],
 			add: [
 				{
 					description: "create a Profiles",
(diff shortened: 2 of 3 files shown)


traffic_portal/test/integration/specs/Profiles.spec.ts
```diff
@@ -46,6 +46,24 @@ profiles.tests.forEach(async profilesData => {
...
+                });
+            });
+            profilesData.toggle.forEach(toggle => {
+                it(toggle.description, async () => {
+                    if(toggle.description.includes('hide')){
+                        expect(await profilesPage.ToggleTableColumn(toggle.Name)).toBe(false);
+                        await profilesPage.OpenProfilesPage();
+                    }else{
+                        expect(await profilesPage.ToggleTableColumn(toggle.Name)).toBe(true);
+                        await profilesPage.OpenProfilesPage();
```
ocket8888, line 62: why are we re-opening the Profiles page at the end of all of these tests? These don't appear to be doing navigation, but either way a better way to reset the location after every test is with an `afterEach` call.

dpham692: I do the reopening Profiles page after each test because if the tests fail it will go back to the Profiles Page and continue with other tests.

dpham692:
@ocket8888 I added aftereach and afterall like you said. Change it multiple way to fit the test suit because it has to log in and log out few times. I also tested it a couple times and make sure the spec clean.

traffic_portal/test/integration/specs/Profiles.spec.ts
```diff
@@ -33,47 +33,57 @@ describe('Setup API for Profiles', () => {
...
-            });
-            it('can open profiles page', async () => {
-                await profilesPage.OpenConfigureMenu();
-                await profilesPage.OpenProfilesPage();
-            });
-            profilesData.add.forEach(add => {
-                it(add.description, async () => {
-                    expect(await profilesPage.CreateProfile(add)).toBeTruthy();
+    profiles.tests.forEach(async profilesData => {
+        profilesData.logins.forEach(login => {
```
ocket8888, line 37: The restructuring here confuses me - this used to be a separate `describe` for each login for each profile, but now it's all under the setup `describe`? It looks that way, but then the `describe`'s description isn't accurate, it's not just setting up it's actually doing all the Profile testing. That's not a terrible way to do it, but given that these specs are asynchronous, there's no guarantee that any one of them will run before any other (and in general you shouldn't assume that to allow for greater degrees of parallelism); that is to say, you don't know that the setup is run to completion before the other tests run. So it should probably go in a `beforeAll` - I actually think it ought to have been done that way in the first place, but the subsequent suite method was working so that change was unnecessary until the `it`s were sharing a `describe`.

By the same reasoning, the teardown logic in the "cleanup" spec should also probably be in an `afterAll`.

Further, we're re-adding `afterEach` `describe`, but it can actually just be added once in the outer `describe` since all spec blocks in a nested suite are also in the outer suite, and so are affected by `beforeEach` and `afterEach` in that outer block.

dpham692: I will fix the structure back to 3 describes. The **beforeAll** is a good idea, I will add that in.  I have to test this out for the **afterall** to answer you because I read somewhere that you can only have one **afterall** and I already use that for the log out function (log out 3 times for each users). @ocket8888

ocket8888: You can have one `afterAll` per `describe`, I think - but also you can just do multiple things in one `afterAll` if that's what you're talking about.

traffic_portal/test/integration/specs/Profiles.spec.ts
```diff
@@ -29,49 +29,70 @@ const loginPage = new LoginPage();
...
 
-describe('Setup API for Profiles', () => {
-    it('Setup', async () => {
-        await api.UseAPI(profiles.setup);
-    });
-});
+beforeAll(async function () {
+    await api.UseAPI(profiles.setup);
+})
+
```
ocket8888, line 35: So you don't wanna do this, because putting a `beforeAll` in the global scope like that is going to run this before every single `it` in the entire integration testing suite.

dpham692: I see that's why the test took so long. I think I will add a 1 beforeAll in the config before all the test run.

ocket8888: Could be, hard to say.

What would that `beforeAll` in the config module be doing? It shouldn't navigate to the Profiles page, that'll have the same problem.

dpham692:
@ocket8888 I changed it back to 3 describe, fix **afterall** and **aftereach**. For the **beforeall** I removed it out right now and will figure how to do with it in a future.

PR #6076, opened 2021-07-30
Adding locks to federation resolvers

srijeet0406:

- [x] This PR is not related to any Issue. It just adds the locks to the federation resolvers endpoints. 

- Traffic Ops

While another user has a hard lock on a cdn, try to link/ unlink federation resolvers for that CDN, and make sure the operation fails.

- master

### DIFF
--- a/traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
+++ b/traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
@@ -25,7 +25,6 @@ import (
 	"database/sql"
 	"fmt"
 	"net/http"
-	"strconv"
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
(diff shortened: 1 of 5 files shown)


traffic_ops/traffic_ops_golang/federation_resolvers/federation_resolvers.go
```diff
@@ -248,6 +249,24 @@ func Delete(w http.ResponseWriter, r *http.Request) {
 	}
 	defer inf.Close()
 
+	fedID, err := strconv.Atoi(inf.Params["id"])
```
rawlinp, line 252: I think we can do `fedID := inf.IntParams["id"]` here since `id` is required and an int param (per L244). Then we don't have a conversion error to handle.

rawlinp (changes requested):

rawlinp (approved):

PR #6081, opened 2021-08-04
Fix DB indices for phys_locations and topology

### LINKED ISSUE #6066: TODB: Incorrect or missing indexes
## I'm submitting a ...

-  bug report

-  Traffic Ops

## Current behavior:
Table `topology` has no `last_updated` index, but `topology_cachegroup` has two including the one for `topology` which is `CREATE INDEX topology_last_updated_idx ON topology_cachegroup ...`.

Table `phys_location` has a typo in it's index. The name is listed as `pys_location_last_updated_idx` instead of `phys_location_last_updated_idx`

## Expected behavior:
For `topology`, each table should have only one last_updated index.
For `phys_location` the index should be named correctly.

Run `db/admin.go` and inspect the schema.

shamrickus:

- [x] This PR fixes #6066.

- Traffic Ops

Run the `db/admin` tool with migrations. Verify the indices are correct. Verify the rollback works correctly.

 - master
 - 5.1.2

### DIFF
--- a/traffic_ops/app/db/migrations/2021080408053529_fix_indices.down.sql
+++ b/traffic_ops/app/db/migrations/2021080408053529_fix_indices.down.sql
@@ -15,8 +15,8 @@
  * the License.
  */
 
-ALTER INDEX IF EXISTS phys_location_last_updated_idx RENAME TO pys_location_last_updated_idx;
+ALTER INDEX phys_location_last_updated_idx RENAME TO pys_location_last_updated_idx;
 
-DROP INDEX IF EXISTS topology_last_updated_idx;
+DROP INDEX topology_last_updated_idx;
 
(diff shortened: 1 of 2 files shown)


zrhoffman (changes requested):
> `IF EXISTS`/`IF NOT EXISTS`

I don't think migrations should succeed error-free if the pre-migration state is not what is expected. The migration version we keep track of should indicate whether these indices already exist and what their names are.

Because the migration version is increased on success, the migration won't rerun anyway, so we shouldn't need the `IF EXISTS`,. If the migration does not do anything, or if it fails, that is potentially useful information that we don't want to hide.

zrhoffman (approved):
* Verified that indices are fixed after applying `2021080408053529_fix_indices.up.sql`
* Verified that a DB dump after applying `2021080408053529_fix_indices.up.sql` and `2021080408053529_fix_indices.down.sql` is identical to a DB dump after running all migrations on the Apache master branch

PR #6089, opened 2021-08-06
Traffic_vault_migrate strip riak formatting from URL Sig keys

shamrickus:

This PR fixes the issue where when moving keys from `riak`to `psql`, the URL Sig keys will retain the `url_sig_{ds}.config` formatting around the key.

- Traffic Ops

Verify when going from Riak->PSQL (A disk dump would work as well) that the URL Sig keys only contain the delivery service name and not the entire formatted string.

 - master


mattjackson220 (approved):
Code looks good! tested locally in ciab and works as expected!

PR #6091, opened 2021-08-06
Fix cache config for https origins

rob05c:
Fixes cache config for https origins.

There was a bug, that https origins would result in edge remap targets of `https://...` and corresponding edge parent lines of port=443 and corresponding mid remap sources of `https:// ... 443`. 

Except that Edge-Mid communication has always been HTTP, and the Mid parents are :80, and Mids are listening on 80, and Mids don't have the certificates necessary to serve HTTPS.

So that config resulted in Edges making TLS HTTPS requests to the Mid's port 80. Which should fail, but seems to sometimes succeed, possibly due to Keep-Alives from other requests.

This fixes HTTPS Origins to properly use `http://...:80` remaps everywhere. We would like to add internal HTTPS support in the future, but that's a much larger feature. Today, ATC doesn't support that. Moreover, that shouldn't depend on whether the Origin is HTTPS. 

The correct bug fix here is to use http and 80 for internal cache communication, regardless of the origin protocol.

Which this fix does.

Includes tests.
Includes changelog.
No docs, no interface change.

Run tests. Generate config with an HTTPS Origin on a Delivery Service that uses mids, verify config is generated to properly use http and port 80 for all internal edge-mid communication.

Probably all of them.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - [#5981](https://github.com/apache/trafficcontrol/issues/5891) - `/deliveryservices/{{ID}}/safe` returns incorrect response for the requested API version
 - [#5984](https://github.com/apache/trafficcontrol/issues/5894) - `/servers/{{ID}}/deliveryservices` returns incorrect response for the requested API version
 - [#6027](https://github.com/apache/trafficcontrol/issues/6027) - Collapsed DB migrations
+- [#6091](https://github.com/apache/trafficcontrol/issues/6091) - Fixed cache config of internal cache communication for https origins
 - [#6066](https://github.com/apache/trafficcontrol/issues/6066) - Fixed missing/incorrect indices on some tables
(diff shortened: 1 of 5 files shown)


traeak (commented):
This works for the simple delivery service case, although there remains a parent.config entry for https (in addition to http) on the edge.
Sadly this doesn't seem to work for the MSO case.  There doesn't seem to be a way to mark a server as priority https and trying to change the dummy origin to https didn't seem to do anything.

rob05c:
>  Sadly this doesn't seem to work for the MSO case. There doesn't seem to be a way to mark a server as priority https and trying to change the dummy origin to https didn't seem to do anything.

It doesn't add the ability to HTTPS MSO, no. That'll take more work (which is on our short list FWIW).

But it should make MSO correctly use http for edge--mid communication, even if the MSO DS OrgServerFQDN field is https (which I think is a bug regardless of MSO, and I think this fixes for both non-MSO and MSO).

> This works for the simple delivery service case, although there remains a parent.config entry for https (in addition to http) on the edge.

I'm looking into this now. I don't think it breaks anything, but it's definitely not ideal. I'll try to find and fix.

traeak (approved):

PR #6099, opened 2021-08-11
Add/Update/Correct GoDoc comments on /lib symbols

ocket8888:
This PR modifies the packages

- `github.com/apache/trafficcontrol/lib/go-atscfg`
- `github.com/apache/trafficcontrol/lib/go-log`
- `github.com/apache/trafficcontrol/lib/go-rfc`
- `github.com/apache/trafficcontrol/lib/go-util`

This PR ensures that each modified package

- Has a GoDoc comment for the package, and that comment is properly recognized
- Has a GoDoc comment for every exported symbol
- Uses complete sentences with proper capitalization and punctuation in GoDoc comments
- Begins every GoDoc comment with either an optional article followed by the name of the documented symbol or the word "Package" followed by the name of the documented package.
- Has no GoDoc comments of the form `// SymbolName ...` where `...` is literal.

This PR also adds deprecation notices to some structures used only in deprecated API versions (2 and 3).

There should be only one actual code change from this PR - fixing a misspelling of "occurred" ("occured") -, but it does move imports around in affected files if necessary to match the pattern:

```go
package packageName

// License header

import (
    "package names for import"
)
```

and possibly moves constants to logical groupings so that a single GoDoc can be written effectively for the group. In at least one case this involved moving a symbol to a different file (but not a different package).

- Documentation (GoDoc comments)

I used the command
```bash
golangci-lint run -c ./.golangci.yml 2>&1 | grep -E '^go-' | grep -v 'move short variable' | grep -v 'error strings should not be capitalized' | grep -v "don't use generic names such as" | grep -vE 'struct field `\w+` should be' | grep -v 'block ends with a' | grep -v 'errors.New(fmt.Sprintf(...))' | grep -vE 'receiver name \w+ should be consistent' | grep -v 'should not use basic type string as key in context.WithValue' | grep -vE 'var `\w+` should be `\w+`' | grep -v 'returns unexported type' | grep -v 'should omit 2nd value from range' | grep -v 'by other packages, and that stutters;' | grep -v "don't use ALL_CAPS in Go names; use CamelCase" | grep -v 'regexrevalidatedotconfig.go:36:7: const'
```
which should output nothing (but exit with a failure because `grep` found no matches), with the following in `./golangci.yml` (under `lib/`):

<details><summary>Click to expand</summary>

```yaml
run:
  concurrency: 4
  timeout: 5m
  issues-exit-code: 0 # TODO: once all enabled linters are satisfied
  tests: false
  skip-dirs-use-default: true

  output:
    format: colored-line-number
    print-issued-lines: true
    print-linter-name: true

issues:
  max-issues-per-linter: 0
  max-same-issues: 0
  exclude-use-default: false

  include:
    - EXC0002
    - EXC0011

linters:
  disable-all: true
  enable:
    - asciicheck
    - bodyclose
    - depguard
    - dogsled
    - durationcheck
    - gochecknoinits
    - godot
    - golint
    - gomodguard
    - govet
    - importas
    - makezero
    - misspell
    - nakedret
    - noctx
    - nolintlint
    - paralleltest
    - predeclared
    - sqlclosecheck
    - structcheck
    - testpackage
    - thelper
    - tparallel
    - typecheck
    - varcheck

```

</details>

Then to find things like `// Foo ...` I did
```bash
grep -RE '^// \w+ \.\.\.' .
```

... also within `lib/`, which should also output no lines (and exit failure b/c `grep`).

## PR submission checklist
- [ ] This PR has tests


lib/go-atscfg/atsdotrules.go
```diff
@@ -37,6 +51,12 @@ type ATSDotRulesOpts struct {
 	HdrComment string
 }
 
+// MakeATSDotRules constructs a '50-ats.rules' file for the given server with
+// the given parameters and header comment content.
+//
+// WARNING: this will panic if parameters with the Names 'Drive_Prefix',
```
rob05c, line 57: This isn't true. It generates fine if any or all params are missing.

```
func TestMakeATSDotRulesNoFiles(t *testing.T) {
	server := makeGenericServer()
	serverProfile := "myProfile"
	server.Profile = &serverProfile

	hdr := "myHeaderComment"

	serverParams := []tc.Parameter{}

	cfg, err := MakeATSDotRules(server, serverParams, &ATSDotRulesOpts{HdrComment: hdr})
	if err != nil {
		t.Fatal(err)
	}
	txt := cfg.Text

	testComment(t, txt, hdr)

	if count := strings.Count(txt, "\n"); count != 1 { // one line for each drive letter, plus 1 comment
		t.Errorf("expected one line for a comment and no others (with no parameters), actual: '%v' count %v", txt, count)
	}
}
```

```
$ go test -failfast -v -run TestMakeATSDotRulesNoFiles
=== RUN   TestMakeATSDotRulesNoFiles
--- PASS: TestMakeATSDotRulesNoFiles (0.00s)
PASS
ok      github.com/apache/trafficcontrol/lib/go-atscfg  0.487s
```

rob05c: Actually that isn't entirely true. There's a known bug, if you have a Prefix but no Letters it generates `KERNEL=="ra", OWNER="ats"`, but that's a data bug anyway, so it hasn't been a huge priority to fix.

But it never panics, AFAIK. If you can demonstrate a panic, definitely file a bug.

ocket8888: My bad, I thought `(map[string]interface{}{})["anything"]` would panic like `([]interface{}{})[100]` does.

It's fixed, but I've probably made similar mistakes elsewhere, given my lack of familiarity with the package. I would greatly appreciate if you'd look over the `lib/go-atscfg` changes yourself, since you're the best candidate for knowing whether descriptions are accurate.

lib/go-atscfg/chkconfig.go
```diff
@@ -26,17 +26,35 @@ import (
...
 const ChkconfigFileName = `chkconfig`
+
+// ChkconfigParamConfigFile is the ConfigFile value of Parameters that affect
+// the generation of the chkconfig configuration file.
 const ChkconfigParamConfigFile = `chkconfig`
+
+// ContentTypeChkconfig is the MIME content type of the contents of the
+// chkconfig configuration file.
+//
+// TODO: The GoDoc for MakeChkconfig says "This is a JSON object, and should be
```
rob05c, line 39: I believe at one point the old ort.pl read the JSON generated by the config generation and executed Linux chkconfig commands based on it.

We now get that data from Traffic Ops, via the `t3c-request` tool, and generate it there, rather than `t3c-generate` (which generates from `lib/go-atscfg`).

Which is to say, it was sort of correct in both places, because the file was never written as JSON. But it's also kind of fuzzy. And I don't think it's used anymore.

But please don't delete it without verifying. I'm not 100% sure, someone will need to verify.

lib/go-atscfg/chkconfig.go
```diff
@@ -26,17 +26,35 @@ import (
...
 type ChkconfigOpts struct {
 }
 
 // MakeChkconfig returns the 'chkconfig' ATS config file endpoint.
-// This is a JSON object, and should be served with an 'application/json' Content-Type.
+//
+// This is a JSON object, and should be served with an 'application/json'
+// Content-Type.
+//
+// TODO: rename/rework? We systemd now, after all.
```
rob05c, line 57: We plan to. I think we're leaning toward removing it altogether, and recommending people manage that with a system management tool (like Ansible or Puppet) rather than ATC.

lib/go-atscfg/hostingdotconfig.go
```diff
@@ -27,15 +27,51 @@ import (
...
+
+// ParamRAMDrivePrefix is the Name of the Parameter that determines the path
+// prefix for RAM caching block devices.
 const ParamRAMDrivePrefix = "RAM_Drive_Prefix"
 
+// ServerHostingDotConfigMidIncludeInactive controls whether or not to include
+// configurations for inactive Delivery Services in hosting.config for mid-tier
+// cache servers.
+//
+// TODO: remove this? It's only used in one place, where it's used as:
```
rob05c, line 59: I don't think we should remove these. They exist to convey intent, and also to make it easy and obvious how to change whether hosting.config includes inactive DSes, if someone wants or needs to fork ATC to do that: change this const.

As public symbols, they could also be used by code to determine whether inactive DSes were included. For example, ATC might want to change that behavior in the future, and someone's script might want to take a different action if that happens.

lib/go-atscfg/ipallowdotconfig.go
```diff
@@ -29,19 +29,80 @@ import (
...
+// LineCommentIPAllowDotConfig is the string that signifies the start of a line
+// comment in the grammar of an ip_allow.config file.
 const LineCommentIPAllowDotConfig = LineCommentHash
 
+// ParamPurgeAllowIP is the Name of Parameters which can specify an IP address
+// to be allowed ALL methods on the cache server (not just PURGE).
+//
+// TODO: replace instances of "purge_allow_ip" with this currently-unused
+// constant.
+// TODO: unexport? This is only used in this package.
```
rob05c, line 49: This shouldn't be removed, it exists for use outside ATC. Someone might want to write a script to manipulate these parameters, and using the canonical ATC `tc.ParamPurgeAllowIP` is the right way to do that.

lib/go-atscfg/chkconfig.go
ocket8888: I won't be deleting anything in this PR. Except possibly comments.

lib/go-atscfg/ipallowdotconfig.go
ocket8888: > ... using the canonical ATC `tc.ParamPurgeAllowIP` is the right way to do that.

but it's actually `atscfg.ParamPurgeAllowIP` - should the comment instead suggest moving it into `lib/go-tc`? Most constants declaring special Parameter Names and/or Config Files are in `lib/go-atscfg` from what I can see, but there's one in `lib/go-tc`.

I can see a case for asking them to be in either place, but it really may come down to where `lib/go-atscfg` will live if and when t3c moves to its own repository, IMO. I think the special Parameter definitions ought to remain with ATC, and t3c would just be in charge of implementing them. That's probably not something we should decide right now, though. TODOs don't even necessarily need to be TODONE.

zrhoffman (approved):

PR #6100, opened 2021-08-12
Remove more unused TM config options

rawlinp:
Remove the following TM config options which appear to be used but
actually aren't, due to them having a corresponding profile parameter:
- cache_health_polling_interval_ms
- cache_stat_polling_interval_ms
- peer_polling_interval_ms

These are actually unused because the following profile parameters take
precedence as soon as the monitoring.json snapshot is retrieved from TO:
- heartbeat.polling.interval
- health.polling.interval
- peers.polling.interval

Additionally, do some minor cleanup of unused functions, unused
parameters, and comments.

- Traffic Monitor
- Automation (Ansible roles)

Ensure that TM still starts up and performs its cache health polling using the interval in the `heartbeat.polling.interval` profile parameter, its cache stat polling using the interval in the `health.polling.interval` profile parameter, and its peer polling using the interval in the `peers.polling.interval` profile parameter.

## PR submission checklist
- [x] This PR updates the tests to remove the unused configuration, but it doesn't really make sense to test the removal of config options.
- [x] These config options were not documented

### DIFF
--- a/.github/actions/tm-integration-tests/entrypoint.sh
+++ b/.github/actions/tm-integration-tests/entrypoint.sh
@@ -120,11 +120,8 @@ go build
 
 cat > ./traffic_monitor.cfg <<- EOF
   {
-      "cache_health_polling_interval_ms": 6000,
-      "cache_stat_polling_interval_ms": 6000,
       "monitor_config_polling_interval_ms": 15000,
       "http_timeout_ms": 8000,
-      "peer_polling_interval_ms": 5000,
       "peer_optimistic": true,
       "max_events": 200,
       "health_flush_interval_ms": 20,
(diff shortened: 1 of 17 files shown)


ocket8888:
> These are actually unused because the following profile parameters take precedence as soon as the monitoring.json snapshot is retrieved from TO

Are they used in the case that said Parameters don't exist?

rawlinp:
No, if the parameters don't exist, the Monitor Config Manager goroutine never sends the monitor config to the other poller goroutines. It will continue polling for a monitoring snapshot in which those parameters actually exist.

ocket8888:
> No, if the parameters don't exist, the Monitor Config Manager goroutine never sends the monitor config to the other poller goroutines. It will continue polling for a monitoring snapshot in which those parameters actually exist.

It looks like that's true for `health.polling.interval` and `peers.polling.interval` but not `heartbeat.polling.interval`. In that last case, it looks like it logs a warning and defaults to `health.polling.interval`, which I'm realizing as I type this means that the default set by the configuration file would still be unused.

zrhoffman (approved):
LGTM

PR #6101, opened 2021-08-12
Add t3c ds tls versions field

rob05c:
Adds t3c using the new Delivery Service TLS Versions field, if it exists.

Also updates t3c to API 4.0, needed for the new field.

Includes tests.
No docs, no interface change (and the DS TLS Versions field is already documented)
No changelog, no interface change (other than the DS TLS Versions field, which is already documented)

Run tests. Run t3c with a DS with the TLS Versions field, verify the values are set in sni.yaml or ssl_server_name.yaml as expected.

Not a bug fix.

## PR submission checklist
- [x] This PR has tests 


cache-config/t3c/t3c.go
```diff
@@ -64,20 +65,19 @@ func main() {
...
-	dir := filepath.Dir(ex)
-	appDir := filepath.Join(dir, app) // TODO use path, not exact dir of this exe
+
+	args := append([]string{app}, os.Args[2:]...)
 
 	env := os.Environ()
 
-	if err := syscall.Exec(appDir, args, env); err != nil {
-		fmt.Fprintf(os.Stderr, "error executing sub-command: "+err.Error()+"\n")
+	if err := syscall.Exec(appPath, args, env); err != nil {
```
rob05c, line 79: This warrants some explanation: this was a sort-of bug I found while working on this.
So `syscall.Exec` is calling `execve`, which takes the full path of the executable, _not_ just the name. Just the name appears to work on some *NIX but not others (specifically, it seems to work on OS X, but not CentOS).

So, what we were doing, was getting the path of the current exe (t3c) and prepending that to the app name (e.g. t3c-apply).

This worked as long as they were in the same directory. Which they are with the provided RPM. But we shouldn't rely on that, apps calling other apps should always use the PATH. Specifically, if you're developing and have apps in different build dirs, this then broke things like `t3c apply`. Most app execution methods do this for you, but `execve` doesn't. (And we very specifically want execve not anything else, because it fully replaces the running app with the called one, rather that a subshell or sub-process or anything else messy).

TLDR this is the right solution: we lookup the executable's path from the OS, and then call `execve` with that full path, which fully replaces the running `t3c` with the subcommand like `t3c-apply`.

jrushford (approved):
Looks good.  Passes all integration and unit tests.

PR #6105, opened 2021-08-15
Remove t3c dispersion flags

rob05c:
Removes the t3c-apply `--dispersion`, `--login-dispersion`, and `--reval-wait-time` flags, and their corresponding features in `t3c`.

When `ort.pl` was slow, there was advantage to doing random sleeps internally, and revalidating periodically (since multiple scripts can't run at once, since they manipulate the same files). But the `t3c` run is fast enough now, there's no advantage to sleeping internally anymore.

There's still reason to randomly sleep before a run interval, to reduce Traffic Ops load. But that is easier and better to do by composing apps, e.g. `sleep "$((RANDOM % 60))" && t3c ...`.

The move in 6.0 from `t3c ...` to `t3c apply` is already a breaking change, so this in-itself is not a breaking change.

Tests exist to verify existing functionality, and this only removes and doesn't add features.
Includes docs.
Includes changelog.

Run tests. Run `t3c`, verify it works as before sans dispersion feature/flags.

Not a bug fix.

## PR submission checklist
- [x] This PR has tests 


jrushford (approved):
Looks good, passes all testing.

PR #6107, opened 2021-08-16
Use Python shebang in Python Postinstall

zrhoffman:
As @ocket8888 mentions in https://github.com/apache/trafficcontrol/pull/4763#discussion_r577819103 , turning a Python script into a polyglot bash script seems convoluted. This PR moves the logic for finding which Python binary to use to the `postinstall` script and uses a standard `python3` shebang for `_postinstall.py`.

- Traffic Ops

* Make sure
* Run the Python postinstall tests, make sure they pass

## PR submission checklist
- [x] Existing tests are sufficient 
- [x] Old shebang was not mentioned in documentation, no documentation change is necessary

### DIFF
--- a/traffic_ops/install/bin/postinstall
+++ b/traffic_ops/install/bin/postinstall
@@ -26,7 +26,7 @@ for arg in "$@"; do
 done
 
 PATH+=:/usr/libexec/
-python_bin="$(command -v {python,platform-python}{3{8,.8,6,.6,},2{7,.7,},} | head -n1)"
+python_bin="$(command -v {python,platform-python}{3{{.,}{9,8,7,6},},2{7,.7,},} | head -n1)"
 if [[ -z "$python_bin" ]]; then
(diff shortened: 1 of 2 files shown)


traffic_ops/install/bin/postinstall
```diff
@@ -25,10 +25,17 @@ for arg in "$@"; do
 	esac
 done
 
+PATH+=:/usr/libexec/
+python_bin="$(command -v {python,platform-python}{3{8,.8,6,.6,},2{7,.7,},} | head -n1)"
```
ocket8888, line 29: I realize this is recreating the supported versions from the old shebang, but why not include 3.7?

zrhoffman: No reason not to. We could include 3.9, too

zrhoffman: Including Python 3.7 and 3.9 in 3f272e6f2f

ocket8888 (approved):
Tests still pass, proper Python version is properly chosen

Issue #6104, opened 2021-08-13
PUT /api/3.0/federations only respects first item in request payload


mitchell852:
## This Bug Report affects these Traffic Control components:
- Traffic Ops

## Current behavior:
A request to PUT /api/3.0/federations with a request body like:

```
[
  {
    "deliveryService":"demo1",
    "mappings":
    {
      "resolve4": ["1.2.3.4"],
      "resolve6": ["2001:db8:3333:4444:5555:6666:7777:8888"]
    }
  },
  {
    "deliveryService":"demo2",
    "mappings":
    {
      "resolve4": ["8.8.8.8"],
      "resolve6": ["2001:db8:3333:4444:5555:6666:7777:9999"]
    }
  }
]
```

will wipe out ALL "Federation to federation resolver mappings" associated with a user’s Delivery Service(s) and replace them with only the first one in the request payload. So if you have 100 mappings and try to add one, you will end up with 1 (instead of 101)

## Expected behavior:
the "Federation to federation resolver mappings" associated with a user’s Delivery Service(s) should be replaced with ALL entries sent in the request payload. not just the first obviously.

## Steps to reproduce:

```
POST /api/3.0/federations with

[
  {
    "deliveryService":"demo1",
    "mappings":
    {
      "resolve4": ["1.2.3.4"],
      "resolve6": ["2001:db8:3333:4444:5555:6666:7777:8888"]
    }
  },
  {
    "deliveryService":"demo2",
    "mappings":
    {
      "resolve4": ["8.8.8.8"],
      "resolve6": ["2001:db8:3333:4444:5555:6666:7777:9999"]
    }
  }
]
```

to create 2 federation mappings for your user. verify that 2 were created by calling GET /api/3.0/federations/all. Now try the PUT /api/3.0/federations endpoint with the same payload as the POST. verify that only 1 mapping exists by calling GET /api/3.0/federations/all.

Side note: what is the difference between the POST and PUT /api/3.0/federations (besides one apparently has a bug and the other does not)

PR #6109, opened 2021-08-16
Fix federations API when using same resolvers for multiple DSes

rawlinp:
Instead of allowing the query to return no row on conflict, ensure it
always returns a row so that we can associate the resolver ID to the
federation.

Closes: #6104

- Traffic Ops

Review the new TO API test and ensure it passes the GitHub Actions CI.

- 4.x
- 5.x
- master

## PR submission checklist
- [x] This PR has tests 
- [x] Bugfix, no docs needed


zrhoffman (approved):
Verified that federation resolvers are created as expected when using the same resolvers for 2 DSes, and that those resolvers are missing on the second DS before this PR.

PR #6115, opened 2021-08-17
Add `cdn.conf` option for enabling new role based permission

shamrickus:
This PR adds a new config option `role_based_permissions` to `cdn.conf`, which will be used to enable the new role based permissions (#5848).

<hr>

- Documentation
- Traffic Ops
- CDN in a Box
- Automation 

Ensure Ansible roles work, GHA pass, CiaB builds successfully, and the `cdn.conf` documentation is correct. 
SSH into the CiaB TO container and verify that `role_based_permissions` is present in the `cdn.conf` file.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/docs/source/admin/traffic_ops.rst
+++ b/docs/source/admin/traffic_ops.rst
@@ -479,8 +479,8 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 
 :role_based_permissions: Toggle whether or not to use Role Based Permissions.
 
-	.. versionadded:: 7.0
-		The blueprint can be seen :pr:`{{5848}}`
+	.. versionadded:: 6.1
+		The blueprint can be seen :pr:`5848`
 
 
 Example cdn.conf
--- a/docs/source/admin/traffic_ops.rst
+++ b/docs/source/admin/traffic_ops.rst
@@ -477,6 +477,12 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 	.. versionadded:: 5.0
 		This is an optional boolean value to enable the handling of the "If-Modified-Since" HTTP request header. Default: false
 
+:role_based_permissions: Toggle whether or not to use Role Based Permissions.
+
+	.. versionadded:: 7.0
+		The blueprint can be seen :pr:`{{5848}}`
+
+
 Example cdn.conf
 ''''''''''''''''
 .. include:: ../../../traffic_ops/app/conf/cdn.conf
--- a/infrastructure/ansible/roles/traffic_ops/templates/cdn.conf.j2
+++ b/infrastructure/ansible/roles/traffic_ops/templates/cdn.conf.j2
@@ -50,6 +50,7 @@
       "address" : "{{ to_smtp_address }}"
    },
    "use_ims": {{ to_use_ims | bool | lower }},
+   "role_based_permissions": {{{ to_use_rbp | bool | lower }},
    "traffic_ops_golang" : {
       "idle_timeout" : {{ to_go_inactive_timeout }},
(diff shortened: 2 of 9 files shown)


docs/source/admin/traffic_ops.rst
```diff
@@ -477,6 +477,12 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 	.. versionadded:: 5.0
 		This is an optional boolean value to enable the handling of the "If-Modified-Since" HTTP request header. Default: false
 
+:role_based_permissions: Toggle whether or not to use Role Based Permissions.
+
+	.. versionadded:: 7.0
+		The blueprint can be seen :pr:`{{5848}}`
```
ocket8888, line 483: I just wrapped it with `{{` and `}}` to show that it was a value to be replaced. This should just be
```rst
:pr:`5848`
```

docs/source/admin/traffic_ops.rst
```diff
@@ -477,6 +477,12 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 	.. versionadded:: 5.0
 		This is an optional boolean value to enable the handling of the "If-Modified-Since" HTTP request header. Default: false
 
+:role_based_permissions: Toggle whether or not to use Role Based Permissions.
+
+	.. versionadded:: 7.0
```
ocket8888, line 482: I'm fine with this for now, but note it'll need to change if we do a 6.1

ocket8888 (changes requested):

srijeet0406 (commented):
Shouldn't the config option be called `permissions_based_roles` instead of `role_based_permissions`?

shamrickus:
I went with `role_based_permissions` to match what the permission design is based on, [`Role-Based Access Control` ](https://en.wikipedia.org/wiki/Role-based_access_control).  

Semantically, either way makes sense; I guess it comes down to whether you consider `permission` to mean: the permission _column_, or the permission _concept_. I would still vote for `role_based_permissions` but I'm not against `permissions_based_roles`.

srijeet0406:
> I went with `role_based_permissions` to match what the permission design is based on, [`Role-Based Access Control` ](https://en.wikipedia.org/wiki/Role-based_access_control).
> 
> Semantically, either way makes sense; I guess it comes down to whether you consider `permission` to mean: the permission _column_, or the permission _concept_. I would still vote for `role_based_permissions` but I'm not against `permissions_based_roles`.

Yeah in my head, the older roles were based on privilege levels, and now they are based on permissions (read-ds, write-ds, etc.), but what you said makes sense too. I'm fine with what you have, no big deal.

srijeet0406 (commented):
Do you mind adding this field to the `test.conf` file as well pls?

ocket8888 (approved):
My comments were addressed, @srijeet0406 can review the rest and I'll merge when he's satisfied.

srijeet0406 (approved):
Looks good to me. Tests pass. Manual testing showed that the flag was being properly set/ unset according to the value in cdn.conf.

rawlinp (approved):

infrastructure/ansible/roles/traffic_ops/templates/cdn.conf.j2
```diff
@@ -50,6 +50,7 @@
       "address" : "{{ to_smtp_address }}"
    },
    "use_ims": {{ to_use_ims | bool | lower }},
+   "role_based_permissions": {{{ to_use_rbp | bool | lower }},
```
rawlinp, line 53: Upon closer review, I believe there is one too many `{` here. I believe it will either render a single `{`, which is not what we want, or Ansible may fail to process this template.

PR #6117, opened 2021-08-18
Prepare database for Roles and Permissions

ocket8888:
This PR removes a foreign key constraint that mandated that a Permission on a Role belong to a row in the `capability` table - this will allow adding arbitrary Permissions to Roles. It also disallows null descriptions as specified in the blueprint (NULLs coalesced to empty strings), and adds the `admin` Role if it does not already exist.

This also includes a migration to ensure that usernames are not NULL, because I found it alarming that they were allowed to be when that would break so much.

<hr/>

- Traffic Ops

Ensure CiaB tests, Traffic Ops API/Go client integration tests, and database tests all pass.

## PR submission checklist
- [x] This PR has tests

### DIFF
--- a/traffic_ops/app/db/migrations/2021081800000000_roles_and_permissions.down.sql
+++ b/traffic_ops/app/db/migrations/2021081800000000_roles_and_permissions.down.sql
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership.  The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+ALTER TABLE ONLY public.role
+ALTER COLUMN "description"
+DROP NOT NULL;
+
+DELETE FROM public.role_capability
+WHERE cap_name NOT IN (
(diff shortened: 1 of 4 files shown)


traffic_ops/app/db/migrations/2021081800000000_roles_and_permissions.down.sql
```diff
@@ -0,0 +1,31 @@
+/*
```
zrhoffman, line 1: > 2021081800000000 with zeros

Why not use `db/admin` to create the migration?

ocket8888: it does that?

zrhoffman: Yep

```shell
go run ./db create_migration my_migration
```

ocket8888: neat

traffic_ops/app/db/migrations/2021081800000001_no_null_username.up.sql
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+ALTER TABLE ONLY public.tm_user
+ALTER COLUMN username
+SET NOT NULL;
```
srijeet0406, line 20: Should we set the `null` usernames to a blank string like we are for `description`, before we add the constraint?

ocket8888: no, I want it to fail intentionally, since a NULL username is something very broken that needs attention - a blank string username is almost certainly not what you actually want.

Besides that, usernames have to be unique - unless they're null. So if more than one user has a NULL username that coalesce won't work anyway.

srijeet0406 (approved):
All unit and API tests pass.
Manually tried to create a new role without a description and it failed.
Manually tried to create a new user without a username and it failed.
Code looks good.

PR #6118, opened 2021-08-18
Feature/refetch

TaylorCFrey:
This PR is to merge the changes necessary to implement the Traffic Ops related changes for the Refetch Blueprint ([PR 5910](https://github.com/apache/trafficcontrol/pull/5910)).

TLDR; Modified the _/jobs_ endpoints for Traffic Ops to take an InvalidationType string that must be either 'REFRESH' or 'REFETCH'. There was some extra work around this modification, but that was its primary purpose.

The changes primarily fall on schema changes for the database and Traffic Ops API changes. Currently this is targeting Traffic Ops API 4.0. 

TO Database:
- Remove unused tables and columns (_job\_status_ table, _job\_agent_ table, various columns in the _job_ table)
- Modify the data within the _job_ table to reflect the new schema and goal (e.g. remove the _keyword_ column and change the _parameters_ column to _ttl\_hr_, etc.)
- ~TODO/Outstanding: Update `app/db/seeds.sql` and `app/db/patches.sql` to account for db schema changes~ Done

TO API:
- Add v4.0 InvalidationJob endpoints and structs to accomplish the requirements in the _blueprint/refetch_ goals.
- Modify the previous endpoints to ensure compatibility with the DB schema changes

TO clients:
- Modify the v4 clients to take advantage of the TO v4.0 API and tc structs
- Ensure the tests pass

TP:
- Modify TP to use the _stable_ (3.1) version of the /jobs endpoint

T3C (formerly ORT):
- Slight modification to ensure tests pass with the DB schema

Documentation:
- TODO/Outstanding: Want to ensure we are actually going to merge this into v4.0 before modifying all the documentation to reflect this. 

<hr>

- Documentation
- Traffic Control Go Client v4 
- Traffic Ops
- Traffic Portal

To test whether REFETCH works specifically or not, you will need to set the parameter `refetch_enabled` to `true` for the _global_ config file in the DB parameter table. Otherwise only REFRESH jobs will be allowed.

The TO database migration scripts will need to be run to ensure the transition from a previous schema works to the new. This is accomplished by using the `db/admin` utility as outlined [here](https://traffic-control-cdn.readthedocs.io/en/latest/development/traffic_ops.html#app-db-admin).

Once the DB migration has taken place, the Traffic Ops client API integration tests can be run. You will need TO to be up and running with the _to\_test_ database set up as described [here](https://traffic-control-cdn.readthedocs.io/en/latest/development/traffic_ops.html#test-cases).
   - To test all currently supported APIs (ensures backwards compatibility as well) you can do something akin to:
   ```bash
   cd github.com/apache/trafficcontrol/traffic_ops/testing/api
   # default location is github.com/apache/trafficcontrol/traffic_ops/testing/api/conf/traffic-ops-test.conf
   go test --cfg <location for your test config> ./...
   ```
   - To test only the invalidation job specific changes for API 4.0 you could run something akin to:
   ```bash
   cd github.com/apache/trafficcontrol/traffic_ops/testing/api/v4
   # default location is github.com/apache/trafficcontrol/traffic_ops/testing/api/conf/traffic-ops-test.conf
   go test -run Jobs --cfg <location for your test config>
   ```

Once the Traffic Ops Integration tests confirm the new and previous implementations of the _/jobs_ endpoints still work then you can proceed to running and testing the TP integration tests still work. This will verify that TP is utilizing the previous stable version of the TO API (v3.1) and not the new _/jobs_ endpoints.

Ensure a local version of TO is up and running with the correct DB schema. This should already be the case based on the TO client integration tests from before.

Ensure Traffic Portal is building and running as outline (here)[https://traffic-control-cdn.readthedocs.io/en/latest/development/traffic_portal.html#traffic-portal].

> By default, the TP integration tests are set up to run against CIAB. If you're running this locally you will need to modify the config file located at github.com/apache/trafficcontrol/traffic_portal/test/integration/config.json. The `params` section will need to be modified:
```json
  "params": {
    "apiUrl": " https://localhost:8443/api/4.0", 
    "baseUrl": "https://localhost:9090",
    "login": {
      "username": "admin",
      "password": "twelve12"
    },
```
> Where
    apiUrl = Running Traffic Ops API URL (This can and should be set to 4.0)
    baseUrl = Running Traffic Portal URL
    login.username = existing administrator account username
    login.password = existing administrator account password
You can also, optionally, remove the `--headless` flag to see the integration tests as they run in Chrome Browser

```bash
cd github.com/apache/trafficcontrol/traffic_portal/test/integration
npm i
npm run start-webdriver
npm run test
```

## PR submission checklist
- [x] This PR has tests 
- [x] This PR has documentation - This is delayed for the time being until accepted or rejected in the current RC (6.0) for ATC. 
- [X] This PR has a CHANGELOG.md entry


lib/go-tc/invalidationjobs.go
```diff
@@ -456,3 +463,210 @@ func (job *UserInvalidationJobInput) Validate(tx *sql.Tx) error {
...
+}
+
+func (job *InvalidationJobCreateV40) GetDSIDfromDSXMLID(tx *sql.Tx) (uint, error) {
+	var dsID uint
+	row := tx.QueryRow(`SELECT id FROM deliveryservice WHERE xml_id=$1`, job.DeliveryService)
+	if err := row.Scan(&dsID); err != nil {
+		if err == sql.ErrNoRows {
+			return 0, fmt.Errorf("No DeliveryService exists matching identifier: %v", job.DeliveryService)
+		}
+		return 0, errors.New("Unknown error occurred")
```
ocket8888, line 660: well it wasn't an unknown error, we just dropped it. This should return that error. I think the reason you're not doing that is because you don't want to expose database errors to the client, and you're returning an error that tells that the DS doesn't exist which will be shown to the client. This is why `dbhelpers` functions typically return _three_ values for things like this - the requested value, whether or not it could be found, and whatever database error might have occurred, which is assumed to be unsafe for the client to view.

lib/go-tc/invalidationjobs.go
```diff
@@ -456,3 +463,210 @@ func (job *UserInvalidationJobInput) Validate(tx *sql.Tx) error {
...
+	}
+	return dsID, nil
+}
+
+// RefetchAllowed checks whether Refetch is allowed and enabled as a global paramter
+// TODO: Error is swallowed from DB...
+func RefetchAllowed(tx *sql.Tx) bool {
+	refetchEnabled := false
+	// Silently ignore paramter error
+	tx.QueryRow(`SELECT value FROM parameter WHERE name='refetch_enabled' AND config_file='global'`).Scan(&refetchEnabled)
```
ocket8888, line 670: I assume that this won't always fail - which is what I would have guessed, but I'm sure you'd have noticed that -, but it for sure can be made to fail. Parameter values are strings, not booleans, so even if this coerces values `'true'` and `'false'` to booleans, it will for sure fail if the value of the Parameter is something like `'foo'`, which means that you can make a Parameter that's perfectly valid that will break an API endpoint so that it only ever returns 500 responses.

We also definitely don't want to be swallowing that error.

Further, the problem with this is that there is no validation on Parameters, and that WHERE clause can even match multiple Parameters. Plus, by convention, we put global Parameters in the GLOBAL Profile - but that Profile could also be deleted or renamed at any time.

So if I have two Parameters:
```json
[{
    "name": "refetch_enabled",
    "configFile": "global",
    "value": "true",
    "secure": false,
    "id": 1
},
{
    "name": "refetch_enabled",
    "configFile": "global",
    "value": "false",
    "secure": false,
    "id": 2
}]
```
Then the one chosen by this method is undefined, even if one of them shows up when you look at the GLOBAL profile, the other still might get chosen.

For all these reasons, I tend to prefer using configuration file parameters (i.e. in "cdn.conf"), whose values can be exposed through the `system/info` endpoint if other components need to know about them.

traffic_ops/app/db/migrations/2021081321025500_refetch.up.sql
```diff
@@ -0,0 +1,65 @@
...
+ * (adding ##REFETCH## to the end of the url) then assign the
+ * correct invalidation_type.
+ */
+UPDATE public.job 
+SET invalidation_type = 'REFETCH'
+WHERE id IN (
+	SELECT id 
+	FROM public.job 
+	WHERE asset_url LIKE '%##REFETCH##'
+);
```
ocket8888, line 58: I don't think you need this nested query. You can just take the `WHERE` clause from the nested query and use that as the `WHERE` clause for your `UPDATE` query.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -42,52 +42,40 @@ import (
...
-       u.username AS createdBy,
-       ds.xml_id AS dsId
-FROM job
-JOIN tm_user u ON job.job_user = u.id
-JOIN deliveryservice ds  ON job.job_deliveryservice = ds.id
-`
+type InvalidationJobV40 struct {
+	api.APIInfoImpl `json:"-"`
+	tc.InvalidationJobV40
+}
```
ocket8888, line 54: Are we using the CRUDer? I thought we wanted to move away from that as a project in favor of middlewares and expanding `APIInfo` structure capabilities?

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -282,6 +493,152 @@ func (job *InvalidationJob) Read(h http.Header, useIMS bool) ([]interface{}, err
...
+		}
+
+		resp, err := json.Marshal(response)
+		if err != nil {
+			api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, fmt.Errorf("Encoding bad request response: %v", err))
+			return
+		}
+
+		w.WriteHeader(http.StatusBadRequest)
+		api.WriteAndLogErr(w, r, append(resp, '\n'))
```
ocket8888, line 529: This should probably just be a one-liner that does the appending for you:
```go
api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, nil, nil)
```
Honestly I don't think `api.WriteAndLogErr` should even be exported

ocket8888 (commented):
I tried to look kind of broad-strokes at the design; there's more to say but it's not worth saying yet since this is just a draft.

Biggest thing that stood out to me was the CRUDer rearing its ugly head again

rawlinp:
> Biggest thing that stood out to me was the CRUDer rearing its ugly head again

This PR doesn't appear to be using the CRUDer any more than the _original_ `/jobs` implementation. It still uses it for `Read` it appears, but create, update, and delete are all as CRUDer-free as the prior version. I'm guessing the prior version functions were mostly just copied then modified to work with the new changes we want.

ocket8888:
I guess it's not fair to expect him to do more work than he needs to. It's just so terrible right now I can't believe he thought any of it was worth keeping.

rawlinp (changes requested):
Didn't manage to complete this review, but here are 22 comments in the meantime. A lot of them are related to using a "major version alias", so let me know if I need to explain that better. The basic idea is that it should be easy to add new minor version structs without having to update hundreds of function signatures all over the place.

lib/go-tc/invalidationjobs.go
```diff
@@ -456,3 +462,209 @@ func (job *UserInvalidationJobInput) Validate(tx *sql.Tx) error {
...
+
+	if len(errs) > 0 {
+		return errors.New(strings.Join(errs, ", "))
+	}
+
+	return nil
+}
+
+// validateDeliveryService ensures the supplied (required) Delivery Service XML ID exists
+func (job *InvalidationJobCreateV40) validateDeliveryService(tx *sql.Tx) error {
```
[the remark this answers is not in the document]
ocket8888: > In fact, we probably already have a function in `dbhelpers.go` that does exactly that.

Looks like we don't, actually. Best you could do looks like using `GetDSTenantIDFromXMLID` and just throwing away the tenant ID return value. Better solution would be to make a simple `DSExists` function in that package, IMO.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -42,52 +42,40 @@ import (
...
-       start_time,
-       u.username AS createdBy,
-       ds.xml_id AS dsId
-FROM job
-JOIN tm_user u ON job.job_user = u.id
-JOIN deliveryservice ds  ON job.job_deliveryservice = ds.id
-`
+type InvalidationJobV40 struct {
+	api.APIInfoImpl `json:"-"`
+	tc.InvalidationJobV40
```
rawlinp, line 53: V40 -> V4 alias

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -98,11 +86,49 @@ RETURNING
...
 	 WHERE tm_user.id=job_user) AS createdBy,
-	keyword,
-	parameters,
+	'PURGE' AS keyword,
+	CONCAT('TTL:', ttl_hr, 'h') AS parameters,
 	start_time
 `
 
+// Almost the same as insertQuery, but returns appropriate values for API 4.0+
+const insertQueryV40 = `
```
rawlinp, line 95: We should probably call this `V4` instead of `V40`, because we'll likely just update this query with new fields if they're ever added.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -130,11 +156,35 @@ WHERE server.status NOT IN (
...
+          start_time
+`
+
+// Almost the same as updateQuery, but returns appropriate values for API 4.0+
+const updateQueryV40 = `
+UPDATE job
+SET asset_url=$1,
+    ttl_hr=$2,
+    start_time=$3,
+	invalidation_type=$4
```
rawlinp, line 185: formatting is off

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -146,20 +196,40 @@ RETURNING job.asset_url,
            WHERE deliveryservice.id=job.job_deliveryservice
           ) AS delivery_service,
           job.id,
-          job.keyword,
-          job.parameters,
-          job.start_time
+		  ttl_hr,
+          start_time,
+		  invalidation_type
```
rawlinp, line 201: formatting is off

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -130,11 +156,35 @@ WHERE server.status NOT IN (
...
+           WHERE deliveryservice.id=job.job_deliveryservice
+          ) AS delivery_service,
+          job.id,
+          'PURGE' as keyword,
+          CONCAT('TTL:', ttl_hr, 'h') AS parameters,
+          start_time
+`
+
+// Almost the same as updateQuery, but returns appropriate values for API 4.0+
+const updateQueryV40 = `
```
rawlinp, line 180: We should probably call this V4 instead of V40, because we'll likely just update this query with new fields if they're ever added.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -146,20 +196,40 @@ RETURNING job.asset_url,
...
+       origin.protocol || '://' || origin.fqdn || rtrim(concat(':', origin.port), ':') AS OFQDN
+FROM job
+INNER JOIN origin ON origin.deliveryservice=job.job_deliveryservice AND origin.is_primary
+INNER JOIN tm_user ON tm_user.id=job.job_user
+INNER JOIN deliveryservice ON deliveryservice.id=job.job_deliveryservice
+WHERE job.id=$1
+`
+
+// Almost the same as putInfoQuery, but returns appropriate values for API 4.0+
+const putInfoQueryV40 = `
```
rawlinp, line 223: We should probably call this V4 instead of V40, because we'll likely just update this query with new fields if they're ever added.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -146,20 +196,40 @@ RETURNING job.asset_url,
...
+SELECT job.id AS id,
+	   tm_user.username AS createdBy,
+	   job.job_user AS createdByID,
+       job.job_deliveryservice AS dsid,
+       deliveryservice.xml_id AS dsxmlid,
+       job.asset_url AS assetURL,
+       job.ttl_hr AS ttlhrs,
        job.start_time AS start_time,
+	   job.invalidation_type as invalidationType,
```
rawlinp, line 232: formatting is off

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -168,11 +238,35 @@ INNER JOIN deliveryservice ON deliveryservice.id=job.job_deliveryservice
...
+           FROM tm_user
+           WHERE tm_user.id=job.job_user
+          ) AS created_by,
+          (
+           SELECT deliveryservice.xml_id
+           FROM deliveryservice
+           WHERE deliveryservice.id=job.job_deliveryservice
+          ) AS deliveryservice,
+          job.id,          
+		  'PURGE' as keyword,
```
rawlinp, line 257: formatting is off

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -168,11 +238,35 @@ INNER JOIN deliveryservice ON deliveryservice.id=job.job_deliveryservice
...
+`
+
+// Almost the same as deleteQuery, but returns appropriate values for API 4.0+
+const deleteQueryV40 = `
+DELETE
+FROM job
+WHERE job.id=$1
+RETURNING 
+		  job.id,
+		  job.asset_url,
```
rawlinp, line 269: formatting is off

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -168,11 +238,35 @@ INNER JOIN deliveryservice ON deliveryservice.id=job.job_deliveryservice
...
+           WHERE deliveryservice.id=job.job_deliveryservice
+          ) AS deliveryservice,
+          job.id,          
+		  'PURGE' as keyword,
+          CONCAT('TTL:', ttl_hr, 'h') AS parameters,
+          job.start_time
+`
+
+// Almost the same as deleteQuery, but returns appropriate values for API 4.0+
+const deleteQueryV40 = `
```
rawlinp, line 263: We should probably call this V4 instead of V40, because we'll likely just update this query with new fields if they're ever added.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -183,9 +277,8 @@ RETURNING job.asset_url,
            FROM deliveryservice
            WHERE deliveryservice.id=job.job_deliveryservice
           ) AS deliveryservice,
-          job.id,
-          job.keyword,
-          job.parameters,
+          ttl_hr,
+		  job.invalidation_type,
```
rawlinp, line 281: formatting is off

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -203,17 +301,124 @@ func selectMaxLastUpdatedQuery(where string) string {
 	select max(last_updated) as t from last_deleted l where l.table_name='job') as res`
 }
 
+// Deprecated, only to be used with versions below 4.0
+const readQuery = `
+SELECT job.id,
+        'PURGE' AS keyword,
```
rawlinp, line 307: formatting is off

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -203,17 +301,124 @@ func selectMaxLastUpdatedQuery(where string) string {
...
+		start_time,
+		u.username as createdBy,
+		ds.xml_id as dsId
+FROM job
+JOIN tm_user u ON job.job_user = u.id
+JOIN deliveryservice ds ON job.job_deliveryservice = ds.id
+`
+
+// Almost the same as readQuery, but returns appropriate values for API 4.0+
+const readQueryV40 = `
```
rawlinp, line 319: We should probably call this V4 instead of V40, because we'll likely just update this query with new fields if they're ever added.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -203,17 +301,124 @@ func selectMaxLastUpdatedQuery(where string) string {
...
+	for rows.Next() {
+		job := tc.InvalidationJobV40{}
+		err := rows.Scan(&job.ID,
+			&job.AssetURL,
+			&job.CreatedBy,
+			&job.DeliveryService,
+			&job.TTLHours,
+			&job.InvalidationType,
+			&job.StartTime)
+		if err != nil {
```
rawlinp, line 395: nit: this error check is normally combined into the Scan like this:
```
if err := rows.Scan(...); err != nil {
    ...
}
```

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -270,6 +475,12 @@ func (job *InvalidationJob) Read(h http.Header, useIMS bool) ([]interface{}, err
 			return nil, nil, fmt.Errorf("parsing db response: %v", err), http.StatusInternalServerError, nil
 		}
 
+		// NamedQuery (nor prepare statements) cannot handle string functions in SELECT such as:
+		// 	CONCAT('TTL:', ttl_hr, 'h') AS parameters
```
rawlinp, line 479: I'm guessing this might be because of the `:` in `TTL:`. I think using `TTL::` might work? That is to say, I think it might actually be possible to use `CONCAT` and other string functions in `NamedQuery`.

rawlinp (changes requested):
Almost done reviewing this entire thing -- just need to finish the 2nd half of `traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go `.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -429,6 +787,198 @@ func Create(w http.ResponseWriter, r *http.Request) {
...
+		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
+		return
+	}
+	defer inf.Close()
+
+	var oFQDN string
+	var dsid uint
+	var uid uint
+	job := tc.InvalidationJobV40{}
+	row := inf.Tx.Tx.QueryRow(putInfoQueryV40, inf.Params["id"])
```
rawlinp, line 802: I'm thinking `id` needs to be a `requiredParam` and `intParam` in the call to `api.NewInfo`.

rawlinp (changes requested):
Ok, I've finally finished going through this entire thing.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -203,17 +301,124 @@ func selectMaxLastUpdatedQuery(where string) string {
...
+		"createdBy":        dbhelpers.WhereColumnInfo{Column: `(SELECT tm_user.username FROM tm_user WHERE tm_user.id=job.job_user)`},
+		"deliveryService":  dbhelpers.WhereColumnInfo{Column: `(SELECT deliveryservice.xml_id FROM deliveryservice WHERE deliveryservice.id=job.job_deliveryservice)`},
+		"dsId":             dbhelpers.WhereColumnInfo{Column: "job.job_deliveryservice", Checker: api.IsInt},
+		"invalidationType": dbhelpers.WhereColumnInfo{Column: "invalidation_type"},
+	}
+
+	where, orderBy, pagination, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(job.APIInfo().Params, queryParamsToSQLCols)
+	if len(errs) > 0 {
+		return nil, util.JoinErrs(errs), nil, http.StatusBadRequest, nil
+	}
```
rawlinp, line 351: I completed this after you started this PR, but please ensure that [my changes](https://github.com/apache/trafficcontrol/pull/6154/files#diff-3310237988f25e4cdf0632f9d8be54da09f3ac609ac7d8e8ba4dd74254d36bec) made to the prior API versions is replicated in API v4. We need the ability to filter the query by `cdn` and `maxRevalDurationDays`.

rawlinp:
I think the DB migrations may need to be updated to a more recent timestamp: ```traffic_ops/app/db/migrations/2021081321025500_refetch.down.sql```

Because there have been other migrations that have come in with more recent timestamps, this one won't be run in dev environments that have already run those more recent migrations.

rawlinp (approved):

rawlinp:
TP integration test failure appears to be a transient issue unrelated to this PR, merging ahead.

PR #6120, opened 2021-08-18
TP tests alert logging

ocket8888:
This PR adds logging `alerts` to the console for any and all raw API requests made by the Traffic Portal integration tests. So that if, e.g. the Roles your TO instance have don't match the expected hard-coded id-to-Role mappings of the tests, you'll be able to see an error that says `no such role` in the console.

This also cleans up the GHA logs by moving TO, TV, and TP logs into build artifacts that only get uploaded on failure. This makes it **_much_** easier to follow testing output and see what went wrong, and the TO/TV logs are less necessary to understand why tests fail now that response alerts are being logged.

<hr>

- Traffic Portal (integration tests)
- Automation (GitHub Actions)

Make sure the test actions pass. To see an example of the alert logging, you could make a branch based on this changeset, comment out the line in Traffic Ops's `routes.go` that adds the `/api/4.0/user/login` endpoint (API version there matters; the tests use 4.0) and push it to your remote. Then you should see a line very near the bottom after between 2 and 4 minutes when the action fails that says:
```
ERROR: POST https://172.18.0.1:8443/api/4.0/user/login (404 Not Found): The requested path '/api/4.0/user/login' does not exist.
```

## PR submission checklist
- [x] This PR has tests 
- [ ] This PR has no documentation because it doesn't change the existing behavior of any component (testing configuration changes are also non-breaking for valid configurations)
- [ ] This PR has no CHANGELOG.md entry because it doesn't change the existing behavior of any component


.github/actions/tp-integration-tests/entrypoint.sh
```diff
@@ -167,56 +182,42 @@ to_build() {
...
+  pushd "${REPO_DIR}/traffic_portal"
   npm ci
   bower install
   grunt dist
 
   cp "${resources}/config.js" ./conf/
   touch tp.log access.log out.log err.log
-  sudo forever --minUptime 5000 --spinSleepTime 2000 -f -o out.log start server.js &
-  tail -f err.log 2>&1 | color_and_prefix "${red_bg}" "Node Err" &
+  sudo forever --minUptime 5000 --spinSleepTime 2000 -f -o out.log start server.js >out.log 2>&1 &
```
shamrickus, line 199: `-o out.log` is redundant.

.github/actions/tp-integration-tests/cdn.json
```diff
@@ -13,11 +13,11 @@
...
 		"port": "6443",
-		"log_location_error": "error.log",
-		"log_location_warning": "warning.log",
-		"log_location_info": "info.log",
-		"log_location_debug": null,
-		"log_location_event": "event.log",
+		"log_location_error": "stderr",
+		"log_location_warning": "stderr",
+		"log_location_info": "stdout",
+		"log_location_debug": "stderr",
```
shamrickus, line 19: Shouldn't this be `stdout`?

traffic_portal/test/integration/config.model.ts
```diff
@@ -0,0 +1,165 @@
...
+ *
+ * @param c The object to check.
+ * @returns `true`, always. When the check fails, it throws an error that
+ * explains why.
+ */
+export function isTestingConfig(c: unknown): c is TestingConfig {
+	if (typeof(c) !== "object") {
+		throw new Error(`testing configuration must be an object, not a '${typeof(c)}'`);
+	}
+	if (c === null) {
```
shamrickus, line 122: nit: null check before `typeof` is marginally faster in case of nulls

shamrickus (changes requested):

.github/actions/tp-integration-tests/cdn.json
ocket8888: It is

.github/actions/tp-integration-tests/entrypoint.sh
ocket8888: Yeah, looking at the docs I'm not sure how `-f -o out.log` would work; would it log stdout from the child process to `out.log` and pip the child's `stderr` through its own `stdout`? Unclear. I'll get rid of `-o out.log`, it seems like that's the best way to ensure all the logs get to where they need to be.

traffic_portal/test/integration/config.model.ts
ocket8888: Yeah, but `typeof` before null check is marginally faster in case of anything that isn't `null` or an object, which is more things than just `null`. I think it'd be more common for the configuration to have a typo or be incomplete so that the input value is actually `undefined` like e.g.:
```json
{
    "arams": {
        "config": "options here"
    }
}
```
... than it would be for it to be like:
```json
{
    "params": null
}
```

.github/actions/tp-integration-tests/cdn.json
ocket8888: or did you mean like "should this _really_ be stdout instead of stderr?" Or were you talking about "log_location_debug"? There was a point when "log_location_info" was set to just "out" instead of "stdout", so I assumed you made this comment before I fixed that typo, but now that I think about it I'm not sure.

shamrickus: I meant since info is stdout shouldn't debug also be stdout

traffic_portal/test/integration/config.model.ts
shamrickus: Fair point

.github/actions/tp-integration-tests/cdn.json
ocket8888: eh, sure. It doesn't make much difference, honestly maybe they should all just be the same stream. Idk, I'll just make it stdout for now.

shamrickus (approved):
LGTM

PR #6124, opened 2021-08-19
Role and User struct changes for permission based roles

srijeet0406:
This PR changes the V4 `Role` and `User` structs to account for the changes as specified in the Roles and Perms blueprint.

<hr>

- Traffic Ops

Make sure all the unit tests and API tests pass.

## PR submission checklist
- [x] This PR has tests 


lib/go-tc/users.go
```diff
@@ -33,6 +33,249 @@ import (
...
+		ret.TenantID = *u.TenantID
+	}
+	if u.Username != nil {
+		ret.Username = *u.Username
+	}
+	return ret
+}
+
+// Downgrade converts a UserV50 to a UserV40 (as seen in API versions 4.x) and User (as seen in API versions < 4.0)
+func (u UserV50) Downgrade() (UserV40, User) {
```
ocket8888, line 194: If we already have a `DowngradeToLegacy`, why the second return value for a legacy `User` here? Would it help if, instead of having ways to upgrade from/downgrade to a regular `User` straight from a `UserV40` if there was a `UserV50.Downgrade` that downgrades to `UserV40`, a `UserV40.Downgrade` that downgrades a V40 to a `User`, a `User.Upgrade` that upgrades a `User` to V40, and a `UserV40.Upgrade` that upgrades a `UserV40` to V5?

That way, if you need a UserV40 and a User generated from the same UserV50, then you can do `userV4 = userV5.Downgrade()` and then when you need a `User` you can just `Downgrade` again.

lib/go-tc/users.go
```diff
@@ -105,6 +348,33 @@ type UserCurrentV40 struct {
...
+	PostalCode        *string    `json:"postalCode"`
+	PublicSSHKey      *string    `json:"publicSshKey"`
+	Role              *string    `json:"role"`
+	StateOrProvince   *string    `json:"stateOrProvince"`
+	Tenant            *string    `json:"tenant"`
+	TenantID          *int       `json:"tenantId"`
+	Token             *string    `json:"-"`
+	UID               *int       `json:"uid"`
+	LastUpdated       *time.Time `json:"lastUpdated"`
+	LastAuthenticated *time.Time `json:"lastAuthenticated"`
```
ocket8888, line 375: `UserV50` has non-pointer types for `LastUpdated`, `LastAuthenticated`, `Role`, and `UserName`, and I think those ought to match here

ocket8888: Also same as above RE: V5 alias

lib/go-tc/users.go
```diff
@@ -136,9 +406,49 @@ type CurrentUserUpdateRequestUser struct {
 	Username           json.RawMessage `json:"username"`
 }
 
+func (u UserCurrentV40) UpgradeUserCurrent() UserCurrentV50 {
```
ocket8888, line 409: GoDoc?

Also, nit: Since this can be only be called from a `UserCurrentV40`, we already know it's upgrading "from a UserCurrentV40", those two words make the name longer without giving us any extra information. Could just be `Upgrade`.

lib/go-tc/users.go
```diff
@@ -334,6 +665,15 @@ type UserRegistrationRequest struct {
...
 
+// UserRegistrationRequestV50 is the request submitted by operators when they want to register a new
+// user in api V5.
+type UserRegistrationRequestV50 struct {
+	Email rfc.EmailAddress `json:"email"`
+	// Role - despite being named "Role" - is actually merely the *ID* of a Role to give the new user.
+	Role     string `json:"role"`
+	TenantID uint   `json:"tenantId"`
+}
+
```
ocket8888, line 676: Same as above RE: V5 alias

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1603,3 +1624,35 @@ func GetDSIDFromStaticDNSEntry(tx *sql.Tx, staticDNSEntryID int) (int, error) {
...
+	}
+	if where == "" {
+		where = BaseWhere + " "
+	} else {
+		where += " AND "
+	}
+	return where + extra
+}
+
+// GetRoleIDFromName returns the ID of the role associated with the supplied name
```
ocket8888, line 1642: GoDoc should end with punctuation.

Also, most functions like this return a `bool` indicating whether or not the identified object could be found, should this one?

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1603,3 +1624,35 @@ func GetDSIDFromStaticDNSEntry(tx *sql.Tx, staticDNSEntryID int) (int, error) {
...
+func GetRoleIDFromName(tx *sql.Tx, roleName string) (int, error) {
+	var id int
+	if err := tx.QueryRow(`SELECT id FROM role WHERE name = $1`, roleName).Scan(&id); err != nil {
+		return -1, errors.New("querying role ID from name: " + err.Error())
+	}
+	return id, nil
+}
+
+// GetRoleNameFromID returns the name of the role associated with the supplied ID
```
ocket8888, line 1651: same as the above comments on GetRoleIDFromName

traffic_ops/traffic_ops_golang/login/register.go
```diff
@@ -164,6 +164,15 @@ func RegisterUser(w http.ResponseWriter, r *http.Request) {
...
 	}
+	// ToDo: uncomment this once the perm based roles and config options are implemented
+	//var reqV50 tc.UserRegistrationRequestV50
+	//if inf.Version.Major >= 5 {
+	//	if userErr = api.Parse(r.Body, tx, &reqV50); userErr != nil {
+	//		api.HandleErr(w, r, tx, http.StatusBadRequest, userErr, nil)
+	//		return
+	//	}
+	//	tenantID = reqV50.TenantID
+	//}
```
ocket8888, line 175: Why can't we do this now? Or if you want to do it in another PR, why leave the comment in?

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +291,522 @@ RETURNING id, last_updated`
...
+	//		return
+	//	}
+	//
+	//	rows, err := tx.Query(updateLegacyRoleQuery(), roleName, roleDesc, roleID)
+	//	if err != nil {
+	//		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("updating role here: "+err.Error()))
+	//		return
+	//	}
+	//	rows.Close()
+	//}
```
ocket8888, line 421: doesn't this being commented-out break all PUT requests to `/roles` in API versions earlier than 5.0?

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +291,522 @@ RETURNING id, last_updated`
...
+
+	if rows, err := result.RowsAffected(); err != nil {
+		log.Errorf("could not check result after inserting role_capability relations: %v", err)
+	} else if expected := len(*permissions); int(rows) != expected {
+		log.Errorf("wrong number of role_capability rows created: %d expected: %d", rows, expected)
+	}
+	return nil, nil, http.StatusOK
+}
+
+func Validate(tx *sqlx.Tx, role tc.Role, roleV50 tc.RoleV50, version *api.Version) error {
```
ocket8888, line 503: Does this need to be exported out of the package?

Instead of using different arguments based on the API version, could we not just have a different validation func for each version of a Role? I think what you're trying to do here is something like
```typescript
export function Validate(tx: sqlx.Tx, role: tc.Role | tc.RoleV50, version: api.Version) {
```
but the reality is that things like that don't work well in Go.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +291,522 @@ RETURNING id, last_updated`
...
+		errs = validation.Errors{
+			"name":        validation.Validate(role.Name, validation.Required),
+			"description": validation.Validate(role.Description, validation.Required),
+			"privLevel":   validation.Validate(role.PrivLevel, validation.Required),
+		}
+		capabilities = role.Capabilities
+	}
+
+	errsToReturn := tovalidate.ToErrors(errs)
+	checkCaps := `SELECT cap FROM UNNEST($1::text[]) AS cap WHERE NOT cap =  ANY(ARRAY(SELECT c.name FROM capability AS c WHERE c.name = ANY($1)))`
```
ocket8888, line 522: We shouldn't do this, the source of truth for what Permissions exist is just the API documentation itself, not the database. That's why I got rid of the foreign key constraint, which enforced this at a database level.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +291,522 @@ RETURNING id, last_updated`
...
+			where = dbhelpers.AppendWhere(where, "permissions @> :can")
+		}
+	}
+	if inf.Config.UseIMS {
+		runSecond, maxTime = ims.TryIfModifiedSinceQuery(inf.Tx, r.Header, queryValues, selectMaxLastUpdatedQuery(where))
+		if !runSecond {
+			log.Debugln("IMS HIT")
+			api.AddLastModifiedHdr(w, maxTime)
+			w.WriteHeader(http.StatusNotModified)
+			api.WriteResp(w, r, []interface{}{})
```
ocket8888, line 760: Is this `api.WriteResp` call actually necessary? I see other IMS hits doing it, but idk why, since the whole point is that clients will just ignore it and use their cache.

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -508,3 +519,426 @@ func (user *TOUser) InsertQuery() string {
...
+
+	where, orderBy, pagination, queryValues, errs := dbhelpers.BuildWhereAndOrderByAndPagination(inf.Params, params)
+	if len(errs) != 0 {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, util.JoinErrs(errs), nil)
+		return
+	}
+
+	tenantIDs, err := tenant.GetUserTenantIDListTx(inf.Tx.Tx, inf.User.TenantID)
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("getting tenant list for user: %v\n", err))
```
ocket8888, line 634: Errors should not contain newlines, also formatting for errors should use `%w` instead of `%v` so that it wraps

ocket8888 (commented):
There is at least one comment in here that won't be relevant when APIv5 is added; I didn't realize that had been pulled out of the PR when I made it.

lib/go-tc/users.go
srijeet0406: Since most of the struct params in `User` and `UserV40` are the same, I took care of both downgrades in one function, instead of duplicating the code. I don't mind changing it to what you're suggesting. Thoughts?

lib/go-tc/roles.go
```diff
@@ -19,6 +28,70 @@ package tc
...
+// RoleV5 is an alias for RoleV50.
+type RoleV5 RoleV50
+
+// RolesResponseV5 is a list of RoleV50 as a response.
+type RolesResponseV5 struct {
+	Response []RoleV5 `json:"response"`
+	Alerts
+}
+
+// RoleResponseV5 is a RoleV50 as a response.
```
ocket8888, line 40: Nit but these are actually `RoleV5`s as responses.

lib/go-tc/roles.go
```diff
@@ -19,6 +28,70 @@ package tc
  * under the License.
  */
 
+// RoleV5 is an alias for RoleV50.
```
ocket8888, line 31: While this is technically true, the purpose of the alias is to always point to the latest minor version for its associated major version.

ocket8888 (commented):
I was starting to review this when I realized it still targets API v5, but we're not doing that at this point, per [mailing list discussion](https://lists.apache.org/thread.html/rdf09f89d935597367cd2b5bbdeac20929066b5e2f286e7757cee39a0%40%3Cdev.trafficcontrol.apache.org%3E).

srijeet0406:
> I was starting to review this when I realized it still targets API v5, but we're not doing that at this point, per [mailing list discussion](https://lists.apache.org/thread.html/rdf09f89d935597367cd2b5bbdeac20929066b5e2f286e7757cee39a0%40%3Cdev.trafficcontrol.apache.org%3E).

Yeah I'm in the process of backing out the v5 changes. Should have it ready today.

docs/source/api/v4/roles.rst
```diff
@@ -83,19 +80,19 @@ Response Structure
...
-			"description": "super-user",
-			"privLevel": 30,
-			"capabilities": [
+			"description": "Has access to everything.",
+			"permissions": [
 				"all-write",
-				"all-read"
-			]
+				"all-read",
+				...,
```
ocket8888, line 93: this line is invalid JSON, which is breaking lexing and therefore syntax highlighting

srijeet0406:
Working on the test fixes.

docs/source/api/v4/roles.rst
ocket8888: If you're concerned that the list takes up too much space, you could always create a new Role with only one or two Permissions and then show that request/response - or don't, it's fine how it is IMO.

srijeet0406: Yeah I just used the `read-only` role instead, because that has lesser permissions than the `admin` role.

traffic_ops/testing/api/v4/cachegroups_test.go
```diff
@@ -101,26 +101,25 @@ func UpdateCachegroupWithLocks(t *testing.T) {
...
+		Role:                 "operations",
 	}
 	user1.Email = util.StrPtr("lockuseremail@domain.com")
-	user1.TenantID = util.IntPtr(1)
+	user1.TenantID = 1
 	user1.FullName = util.StrPtr("firstName LastName")
 	_, _, err = TOSession.CreateUser(user1, client.RequestOptions{})
 	if err != nil {
-		t.Fatalf("could not create test user with username: %s", *user1.Username)
+		fmt.Println(err)
```
ocket8888, line 116: Looks like you left in a print statement you used for debugging - if the error was useful, it should probably get logged in the `t.Fatalf` call (and I can't imagine a more useful piece of information that could be logged in the event of a failure here)

traffic_ops/testing/api/v4/roles_test.go
```diff
@@ -63,30 +62,23 @@ func UpdateTestRolesWithHeaders(t *testing.T, header http.Header) {
...
-		t.Fatal("Traffic Ops returned a representation for a Role with null or undefined ID")
-	}
-
-	expectedRole := "new admin2"
-	remoteRole.Name = &expectedRole
+	expectedDescription := "new description"
+	remoteRole.Description = expectedDescription
 	opts.QueryParameters.Del("name")
-	_, reqInf, _ := TOSession.UpdateRole(*remoteRole.ID, remoteRole, opts)
+	_, reqInf, err := TOSession.UpdateRole(remoteRole.Name, remoteRole, opts)
```
ocket8888, line 81: you're now capturing the error, but the value is never read after this assignment

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1603,3 +1624,41 @@ func GetDSIDFromStaticDNSEntry(tx *sql.Tx, staticDNSEntryID int) (int, error) {
 	}
 	return dsID, nil
 }
+
+// AppendWhere appends 'extra' safely to the WHERE clause 'where'. What is
+// returned is guaranteed to be a valid WHERE clause (including a blank string).
```
ocket8888, line 1629: I think I was being a little cocky when I wrote this docstring - it's only guaranteed to be a valid WHERE clause when `where` is already a valid WHERE clause and `extra` is a valid extra condition statement.

Which is important to note, because it is always `AND`ed - you can't add `OR some = 'thing'` or it'll break. Although tests should probably catch that pretty fast.

I won't ask you to change the GoDoc, but it's a bit more optimistic than it should be.

docs/source/api/v4/roles.rst
```diff
@@ -327,3 +327,5 @@ Response Structure
 		"text": "role was deleted.",
 		"level": "success"
 	}]}
+
+.. [#permissions] ``permissions`` cannot include permissions that are not included in the permissions of the requesting user.
```
ocket8888, line 331: This footnote is never referenced, it should be applied to applicable request and/or response properties

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+			return
+		}
+		if err := roleV4.Validate(); err != nil {
+			api.HandleErr(w, r, tx, http.StatusBadRequest, err, nil)
+			return
+		}
+		roleName = roleV4.Name
+		roleDesc = roleV4.Description
+		privLevel = inf.User.PrivLevel
+		roleCapabilities = &roleV4.Permissions
```
ocket8888, line 654: This should be checking for privilege escalation - you shouldn't be able to create a Role with Permissions you don't have.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+			RoleV11: tc.RoleV11{
+				ID:          util.IntPtr(roleID),
+				Name:        util.StrPtr(roleName),
+				Description: util.StrPtr(roleDesc),
+				PrivLevel:   util.IntPtr(privLevel),
+			},
+			Capabilities: roleCapabilities,
+		}
+	}
+	api.WriteAlertsObj(w, r, http.StatusOK, alerts, roleResponse)
```
ocket8888, line 715: Per API guidelines, the response code should be `201 Created`, ideally with a `Location` header that contains a URI for the created resource.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -100,6 +118,32 @@ func (role *TORole) SetKeys(keys map[string]interface{}) {
...
+	errs := make(map[string]error)
+	errs = validation.Errors{
+		"name":        validation.Validate(role.Name, validation.Required),
+		"description": validation.Validate(role.Description, validation.Required),
+		"privLevel":   validation.Validate(role.PrivLevel, validation.Required),
+	}
+	capabilities = role.Capabilities
+
+	errsToReturn := tovalidate.ToErrors(errs)
+	checkCaps := `SELECT cap FROM UNNEST($1::text[]) AS cap WHERE NOT cap =  ANY(ARRAY(SELECT c.name FROM capability AS c WHERE c.name = ANY($1)))`
```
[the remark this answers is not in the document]
ocket8888: but actually, the database no longer stores a set of the Permissions that are valid, so this check just shouldn't be taking place.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+			return
+		}
+
+		if !api.IsUnmodified(r.Header, *existingLastUpdated) {
+			api.HandleErr(w, r, tx, http.StatusPreconditionFailed, api.ResourceModifiedError, nil)
+			return
+		}
+		rows, err := tx.Query(updateRoleQuery(), roleDesc, roleName)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("updating role: "+err.Error()))
```
ocket8888, line 386: errors constructed from errors should be wrapped using `fmt.Errorf` and `%w`

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+		}
+		defer rows.Close()
+		if !rows.Next() {
+			api.HandleErr(w, r, tx, http.StatusNotFound, errors.New("no role found with this ID"), nil)
+			return
+		}
+		lastUpdated := tc.TimeNoMod{}
+		for rows.Next() {
+			if err := rows.Scan(&lastUpdated); err != nil {
+				api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("scanning lastUpdated from role update: "+err.Error()))
```
ocket8888, line 397: same as above RE: errors

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+		}
+
+		if !api.IsUnmodified(r.Header, *existingLastUpdated) {
+			api.HandleErr(w, r, tx, http.StatusPreconditionFailed, api.ResourceModifiedError, nil)
+			return
+		}
+
+		rows, err := tx.Query(updateLegacyRoleQuery(), roleName, roleDesc, roleID)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("updating role here: "+err.Error()))
```
ocket8888, line 446: same as above, RE: errors

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, err)
+			return
+		}
+
+		if !api.IsUnmodified(r.Header, *existingLastUpdated) {
+			api.HandleErr(w, r, tx, http.StatusPreconditionFailed, api.ResourceModifiedError, nil)
+			return
+		}
+
+		rows, err := tx.Query(updateLegacyRoleQuery(), roleName, roleDesc, roleID)
```
ocket8888, line 444: If we aren't reading the rows, this can just be an `Exec` - but I think we should be reading back `lastUpdated` at least.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+$2,
+$3
+)
+RETURNING id, last_updated`
+}
+
+func createRoleCapabilityAssociations(tx *sqlx.Tx, roleID int, permissions *[]string) (error, error, int) {
+	result, err := tx.Exec(associateCapabilities(), roleID, pq.Array(permissions))
+	if err != nil {
+		return nil, errors.New("creating role capabilities: " + err.Error()), http.StatusInternalServerError
```
ocket8888, line 523: same as above RE: errors

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+	}
+
+	if rows, err := result.RowsAffected(); err != nil {
+		log.Errorf("could not check result after inserting role_capability relations: %v", err)
+	} else if expected := len(*permissions); int(rows) != expected {
+		log.Errorf("wrong number of role_capability rows created: %d expected: %d", rows, expected)
+	}
```
ocket8888, line 530: Should these conditions cause the request to fail?

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+		if roleIDParam, ok := inf.Params["id"]; !ok {
+			api.HandleErr(w, r, tx, http.StatusBadRequest, errors.New("must supply a role ID to delete"), nil)
+			return
+		} else {
+			roleID, err = strconv.Atoi(roleIDParam)
+			if err != nil {
+				api.HandleErr(w, r, tx, http.StatusBadRequest, errors.New("must supply an integral role ID to delete"), nil)
+				return
+			}
+		}
```
ocket8888, line 577: nit but this doesn't need to be in an `else` - the preceding `if` ends with a `return`.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+	alerts := tc.CreateAlerts(tc.SuccessLevel, "role was deleted.")
+	api.WriteAlerts(w, r, http.StatusOK, alerts)
+	changeLogMsg := fmt.Sprintf("ROLE: %s, ID: %d, ACTION: Deleted Role", roleName, roleID)
+	api.CreateChangeLogRawTx(api.ApiChange, changeLogMsg, inf.User, tx)
+}
+
+func deleteRoleCapabilityAssociations(tx *sqlx.Tx, roleID int) (error, error, int) {
+	result, err := tx.Exec(deleteAssociatedCapabilities(), roleID)
+	if err != nil {
+		return nil, errors.New("deleting role capabilities: " + err.Error()), http.StatusInternalServerError
```
ocket8888, line 609: same as above RE: errors

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+	}
+	if _, err = result.RowsAffected(); err != nil {
+		log.Errorf("could not check result after inserting role_capability relations: %v", err)
+	}
```
ocket8888, line 613: Should this cause the request to fail.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+		}
+		roleName = *role.Name
+		roleDesc = *role.Description
+		privLevel = *role.PrivLevel
+		roleCapabilities = role.Capabilities
+	}
+
+	rows, err := tx.Query(createQuery(), roleName, roleDesc, privLevel)
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("creating role: "+err.Error()))
```
ocket8888, line 672: same as above RE: errors

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+		return
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var throwaway interface{}
+		if err := rows.Scan(&roleID, &throwaway); err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("role create: scanning role ID: "+err.Error()))
```
ocket8888, line 680: same as above RE: errors

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +317,502 @@ RETURNING id, last_updated`
...
+		api.WriteResp(w, r, roles)
+	}
+}
+
+func selectMaxLastUpdatedQuery(where string) string {
+	return `SELECT max(t) FROM (
+		SELECT max(r.last_updated) AS t from role r ` + where + ` UNION ALL
+		SELECT max(l.last_updated) AS t from last_deleted l WHERE l.table_name='role' OR l.table_name='role_capability' UNION ALL
+		SELECT max(rc.last_updated) AS t from role_capability rc INNER JOIN role ON rc.role_id = role.id)
+		as res`
```
ocket8888, line 817: Nit: FROM and AS are SQL keywords, should be ALL_CAPS like the others

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +292,529 @@ RETURNING id, last_updated`
...
+	}
+	defer inf.Close()
+
+	version := inf.Version
+	if version == nil {
+		middleware.NotImplementedHandler().ServeHTTP(w, r)
+		return
+	}
+	tx := inf.Tx.Tx
+	if version.Major >= 4 {
```
ocket8888, line 321: Based on `routes.go`, this handler is **only** for APIv4 Roles. I think your changes in this file are building off of my branch, but in my branch I removed the old CRUDer stuff and used each method function as the entrypoint for all versions of that request method.

If you're going to leave the CRUDer in and still use it serve old API versions, then there's no reason to handle different API versions separately in these functions, because it actually only handles one API version.

ocket8888 (changes requested):
The handlers for API versions <4 need to also check that more Permissions than the user has aren't being given to a Role, otherwise you could escalate your privileges by just using an older API version.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+		api.HandleErr(w, r, tx, http.StatusBadRequest, errors.New("no ID exists for the supplied role name"), nil)
+		return
+	} else if !ok {
+		api.HandleErr(w, r, tx, http.StatusNotFound, errors.New("no such role"), nil)
+		return
+	}
+
+	existingLastUpdated, found, err := api.GetLastUpdated(inf.Tx, roleID, "role")
+	if err == nil && found == false {
+		api.HandleErr(w, r, tx, http.StatusNotFound, errors.New("no role found with that ID"), nil)
```
ocket8888, line 358: The user didn't specify an ID - this error message should probably be the same as the above error for when the Role cannot be found. Not that I'm particularly attached to that wording either (personally I'd probably do something like `fmt.Errorf("no such Role: %s", roleName)`), but they should probably match.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+		return
+	}
+	rows, err := tx.Query(updateRoleQuery(), roleDesc, roleName)
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("updating role: %w", err))
+		return
+	}
+	defer rows.Close()
+	if !rows.Next() {
+		api.HandleErr(w, r, tx, http.StatusNotFound, errors.New("no role found with this ID"), nil)
```
ocket8888, line 377: same as above RE: error messages including ID when ID wasn't specified and matching other messages for the same error

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+		return
+	}
+	lastUpdated := tc.TimeNoMod{}
+	for rows.Next() {
+		if err := rows.Scan(&lastUpdated); err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("scanning lastUpdated from role update: %w", err))
+			return
+		}
+		roleV4.LastUpdated = &lastUpdated.Time
```
ocket8888, line 386: Instead of scanning into a `tc.TimeNoMod` and then storing a reference to a member struct thereof, why not just cut out the middle man and use a `time.Time`?

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+
+func createRoleCapabilityAssociations(tx *sqlx.Tx, roleID int, permissions *[]string) (error, error, int) {
+	result, err := tx.Exec(associateCapabilities(), roleID, pq.Array(permissions))
+	if err != nil {
+		return nil, fmt.Errorf("creating role capabilities: %w", err), http.StatusInternalServerError
+	}
+
+	if rows, err := result.RowsAffected(); err != nil {
+		return nil, fmt.Errorf("could not check result after inserting role_capability relations: %w", err), http.StatusInternalServerError
+	} else if expected := len(*permissions); int(rows) != expected {
```
ocket8888, line 453: This should cast the `len` to an `int64` instead of `rows` to an `int` to avoid overflow errors from the (potential) loss of precision.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+		return fmt.Errorf("wrong number of role_capability rows created: %d expected: %d", rows, expected), nil, http.StatusBadRequest
```
ocket8888, line 454: This should be an internal server error, because that's what actually happened; there's nothing about the user's request that they can change to make the request succeed. Also client-facing errors should never use database table names, as that's an implementation detail of the API - but that doesn't matter for system errors, which this should be, so no need to change the actual message

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+}
+
+func deleteRoleCapabilityAssociations(tx *sqlx.Tx, roleID int) (error, error, int) {
+	result, err := tx.Exec(deleteAssociatedCapabilities(), roleID)
+	if err != nil {
+		return nil, fmt.Errorf("deleting role capabilities: %w", err), http.StatusInternalServerError
+	}
+	if _, err := result.RowsAffected(); err != nil {
+		return nil, fmt.Errorf("could not check result after inserting role_capability relations: %w", err), http.StatusInternalServerError
+	}
```
ocket8888, line 466: but actually it doesn't really seem to be checking the resulting rows affected, since that value is thrown away to `_` - should it be doing something with the real information returned by this call?

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+		return
+	}
+	roleID, ok, err = dbhelpers.GetRoleIDFromName(tx, roleName)
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, errors.New("no ID exists for the supplied role name"), nil)
+		return
+	} else if !ok {
+		api.HandleErr(w, r, tx, http.StatusNotFound, errors.New("no such role"), nil)
+		return
+	}
```
ocket8888, line 495: same as above from the `PUT` handler that does the same thing

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+
+	assignedUsers := 0
+	if err := inf.Tx.Get(&assignedUsers, "SELECT COUNT(id) FROM tm_user WHERE role=$1", roleID); err != nil {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("role delete counting assigned users: "+err.Error()))
```
ocket8888, line 499: errors constructed from errors should wrap those underlying errors

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+		return
+	} else if assignedUsers != 0 {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, fmt.Errorf("can not delete a role with %d assigned users", assignedUsers), nil)
+		return
+	}
+
+	rows, err := tx.Query(deleteRoleQuery(), roleID)
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("deleting role: "+err.Error()))
```
ocket8888, line 508: same as above RE: error-wrapping

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+		return
+	}
+	defer inf.Close()
+
+	tx := inf.Tx.Tx
+	if roleName, ok = inf.Params["name"]; !ok {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, errors.New("must supply a role name to delete"), nil)
+		return
+	}
+	roleID, ok, err = dbhelpers.GetRoleIDFromName(tx, roleName)
```
ocket8888, line 488: I'm fairly sure the Role's ID isn't actually needed in the Go code anywhere, it's only used in queries - where it could just be fetched like: `... WHERE role = (SELECT id FROM public.role WHERE "name"=$1)`. You don't need to change it, but it gets rid of an extra DB query and the two error-handling blocks that follow.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("deleting role: "+err.Error()))
+		return
+	}
+	rows.Close()
+	userErr, sysErr, errCode = deleteRoleCapabilityAssociations(inf.Tx, roleID)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, tx, errCode, userErr, sysErr)
+		return
+	}
```
ocket8888, line 516: If this is actually working already, it means that the DELETE is cascading to the Permission associations already, because it would violate the foreign key constraints on that "join table" for a row to exist with a Role ID that doesn't exist - seems like we don't need this whole block.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+	if err := roleV4.Validate(); err != nil {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, err, nil)
+		return
+	}
+	current, err := auth.GetCurrentUser(r.Context())
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, err)
+		return
+	}
```
ocket8888, line 552: same as above RE: using the `inf.User` property

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+	defer rows.Close()
+
+	for rows.Next() {
+		var throwaway interface{}
+		if err := rows.Scan(&roleID, &throwaway); err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("role create: scanning role ID: %w", err))
+			return
+		}
```
ocket8888, line 575: `createQuery` is only called in one place, because legacy handlers are using `insertQuery` instead. Therefore, there's no reason to have a throwaway here; the query can just return only and exactly what it needs to - which should include the value for the response's `lastUpdated` field, by the way.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+	api.CreateChangeLogRawTx(api.ApiChange, changeLogMsg, inf.User, tx)
+}
+
+// Create will create a new role based on the struct supplied.
+func Create(w http.ResponseWriter, r *http.Request) {
+	var roleID int
+	var roleName string
+	var roleDesc string
+	var privLevel int
+	var roleCapabilities *[]string
```
ocket8888, line 529: this being a pointer to a slice is only causing unnecessary referencing/dereferencing and `nil` checks - it should just be immediately set to the value of the requested Permissions when they're available.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+	inf, userErr, sysErr, errCode := api.NewInfo(r, nil, nil)
+	if userErr != nil || sysErr != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, errCode, userErr, sysErr)
+		return
+	}
+	defer inf.Close()
+
+	tx := inf.Tx.Tx
+	params := make(map[string]dbhelpers.WhereColumnInfo, 0)
+	params["name"] = dbhelpers.WhereColumnInfo{Column: "name"}
```
ocket8888, line 614: nit but since we know at east one entry is going in that map, using `0` in the `make` is just forcing at least one re-allocation.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +301,369 @@ RETURNING id, last_updated`
...
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("querying Roles: %w", err))
+		return
+	}
+	defer log.Close(rows, "reading in Roles from the database")
+
+	var roleV4 tc.RoleV4
+	rolesV4 := []tc.RoleV4{}
+
+	for rows.Next() {
+		throwAway := new(interface{})
```
ocket8888, line 653: There's no reason we need a throwaway here, since the query used for this request is only used in one place it can just return only what it needs to return for the response.

traffic_ops/traffic_ops_golang/role/roles.go
srijeet0406: Actually, it shouldn't matter if the affected rows is 0 or more than 0, hence, I think we don't need this check anymore at all.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1604,6 +1667,45 @@ func GetDSIDFromStaticDNSEntry(tx *sql.Tx, staticDNSEntryID int) (int, error) {
...
+}
+
+// GetRoleIDFromName returns the ID of the role associated with the supplied name.
+func GetRoleIDFromName(tx *sql.Tx, roleName string) (int, bool, error) {
+	var id int
+	if err := tx.QueryRow(`SELECT id FROM role WHERE name = $1`, roleName).Scan(&id); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return id, false, nil
+		}
+		return id, false, errors.New("querying role ID from name: " + err.Error())
```
ocket8888, line 1692: Errors constructed from errors should wrap

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1604,6 +1667,45 @@ func GetDSIDFromStaticDNSEntry(tx *sql.Tx, staticDNSEntryID int) (int, error) {
...
+}
+
+// GetRoleNameFromID returns the name of the role associated with the supplied ID.
+func GetRoleNameFromID(tx *sql.Tx, roleID int) (string, bool, error) {
+	var name string
+	if err := tx.QueryRow(`SELECT name FROM role WHERE id = $1`, roleID).Scan(&name); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return name, false, nil
+		}
+		return name, false, errors.New("querying role name from ID: " + err.Error())
```
ocket8888, line 1704: Same as above RE: error-wrapping

traffic_ops/traffic_ops_golang/login/register.go
```diff
@@ -159,43 +167,74 @@ func RegisterUser(w http.ResponseWriter, r *http.Request) {
...
+			api.HandleErr(w, r, tx, errCode, userErr, nil)
+			return
+		}
+	} else {
+		req.Email = reqV4.Email
+		req.TenantID = reqV4.TenantID
+		dbhelpers.GetRoleIDFromName(tx, reqV4.Role)
+		roleID, ok, err := dbhelpers.GetRoleIDFromName(tx, reqV4.Role)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("no ID exists for the supplied role name"))
```
ocket8888, line 231: the error shouldn't be dropped, it should be logged to aid in fixing problems

traffic_ops/traffic_ops_golang/login/register.go
```diff
@@ -159,43 +167,74 @@ func RegisterUser(w http.ResponseWriter, r *http.Request) {
...
-	if privLevel > inf.User.PrivLevel {
-		userErr = errors.New("Cannot register a user with a role with higher privileges than yourself")
-		errCode = http.StatusForbidden
-		api.HandleErr(w, r, tx, errCode, userErr, nil)
-		return
+	// ToDo: Add checks for permission based role checking here, if the version is >=5 and the config option is turned on.
+	if inf.Version.Major < 4 {
+		privLevel, ok, err := dbhelpers.GetPrivLevelFromRoleID(tx, int(req.Role))
+		if err != nil {
+			sysErr = fmt.Errorf("Checking role #%d privilege level: %v", req.Role, err)
```
ocket8888, line 208: constructing an error from an error should wrap

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -273,3 +300,332 @@ RETURNING id, last_updated`
...
+		return
+	}
+	missing := inf.User.MissingPermissions(roleV4.Permissions...)
+	if len(missing) != 0 {
+		api.HandleErr(w, r, tx, http.StatusForbidden, fmt.Errorf("cannot request more than assigned permissions, current user needs %s permissions", strings.Join(missing, ",")), nil)
+		return
+	}
+	roleDesc = roleV4.Description
+	roleCapabilities = roleV4.Permissions
+	if roleName, ok = inf.Params["name"]; !ok {
```
ocket8888, line 337: if a string appears in the "required Parameters" argument of `api.NewInfo`, it'll do this check for you.

So instead of
```go
inf, userErr, sysErr, errCode := api.NewInfo(r, nil, nil)
```
doing
```go
inf, userErr, sysErr, errCode := api.NewInfo(r, []string{"name"}, nil)
```
will guarantee that `inf.Params` contains `"name"` as a key (or the proper error(s)/response code will be returned from `NewInfo`).

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -508,3 +578,483 @@ func (user *TOUser) InsertQuery() string {
...
+	*userV4.LocalPassword, err = auth.DerivePassword(*userV4.LocalPassword)
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusBadRequest, err, nil)
+		return
+	}
+
+	var resultRows *sqlx.Rows
+	_, ok, err := dbhelpers.GetRoleIDFromName(tx, userV4.Role)
+	if err != nil {
+		api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, errors.New("no ID exists for the supplied role name"))
```
ocket8888, line 840: this shouldn't swallow the error

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -508,3 +578,483 @@ func (user *TOUser) InsertQuery() string {
...
+	var lastUpdated time.Time
+	var tenant string
+	var rolename string
+	var changeLogMsg string
+
+	rowsAffected := 0
+	for resultRows.Next() {
+		rowsAffected++
+		if err = resultRows.Scan(&id, &lastUpdated, &tenant, &rolename); err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("could not scan after insert: %s\n)", err))
```
ocket8888, line 864: error strings should not include newlines

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -508,3 +578,483 @@ func (user *TOUser) InsertQuery() string {
...
+	userV4.LastUpdated = lastUpdated
+	userV4.Tenant = &tenant
+	userV4.Role = rolename
+	userV4.LocalPassword = nil
+
+	userResponse := tc.UserResponseV4{
+		Response: userV4,
+		Alerts:   tc.CreateAlerts(tc.SuccessLevel, "user was created."),
+	}
+	api.WriteAlertsObj(w, r, http.StatusOK, userResponse.Alerts, userResponse.Response)
```
ocket8888, line 887: status code for object creation should be `200 OK`, ideally with an HTTP `Location` header set to a URI that can be used to retrieve a representation of the created object.

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -508,3 +578,483 @@ func (user *TOUser) InsertQuery() string {
...
+}
+
+func validateUserV4(user tc.UserV4) error {
+	validateErrs := validation.Errors{
+		"email":    validation.Validate(user.Email, validation.Required, is.Email),
+		"fullName": validation.Validate(user.FullName, validation.Required),
+		"role":     validation.Validate(user.Role, validation.Required),
+		"username": validation.Validate(user.Username, validation.Required),
+		"tenantID": validation.Validate(user.TenantID, validation.Required),
+	}
```
ocket8888, line 778: At some point during creation or update of users, there must be a check that the Role is not changed to one with greater permissions than the requesting user. Old API versions also do `privCheck`, but should also be doing the permissions check (if Permissions are enabled in the configuration? or just always?) so that privileges cannot be escalated by just using older API versions.

ocket8888 (changes requested):

traffic_ops/traffic_ops_golang/user/user.go
srijeet0406: Do you mean a `201 Created`?

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -884,7 +899,8 @@ func Create(w http.ResponseWriter, r *http.Request) {
 		Response: userV4,
 		Alerts:   tc.CreateAlerts(tc.SuccessLevel, "user was created."),
 	}
-	api.WriteAlertsObj(w, r, http.StatusOK, userResponse.Alerts, userResponse.Response)
+	w.Header().Set("Location", fmt.Sprintf("/api/4.0/users?id=%d", *userV4.ID))
```
ocket8888, line 902: For future-proofing, this should probably just interpolate the `inf.Version`. I'm not sure off the top of my head if there's a `Version.String` or if that's just sitting in one of my branches somewhere, but if there is you probably wanna make use of it instead of using `inf.Version.Major` and `inf.Version.Minor` separately.

traffic_ops/traffic_ops_golang/user/user.go
```diff
@@ -899,7 +899,8 @@ func Create(w http.ResponseWriter, r *http.Request) {
 		Response: userV4,
 		Alerts:   tc.CreateAlerts(tc.SuccessLevel, "user was created."),
 	}
-	w.Header().Set("Location", fmt.Sprintf("/api/4.0/users?id=%d", *userV4.ID))
+	version := fmt.Sprintf("%s.%s", strconv.FormatUint(inf.Version.Major, 10), strconv.FormatUint(inf.Version.Minor, 10))
```
ocket8888, line 902: Since you're using `fmt.Sprintf` already, there's no need to bring `strconv` into it. Just use `%d` instead of `%s`.

ocket8888:
TP Role creation form appears to be broken: 
image

srijeet0406:
> TP Role creation form appears to be broken: image
> 
> The controls are all broken functionally, as well as visually

Should be fixed now.

Issue #6127, opened 2021-08-19
Rolling back `2021080408053529_fix_indices` in TO DB tests: `pys_location_last_updated_idx" already exists`


zrhoffman:
## This Bug Report affects these Traffic Control components:

- Traffic Ops

## Current behavior:

If run using a database dump, the database tests fail to roll back the `2021080408053529_fix_indices` migration:

```
+ ./db/admin --env=production down
Error running migrate down: migration failed: relation "pys_location_last_updated_idx" already exists in line 0: /*
```

Line it fails at:
https://github.com/apache/trafficcontrol/blob/308db857ab624bfd7f4c9c9d642fd22e8872efa9/traffic_ops/app/db/migrations/2021080408053529_fix_indices.down.sql#L18

Somehow, an index named `pys_location_last_updated_idx` still exists after running

https://github.com/apache/trafficcontrol/blob/308db857ab624bfd7f4c9c9d642fd22e8872efa9/traffic_ops/app/db/migrations/2021080408053529_fix_indices.up.sql#L18

I cannot reproduce this issue outside of the DB tests, but this is not a [Migrate](https://github.com/golang-migrate/migrate)-specific issue. If you run the DB tests with Goose at [`3cc78ea~`](https://github.com/apache/trafficcontrol/commits/3cc78ea53a~/traffic_ops/app/db/migrations) but add the Goose equivalent of `2021080408053529_fix_indices`:

```sql
-- +goose Up
-- SQL in section 'Up' is executed when this migration is applied

ALTER INDEX pys_location_last_updated_idx RENAME TO phys_location_last_updated_idx;
DROP INDEX topology_last_updated_idx;
CREATE INDEX topology_last_updated_idx ON topology (last_updated DESC NULLS LAST);

-- +goose Down
-- SQL section 'Down' is executed when this migration is rolled back

ALTER INDEX phys_location_last_updated_idx RENAME TO pys_location_last_updated_idx;
DROP INDEX topology_last_updated_idx;
CREATE INDEX topology_last_updated_idx ON topology_cachegroup (last_updated DESC NULLS LAST);
```

then the TO DB tests will still fail.

```
+ ./db/admin --env=production down
Running goose down...
goose: migrating db environment 'production', current version: 2021080408053529, target: 2021070800000000
2021/08/19 18:52:11 FAIL 2021080408053529_fix_indices.sql (pq: relation "pys_location_last_updated_idx" already exists), quitting migration.
Can't run goose: exit status 1
```

[...]

PR #6096, opened 2021-08-10
`db/admin migrate`: Migrate first migration explicitly if DB version's migration is squashed

zrhoffman:
This PR fixes an issue encountered migrating the database if a migration for the current DB migration version does not exist, typically because a migration at that version has been squashed into `create_table.sql`:

```
trafficops-db-admin_1  | error: no migration found for version 2021012200000000: read down for version 2021012200000000 : file does not existError running migrate up: no migration found for version 2021012200000000: read down for version 2021012200000000 : file does not exist
trafficops-db-admin_1  | DB upgrade failed!
```

-----------------------------------------------

## Which Traffic Control components are affected by this PR?

- Documentation
- Traffic Ops

## What is the best way to verify this PR?

1. Migrate the DB using Goose at [`d96c13c~`](https://github.com/apache/trafficcontrol/commit/d96c13cb40~) 
2. Dump the database and move the dump to `traffic_ops_db/test/docker/initdb.d`
3. Run the Traffic Ops DB tests

## If this is a bugfix, which Traffic Control versions contained the bug?

- master branch only

## PR submission checklist
- [x] Existing tests are sufficient 
- [x] This PR has documentation 
- [ ] This PR has a CHANGELOG.md entry 
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://apache.org/security) for details)

PR #6128, opened 2021-08-20
Skip check for the past dbversion's migration existing if it is squashed

zrhoffman:
This PR
- Fixes an issue introduced by #6096 where, where when migrating from Goose, `db/admin dbversion` would return `0` instead of Goose migration version (e080781e61)
- Fixes the issue that #6096 was supposed to fix: Running the first unsquashed migration without checking whether the past migration exists if migrating from the last squashed migration (f102c654ff).
- Fixes a bug in the Traffic Ops DB tests where `dropdb` did not specify the database to drop

<hr>

- Traffic Ops

1. Delete the `2021080408053529_fix_indices` migrations (unnecessary once #6127 is fixed)
2. Migrate the Traffic Ops database at [ATC 5.1.2](https://github.com/apache/trafficcontrol/commits/RELEASE-5.1.2)
3. Dump the database, move the dump to `traffic_ops_db/test/docker/initdb.d`, and rebuild the `db` service
4. Build a Traffic Ops RPM, move it to `traffic_ops_db/test/docker/traffic_ops.rpm` and rebuild the `trafficops-db-admin` service
5. Run the Traffic Ops DB tests

- 6.0.x release branch

## PR submission checklist

### DIFF
--- a/traffic_ops_db/test/docker/Dockerfile-db-admin
+++ b/traffic_ops_db/test/docker/Dockerfile-db-admin
@@ -23,25 +23,43 @@ FROM centos:7
 ARG POSTGRES_VERSION=13.2
 ENV POSTGRES_VERSION=$POSTGRES_VERSION
 
-# NOTE: temporary workaround for removal of golang packages from CentOS 7 base repo
 RUN yum install -y \
-    epel-release \
-    centos-release-scl-rh \
-    https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
+		epel-release \
+		centos-release-scl-rh \
+		https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm \
     git && \
-    yum -y install golang
+    yum -y install           \
+        cpanminus            \
+        expat-devel          \
+        libcap               \
+        libcurl-devel        \
+        libidn-devel         \
+        libpcap-devel        \
+        mkisofs              \
+        openssl-devel        \
+        perl-core            \
+        perl-Crypt-ScryptKDF \
+        perl-DBD-Pg          \
+        perl-DBI             \
+        perl-Digest-SHA1     \
+        perl-JSON            \
+        perl-libwww-perl     \
+        perl-TermReadKey     \
+        perl-Test-CPAN-Meta  \
+        perl-WWW-Curl        \
+        postgresql13         \
+        postgresql13-devel &&\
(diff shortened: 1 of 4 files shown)


zrhoffman:
Rebased to get the fix for #6127. Removing the `2021080408053529_fix_indices` migration is no longer necessary to test this PR.

traffic_ops_db/test/docker/Dockerfile-db-admin
```diff
@@ -23,25 +23,43 @@ FROM centos:7
...
 
 # Override TRAFFIC_OPS_RPM arg to use a different one using --build-arg TRAFFIC_OPS_RPM=...  Can be local file or http://...
 ARG TRAFFIC_OPS_RPM=traffic_ops.rpm
 ADD $TRAFFIC_OPS_RPM /
-RUN yum install -y \
-        /$(basename $TRAFFIC_OPS_RPM) && \
-        rm /$(basename $TRAFFIC_OPS_RPM) && \
-    yum clean all
+RUN rpm -Uvh $(basename $TRAFFIC_OPS_RPM) && \
+    rm $(basename $TRAFFIC_OPS_RPM)
```
rawlinp, line 58: Just for my own edification -- why was this change necessary?

rawlinp:
Uh oh, looks like the DB tests will fail perpetually until they're fixed to ignore the migration that was edited.

zrhoffman:
> Uh oh, looks like the DB tests will fail perpetually until they're fixed to ignore the migration that was edited.

cf59b77ede: No longer failing if that migration is not changed in the PR

traffic_ops_db/test/docker/Dockerfile-db-admin
zrhoffman: Installing the TO RPM dependencies in a layer before installing the TO RPM is just an optimization to allow a user to rebuild the `trafficops-db-admin` image with different TO RPMs without needing to download all the Yum packages every time.

rawlinp: Oh ok so `rpm` will still install the required dependencies even if they change in the RPM but aren't updated in the packages above?

zrhoffman: It will not, but if those dependencies are not satisfied when `rpm` runs, `rpm` will fail.

rawlinp: Ok, good to know. So it will be obvious when the TO RPM dependencies change that this list will need to be updated as well. Thanks!

rawlinp (approved):

PR #6135, opened 2021-08-24
Endpoint Required Permissions

ocket8888:
This PR adds a default middleware to all routes that checks if an authenticated user has the Permissions required when a route requires Permissions. This allows routes in the `routes.go` file to specify what Permissions, if any, they require to be used at all by a client, just like how PrivLevel is currently specified. Right now, no routes require any Permissions, which reduces the middleware to a "no-op".

Permissions are only checked if the configuration's `role_based_permissions` property is set to `true`, in which case PrivLevels are now ignored and only Permissions are checked.

<hr>

- Traffic Ops

I feel the tests are fairly comprehensive, but to manually test what I did was add a route definition for `GET /test-ping` in API version 4.0 to routes.go, requiring the non-existent Permission "foo". The route definition simply wrote `"success
n"` to the response. I then compiled and ran Traffic Ops  with `role_based_permissions` set to `true` and - after authenticating - attempted to send a GET request to `/api/4.0/test-ping`, which failed with an error about missing the Permission `foo`.

## PR submission checklist
- [x] This PR has tests


ocket8888:
I'd just like to say that since I added a field to the `Route` struct that ~2000 lines changed is a bit misleading - about 1400 of those were from the route definitions alone

traffic_ops/traffic_ops_golang/auth/authorize.go
```diff
@@ -42,6 +42,26 @@ type CurrentUser struct {
 	TenantID     int            `json:"tenantId" db:"tenant_id"`
 	Role         int            `json:"role" db:"role"`
 	Capabilities pq.StringArray `json:"capabilities" db:"capabilities"`
+	perms        map[string]struct{}
```
shamrickus, line 45: Any reason `perms` is lowercase?

ocket8888: yes, so that it isn't exported. That way no code outside of the auth package can mutate it, so nothing can accidentally give a user more Permissions than they actually have.

PR #6144, opened 2021-08-26
Document endpoint perms for role based permissions

shamrickus:
This PR adds comments to each v4 route denoting the required permission. For each permission they are of the form `{permission-name}:{verb}` where `verb` is one of `READ`, `CREATE`, `UPDATE`, or `DELETE`.

The current list of new permissions is:
`ACME`, `API-Capability`, `ASN`, `Async-Status`, `Cache-Group`, `Capability`, `CDN`, `CDN-Security-Key`, `CDN-Snapshot`, `Consistent-Hash`, `Coordinates`, `DBDump`, `Division`, `DS`, `DSR`, `DS-Required-Capability`, `DSS`, `DS-Security-Key`, `Federation`, `ISO`, `Jobs`, `Logs`, `Origin`, `Parameter`, `Physical-Location`, `Profile`, `Profile-Parameter`, `Region`, `Role`, `Server`, `Server-Capability`, `Server-Capability-Server`, `Server-Check`, `Servers`, `Service-Category`, `Static-DNS-Entry`, `Stats`, `Status`, `Steering`, `Tenant`, `Topology`, `Traffic-Vault`, `Type`, `User`,

<hr>

- Documentation

Look at the routes in the docs for v4 and compare that to the previous Roles & Capabilities endpoint security to ensure that
the new permissions are _at least_ as descriptive as before. Also confirm that the new permissions make sense and are sufficient for the endpoint they are attached to.

## PR submission checklist

### DIFF
--- a/docs/source/api/v4/statuses_id.rst
+++ b/docs/source/api/v4/statuses_id.rst
@@ -98,7 +98,7 @@ Response Structure
 
 ``PUT``
 =======
-Updates a Server's :term:`Status`.
+Updates a :term:`Status`.
 
 :Auth. Required: Yes
 :Roles Required: None


docs/source/api/v4/cdn_locks.rst
```diff
@@ -71,6 +72,7 @@ Allows user to acquire a lock on a CDN.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: CDNLOCK:CREATE
```
ocket8888, line 75: Should things like this also require `CDN-READ` (or equivalent)? Like, a user is allowed to delete a CDN that they aren't allowed to know exists. But the information is exposed by being able to manipulate the objects anyway, and then if e.g. in this case, if the CDN is already locked and it returns an error like "CDN 'CDN-Name' is already locked" then should it check for Permissions before displaying the CDN name in the error message?

shamrickus: By similar routes do you mean those not having the read permission on an endpoint that deletes/updates/creates?

ocket8888: Yeah, or related things like locks, capability associations, and snapshots

shamrickus: I'll explicitly require the `READ` perm on the 'same' route and directly related routes.

I'm torn on `snapshots`, it would pretty much need read permission for everything which doesn't seem optimal; hence why I have the permissions specifically for snapshots.

ocket8888: That's fair, I don't have any better ideas for Snapshots either.

shamrickus: I've updated the routes. Each route requires read permissions for the _required_ fields.

docs/source/api/v4/server_server_capabilities.rst
```diff
@@ -25,6 +25,7 @@ Gets all associations of :term:`Server Capabilities` to :term:`cache servers`.
 
 :Auth. Required: Yes
 :Roles Required: None
+:Permissions Required: SERVER-CAPABILITY-SERVER:READ, SERVER:READ, SERVER-CAPABILITY:READ
```
ocket8888, line 28: idk if this needs a separate Permission set, these are really just a server field that got put in a separate endpoint for some reason.

docs/source/api/v4/servers.rst
```diff
@@ -236,6 +237,7 @@ Allows a user to create a new server.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: SERVER:CREATE, CACHE-GROUP:UPDATE, DELIVERY-SERVICE:UPDATE, CDN:UPDATE, SERVER:READ, DELIVERY-SERVICE:READ, CDN:READ, PHYSICAL-LOCATION:READ, CACHE-GROUP:READ, TYPE:READ, PROFILE:READ
```
ocket8888, line 240: Does this need `DELIVERY-SERVICE:UPDATE`, `CACHE-GROUP:UPDATE`, and `CDN-UPDATE`?

docs/source/api/v4/servers_id.rst
```diff
@@ -25,6 +25,7 @@ Allow user to edit a server.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: SERVER:UPDATE, SERVER:READ, CACHE-GROUP:UPDATE, DELIVERY-SERVICE:UPDATE, CDN:UPDATE, SERVER:READ, DELIVERY-SERVICE:READ, CDN:READ, PHYSICAL-LOCATION:READ, CACHE-GROUP:READ, TYPE:READ, PROFILE:READ
```
ocket8888, line 28: Same as above RE: `DELIVERY-SERVICE:UPDATE`, `CACHE-GROUP:UPDATE`, and `CDN:UPDATE`.

docs/source/api/v4/servers_id.rst
```diff
@@ -326,6 +327,7 @@ Allow user to delete server through api.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: SERVER:DELETE, SERVER:READ, CACHE-GROUP:UPDATE, DELIVERY-SERVICE:UPDATE, CDN:UPDATE, SERVER:READ, DELIVERY-SERVICE:READ, CDN:READ, PHYSICAL-LOCATION:READ, CACHE-GROUP:READ, TYPE:READ, PROFILE:READ
```
ocket8888, line 330: Same as above RE: `DELIVERY-SERVICE:UPDATE`, `CACHE-GROUP:UPDATE`, and `CDN:UPDATE`.

docs/source/api/v4/servers_id_deliveryservices.rst
```diff
@@ -25,6 +25,7 @@ Retrieves all :term:`Delivery Services` assigned to a specific server.
 
 :Auth. Required: Yes
 :Roles Required: None\ [#tenancy]_
+:Permissions Required: DS-SERVER:READ, DELIVERY-SERVICE:READ, SERVER:READ
```
ocket8888, line 28: Same as above RE: `DS-SERVER:*`

docs/source/api/v4/servers_id_deliveryservices.rst
```diff
@@ -256,6 +257,7 @@ Assign an arbitrary number of :term:`Delivery Services` to a single server.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: DS-SERVER:CREATE, DS-SERVER:READ, DELIVERY-SERVICE:READ, SERVER:READ
```
ocket8888, line 260: Same as above RE: `DS-SERVER:*`

docs/source/api/v4/service_categories_name.rst
```diff
@@ -0,0 +1,133 @@
...
+Update a service category.
+
+:Auth. Required: Yes
+:Roles Required: "admin" or "operations"
+:Permissions Required: SERVICE-CATEGORY:UPDATE, SERVICE-CATEGORY:READ
+:Response Type:  Object
+
+Request Structure
+-----------------
+:name:        This :term:`Service Category`'s name
```
ocket8888, line 33: This should be in a Path Parameters table, field lists are for request (and response) bodies

docs/source/api/v4/service_categories_name.rst
```diff
@@ -0,0 +1,133 @@
...
+
+:Auth. Required: Yes
+:Roles Required: "admin" or "operations"
+:Permissions Required: SERVICE-CATEGORY:DELETE, SERVICE-CATEGROY:READ
+:Response Type:  ``undefined``
+
+
+Request Structure
+-----------------
+
```
ocket8888, line 95: There's a Path Parameter for this method of this endpoint that should be documented.

docs/source/api/v4/staticdnsentries.rst
```diff
@@ -25,6 +25,7 @@ Retrieve all static DNS entries configured within Traffic Control
 
 :Auth. Required: Yes
 :Roles Required: None
+:Permissions Required: STATIC-DNS:READ
```
ocket8888, line 28: Should this require `DELIVERY-SERVICE:READ` and/or `CACHE-GROUP:READ`?

docs/source/api/v4/staticdnsentries.rst
```diff
@@ -129,6 +130,7 @@ Creates a new, static DNS entry.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: STATIC-DNS:CREATE, STATIC-DNS:READ
```
ocket8888, line 133: Should this require `DELIVERY-SERVICE:READ` and/or `CACHE-GROUP:READ`?

docs/source/api/v4/staticdnsentries.rst
```diff
@@ -324,6 +326,7 @@ Delete staticdnsentries.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: STATIC-DNS:DELETE, STATIC-DNS:READ
```
ocket8888, line 329: Should this require `DELIVERY-SERVICE:READ` and/or `CACHE-GROUP:READ`?

docs/source/api/v4/statuses.rst
```diff
@@ -94,3 +95,65 @@ Response Structure
...
+:Auth. Required: Yes
+:Roles Required: None
+:Permissions Required: STATUS:CREATE, STATUS:READ
+:Response Type:  Array
+
+Request Structure
+-----------------
+:description:	Return only :term:`Statuses` with this *exact* description                                           |
+:id:			Return only the :term:`Status` with this integral, unique identifier                                 |
+:name:			Return only :term:`Statuses` with this name                                                          |
```
ocket8888, line 112: danging <kbd>|</kbd>

docs/source/api/v4/statuses.rst
```diff
@@ -94,3 +95,65 @@ Response Structure
...
+
+:Auth. Required: Yes
+:Roles Required: None
+:Permissions Required: STATUS:CREATE, STATUS:READ
+:Response Type:  Array
+
+Request Structure
+-----------------
+:description:	Return only :term:`Statuses` with this *exact* description                                           |
+:id:			Return only the :term:`Status` with this integral, unique identifier                                 |
```
ocket8888, line 111: ID cannot appear in a POST since it's not known at the time of creation.

docs/source/api/v4/statuses_id.rst
```diff
@@ -0,0 +1,159 @@
...
+
+:Auth. Required: Yes
+:Roles Required: None
+:Permissions Required: STATUS:UPDATE, STATUS:READ
+:Response Type:  Array
+
+Request Structure
+-----------------
+:description:	Return only :term:`Statuses` with this *exact* description                                           |
+:id:			Return only the :term:`Status` with this integral, unique identifier                                 |
```
ocket8888, line 111: ID is a path parameter that should be in a Path Parameters table, not the body field list.

docs/source/api/v4/statuses_id.rst
```diff
@@ -0,0 +1,159 @@
...
+:name:			Return only :term:`Statuses` with this name                                                          |
```
ocket8888, line 112: dangling <kbd>|</kbd>

docs/source/api/v4/steering_id_targets_targetID.rst
```diff
@@ -105,6 +106,7 @@ Removes a specific target mapping from a specific :term:`Delivery Service`
 
 :Auth. Required: Yes
 :Roles Required: Portal, Steering, Federation, "operations" or "admin"
+:Permissions Required: STEERING:DELETE, STEERING:READ, DELIVERY-SERVICE:READ, TYPE:READ, DELIVERY-SERVICE:UPDATE
```
ocket8888, line 109: I don't think we should require `DELIVERY-SERVICE:UPDATE` here, just because there is a desire for users without the ability to directly manage Delivery Services to be able to manipulate steering assignments, which would be made impossible by requiring that Permission.

docs/source/api/v4/topologies.rst
```diff
@@ -152,6 +153,7 @@ Create a new :term:`Topology`.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: TOPOLOGY:CREATE, TOPOLOGY:READ, CACHE-GROUP:READ, CACHE-GROUP:UPDATE
```
ocket8888, line 156: Does this need `CACHE-GROUP:UPDATE`?

docs/source/api/v4/topologies.rst
```diff
@@ -540,6 +542,7 @@ Deletes a specific :term:`Topology`.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: TOPOLOGY:DELETE, TOPOLOGY:READ, CACHE-GROUP:READ, CACHE-GROUP:UPDATE
```
ocket8888, line 545: Does this need `CACHE-GROUP:UPDATE`?

docs/source/api/v4/user_login.rst
```diff
@@ -25,6 +25,7 @@ Authentication of a user using username and password. Traffic Ops will send back
 
 :Auth. Required: No
 :Roles Required: None
+:Permissions Required: None
```
ocket8888, line 28: Currently we have undocumented special handling for a Role with the exact name "disallowed" that prevents users from logging in. That's because some organizations are not allowed to delete user data (and also to preserve foreign keys on e.g. changelogs and inactive DS Requests), so to prevent old accounts from being used after their owner no longer has need for them, we need to be able to disallow login for some users. We could add an `AUTH` Permission to fill that role, or add a flag to Roles or Users to define the concept. What do you think?

docs/source/api/v4/user_login_oauth.rst
```diff
@@ -25,6 +25,7 @@ Authentication of a user by exchanging a code for an encrypted JSON Web Token fr
 
 :Auth. Required: No
 :Roles Required: None
+:Permissions Required: None
```
ocket8888, line 28: Same as above RE: auth

docs/source/api/v4/user_login_token.rst
```diff
@@ -25,6 +25,7 @@ Authentication of a user using a token. Normally, the token is obtained via a ca
 
 :Auth. Required: No
 :Roles Required: None
+:Permissions Required: None
```
ocket8888, line 28: Same as above RE: auth

docs/source/api/v4/user_reset_password.rst
```diff
@@ -25,6 +25,7 @@ Sends an email to reset a user's password.
 
 :Auth. Required: No
 :Roles Required: None
+:Permissions Required: None
```
ocket8888, line 28: Same as above RE: auth (maybe?)

docs/source/api/v4/servers_id_status.rst
```diff
@@ -25,6 +25,7 @@ Updates server status and queues updates on all descendant :term:`Topology` node
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: SERVER:UPDATE, SERVER:READ
```
ocket8888, line 28: Does this need `STATUS:READ`?

docs/source/api/v4/cdns_name_configs_monitoring.rst
```diff
@@ -27,6 +27,7 @@ Retrieves information concerning the monitoring configuration for a specific CDN
 
 :Auth. Required: Yes
 :Roles Required: None
+:Permissions Required: CDN:READ, SERVER:READ, TYPE:READ, PROFILE:READ
```
ocket8888, line 30: seems like this endpoint has the same problems as snapshots, so you may just want to employ the same solution where you just give it its own Permission set and throw up your hands about all the data therein exposed.

ocket8888 (commented):
Regarding plurality: I've noticed mostly things are singular, but some aren't, like `DS-REQUESTS`. I think we ought to consistently pick plural or not. 

Of note, I think this mix has potentially already caused an error of sorts: I'm noticing both `FEDERATION:*` and `FEDERATIONS:*` Permissions exist, and also both `PARAMETER:*` and `PARAMETERS:*`.

Also, instead of `PROFILE-PARAMETER:CREATE` (and `:UPDATE` and `:DELETE`) what do you think of just requiring `PROFILE:UPDATE` for manipulating the Parameters of a Profile? Then `PROFILE:READ` and `PARAMETER:READ` together would encompass being able to see the Parameters of a Profile, which makes total sense to me since they're returned from a GET request to `/profiles` already. Which also means that perhaps custom logic will need to be implemented to not return Parameters if the requesting user doesn't have `PARAMETER:READ`. But that's a problem for a later date.

Also, should update queueing for a Server be a different Permission than `SERVER:UPDATE`? seems like a much lower bar, so that could be reflected by the Permissions.

docs/source/api/v4/cdns_name_health.rst
```diff
@@ -25,6 +25,7 @@ Retrieves the health of all :term:`Cache Groups` for a given CDN.
 
 :Auth. Required: Yes
 :Roles Required: None
+:Permissions Required: CDN:READ, CACHE-GROUP:READ
```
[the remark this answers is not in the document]
shamrickus: `STAT` is based on `stats_summary`, I actually think they should be removed from all endpoints besides `stats_summary`/`deliveryserivce_stats`.

docs/source/api/v4/deliveryservice_request_comments.rst
```diff
@@ -109,6 +110,7 @@ Allows user to create a :term:`Delivery Service Request` comment.
 
 :Auth. Required: Yes
 :Roles Required: "admin", "Federation", "operations", "Portal", or "Steering"
+:Permissions Required: DS-REQUESTS:CREATE, DELIVERY-SERVICE:READ
```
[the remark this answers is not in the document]
shamrickus: I was trying to avoid making a perm for comments, `UPDATE` makes more sense here.

docs/source/api/v4/deliveryservice_requests_id_assign.rst
```diff
@@ -75,6 +76,7 @@ The response is the username of the user to whom the :term:`Delivery Service Req
 =======
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: DS-REQUESTS:UPDATE, DS-REQUESTS:READ
```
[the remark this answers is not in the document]
shamrickus: I think `federations/{{id}}/users` should use `USER:READ`, only tenancy should require `UPDATE` IMO

docs/source/api/v4/isos.rst
```diff
@@ -26,6 +26,7 @@ Generates an ISO from the requested ISO source.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
+:Permissions Required: ISO:CREATE, ISO:READ
```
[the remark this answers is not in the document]
shamrickus: Fair point, I was trying to stick to the four verbs, but I think I don't think that's valuable (especially given `queuing`).

docs/source/api/v4/service_categories_name.rst
shamrickus: It would need both.

docs/source/api/v4/user_login.rst
shamrickus: Why don't we just have a `locked` column on the user table or something of that sort? This does not seem like an issue we should be tackling with permissions.

PR #6148, opened 2021-08-30
Remove bower from TP

shamrickus:
This PR removes bower and migrates to npm. A few libraries are named differently and/or have different paths. jQuery was updated to `2.1.4` (from `2.0.0`) as `2.1.x` is the first version listed on npm. The release notes can be seen [here](https://blog.jquery.com/2014/01/24/jquery-1-11-and-2-1-released/).
<hr>

- Documentation
- Traffic Portal
- CDN in a Box
- Automation (GHA)

Make sure the GHA pass, TP can build both locally and through make. Run the TP tests and ensure they still work. Make sure TP still functions and returns no console errors.
Verify the TP documentation is correct & makes sense.
Confirm that bower is completely removed from the repo and is not installed in any form.
 

## PR submission checklist
- [x] This PR has tests 


traffic_portal/package.json
```diff
@@ -50,12 +50,12 @@
...
-    "bootstrap-sass": "^3.3.6",
-    "es5-shim": "^4.5.6",
-    "flot": "^2.1.6",
-    "font-awesome": "^4.5.0",
-    "jquery": "^2.1.4",
-    "jquery.flot.tooltip": "^0.9.0",
-    "json3": "^3.3.2"
+    "bootstrap-sass": "3.3.6",
+    "es5-shim": "4.5.6",
+    "flot": "2.1.6",
```
zrhoffman: This downgrades `flot` from `2.3.2`, since `^` can refer to future minor or patch versions. The exact version should refer to what was previously the calculated `^` version to keep it from downgrading. So, `flot` should be `2.3.2`.

CHANGELOG.md
```diff
@@ -3,7 +3,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
-## [unreleased]
+## unreleased
+### Changed
+* Migrated completely off of bower in favor of npm
+
+## [6.0.x](https://github.com/apache/trafficcontrol/tree/6.0.x) - branched
```
zrhoffman, line 10: Put it in the current *Changed* changelog section, I'll take care of the 6.0.0 part.

zrhoffman (changes requested):

traffic_portal/package.json
shamrickus: Good catch, however when I added flot to `package.json` I should've used exact versioning to follow what we do everywhere else (or at least only allow patch release updates). I'll upgrade it to `2.3.2` since that's what `package-lock` was installing anyways.

zrhoffman (approved):
Verified that the packages Traffic Portal files are the same before and after this PR, other than timestamps and explicitly binding `window.$` and `window.jQuery`, which is necessary for the jQuery update.

```diff
diff --git a/opt/traffic_portal/public/resources/assets/js/shared-libs.js b/opt/traffic_portal/public/resources/assets/js/shared-libs.js
index c944474..4153bf7 100644
--- a/opt/traffic_portal/public/resources/assets/js/shared-libs.js
+++ b/opt/traffic_portal/public/resources/assets/js/shared-libs.js
@@ -44,7 +44,7 @@ require('ui-bootstrap-tpls');
 require('contextMenu');
 
 // jquery
-require('jquery');
+window.$ = window.jQuery = require('jquery');
 
 // flot charts
```

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
-## [unreleased]
+## unreleased
```
zrhoffman, line 6: Missed this, the square brackets should not be removed.

Issue #5674, opened 2021-03-24
Add query parameter option to Traffic Ops GET /jobs API to return only jobs created during maxRevalDurationDays window


rawlinp:
## I'm submitting a ...
-  improvement request (usability, performance, tech debt, etc.)

## Traffic Control components affected ...
-  Traffic Ops
-  Traffic Ops ORT

## Current behavior:
The Traffic Ops `GET /api/x/jobs` API currently returns all invalidation jobs ever created. For applications like ORT that really only need to get _unexpired_ invalidation jobs, this is very inefficient.

## New behavior:
Provide a query parameter option to return only invalidation jobs created within the `maxRevalDurationDays` window. Update ORT to use the new query parameter option. For performance reasons, we might want to add a DB index on the `start_time` column w/ `DESC NULLS LAST`, similar to the `last_updated` index.

## Minimal reproduction of the problem with instructions:
Request the TO `GET /api/x/jobs` route, observe that every job ever created is returned.

PR #6154, opened 2021-09-01
Add support for cdn and maxRevalDurationDays qparams to GET /jobs

rawlinp:
For the `GET /jobs` API, add support for the following query parameters:
- `cdn` to return jobs for delivery services in the given CDN with this name
- `maxRevalDurationDays` to return jobs with a start_time that is
  greater than now() - $maxRevalDurationDays (a parameter in the GLOBAL
  profile which controls how long any created invalidation job can last)

Related: #5674

The intention is to use these new query parameters in requests made by `t3c` in order to reduce load on TODB when processing invalidation requests.

<hr>

- Documentation
- Traffic Ops

Build the documentation, ensure it builds successfully, and view the rendered documentation. Review the added TO API tests and ensure it passes in the GitHub Actions CI.

## PR submission checklist
- [x] This PR has tests 


ocket8888:
does it need to be CDN _ID_ ? Could we do name instead?

rawlinp:
Normally I'd be with you on that, but in this case since I'm trying to reduce load on TODB when invalidations are submitted, using the CDN ID means I don't have to add an extra `JOIN` to the query.

ocket8888:
Well you shouldn't need to do a full JOIN, just a `column IN (SELECT FIRST(id) FROM cdn WHERE name = $n)` statement would work, which I don't think would cause the same load. but I could be wrong.

rawlinp:
How strongly do you feel it should be name instead of ID? I agree your query would probably work but it does slightly complicate the query even more when just using the ID is fine.

rawlinp:
I should add, `t3c` already has access to both the cdn name and ID from its `GET /servers` request, so it's not like it's also having to `GET /cdns` in order to get the ID for a given name.

ocket8888:
Well, what's your concern with the query being too complicated: is it that the query that _runs_ will be too complicated, or that the query as _written/implemented_  will me too complicated? If it's the former then maybe we could do both?

ocket8888:
In that case, you don't actually need to do it here, I suppose. If I really care I could add it myself, but I don't want to do that if you think it's a bad idea outright.

rawlinp:
Not a big deal, I'd rather do one than both, and even though it slightly complicates the query building (due to not being able to make use of the `WhereColumnInfo` stuff without making the query 4x more inefficient), the resulting performance is nearly the same as just using `cdnId`.

docs/source/api/v2/jobs.rst
```diff
@@ -31,23 +31,27 @@ Request Structure
...
+	| id                   | no       | Return only the single invalidation job identified by this integral, unique identifer                                                                            |
+	+----------------------+----------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+	| keyword              | no       | Return only invalidation jobs that have this "keyword" - only "PURGE" should exist                                                                               |
+	+----------------------+----------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+	| maxRevalDurationDays | no       | Return only invalidation jobs with a start_time that is within the window defined by the ``maxRevalDurationDays`` :term:`Parameter` in :ref:`the-global-profile` |
```
ocket8888, line 51: The field returned by the API is `startTime` - `start_time` is an internal DB column name with which consumers of the API need not concern themselves.

traffic_ops/testing/api/v4/jobs_test.go
```diff
@@ -482,6 +482,62 @@ func GetTestJobsByValidData(t *testing.T) {
 	} else {
 		t.Error("No user response available for get user by name")
 	}
+
+	// get maxRevalDurationDays parameter value
+	opts = client.NewRequestOptions()
+	opts.QueryParameters.Set("name", "maxRevalDurationDays")
+	params, _, err := TOSession.GetParameters(opts)
+	if err != nil {
+		t.Fatalf("unexpected error getting maxRevalDurationDays parameter")
```
ocket8888, line 491: This should print the error that occurred

traffic_ops/testing/api/v4/jobs_test.go
```diff
@@ -482,6 +482,62 @@ func GetTestJobsByValidData(t *testing.T) {
...
+	opts = client.NewRequestOptions()
+	opts.QueryParameters.Set("maxRevalDurationDays", "")
+	maxRevalJobs, _, err := TOSession.GetInvalidationJobs(opts)
+	if err != nil {
+		t.Errorf("unexpected error getting jobs by maxRevalDurationDays: %v", err)
+	} else if len(maxRevalJobs.Response) < 1 {
+		t.Errorf("GET /jobs by maxRevalDurationDays - expected at least 1 job")
+	}
+	for _, j := range maxRevalJobs.Response {
+		if time.Since((*j.StartTime).Time) > time.Duration(maxRevalDurationDays)*24*time.Hour {
```
ocket8888, line 511: This line will segfault if the API returns a malformed response with a null or undefined `startTime`

traffic_ops/testing/api/v4/jobs_test.go
```diff
@@ -482,6 +482,62 @@ func GetTestJobsByValidData(t *testing.T) {
...
+	}
+
+	// create DS xml_id -> cdn_id lookup map
+	dses, _, err := TOSession.GetDeliveryServices(client.NewRequestOptions())
+	if err != nil {
+		t.Fatalf("unexpectd error getting delivery services: %v", err)
+	}
+	dsToCDN := make(map[string]string, len(dses.Response))
+	for _, ds := range dses.Response {
+		dsToCDN[*ds.XMLID] = *ds.CDNName
```
ocket8888, line 523: This line will segfault if the API returns a malformed response with null or undefined `xmlId` and/or `cdnName`

traffic_ops/testing/api/v4/jobs_test.go
```diff
@@ -482,6 +482,62 @@ func GetTestJobsByValidData(t *testing.T) {
...
+	opts.QueryParameters.Set("cdn", cdn)
+	cdnJobs, _, err := TOSession.GetInvalidationJobs(opts)
+	if err != nil {
+		t.Errorf("unexpected error getting jobs by cdn: %v", err)
+	} else if len(cdnJobs.Response) < 1 {
+		t.Errorf("GET /jobs by cdn - expected at least 1 job")
+	}
+	for _, j := range cdnJobs.Response {
+		if dsToCDN[*j.DeliveryService] != cdn {
+			t.Errorf("GET /jobs by cdn returned job that does not belong to CDN %s: {%s, %s, %v}", cdn, *j.DeliveryService, *j.AssetURL, *j.StartTime)
```
ocket8888, line 538: These lines will segfault if the API returns a malformed response that has missing or null fields

rawlinp (commented):
Woops, forgot to submit my review comments with that last commit. Here they are :)

ocket8888:
I would argue that stack traces obfuscate test failures. Both by being harder to sift through to find out where the problem occurred and by crashing the test suite so that any further errors - even in unrelated tests - aren't reported.

ocket8888:
re-running client tests - appears to be that flaky CDN SSL keys test

ocket8888:
t3c testing error looks like it was having trouble connecting to downloads.postgresql.org - I restarted that one too, hopefully it was a transient issue that's passed us by

rawlinp:
> I would argue that stack traces obfuscate test failures. Both by being harder to sift through to find out where the problem occurred and by crashing the test suite so that any further errors - even in unrelated tests - aren't reported.

Sure, it might be more obfuscated, but it still tells you the exact line that it occurs on via the stacktrace. So in the rare case it actually happens, it would be pretty obvious what needs to be fixed (the API returning null values where they're expected to be non-null).

ocket8888:
Yeah, that's true, it tells _you_ exactly what line the error came from. Tools have issues though. If a test fails, I can go straight to the failing line with <kbd>F4</kbd>, but when a segfault occurs I have to go dig through the output by hand.

It's not like it's going to cause tests to fail or pass that otherwise wouldn't have, though, so if you feel strongly about it, I'll just add the checks if/when I run into the segfaults. Or anyone else could, if they wanted to volunteer.

rawlinp:
So in the off-chance that someone inadvertently breaks the API to where it no longer returns required fields (which is _highly_ unlikely IMO), adding in hundreds of lines across the entire TO API test suite saves you a few seconds to find the failing line? I don't think that is a good trade-off for the readability that is lost with all the unnecessary nil-checks.

ocket8888:
I don't disagree that the nil checks are obnoxious rough to read through. I just think the better solution is not use pointers for everything.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
```diff
@@ -228,10 +228,28 @@ func (job *InvalidationJob) Read(h http.Header, useIMS bool) ([]interface{}, err
 	if err != nil {
 		return nil, nil, fmt.Errorf("getting accessible tenants for user - %v", err), http.StatusInternalServerError, nil
 	}
+	cdn := ""
+	if cdnName, ok := job.APIInfo().Params["cdn"]; ok {
+		queryValues["cdn"] = cdnName
+		cdn = ` AND ds.cdn_id = (SELECT id FROM cdn WHERE name = :cdn) `
```
ocket8888, line 234: fwiw, I think this actually could be a `dbhelpers.WhereInfo` if you did it like:
```go
"cdn": dbhelpers.WhereColumnInfo{Column: `(SELECT cdn.name FROM public.cdn WHERE cdn.id=ds.cdn_id)`},
```
but I haven't tested that, and what you have works fine.

ocket8888 (approved):
Everything appears to be working. I misunderstood what the param was supposed to be doing at first, but I got it now.

The only thing I'll say is that it's a little weird that I can do `GET /api/4.0/jobs?maxRevalDurationDays=false` and it means the same thing as `GET /api/4.0/jobs?maxRevalDurationDays=true` - or even `GET /api/4.0/jobs?maxRevalDurationDays={{!!PINEAPPLES!!}}`.

traffic_ops/traffic_ops_golang/invalidationjobs/invalidationjobs.go
rawlinp: I tried that out at first, but the performance was atrocious. I think that subquery is run per row, whereas this version just runs it once to get the CDN ID.

ocket8888: Oh. That's interesting. Wonder if the others done that way in that query are having similar problems.

rawlinp: Possibly, but since `t3c` isn't using those filters I decided not to look into them.

PR #6157, opened 2021-09-01
Error wrapping in t3c

ocket8888:
This makes errors constructed in from other errors in t3c (everything under `cache-config/`) use error-wrapping formatting so that the identity of the underlying error is not destroyed. It also ensures that:

- Error strings do not start with a capital letter
- Error strings do not end with punctuation
- Error identity is checked with `errors.Is`
- Error type is checked with `errors.As`

Originally, I was going to do that for everything in the repo at once, but I drastically underestimated how large that changeset would be. So I just started with the first component alphabetically. Partially because I did originally want this to encompass the whole repo, there's also a change in here that that removes a newline from an error string in `traffic_vault_migrate` and some changes to the db/admin program - I left those in because I figured it was harmless to do so, but I can take them out at a reviewer's request.

<hr>

Make sure all the tests still pass. To check for non-wrapping errors, I grepped with the regular expression:
```regexp
(fmt\.Errorf|errors\.New)(.+(\.Error\b|%v.+ \w*[eE]r\w*[,)]|\\n|[!\.?]")|\("[A-Z])
```

which does output some false positives where an error is passed to `fmt.Errorf` that just so happens to use `%v` to format something that isn't an error. I also used `errorlint`, which is available through golangci-lint, to catch cases where errors.Is and errors.As should be used (but that won't catch all the things the regexp does, it'll miss things like `errors.New(err.Error())`). Finally, I checked for matches to

```regexp
errors\.New\(fmt\.Sprintf\(
```
but that yielded no results, so no changes were necessary.

## PR submission checklist


rob05c:
See my comment on https://github.com/apache/trafficcontrol/pull/6159#issuecomment-920952811 about performance.

`t3c` isn't nearly as performance-critical as Grove, but we're trying to reduce the runtime of cache. As we get down from minutes to seconds, it'd be good if we didn't add things that make the whole run take say, 5 seconds instead of 2.

Again, I support the idea of this. And I can't say how much an impact this has without testing. But it'd be really good if, before merging something potentially performance-impacting like this, if someone benchmarked `t3c-apply` runs for every concievable ordinary configuration on a large CDN, and confirmed it doesn't significantly impact performance.

PR #6158, opened 2021-09-02
Add T3C README-based docs to the documentation build

ocket8888:
This PR adds the docs for `t3c` - which are currently in its README.md files that get turned into manual pages on the fly - to the build system for our documentation. It does this by first making a makefile for the cache-config docs (which would be much simpler if we didn't care about putting the output into the same directory as the program - but that's how the current build system works so I decided to copy it) and then using that makefile in a sub-make in the docs makefile.

The makefile for `cache-config` is also capable of outputting manual pages.

This _does_ add a dependency on Pandoc to our docs build.

<hr>

- Documentation
- GItHub Actions
- unknown

Make sure it builds the documentation with no warnings, and check to make sure the t3c documentation is in it (under _Administration_).

## PR submission checklist
- [x] This PR has tests 


ocket8888:
docs build failure seems to be upstream: https://github.com/ammaraskar/sphinx-action/issues/33

I'll monkey with this more some other time.

PR #6163, opened 2021-09-02
t3c check for live service before adding to ramdisk

ezelkow1:
Add a continue in the case the DS is not a LIVE service, so that it does not get added to the ramdisk hosting list

<hr>

- Documentation

Add a DS without a toplogy that is not a LIVE service, verify its hosting.config does not place it on a ramdisk. Add it to a topology and ensure that it is still not on a ramdisk

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/lib/go-atscfg/hostingdotconfig.go
+++ b/lib/go-atscfg/hostingdotconfig.go
@@ -196,6 +196,9 @@ func MakeHostingDotConfig(
 					if !topoHasServer {
 						continue
 					}
+                                        if !strings.Contains(string(*ds.Type), "LIVE") {
+                                                continue
+                                        }
 				}
 			}
 
--- a/lib/go-atscfg/hostingdotconfig.go
+++ b/lib/go-atscfg/hostingdotconfig.go
@@ -196,9 +196,9 @@ func MakeHostingDotConfig(
 					if !topoHasServer {
 						continue
 					}
-                                        if !strings.Contains(string(*ds.Type), "LIVE") {
-                                                continue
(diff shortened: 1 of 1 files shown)


rawlinp:
I believe this code change is correct in that it prevents topology-based, non-live DSes from getting assigned to the RAM disk, but it looks like the unit tests reinforce the original, incorrect behavior.

For the record, I think we should totally simplify the logic for creating this file and just add a line for every live-type DS origin, irregardless of whether or not the DS is actually assigned to the server (via DSS or topology) and whether or not the DS is a mid or an edge. It's not "ideal" in that it adds lines for DSes that don't use that particular server, but I think the gains in simplicity of this code would be worth it. That, and it probably doesn't hurt ATS to have extra entries in there that don't actually get used.

ezelkow1:
Closing this one, further discussions with @rob05c that we can remove the toplogy != nil check before this occurs and just always do these type checks. So we will have a different PR up to remove that check

rob05c:
> For the record, I think we should totally simplify the logic for creating this file and just add a line for every live-type DS origin, regardless of whether or not the DS is actually assigned to the server (via DSS or topology) and whether or not the DS is a mid or an edge. It's not "ideal" in that it adds lines for DSes that don't use that particular server, but I think the gains in simplicity of this code would be worth it. That, and it probably doesn't hurt ATS to have extra entries in there that don't actually get used.

I would be +1 on that. I've suggested in the past that we do that for all configs, for safety; I was outvoted.

PR #6166, opened 2021-09-07
T3c hosting

ezelkow1:
Topologies are being skipped for live/non-live checks, so anything with a topology will land on a ramdisk. This adds that check back in

<hr>

Test without the change with a non-live DS assigned to a topology, should see it get added to the ATS hosting.config ram disk volume. run with the change and it should not be listed and fall into the default hosting

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/lib/go-atscfg/hostingdotconfig.go
+++ b/lib/go-atscfg/hostingdotconfig.go
@@ -135,22 +135,26 @@ func MakeHostingDotConfig(
 				continue
 			}
 
-			// mids: include all DSes with at least one server assigned
-			if len(dsServerMap[*ds.ID]) == 0 {
-				continue
+			if ds.Topology == nil {
+				// mids: include all DSes with at least one server assigned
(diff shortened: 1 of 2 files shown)


lib/go-atscfg/hostingdotconfig.go
```diff
@@ -196,6 +199,9 @@ func MakeHostingDotConfig(
 					if !topoHasServer {
 						continue
 					}
+					if !strings.Contains(string(*ds.Type), "LIVE") {
```
rob05c, line 202: Nitpick: better to use `https://github.com/apache/trafficcontrol/blob/master/lib/go-tc/enum.go#L870`

`if !ds.Type.IsLive()`

rob05c (approved):
LGTM, code looks good, manually tested, verified Topology DSes with LIVE are still put in hosting.config, and non-LIVE aren't.

Issue #6034, opened 2021-07-15
Add support for cdn query parameter filter to GET /deliveryserviceserver route


rawlinp:
## I'm submitting a ...
-  improvement request (usability, performance, tech debt, etc.)

## Traffic Control components affected ...
-  Traffic Ops
-  Traffic Ops ORT

## Current behavior:
The `GET /deliveryserviceserver` route does not support filtering by CDN.

## New behavior:
The `GET /deliveryserviceserver` route should have support for a `?cdn=my-cdn-name` query parameter to filter the response by CDN. Once added, we should make `t3c` use it to filter by CDN instead of requesting the entirety of ds-server assignments across all CDNs. `t3c` has no need for ds-server assignments for CDNs outside of its own CDN, so this leads to unnecessary network bandwidth usage and database load.

## Minimal reproduction of the problem with instructions:
`GET /deliveryserviceserver?cdn=my-cdn`, notice how it doesn't filter by the given CDN name.

PR #6167, opened 2021-09-07
Add support for "cdn" query parameter in GET /deliveryserviceserver API

rawlinp:
Allow filtering by CDN name via a new "cdn" query parameter. Also, allow clients to send an `orderby` query parameter with an empty value to prevent unnecessary sorting of the response (in order to reduce unnecessary load on TODB).

Related: #6034 

<hr>

- Documentation
- Traffic Ops

Review the added TO API test, ensure it passes in the GitHub Actions CI run.

## PR submission checklist
- [x] This PR has tests 


traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
```diff
@@ -191,9 +192,6 @@ func (dss *TODeliveryServiceServer) readDSS(h http.Header, tx *sqlx.Tx, user *au
 		}
 		offset *= limit
 	}
-	if orderby == "" {
-		orderby = "deliveryService"
-	}
```
ocket8888, line 196: Does this really result in significant performance gains? Because it makes the API less "functional", in the sense that two instances with the same data might return different responses for the same request. Or the same instance with unchanged data might return different responses for identical requests - though I think that's less likely.

rawlinp: It definitely makes a significant gain in `cost` as determined by `EXPLAIN ANALYZE ...`. Just selecting the `server` and `deliveryService` columns, I get `4311.89` _without_ ordering and `8031.62` _with_ ordering. That's about a 46% reduction in cost.

That said, I was kind of on the fence about removing this default ordering vs giving the client the ability to prevent any ordering -- e.g. `?orderby` without a value. IIRC the existing code would still give that a default ordering since it doesn't check if the param actually exists in the request URL. So I would either remove the default ordering or make it check if the param exists in the URL. If the param exists without a value, it wouldn't give it a default. What do you think of that?

rawlinp: Done.

ocket8888 (approved):
Code looks good, appears to be working based on test results and a couple of manual requests

PR #6169, opened 2021-09-07
Fix t3c update when fallback occurs

rob05c:
Fixed t3c-update when a fallback to a previous Traffic Ops API version occurs.

The bug was that when a fallback occurred, the "old" client was doing a Get instead of a Set.

This also adds a log message of exactly the client's latest version and the version being fallen back to, e.g. `INFO: client.go:112: 2021-09-07T19:56:59.82297Z: cache-config Traffic Ops client: 4.0 not supported, falling back to 3.1`, which should be helpful in debugging, and in manually testing this fix.

No automated tests, unfortunately this can't be reasonably unit-tested since it's the TO request function itself, and the t3c Integration Test Framework doesn't yet have the ability to use multiple TO versions and fall back between them, and that'd be a pretty big feature to add. FWIW I manually tested and verified it's fixed, and sets the update flags in TO appropriately on a TO without the latest major in the client.
Includes changelog.
No docs, no interface change.

Run t3c-update against a TO without the latest major version in the client, verify flags are set and unset as specified.

## PR submission checklist
- ~~[x] This PR has tests~~ explained above, can't be tested without a significant test framework enhancement


jrushford (approved):
looks good.  tested and it fixes the bug

PR #6178, opened 2021-09-08
fixes credential handling in the tm-health-client

jrushford:
Fixes some credential handling with the tm-health-client

- Changes the behavior at startup to allow the credentials specified in the json config file to be used.  If they are not defined in the json config file then the external "to-credential-file" defined in the json config file will be used.
- Fixes a crash that may occur when a session cannot be established with traffic ops.

Verify that the tm-health-client successfully starts with TO credentials defined in the tm-health-client.json config file and separately with a "tm-credential-file" specified in the tm-health-client.json file with no other credentials specified there.

Verify that the tm-health-client does not crash when invalid credentials are used.

## PR submission checklist
- [x] This PR has tests 
- [] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://apache.org/security) for details)

### DIFF
--- a/cache-config/tm-health-client/config/config.go
+++ b/cache-config/tm-health-client/config/config.go
@@ -23,6 +23,7 @@ import (
 	"bufio"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


smalenfant (approved):
Changes looks good. The only thing I wish we would do is using the environment like t3c-apply does. That could simply be an enhancement later.

cache-config/tm-health-client/config/config.go
```diff
@@ -196,8 +196,13 @@ func GetConfig() (Cfg, error, bool) {
 		return Cfg{}, errors.New(err.Error() + "\n"), false
 	}
 
-	if err = ReadCredentials(&cfg); err != nil {
-		return cfg, err, false
+	if cfg.TOPass == "" || cfg.TOUser == "" || cfg.TOUrl == "" {
+		if cfg.TOCredentialFile == "" {
+			return Cfg{}, errors.New("cannot continue, no TO credentials have been specified."), false
```
ocket8888, line 201: error strings shouldn't end with punctuation.

cache-config/tm-health-client/config/config.go
```diff
@@ -219,6 +224,9 @@ func GetTrafficMonitors(cfg *Cfg) error {
 
 	// login to traffic ops.
 	session, _, err := toclient.LoginWithAgent(cfg.TOUrl, cfg.TOUser, cfg.TOPass, true, "tm-health-client", false, GetRequestTimeout())
+	if err != nil {
+		return errors.New("could not establish a TrafficOps session: " + err.Error())
```
ocket8888, line 228: constructing errors from errors should use `fmt.Errorf("... %w ...", err)` instead of `errors.New("... " + err.Error() + " ...")` so that the identity of the original error is not destroyed

cache-config/tm-health-client/config/config.go
jrushford: made the change

cache-config/tm-health-client/config/config.go
jrushford: made the change

ocket8888 (approved):

PR #3175, opened 2019-01-02
WIP Add Delivery Service Snapshots

rob05c:
WIP - Adds Delivery Service Snapshots.

Still WIP, needs work, but I thought I'd make it available for public comment.

See https://cwiki.apache.org/confluence/display/TC/Traffic+Control++Self+Service+Proposal+for+Change+Integrity

This follows the plan, with one major exception: it creates new "snapshot" tables, which are copied into on snapshot. This is because Perl touches many of these tables, and the Perl can't be easily fixed.

The idea is, if we want DS snapshots now, we can implement them much sooner, without rewriting all of Perl, with these "snapshot" tables. Then, when Perl finally goes away, we can simply create a database migration to rename the "snapshot" tables.

IMO This shouldn't be merged without 304 Not Modified support, either. Because the CRConfig is now generated every time, it takes longer, and the Monitors are constantly polling, so this adds time to that poll and work to the TO/DB. It's pretty easy to add, IMO there's no legitimate reason to merge this without adding that support, to mitigate the cost increase.

*Specifically Includes*
- database migrations to create the "_snapshot" tables.
- functions to copy "real" data into snapshot tables
- a new API endpoint, `deliveryservices/{ds-name}/snapshot`
- changes CRConfig Snapshot function to just insert the timestamp, and copy into snapshot tables
- changes CRConfig Get Snapshot function to build the CRConfig on-the-fly, from the timestamp and snapshot tables
- changes CRConfig Get Live Snapshot to copy into snapshot tables, and query their latest values (reproducing existing live behavior)

*Things That Need To Happen Before This Can Be Merged*
- [ ] Mailing List Consensus
- [ ] Fix SQL injection vulnerabilities in `WithCDNSnapshotTimeReal` and `UpdateDSDSSSnapshot`
- [ ] Traffic Portal UI button to Snapshot ("Save") Delivery Service
- [ ] Change `monitoring.json` to also build on-the-fly from snapshot tables
- [ ] Parity Testing
- [ ] Decide how to deal with the next TO upgrade requiring all DSes be snapshotted
- [ ] Update DB migration dates
- [ ] Add 304 Not Modified Support

#### Which TC components are affected by this PR?

- [ ] Documentation
- [ ] Grove
- [ ] Traffic Analytics
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [ ] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
[...]

PR #6180, opened 2021-09-09
Delivery Service Snapshots (Blueprint)

ocket8888:
This PR adds a blueprint for Delivery Service Snapshots that would allow changes to be made to and deployed for a single Delivery Service without affecting any others.

Transcribed from the Confluence wiki "Self Service Change Integrity" page with some re-wording and updates for new features (mainly Topologies).

Disclaimer: This blueprint details a massive amount of changes to the Traffic Ops API and a body of work on `t3c`, the size of which I can't accurately guage. Some of the ideas sound good to me, some don't, and while I'm happy to update the blueprint as any conversation around it evolves, I am not the originator of this idea, so I may not be able to answer all questions about it.

<hr>

None

Read the blueprint.

## PR submission checklist


ocket8888:
related: #3175 (needs major updates but purports to implement this blueprint at least in part)

ocket8888:
After 3 years of no activity, I feel confident we are not interested in this idea.

Issue #6093, opened 2021-08-10
TR dnschallenges: exception when regexp doesn't match the xmlId


smalenfant:
## This Bug Report affects these Traffic Control components:

- Traffic Router

## Current behavior:

Exception when DS regexp doesn't match the xmlId. Certificate not created.

```
WARN  2021-08-10T15:37:49.750 [pool-2-thread-1] com.comcast.cdn.traffic_control.traffic_router.core.ds.LetsEncryptDnsChallengeWatcher - Failed updating dns challenge txt record with data from https://${toHostname}/api/2.0/letsencrypt/dnsrecords/:
java.lang.NullPointerException
	at com.comcast.cdn.traffic_control.traffic_router.core.ds.LetsEncryptDnsChallengeWatcher.updateStaticEntries(LetsEncryptDnsChallengeWatcher.java:152)
	at com.comcast.cdn.traffic_control.traffic_router.core.ds.LetsEncryptDnsChallengeWatcher.lambda$useData$0(LetsEncryptDnsChallengeWatcher.java:85)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
	at com.comcast.cdn.traffic_control.traffic_router.core.ds.LetsEncryptDnsChallengeWatcher.useData(LetsEncryptDnsChallengeWatcher.java:59)
	at com.comcast.cdn.traffic_control.traffic_router.core.util.AbstractResourceWatcher.loadDatabase(AbstractResourceWatcher.java:108)
	at com.comcast.cdn.traffic_control.traffic_router.core.loc.AbstractServiceUpdater.updateDatabase(AbstractServiceUpdater.java:117)
	at com.comcast.cdn.traffic_control.traffic_router.core.loc.AbstractServiceUpdater$1.run(AbstractServiceUpdater.java:86)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #6184, opened 2021-09-09
Le regex fix

mattjackson220:
Closes: #6093 

This PR updates the Let's Encrypt workflow to be able to complete the DNS challenges for DSes where the regex has been updated so that the domain no longer contains the xmlid.

<hr>

- Traffic Ops
- Traffic Router

Verify that the TR and TO tests pass
Verify that the normal Let's Encrypt workflow works as before
Create a DS with a regex that has the domain without the xmlId
Verify that the LE workflow works for this new DS as well

all of them since LE was introduced

## PR submission checklist
This does not have test or documentation updates since it is a bug fix

- [x] This PR has tests 


rawlinp (approved):
These changes look sound to me!

PR #6190, opened 2021-09-10
Fixing tm-health-client stopping trafficserver

smalenfant:
It was found during testing that the fix made to the tm-health-client systemd service file was stopping trafficserver. 

`systemctl stop tm-health-client` should not stop trafficserver. Although a bit more extensive testing might be required. 

- master

## PR submission checklist
- [] This PR has tests 
- [] This PR has documentation 


jrushford (approved):
Tests ok and trafficserver is no-longer stopped when stopping the tm-health-client

PR #6191, opened 2021-09-10
Traffic Portal Server Options

ocket8888:
This PR adds the ability to run Traffic Portal using an arbitrary configuration file via the new `-c` flag to its `server.js` main script, an environment variable to control passing said flag to `grunt`, and enables logging to `stdout` when logging is not configured in the configuration file.

This also adds documentation for Traffic Portal's main configuration file.

<hr>

- Documentation
- Traffic Portal

Make sure existing tests pass, test the new behavior by making a copy of `configDev.js` (or `config.js`) somewhere, changing some configuration (e.g. `sslPort`) and running `grunt` with `TP_SERVER_CONFIG_FILE` set to the location of the modified copy. If the server that's run has the modified settings, then you know it used the right configuration file, which is how I tested it.

## PR submission checklist


traffic_portal/server.js
```diff
@@ -28,19 +28,29 @@ var constants = require('constants'),
...
+    }
 }
 
-var logStream = fs.createWriteStream(config.log.stream, { flags: 'a' }),
-    useSSL = config.useSSL;
-
+const useSSL = config.useSSL;
+let logStream;
+if (!Object.prototype.hasOwnProperty.call(config, "log") || !config.log || !Object.prototype.hasOwnProperty.call(config.log, "stream") || !config.log.stream) {
+    logStream = null;
```
shamrickus, line 50: I think we would want this to halt execution in this state.

shamrickus (changes requested):

traffic_portal/server.js
ocket8888: No, that's intentional, and causes logging to go to stdout per
> ... enables logging to `stdout` when logging is not configured in the configuration file.

Previously it *would* halt execution because it would be trying to either access a property of a "falsey" value or trying to open a writestream to something that isn't a valid filename. That's no longer desired behavior, which is why I do the intentional checks here.

traffic_portal/package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "traffic_portal",
```
shamrickus, line 8: I don't understand why this wasn't added a long time ago, the name field in `package.json` has been set for like 4 years.

shamrickus (approved):
LGTM

PR #6200, opened 2021-09-13
t3c-apply report mode. Ignore update server flag

smalenfant:
This update makes a few things more inline with what it was before.

It's adding the `--ignore-update-flag` to `-m report`.
It is also updating the traffic_ops_ort wrapper to set the default verbosity of t3c-apply. It also enables the different logging mode.

Compare functionality to previous version. There is nothing much changed here.

- master
- 6.0.0

## PR submission checklist
- [] This PR has tests - Looking into the test
- [] This PR has documentation - Working on documentation
- [] This PR has a CHANGELOG.md entry - Doesn't warrant a changelog entry (new feature)

### DIFF
--- a/cache-config/traffic_ops_ort.pl
+++ b/cache-config/traffic_ops_ort.pl
@@ -16,6 +16,7 @@
 use strict;
 use warnings;
 use Getopt::Long;
+use feature qw(switch);
 
 my $dispersion = undef;
 my $retries = 5;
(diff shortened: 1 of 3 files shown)


cache-config/traffic_ops_ort.pl
```diff
@@ -102,6 +103,23 @@
...
+
+if ( defined( $ARGV[1] ) ) {
+	my $log_level = "";
+        my $ort_log_level = uc($ARGV[1]);
+	given ( $ort_log_level ) {
+                when ("ALL")   { $log_level = "-vv"; }
+                when ("TRACE") { $log_level = "-vv"; }
+                when ("DEBUG") { $log_level = "-vv"; }
+                when ("INFO")  { $log_level = "-v"; }
+                when ("WARN")  { $log_level = ""; }
```
rob05c, line 115: `-v` means warn, `-vv` means info. I think this line should be `-v` and the above should be `-vv`.

smalenfant: Ok. I can update that. Without tc3-apply working in report, it's hard to know what should be expected.

zrhoffman:
Superseded #6220

PR #6211, opened 2021-09-14
Fixed crs/stats fields to report time correctly and not overflow and …

mattjackson220:
…removed unused fields from TO

This PR fixes Traffic Router to correctly keep track of time in HTTP and DNS average stats, prevent overflow in the stats in crs/stats, and removes the unused fields from TO.

<hr>

- Traffic Ops
- Traffic Router

Verify that the `averageHttpTime` and `averageDnsTime` values at `crs/stats` look reasonable (previously the current epoch time in milliseconds was being added)
Verify that if the `totalHttpTime` or `totalDnsTime` variables get too large that they would overflow a `long` (will probably need to be manually adjusted to do this since now the time added in will be very small) that they instead reset to 0 and reset the corresponding count to 0 too.
Verify that the TR and TO tests pass

## PR submission checklist
This is a small bug fix so no tests or docs were updated
- [x] This PR has tests 

### DIFF
--- a/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/router/StatTracker.java
+++ b/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/router/StatTracker.java
@@ -368,6 +368,11 @@ public class StatTracker {
 			if (t.routeType == RouteType.DNS) {
 				totalDnsCount++;
 				totalDnsTime += t.time;
+				if (totalDnsTime < 0 || totalDnsCount < 0) {
+					this.resetDnsStatsFromOverflow();
+					totalDnsCount++;
+					totalDnsTime += t.time;
+				}
 				map = dnsMap;
 
 				if (t.resultDetails == Track.ResultDetails.LOCALIZED_DNS) {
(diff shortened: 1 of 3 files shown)


traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/router/StatTracker.java
```diff
@@ -452,11 +462,20 @@ public void initialize(final Map<String, List<String>> initMap, final CacheRegis
...
 	}
+
+	private void resetHttpStatsFromOverflow() {
+		totalHttpCount = 0;
+		totalHttpTime = 0;
+	}
+
+	private void resetDnsStatsFromOverflow() {
+    	totalDnsCount = 0;
+    	totalDnsTime = 0;
```
rawlinp, line 479: these lines are missing an indentation

rawlinp (changes requested):
This looks good to me from a code perspective -- it's just missing a little indentation.

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/router/StatTracker.java
rawlinp: actually, this indentation is using spaces instead of tabs, and it appears this is a tab-indented file

rawlinp (approved):
Looks good!

PR #6212, opened 2021-09-14
Increasing astats string buffer size

ezelkow1:
We have seen astats config ip lists going beyond the existing 1k limit and being truncated. There is a proper rewrite on the way but this temporary fix can suffice for a while

<hr>

- traffic server

Use astats with a very long list of allowable IPs, if you enable debug you can see at a certain point that parsing will be truncated. With the new version you can have a much larger list

## PR submission checklist
- [x] This PR has tests 


rob05c (approved):
LGTM. I haven't tested, but I don't see how it could cause any issues. It's used exclusively for the size of an array on the stack.

Issue #2564, opened 2018-07-19
Legacy Naming Conventions (Rascal, RIAK, CCR et. al.)


ocket8888:
It's super confusing that Traffic Monitor is referred to in hundreds of places as "Rascal", Traffic Vault as "RIAK" and Traffic Router as CCR. We should change those.

That would also entail removing legacy constraints related to those names, e.g. Traffic Monitor profiles currently must start with a "RASCAL" prefix for no good reason I've ever heard. Database tables and columns ought to be renamed as well (if applicable, I'm not sure), but that probably would make production upgrades much more complicated.

mitchell852:
"low effort" @ocket8888 ?

PR #6213, opened 2021-09-15
Remove Traffic Router Profile name restrictions

ocket8888:
This PR removes restrictions previously placed on the names of Profiles used by Traffic Routers. From the docs:

> For legacy reasons, the names of Profiles of this type must begin with `CCR_` or `TR_`. This is **not** enforced by the [Traffic Ops API](https://traffic-control-cdn.readthedocs.io/en/latest/api/index.html#to-api) or Traffic Portal, but certain Traffic Control operations/components expect this and will fail to work otherwise!

However, as can been seen in these changes, that's not even actually strictly true. `TR_` prefixes were not supported by the `/cdns/domains` Traffic Ops API endpoint.

After this PR, however, TO only checks for the correct Profile *type* (`TRAFFIC_ROUTER`), **not** pattern-matching on the name at all.

This is potentially breaking, but only in the extremely weird scenario where Traffic Router servers were using a Profile for some other type of server (since Profile types are immutable, it's not conceivable that someone would have e.g. `TRAFFIC_ROUTER_1` as a Profile type, so the Profile type must be some other valid type e.g. `TRAFFIC_PORTAL`). I believe Traffic Portal would disallow such an assignment anyway, but I'm not certain.

This addresses - but doesn't itself close, without #6218 - #2564.

<hr>

- Documentation
- Traffic Ops
- CDN in a Box

Make sure that all tests pass - existing tests do exercise these endpoints, which should prove that `CCR_` as a prefix still works. To prove that other things still work, CiaB helps by simply being able to run without the prefix. 

## PR submission checklist
- [x] This PR has tests, of a sort, by affecting CiaB which is itself adequately tested.

### DIFF
--- a/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
+++ b/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
@@ -375,7 +375,7 @@ WITH cdn_profile_ids AS (
     MAX(p.id) as profile_id -- We only want 1 profile, so get the probably-newest if there's more than one.
   FROM
     cdn c
-    LEFT JOIN profile p ON c.id = p.cdn AND (p.name like 'CCR%' OR p.name like 'TR%')
+    LEFT JOIN profile p ON c.id = p.cdn AND (p.type = 'TR_PROFILE')
     GROUP BY c.name, c.dnssec_enabled, c.domain_name
 )
 SELECT
--- a/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
+++ b/traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
@@ -375,7 +375,7 @@ WITH cdn_profile_ids AS (
     MAX(p.id) as profile_id -- We only want 1 profile, so get the probably-newest if there's more than one.
   FROM
     cdn c
-    LEFT JOIN profile p ON c.id = p.cdn AND (p.type = 'TR_PROFILE')
+    LEFT JOIN profile p ON c.id = p.cdn AND (p.type = '` + tc.TrafficRouterProfileType + `')
     GROUP BY c.name, c.dnssec_enabled, c.domain_name
 )
 SELECT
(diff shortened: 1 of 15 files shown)


traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
```diff
@@ -375,7 +375,7 @@ WITH cdn_profile_ids AS (
     MAX(p.id) as profile_id -- We only want 1 profile, so get the probably-newest if there's more than one.
   FROM
     cdn c
-    LEFT JOIN profile p ON c.id = p.cdn AND (p.name like 'CCR%' OR p.name like 'TR%')
+    LEFT JOIN profile p ON c.id = p.cdn AND (p.type = 'TR_PROFILE')
```
rawlinp, line 378: We should probably have a constant for `"TR_PROFILE"`

rawlinp (changes requested):

traffic_ops/traffic_ops_golang/cdn/dnssecrefresh.go
ocket8888: Probably. But there's also no constant for any other profile type - except `GROVE_PROFILE` which has a constant in the `github.com/apache/trafficcontrol/grove/grovetccfg/main` package, but that hardly counts since it's in `main` and therefore isn't exported and thus unusable anywhere else.

What I'm getting at is all of these should have constants, and while I'm prepared to do that it would make the PR a bit bigger. Not by much, since it turns out most of them aren't used anywhere (only `GROVE_PROFILE`, `ATS_PROFILE`, and `TR_PROFILE` are used in Go source files, that I could find). So if you want me to do that lmk and I will - I just don't feel good about leaving it half-done by only making one constant for TR_PROFILE.

rawlinp: Sure, I agree that it doesn't seem like there are many usages of other profile type strings, so I don't mind if you want to add that to this PR.

rawlinp (approved):

PR #6214, opened 2021-09-15
Handle blocked IPs as soon as possible

ezelkow1:
Add a status set to forbidden and event error to try to stop the SM from handling blocked IPs. With ATS 9 some of the loop handling has changed and now astats can trigger a loop detection because when it blocks an IP it just hands it back to the state machine, which then checks the destination which is the cache itself and triggers a loop detection. This code changes that so the status is forbidden and returns an HTTP error.

<hr>

- traffic server

Test a cache with ATS 9 that is blocked via astats, you may see an ERR_LOOP_DETECTED crc. Try again with the patch and it should be forbidden

## PR submission checklist
- [x] This PR has tests 


traeak (approved):
Tested with both ats81 and ats91, both pass and fail conditions.  Looks good.

Issue #5981, opened 2021-06-28
Traffic Stats uses unmaintained logging library


rawlinp:
## I'm submitting a ...
-  improvement request (usability, performance, tech debt, etc.)

## Traffic Control components affected ...
-  Traffic Stats

## Current behavior:
Traffic Stats currently uses this logging library which hasn't had a commit since Jan. 30, 2017: https://github.com/cihub/seelog https://github.com/apache/trafficcontrol/blob/dfdbfe921af874c17e8da4488624b49a91cd6244/traffic_stats/traffic_stats.go#L42

## New behavior:
Since all of our other Go components use the in-repo [lib/go-log](https://github.com/apache/trafficcontrol/tree/dfdbfe921af874c17e8da4488624b49a91cd6244/lib/go-log) library, it might make sense for TS to use that logging library as well.

PR #6217, opened 2021-09-16
Traffic Stats logging use lib/go-log

ocket8888:
This is related to #5981 but does not fix it.

This adds the ability to use a standard `lib/go-log`-based configuration for TS logging. It also deprecates the use of a seelog configuration file.

<hr>

- Documentation
- Traffic Stats

Play around with logging configurations, make sure logging is still taking place appropriately when configured, if the docs don't help figure out what configurations should yield what kind of logging then the docs changes are bad.

## PR submission checklist

### DIFF
--- a/traffic_stats/traffic_stats.go
+++ b/traffic_stats/traffic_stats.go
@@ -136,58 +136,58 @@ type Timers struct {
 
 func info(args ...interface{}) {
 	if useSeelog {
-		seelog.Info(args)
+		seelog.Info(args...)
 	} else {
-		log.Infoln(args)
+		log.Infoln(args...)
 	}
 }
 func infof(format string, args ...interface{}) {
 	if useSeelog {
-		seelog.Infof(format, args)
+		seelog.Infof(format, args...)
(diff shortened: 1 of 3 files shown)


traffic_stats/traffic_stats.go
```diff
@@ -131,24 +224,24 @@ func main() {
 	tickers = setTimers(config)
 
 	termChan := make(chan os.Signal, 1)
-	signal.Notify(termChan, syscall.SIGKILL, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	signal.Notify(termChan, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
```
shamrickus, line 227: Why'd you remove `SIGKILL`?

ocket8888: Programs can't actually catch SIGKILL, so it was unnecessary at best, misleading at worst.

shamrickus (approved):
LGTM

PR #6218, opened 2021-09-16
Remove outdated restrictions on Profile Names for specific Types

ocket8888:
The documentation incorrectly specifies name restrictions for some Profile types - they were true when they were written in February of 2019, but with the changes since then, these restrictions are no longer in place.

This addresses - but doesn't itself close, without #6213 - #2564.

<hr>

- Documentation

Make sure the docs are accurate and thorough, and build without warnings or errors.

All since about mid 2020

## PR submission checklist
- Documentation has no tests


srijeet0406 (approved):
Looks good.

PR #6232, opened 2021-09-22
Migrate type test from TP E2E to TP integration test

dpham692:
<hr>

- Traffic Portal
- CDN in a Box
- GitHub Action automation
- unknown

Make sure all tests are migrate in integration/Data/type.ts directory.
https://github.com/apache/trafficcontrol/blob/master/traffic_portal/test/e2e_deprecated/types/types-spec.js
Go to traffic_portal/test/integration and run
`npm test -- --params.baseUrl https://localhost --capabilities.shardTestFiles true` and make sure test all pass.
Make sure CI tests pass also.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_portal/test/integration/specs/Profiles.spec.ts
+++ b/traffic_portal/test/integration/specs/Profiles.spec.ts
@@ -94,12 +94,9 @@ profiles.tests.forEach(async profilesData => {
         });
     });
 });
-
 describe('Clean up API for Profiles', () => {
-    afterAll(async function () {
-        it('Cleanup', async () => {
-            await api.UseAPI(profiles.cleanup);
-        });
-    })
+    it('Cleanup', async () => {
+        await api.UseAPI(profiles.cleanup);
+    });
 });
 
--- a/traffic_portal/test/integration/specs/Types.spec.ts
+++ b/traffic_portal/test/integration/specs/Types.spec.ts
@@ -36,38 +36,49 @@ describe('Setup API for Types Test', () => {
 types.tests.forEach(async typesData => {
     typesData.logins.forEach(login => {
         describe(`Traffic Portal - Types - ${login.description}`, () => {
+            afterEach(async function () {
+                await typesPage.OpenTypesPage();
+            });
+            afterAll(async function () {
+                expect(await topNavigation.Logout()).toBe(true);
+            })
             it('can login', async () => {
                 browser.get(browser.params.baseUrl);
                 await loginPage.Login(login);
-                expect(await loginPage.CheckUserName(login)).toBeTruthy();
-            });
-            it('can open types page', async () => {
+                expect(await loginPage.CheckUserName(login)).toBe(true);
                 await typesPage.OpenConfigureMenu();
-                await typesPage.OpenTypesPage();
             });
-
+            typesData.toggle.forEach(toggle => {
+                it(toggle.description, async () => {
+                    if(toggle.description.includes('hide')){
+                        expect(await typesPage.ToggleTableColumn(toggle.Name)).toBe(false);
+                    }else{
+                        expect(await typesPage.ToggleTableColumn(toggle.Name)).toBe(true);
+                    }
+                    
+                });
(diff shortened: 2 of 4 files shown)


traffic_portal/test/integration/specs/Profiles.spec.ts
```diff
@@ -94,12 +94,9 @@ profiles.tests.forEach(async profilesData => {
...
-
 describe('Clean up API for Profiles', () => {
-    afterAll(async function () {
-        it('Cleanup', async () => {
-            await api.UseAPI(profiles.cleanup);
-        });
-    })
+    it('Cleanup', async () => {
+        await api.UseAPI(profiles.cleanup);
+    });
```
ocket8888, line 100: Was this file supposed to be changed?

traffic_portal/test/integration/specs/Types.spec.ts
```diff
@@ -36,39 +36,50 @@ describe('Setup API for Types Test', () => {
...
-                await typesPage.OpenTypesPage();
             });
-
+            typesData.toggle.forEach(toggle => {
+                it(toggle.description, async () => {
+                    if(toggle.description.includes('hide')){
+                        expect(await typesPage.ToggleTableColumn(toggle.Name)).toBe(true);
+                    }else{
+                        expect(await typesPage.ToggleTableColumn(toggle.Name)).toBe(true);
+                    }
```
ocket8888, line 57: This looks like it's doing the same thing regardless of which path is taken - is that check necessary?

traffic_portal/test/integration/specs/Profiles.spec.ts
dpham692: After couple test and research. The afterAll can only be use one in a spec file regardless how many describe you put it in. So with this afterAll, the clean up API will not work.

traffic_portal/test/integration/specs/Types.spec.ts
dpham692: Fixed the check. Apparently, the function check was not cap.

traffic_portal/test/integration/specs/Profiles.spec.ts
ocket8888: > After couple test and research. The afterAll can only be use one in a spec file regardless how many describe you put it in.

Where did you hear that? [The Jasmine `afterAll` docs](https://jasmine.github.io/api/3.9/global.html#afterAll) say:

> Run some shared teardown once after all of the specs in the `describe` are run.

So you should be able to have one per `describe`. After all, hasn't this been working for a while now?

dpham692: @ocket8888 I tested it. You can look at the old test packaged that the clean up hasn't been run for a while. Only 1 afterAll run per spec doesn't matter you put the second afterAll in another describe or outside other describe. The test working for a while because the cleanup for the profiles did not touch other test.

dpham692: You can check it in this PR for example: https://github.com/apache/trafficcontrol/pull/6233/checks?check_run_id=3676058657

ocket8888: Yeah, I wouldn't expect it to actually run. Like the docs say:

> Run some shared teardown once after all of the specs in the `describe` are run.

That `describe` contains no specs. The `afterAll` should be in the same `describe` as the specs _after all_ of which the `afterAll` block will run.

See in this demo:

https://jsfiddle.net/62pefumq/

After running it, there should be two debug-level logs in the console: "in afterAll 1" and "in afterAll 3". "in afterAll 2" doesn't get printed because its `describe` contains no specs.

dpham692: @ocket8888 Ok, got you. I fixed it.

ocket8888 (approved):
LGTM

PR #6235, opened 2021-09-24
Removed default limits values in logs API

### LINKED ISSUE #6228: GET logs?days=X limits results to 1000 rows max
- Traffic Ops
- Traffic Portal

## Current behavior:
GET logs?days=X where the `limit` query param is not defined limits results rows to 1000 rows max. This appears to be a regression as this is not how it behaved before. This affects the ability to pull change logs for X days thru the API (obviously) as well as thru TP (https://tp.domain.tld/#!/change-logs).

This appears to be a regression on API version 2.x, 3.x and 4.x.

## Expected behavior:
GET logs?days=X where limit query param is not defined does NOT limit the number of rows returned.

## Steps to reproduce:
Call the API endpoint or change the number of days on https://tp.domain.tld/#!/change-logs and observe that never are more than 1000 rows returned.

dmohan001c:

- [x] This PR fixes #6228   

This removed the logs api which has default limit value.
POST  /logs

- CDN in a Box
- Traffic Control Client 
- Traffic Ops
- CI tests

* Master

### DIFF
--- a/docs/source/api/v4/logs.rst
+++ b/docs/source/api/v4/logs.rst
@@ -37,7 +37,7 @@ Request Structure
 	+===========+==========+=====================================================================================================================================+
 	| days      | no       | An integer number of days of change logs to return                                                                                  |
 	+-----------+----------+-------------------------------------------------------------------------------------------------------------------------------------+
-	| limit     | no       | The number of records to which to limit the response, by default it will display all the response                                                                               |
+	| limit     | no       | The number of records to which to limit the response, by default there is no limit applied                                          |
 	+-----------+----------+-------------------------------------------------------------------------------------------------------------------------------------+
 	| offset    | no       | The number of results to skip before beginning to return results. Must use in conjunction with limit                                |
 	+-----------+----------+-------------------------------------------------------------------------------------------------------------------------------------+
--- a/docs/source/api/v4/logs.rst
+++ b/docs/source/api/v4/logs.rst
@@ -37,7 +37,7 @@ Request Structure
 	+===========+==========+=====================================================================================================================================+
 	| days      | no       | An integer number of days of change logs to return                                                                                  |
 	+-----------+----------+-------------------------------------------------------------------------------------------------------------------------------------+
-	| limit     | no       | The number of records to which to limit the response                                                                                |
+	| limit     | no       | The number of records to which to limit the response, by default it will display all the response                                                                               |
 	+-----------+----------+-------------------------------------------------------------------------------------------------------------------------------------+
 	| offset    | no       | The number of results to skip before beginning to return results. Must use in conjunction with limit                                |
 	+-----------+----------+-------------------------------------------------------------------------------------------------------------------------------------+
--- a/traffic_ops/traffic_ops_golang/logs/log.go
+++ b/traffic_ops/traffic_ops_golang/logs/log.go
@@ -63,19 +63,13 @@ func Get(w http.ResponseWriter, r *http.Request) {
 	if pLimit, ok := inf.IntParams["limit"]; ok {
 		limit = pLimit
 	}
-	a := tc.Alerts{}
 	setLastSeenCookie(w)
 	logs, count, err := getLog(inf, days, limit)
 	if err != nil {
-		a.AddNewAlert(tc.ErrorLevel, err.Error())
-		api.WriteAlerts(w, r, http.StatusInternalServerError, a)
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, err, err)
 		return
 	}
-	if a.HasAlerts() {
-		api.WriteAlertsObj(w, r, 200, a, logs)
-	} else {
-		api.WriteRespWithSummary(w, r, logs, count)
-	}
+	api.WriteRespWithSummary(w, r, logs, count)
 }
 
 // Get is the handler for GET requests to /logs V4.0.
@@ -177,7 +171,7 @@ func getLogV40(inf *api.APIInfo, days int) ([]tc.Log, uint64, error) {
 		return nil, 0, util.JoinErrs(errs)
 	}
 
-	timeInterval := fmt.Sprintf("l.last_updated > now() - INTERVAL '%v' DAY", days)
+	timeInterval := fmt.Sprintf("l.last_updated > now() - INTERVAL '%d' DAY", days)
 	if where != "" {
 		whereCount = ", tm_user as u\n" + where + " AND l.tm_user = u.id"
 		where = where + " AND " + timeInterval
--- a/traffic_ops/traffic_ops_golang/logs/log.go
+++ b/traffic_ops/traffic_ops_golang/logs/log.go
@@ -188,26 +188,26 @@ func getLogV40(inf *api.APIInfo, days int) ([]tc.Log, uint64, error) {
 	queryCount := countQuery + whereCount
 	rowCount, err := inf.Tx.NamedQuery(queryCount, queryValues)
 	if err != nil {
-		return nil, count, errors.New("querying log count for a given user: " + err.Error())
+		return nil, count, fmt.Errorf("querying log count for a given user: %w", err)
 	}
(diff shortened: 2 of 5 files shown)


ocket8888 (changes requested):
Looks like this isn't setting default limits in API versions earlier than 4.0:
```console
$ toget -k logs | jq '.response[]|.id' | wc -l
2151
$ toget -k 'logs?days=1' | jq '.response[]|.id' | wc -l                                                                                           
2151
$ toget -k logs | jq .summary.count
2151
```
This should also come with docs updates that make it clear that `limit` does, in fact, have a default value (that changes depending on whether or not `days` was given.

docs/source/api/v4/logs.rst
```diff
@@ -37,7 +37,7 @@ Request Structure
 	+===========+==========+=====================================================================================================================================+
 	| days      | no       | An integer number of days of change logs to return                                                                                  |
 	+-----------+----------+-------------------------------------------------------------------------------------------------------------------------------------+
-	| limit     | no       | The number of records to which to limit the response                                                                                |
+	| limit     | no       | The number of records to which to limit the response, by default it will display all the response                                                                               |
```
ocket8888, line 40: This table is now malformed, needs to be aligned - but also, you don't need to make mention of there not being a default, instead the older API versions should make it clear that there _is_ a default. One would expect that if a limit is not given, then there is no limit. The docs should just clarify when those expectations are not met.

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -79,6 +78,35 @@ func Get(w http.ResponseWriter, r *http.Request) {
...
+	a := tc.Alerts{}
+	setLastSeenCookie(w)
+	logs, count, err := getLogV40(inf, days)
+
+	if err != nil {
+		a.AddNewAlert(tc.ErrorLevel, err.Error())
+		api.WriteAlerts(w, r, http.StatusInternalServerError, a)
+		return
+	}
+	if a.HasAlerts() {
```
ocket8888, line 103: this will never be the case. The only code path that adds an alert writes it and then returns, so this will never evaluate to true

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -79,6 +78,35 @@ func Get(w http.ResponseWriter, r *http.Request) {
...
+		days = pDays
+	}
+
+	a := tc.Alerts{}
+	setLastSeenCookie(w)
+	logs, count, err := getLogV40(inf, days)
+
+	if err != nil {
+		a.AddNewAlert(tc.ErrorLevel, err.Error())
+		api.WriteAlerts(w, r, http.StatusInternalServerError, a)
```
ocket8888, line 100: this should probably just use `api.HandleErr` instead of crafting alerts in here

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -135,11 +163,64 @@ FROM "log" as l JOIN tm_user as u ON l.tm_user = u.id`
...
+		"username": {Column: "u.username", Checker: nil},
+	}
+	where, _, pagination, queryValues, errs :=
+		dbhelpers.BuildWhereAndOrderByAndPagination(inf.Params, queryParamsToQueryCols)
+
+	if len(errs) > 0 {
+		return nil, 0, util.JoinErrs(errs)
+	}
+
+	timeInterval := fmt.Sprintf("l.last_updated > now() - INTERVAL '%v' DAY", days)
```
ocket8888, line 180: nit but formatting that isn't meant to layout data structures (which should, itself, be rare) should not use `%v`, since a more specific formatting parameter would enforce a compile-time check on the type of the related argument

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -135,11 +163,64 @@ FROM "log" as l JOIN tm_user as u ON l.tm_user = u.id`
...
+		whereCount = ", tm_user as u\n" + where + " AND l.tm_user = u.id"
+		where = where + " AND " + timeInterval
+	} else {
+		whereCount = where
+		where = "\nWHERE " + timeInterval
+	}
+	queryCount := countQuery + whereCount
+	rowCount, err := inf.Tx.NamedQuery(queryCount, queryValues)
+	if err != nil {
+		return nil, count, errors.New("querying log count for a given user: " + err.Error())
```
ocket8888, line 191: errors created from errors should wrap using e.g. `fmt.Errorf("querying log count for a given user: %w", err)`

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -135,11 +163,64 @@ FROM "log" as l JOIN tm_user as u ON l.tm_user = u.id`
...
+	}
+	defer rowCount.Close()
+	for rowCount.Next() {
+		if err = rowCount.Scan(&count); err != nil {
+			return nil, count, errors.New("scanning logs: " + err.Error())
```
ocket8888, line 196: Same as above RE: error wrapping

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -135,11 +163,64 @@ FROM "log" as l JOIN tm_user as u ON l.tm_user = u.id`
...
+		}
+	}
+
+	query := selectFromQuery + where + "\n ORDER BY last_updated DESC" + pagination
+	rows, err := inf.Tx.NamedQuery(query, queryValues)
+	if err != nil {
+		return nil, count, errors.New("querying logs: " + err.Error())
```
ocket8888, line 203: Same as above RE: error wrapping

traffic_ops/traffic_ops_golang/logs/log.go
```diff
@@ -135,11 +163,64 @@ FROM "log" as l JOIN tm_user as u ON l.tm_user = u.id`
...
+	}
+	defer rows.Close()
+	ls := []tc.Log{}
+	for rows.Next() {
+		l := tc.Log{}
+		if err = rows.Scan(&l.ID, &l.Level, &l.Message, &l.User, &l.TicketNum, &l.LastUpdated); err != nil {
+			return nil, count, errors.New("scanning logs: " + err.Error())
```
ocket8888, line 210: Same as above RE: error wrapping

ocket8888 (changes requested):

docs/source/api/v4/logs.rst
dmohan001c: Fixed the malformed tables.

traffic_ops/traffic_ops_golang/logs/log.go
dmohan001c: Fixed

traffic_ops/traffic_ops_golang/logs/log.go
dmohan001c: Fixed.

traffic_ops/traffic_ops_golang/logs/log.go
dmohan001c: Fixed.

traffic_ops/traffic_ops_golang/logs/log.go
dmohan001c: Fixed.

traffic_ops/traffic_ops_golang/logs/log.go
dmohan001c: Fixed.

traffic_ops/traffic_ops_golang/logs/log.go
dmohan001c: Fixed.

traffic_ops/traffic_ops_golang/logs/log.go
dmohan001c: updated.

ocket8888 (approved):
This seems to be working

PR #6243, opened 2021-09-30
Add cdn.conf option to disable automatic cert deletion

rawlinp:
Some operators may not want to automatically delete certificates for
delivery services that no longer exist, in case they need to be
resurrected. Additionally, if a Traffic Vault backend is shared between
different environments, you don't want to inadvertently delete
certificates out of other environments.

<hr>

- Documentation
- Traffic Ops
- CDN in a Box
- Automation (GHA, ansible roles)

In your cdn.conf, set `"disable_auto_cert_deletion": true`, start TO, then take a CDN snapshot while tailing the TO log file. You should see the INFO log `automatic certificate deletion is disabled, returning without cleaning up old certificates` to indicate that the new config option is working as expected and skipped certificate deletion.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/.github/actions/to-integration-tests/cdn.json
+++ b/.github/actions/to-integration-tests/cdn.json
@@ -9,6 +9,7 @@
 		"pid_file": "/var/run/traffic_ops.pid",
 		"workers": 12
 	},
+	"disable_auto_cert_deletion": false,
 	"use_ims": true,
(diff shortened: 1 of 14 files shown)


shamrickus (approved):
LGTM

PR #6251, opened 2021-10-04
Remove user_role and api_capability tables, and the associated endpoints

srijeet0406:
This PR removes the `user_role` and `api_capability` tables, and the associated endpoints.

<hr>

- Documentation
- Traffic Control Client 
- Traffic Ops

Make sure all the unit and API tests pass.

- master

## PR submission checklist
- [x] This PR has tests 


srijeet0406:
> We can't actually remove the routes from v2/v3 or even drop the `api_capability` table. The old routes should be marked deprecated for API consumers to have time to react as well as preserve interop between two different version of TO.

You're correct. It should be fixed now.

Issue #5927, opened 2021-06-10
CDN in a Box still runs Riak server


zrhoffman:
## I'm submitting a ...

-  improvement request (usability, performance, tech debt, etc.)

## Traffic Control components affected ...

-  CDN in a Box

## Current behavior:

The Riak `trafficvault` container is a default service in CDN in a Box.

## New behavior:

The `trafficvault` service should not be included in CDN in a Box by default. It (and the *traffic_vault* directory) should be moved to the *optional* directory.

## Minimal reproduction of the problem with instructions:

1.  Start CDN in a Box
2.  See that the Riak `trafficvault` container is running
    ```shell
    docker-compose ps trafficvault
    ```

PR #6252, opened 2021-10-05
Removed Riak from default CiaB

mattjackson220:
Closes: #5927 

This PR moves the Riak container in CiaB to only run if the optional flag is provided instead of by default.

<hr>

- Documentation
- CDN in a Box

Run CiaB and verify that the `trafficvault` container is not run
Verify that CiaB works normally
Run CiaB with the `-f optional/docker-compose.traffic-vault.yml` flag and the `TV_BACKEND` variable in `variables.env` updated to use Riak
Verify that Riak can successfully be used when the flag is provided
Verify that CiaB works normally with Riak

## PR submission checklist
- [x] This PR does not require testing updates since it is a CiaB change.

### DIFF
--- a/infrastructure/cdn-in-a-box/optional/docker-compose.traffic-vault.yml
+++ b/infrastructure/cdn-in-a-box/optional/docker-compose.traffic-vault.yml
@@ -31,8 +31,6 @@ services:
     build:
       context: ""
       dockerfile: optional/traffic_vault/Dockerfile
-    depends_on:
-      - enroller
     environment:
       - CLUSTER_NAME=trafficvault
     labels:
--- a/infrastructure/cdn-in-a-box/optional/traffic_vault/Dockerfile
+++ b/infrastructure/cdn-in-a-box/optional/traffic_vault/Dockerfile
@@ -27,11 +27,11 @@ RUN echo 'APT::Install-Recommends 0;' >> /etc/apt/apt.conf.d/01norecommends \
  && DEBIAN_FRONTEND=noninteractive apt-get install -y net-tools ca-certificates dnsutils gettext-base \
  && rm -rf /var/lib/apt/lists/* && rm -rf /etc/apt/apt.conf.d/docker-gzip-indexes
 
-ADD traffic_vault/prestart.d/* /etc/riak/prestart.d/
-ADD traffic_vault/poststart.d/* /etc/riak/poststart.d/
+ADD optional/traffic_vault/prestart.d/* /etc/riak/prestart.d/
+ADD optional/traffic_vault/poststart.d/* /etc/riak/poststart.d/
 ADD enroller/server_template.json \
-    traffic_vault/run.sh \
-    traffic_vault/sslkeys.xml \
+    optional/traffic_vault/run.sh \
+    optional/traffic_vault/sslkeys.xml \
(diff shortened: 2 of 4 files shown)


infrastructure/cdn-in-a-box/optional/docker-compose.traffic-vault.yml
```diff
@@ -0,0 +1,56 @@
...
+    labels:
+      - "com.basho.riak.cluster.name=trafficvault"
+    volumes:
+      - schemas:/etc/riak/schemas
+      - shared:/shared
+    domainname: infra.ciab.test
+    env_file:
+      - variables.env
+    hostname: trafficvault
+    ports:
```
rawlinp, line 47: I think since the expose-ports version is for exposing the ports, this one shouldn't also expose the ports, right?

rawlinp (changes requested):

rawlinp:
Update: that fixed the build, but it seems running w/ Riak is not working because the enroller is trying to create the DS before Riak is ready, which makes it 500 since we now generate self-signed certs at creation:
```
INFO: enroller.go:1054: 2021-10-06T23:16:43.063716314Z: new file : /shared/enroller/deliveryservices/010-demo1.json
INFO: enroller.go:1061: 2021-10-06T23:16:43.063750984Z: creating deliveryservices from /shared/enroller/deliveryservices/010-demo1.json
INFO: enroller.go:234: 2021-10-06T23:16:43.192919103Z: error creating Delivery Service: error requesting Traffic Ops: path 'https://trafficops.infra.ciab.test:443/api/4.0/deliveryservices' gave HTTP error 500 Internal Server Error - error-level alerts: Internal Server Error - alerts: {Alerts:[{Text:Internal Server Error Level:error}]}
INFO: enroller.go:1084: 2021-10-06T23:16:43.192980923Z: error creating deliveryservices from /shared/enroller/deliveryservices/010-demo1.json: error requesting Traffic Ops: path 'https://trafficops.infra.ciab.test:443/api/4.0/deliveryservices' gave HTTP error 500 Internal Server Error - error-level alerts: Internal Server Error
```

rawlinp:
I was also considering just getting rid of Riak in CIAB entirely, but I came to the same conclusion as Zach. Until we're fully migrated onto Postgres and completely trust that the migration tool won't need any Riak-related bugfixes, we should probably keep it in CIAB.

ocket8888:
The only reason CiaB needed TO-Perl was because it _wasn't _ optional, since there were API 1.x endpoints that were never re-written into Go, and because it provided the static HTTP file server that Traffic Router needs to work. If it were possible I'd have ripped it out much earlier.

I don't really think CiaB needs to support every setup that the project as a whole supports, especially when it's deprecated.

All that being said, I don't think supporting legacy Riak KV-based TV makes CiaB _worse_, really, so if someone's willing to do the work anyway, go for it, I guess. Just my two cents.

rawlinp:
It's almost fully supported already, we just need to make the enroller wait to create DSes until Riak is ready or just have the enroller retry failed requests indefinitely. Which, maybe it should do that, because today it just ignores the error and moves on, which makes it harder to realize that something failed and that your CIAB won't start up properly.

PR #6253, opened 2021-10-05
Fix TP roles navigation

srijeet0406:
This PR is not related to any issue. Some things were broken as a result of https://github.com/apache/trafficcontrol/pull/6124. This PR just fixes them.

<hr>

- Traffic Portal

Make sure the following things are fixed in TP:

`1. when creating a delivery service, as a user with the admin role, I no longer have the option to Fulfill Request Immediately`
`2. when I go to a specific user, then click on "view details" under the role name, it takes me back to the main dashboard. I would expect this to take me to that specific role.`
`3. when I go to a specific role, then click on "view users", it takes me back to the main dashboard. I would expect this to take me to a list of users with that specific role.`

- master

## PR submission checklist
- [] Just a fix related to the above mentioned PR, tests seemed unnecessary 
- [] Just a fix related to the above mentioned PR, documentation seemed unnecessary 


rawlinp (approved):
Navigation is fixed, and the `dsRequests.overrideRole` functionality appears to be working again. Just waiting for the GHA to pass before merging.

PR #6278, opened 2021-10-13
TP: replace compass/ruby

shamrickus:
This PR changes the TP sass compiler from `compass` (deprecated) to `dart-sass` (official sass compiler). As a result, ruby is no longer a dependency of TP. 

It also changes the TP build docker file to always install nodejs 12.

<hr>

- Documentation
- Traffic Portal
- CDN in a Box

***Note***: CiaB (and associated CI) for TP _build_ will always fail (until merged) as it is using the dockerhub TP image, which was built using nodejs 10. `gulp-dart-sass` requires nodejs 12.

Build TP locally and confirm that it works.
Build TP using `pkg` with the `-b` command (so it uses the correct nodejs version).
Verify that TP looks correct.
Check the docs and make sure `ruby`/`compass` are no longer mentioned.

It is also possible to diff `dist/resources/styles` and see that the CSS is mostly the same with a few exceptions. First, sourceMaps are now specified by comments instead of through media queries. Second, the `compass` minification doesn't remove some chars that `dart-sass` does, as a result the new CSS should always be at least as small.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_portal/grunt/watch.js
+++ b/traffic_portal/grunt/watch.js
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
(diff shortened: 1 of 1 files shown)


zrhoffman:
#6278's TP GHA failed due to #6283 needs to be rebased onto master now that #6284 is merged in order to pass

Issue #6268, opened 2021-10-12
DNSSEC refresh change log entry unnecessary if no keys are refreshed


rawlinp:
## This Improvement request (usability, performance, tech debt, etc.) affects these Traffic Control components:

- Traffic Ops

## Current behavior:

When making a request to `/cdns/dnsseckeys/refresh`, a change log entry is created even if there were no DNSSEC keys refreshed:
> Refreshed 0 DNSSEC keys

## New behavior:

If 0 DNSSEC keys are refreshed via a request to `/cdns/dnsseckeys/refresh`, a change log entry should not be created. While it's nice to know that the cron job is working, this is mostly useless information. A change log entry is really only necessary if any keys are actually refreshed.

PR #6279, opened 2021-10-13
Reduce DNSSEC refresh change log noise

rawlinp:
Only create a changelog entry if any keys actually changed or an error
occurred. Also, differentiate between errors storing into Traffic Vault
and other types of errors.

Closes: #6268

<hr>

- Traffic Ops

Attempt to refresh the DNSSEC keys in quick succession. If there are no keys refreshed, there should be no change log entry created.

6.0.0

## PR submission checklist
- [x] Change log entry creation not critical enough for testing
- [x] No docs necessary


rawlinp:
For some reason, the TP tests are failing to get past the initialize stage, and it seems to be an issue unrelated to this PR.

PR #6294, opened 2021-10-16
fixed incorrect field name

### LINKED ISSUE #6130: Docs for /users has incorrect field name
- Documentation

## Current behavior:
The docs say that there's a field containing the Name of the User's Role called `roleName`.

## Expected behavior:
The docs should accurately document the API's output.

## Steps to reproduce:
`toget -kp 'users?limit=1' | jq '.response[0]'` - returns:
```json
{
        "rolename": "admin",
        "username": "admin",
        "registrationSent": null,
        "addressLine1": null,
        "addressLine2": null,
        "city": null,
        "company": null,
        "country": null,
        "email": null,
        "fullName": null,
        "gid": null,
        "id": 2,
        "newUser": false,
        "phoneNumber": null,
        "postalCode": null,
        "publicSshKey": null,
        "role": 1,
        "stateOrProvince": null,
        "tenant": "root",
        "tenantId": 1,
        "uid": null,
        "lastUpdated": "2021-08-20 17:47:38+00"
}
```

Note the field name is `rolename`, not `roleName`.

It's unclear whether this is a regression without digging further into git history than I feel like doing right now, but I think that's unlikely, and it's probably just that the docs have been copied with each successive major API version release from the previous version since v2 was added, and the GoDoc for the field on the `lib/go-tc` struct related to API User representations mentions a difference between API versions in the name of that property. Interestingly, though, the API docs for v2 get the name right, while v3 and v4 both use the incorrect camelCase form.

Ayushsunny:
Fixes #6130 

<hr>

- Documentation

### DIFF
--- a/docs/source/api/v2/users.rst
+++ b/docs/source/api/v2/users.rst
@@ -209,7 +209,7 @@ Response Structure
 :publicSshKey:     The user's public key used for the SSH protocol
 :registrationSent: If the user was created using the :ref:`to-api-users-register` endpoint, this will be the date and time at which the registration email was sent - otherwise it will be ``null``
 :role:             The integral, unique identifier of the highest-privilege role assigned to this user
-:roleName:         The name of the highest-privilege role assigned to this user
+:rolename:         The name of the highest-privilege role assigned to this user
 :stateOrProvince:  The name of the state or province where this user resides
 :tenant:           The name of the tenant to which this user belongs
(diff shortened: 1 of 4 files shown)


Ayushsunny:
@ocket8888 review

Ayushsunny:
Done @rawlinp. 
could you please check now?

Issue #6106, opened 2021-08-16
Traffic Ops database is not locked during Postinstall


zrhoffman:
## This Bug Report affects these Traffic Control components:

- Traffic Ops

## Current behavior:

We have no mechanism to verify that multiple Traffic Ops instances are not trying to run the TO Postinstall at the same time.

Previously filed in #4984; sort of fixed in #5138 by making `create_tables.sql` alone idempotent; idempotency is lost as soon as `squash_migrations.sh` from #6096 is used.

## Expected behavior:

When the Postinstall script starts, it should check if a postinstall lock exists in the DB, and if so, wait for it to be cleared. If there is no lock, it should set one and clear it once Postinstall completes (with or without an error).

The `cdn_lock` table cannot be used because the `username` and `cdn_name` are non-nullable foreign keys.

Also, we should be using `--single-transaction` for all of our `psql` commands in the Postinstall script and in `db/admin`.

## Steps to reproduce:

Try to run the Traffic Ops Postinstall on the same database from two separate TO instances.

zrhoffman:
To be clear, the Postinstall script is designed to be able to be run more than once:
https://github.com/apache/trafficcontrol/blob/9516de39fffa3ddda0346720ef98edabc8efa07e/traffic_ops/install/bin/_postinstall.py#L1113-L1123

https://github.com/apache/trafficcontrol/blob/9516de39fffa3ddda0346720ef98edabc8efa07e/traffic_ops/install/bin/_postinstall.pl#L600-L608

but if the Postinstall script is being run simultaneously, for the first time,. ,on 2 separate Traffic Ops hosts against the same TO DB, the check may still fail to find existing tables.

PR #6307, opened 2021-10-21
Add logic to every endpoint to be able to use the new permissions based authorization system

srijeet0406:
This PR is not related to any issue. It adds the required permissions to the 4.0 routes.

<hr>

- Documentation
- Traffic Ops

Make sure the TO and TP tests pass. Manually test the 4.0 routes (as many as you can) with different roles to make sure that only a role (associated with a user) with the required permissions can access those routes.

- master

## PR submission checklist
- [x] This PR has tests 


ocket8888 (changes requested):
We shouldn't remove `/capabilities` with this PR, it's an unrelated body of work, and one that constitutes more than is being done here from what I can tell. I opened #6319 to track it.

traffic_ops_db/test/docker/run-db-test.sh
```diff
@@ -99,12 +99,6 @@ fi
 
 ./db/admin --env=production upgrade || { echo "DB upgrade failed!"; exit 1; }
 
-if ! ./db/admin -env=production load_schema ||
-  ! ./db/admin -env=production load_schema; then
-  echo 'Could not re-run create_tables.sql!'
-  exit 1
-fi;
-
```
zrhoffman, line 107: As pointed out to me by @ocket8888, this means #4984 becomes reproducible again, which would be fixed by fixing #6106.

ocket8888: I'm good with that, if you are.

ocket8888 (approved):
Verified manually that a user without enough Permissions to use an endpoint now gets rejected with an error describing said Permissions when attempting to use it anyway - except a user with the `admin` Role, which is always granted access, regardless of Permissions.

traffic_ops_db/test/docker/run-db-test.sh
zrhoffman: Yep that's fine with me, as long as #6106 gets fixed at some point.

traffic_ops/app/db/migrations/202110191500000000_add_permissions.down.sql
```diff
@@ -0,0 +1,138 @@
+/*
```
rawlinp, line 1: I'm not sure if it's a big deal or not, but this file has a couple extra digits in the timestamp compared to prior timestamps:
```
2021100114200000_remove_user_role.down.sql
2021100114200000_remove_user_role.up.sql
202110191500000000_add_permissions.down.sql
202110191500000000_add_permissions.up.sql
```
was that intentional?

ocket8888: @srijeet0406 Did you generate these with `admin` or make the filenames yourself?

zrhoffman: The extra digits should have made the *todb-tests* action fail (was it even triggered on this PR?):
https://github.com/apache/trafficcontrol/blob/631d8793eb7e83f9db2624f1cc0723f9f2d4ff28/.github/actions/todb-tests/entrypoint.sh#L86-L96

srijeet0406: I named them myself, I wasn't even aware that I could generate them using `admin`. I'll create a follow up PR.

ocket8888: > was it even triggered on this PR?

Yeah, this is the PR where we had to stop checking for being able to run `create_tables` twice, remember? The problem is that `VIOLATING_FILES` isn't being set properly, the set is empty after that `sed` call:

```console
$ ls | sort | cut -d _ -f 1 | sed -n -e '/2020061622101648/,$p' 
$
```

zrhoffman: Weird that it doesn't show up in the list of actions run above. It shows up when [searching runs by branch `CDN-13621`](https://github.com/apache/trafficcontrol/actions/workflows/traffic.ops.database.yml?query=branch:CDN-13621), though.

PR #6320, opened 2021-11-01
Adds support to the tc-health-client for querying Traffic Monitors via a caching Proxy server.

jrushford:
Adds support to the tc-health-client for querying Traffic Monitors via a caching Proxy server.

- adds a tm-proxy-url config property
- updates the traffic_monitor/tmclient to use an optional http.Transport
  object needed to support an Http Proxy.

On an edge cache, configure the tc-health-client to run with the tm-proxy-url set in the /etc/trafficcontrol/tc-health-client.json config file.  With the proxy server configured to proxy traffic monitor requests from the health-client, verify traffic monitor requests are successful, see /var/log/trafficcontrol/tc-health-client.log

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/tc-health-client/config/config.go
+++ b/tc-health-client/config/config.go
@@ -60,6 +60,7 @@ type Cfg struct {
 	TOPass                  string          `json:"to-pass"`
 	TOUrl                   string          `json:"to-url"`
 	TOUser                  string          `json:"to-user"`
+	TmProxyURL              string          `json:"tm-proxy-url"`
 	TmPollIntervalSeconds   string          `json:"tm-poll-interval-seconds"`
 	TmUpdateCycles          int             `json:"tm-update-cycles"`
 	TrafficServerConfigDir  string          `json:"trafficserver-config-dir"`
@@ -67,6 +68,7 @@ type Cfg struct {
 	TrafficMonitors         map[string]bool `json:"trafficmonitors,omitempty"`
 	HealthClientConfigFile  util.ConfigFile
 	CredentialFile          util.ConfigFile
+	ParsedProxyURL          *url.URL
 }
 
 type LogCfg struct {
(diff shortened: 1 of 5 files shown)


traffic_monitor/tmclient/tmclient.go
```diff
@@ -198,6 +199,10 @@ func (c *TMClient) ConfigDoc() (handler.OpsConfig, error) {
 func (c *TMClient) getBytes(path string) ([]byte, error) {
 	url := c.url + path
 	httpClient := http.Client{Timeout: c.timeout}
+	// add the http.Transport object if specified.
+	if c.Transport != nil {
```
rob05c, line 203: Nitpick: the httpClient `Transport` is nil by default here, so the `if` check isn't necessary

tc-health-client/config/config.go
```diff
@@ -328,6 +331,22 @@ func LoadConfig(cfg *Cfg) (bool, error) {
...
+			if ProxyURL, err = url.Parse(cfg.TmProxyURL); err != nil {
+				ProxyURL = nil
+				return false, errors.New("parsing TmProxyUrl: " + err.Error())
+			}
+			if ProxyURL.Port() == "" {
+				ProxyURL = nil
+				return false, errors.New("TmProxyUrl invalid port specified")
+			}
+			log.Infof("TM queries will use the proxy: %s", cfg.TmProxyURL)
+		} else {
```
rob05c, line 347: Nitpick: the `ProxyURL` is already nil, and nothing outside this `if` changes it, so the `else { ProxyURL = nil` isn't necessary

traffic_monitor/tmclient/tmclient.go
jrushford: @rob05c I added this so that I can inject an http.Transport into the httpClient.  I only want to inject an http.Transport when I've created and added one to the TMClient which is what's done when using a Proxy.

tc-health-client/config/config.go
jrushford: @rob05c When the config is reloaded and the tm-proxy-url parameter has been removed, the ProxyURL is set to nil and no proxy will be used the next time a TM is queried.

rob05c: Ah, I see, because `ProxyURL` is a global variable, if `LoadConfig` is called twice it'll be different.
This is why I don't like Global Variables. But yes, you're right, disregard my comment.

traffic_monitor/tmclient/tmclient.go
jrushford: @rob05c it's needed so as not to eliminate the Default Transport.  It overrides the Default Transport with the one created for using a proxy.  If I use this without the nil check, the default Transport is eliminated and the TM integrations tests fail:

httpClient := http.Client{Timeout: c.timeout, Transport: c.Transport}

rob05c: Ah, because `TMClient.Transport` a `*http.Transport` and `http.Client` is a `http.RoundTripper`, there's an implicit cast happening, and it's creating a non-nil interface pointing to a nil pointer. Which is where the weirdness and test failure is coming from.

If `TMClient.Transport` was a `RoundTripper` it'd avoid the cast, and work as-expected. But interfaces are really pointers, and pointers-to-pointers where the outer interface pointer is non-nil with an object that's a nil pointer get weird.

That's fine, I'd just leave it. Sorry, I wouldn't have suggested it if I'd noticed the types were different. Non-nil interfaces with potentially nil objects are one of Go's pain points, not worth dealing with.

jpappa200 (approved):
👍  looks good to me.

rob05c (approved):
LGTM, code looks good, tests pass, and I trust Joe's review.

PR #6321, opened 2021-11-01
Fix the names of the migrations for roles and perms

srijeet0406:
This PR is not related to any issue. It just fixes the names of the migrations added in https://github.com/apache/trafficcontrol/pull/6307
<hr>

- Traffic Ops

Make sure the migrations run successfully. Also make sure that the GHA todb_tests passes.

- master

## PR submission checklist
- [ ] This PR has tests 


rawlinp (changes requested):
The underlying issue is more than just these 2 migrations. Since we backported `le_dns_challenge_xml_id` to 6.0, every migration between that one and `fix_indices` needs to be renamed to be later than `le_dns_challenge_xml_id`.

rawlinp:
Oh, I see this is just for removing the extra digits. The issue I just referred to is different, but still needs fixed if you'd like to just take care of that in this PR.

srijeet0406:
> Oh, I see this is just for removing the extra digits. The issue I just referred to is different, but still needs fixed if you'd like to just take care of that in this PR.

Yeah I'll rename all those migrations in this PR.

zrhoffman:
#6321 should be rebased now that @ocket8888's PR to fix for the TO DB GHA in #6322 is merged.

rawlinp (approved):

PR #6322, opened 2021-11-02
Fix GHA that checks db migration filenames to properly check timestamp digits

ocket8888:
At some point our GHA that is supposed to ensure the proper number of digits for database migration file name timestamps broke and allowed in something that should've been invalid in #6307. This PR fixes the GHA to properly mark those files as in violation of the naming scheme.

<hr>

- Automation (GitHub Actions)

This PR _should_ fail that check, because our master branch currently _should_ fail that check but doesn't - that's the bug this fixes. If you create a new branch on your fork based on this branch, delete the migration files added in #6307, then push, the check should pass (though others may fail).

- master

## PR submission checklist
- [x] This PR fixes tests
- [x] This PR has no documentation
- [x] This PR has no CHANGELOG.md entry

### DIFF
--- a/.github/actions/todb-tests/entrypoint.sh
+++ b/.github/actions/todb-tests/entrypoint.sh
@@ -85,7 +85,7 @@ for migration_dir in ${migration_dirs[@]}; do
 
 	set +e;
 	# All new migrations must use 16-digit timestamps.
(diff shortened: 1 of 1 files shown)


zrhoffman (approved):
*Traffic Ops and Traffic Vault Database Tests* GHA workflow gives

```
ERROR: traffic_ops/app/db/migrations/202110191500000000
202110191500000000(name).sql: wrong filename, all migrations after 2020-06-16 must use 16-digit timestamps in their filenames
```

which is accurate.

PR #6324, opened 2021-11-02
Disallow affecting the "admin" Role.

ocket8888:
This PR adds validation to the TO API that forbids modifying or deleting the special "admin" Role. It also adds a database migration that changes the "admin" Role's Permission set to the one specified by the original blueprint: `["ALL"]`.

Changes were also made to Traffic Portal to make it clear that the "admin" Role cannot be deleted or modified.

This PR also updates the patches and seeds SQL files to use new Permissions instead of legacy API-method-and-endpoint-to-Legacy-Capability mappings.

<hr/>

- Documentation
- Traffic Ops
- Traffic Portal

Make sure the provided test passes

## PR submission checklist
- [x] This PR has tests

### DIFF
--- a/traffic_ops/traffic_ops_golang/role/roles.go
+++ b/traffic_ops/traffic_ops_golang/role/roles.go
@@ -226,7 +226,8 @@ func (role *TORole) Update(h http.Header) (error, error, int) {
 	var isAdmin bool
 	if err := role.ReqInfo.Tx.Get(&isAdmin, isAdminQuery, role.ID); err != nil {
 		return nil, fmt.Errorf("checking if Role to be modified is '%s': %w", tc.AdminRoleName, err), http.StatusInternalServerError
-	} else if isAdmin {
+	}
+	if isAdmin {
 		return cannotModifyAdminError, nil, http.StatusBadRequest
 	}
 
@@ -265,7 +266,8 @@ func (role *TORole) Delete() (error, error, int) {
 	var isAdmin bool
 	if err := role.ReqInfo.Tx.Get(&isAdmin, isAdminQuery, role.ID); err != nil {
 		return nil, fmt.Errorf("checking if Role to be deleted is '%s': %w", tc.AdminRoleName, err), http.StatusInternalServerError
-	} else if isAdmin {
+	}
+	if isAdmin {
 		return cannotModifyAdminError, nil, http.StatusBadRequest
 	}
 
--- a/traffic_ops/traffic_ops_golang/role/roles.go
+++ b/traffic_ops/traffic_ops_golang/role/roles.go
@@ -41,6 +41,14 @@ import (
 	"github.com/lib/pq"
 )
 
+type roleError string
+
+func (e roleError) Error() string {
+	return string(e)
+}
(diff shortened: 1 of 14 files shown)


traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -239,12 +256,19 @@ func (role *TORole) Update(h http.Header) (error, error, int) {
...
+	if err := role.ReqInfo.Tx.Get(&assignedUsers, "SELECT COUNT(id) FROM public.tm_user WHERE role=$1", role.ID); err != nil {
 		return nil, errors.New("role delete counting assigned users: " + err.Error()), http.StatusInternalServerError
 	} else if assignedUsers != 0 {
 		return fmt.Errorf("can not delete a role with %d assigned users", assignedUsers), nil, http.StatusBadRequest
 	}
 
+	var isAdmin bool
+	if err := role.ReqInfo.Tx.Get(&isAdmin, isAdminQuery, role.ID); err != nil {
+		return nil, fmt.Errorf("checking if Role to be deleted is '%s': %w", tc.AdminRoleName, err), http.StatusInternalServerError
+	} else if isAdmin {
```
srijeet0406, line 268: Is this `else` needed?

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -213,6 +223,13 @@ func (role *TORole) Read(h http.Header, useIMS bool) ([]interface{}, error, erro
 }
 
 func (role *TORole) Update(h http.Header) (error, error, int) {
+	var isAdmin bool
+	if err := role.ReqInfo.Tx.Get(&isAdmin, isAdminQuery, role.ID); err != nil {
+		return nil, fmt.Errorf("checking if Role to be modified is '%s': %w", tc.AdminRoleName, err), http.StatusInternalServerError
+	} else if isAdmin {
```
srijeet0406, line 229: Is this `else` needed?

traffic_ops/traffic_ops_golang/role/roles.go
ocket8888: no, I don't think so. That's just habit, because `isAdmin` is assigned a value within the `if` statement I thought of it as only accessible within that scope. But that's not true.

traffic_ops/app/db/migrations/2021110207242921_admin_all_permission.up.sql
```diff
@@ -0,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
srijeet0406, line 2: This file and the corresponding `down` migration file should be renamed so that the timestamp is after the latest refetch migration.

srijeet0406 (approved):
Code looks good. 
Api and unit tests pass.
Ran the migrations and made sure things looked correct.
Manually tested with locally run TO and TP, and verified that the admin role cannot be modified. Also, another admin role cannot be created.
Made sure that the admin role cannot be modified via API calls.

PR #6325, opened 2021-11-03
Feature/tp refetch

TaylorCFrey:
This is a continuation of the work in ([PR 6118](https://github.com/apache/trafficcontrol/pull/6118)) as defined in the ([Refetch Blueprint](https://github.com/apache/trafficcontrol/pull/5910)). These changes allow Traffic Portal to take advantage of the changes to the Traffic Ops API and Invalidation Jobs.

The table listing invalidation jobs has been modified to show the TTL explicitly (rather than Parameters) and the create form has been modified to account for the new Invalidation Type field. Traffic Portal integration test was updated to account for these changes as well.

There was also a minor change cleaning up the _roles_ that were defined in some tests that may have been conflicting (`read-only user` should be `read-only`).

<hr>

- Traffic Portal

Manually, to launch and run Traffic Ops and Traffic Portal. Navigate to the Invalidation Content page under the Tools menu in Traffic Portal. Note the `Parameters` column is now titled `TTL (Hours)` as well as the addition of the `Invalidation Type`.

Manually, create an Invalidation Request. Note the new required field for `Invalidation Type`, which defaults to _REFRESH_.

Repeat the same observation and test as above, however navigate to Delivery Services. Select a Delivery Service. In the upper right drop down menu _More_ -> _Manage Invalidation Requests_. The difference here is that the table and create form are specific to this particular DS.

Automatic integration tests must pass.

## PR submission checklist
- [x] This PR has tests 
- [ ] This PR has documentation - Documentation will be addressed in a separate PR 

### DIFF
--- a/traffic_portal/app/src/common/modules/form/deliveryServiceJob/form.deliveryServiceJob.tpl.html
+++ b/traffic_portal/app/src/common/modules/form/deliveryServiceJob/form.deliveryServiceJob.tpl.html
@@ -40,14 +40,22 @@ under the License.
                     <span ng-show="hasError(jobForm.regex)" class="form-control-feedback"><i class="fa fa-times"></i></span>
                 </div>
             </div>
-            <div class="form-group" ng-class="{'has-error': hasError(jobForm.ttl), 'has-feedback': hasError(jobForm.ttl)}">
+            <div class="form-group" ng-class="{'has-error': hasError(jobForm.ttlhours), 'has-feedback': hasError(jobForm.ttlhours)}">
                 <label class="control-label col-md-2 col-sm-2 col-xs-12">TTL (hours) *</label>
                 <div class="col-md-10 col-sm-10 col-xs-12">
-                    <input name="ttl" type="number" class="form-control" placeholder="Number of hours until the invalidation request expires" ng-model="job.ttl" ng-maxlength="3" ng-pattern="/^\d+$/" required autofocus>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'required')">Required Whole Number</small>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'maxlength')">Too Long</small>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'pattern')">Whole Number</small>
-                    <span ng-show="hasError(jobForm.ttl)" class="form-control-feedback"><i class="fa fa-times"></i></span>
+                    <input name="ttlhours" type="number" class="form-control" placeholder="Number of hours until the invalidation request expires" ng-model="job.ttlhours" ng-maxlength="3" ng-pattern="/^\d+$/" required autofocus>
+                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttlhours, 'required')">Required Whole Number</small>
+                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttlhours, 'maxlength')">Too Long</small>
+                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttlhours, 'pattern')">Whole Number</small>
+                    <span ng-show="hasError(jobForm.ttlhours)" class="form-control-feedback"><i class="fa fa-times"></i></span>
+                </div>
+            </div>
+            <div class="form-group" ng-class="{'has-error': hasError(jobForm.invalidationtype), 'has-feedback': hasError(jobForm.invalidationtype)}">
+                <label class="control-label col-md-2 col-sm-2 col-xs-12">Invalidation Type *</label>
+                <div class="col-md-10 col-sm-10 col-xs-12">
+                    <select name="invalidationtype" class="form-control" ng-model="job.invalidationType" ng-options="invalidationtype as invalidationtype for invalidationtype in invalidationtypes" required>
+                    </select>
+                    <small class="input-error" ng-show="hasPropertyError(jobForm.invalidationtype, 'required')">Required</small>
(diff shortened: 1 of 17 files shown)


traffic_portal/app/src/common/modules/form/job/form.job.tpl.html
```diff
@@ -47,14 +47,22 @@
...
                 <label class="control-label col-md-2 col-sm-2 col-xs-12">TTL (hours) *</label>
                 <div class="col-md-10 col-sm-10 col-xs-12">
-                    <input name="ttl" type="number" class="form-control" placeholder="Number of hours until the invalidation request expires" ng-model="job.ttl" ng-maxlength="3" ng-pattern="/^\d+$/" required autofocus>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'required')">Required Whole Number</small>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'maxlength')">Too Long</small>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'pattern')">Whole Number</small>
-                    <span ng-show="hasError(jobForm.ttl)" class="form-control-feedback"><i class="fa fa-times"></i></span>
+                    <input name="ttlhours" type="number" class="form-control" placeholder="Number of hours until the invalidation request expires" ng-model="job.ttlhours" ng-maxlength="3" ng-pattern="/^\d+$/" required autofocus>
```
ocket8888, line 53: I'm just looking through this for docs reasons, but I can't stop myself from pointing out: this is an `input[type="number"]:required` so setting a pattern of `/^\d+$/` has no effect, since only things that would match that pattern (which can be just `pattern` too, since it's static - just like how `ng-maxlength=3` is no different from `maxlength=3` since `3` is a constant value but **also** also on a numeric input the controls will behave more nicely if you just set `max="999"` instead of `maxlength="3"`) are valid inputs anyway. And also, since this is a number, you don't need to be `parseInt`-ing down below.

TaylorCFrey: This is good stuff. I didn't know this. I honestly tried to touch as little as possible in this spot of the code, so what we see is the original markup. From what I am understanding, since it is `required` and `type="number"` we can do the following:
- Remove the `ng-pattern` attribute
- Replace the `ng-maxlength` with `max="999"` (Do/should we set a `min="0"` or `min="1"` as well?)

As for the `parseInt()` call in the controller, I was to understand that was parsing the input as it was returned from the server, which until this time was the `parameters` field and a string ("TTL:36h"). Is it safe to assume that since the API has been changed to return an uint it is now a JavaScript number and doesn't need to be "parsed"?

ocket8888: Yes, the marshaler in Go will convert a `uint` to a JSON number, which `JSON.parse` will convert into a `number`.

There should probably be a minimum, probably 1.

zrhoffman:
In order for CI to pass, #6325 needs to be rebased onto the [`apache:master`](https://github.com/apache/trafficcontrol/commits/master) branch now that the CI fix in apache/trafficcontrol#6329 is merged:
```shell
git remote add apache git@github.com:apache/trafficcontrol.git
git fetch apache
git rebase apache
git push your-fork your-branch --force
```

shamrickus (approved):
LGTM

traffic_portal/app/src/common/modules/form/deliveryServiceJob/form.deliveryServiceJob.tpl.html
```diff
@@ -40,14 +40,20 @@
...
                 <label class="control-label col-md-2 col-sm-2 col-xs-12">TTL (hours) *</label>
                 <div class="col-md-10 col-sm-10 col-xs-12">
-                    <input name="ttl" type="number" class="form-control" placeholder="Number of hours until the invalidation request expires" ng-model="job.ttl" ng-maxlength="3" ng-pattern="/^\d+$/" required autofocus>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'required')">Required Whole Number</small>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'maxlength')">Too Long</small>
-                    <small class="input-error" ng-show="hasPropertyError(jobForm.ttl, 'pattern')">Whole Number</small>
-                    <span ng-show="hasError(jobForm.ttl)" class="form-control-feedback"><i class="fa fa-times"></i></span>
+                    <input name="ttlhours" type="number" class="form-control" placeholder="Number of hours until the invalidation request expires" ng-model="job.ttlhours" min="1" max="999" required autofocus>
```
ocket8888, line 46: FWIW, `"Number of hours until the invalidation request expires"` is not a an appropriate `placeholder` value for an `input[type="number"]`.

From [the MDN section on input labeling](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#labels):
> The placeholder should never be required to understand your forms. It is not a label, and should not be used as a substitute, because it isn't. The placeholder is used to provide a hint as to what an inputted value should look like, not an explanation or prompt.

So it's supposed to be an example input, and this looks more like a label. Traditionally we'd put that in the "helptext" tooltips for the input labels, though for all I care it could just be the entire `label` text content. The `title` attribute might also be appropriate, though whether or not any user would encounter that is very implementation-dependent (and since we use `form[novalidate]`s no non-screen-reading user would ever see them unless they decide to hover it in a Chromium or Gecko engine browser).

ocket8888 (approved):

PR #6337, opened 2021-11-05
updated database and added script to pull expiration info from cert data

mattjackson220:
This PR adds columns to the Postgres TV database for expiration date and provider and includes a script to fill those in from existing certs.  This will be used with later updates (separate PRs that will reference this)

<hr>

- Traffic Vault

make sure your environment has TV Postgres setup and has some certificates in the sslkeys table.
run the `admin` tool with `--trafficvault` to upgrade the traffic vault DB
verify the new columns are there
run the script and make sure those columns get filled in but nothing else is impacted

## PR submission checklist
This PR does not have tests or documentation since its a small DB update. It does not have a changelog update since that will come in the last PR
- [x] This PR has tests 


traffic_ops/app/db/trafficvault/scripts/fill_expiration_and_provider.conf
```diff
@@ -0,0 +1,8 @@
+{
+	"dbname": "",
+	"hostname": "",
+	"user": "",
+	"password": "",
+	"port": 5432,
+	"ssl": false
+}
```
ocket8888, line 8: Can you name this file with `.json`? Some editors aren't smart enough to figure out the grammar from the contents and rely on "file extensions".

traffic_ops/app/db/trafficvault/scripts/fill_expiration_and_provider.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+		id := strings.Join([]string{ds, cdn, version}, ", ")
+		jsonKeys, err := util.AESDecrypt(encryptedSslKeys, aesKey)
+		if err != nil {
+			die("reading SSL Keys: " + err.Error())
+		}
+
+		if !bytes.HasPrefix(jsonKeys, []byte("{")) {
+			die("decrypted SSL Key did not have prefix '{' for id " + id)
+		}
```
ocket8888, line 116: Do you need to do this check? Wouldn't the `json.Unmarshal` call below catch it?

traffic_ops/app/db/trafficvault/scripts/fill_expiration_and_provider.go
```diff
@@ -0,0 +1,217 @@
...
+
+		sslKeyMap[id] = expiryAndProvider{
+			Provider:   sslKey.AuthType,
+			Expiration: x509cert.NotAfter,
+		}
+	}
+
+	for id, info := range sslKeyMap {
+		idParts := strings.Split(id, ", ")
```
ocket8888, line 147: If the map key doesn't contain exactly two commas, should this exit here? Currently, more than two commas will be ignored and any text after the third comma is stripped and discarded - is that proper? Or should it consider that part of the `version`? If it's less than two commas, then this will segfault and exit here. IMO it would be better to print a single line error and exit with an error code, but alternatively, would you want it to continue in that case and try to read the other key ids? If so, you'd need to handle that here.

traffic_ops/app/db/trafficvault/scripts/fill_expiration_and_provider.go
```diff
@@ -0,0 +1,217 @@
...
+		ds := idParts[0]
+		cdn := idParts[1]
+		version := idParts[2]
+		res, err := tx.Exec(`UPDATE sslkey SET provider = $1, expiration = $2 WHERE deliveryservice = $3 AND cdn = $4 AND version = $5`, info.Provider, info.Expiration, ds, cdn, version)
+		if err != nil {
+			die(fmt.Sprintf("updating SSL Keys for %s, %s", id, err.Error()))
```
ocket8888, line 153: nit but calling `.Error()` is already the default string coercion for errors, so if you know it isn't `nil` `%s` formatting suffices: https://play.golang.org/p/FtjDb-2ze0Y (if it might be `nil` you'd need to use `%v` instead to avoid segfaulting in that case - but still don't need to explicitly call `.Error()`).

traffic_ops/app/db/trafficvault/scripts/fill_expiration_and_provider.go
```diff
@@ -0,0 +1,217 @@
...
+	Password string `json:"password"`
+	Port     int    `json:"port"`
+	SSL      bool   `json:"ssl"`
+}
+
+func readKey(keyLocation string) ([]byte, error) {
+	var keyBase64 string
+	keyBase64Bytes, err := ioutil.ReadFile(keyLocation)
+	if err != nil {
+		return []byte{}, errors.New("reading file '" + keyLocation + "':" + err.Error())
```
ocket8888, line 178: constructing errors from errors should use wrapping

traffic_ops/app/db/trafficvault/scripts/fill_expiration_and_provider.go
```diff
@@ -0,0 +1,215 @@
...
+
+	_ "github.com/lib/pq"
+)
+
+const PROPERTIES_FILE = "./fill_expiration_and_provider_conf.json"
+
+func main() {
+	aesKeyLocation := flag.String("aes-key", "/opt/traffic_ops/app/conf/aes.key", "(Optional) The file path for the previous base64 encoded AES key. Default is /opt/traffic_ops/app/conf/aes.key.")
+	cfg := flag.String("cfg", PROPERTIES_FILE, "(Optional) The path for the configuration file. Default is "+PROPERTIES_FILE+".")
+	help := flag.Bool("help", false, "(Optional) Print usage information and exit.")
```
ocket8888, line 45: Nit, but in general people consider options optional, so I don't think you need to be explicit about it.

ocket8888 (commented):
This all seems to be working fine, I just have one question: Does every TO installation need this binary? If so, it should be compiled (also the default name of the output binary is the directory name of the main package, so in this case it would be "scripts") because Traffic Ops doesn't require Go at runtime. If not, I think it should be in `tools/` instead of `traffic_ops/app/db/traffic_vault/scripts`, so that it doesn't get bundled in the RPM

ocket8888 (approved):

PR #6338, opened 2021-11-05
Cert Expirations Traffic Ops Endpoint

mattjackson220:
Requires PR #6337 to be merged first!

This PR adds a Traffic Ops API endpoint to retrieve data about SSL certificate expirations

<hr>

- Documentation
- Traffic Ops

Run Traffic OPs and verify that the new `sslkey_expirations` endpoint works and provides the correct information from the Traffic Vault database (must be Postgres, unsupported in Riak as of now)

## PR submission checklist
This PR does not have a changelog entry since that will be in the last PR with the TP updates
- [x] This PR has tests 


docs/source/api/v4/sslkey_expirations.rst
```diff
@@ -0,0 +1,92 @@
...
+**********************
+``sslkey_expirations``
+**********************
+
+``GET``
+=======
+Retrieves SSL certificate expiration information.
+
+:Auth. Required: Yes
+:Roles Required: "admin"
```
ocket8888, line 27: What Permissions does this require?

lib/go-tc/deliveryservice_ssl_keys.go
```diff
@@ -93,6 +93,15 @@ type DeliveryServiceSSLKeysV15 struct {
 	Expiration time.Time `json:"expiration,omitempty"`
 }
 
+type SSLKeyExpirationInformation struct {
```
ocket8888, line 96: GoDoc?

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -130,6 +130,9 @@ func Routes(d ServerData) ([]Route, http.Handler, error) {
 		 * 4.x API
 		 */
 
+		// SSL Keys
+		{api.Version{Major: 4, Minor: 0}, http.MethodGet, `sslkey_expirations$`, deliveryservice.GetSSlKeyExpirationInformation, auth.PrivLevelAdmin, []string{"ACME:CREATE", "ACME:READ"}, Authenticated, nil, 41357729075},
```
ocket8888, line 134: Why make a new endpoint instead of just adding expiration info to the keys already returned by other endpoints?

ocket8888 (commented):
It's probably not a big deal, but you may want to  mark things that depend other PRs as drafts, so that they get reviewed in the right order and nobody duplicates review effort

traffic_ops/traffic_ops_golang/routing/routes.go
mattjackson220: we dont really have an endpoint that would fill the need here since this is for all certs and not specific to a DS or CDN.  we could add these new fields to those other endpoints as well but this would require a new one anyway.  plus its also good because otherwise we'd return a ton of info we dont need (like all certs and private keys) and it would be a bigger security concern instead of just generic expiration info. i have it as `admin` now but it could be opened up to a lower auth level if we wanted to since it doesnt include the private info

ocket8888: it's tough because CDNs aren't tenanted. I'm tempted to say we should just have `/ssl_keys` with DS and CDN filters. But that's a lot of work too.

btw, this should probably allow a trailing `/`, since very nearly all other endpoints do in APIv4.

docs/source/api/v4/sslkey_expirations.rst
```diff
@@ -0,0 +1,93 @@
...
+	Set-Cookie: mojolicious=...; Path=/; Expires=Mon, 07 Jun 2021 22:52:20 GMT; Max-Age=3600; HttpOnly
+	Vary: Accept-Encoding
+	X-Server-Name: traffic_ops_golang/
+	Date: Mon, 07 Jun 2021 21:52:20 GMT
+	Content-Length: 384
+
+	{ "response": [
+		{
+			"deliveryservice": "foo1",
+			"deliveryservice_id": 1,
```
ocket8888, line 79: Two things here:
1. I don't really think we need two ways to identify the Delivery Service here. I'd go with XMLID, both because it's easier to understand and because it has some bearing on the certs.

b) response payload parameters should be `camelCase` not `snake_case`

lib/go-tc/deliveryservice_ssl_keys.go
```diff
@@ -93,6 +93,16 @@ type DeliveryServiceSSLKeysV15 struct {
...
 }
 
+// SSLKeyExpirationInformation contains information about an SSL key's expiration.
+type SSLKeyExpirationInformation struct {
+	DeliveryService   *string    `json:"deliveryservice"`
+	DeliveryServiceId *int       `json:"deliveryservice_id"`
+	Cdn               *string    `json:"cdn"`
+	Provider          *string    `json:"provider"`
+	Expiration        *time.Time `json:"expiration"`
+	Federated         *bool      `json:"federated"`
```
ocket8888, line 103: None of these could ever be null, right? Or maybe `DeliveryService` could. But if properties can't be null, then they don't need to be pointers, just causes otherwise unnecessary checking and/or segfaults

docs/source/api/v4/sslkey_expirations.rst
```diff
@@ -0,0 +1,93 @@
...
+``sslkey_expirations``
+**********************
+
+``GET``
+=======
+Retrieves SSL certificate expiration information.
+
+:Auth. Required: Yes
+:Roles Required: "admin"
+:Permissions Required: ACME:READ
```
ocket8888, line 28: This reads more than just ACME keys right? Besides, right now, ACME:READ is used to express the ability to read ACME account details. The ability to read SSL keys is behind DS-SECURITY-KEY:READ. If this is supposed to be a separate endpoint _because_ we want to separate the ability to read expirations from the ability to read keys, then this should probably have a new Permission entirely to encapsulate that.

traffic_ops/traffic_ops_golang/deliveryservice/keys.go
```diff
@@ -136,6 +136,39 @@ func AddSSLKeys(w http.ResponseWriter, r *http.Request) {
...
+	expirationInfos, ok, err := inf.Vault.GetExpirationInformation(inf.Tx.Tx, r.Context(), daysParam)
+	if err != nil {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("getting SSL keys expiration information from Traffic Vault: "+err.Error()))
+		return
+	}
+	if !ok {
+		api.HandleErr(w, r, inf.Tx.Tx, http.StatusInternalServerError, nil, errors.New("getting SSL keys expiration information from Traffic Vault: "))
+		return
+	}
+	api.WriteResp(w, r, expirationInfos)
```
ocket8888, line 169: The type of the response for this is `null | [ExpirationInfo, ...ExpirationInfo[]]`, I think it'd be better if `GetExpirationInformation` returned empty lists instead of `nil` and then we'd just have `ExpirationInfo[]`

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -156,6 +158,69 @@ func (p *Postgres) GetDeliveryServiceSSLKeys(xmlID string, version string, tx *s
...
+		fedList = append(fedList, fedString)
+	}
+
+	query := "SELECT deliveryservice, cdn, provider, expiration FROM sslkey WHERE version='latest'"
+
+	if days != "" {
+		query = query + " AND expiration <= (now() + '" + days + " days'::interval)"
+	}
+
+	var expirationInfos []tc.SSLKeyExpirationInformation
```
ocket8888, line 190: This is what I'm talking about right here; the "zero value" of a slice is `nil`, so if you don't find any expiration key info within the selected interval or where `version` is `'latest'`, then this function returns `nil` in its first return value.

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -156,6 +158,69 @@ func (p *Postgres) GetDeliveryServiceSSLKeys(xmlID string, version string, tx *s
...
+		expirationInfo.Federated = util.BoolPtr(false)
+		for _, fed := range fedList {
+			if fed == *expirationInfo.DeliveryService {
+				expirationInfo.Federated = util.BoolPtr(true)
+			}
+		}
+
+		dsID, _, ok, err := dbhelpers.GetDSIDAndCDNFromName(tx, *expirationInfo.DeliveryService)
+		if err != nil {
+			return []tc.SSLKeyExpirationInformation{}, false, fmt.Errorf("deliveryservice.GenerateLetsEncryptCertificates: getting DS ID from name: %v", err)
```
ocket8888, line 212: constructing errors from errors should wrap with `%w`

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -181,8 +246,17 @@ func (p *Postgres) PutDeliveryServiceSSLKeys(key tc.DeliveryServiceSSLKeys, tx *
 		return errors.New("encrypting keys: " + err.Error())
 	}
 
+	err = deliveryservice.Base64DecodeCertificate(&key.Certificate)
+	if err != nil {
+		return errors.New(fmt.Sprintf("decoding SSL keys, %s", err.Error()))
+	}
+	expiration, _, err := deliveryservice.ParseExpirationAndSansFromCert([]byte(key.Certificate.Crt), key.Hostname)
+	if err != nil {
+		return errors.New("parsing expiration from certificate: " + err.Error())
```
ocket8888, line 255: should use `fmt.Errorf(...)` instead of `errors.New(fmt.Sprintf(...))`

docs/source/api/v4/sslkey_expirations.rst
mattjackson220: so i originally was only going to return the XMLID but in TP it needs the id to navigate to the SSL Key page when the row gets clicked so i returned both so TP can show the XMLID but navigate to the page

ocket8888: Well if that's really easier, then can't we just stick with `deliveryservice: number` and drop the XMLID? Alternatively, TP can just fetch DSes at the same time by just adding it to the `resolve` in the index.js and then use that to construct a map of XMLID to ID (which could also be done right in the `resolve` function) that can be used to construct the location on navigation, since that's a click handler and not a link it's really easy to do.

mattjackson220: 👍

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -156,6 +157,61 @@ func (p *Postgres) GetDeliveryServiceSSLKeys(xmlID string, version string, tx *s
 	return sslKey, true, nil
 }
 
+// GetExpirationInformation returns the expiration information for all SSL Keys.
+func (p *Postgres) GetExpirationInformation(tx *sql.Tx, ctx context.Context, days string) ([]tc.SSLKeyExpirationInformation, bool, error) {
```
ocket8888, line 161: This never returns `false` for the second value with a `nil` error in the third value, and it never returns `true` except when the third value _is_ `nil`. So what information is that second return value meant to convey?

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -181,8 +237,17 @@ func (p *Postgres) PutDeliveryServiceSSLKeys(key tc.DeliveryServiceSSLKeys, tx *
 		return errors.New("encrypting keys: " + err.Error())
 	}
 
+	err = deliveryservice.Base64DecodeCertificate(&key.Certificate)
+	if err != nil {
+		return fmt.Errorf("decoding SSL keys, %w", err)
+	}
+	expiration, _, err := deliveryservice.ParseExpirationAndSansFromCert([]byte(key.Certificate.Crt), key.Hostname)
+	if err != nil {
+		return errors.New("parsing expiration from certificate: " + err.Error())
```
ocket8888, line 246: building errors from other errors should wrap them like you did 4 lines above

lib/go-tc/deliveryservice_ssl_keys.go
```diff
@@ -93,6 +93,15 @@ type DeliveryServiceSSLKeysV15 struct {
 	Expiration time.Time `json:"expiration,omitempty"`
 }
 
+// SSLKeyExpirationInformation contains information about an SSL key's expiration.
+type SSLKeyExpirationInformation struct {
+	DeliveryService string    `json:"deliveryservice"`
+	Cdn             string    `json:"cdn"`
```
ocket8888, line 99: nit: initialism should be `CDN` not `Cdn`

lib/go-tc/deliveryservice_ssl_keys.go
```diff
@@ -93,6 +93,15 @@ type DeliveryServiceSSLKeysV15 struct {
...
+	Provider        *string   `json:"provider"`
```
ocket8888, line 100: This is still a reference, and the underlying database allows null values for `provider`, but in practice they're all empty strings:
```plpgsql
$ SELECT deliveryservice, cdn, provider, expiration FROM sslkey WHERE provider IS NULL;
 deliveryservice | cdn | provider | expiration 
-----------------+-----+----------+------------
(0 rows)

$ SELECT deliveryservice, cdn, provider, expiration FROM sslkey WHERE provider = '';
 deliveryservice |     cdn      | provider |       expiration       
-----------------+--------------+----------+------------------------
 demo1           | CDN-in-a-Box |          | 2022-03-17 19:10:49+00
 demo1           | CDN-in-a-Box |          | 2022-03-17 19:10:49+00
 demo2           | CDN-in-a-Box |          | 2022-03-17 19:10:49+00
 demo2           | CDN-in-a-Box |          | 2022-03-17 19:10:49+00
(4 rows)
```

Should that be inserting `NULL` instead of `''`? Or maybe that column should have a `NOT NULL` constraint instead?

mattjackson220: i had it this way because when the DB migrate is run it has `NULL` in there and errored out if i had the `NOT NULL` requirement but after the `fill_expiration_and_provider` script is run then it should all be filled in or empty string. is there a better way to do that? im game to change it to all `NULL` but if theres a better way id prefer that stay as empty string

mattjackson220: i was hoping to not add another migration but thats the right way to go. updating it.

ocket8888 (approved):
Seems to be working fine

Issue #2526, opened 2018-07-10
RFE: New Calendar view for SSL Certificate Expiration


jhg03a:
As a SRE I want to be able to click a link in traffic portal and see a calendar with a list of upcoming dates for which HTTPS enabled delivery services will have their certificates expire.  As a secondary goal, I would like to have a list or api endpoint that produces a list of expired SSL certificates.  As a third goal, I would also like the option to download this information in ics format for use in other tools such as gcal or outlook.  This would make staying on top of SSL certificate expirations much easier so manual records or 3rd party notifications would no longer be required and prevent expirations from occurring in production.

This data is all presently available in trafficvault and could be reconstituted via api+scripting, but I'd prefer to see it inside TP itself.

mitchell852:
mind if i close @jhg03a ? probably not going to happen and there are probably better tools for this.

jhg03a:
There actually isn't a better tool I can find.  Only TO knows what those cert expiration dates are without constant exporting and potential private key disclosure.

jhg03a:
Unless you are 100% LE with no vanity DS url, you have this usability problem.

rob05c:
> In the future, ssl certs will be a Permission, not admin-only, so that filtering will be exposed

Will we need 2 Permissions: certs-and-keys, and only-public-certs? In fact, I think an only-public-certs would be safe to let anyone read, that should probably be added to a "default" Permission set if such a thing exists.

ocket8888:
Sounds like too much work for TPv1 IMO, could be a table in that version and then we revamp it later

PR #6339, opened 2021-11-05
Cert expirations tp

mattjackson220:
Closes: #2526 

Requires PR #6337 and PR #6338 to be merged first!

This PR creates a page in Traffic Portal with a table of all SSL certificate expirations.

<hr>

- Traffic Portal

Run Traffic Portal and verify that the Certificate Expirations button is available in the navigation bar on the left side under Services
Verify that the page loads successfully and shows the correct SSL cert expiration data
Verify that column filtering and searching works
Verify that the `Days` search works

## PR submission checklist
- [x] This PR has tests 


traffic_portal/app/src/app.js
```diff
@@ -372,6 +374,7 @@ var trafficPortal = angular.module('trafficPortal', [
         require('./common/modules/table/cdnNotifications').name,
         require('./common/modules/table/cdnProfiles').name,
         require('./common/modules/table/cdnServers').name,
+		require('./common/modules/table/certExpirations').name,
```
ocket8888, line 377: the rest of the file (unfortunately) uses spaces for indentation, as do your other two added lines. If you wanna change the indentation of the file to use actual indentation, I'd support that fully. As-is, though, this should conform.

traffic_portal/app/src/common/api/index.js
```diff
@@ -24,7 +24,8 @@ module.exports = angular.module('trafficPortal.api', [])
     .service('cacheStatsService', require('./CacheStatsService'))
     .service('capabilityService', require('./CapabilityService'))
     .service('cdnService', require('./CDNService'))
-    .service('changeLogService', require('./ChangeLogService'))
+	.service('certExpirationsService', require('./CertExpirationsService'))
+	.service('changeLogService', require('./ChangeLogService'))
```
ocket8888, line 28: same as above RE: indentation

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
```diff
@@ -0,0 +1,317 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+var TableCertExpirationsController = function(tableName, certExpirations, filter, $document, $scope, $state, $filter, locationUtils, certExpirationsService) {
+
+    let table;
```
ocket8888, line 22: indentation is inconsistent in this file, some lines uses spaces and some lines use the One True indentation. There's no mandate on which to use throughout TP, but you should pick one to stick with for the file to prevent unintentionally massive diffs if the file is ever changed again by someone who's editor consolidates the indentation to a single style (fwiw tabs are mandated in TPv2, so it's a solid choice IMO).

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
```diff
@@ -0,0 +1,317 @@
...
+			field: "federated",
+			hide: false,
+			valueFormatter: federatedCellFormatter
+		},
+	];
+
+	$scope.pageSize = 100;
+
+	/** Options, configuration, data and callbacks for the ag-grid table. */
+	$scope.gridOptions = {
```
ocket8888, line 93: It looks like you might be doing more work here than you need to. Can you make use of the existing AG-Grid table component?

traffic_portal/test/integration/PageObjects/SideNavigationPage.po.ts
```diff
@@ -36,6 +36,7 @@ export class SideNavigationPage extends BasePage{
     private lnkDeliveryServices = element(by.xpath("//a[@href='/#!/delivery-services']"));
     private lnkDeliveryServiceRequest = element(by.xpath("//a[@href='/#!/delivery-service-requests']"));
     private lnkServiceCategories = element(by.xpath("//a[@href='/#!/service-categories']"));
+    private lnkCertExpirations = element(by.xpath("//a[@href='/#!/cert-expirations']"));
```
ocket8888, line 39: CSS selectors are much more familiar and easy to read than XPath, should use those instead

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
mattjackson220: i tried that route originally but i couldnt get the `Days` input to be with the other inputs or to get it to update the table correctly with the new information. i can try again next week on that but i got the other updates in there

traffic_portal/app/src/common/modules/table/certExpirations/table.certExpirations.tpl.html
```diff
@@ -0,0 +1,70 @@
...
+            <li class="active">Certificate Expirations</li>
+        </ol>
+        <div class="pull-right">
+            <div class="form-inline" role="search">
+                <input id="quickSearch" name="quickSearch" type="search" class="form-control text-input" placeholder="Quick search..." ng-model="quickSearch" ng-change="onQuickSearchChanged()" aria-label="Search"/>
+                <div class="input-group text-input">
+                    <span class="input-group-addon">
+                        <label for="days">Days</label>
+                    </span>
+                    <input id="days" name="days" type="number" min="1" class="form-control" placeholder="" ng-model="days" ng-change="onDaysChanged(days)" aria-label="Days"/>
```
ocket8888, line 32: You don't need `aria-label` since the element is explicitly labeled. An empty `placeholder` is the default, so it need not be specified.

ocket8888: Passing `days` to the `onDaysChanged` expands expressions of non-constant variables to scope properties; this is calling `$scope.onDaysChanged($scope.days)`. This is a bit of a nitpick, but since `$scope.days` is in-scope of the function `$scope.onDaysChanged`, you don't need to pass it an argument, you can just access the model value directly.

mattjackson220: i removed the `days` aspect completely and used the common ag grid instead

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
```diff
@@ -17,27 +17,24 @@
...
-var TableCertExpirationsController = function(tableName, certExpirations, dsXmlToIdMap, filter, $document, $scope, $state, $filter, locationUtils, certExpirationsService) {
-
-	let table;
+var TableCertExpirationsController = function(tableName, certExpirations, dsXmlToIdMap, filter, $document, $scope, $state, $filter, locationUtils) {
+
+	/** All of the expiration fields converted to actual Dates */
+	$scope.certExpirations = certExpirations.map(
+		function(x) {
+			// need to convert this to a date object for ag-grid filter to work properly
+			x.expiration = new Date(x.expiration.replace("+00", "Z"));
```
ocket8888, line 26: The timestamps returned by the API are RFC3339 with nanosecond precision, so this replacement is unnecessary.

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
```diff
@@ -0,0 +1,94 @@
...
+
+	/**
+	 * Formats the contents of a 'federation' column cell as "True" or blank for visibility.
+	 */
+	function federatedCellFormatter(params) {
+		if (!params.value) {
+			return '';
+		} else {
+			return params.value;
+		}
```
ocket8888, line 44: This will make it always 'true' or blank - shouldn't it be 'true' or 'false'?

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
```diff
@@ -0,0 +1,94 @@
...
+			hide: false,
+			valueFormatter: federatedCellFormatter
+		},
+	];
+
+	/** Options, configuration, data and callbacks for the ag-grid table. */
+	$scope.gridOptions = {
+		onRowClick: function(params) {
+			const selection = window.getSelection().toString();
+			if(selection === "" || selection === $scope.mouseDownSelectionText) {
```
ocket8888, line 82: `$scope.mouseDownSelectionText` is either undefined or the empty string, and `window.getSelection.toString()` will always return a string. when `$scope.mouseDownSelectionText` is an empty string, the second part of this condition is redundant. `undefined` and `string` types have no overlap, so this condition will always evaluate to `false` in the case that `$scope.mouseDownSelectionText` is `undefined`.. Furthermore, the empty string is the only false-y string value, so all in all this condition can be written as simply `if (!selection) {`.

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
```diff
@@ -0,0 +1,94 @@
...
+				locationUtils.navigateToPath('/delivery-services/' + $scope.dsXmlToIdMap[params.data.deliveryservice] + '/ssl-keys');
+				// Event is outside the digest cycle, so we need to trigger one.
+				$scope.$apply();
+			}
+			$scope.mouseDownSelectionText = "";
```
ocket8888, line 87: Since this value is never (usefully) read, this line can be removed.

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+var TableCertExpirationsController = function(tableName, certExpirations, dsXmlToIdMap, filter, $document, $scope, $state, $filter, locationUtils) {
```
ocket8888, line 20: `filter`, `$document`, `$state`, and `$filter` are all unused

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
```diff
@@ -0,0 +1,94 @@
...
+				// Event is outside the digest cycle, so we need to trigger one.
+				$scope.$apply();
+			}
+			$scope.mouseDownSelectionText = "";
+		}
+	};
+
+};
+
+TableCertExpirationsController.$inject = ['tableName', 'certExpirations', 'dsXmlToIdMap', 'filter', '$document', '$scope', '$state', '$filter', 'locationUtils'];
```
ocket8888, line 93: same as above RE unused dependency injections.

traffic_portal/app/src/modules/private/certExpirations/list/index.js
```diff
@@ -0,0 +1,55 @@
...
+						resolve: {
+							certExpirations: function(certExpirationsService) {
+								return certExpirationsService.getCertExpirations();
+							},
+							tableName: function() {
+								return 'cert-expirations';
+							},
+							filter: function() {
+								return null;
+							},
```
ocket8888, line 38: I'm not sure what this is for, but it's never used in the controller for this state, so it should just be removed.

traffic_portal/app/src/modules/private/certExpirations/list/index.js
```diff
@@ -0,0 +1,55 @@
...
+							deliveryservices: function(deliveryServiceService) {
+								return deliveryServiceService.getDeliveryServices();
+							},
+							dsXmlToIdMap: function (deliveryservices) {
+								var result = {};
+								for (i in deliveryservices) {
+									result[deliveryservices[i].xmlId] = deliveryservices[i].id;
+								}
+								return result;
+							}
```
ocket8888, line 48: This doesn't represent a state; it's pure functional logic. This should probably be moved to the controller instead of injected as a dependency thereof.

Also, `result` is never reassigned, so it could be `const`, and can in fact be constructed as a mapping on the `deliveryservices` collection like so:

```javascript
$scope.dsXMLIDToIDMap = Object.fromEntries(deliveryservices.map(ds=>[ds.xmlId, ds.id]));
```

It can also be made into an ES6 Map in much the same way, if that interests you at all:
```javascript
$scope.dsXMLIDToIDMap = new Map(deliveryservices.map(ds=>[ds.xmlId, ds.id]));
```

traffic_portal/test/integration/PageObjects/SideNavigationPage.po.ts
```diff
@@ -157,6 +158,11 @@ export class SideNavigationPage extends BasePage{
         await browser.actions().mouseMove(this.lnkServiceCategories).perform();
         await browser.actions().click(this.lnkServiceCategories).perform();
     }
+    async NavigateToCertExpirationsPage(){
+        await browser.wait(ExpectedConditions.visibilityOf(this.lnkCertExpirations), 2000);
+        await browser.actions().mouseMove(this.lnkCertExpirations).perform();
+        await browser.actions().click(this.lnkCertExpirations).perform();
```
ocket8888, line 164: I don't think the part about moving the mouse is under test here, so we can just do this on one line with
```
return this.lnkCertExpirations.click();
```
This method is also (like most others) missing a return type (should be `Promise<void>`), and IMO ought to have a visibility modifier (`public`) and jsDoc comment (but neither of those are enforced, so don't feel like you have to do them - return type though is something I've been requesting on all callables, if not actually enforced).

ocket8888 (commented):
Page visually looks good, all UX controls seem to be working

traffic_portal/app/src/common/modules/table/certExpirations/table.certExpirations.tpl.html
```diff
@@ -0,0 +1,23 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<div class="x_panel">
+    <common-grid-controller title="Certificate Expirations" table-name="certExpirations" options="gridOptions" data="certExpirations"
```
shamrickus, line 21: Just merged [6415](https://github.com/apache/trafficcontrol/pull/6415), `title` needs to be `table-title` now.

traffic_portal/app/src/common/modules/table/certExpirations/TableCertExpirationsController.js
mattjackson220: yea i put it as 'true' or blank because when it was 'true' or 'false' it was not super clear at just a glance. so this way its easy to see quickly

ocket8888: Okay, that's fine with me. You could also change it to a checkmark or something if you like that better - I think the cache checks page has an example of that - but as far as I'm concerned this is fine.

traffic_portal/app/src/modules/private/certExpirations/list/index.js
```diff
@@ -32,19 +32,6 @@ module.exports = angular.module('trafficPortal.private.certExpirations.list', []
 							},
 							tableName: function() {
 								return 'cert-expirations';
-							},
-							filter: function() {
-								return null;
-							},
-							deliveryservices: function(deliveryServiceService) {
-								return deliveryServiceService.getDeliveryServices();
-							},
```
ocket8888: this is most of what `resolve` values are for: you get a promise that asynchronously outputs data, and Angular handles resolving it to an actual value injected into your controller at component initialization time. If you keep `deliveryservices` here, you can input it as an injected dependency like it used to be, and instead of a function like
```javascript
function init() {
    doSomething();
}
init();
```
(which, incidentally is in no way any different than just calling `doSomething` immediately - the only real differences are `this` binding and the values of some global variables that no one should ever use like `name` and `arguments`)

you can just directly construct the map immediately like:
```javascript
	$scope.dsXmlToIdMap = new Map(deliveryservices.map(ds=>[ds.xmlId, ds.id]));
```
no extra function wrapping or deferred calling or `.then`-ing a promise necessary at all.

traffic_portal/test/integration/PageObjects/SideNavigationPage.po.ts
```diff
@@ -158,10 +158,9 @@ export class SideNavigationPage extends BasePage{
         await browser.actions().mouseMove(this.lnkServiceCategories).perform();
         await browser.actions().click(this.lnkServiceCategories).perform();
     }
-    async NavigateToCertExpirationsPage(){
-        await browser.wait(ExpectedConditions.visibilityOf(this.lnkCertExpirations), 2000);
-        await browser.actions().mouseMove(this.lnkCertExpirations).perform();
-        await browser.actions().click(this.lnkCertExpirations).perform();
+    // NavigateToCertExpirationsPage verifies that the link to the Certificate Expirations page is clickable.
```
ocket8888, line 161: JSDoc comments need to be started with `/**` (and therefore ended with `*/`) in order to meet the spec. IDEs and the `jsdoc` program that they use won't pick it up otherwise.

ocket8888 (approved):

PR #6341, opened 2021-11-08
Changes the t3c process locking so that no two simultaneous t3c instances run.

jrushford:
Changes the t3c process locking so that no two simultaneous t3c instances run. The locking now occurs before t3c attempts logging into traffic ops.

<hr>

Try to start two or more t3c-apply processes on the same machine.  Only one will run, the others should
exit with a log message indicating another t3c-apply process is already running.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/cache-config/t3c-apply/t3c-apply.go
+++ b/cache-config/t3c-apply/t3c-apply.go
@@ -69,6 +69,9 @@ func main() {
 	} else if cfg == (config.Cfg{}) { // user used the --help option
 		os.Exit(Success)
 	}
+	if !lock.GetLock("/var/run/t3c.lock") {
+		os.Exit(AlreadyRunning)
+	}
 
 	if cfg.UseGit == config.UseGitYes {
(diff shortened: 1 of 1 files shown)


cache-config/t3c-apply/t3c-apply.go
```diff
@@ -69,6 +69,9 @@ func main() {
 	} else if cfg == (config.Cfg{}) { // user used the --help option
 		os.Exit(Success)
 	}
+	if !lock.GetLock("/var/run/t3c.lock") {
```
rob05c, line 72: Does this need a `if !cfg.ReportOnly {` check like the old code had, to allow running in Report Mode while syncds runs?

jrushford: Not in my opinion.  The idea with the locking is limit t3c processes and logins to TO.  Why allow a t3c process running in ReportOnly, it still has to login to TO.

rob05c (approved):
LGTM. Tests pass, code looks good

PR #6346, opened 2021-11-10
Fix TO login DB contexts to inherit from the request context

rawlinp:
By doing this, the DB queries will be cancelled if the client closes its
connection in the middle of request handling.

<hr>

- Traffic Ops

Log in like you normally would, make sure it works and sets the `last_authenticated` field properly still.

- 5.x
- 6.x

## PR submission checklist
- [x] This PR does not add tests b/c login is fairly well-tested already, and testing the specific behavior of cancelling DB queries if the client closes their connection early is quite difficult. However, this is how the vast majority of our DB transaction contexts are created/used and is fairly ubiquitous.


mattjackson220 (approved):
looks good to me! code looks good, tests pass, tested locally and works as expected

Issue #6348, opened 2021-11-12
traffic_vault_migrate does not handle `-` correctly.


shamrickus:
## This Bug Report affects these Traffic Control components:

- Tools

## Current behavior:
If a CDN has a `-` in it's name, the SSL key version will be incorrectly extracted from the Riak database; which will then insert it into the destination DB without error (excluding duplicates which is a very common case).

## Expected behavior:
All valid CDN/DS names should be supported without issue.

## Steps to reproduce:
While using Riak, create CDN with a `-` in the name, create a corresponding DS with SSL keys. Run the `traffic_vault_migrate` tool which will then insert the bad version data into postgres if there are no duplicates.

PR #6353, opened 2021-11-15
Handle `-` correctly in `traffic_vault_migrate`

shamrickus:
Closes: #6348
Also changes it so when a orphaned key is found in Riak, a warning is output instead of halting with an error.

<hr>

- Tools

Using Riak TV, create a CDN with a `-` in the name. Create child DS with SSL keys. Run the `traffic_vault_migrate` tool and either the version in postgres will be incorrect, or the insert query will fail due to duplicate inserts.

- 6.0.1
- master

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/app/db/traffic_vault_migrate/riak.go
+++ b/traffic_ops/app/db/traffic_vault_migrate/riak.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/basho/riak-go-client"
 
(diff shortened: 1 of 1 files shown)


mattjackson220 (approved):
looks good to me! code looks good and tests work as expected! dashes are handled correctly!

PR #6357, opened 2021-11-17
Add t3c retrying app lock

rob05c:
Adds t3c retrying the app lock.

This allows operators to run multiple t3c instances on timers, such as for syncds and reval modes, and have them succeed when run around the same time.

It retries the lock every second for a minute. We can make that configurable in the future, but there are already a lot of config options, and if checking a file every second is a performance issue, or if t3c-apply takes more than a minute to run on even a very large CDN, the system is almost certainly in a very bad state. At this point we have so many options, I'd prefer to avoid adding new ones until necessary.

Includes tests.
Includes changelog.
No docs, no interface change, lockfile behavior isn't documented.

Run tests. Run multiple t3c-apply instances, verify the second runs after the first finishes.

Not a bug fix.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/cache-config/testing/ort-tests/t3c-lockfile_test.go
+++ b/cache-config/testing/ort-tests/t3c-lockfile_test.go
@@ -33,7 +33,7 @@ func TestLockfile(t *testing.T) {
 		tcdata.CacheGroups, tcdata.Servers, tcdata.Topologies,
 		tcdata.DeliveryServices}, func() {
 
-		hostName := "atlanta-edge-03"
+		const hostName = `atlanta-edge-03`
 		const fileName = `records.config`
 
 		firstOut := ""
--- a/cache-config/testing/ort-tests/t3c-lockfile_test.go
+++ b/cache-config/testing/ort-tests/t3c-lockfile_test.go
@@ -52,12 +52,12 @@ func TestLockfile(t *testing.T) {
 		out, code := t3cUpdateReload(hostName, "badass")
 		t.Logf("TestLockFile second t3c finished %v", time.Now())
 		if code != 0 {
-			t.Fatalf("second t3c apply badass failed: output '''%v''' code %v\n", out, code)
+			t.Fatalf("second t3c apply badass failed: output '''%v''' code %v", out, code)
 		}
 
 		wg.Wait()
 		if firstCode != 0 {
-			t.Fatalf("first t3c apply badass failed: output '''%v''' code %v\n", firstOut, firstCode)
+			t.Fatalf("first t3c apply badass failed: output '''%v''' code %v", firstOut, firstCode)
 		}
 
 		outStr := string(out)
--- a/cache-config/testing/ort-tests/t3c-lockfile_test.go
+++ b/cache-config/testing/ort-tests/t3c-lockfile_test.go
@@ -42,22 +42,22 @@ func TestLockfile(t *testing.T) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
-			t.Logf("TestLockFile first t3c starting %v", time.Now())
+			t.Logf("TestLockFile first t3c starting %s", time.Now())
 			firstOut, firstCode = t3cUpdateReload(hostName, "badass")
-			t.Logf("TestLockFile first t3c finished %v", time.Now())
+			t.Logf("TestLockFile first t3c finished %s", time.Now())
 		}()
 
 		time.Sleep(time.Millisecond * 100) // sleep long enough to ensure the concurrent t3c starts
-		t.Logf("TestLockFile second t3c starting %v", time.Now())
+		t.Logf("TestLockFile second t3c starting %s", time.Now())
 		out, code := t3cUpdateReload(hostName, "badass")
-		t.Logf("TestLockFile second t3c finished %v", time.Now())
+		t.Logf("TestLockFile second t3c finished %s", time.Now())
 		if code != 0 {
-			t.Fatalf("second t3c apply badass failed: output '''%v''' code %v", out, code)
+			t.Fatalf("second t3c apply badass failed: output '''%s''' code %d", out, code)
 		}
 
 		wg.Wait()
 		if firstCode != 0 {
-			t.Fatalf("first t3c apply badass failed: output '''%v''' code %v", firstOut, firstCode)
+			t.Fatalf("first t3c apply badass failed: output '''%s''' code %d", firstOut, firstCode)
 		}
 
 		outStr := string(out)
@@ -76,7 +76,7 @@ func TestLockfile(t *testing.T) {
 		}
 
 		if acquireStartLine == "" || acquireEndLine == "" {
-			t.Fatalf("t3c apply output expected to contain 'Trying to acquire app lock' and 'Acquired app lock', actual: '''%v'''", out)
+			t.Fatalf("t3c apply output expected to contain 'Trying to acquire app lock' and 'Acquired app lock', actual: '''%s'''", out)
 		}
 
 		acquireStart := parseLogLineTime(acquireStartLine)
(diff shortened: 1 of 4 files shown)


ezelkow1 (approved):

cache-config/testing/ort-tests/t3c-lockfile_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/apache/trafficcontrol/cache-config/testing/ort-tests/tcdata"
+)
+
+func TestLockfile(t *testing.T) {
+	testName := "TestLockfile"
+	t.Logf("------------- Starting " + testName + " ---------------")
```
ocket8888, line 28: Is this (and the corresponding ending line) necessary? `go test` prints the name of the test when that test starts when you use `-v` or when the test fails (which are the only times you'll see `testing.T.Log` output anyway).

cache-config/testing/ort-tests/t3c-lockfile_test.go
```diff
@@ -0,0 +1,113 @@
...
+		if acquireEnd == nil {
+			t.Fatalf("t3c apply acquire line failed to parse time, line '" + acquireEndLine + "'")
+		}
+
+		minDiff := time.Second * 1 // checking the file lock should never take 1s, so that's enough to verify it was hit
+		if diff := acquireEnd.Sub(*acquireStart); diff < minDiff {
+			t.Fatalf("t3c apply expected time to acquire while another t3c is running to be at least %v, actual %v start line '%v' end line '%v'", minDiff, diff, acquireStartLine, acquireEndLine)
+		}
+
+		t.Logf(testName + " succeeded")
```
ocket8888, line 96: Is this necessary? The output of `go test` informs you whether the test succeeded or failed

cache-config/testing/ort-tests/t3c-lockfile_test.go
```diff
@@ -0,0 +1,113 @@
...
+			firstOut, firstCode = t3cUpdateReload(hostName, "badass")
+			t.Logf("TestLockFile first t3c finished %v", time.Now())
+		}()
+
+		time.Sleep(time.Millisecond * 100) // sleep long enough to ensure the concurrent t3c starts
+		t.Logf("TestLockFile second t3c starting %v", time.Now())
+		out, code := t3cUpdateReload(hostName, "badass")
+		t.Logf("TestLockFile second t3c finished %v", time.Now())
+		if code != 0 {
+			t.Fatalf("second t3c apply badass failed: output '''%v''' code %v\n", out, code)
```
ocket8888, line 55: I have two nits here:
- you don't need the `\n`, the logging makes sure it ends with a newline character for you
- using `%v` instead of a more specific formatting parameter for primitive types/`fmt.Stringer`s makes it harder to catch obscure issues like printing a memory address instead of the value stored there, so I encourage people to be more specific when possible. Not that you're printing pointer values on this particular line, or anything. Just a good practice IMO.

cache-config/testing/ort-tests/t3c-lockfile_test.go
```diff
@@ -0,0 +1,113 @@
...
+	t.Logf("------------- Starting " + testName + " ---------------")
+	tcd.WithObjs(t, []tcdata.TCObj{
+		tcdata.CDNs, tcdata.Types, tcdata.Tenants, tcdata.Parameters,
+		tcdata.Profiles, tcdata.ProfileParameters, tcdata.Statuses,
+		tcdata.Divisions, tcdata.Regions, tcdata.PhysLocations,
+		tcdata.CacheGroups, tcdata.Servers, tcdata.Topologies,
+		tcdata.DeliveryServices}, func() {
+
+		hostName := "atlanta-edge-03"
+		const fileName = `records.config`
```
ocket8888, line 37: why is one of these `const` but not the other? The both look like they aren't modified.

cache-config/testing/ort-tests/t3c-lockfile_test.go
```diff
@@ -0,0 +1,113 @@
...
+			defer wg.Done()
+			t.Logf("TestLockFile first t3c starting %v", time.Now())
+			firstOut, firstCode = t3cUpdateReload(hostName, "badass")
+			t.Logf("TestLockFile first t3c finished %v", time.Now())
+		}()
+
+		time.Sleep(time.Millisecond * 100) // sleep long enough to ensure the concurrent t3c starts
+		t.Logf("TestLockFile second t3c starting %v", time.Now())
+		out, code := t3cUpdateReload(hostName, "badass")
+		t.Logf("TestLockFile second t3c finished %v", time.Now())
```
ocket8888, line 53: Should timing stuff be in a benchmark?

cache-config/testing/ort-tests/t3c-lockfile_test.go
rob05c: This is following the existing convention, e.g. https://github.com/apache/trafficcontrol/blob/master/cache-config/testing/ort-tests/to_requester_test.go#L36
I didn't create the convention, and I agree, but it seemed better to follow the convention as long as it exists.

cache-config/testing/ort-tests/t3c-lockfile_test.go
rob05c: I added that line to distinguish the `tcd.WithObjs(t, []tcdata.TCObj{` block getting to the end successfully, versus the outer test itself. I found it very useful when working on the test. I can remove it; but it exists because it was very useful in-practice writing, fixing, and figuring out why and where the test was failing.

cache-config/testing/ort-tests/t3c-lockfile_test.go
rob05c: I removed the newline.

I don't agree about `%v`. In my professional opinion, "stronger types" in format strings make the code more brittle and likely to break without actually gaining anything. Unless the format string is used for machine-read data, which is rare, there's little value in forcing it to break when the type changes. For pointers in particular, it's sometimes valuable to print the address, so it's valuable to make it explicit to readers that the pointer or value is being printed, by passing the pointer or value explicitly. Calling `fmt.Sprintf("%d", pointer)` will simply print it as decimal instead of hex, it neither fails `go vet` nor fails to compile.

cache-config/testing/ort-tests/t3c-lockfile_test.go
rob05c: Changed.

cache-config/testing/ort-tests/t3c-lockfile_test.go
rob05c: This isn't a benchmark, it's testing functional behavior. The logs here are for debugging the test when it fails, making it easier to figure out why and how it failed.

The test is to verify a t3c run while another t3c is running waits and then runs, rather than either not running or wrongly running at the same time. It's not benchmarking the run.

We could add benchmarks, but that would be more work, and unrelated to this PR.

cache-config/testing/ort-tests/t3c-lockfile_test.go
rawlinp: +1 on the nitpick (more specific types used in format strings). IMO it's tremendously more likely for a developer to accidentally print a pointer address when they really meant to print the actual value, than it is that a type changed and needs a changed format string in various places. Using specific types forces the developer to actually think about what is getting printed and help nudge them into things like dereferencing the pointer where needed. But, like you said, for things like printing pointers to ints, you can't always rely on the compiler or `go vet`.

That said, it is definitely a nitpick, which by definition, I wouldn't necessarily hold up the PR for.

rob05c: Changed.

cache-config/testing/ort-tests/t3c-lockfile_test.go
ocket8888: I'm not sure I understand the use-case, but it's not a big deal - and any rate there are already some extraneous log lines, so it may just be best addressed all at once.

cache-config/testing/ort-tests/t3c-lockfile_test.go
ocket8888: That's fine, then, I suppose.

ocket8888 (approved):

ocket8888:
I merged past CodeQL because that seems to be having issues today, being queued for hours.

PR #5950, opened 2021-06-16
Add GHA for TM Integration Tests

shamrickus:
## What does this PR (Pull Request) do?

- [x] This PR is not related to any Issue. 

## Which Traffic Control components are affected by this PR?

- CI tests

## What is the best way to verify this PR?
Look at GHA, make a branch and push to test run.

## The following criteria are ALL met by this PR

- [x] This PR includes tests OR I have explained why tests are unnecessary
- [x] This PR includes documentation OR I have explained why documentation is unnecessary
- [x] This PR includes an update to CHANGELOG.md OR such an update is not necessary
- [x] This PR includes any and all required license headers
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #6358, opened 2021-11-18
Traffic Router Ultimate Test Harness blueprint

zrhoffman:
This Pull Request adds a blueprint for the Traffic Router Ultimate Test Harness.

### DIFF
--- a/blueprints/tr-load-tests.md
+++ b/blueprints/tr-load-tests.md
@@ -163,7 +163,7 @@ Instructions for using the Traffic Router Ultimate Test Harness should be added 
         - information about Traffic Router-type Servers in those CDNs
         - Type information
         - Delivery Service information
-    - Documentation of each CLI option
+    - Documentation of each option
     - Example commands
 
 ### Testing Impact
@@ -196,6 +196,26 @@ The Router Ultimate Test Harness should include a load test for HTTP-routed Deli
 | Path count                      | The number of paths to generate for use in requests to Delivery Services                                                                                                                                | HTTP                  | 10000                       |
 | Maximum path length             | The maximum string length for each generated path                                                                                                                                                       | HTTP                  | 100                         |
 | Use location header             | Whether the HTTP HTTP Delivery service should redirect the user or server the routing information as a JSON response.                                                                                   | HTTP                  | True                        |
+
+These options will be structured in a config file:
+
+```jsonc
+{
+    "all": {
+        "cdn_name": "Kabletown"
+    },
+    "http": {
+        "ipv4_only": true,
+        /* more options */
+        "path_count": 5000
+    },
+    "dns": {
+        "delivery_service_name": "static"
+    }
+}
+```
+
+Options that should apply to a specific type of test should go under a key named that test type, while options applying to all tests can go under the `"all"` key.
 
 #### Other Tests
 
--- a/blueprints/tr-load-tests.md
+++ b/blueprints/tr-load-tests.md
@@ -65,7 +65,14 @@ No Traffic Portal impact is anticipated.
 <!--
 *How* will this impact Traffic Ops (at a high level)?
 -->
-No Traffic Ops impact is anticipated.
+The Operations-level TR Ultimate Test Harness user will update the status of any Traffic Router that is currently undergoing load testing to `"OFFLINE"` in Traffic Ops. It should also update the status of that Traffic Router to its previous status once the load testing is complete, regardless of whether the test failed or succeeded.
+
+The Operations-level TR Ultimate Test Harness user will also be used to get
+* CDN snapshots
+* CDN information
+* information about Traffic Router-type Servers in those CDNs
+* Type information
+* Delivery Service information
 
 #### REST API Impact
 <!--
@@ -102,7 +109,7 @@ How will existing tables and columns be changed?
 What are the column data types and modifiers?
 What are the FK references and constraints?
 -->
-No Data Model impact is anticipated.
+A database migration should be added to seed a new Operations-level user for use with the TR Ultimate Test Harness.
 
 ### Cache Config Impact
 <!--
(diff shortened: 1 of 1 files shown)


mitchell852:
@zrhoffman - do you envision the TRUTH :) will someday run via a GitHub Action triggered by any code changes to TR and will fail if those changes negatively impact TR performance?

zrhoffman:
> @zrhoffman - do you envision the TRUTH :) will someday run via a GitHub Action triggered by any code changes to TR and will fail if those changes negatively impact TR performance?

Yes, although only if we can get that GHA to consistently meet a chosen threshold. @shamrickus added a similar GHA for Traffic Monitor in #5950, but because the performance of GitHub Actions runners is so variable, it only passed 14 of 20 times in the last 20 commits to the [`master`](https://github.com/apache/trafficcontrol/commits/master) branch.

So if we can make a TRUTH GHA consistent, then yes. Otherwise, a GHA would have to wait.

zrhoffman:
> @zrhoffman - do you envision the TRUTH :) will someday run via a GitHub Action triggered by any code changes to TR and will fail if those changes negatively impact TR performance?

*Automation Impact* section added in 4d0e5594c2.

limited:
Hey @zrhoffman - 
  I fully support improving performance testing of TR. I've used both `wrk` and `flamethrower` before to decent success. Configuring those tools is definitely a challenge, especially across large configurations. This looks like it will simplify that barrier for many others

blueprints/tr-load-tests.md
```diff
@@ -102,7 +109,7 @@ How will existing tables and columns be changed?
 What are the column data types and modifiers?
 What are the FK references and constraints?
 -->
-No Data Model impact is anticipated.
+A database migration should be added to seed a new Operations-level user for use with the TR Ultimate Test Harness.
```
ocket8888, line 112: Do all install environments really need to support running the test harness? Surely we only want to run it against testing environments? What new Permission would this user need that existing users don't have?

blueprints/tr-load-tests.md
```diff
@@ -0,0 +1,296 @@
...
+*How* will this impact Traffic Ops (at a high level)?
+-->
+The Operations-level TR Ultimate Test Harness user will update the status of any Traffic Router that is currently undergoing load testing to `"OFFLINE"` in Traffic Ops. It should also update the status of that Traffic Router to its previous status once the load testing is complete, regardless of whether the test failed or succeeded.
+
+The Operations-level TR Ultimate Test Harness user will also be used to get
+* CDN snapshots
+* CDN information
+* information about Traffic Router-type Servers in those CDNs
+* Type information
+* Delivery Service information
```
ocket8888, line 75: This sounds more like a step in the testing setup than actual changes to Traffic Ops

blueprints/tr-load-tests.md
```diff
@@ -0,0 +1,296 @@
...
+<!--
+*How* will this impact testing?
+
+What is the high-level test plan?
+How should this be tested?
+Can this be tested within the existing test frameworks?
+How should the existing frameworks be enhanced in order to test this properly?
+-->
+#### Load Tests
+The Router Ultimate Test Harness should include a load test for HTTP-routed Delivery Services and for DNS-routed Delivery Services
```
ocket8888, line 180: Do we care about STEERING/federated Delivery Services?

blueprints/tr-load-tests.md
ocket8888: As users cannot be deleted through the API and any valid installation of ATC follows the setup process that runs `postinstall` which will create a default user, no supported environments have no users, and as of #6324 a Role will always exist that has all Permissions. So why do all new installs/upgrades of ATC need to create a new user just for running tests?

zrhoffman: Good points, removed mention of the *user creation* migration section in f202723c5c.

blueprints/tr-load-tests.md
zrhoffman: True, moved to the *documentation* section in 8e5b1f14d4.

blueprints/tr-load-tests.md
zrhoffman: > Do we care about STEERING

Testing STEERING specifically should not be necessary, but we should at least test HTTP Delivery Services with no Location header (`trred=false`). Added in 7d5fec0e1f

> /federated Delivery Services?

Federations are already mentioned:

https://github.com/apache/trafficcontrol/blob/f202723c5c2475f51c7558c7a83b80ecf6ee9774/blueprints/tr-load-tests.md#L201

blueprints/tr-load-tests.md
```diff
@@ -0,0 +1,296 @@
...
+Will a new field be added to an existing form?
+How will the user interact with the new UI changes?
+-->
+No Traffic Portal impact is anticipated.
+
+### Traffic Ops Impact
+<!--
+*How* will this impact Traffic Ops (at a high level)?
+-->
+The Operations-level TR Ultimate Test Harness user will update the status of any Traffic Router that is currently undergoing load testing to `"OFFLINE"` in Traffic Ops. It should also update the status of that Traffic Router to its previous status once the load testing is complete, regardless of whether the test failed or succeeded.
```
rawlinp, line 68: There seems to be a minor discrepancy between this line and L260, which states that an operator may want to set the TR to OFFLINE manually. Which line is correct? Also, I'm not positive, but I _think_ setting the TR under test to OFFLINE would affect its responses in that it would exclude itself from any responses.

blueprints/tr-load-tests.md
```diff
@@ -0,0 +1,296 @@
...
+
+What is the high-level test plan?
+How should this be tested?
+Can this be tested within the existing test frameworks?
+How should the existing frameworks be enhanced in order to test this properly?
+-->
+#### Load Tests
+The Router Ultimate Test Harness should include a load test for HTTP-routed Delivery Services and for DNS-routed Delivery Services
+
+##### Load Test Options
```
rawlinp, line 181: Are these going to be CLI options? Will there be a config file? Some of these options seem more generic (e.g. CDN name), while some of them are load-test specific (e.g. req/s threshold). If we extend the test harness with more than just load testing, would each different type of test have its own config options and/or file?

blueprints/tr-load-tests.md
zrhoffman: > Also, I'm not positive, but I _think_ setting the TR under test to OFFLINE would affect its responses in that it would exclude itself from any responses.

You're right, setting the TR status to OFFLINE would mean needing a separate ONLINE TR. To avoid that complexity, mention of setting the status to OFFLINE is removed in 9d06008767.

blueprints/tr-load-tests.md
zrhoffman: > Are these going to be CLI options?

No

> Will there be a config file?

Config file structure clarified in c0bd076c46

> If we extend the test harness with more than just load testing, would each different type of test have its own config options and/or file?

Each test can specify its own options, but options applying to more than one test can go under `"all"`.

PR #6361, opened 2021-11-18
Add flags to TV Riak util

TaylorCFrey:
<hr>

- Documentation
- Tools - Traffic Vault Util

Attempt to connect to an instance of Riak with appropriate parameters. If, when used previously, the connection was to ignore TLS configuration, then pass the new `skip_tls=true` flag as a program argument. If TLS is required, then `server_name SERVER_NAME` will be also required to validate the TLS certificates.

Verify the documentation (Tools -> Traffic Vault Util) has the correct and appropriate changes.

## PR submission checklist
- [x] This PR has tests - No tests available against the tool.
- [x] This PR has documentation - Documentation was updated to reflect the changes

### DIFF
--- a/docs/source/tools/traffic_vault_util.rst
+++ b/docs/source/tools/traffic_vault_util.rst
@@ -28,7 +28,7 @@ The ``traffic_vault_util`` tool - located at :file:`tools/traffic_vault_util.go`
 
 Usage
 =====
-``traffic_vault_util [--dry_run] --vault_ip IP --vault_action ACTION [--vault_user USER] [--vault_password PASSWD] [--vault_port PORT] [--skip_tls] [--server_name SERVER_NAME]``
+``traffic_vault_util [--dry_run] --vault_ip IP --vault_action ACTION [--vault_user USER] [--vault_password PASSWD] [--vault_port PORT] [--insecure] [--server_name SERVER_NAME]``
 
 .. option:: --dry_run
 
@@ -69,7 +69,7 @@ Usage
 
 	.. warning:: Although this flag is optional, the utility will not work without it. It will try, but it will fail\ [1]_.
 
-.. option:: --skip_tls
+.. option:: --insecure
 
 	An optional flag which, if given, specifies whether to utilize TLS certificate checks when establishing a connection. Defaults to false.
 
--- a/docs/source/tools/traffic_vault_util.rst
+++ b/docs/source/tools/traffic_vault_util.rst
@@ -28,7 +28,7 @@ The ``traffic_vault_util`` tool - located at :file:`tools/traffic_vault_util.go`
 
 Usage
 =====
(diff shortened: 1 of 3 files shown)


ocket8888:
FWIW, Riak is deprecated, so we're getting rid of this soon. Not that I'm -1 on the change, but it's a change I decided wasn't worth my own time to make, for that reason.

docs/source/tools/traffic_vault_util.rst
```diff
@@ -28,7 +28,7 @@ The ``traffic_vault_util`` tool - located at :file:`tools/traffic_vault_util.go`
 
 Usage
 =====
-``traffic_vault_util [--dry_run] --vault_ip IP --vault_action ACTION [--vault_user USER] [--vault_password PASSWD] [--vault_port PORT]``
+``traffic_vault_util [--dry_run] --vault_ip IP --vault_action ACTION [--vault_user USER] [--vault_password PASSWD] [--vault_port PORT] [--skip_tls] [--server_name SERVER_NAME]``
```
rawlinp, line 31: nit: in most places we just call this `--insecure` instead of `--skip_tls`

docs/source/tools/traffic_vault_util.rst
```diff
@@ -69,4 +69,12 @@ Usage
 
 	.. warning:: Although this flag is optional, the utility will not work without it. It will try, but it will fail\ [1]_.
 
+.. option:: --skip_tls
+
+	An optional flag which, if given, specifies whether to utilize TLS certificate checks when establishing a connection. Defaults to false.
+
+.. option:: --server_name
```
rawlinp, line 76: Is this option really necessary? I.e. if we pass a hostname using `--vault_ip` (as the help text seems to indicate is ok), requiring this `--server_name` option seems redundant.

docs/source/tools/traffic_vault_util.rst
TaylorCFrey: Dig it, done.

docs/source/tools/traffic_vault_util.rst
TaylorCFrey: I was curious about this to. I ran into it when testing. When I set the TLS configuration value `InsecureSkipVerify` to `true` I received a connection error:

```[connectionManager] tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config```

I believe it is trying to match the ServerName to information in Certificate. Should I pass the `vault_ip` into the TLS configuration for `ServerName`?

rawlinp: Oh, huh, TIL! I guess this isn't necessary when using the tlsConfig within an http client, but the riak client needs it. Yeah, let's just pass the `vault_ip` into the TLS config for `ServerName`. That would keep it from being redundant.

TaylorCFrey: Done!

rawlinp (approved):

PR #6373, opened 2021-11-22
Logs change

### LINKED ISSUE #6371: Logs have useless fields
## This Improvement request (usability, performance, tech debt, etc.) affects these Traffic Control components:
- Traffic Ops

## Current behavior:
Each changelog entry currently includes the following fields:
```typescript
export interface Log {
	id: number;
	lastUpdated: Date;
	level: "APICHANGE";
	message: string;
	ticketNum: null;
	user: string;
}
```

## New behavior:
IDs have no purpose for changelog entries because they are immutable by nature. "`level`" used to have meaning when changes could be made through either the now-removed Traffic Ops UI or the API, where it would show that distinction. But that no distinction no longer exists (and was always of questionable merit at any rate IMO, so I'm not sure historical preservation is a good reason to keep it around). `ticketNum` has not only no known purpose, but also there's no way to populate it, so it's just always `null`. There's no reasonable way I can think of to make use of that field. Finally, `lastUpdated` is not necessarily "useless", it's just a bad name IMO. A changelog entry is never "updated", it is only created.

With all this in mind, I propose a new format for `/logs` output:
```typescript
export interface Log {
	/**
	 * This should be an RFC3339 timestamp instead
	 * of the custom format used today for "lastUpdated"
	 */
	date: Date;
	message: string;
	user: string;
}
```

This isn't a big enough change to merit a blueprint, IMO, but I could do that too.

ocket8888:
This PR closes #6371 by making the changes therein suggested.

<hr/>

- Documentation
- Traffic Ops Client (Go)
- Traffic Ops

Make sure the API tests pass

## PR submission checklist
- [x] This PR has tests


traffic_ops/app/db/migrations/2021112117412459_logs_change.up.sql
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+CREATE TABLE public.log_new (
+	"message" text NOT NULL,
+	"user" bigint NOT NULL,
+	"date" timestamp with time zone NOT NULL DEFAULT NOW()
```
rawlinp, line 21: `last_updated` is kind of ubiquitous in our API, even though it doesn't necessarily make sense for non-updatable things. Making it unique in this sense means we can't easily reuse queries for things like IMS using the ubiquitous `last_updated` column.

ocket8888: Well I could always leave the column name alone but just change the struct property name. You think that's okay?

rawlinp: That might be confusing

ocket8888: I think it'd be fine, especially if I added `db` struct tags to make the relationship explicit. But if you don't agree I'll just make it `lastUpdated` the whole way through and make note in the documentation that the name was chosen for consistency rather than clarity.

rawlinp: IMO I think the consistency across the API is more useful here than the clarity of this particular endpoint

PR #6388, opened 2021-11-30
Print single quote in heredoc in subshell for Bash 3 compatibility

zrhoffman:
This PR fixes an issue in the `pkg` script when using Bash 3 (as macOS does, since it is the last major version to use GPL 2 and they don't want to touch GPL 3). In Bash 3, a heredoc containing an unmatched single quote is a syntax error:
```
computer:trafficcontrol user$ ./pkg -v
./pkg: line 157: unexpected EOF while looking for matching `)'
./pkg: line 266: syntax error: unexpected end of file
```
which affects `pkg` because of this single quote: https://github.com/apache/trafficcontrol/blob/9516de39fffa3ddda0346720ef98edabc8efa07e/pkg#L166
<hr>

- Build system 

1. Ensure `env bash --version` returns a 3.x.x version
2. Test `pkg` by building a package

- master (9516de39ff)

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/pkg
+++ b/pkg
@@ -163,7 +163,7 @@ Usage: $SELF [options] [projects]
   -d           Disable compiler optimizations for debugging.
   -f <file>    Use <file> as the docker-compose. Default: $COMPOSE_FILE
   -h           Print help message and exit
-  -L           Don't write logs to files - respects output levels on stderr/stdout as set by -q/-v
+  -L           Don$(echo \')t write logs to files - respects output levels on stderr/stdout as set by -q/-v
   -l           List available projects.
   -o           Build from the optional list. Same as -f "$COMPOSE_FILE_OPT"
   -p           Pull builder Docker images, do not build them (default)


pkg
```diff
@@ -163,7 +163,7 @@ Usage: $SELF [options] [projects]
   -d           Disable compiler optimizations for debugging.
   -f <file>    Use <file> as the docker-compose. Default: $COMPOSE_FILE
   -h           Print help message and exit
-  -L           Don't write logs to files - respects output levels on stderr/stdout as set by -q/-v
+  -L           Don$(echo \')t write logs to files - respects output levels on stderr/stdout as set by -q/-v
```
ocket8888, line 166: does this have to be a subshell? Can it not just be escaped?

zrhoffman: Nope, if it's `\'`, then both characters get printed.

Both characters get printed if you do `''`, too.

ocket8888: stunning

ocket8888 (approved):

Issue #6367, opened 2021-11-21
PUT /user/current doesn't work


ocket8888:
## This Bug Report affects these Traffic Control components:
- Traffic Ops

## Current behavior:
A PUT request `/user/current` is incapable of modifying anything about the current user - except that it does update `lastUpdated`.

## Expected behavior:
You should be able to update your user using a PUT request to `/user/current`, as this was possible in previous Traffic Ops versions.

## Steps to reproduce:
1. GET `/user/current` e.g.
```json
{
	"response": {
		"username": "quest",
		"localUser": true,
		"roleName": "admin",
		"addressLine1": null,
		"addressLine2": null,
		"city": null,
		"company": null,
		"country": null,
		"email": "email@addr.com",
		"fullName": "test",
		"gid": null,
		"id": 13,
		"newUser": false,
		"phoneNumber": null,
		"postalCode": null,
		"publicSshKey": null,
		"role": 1,
		"stateOrProvince": null,
		"tenant": "root",
		"tenantId": 1,
		"uid": null,
		"lastUpdated": "2021-11-21 20:22:31+00"
	}
}
```
2. Change anything, or even everything if you want, and submit it in a PUT request to `/user/current` e.g.
```json
{
        "username": "queste",
        "registrationSent": "2021-11-21 20:11:24+00",
        "roleName": "admin",
        "addressLine1": "addressLine1",
        "addressLine2": "addressLine2",
        "city": "city",
        "company": "company",
        "country": "country",
        "email": "address@mail.com",
        "fullName": "testquest",
        "gid": 7,
        "id": 18,
        "newUser": true,
        "phoneNumber": "phoneNumber",
        "postalCode": "postalCode",
        "publicSshKey": "SSH",
        "role": 2,
        "stateOrProvince": "state",
        "tenant": "roote",
        "tenantId": 2,
        "uid": 8,
        "lastUpdated": "2021-11-21 20:11:25+00",
        "token": "token",
        "localUser": false
}
```
3. Notice the response changed nothing e.g.
```json
{
	"alerts": [
		{
			"text": "User profile was successfully updated",
			"level": "success"
		}
	],
	"response": {
		"username": "quest",
		"registrationSent": null,
		"roleName": "admin",
		"addressLine1": null,
		"addressLine2": null,
		"city": null,
		"company": null,
		"country": null,
		"email": "email@addr.com",
		"fullName": "test",
		"gid": null,
		"id": 13,
		"newUser": false,
[...]

PR #6389, opened 2021-12-01
Update PUT user/current to work with v4 User Roles

TaylorCFrey:
Closes: #6367

The `user/current` PUT requests appeared to be failing on v4 API calls. However after testing I was able to successfully change a current user's information, but the response did not display said information. This will fix the response to ensure the correct information is returned.

Additionally, a check was added to ensure the request was wrapped in a legacy `user{}` object.

Lastly, the API integration tests were checking for values to update by issuing an _update_ (PUT) request then performing a _read_ (GET) request. This appears to work as intended for all API versions, it was only the API response for v4.0 that was incorrect.

<hr>

- Traffic Ops

Ensure a valid user is created via the API with appropriate role (say _operations_):
````bash
curl --request POST \
  --url https://<traffic_ops_host>/api/4.0/users \
  --header 'Content-Type: application/json' \
  --data '{
	"addressLine1":"1234 Somewhere ln",
	"city":"Manhattan",
	"company":"Stihl",
	"confirmLocalPasswd":"passwdops",
	"country":"USA",
	"email":"john.frank@local.tld",
	"fullName":"John",
	"localPasswd":"passwdops",
	"phoneNumber":"555-123-4567",
	"postalCode":"10002",
	"role":"operations",
	"stateOrProvince":"New York",
	"username":"operations",
	"tenantId":1
}'
````

Then log in with said user:
````bash
curl --request POST \
  --url https://<traffic_ops_host>/api/4.0/user/login \
  --header 'Content-Type: application/json' \
  --data '{"u":"operations","p":"passwdops"}'
````

Send GET request to `user/current` to ensure correct user is returned

_Request_:
````bash
curl --request GET \
  --url https://<traffic_ops_host>/api/4.0/user/current/
````
_Response_:
````json
{
  "response": {
    "username": "operations",
    "localUser": true,
    "addressLine1": "1234 Somewhere ln",
    "addressLine2": null,
    "city": "Manhattan",
    "company": "Stihl",
    "country": "USA",
    "email": "john.frank@local.tld",
    "fullName": "John",
    "gid": null,
    "id": 89,
    "newUser": false,
    "phoneNumber": "555-123-4567",
    "postalCode": "10002",
    "publicSshKey": null,
    "role": "operations",
    "stateOrProvince": "New York",
    "tenant": "root",
    "tenantId": 1,
    "uid": null,
    "lastUpdated": "2021-12-01T14:13:14.261472-07:00",
    "lastAuthenticated": "2021-12-01T14:13:14.261472-07:00"
  }
}
````

Send PUT request to `user/current` with editable fields modified (say `addressLine1`, `city`, `country`, `postalCode` for example):
````bash
curl --request PUT \
  --url https://<traffic_ops_host>/api/4.0/user/current \
  --header 'Content-Type: application/json' \
  --data '{
	"user": {
		"username": "operations",
		"addressLine1": "5432 Somewhere ln",
		"addressLine2": null,
		"city": "Denver",
		"company": "Stihl",
		"country": "US",
		"email": "john.frank@local.tld",
		"fullName": "John",
		"id": 89,
		"phoneNumber": "555-123-4567",
		"postalCode": "80221",
		"publicSshKey": null,
		"role": "operations",
		"stateOrProvince": "New York",
		"tenant": "root",
		"tenantId": 1
	}
}'
````

Verify fields were updated correctly in the response:
````json
{
  "alerts": [
    {
      "text": "User profile was successfully updated",
      "level": "success"
    }
  ],
  "response": {
    "addressLine1": "5432 Somewhere ln",
    "addressLine2": null,
    "changeLogCount": null,
    "city": "Denver",
    "company": "Stihl",
    "country": "US",
    "email": "john.frank@local.tld",
    "fullName": "John",
    "gid": null,
    "id": 89,
    "lastAuthenticated": null,
    "lastUpdated": "2021-12-01T14:13:14.261472-07:00",
    "newUser": false,
    "phoneNumber": "555-123-4567",
    "postalCode": "80221",
    "publicSshKey": null,
    "registrationSent": null,
    "role": "operations",
    "stateOrProvince": "New York",
    "tenant": "root",
    "tenantId": 1,
    "uid": null,
    "username": "operations"
  }
}
````

Verify updated fields are also returned in GET `user/current` as well as GET `users/:id`

Perform the same above tests for API 3.0, *BUT* be aware that the `role` field must be a valid ID, not a string with API 3.0.

Lastly, attempt to send the body for a PUT request to `user/current` _without_ the `user` containing object. This can be done with either v4.0 or v3.x of the API. An appropriate error message should be returned.

_Request_:
````bash
curl --request PUT \
  --url https://<traffic_ops_host>/api/3.0/user/current \
  --header 'Content-Type: application/json' \
  --data '{
		"username": "operations",
		"addressLine1": "5432 Somewhere ln",
		"addressLine2": null,
		"city": "Denver",
		"company": "Stihl",
		"country": "US",
		"email": "john.frank@local.tld",
		"fullName": "John",
		"id": 89,
		"phoneNumber": "555-123-4567",
		"postalCode": "80212",
		"publicSshKey": null,
		"role": 2,
		"stateOrProvince": "Colorado",
		"tenant": "root",
		"tenantId": 1
}'
````

_Response_:
````json
{
  "alerts": [
    {
      "text": "missing required 'user' object",
      "level": "error"
    }
  ]
}
````

## PR submission checklist
- [X] This PR has tests - API tests exist and do validate functionality, only the v4.0 API response was invalid.
- [ ] This PR has documentation - No changes to documentation were necessary

### DIFF
--- a/traffic_ops/traffic_ops_golang/user/current.go
+++ b/traffic_ops/traffic_ops_golang/user/current.go
@@ -33,8 +33,6 @@ import (
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/tenant"
-
-	"github.com/jmoiron/sqlx"
 )
 
 const replacePasswordQuery = `
@@ -169,7 +167,6 @@ WHERE u.id=$1
 
 func ReplaceCurrent(w http.ResponseWriter, r *http.Request) {
 	var useV4User bool
-	var userV4 tc.UserV4
 	inf, userErr, sysErr, errCode := api.NewInfo(r, nil, nil)
 	tx := inf.Tx.Tx
 	if userErr != nil || sysErr != nil {
@@ -316,16 +313,7 @@ func ReplaceCurrent(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	if useV4User {
-		userV4 = user.Upgrade()
-	}
-
-	if useV4User {
-		err = updateUserV4(userV4, inf.Tx)
-	} else {
-		err = updateUser(&user, tx, changePasswd, changeConfirmPasswd)
-	}
-	if err != nil {
+	if err = updateUser(&user, tx, changePasswd, changeConfirmPasswd); err != nil {
 		errCode = http.StatusInternalServerError
 		sysErr = fmt.Errorf("updating user: %v", err)
 		api.HandleErr(w, r, tx, errCode, nil, sysErr)
@@ -333,26 +321,10 @@ func ReplaceCurrent(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if useV4User {
-		api.WriteRespAlertObj(w, r, tc.SuccessLevel, "User profile was successfully updated", userV4)
+		api.WriteRespAlertObj(w, r, tc.SuccessLevel, "User profile was successfully updated", user.Upgrade())
 	} else {
 		api.WriteRespAlertObj(w, r, tc.SuccessLevel, "User profile was successfully updated", user)
 	}
-}
-
-func updateUserV4(u tc.UserV4, tx *sqlx.Tx) error {
-	resultRows, err := tx.NamedQuery(UpdateQueryV40(), u)
-	if err != nil {
-		return err
-	}
-	defer resultRows.Close()
-
-	for resultRows.Next() {
-		if err := resultRows.Scan(&u.LastUpdated, &u.Tenant, &u.Role); err != nil {
-			return err
-		}
-	}
-
-	return nil
 }
 
 func updateUser(u *tc.User, tx *sql.Tx, changePassword bool, changeConfirmPasswd bool) error {
--- a/traffic_ops/traffic_ops_golang/user/current.go
+++ b/traffic_ops/traffic_ops_golang/user/current.go
@@ -28,13 +28,13 @@ import (
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/lib/go-util"
-	"github.com/jmoiron/sqlx"
 
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/tenant"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by mitchell852** [comment]
so this bug snuck thru due to zero tests for this endpoint, right? think you can create a simple positive test to prevent this from happening again @TaylorCFrey ?

**Comment by TaylorCFrey** [comment]
Good question @mitchell852. There are no unit tests, but there does appear to be integration API test. I'll look into those asap.

File: traffic_ops/traffic_ops_golang/user/current.go [hunk 1]
```diff
@@ -309,7 +316,16 @@ func ReplaceCurrent(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	if err = updateUser(&user, tx, changePasswd, changeConfirmPasswd); err != nil {
+	if useV4User {
+		userV4 = user.Upgrade()
+	}
+
+	if useV4User {
+		err = updateUserV4(userV4, inf.Tx)
```
rawlinp, line 324: So, as far as I can tell, the difference between userV4 and user is really just `Role`, which isn't even used/updated in the `updateUser` function. So, why doesn't just calling `updateUser(&user, tx, changePasswd, changeConfirmPasswd)` actually work?

TaylorCFrey: Good catch, and actually I didn't want to use `UpdateQueryV40` anyways. Hold pls for an update.

mitchell852:
> Good question @mitchell852. There are no unit tests, but there does appear to be integration API test. I'll look into those asap.

yeah, i would think a simple api integration test to verify that calling PUT /user/current with a valid user object succeeds should be sufficient.

rawlinp:
I think the real mystery is how [the existing TO API](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/testing/api/v4/user_test.go#L247-L275) test wasn't failing, because it definitely attempts to change fields then checks to see that the changes stuck.

TaylorCFrey:
After some exploration I believe I found two issues. First, the request body from #6367 needs to be wrapped in a `"user":{...}` object. Without the wrapper, the API would not parse the body resulting in the API call believing nothing has changed. The response would return a success thinking all fields were "unedited". I've modified the code to check for the existence of the `user` wrapper and submitted an issue to explore it's removal #6396

Second, if the request was for an API v4.0, it would appear as though some of the fields were not edited, but a success would be returned. Subsequent `GET` requests would appear to work and show that the field was updated. This was because we were saving the state of the user earlier in the function then returning that state, rather than the update state. This was not the case for v3.x API calls. And it only impacted the response body from the `PUT` requests. This explains why the API integration client tests were succeeding as they do not directly validate the response body from the `PUT` request, but rather initiate a `PUT` followed by a `GET`.

rawlinp (approved):
Manually verified that not providing the `user` wrapper object returns an error as expected and that the PUT response is correct for 4.0 now.

Issue #6299, opened 2021-10-19
User representations don't match


ocket8888:
## This Improvement request (usability, performance, tech debt, etc.) affects these Traffic Control components:
- Traffic Ops
- Documentation

## Current behavior:
The representations of Users are different depending on the API endpoint used to view them, e.g. `/user/current` uses the field `roleName` to hold the name of the User's Role, while `/users` shows the same information in the `rolename` property.

The current model for users (API v3.0, 4.0 makes some structural changes in some contexts) is such that requests and responses have different "shapes" depending on the request method - and some fields are even renamed between requests and responses.

<details><summary>A TypeScript model of Users in APIv3</summary>

```typescript
export interface PostRequestUser {
	addressLine1?: string | null;
	addressLine2?: string | null;
	city?: string | null;
	company?: string | null;
	confirmLocalPasswd: string;
	country?: string | null;
	email: string;
	fullName: string;
	gid?: number | null;
	id?: never;
	lastUpdated?: never;
	localPasswd: string;
	newUser?: boolean | null;
	phoneNumber?: string | null;
	postalCode?: string | null;
	publicSshKey?: string | null;
	registrationSent?: never;
	role: number;
	roleName?: never;
	rolename?: never;
	stateOrProvince?: string | null;
	tenant?: never;
	tenantId?: never;
	tenantID: number;
	uid?: number | null;
	username: string;
}

interface PutRequestNotChangingPasswordUser {
	addressLine1?: string | null;
	addressLine2?: string | null;
	city?: string | null;
	company?: string | null;
	confirmLocalPasswd?: never;
	country?: string | null;
	email: string;
	fullName: string;
	gid?: number | null;
	id?: never;
	lastUpdated?: never;
	localPasswd?: never;
	newUser?: boolean | null;
	phoneNumber?: string | null;
	postalCode?: string | null;
	publicSshKey?: string | null;
	registrationSent?: never;
	role: number;
	roleName?: never;
	rolename?: never;
	stateOrProvince?: string | null;
	tenant?: never;
	tenantId?: never;
	tenantID: number;
	uid?: number | null;
	username: string;
}

export type PutRequestUser = PostRequestUser | PutRequestNotChangingPasswordUser;

interface ResponseUser {
	addressLine1: string | null;
	addressLine2: string | null;
	city: string | null;
	company: string | null;
	country: string | null;
	email: string;
	fullName: string;
	gid: number | null;
	id: number;
	lastUpdated: Date;
[...]

PR #6263, opened 2021-10-08
TPv2 Structure/Module changes

shamrickus:
This PR makes some organizational changes to how TPv2 is structured. First, files are no longer grouped by Angular type, but by "feature". Second, the code has been refactored into 3 modules (`App`, `Shared`, `Core`). The `Core` module is lazy loaded only after a user has been authenticated.

<hr>

## Which Traffic Control components are affected by this PR?

- Traffic Portal 2

## What is the best way to verify this PR?
Run the tests, make sure site works. You can also look at the network tab and see that `core` is only loaded when authenticated and a relevant route has been requested.

## PR submission checklist
- [x] This PR has tests 
- [x] This PR has documentation 
- [x] This PR has a CHANGELOG.md entry 
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://apache.org/security) for details)

PR #6391, opened 2021-12-02
TPv2 Update Angular and make compilation stricter

ocket8888:
This PR updates Angular to version 13 (13.0.3 at the time of this writing) for TPv2, and adds some options to get type-checking in templates, as well as adding some more strict type checking generically. Mostly changes that needed to made to make it compile came from the strict template typing. And the fact that it was using a "user" as the same thing as a "current user", which unfortunately cannot be safely done (see #6299).

This also fixes some broken tests from the module restructure done in #6263, and squashes an error about Chart JS being bulky and causing optimization problems. That's all true, but we know about it, so there's no reason to see a warning on every compilation.

Also, the new version of Angular allows the use of ES2020 in source, and because IE is dead and Edge is now Chromium-based, our output has moved up to ES2017, which is nice.

<hr/>

- Traffic Portal (experimental v2)

Make sure the unit and end-to-end tests pass. Also probably play around with the UI a bit, because our end-to-end tests only cover the login page.

- master, but who cares, it's experimental

## PR submission checklist
- [x] This PR has tests


ocket8888:
I know this PR looks big, but 49.k lines of changes are coming from just package-lock.json.

experimental/traffic-portal/package.json
```diff
@@ -22,60 +22,59 @@
...
+    "@angular-eslint/template-parser": "13.0.1",
+    "@angular/cli": "~13.0.3",
+    "@angular/compiler-cli": "~13.0.2",
     "@types/argparse": "^2.0.2",
-    "@types/chart.js": "^2.9.32",
+    "@types/chart.js": "^2.9.34",
     "@types/express": "^4.17.0",
     "@types/jasmine": "~3.6.0",
     "@types/jasminewd2": "~2.0.3",
     "@types/node": "^12.11.1",
```
shamrickus, line 69: Should be upgrade to 14?

ocket8888: yeah, it seems like it should.

shamrickus (approved):
LGTM

PR #6395, opened 2021-12-02
Set Key for Lets Encrypt Autorenewal

Vijay-1:
@ocket8888 @mattjackson220 

<hr>

- Traffic Ops

topost /letsencrypt/autorenew -k -a 2.0 --to-url https://traffic-ops.net --to-user ****--to-password ****--request-payload --request-headers

Will result in renewal of certificates

4.x, 5.x,6,x


mattjackson220 (approved):
looks good!

PR #6404, opened 2021-12-03
Upgrade Profiles Traffic Portal view to use AG-Grid tables

ocket8888:
This updates the "Profiles" Traffic Portal page to use AG-Grid instead of JQuery datatables. It's mostly a drop-in replacement, except that the "CDN" column now displays the CDN ID on hover, Types now display a more readable explanation of their meaning on hover (e.g. "ES_PROFILE" now shows "Elasticsearch Server Profile" in the on-hover tooltip) and columns were added for ID and Last Updated - the data for those two were always available but just not shown previously. They're hidden by default to preserve the old table's structure a bit more.

This PR _also_ adds some general purpose functionality that should help greatly reduce duplicated components for special filtering of tables. As an example, this PR removes the "CDN Profiles" view (previously accessible from the "Manage Profiles" context menu item of a row in the CDN table, or the "View Profiles" item in the "More" dropdown on a CDN's details page). Instead, the places whence a user has been directed to that view now instead direct the user to the selfsame Profiles table herein upgraded to AG-Grid, with a filter set for that CDN. This was accomplished by making _all_ AG-Grid tables support filtering through query string parameters. For example, after this PR, the path/fragment `#!/profiles?cdnName=CDN-in-a-Box` will **always** take you to a Profile table filtered by that CDN. Any other filters the user has applied to the table, e.g. Type, will be saved and re-applied as normal. This new filter state, though, is _not_ saved immediately. The user's saved filters are only overwritten when they alter some filter. So, going to `#!/profiles?cdnName=CDN-in-a-Box` will show a list of Profiles filtered by CDN, but then immediately going to `#!/profiles` will load the user's normal filters and not filter by CDN unless they normally do so (and of course it'll filter by whatever CDN they normally use). This new filtering works with any property of any object shown in a table that has an API property name. For example, one can link a person to Profiles filtered to only show `ATS_PROFILE` using `#!/profiles?type=ATS_PROFILE`. Furthermore, these can even be combined; `#!/profiles?cdnName=CDN-in-a-Box&type=ATS_PROFILE&routingDisabled=true` is a link to a table that filters Profiles to (at least) those within the CDN-in-a-Box CDN that are ATS cache server Profiles and are routing-disabled. Finally, all numeric and textual (and boolean because those are treated as textual) fields support a _second_ value that will be combined using OR. For example, `#!/profiles?type=ATS_PROFILE&type=GROVE_PROFILE` sets the table to filter for Profiles that can be used for cache servers, whether they're ATS or Grove.

<hr/>

- Traffic Portal

Make sure all the TP tests pass, make sure the new table looks okay. Check out both aforementioned ways of viewing a CDN's Profiles and make sure they both still work. You can also try to test out some query string parameters, if you like. I ran it through the examples above and it seems to be working fine, but the only situation that _has_ to work right now is the CDN one because that's the only one that's accessible without directly altering the URL.

## PR submission checklist
- [x] This PR has tests


shamrickus:
Rebase please (has CiaB fix)

shamrickus (approved):
LGTM

PR #6405, opened 2021-12-05
Fix t3c version in build

rob05c:
Fixes the t3c build to properly add the ATC version, changeset,
and monotonic version to the t3c version commands,
as well as to t3c-generate generated files.

Also adds the version flag to apps that were missing it.

I see this as a pretty big win for operations. Every time I install a new t3c version, the first thing I do is run it, and look at how the config files changed. Because production runs are on a schedule, it's difficult or impossible to figure out which exact git commit was the first new version run. This makes it easy to see in the file headers.

Includes changelog
Includes docs
Includes tests

Labelling `improvement` rather than `bug`, because even though some code existed the feature has never worked.

Install `trafficcontrol-cache-config`, run `t3c-* --version`, verify version matches ATC version and includes git changeset.
Run t3c, verify config file comment headers include ATC/RPM version and changeset.

Not a bug.

## PR submission checklist
- [x] This PR has tests 


ezelkow1 (approved):
lgtm

PR #6414, opened 2021-12-09
Remove t3c unused plugin symbols

rob05c:
t3c wasn't sending the list of installed plugins to t3c-check-reload
like it's designed to take, but that was actually unnecessary,
because installed plugins always require a restart, but syncds
never installs plugins, and badass always restarts. So the entire
check served no purpose.

No new tests, existing tests cover the behavior, and this doesn't change behavior, just removes unused code.
No docs, no interface change.

Run t3c, verify behavior is unchanged.

Not a bug fix

## PR submission checklist


jpappa200 (approved):
LGTM 👍  
Built and ran as expected.

zrhoffman:
#6414 has a merge conflict

Issue #2488, opened 2018-06-29
Create a permission that can be used to retrieve secure parameter values


mitchell852:
GET /api/parameters can contain secure values.

currently if you are an admin (priv level=30), then the values for "secure" parameters will be returned. if you are not an admin, the value will be replaced with '**********'

as we move towards capability-based permissions (as opposed to priv_level based), we will need to key off of a capability (instead of a role) to determine whether to show or hide the value of secure parameters.

One suggestion was to create a capability called "secure-read" or something along those lines. The presence of this capability would allow you to see the value of "secure" parameters.

So for example,

to access GET /api/parameters and see value of secure and non-secure parameters you'd need the following capabilities:

params-read
secure-read

to access GET /api/parameters and see value of ONLY non-secure parameters you'd need the following capabilities:

params-read

mitchell852:
related to #2037

mitchell852:
assigning this to @rob05c to ensure he has this accounted for in his open roles/capabilities PR

rob05c:
Yes, it is accounted for. E.g. https://github.com/apache/trafficcontrol/pull/2791/files#diff-aaa675dae4ffc0279136f6687d6c7f69R398

mitchell852:
can this be closed @srijeet0406 @ocket8888 @shamrickus?

ocket8888:
No. That's actually kind of an important detail we missed. With Permissions checking disabled, anyone with priv_level >= 30 can see "secure" Parameter Values. If you turn on Permissions checking, that doesn't change and there's no Permission you can take away from someone to hide them or give someone to show those Values. That should be fixed.

srijeet0406:
@mitchell852 I can take this one

PR #6417, opened 2021-12-10
Create a secure parameter read permission to be able to see the values of secure parameters

srijeet0406:
Closes: #2488 

<hr>

- Traffic Ops

Make a call to `GET /api/parameters` as a user that has the `PARAMETER-SECURE:READ` permission associated with it's role. This should get the value of all secure and insecure parameters.
Now, make the same call from a user that does not have the said permission associated with it's role. This should display the values of the secure parameters as `*****`.
All tests should pass. 

- master

## PR submission checklist
- [x] This PR has tests 
- [x] This PR does not require documentation 
- [x] This PR does not have a CHANGELOG.md entry 

### DIFF
--- a/traffic_ops/traffic_ops_golang/parameter/parameters.go
+++ b/traffic_ops/traffic_ops_golang/parameter/parameters.go
@@ -21,6 +21,7 @@ package parameter
 
 import (
 	"errors"
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
 	"net/http"
 	"strconv"
 	"time"
@@ -31,7 +32,6 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-tc/tovalidate"
 	"github.com/apache/trafficcontrol/lib/go-util"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
-	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/util/ims"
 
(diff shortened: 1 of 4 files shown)


traffic_ops/traffic_ops_golang/parameter/parameters.go
```diff
@@ -175,8 +175,14 @@ func (param *TOParameter) Read(h http.Header, useIMS bool) ([]interface{}, error
 		if err = rows.StructScan(&p); err != nil {
 			return nil, nil, errors.New("scanning " + param.GetType() + ": " + err.Error()), http.StatusInternalServerError, nil
 		}
-		if p.Secure != nil && *p.Secure && param.ReqInfo.User.PrivLevel < auth.PrivLevelAdmin {
-			p.Value = &HiddenField
+		if p.Secure != nil && *p.Secure {
+			if param.ReqInfo.Config.RoleBasedPermissions {
```
ocket8888, line 179: `RoleBasedPermissions` is only supposed to have an effect API versions 4 and later, and this handler is shared across all versions.

ocket8888 (changes requested):

shamrickus (approved):
LGTM
I'll let @ocket8888 merge in case he wants to re-review

traffic_ops/traffic_ops_golang/parameter/parameters.go
```diff
@@ -31,7 +32,6 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-tc/tovalidate"
 	"github.com/apache/trafficcontrol/lib/go-util"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
-	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
```
ocket8888, line 34: This import shouldn't have been moved up into the standard library

PR #6423, opened 2021-12-14
Fix TM stats_over_http parsing to sum DS data

rawlinp:
The stats_over_http parsing was replacing stat values instead of summing
them for delivery services with multiple regexes. Since stats_over_http
response output can change order from request to request, this was
causing TM to report "new stat is lower than last stat" warnings when
the response order changed (which happens quite frequently). By properly
summing together the stats that match the same underlying delivery
service, these warnings disappear, and the reported data from TM is more
accurate.

Additionally, some things like unused parameters were cleaned up while
trying to find the underlying cause of these warnings.

<hr>

- Traffic Monitor

In an environment that contains delivery services with multiple regexes in use and servers that use the stats_over_http plugin (instead of astats), run TM prior to this version. You will likely see many "new stat is less than last stat" warnings. Then run this version of TM, and you shouldn't see any of those warnings anymore.

- 5.x
- 6.x

## PR submission checklist
- [x] This PR updates some tests but doesn't really add a new test for this particular case
- [x] Bug fix, no docs necessary


mattjackson220 (approved):
Looks good to me! tests pass, tested locally and i dont see the "new stat is less than last stat" log lines

PR #6425, opened 2021-12-14
adds note to blueprint regarding lazy consensus option

mitchell852:
This PR adds a note to the blueprint section of CONTRIBUTING.md regarding the option to invoke [lazy consensus](https://community.apache.org/committers/lazyConsensus.html) on blueprints if needed.

- Documentation

Read Contributing.md file, blueprint section

## PR submission checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,7 +68,7 @@ list with a short description of your feature plus a link to the blueprint PR.
 1. Wait for feedback to be applied to your blueprint PR. Because it's a PR,
 line-specific feedback can be given.
 1. Just like any PR, once all the concerns have been addressed, the blueprint is
-merged into the blueprints directory (if accepted) or closed (if rejected).
+merged into the blueprints directory (if accepted) or closed (if rejected). Note: Blueprint authors have the option of invoking <a href="https://community.apache.org/committers/lazyConsensus.html" target="_blank">lazy consensus</a> to facilitate the merge of the blueprint if community feedback is not being provided or feedback has stalled. To invoke lazy consensus, please make the community aware via an email to the dev mailing list that you intend to invoke lazy consensus after X hours with no feedback.
 1. Start work on the feature. Optionally, you can open a draft PR if you want
 feedback during development.
 1. Submit your PR for review and reference the blueprint in the PR description.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,7 +68,7 @@ list with a short description of your feature plus a link to the blueprint PR.
 1. Wait for feedback to be applied to your blueprint PR. Because it's a PR,
 line-specific feedback can be given.
 1. Just like any PR, once all the concerns have been addressed, the blueprint is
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -68,7 +68,7 @@ list with a short description of your feature plus a link to the blueprint PR.
 1. Wait for feedback to be applied to your blueprint PR. Because it's a PR,
 line-specific feedback can be given.
 1. Just like any PR, once all the concerns have been addressed, the blueprint is
-merged into the blueprints directory (if accepted) or closed (if rejected).
+merged into the blueprints directory (if accepted) or closed (if rejected). Note: Blueprint authors have the option of invoking <a href="https://community.apache.org/committers/lazyConsensus.html" target="_blank">lazy consensus</a> to facilitate the merge of the blueprint if community feedback is not being provided or feedback has stalled. To invoke lazy consensus, please make the community aware via an email to the dev mailing list that you intend to invoke lazy consensus after X hours with no feedback.
```
ocket8888, line 71: Why are you using an HTML anchor tag instead of just a markdown link?

ocket8888 (approved):
Seems fine to me

CONTRIBUTING.md
```diff
@@ -68,7 +68,7 @@ list with a short description of your feature plus a link to the blueprint PR.
 1. Wait for feedback to be applied to your blueprint PR. Because it's a PR,
 line-specific feedback can be given.
 1. Just like any PR, once all the concerns have been addressed, the blueprint is
-merged into the blueprints directory (if accepted) or closed (if rejected).
+merged into the blueprints directory (if accepted) or closed (if rejected). Note: Blueprint authors have the option of invoking [lazy consensus](https://community.apache.org/committers/lazyConsensus.html) to facilitate the merge of the blueprint if community feedback is not being provided or feedback has stalled. To invoke lazy consensus, please make the community aware via an email to the [dev@trafficcontrol.apache.org](mailto:dev@trafficcontrol.apache.org) mailing list that you intend to invoke lazy consensus after X hours with no feedback.
```
zrhoffman, line 71: Maybe mention that lazy consensus cannot be invoked once any feedback is left.

zrhoffman (changes requested):

CONTRIBUTING.md
mitchell852: isn't that sort of implied with

> or feedback has stalled

> you intend to invoke lazy consensus after X hours with no feedback

zrhoffman: "feedback has stalled" implies that some feedback has already been left. Is there a precedent for lazy consensus being used, in this case?

rawlinp: I would leave that up to the discretion of the blueprint submitter IMO. Let's say someone reviews your blueprint then goes on vacation for two weeks. Depending on their feedback, you might think they'd be ok with the blueprint merging once their feedback is addressed, rather than forcing you to wait 2 weeks for them to get back. Plus, it's not like once a blueprint is reviewed and merged that the design cannot change after that. Merging a blueprint just generally means you should be able to move forward with the proposed design, although that's not a hard rule either. In the above scenario, if the person who left feedback still didn't agree with the blueprint after it was merged, the blueprint could always be changed (assuming general consensus on the new change).

zrhoffman: Sounds good to me

zrhoffman (approved):

PR #6433, opened 2021-12-16
Deprecate/remove `/capabilities`

### LINKED ISSUE #6319: Deprecate/remove /capabilities TO API endpoint
## This Improvement request (usability, performance, tech debt, etc.) affects these Traffic Control components:
- Traffic Ops Client (Go)
- Traffic Ops
- Traffic Portal
- Documentation

## Current behavior:
With Roles and Permissions being implemented, "Capabilities" have no plans for future use, so they should be deprecated in API versions earlier than 4 and removed from APIv4 and later. Since they are not enforced anywhere, removing the endpoint from 4.0 won't break any permissions checking, even though that version still supports the old priv_level method of permissions checking.

If done before Permissions are fully integrated into TP, the TP services can just be downgraded to use the stable API in lieu of stripping the functionality out entirely; we're going to need to rework that for Permissions soon anyway, so there's no need to make bigger changes than are necessary.

## New behavior:
`/capabilities` is deprecated in API v<4 and removed in v>=4

ocket8888:
This PR closes #6319 by removing `/capabilities` from Traffic Ops API version 4 and adding deprecation notices to earlier versions.

<hr/>

- Documentation
- Traffic Ops Client (Go)
- Traffic Ops

Make sure nothing else is broken, endpoints still work in API versions earlier than 4, and they output deprecation warnings in those earlier versions.

## PR submission checklist
- [ ] This PR has tests


rawlinp (approved):
LGTM

PR #6380, opened 2021-11-25
Upgrade github.com/lestrrat-go/jwx to 1.2.12

lestrrat:
Hi,

I am the author of github.com/lestrrat-go/jwx. During a routine search I noticed that this project uses a rather antiquated version of that library.

After reviewing I realized that you probably have not upgraded the module because `jwt.Set` has been turned into an interface during the v0.9.x to v1.x transition. You probably needed to allow private parameters in the `jwk.Set` deserialization. I have tentatively implemented it in the main branch for github.com/lestrrat-go/jwx, and applied the changes I _think_ are required to make things work on your end. Currently the PR uses an unreleased version of github.com/lestrrat-go/jwx. If these changes actually work for you, I should probably release 1.2.12 first, and then have this PR merged.

Please note that I have only verified that `make unit` mostly works -- I did not run the `TestRemapConfig` and its related tests, and did not verify `trafficcontrol` running in any other form.

Also, I reviewed your PR template, but didn't think this particular PR really fits into your regular type of PR. I especially have no clue how `trafficcontrol` works, and just changed the minimal code so that it works with the new version of jwx. And to that extent, I made the judgement call to set aside your templates, as I do not believe I can really respond to your PR template.  I feel bad for ignoring the rules, but I hope you understand my position.

PR #6434, opened 2021-12-17
Replace github.com/dgrijalva/jwt-go with github.com/lestrrat-go/jwx/jwt

lestrrat:
Hello, this is a PR similar to #6380, and it aims to replace github.com/dgrijalva/jwt-go with github.com/lestrrat-go/jwx/jwt. I hope breaking protocol with the PR template is okay for this particular case.

For this PR, I have added various new features in github.com/lestrrat-go/jwx around handling `jku`s for traffic_ops/traffic_ops_golang/login/login.go, and hence the PR uses an as-of-yet unreleased version of github.com/lestrrat-go/jwx again. As for the new features, most notable to this project is the ability to verify a token using `jku` in a relatively safe manner. It by default checks for URLs having `https` scheme, it uses whitelists, and it now also uses `jwk.AutoRefresh` which keeps track of JWKS and only makes HTTP requests when required.

For relevant PRs around these features on github.com/lestrrat-go/jwx side, please see lestrrat-go/jwx#511, lestrrat-go/jwx#513, lestrrat-go/jwx#515, lestrrat-go/jwx#517, and lestrrat-go/jwx#519. 

The error handling logic in `login.go` is ever so slightly different from before, but as far as I can see, there were no test failures. The change is mainly in that github.com/lestrrat-go/jwx/jwt has no notion of a parsable, and yet invalid token (`token.Valid` is not a thing). `jwt.Parse` will simply return an error whether it's a malformed JWT(JWS) or if verification fails. So there were times when error handling had to be tweaked.

For other occurrences, there were a couple of files in `experimental` directory that used `dgrijalva/jwt-go`. However, these seemed half baked, and without tests, so I only made sure that they would build.

Finally. I tried my best, but I just cannot remove github.com/dgrijalva/jwt-go from `go.sum`. I don't know what's referencing it. Pointers welcome


zrhoffman (changes requested):
Since `github.com/dgrijalva/jwt-go` is no longer a vendored dependency, these lines should be removed from `LICENSE`:
https://github.com/apache/trafficcontrol/blob/33a1ee0f30ccb0b927105e2a96187883a0a3b6d8/LICENSE#L565-L569

go.sum
```diff
@@ -386,8 +386,6 @@ github.com/denisenkom/go-mssqldb v0.10.0/go.mod h1:xbL0rPBG9cCiLr28tMa8zpbdarY27
 github.com/denverdino/aliyungo v0.0.0-20190125010748-a747050bb1ba/go.mod h1:dV8lFg6daOBZbT6/BDGIz6Y3WFGn8juu6G+CQ6LHtl0=
 github.com/dgrijalva/jwt-go v0.0.0-20170104182250-a601269ab70c/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
 github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
```
zrhoffman, line 388: > Finally. I tried my best, but I just cannot remove github.com/dgrijalva/jwt-go from `go.sum`. I don't know what's referencing it. Pointers welcome

It is required by 2 modules:
* `github.com/influxdata/influxdb`@`v1.9.5`, a direct dependency
* `github.com/Azure/go-autorest/autorest/adal`, required by `github.com/golang-migrate/migrate`, required at https://github.com/apache/trafficcontrol/blob/33a1ee0f30ccb0b927105e2a96187883a0a3b6d8/traffic_ops/app/db/admin.go#L36

Since `github.com/dgrijalva/jwt-go` is no longer a vendored dependency, I don't see a problem with these lines staying in `go.sum`, from ATC's perspective.

lestrrat:
@zrhoffman PTAL

lestrrat:
@zrhoffman I figure it's Christmas and the New Year's and what not, but just a mild nudge :)

zrhoffman:
Thank you for your patience over the holidays. The code seems good, but it still needs to be tested.

mitchell852:
@lestrrat - can you fix the merge conflict?

lestrrat:
I resolved it, but I'm not sure that the error in CI was

zrhoffman (approved):
Looks good! Token login works, and the changes I requested have been addressed.

zrhoffman:
@lestrrat The changes look good. Would you please update `github.com/lestrrat-go/jwx/jwt` to the latest release and revendor dependencies (which should fix the merge conflict)?

zrhoffman:
Never mind, I was able to fix the merge conflicts myself, and the version is identical to `v1.2.14` besides the `Changes` file. Thank you for your contribution!

lestrrat:
@zrhoffman thanks!

Issue #6382, opened 2021-11-25
Geo Limit Countries should be an array


ocket8888:
## This Improvement request (usability, performance, tech debt, etc.) affects these Traffic Control components:
- Traffic Ops

## Current behavior:
Currently, it's a string that represents an array of strings separated by commas e.g. `"US,CA"`.

## New behavior:
Arrays should be arrays e.g. `["US", "CA"]`. If the parser is smart enough (and if we care to do it this way), that could be done in a backward-compatible manner such that if the field is a string it's parsed as a comma-delimited array, and if it's an array it's parsed as such. With the right validation (currently it has none, but country codes can't, for example, legally contain commas) the storage format could even be kept the same, which not only obviates the need for a migration but trivializes supporting old API versions.

srijeet0406:
@ocket8888 I can take this one

PR #6441, opened 2021-12-21
Geo limit countries can now be parsed as an array and a string

srijeet0406:
This PR closes https://github.com/apache/trafficcontrol/issues/6382 Geo Limit Countries should be an array

<hr>

- Traffic Ops
- Traffic Portal

Make sure all the tests pass.
Using the API, create a DS (or update an existing one) with `GeoLimit` set to 0.
Try setting the `Geo Limit Countries` with a string -> this should pass
Try setting the `Geo Limit Countries` with a string array (`["US","CA"]`)-> this should pass
Try setting the `Geo Limit Countries` with a malformed string array (`"US","CA"]`)-> this should fail
Try setting the `Geo Limit Countries` with a string array with alphabets and numbers (`["US","CA","12"]`)-> this should fail

- master

## PR submission checklist
- [x] This PR has tests 
- [x] This PR does not have documentation 


traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
```diff
@@ -941,7 +941,7 @@ <h3 ng-if="!open()">Previous Value</h3>
                         </div>
                         </label>
                         <div class="col-md-10 col-sm-10 col-xs-12">
-                            <input id="geoLimitCountries" name="geoLimitCountries" type="text" class="form-control" ng-model="deliveryService.geoLimitCountries" maxlength="255" pattern="[A-Z]+(,[A-Z]+)*">
+                            <input id="geoLimitCountries" name="geoLimitCountries" type="text" class="form-control" ng-model="deliveryService.geoLimitCountries" maxlength="255" pattern="(\[[A-Z]+(,[A-Z]+)*\]|[A-Z]+(,[A-Z]+)*)">
```
ocket8888, line 944: Users should not be writing loose JSON syntax into our inputs, they should be able to handle that for the user. This document fragment actually doesn't need to change at all to make this change to Delivery Services, though IMO it would be an improvement to make it a true variable list of inputs like the "Consistent Hash Query Parameters".

ocket8888 (changes requested):
It seems like instead of allowing arrays as input, this is allowing a string that may optionally start with `[` and end with `]`. This changeset seems to allow e.g.
```json
{
  "geoLimitCountries": "[US,FR]"
}
```
but it should be allowing
```json
{
  "geoLimitCountries": [ "US", "FR" ]
}
```

traffic_portal/app/src/common/modules/form/deliveryService/form.deliveryService.DNS.tpl.html
srijeet0406: Alright, so I'm going to back out the TP changes, but on the API side, I'll add the logic for the `geoLimitCountries` to be able to be either a string or an array of strings. Does that sound good?

ocket8888: yeah, that should be fine.

lib/go-tc/deliveryservices.go
```diff
@@ -643,7 +643,7 @@ type DeliveryServiceNullableFieldsV11 struct {
 	// countries within which the Delivery Service's content ought to be made
 	// available. This has no effect if GeoLimit is not a pointer to exactly the
 	// value 2.
-	GeoLimitCountries *string `json:"geoLimitCountries" db:"geo_limit_countries"`
+	GeoLimitCountries *interface{} `json:"geoLimitCountries" db:"geo_limit_countries"`
```
ocket8888, line 646: When `geoLimitCountries` is passed into the API as a string, it's still really an array, just one that's represented as a string. Because of that, I think if we're going to make a type change here it makes the most sense to use a string slice. To get it to parse independent of input type, there are a few ways but what I'd recommend is a type alias for `[]string` that implements [`encoding/json.Unmarshaler`](https://pkg.go.dev/encoding/json#Unmarshaler) to handle the two situations and parse country code array strings out into actual arrays immediately, which is currently being enforced as validation (and so in that case could/would need to be removed from the validation). 

Also, the types of fields in legacy API version structures cannot be changed. While the change is non-breaking for the API, it is breaking the call signature of legacy clients using our Go client library. If anyone is using the `GeoLimitCountries` field of that struct - or any that nest it - then their programs will no longer compile after this change: https://go.dev/play/p/Dk7Eu8XbKt4

lib/go-tc/deliveryservices.go
```diff
@@ -245,7 +245,8 @@ type DeliveryServiceV40 struct {
 
 	// TLSVersions is the list of explicitly supported TLS versions for cache
 	// servers serving the Delivery Service's content.
-	TLSVersions []string `json:"tlsVersions" db:"tls_versions"`
+	TLSVersions           []string               `json:"tlsVersions" db:"tls_versions"`
+	GeoLimitCountriesList *GeoLimitCountriesType `json:"geoLimitCountriesList"`
```
ocket8888, line 249: I don't think we should add a new property for this, the existing `geoLimitCountries` JSON property can be re-used; when `encoding/json` unmarshals into a structure, it uses the same rules as the Go language to determine which property to unmarshal into.

Example: https://go.dev/play/p/chpKFwvn9X0

Because we rely heavily on nesting structures to avoid duplicating unchanged fields, we have a few duplicated data fields in the API structures, and accessing properties of nested structures instead of just directly (i.e. `fooV40.fooV31.Prop` instead of `fooV40.Prop` even if `Prop` only exists on `fooV40.fooV31`) is not supported by `lib/go-tc`.

lib/go-tc/deliveryservices.go
```diff
@@ -245,7 +245,8 @@ type DeliveryServiceV40 struct {
 
 	// TLSVersions is the list of explicitly supported TLS versions for cache
 	// servers serving the Delivery Service's content.
-	TLSVersions []string `json:"tlsVersions" db:"tls_versions"`
+	TLSVersions       []string               `json:"tlsVersions" db:"tls_versions"`
+	GeoLimitCountries *GeoLimitCountriesType `json:"geoLimitCountries"`
```
ocket8888, line 249: `GeoLimitCountriesType` is already a reference type - `nil` is a valid value for a `[]string` - so I don't think this needs to be a pointer.

lib/go-tc/deliveryservices.go
```diff
@@ -577,6 +578,36 @@ type DeliveryServiceNullableV11 struct {
 	DeliveryServiceRemovedFieldsV11
 }
 
+type GeoLimitCountriesType []string
```
ocket8888, line 581: This type should probably have a GoDoc comment. Personally I would also add one to its `UnmarshalJSON` and `MarshalJSON` methods, but those are extremely common especially in our codebase so it probably doesn't add much value.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added definition for `heartbeat.polling.interval` for CDN Traffic Monitor config in API documentation.
 - New `pkg` script options, `-h`, `-s`, `-S`, and `-L`.
 - Added `Invalidation Type` (REFRESH or REFETCH) for invalidating content to Traffic Portal.
+- Added a new field `geoLimitCountriesList` to the `DeliveryService` structure to accept both an array as well as single strings for geo limit country codes
```
ocket8888, line 20: This is no longer true - also currently

> an array as well as single strings

isn't true currently, see above comment

docs/source/api/v4/deliveryservices.rst
```diff
@@ -105,6 +105,7 @@ Response Structure
 :fqPacingRate:              The :ref:`ds-fqpr`
 :geoLimit:                  An integer that defines the :ref:`ds-geo-limit`
 :geoLimitCountries:         A string containing a comma-separated list defining the :ref:`ds-geo-limit-countries`
+:geoLimitCountriesList:     A string or string array containing a comma-separated list defining the :ref:`ds-geo-limit-countries-list`
```
ocket8888, line 108: This isn't accurate anymore

docs/source/api/v4/deliveryservices.rst
```diff
@@ -297,6 +298,7 @@ Request Structure
 :fqPacingRate:              The :ref:`ds-fqpr`
 :geoLimit:                  An integer that defines the :ref:`ds-geo-limit`
 :geoLimitCountries:         A string containing a comma-separated list defining the :ref:`ds-geo-limit-countries`\ [#geolimit]_
+:geoLimitCountriesList:     A string or string array containing a comma-separated list defining the :ref:`ds-geo-limit-countries-list`
```
ocket8888, line 301: This isn't accurate anymore

docs/source/api/v4/deliveryservices.rst
```diff
@@ -448,6 +450,7 @@ Response Structure
 :fqPacingRate:              The :ref:`ds-fqpr`
 :geoLimit:                  An integer that defines the :ref:`ds-geo-limit`
 :geoLimitCountries:         A string containing a comma-separated list defining the :ref:`ds-geo-limit-countries`
+:geoLimitCountriesList:     A string or string array containing a comma-separated list defining the :ref:`ds-geo-limit-countries-list`
```
ocket8888, line 453: This isn't accurate anymore

docs/source/api/v4/deliveryservices_id.rst
```diff
@@ -53,6 +53,7 @@ Request Structure
 :fqPacingRate:              The :ref:`ds-fqpr`
 :geoLimit:                  An integer that defines the :ref:`ds-geo-limit`
 :geoLimitCountries:         A string containing a comma-separated list defining the :ref:`ds-geo-limit-countries`\ [#geolimit]_
+:geoLimitCountriesList:     A string or string array containing a comma-separated list defining the :ref:`ds-geo-limit-countries-list`
```
ocket8888, line 56: This isn't accurate anymore

docs/source/api/v4/deliveryservices_id.rst
```diff
@@ -205,6 +206,7 @@ Response Structure
 :fqPacingRate:              The :ref:`ds-fqpr`
 :geoLimit:                  An integer that defines the :ref:`ds-geo-limit`
 :geoLimitCountries:         A string containing a comma-separated list defining the :ref:`ds-geo-limit-countries`
+:geoLimitCountriesList:     A string or string array containing a comma-separated list defining the :ref:`ds-geo-limit-countries-list`
```
ocket8888, line 209: This isn't accurate anymore

docs/source/api/v4/servers_id_deliveryservices.rst
```diff
@@ -92,6 +92,7 @@ Response Structure
 :fqPacingRate:              The :ref:`ds-fqpr`
 :geoLimit:                  An integer that defines the :ref:`ds-geo-limit`
 :geoLimitCountries:         A string containing a comma-separated list defining the :ref:`ds-geo-limit-countries`
+:geoLimitCountriesList:     A string or string array containing a comma-separated list defining the :ref:`ds-geo-limit-countries-list`
```
ocket8888, line 95: This isn't accurate anymore

docs/source/overview/delivery_services.rst
```diff
@@ -288,6 +288,12 @@ When `Geo Limit`_ is being used with this Delivery Service (and is set to exactl
 	|                  |                                                                           | country code - one should exist for each allowed country code                                  |
 	+------------------+---------------------------------------------------------------------------+------------------------------------------------------------------------------------------------+
 
+.. _ds-geo-limit-countries-list:
+
+Geo Limit Countries List
+------------------------
+When `Geo Limit`_ is being used with this Delivery Service (and is set to exactly ``2``), this is optionally a list of country codes to which access to content provided by the Delivery Service will be restricted. This can either be a single string representing a country code, or an array of strings. When creating a Delivery Service with this field or modifying the Geo Limit Countries List field on an existing Delivery Service, any amount of whitespace between country codes is permissible, as it will be removed on submission, but responses from the :ref:`to-api` should never include such whitespace.
+
```
ocket8888, line 296: This property doesn't exist in this PR anymore

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -606,7 +609,8 @@ func (ds *TODeliveryService) Read(h http.Header, useIMS bool) ([]interface{}, er
 		switch {
 		// NOTE: it's required to handle minor version cases in a descending >= manner
 		case version.Major > 3:
-			returnable = append(returnable, ds.RemoveLD1AndLD2())
+			dsV4 := ds.RemoveLD1AndLD2()
+			returnable = append(returnable, dsV4)
```
ocket8888, line 613: this doesn't appear to change anything - was this supposed to be included in this PR?

traffic_ops/traffic_ops_golang/deliveryservice/deliveryservices.go
```diff
@@ -2136,9 +2174,13 @@ func setNilIfEmpty(ptrs ...**string) {
 	}
 }
 
-func sanitize(ds *tc.DeliveryServiceV4) {
+func sanitize(ds *tc.DeliveryServiceV4) error {
```
ocket8888, line 2177: It is not possible, it seems, for this function to return a non-`nil` error - why bother returning an error value that never needs to be checked in practice?

ocket8888 (changes requested):
According to the title of this PR:

> Geo limit countries can now be parsed as an array **_and_** a string

(emphasis mine) but actually the old string-as-an-array doesn't work anymore: https://go.dev/play/p/AWo8KqQ6yvt

I'm fine with making that breaking change, and since it's in a new major API version we are free to do that, but I want to be sure that the intention is to make that breaking change. Furthermore, if we're going to be breaking it in that way, then we don't need a type alias at all, it can just be `[]string` since that's the only valid input. It's also worth noting that the output (i.e. in API responses) type is now changed as well to be an array of strings, which is already a breaking change anyway.

srijeet0406:
> According to the title of this PR:
> 
> > Geo limit countries can now be parsed as an array **_and_** a string
> 
> (emphasis mine) but actually the old string-as-an-array doesn't work anymore: https://go.dev/play/p/AWo8KqQ6yvt
> 
> I'm fine with making that breaking change, and since it's in a new major API version we are free to do that, but I want to be sure that the intention is to make that breaking change. Furthermore, if we're going to be breaking it in that way, then we don't need a type alias at all, it can just be `[]string` since that's the only valid input. It's also worth noting that the output (i.e. in API responses) type is now changed as well to be an array of strings, which is already a breaking change anyway.

I have changed the logic to be able to accept something like `"US,CA,IN"`, as well as `["US","CA","IN"]`.
As for the response part, yes, it will be an array of strings starting api v4.

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -98,6 +99,64 @@ func TestDeliveryServices(t *testing.T) {
...
+	deliveryServices, _, err := TOSession.GetDeliveryServices(opts)
+	if err != nil {
+		t.Fatalf("couldn't get ds: %v", err)
+	}
+	if len(deliveryServices.Response) != 1 {
+		t.Fatal("couldn't get exactly one ds in the response, quitting")
+	}
+	dsID := deliveryServices.Response[0].ID
+	customDS.GeoLimitCountries = []string{"US   ", "CA", "12"}
+	_, _, err = TOSession.UpdateDeliveryService(*dsID, customDS, client.RequestOptions{})
```
ocket8888, line 149: This line (and L153) will segfault if Traffic Ops erroneously returns a Delivery Service with a null or omitted ID

ocket8888 (commented):
Functionally this works perfectly, just have a few nitpicks.

ocket8888 (approved):

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -98,6 +99,67 @@ func TestDeliveryServices(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unable to get Types: %v - alerts: %+v", err, types.Alerts)
+	}
+	if len(types.Response) < 1 {
+		t.Fatal("expected at least one type")
+	}
+	customDS := getCustomDS(cdn.ID, types.Response[0].ID, "geo-limit-countries-test-ds-name", "edge", "https://test-geo-limit.com", "geo-limit-countries-test-ds-xml-id")
+	customDS.Protocol = util.IntPtr(0)
+	customDS.GeoLimit = util.IntPtr(2)
+	//geoLimitCountries := []string{"US   ", "CA"}
```
zrhoffman, line 121: Can this commented line be removed?

PR #6444, opened 2022-01-03
Remove unnecessary logging of SSL cert/key paths

ocket8888:
Currently, if `cdn.conf` specifies cert/key paths that don't match the values supplied to postinstall, it will log all four specified SSL key/cert paths. This is a bit unnecessary, since anyone with the permissions to be setting those values usefully will already have access to all four, and we need not clutter the logging output with it.

<hr/>

- Traffic Ops (postinstall)

Make sure postinstall tests still pass, make a Traffic Ops config then run postinstall with mismatched values for the ssl key/cert paths and make sure they don't get logged.

## PR submission checklist
- [ ] This PR has tests


zrhoffman (approved):
Looks good!

Issue #5118, opened 2020-10-07
Traffic Stats -- Support Kafka as an output


dneuman64:
## I'm submitting a ...

-  new feature / enhancement request

## Traffic Control components affected ...

-  Traffic Stats

## Current behavior:

Currently, Traffic Stats can only be configured to send TS data to InfluxDB.

## New behavior:

Traffic Stats can be configured to send data to InfluxDB, Kafka, or Both. 
The configuration will support all information needed to send to Kafka, including TLS support.

smalenfant:
@dneuman64 We accomplished this goal by sending to telegraf and using the Kafka output. This was mostly to bridge the data to Elasticsearch. Although a direct implementation would be more robust. We used the influx_relay output to telegraf. 

```
[[inputs.influxdb_listener]]
  service_address = ":8186"
  read_timeout = "10s"
  max_body_size = 0
  database_tag = "database"

[[outputs.kafka]]
  ## URLs of kafka brokers
  brokers = ["kafka:9092"]
  topic = "telegraf"
  data_format = "json"
  
  [outputs.kafka.tagpass]
    database = ["deliveryservice_stats"]
```

PR #6451, opened 2022-01-07
Adding Support for Kafka to Traffic Stats

henryluimcit:
This PR closes [#5118](https://github.com/apache/trafficcontrol/issues/5118)

<hr>

- Traffic Stats

Within the `traffic_stats.cfg` file, change the `kafkaConfig` to connect to the appropriate local or TLS authenticated server:

```cfg
	"kafkaConfig": {
	    "enable": false,
	    "brokers": "localhost:9092",
	    "topic": "traffic-stats",
	    "required_acks": 1,
	    "enable_tls": false,
	    "root_ca": "",
	    "client_cert": "",
	    "client_cert_key": ""
	},
```

To enable sending data to Kafka, set `enable` to `true`. Add the broker(s) and topic. If using TLS authentication, set `enable_tls` to `true`, and add the path to the CA certificate, client certificate, and client certificate key respectively.

Make sure that the Kafka server is running before starting Traffic Stats. If starting a Kafka environment for the first time, instructions can be found [here](https://kafka.apache.org/quickstart) using steps 1-3.

In order to see the messages pushed by the Kafka producer locally, a consumer can fetch the data. Here's example local consumer using the broker and topic from above:

```
bin/kafka-console-consumer.sh --bootstrap-server localhost:9092 --topic traffic-stats --group my-traffic-stats-app --from-beginning
```

InfluxDB can also be disabled in `traffic_stats.cfg`. Set `enableInflux` to `false` if Influx is not needed. However, daily stats cannot be written since Traffic Stats queries InfluxDB for daily stats, and Grafana queries InfluxDB and thus will not display charts in Traffic Portal.

## PR submission checklist
- [x] This PR does not have tests. A Kafka environment must be set up to test manually.


traffic_stats/traffic_stats.go
```diff
@@ -243,11 +273,18 @@ func main() {
...
+				if c != nil {
+					publishToKafka(config, val, c)
+				}
 				sendMetrics(config, val, false)
 			}
+			startShutdown(c)
 			os.Exit(0)
 		case <-tickers.Publish:
 			for key, val := range Bps {
+				if c != nil {
```
rawlinp, line 285: So, rather than checking if the kafka client isn't nil and if InfluxDB is enabled, I think it would make sense to add some kind of `type DataExporter interface { ExportData(bps influx.BatchPoints) }` of which there are two concrete implementations -- one for InfluxDB and one for Kafka. Then, we would have a `dataExporters := []DataExporter{}` which we would add the InfluxDB and Kafka instances to (if enabled). Then, the main goroutine would iterate through the `[]DataExporter` and call `ExportData(...)` on each one. This way, we don't have to constantly check all over the code if certain data exporters are enabled or not.

That said, for `calcDailySummary()`, if that's just going to be an influxDB thing only, then it might make sense to still check if influxDB is enabled in that function.

In the future, if someone wanted to add another data exporter implementation (e.g. Prometheus), this would make it a lot easier. What do you think of this idea?

traffic_stats/traffic_stats.go
```diff
@@ -92,6 +95,7 @@ type StartupConfig struct {
 	ToUser                      string   `json:"toUser"`
 	ToPasswd                    string   `json:"toPasswd"`
 	ToURL                       string   `json:"toUrl"`
+	EnableInflux                bool     `json:"enableInflux"`
```
rawlinp, line 98: Since influx is enabled by default and should remain so for backwards-compatibility, it might make sense to call this `DisableInflux` (and reverse the logic) since it will default to `false` if left out of existing configs (which is what we want). We don't want to require preexisting config files to have to add `enableInflux: true` in order to continue working as is.

traffic_stats/traffic_stats.go
```diff
@@ -243,11 +273,18 @@ func main() {
 		case <-termChan:
 			info("Shutdown Request Received - Sending stored metrics then quitting")
 			for _, val := range Bps {
+				if c != nil {
+					publishToKafka(config, val, c)
```
rawlinp, line 277: This line ignores any returned error

traffic_stats/traffic_stats.go
```diff
@@ -243,11 +273,18 @@ func main() {
...
+				}
 				sendMetrics(config, val, false)
 			}
+			startShutdown(c)
 			os.Exit(0)
 		case <-tickers.Publish:
 			for key, val := range Bps {
+				if c != nil {
+					go publishToKafka(config, val, c)
```
rawlinp, line 286: This line ignores any returned error

rawlinp (changes requested):

traffic_stats/traffic_stats.go
```diff
@@ -20,10 +20,13 @@ under the License.
 package main
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/Shopify/sarama"
```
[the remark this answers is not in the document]
henryluimcit: Fixed in latest commit

traffic_stats/traffic_stats.go
```diff
@@ -280,6 +317,131 @@ func main() {
 	}
 }
 
+func startShutdown(c *KafkaCluster) {
+	if c == nil {
+		return
+	}
+	go func() {
```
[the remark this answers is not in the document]
henryluimcit: Yes, the separate goroutine is unneeded in this case. I removed it in the latest commit

traffic_stats/traffic_stats.go
```diff
@@ -280,6 +317,131 @@ func main() {
...
+
+	for _, point := range bps.Points() {
+
+		var KafkaJSON KafkaJSON
+		KafkaJSON.Name = point.Name()
+		KafkaJSON.Tags = point.Tags()
+		KafkaJSON.Fields, _ = point.Fields()
+		KafkaJSON.Time = point.Time()
+
+		message, err := json.Marshal(KafkaJSON)
```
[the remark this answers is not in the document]
henryluimcit: Good catch - I put the error in the correct spot

traffic_stats/traffic_stats.go
henryluimcit: I created a DataExporter interface with a couple caveats: 

1. The retry (bool) parameter is only used for Influx
2. I used the startup config object to implement the method for ExportData (line 152) for Influx (line 284). I wasn't sure what else to use since nothing was needed.

traffic_stats/traffic_stats.go
henryluimcit: That makes sense - I changed to `DisableInflux`

traffic_stats/traffic_stats.go
henryluimcit: Fixed with ExportData method (line 147)

traffic_stats/traffic_stats.go
henryluimcit: Fixed with ExportData method (line 147)

henryluimcit: Fixed with ExportData method (line 147)

traffic_stats/traffic_stats.go
```diff
@@ -108,6 +112,45 @@ type StartupConfig struct {
...
+}
+
+func (c *KafkaCluster) ExportData(config StartupConfig, bps influx.BatchPoints, retry bool) {
+	err := publishToKafka(config, bps, c)
+	if err != nil {
+		log.Errorln("Unable to export to Kafka", err)
+	}
+}
+
+func (influx StartupConfig) ExportData(config StartupConfig, bps influx.BatchPoints, retry bool) {
```
rawlinp, line 152: Rather than tying this method to `StartupConfig`, it might make more sense to create a new `InfluxClient` struct (or some other name) to associate this interface method with.

traffic_stats/traffic_stats.go
```diff
@@ -229,6 +274,16 @@ func main() {
 	hupChan := make(chan os.Signal, 1)
 	signal.Notify(hupChan, syscall.SIGHUP)
 
+	dataExporters := []DataExporter{}
+
+	if config.KafkaConfig.Enable == true && c != nil {
```
rawlinp, line 279: This can be simplified to `if config.KafkaConfig.Enable && c != nil`

traffic_stats/traffic_stats.go
```diff
@@ -229,6 +274,16 @@ func main() {
...
+		dataExporters = append(dataExporters, c)
+	}
+
+	if config.DisableInflux == false {
```
rawlinp, line 283: This can be simplified to `if !config.DisableInflux`

traffic_stats/traffic_stats.go
```diff
@@ -280,6 +340,131 @@ func main() {
...
+	}
+	if err := (*c.client).Close(); err != nil {
+		log.Warnf("Error closing client for cluster:  %v", err)
+	}
+	c.producer = nil
+	log.Infof("Finished cluster shutdown")
+}
+
+func TlsConfig(config KafkaConfig) (*tls.Config, error) {
+	if config.EnableTls == false {
```
rawlinp, line 360: This can be simplified to `if !config.EnableTls`

traffic_stats/traffic_stats.go
```diff
@@ -280,6 +340,131 @@ func main() {
...
+			return nil, err
+		}
+		c.Certificates = []tls.Certificate{cert}
+	}
+	return c, nil
+}
+
+func newKakfaCluster(config KafkaConfig) *KafkaCluster {
+
+	if config.Enable == false {
```
rawlinp, line 400: This can be simplified to `if !config.Enable`

traffic_stats/traffic_stats.go
```diff
@@ -349,6 +534,10 @@ func loadStartupConfig(configFile string, oldConfig StartupConfig) (StartupConfi
 		warn("No logging configuration found in configuration file - default logging to stderr will be used")
 	}
 
+	if config.DisableInflux == true {
```
rawlinp, line 537: This can be simplified to `if config.DisableInflux`

traffic_stats/traffic_stats.go
```diff
@@ -370,6 +559,10 @@ func loadStartupConfig(configFile string, oldConfig StartupConfig) (StartupConfi
 }
 
 func calcDailySummary(now time.Time, config StartupConfig, runningConfig RunningConfig) {
+	if config.DisableInflux == true {
```
rawlinp, line 562: This can be simplified to `if config.DisableInflux`

rawlinp (changes requested):

traffic_stats/traffic_stats.go
ocket8888: I'd like to point out once again that with this change our configuration file for TO will have both `disableInflux` and `influxEnabled` - and I don't think that's ideal

rawlinp: This isn't the TO config file -- this is the Traffic Stats config file.

rawlinp:
This looks good to me, but there are merge conflicts to address.

rawlinp (commented):
Oh, we vendor our dependencies in the repo currently, so I think all that's left is to run `go mod vendor` in the `trafficcontrol` dir and add those new dependencies.

vendor/modules.txt
```diff
@@ -10,6 +10,11 @@ github.com/GehirnInc/crypt
 github.com/GehirnInc/crypt/common
```
rawlinp, line 10: were there no other additions to the `vendor` directory?

henryluimcit: Added vendor files in most recent commit

go.mod
```diff
@@ -51,6 +51,7 @@ require (
 	golang.org/x/crypto v0.0.0-20211215153901-e495a2d5b3d3
```
zrhoffman, line 51: The following dependencies are unused and can be removed from `go.mod`:
* github.com/hashicorp/go-uuid
* github.com/jcmturner/gofork
* github.com/klauspost/compress
* gopkg.in/jcmturner/aescts.v1
* gopkg.in/jcmturner/dnsutils.v1
* gopkg.in/jcmturner/gokrb5.v7
* gopkg.in/jcmturner/rpc.v1

henryluimcit: These are required dependencies for this PR

zrhoffman: Yep
> Never mind, I'm seeing all of these deps in `go.mod` when regenerating it from scratch.

rawlinp (approved):

PR #6455, opened 2022-01-07
Remove vendored dependencies

ocket8888:
Now that we have switched to Go modules¹, there's no need to commit dependencies in the vendoring directory to the repository

<hr/>

- Documentation
- Traffic Ops Client (Go)
- Traffic Monitor
- Traffic Ops
- Traffic Stats
- Grove
- CDN in a Box
- Automation (GHA, anything that relies on Go)

Make sure everything still works and there are no license issues.

## PR submission checklist
- [ ] This PR has tests
- [ ] This PR has a CHANGELOG.md entryhave a changelog entry.


go.sum
```diff
@@ -1317,6 +1317,7 @@ github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9de
 github.com/yuin/goldmark v1.1.32/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
```
zrhoffman, line 1317: Does the _Check Go Modules_ action pass without these changes to `go.mod` and `go.sum`?

zrhoffman (changes requested):

rob05c:
Vendoring prevents the build from breaking in the future, when people delete their github repos that TC depends on.

Internet resources disappearing is a common problem, and it makes it impossible to build older versions, if we don't vendor and include everything necessary to build without the internet. See: `left-pad`.

It also makes it possible to build if someone doesn't have an internet connection, or is in an area where one of the dependency URLs is blocked by a firewall.

I really think we should vendor, for those reasons.

go.sum
ocket8888: I'm not sure, but it also passes _with_ those changes. If it passes under both conditions there must be something wrong with the check, right? Surely it shouldn't be able to pass in both scenarios

zrhoffman: ec4e9b44eb upgrades `golang.org/x/tools` to `v0.1.8` and commits the corresponding `go.sum`. I also get those `go.sum` changes if I run `go mod tidy` after upgrading `golang.org/x/tools` to `v0.1.8`. So the check is not incorrect, just the https://github.com/apache/trafficcontrol/blob/ec4e9b44eb326b4f7d9a436cd7ec1300aa516b9a/go.mod#L78 line.

ocket8888: That change is inserted by `go mod tidy`, though. I did not manually alter either `go.mod` or `go.sum`. If the check (and all other tests) pass after the change, shouldn't it stay in? The master branch also updates those files whenever I build a Go component, so the underlying issue isn't introduced in this PR.

zrhoffman: > If you download this changeset and run `go mod tidy`, does it change anything for you?

no

rawlinp:
Yeah, this seems like it needs a mailing list discussion. There are pros and cons to keeping the vendored dependencies in the repo (@rob05c pointed out a big one). Another pro of vendoring dependencies is that it helps to understand the amount of code changed whenever a dependency is upgraded/added/removed.

ocket8888:
> this seems like it needs a mailing list discussion

Alright, no problem. I really just didn't expect anyone to come to the defense of vendoring :P

PR #6456, opened 2022-01-07
Resolves issue #6439 by changing the TM endpoint used to get cache status using publish/CrStates

### LINKED ISSUE #6439: TC Health Client uses TM's /api/cache-statuses API which does not participate in the health protocol

- Traffic Control Health Client (t3c-health-client)

## Current behavior:

The TC Health Client uses TM's `/api/cache-statuses` API for getting cache availability states in order to mark caches up/down. This is very dangerous because this API returns the TM's _local_ cache states rather than the combined cache states of itself and its peers using the optimistic health protocol. Using this API, the TC health client will set caches down when only the single requested TM thinks they're down. However, if that singular TM is on a network partition and can't reach those caches because of it, it will mark those caches down even though they might be perfectly reachable by its peers (and the broader client base).

## Expected behavior:

The TC Health Client should use TM's `/publish/CrStates` API for getting cache availability states. This API returns the _combined_ cache states of itself and its peers participating in the optimistic health protocol, and this API is what Traffic Router actually uses for routing clients to available caches.

In addition to the dangers of using TM's local cache states, `/api/cache-statuses` includes a lot of unnecessary data that the tc-health-client doesn't really need, whereas `/publish/CrStates` is simply just cache availability. We also shouldn't have Traffic Router and the tc-health-client using different APIs for getting cache states from TM.

## Steps to reproduce:

Disconnect a TM from the cache network or otherwise make it unable to reach one or more caches. Request `/api/cache-statuses` and `/publish/CrStates` and compare the results. As long as a cache is reachable and otherwise healthy, the other peers will consider it available, so the caches will be available in `/publish/CrStates` but not in `/api/cache-statuses`.

jrushford:
Closes #6439  by changing the TM endpoint used to get cache status using publish/CrStates

### DIFF
--- a/tc-health-client/tmagent/tmagent.go
+++ b/tc-health-client/tmagent/tmagent.go
@@ -38,7 +38,6 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/tc-health-client/config"
 	"github.com/apache/trafficcontrol/tc-health-client/util"
-	"github.com/apache/trafficcontrol/traffic_monitor/datareq"
 	"github.com/apache/trafficcontrol/traffic_monitor/tmclient"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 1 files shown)


jpappa200 (approved):
👍 LGTM

tc-health-client/tmagent/tmagent.go
```diff
@@ -38,7 +38,7 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/tc-health-client/config"
 	"github.com/apache/trafficcontrol/tc-health-client/util"
-	"github.com/apache/trafficcontrol/traffic_monitor/datareq"
+	//"github.com/apache/trafficcontrol/traffic_monitor/datareq"
```
zrhoffman, line 41: This comment should be removed:

zrhoffman (changes requested):

tc-health-client/tmagent/tmagent.go
jrushford: @zrhoffman yeah, agreed.  I overlooked that.

zrhoffman (approved):

Issue #6457, opened 2022-01-08
User registration and password reset are broken due to the `last_authenticated` value being null


srijeet0406:
## This Bug Report affects these Traffic Control components:

- Traffic Ops
- Traffic Portal

## Current behavior:

In Traffic Portal, try registering a new user, or resetting the password of an existing user. You'll notice that the operation fails. If you tail the logs, you'll see something similar to:
`getting current user: querying current user: sql: Scan error on column index 9, name "last_authenticated": unsupported Scan, storing driver.Value type <nil> into type *time.Time`

## Expected behavior:

Registration/ password reset should work as expected.

## Steps to reproduce:

In Traffic Portal, try registering a new user, or resetting the password of an existing user.

PR #6458, opened 2022-01-08
User registration and password reset are broken due to the last_authenticated value being null

srijeet0406:
This PR closes https://github.com/apache/trafficcontrol/issues/6457
<hr>

- Traffic Ops
- Traffic Portal

Run TP and TO, and verify that you can successfully register new users and reset passwords of existing users.

- master

## PR submission checklist
- [x] This PR does not have tests 
- [x] This PR does not have documentation 


traffic_ops/app/db/migrations/2022010715281600_remove_null_last_authenticated.up.sql
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+UPDATE public.tm_user
+SET last_authenticated = '1970-01-01 00:00:00.000000'
+WHERE last_authenticated IS NULL;
```
ocket8888, line 20: Is this migration necessary now that the `last_authenticated` column is being properly updated on token/oauth login?

rawlinp: It doesn't hurt really -- I think we could even make this column non-nullable with a default of beginning of epoch. There'd be one less nullable column to worry about at least.

ocket8888: This is one thing that I think ought to be "nullable", because if a user has never logged in `null` is correct but the epoch value is a lie. Yes, it is impossible for a user to have actually logged in before ATC existed, but it's just nice to never lie, so that users/clients don't need to translate our responses themselves into what they actually probably mean - our system should be doing that for them, IMO.

I won't hold up the PR for it, but if we're gonna make it non-nullable the Go structures should use `time.Time` instead of `*time.Time` to reflect that.

srijeet0406: I had it in there just as an added layer of protection, as Rawlin mentioned. I do agree with Brennan that if it is changed to non- nullable, we should change the Go structures as well. I'll let you guys decide which way we want to go.

ocket8888: well, I think you have to be the tiebreaker here. I vote we let it be null and it looks to me like Rawlin is voting we don't, at the risk of speaking for him.

traffic_portal/app/src/common/api/UserService.js
```diff
@@ -83,12 +83,26 @@ var UserService = function($http, locationUtils, userModel, messageModel, ENV) {
...
             function(result) {
-                if (userModel.user.id === user.id) {
+                if (userData.user != undefined) {
+                    userData = userData.user;
+                }
+                console.log(userData);
+                if (userModel.user.id === userData.id) {
                     // if you are updating the currently logged in user...
-                    userModel.setUser(user);
+                    userModel.setUser(userData);
```
ocket8888, line 105: To keep things simple, I'd recommend just adding a new method e.g. `updateCurrentUser` and have the controller determine which method to call, the same way we have `getUsers` as well as `getCurrentUser`.

traffic_ops/app/db/migrations/2022010715281600_remove_null_last_authenticated.up.sql
rawlinp: A null last_authenticated is still technically a lie for users that have authenticated before the feature was added, but I don't really care. We can't really answer the question of who has ever authenticated or not, only who has authenticated since this feature was deployed. If @ocket8888 prefers to keep the null values and delete these migrations, I think that's fine. I just generally like the idea of reducing the amount of nullable columns is all.

ocket8888: For the record, I also like that idea, in general, just not in this case. It does feel a little strange arguing _in favor_ of nullability.

We could improve it a bit - without, still, being able to always tell the complete truth in all situations - by adding a migration that sets `last_authenticated = now() WHERE last_authenticated IS NULL` with no "down" statements and no `NOT NULL` constraint/nullability change. That way you know that all users haven't logged in since _at least_ their `lastAuthenticated` time, which is set to the date/time at which the feature was added (roughly - it's been at least one minor release since then in reality, but it's behind an unstable API atm). I'd find that more acceptable than setting it to Jan 1, 1970. That way there's no synonym for `null` being used and `null` is still allowed for new users moving forward.

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -371,6 +378,13 @@ func OauthLoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
 			log.Errorf("checking local user: %s\n", err.Error())
 		}
 
+		_, dbErr := db.Exec(UpdateLoginTimeQuery, form.Username)
```
ocket8888, line 381: The "Last Authenticated" time of a user should only be updated if the user successfully authenticates, which isn't checked until line 388 currently.

traffic_portal/app/src/modules/private/user/UserController.js
```diff
@@ -51,6 +51,9 @@ var UserController = function($scope, $state, $location, $uibModal, formUtils, l
     $scope.user = userModel.user;
 
     $scope.confirmSave = function(user, usernameField) {
+        if (usernameField == undefined) {
```
ocket8888, line 54: comparisons should use `===` - under what conditions was this occurring?

traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -232,6 +232,13 @@ func TokenLoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
 			return
 		}
 
+		_, dbErr := db.Exec(UpdateLoginTimeQuery, username)
+		if dbErr != nil {
+			log.Errorf("unable to update authentication time for a given user: %s\n", dbErr.Error())
```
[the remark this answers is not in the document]
ocket8888: Actually, disregard, because it looks like you're double-logging this, because `api.HandleErr` will log it too. It looks like what you actually want something like:
```go
if dbErr != nil {
    dbErr = fmt.Errorf("unable to update authentication time for user '%s': %w", username, dbErr)
    api.HandleErr(w, r, nil, http.StatusInternalServerError, nil, dbErr)
    return
}
```
(I think the username is helpful information)

traffic_ops/app/db/migrations/2022010715281600_remove_null_last_authenticated.up.sql
srijeet0406: Sounds good, I'll set it to `now()` wherever it is `null`. Fun fact, I had it as `now()` when I first started debugging this issue, and then changed it to epoch start later.

traffic_portal/app/src/modules/private/user/UserController.js
srijeet0406: This happens when the user tries to Update the password after clicking on the `Reset Password` link in TP

traffic_portal/app/src/common/api/UserService.js
```diff
@@ -82,14 +82,14 @@ var UserService = function($http, locationUtils, userModel, messageModel, ENV) {
         );
     };
 
-    // todo: change to use query param when it is supported
-    this.updateUser = function(user) {
-        return $http.put(ENV.api.unstable + "users/" + user.id, user).then(
+    this.updateCurrentUser = function(user) {
+        // We should be using PUT 'user/current' to update the current user
+        var currUser = {
```
ocket8888, line 87: `currUser` is never reassigned, could be `const`.

traffic_portal/app/src/common/api/UserService.js
```diff
@@ -82,14 +82,14 @@ var UserService = function($http, locationUtils, userModel, messageModel, ENV) {
...
+            user: user
```
ocket8888, line 88: nit but `: user` is superfluous; `{ user }` is equivalent to `{ user: user }`. I find that to be cleaner, but if you're not used to it I bet you could be confused by that syntax. So change it or don't, just wanted to point it out.

srijeet0406: Yeah, I'm not used to the JS syntax at all. But I'll change it 👍

ocket8888 (approved):
This appears to fix the issue it's trying to fix, but I've discovered something troubling: the user is not required to set a password. The form can be submitted without setting a password, and that makes the token get dropped from the database, leaving the user with no way to authenticate, and that's not fixable without direct database manipulation. Also their cookie is immediately invalid so that the very next request 401's and there's no way to obtain a new cookie.

To reproduce: register a user, then go to the link in the email, set the full name as indicated by the form as being required, then submit the form. If you just wait like a minute a "newcount" logs request will be made that will cause a "user not found" error to be returned from the API and you'll be dropped back on the login page with no way to log in since you don't have a password.

This should be fixed, but if you think it's out-of-scope for this PR/just don't want to do it, I'll just merge it and open a new issue.

ocket8888:
Solving that problem is more complicated than I initially thought, because of LDAP and OAuth users. I'll open an issue for it probably tomorrow. This PR doesn't need to try to fix the problem.

PR #6507, opened 2022-01-11
Remove unneeded WriteHeader

shamrickus:
This PR removes an extra call to `(http.ResponseWriter).WriteHeader` causing the trafficops error log to be polluted with messages.

<hr/>

- Traffic Ops

Run the unit tests and inspect the TO `error.log` and you will see many entries such as:
```
ERROR: server.go:3170: http: superfluous response.WriteHeader call from 
  github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware.GzipResponse 
  (wrappers.go:240)
```
Run again with changes and verify that no error messages like that in the logs.

- 6.0.x
- master

## PR submission checklist
- [x] This PR has tests 


ocket8888 (approved):
Gets rid of the log spam, tests all still pass

PR #6523, opened 2022-01-25
Fixes the CrStates query used by the tc-health-client.

jrushford:
Fixes the CrStates query used by the tc-health-client.  Should not be using raw == true as that requests only
the local state.

Run the unit tests.

Examples:
- master

## PR submission checklist
- [x] This PR has tests 


rob05c (approved):
LGTM

PR #6537, opened 2022-02-03
Fix `traffic_vault_migrate` for PG backend

### LINKED ISSUE #6526: traffic_vault_migrate can no longer insert into PG

- Traffic Ops

## Current behavior:
Since #6337, a new column has been added to the SSLKeys table. This value is _not null_ and is not accounted for in traffic_vault_migrate causing all inserts to fail with `... pq: null value in column "provider" of relation "sslkey" violates not-null constraint`

## Expected behavior:
TVM should not fail when importing into PG.

## Steps to reproduce:
Create a SSL Key in Riak, attempt to use `traffic_vault_migrate` to migrate from Riak to PG, which will fail with a not-null constraint error.

shamrickus:
Fixes #6526 
<hr/>

- Tools

Run CiaB, dump the TV db using `./traffic_vault_migrate -t PG -d`. Then setup a local TV using `db/admin`. Finally import the keys from the dump into the local TV using `./traffic_vault_migrate -o PG --fill=dump`. The last step should succeed, where it would fail previously.

 - master
 - 6.1.x

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/app/db/traffic_vault_migrate/postgres.go
+++ b/traffic_ops/app/db/traffic_vault_migrate/postgres.go
@@ -348,8 +348,8 @@ type pgSSLKeyTable struct {
 }
 
 func (tbl *pgSSLKeyTable) insertKeys(db *sql.DB) error {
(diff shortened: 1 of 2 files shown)


mattjackson220 (approved):
LGTM!

PR #6538, opened 2022-02-04
Correcting use of httpsPort instead of secure.port

mikeV02:
I was trying to change the TrafficRouter SecurePort value to other than 443, and tried to use the parameter secure.port added to the server.xml on the CCR profile in Traffic Portal. But noticed, that TcOps/TcMon, do not add that parameter to the CrCronfig.json.

I digged into the code and fount that the use of secure.port was added to TrafficRouter on the following commit: https://github.com/apache/trafficcontrol/commit/8ff9006beb0639a21ac3235cdeaa15bf852cfc1a However, no counter part was added to TcOps to actually state the value secure.port from the TrafficPortal paramenters. The value that TcOps adds to the CrConfig.json is the httpsPort, from the server configuration itself on TrafficPortal, but that parameter is not used anywhere by Traffic Router (from what I saw).

I see that it would require just two lines edits on traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/LanguidState.java to make use of httpsPort and not the unexisting secure.port.

Basically, edit lines 86 and 87 of the file, and replace secure.port for httpsPort
<hr/>

- Traffic Router

Try adding the secure.port parameter to the CCR profile on Traffic Portal. It does not matter if you add it to server.xml or CrConfig,json, Traffic Ops never assigns secure.port to the traffic routers under that profile. Applying the change on the file on this PR, makes TR use the httpsPort configured under the server configuration un Traffic Portal.

All

## PR submission checklist


traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/LanguidState.java
```diff
@@ -83,8 +83,8 @@ private void initPorts(final JsonNode routerJson) {
 			trafficRouterManager.setApiPort(apiPort);
 		}
 
-		if (routerJson.has("secure.port")) {
-			setSecurePort(routerJson.get("secure.port").asInt());
+		if (routerJson.has("httpsPort")) {
```
rawlinp, line 86: I believe this needs to use `hasNonNull` instead of `has`; otherwise, L87 could cause a null pointer exception for TRs that don't have `httpsPort` set (and hence have a `null` value). In fact, just for extra safety/consistency, we should probably change all of the `has` checks into `hasNonNull` in this method.

rawlinp (changes requested):

rawlinp:
Also, I just want to point out that if you use a non-standard `httpsPort` for your TRs, all the delivery service example URLs for https will be wrong, since they'll be missing TR's non-standard https port suffix. There's not much we can do to fix that, but it shouldn't _really_ matter as long as the clients are given the correct URL with the non-standard https port. If we were to make `secure.port` a real per-CDN TR profile parameter, we could fix the example URLs for the delivery services.

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/LanguidState.java
mikeV02: You are totally correct on the use of hasNonNull. I changed line 86 to use the correct method. I did not change the other lines.

If for some reason someone does not put an http port, the field is not showed in the CrConfig.json. Thus, the if statement will evaluate false, and no exception will be triggered (at least not because of that line).

And if for the very odd reason someone creates the api.port and/or secure.api.port, assigns them to the CCR profile, but leave them as empty value. The CrConfig.json will show them as empty strings `secure.api.port: ""`. I did the test and TR seems to simply ignore them, and uses the default ones 3333/3443.

rawlinp (approved):
This looks good, but there are merge conflicts to resolve (most likely in CHANGELOG.md).

rawlinp (approved):

PR #6544, opened 2022-02-08
Layered profile `servers/` API {GET, POST,  PUT & DELETE}

rimashah25:
PR to update API servers/ and servers/{id} [GET, POST, PUT, DELETE only] with respect to server layered_profile feature https://github.com/apache/trafficcontrol/pull/6095

<hr/>

- Documentation
- Traffic Ops
- Traffic Control Client
- Traffic Ops ORT
- Traffic Portal
-  CDN-In-A-Box

Test GET, POST servers/, GET servers/{id}, PUT, DELETE servers/{id} with curl cmds and correct JSON requests.
For v4, POST and PUT have the following in json struct,
```
"profileNames": ["TRAFFIC_STATS"] 
OR 
"profileNames": ["TRAFFIC_PORTAL","GLOBAL"]
``` 
For v3, POST and PUT have the following in json struct,
```
        "profile": "TRAFFIC_OPS",
	"profileDesc": "Traffic Ops profile",
	"profileId": 3,
```

## PR submission checklist
- [x] This PR has tests 


lib/go-atscfg/parentdotconfig.go
```diff
@@ -1897,8 +1884,8 @@ func getProfileParentConfigParams(tcParentConfigParams []tc.Parameter) (map[stri
 func getServerParentConfigParams(server *Server, profileParentConfigParams map[string]map[string]string) map[string]string {
 	// We only need parent.config params, don't need all the params on the server
 	serverParams := map[string]string{}
-	if server.Profile == nil || *server.Profile != "" { // TODO warn/error if false? Servers requires profiles
-		for name, val := range profileParentConfigParams[*server.Profile] {
+	if server.Profiles == nil || (*server.Profiles)[0] != "" { // TODO warn/error if false? Servers requires profiles
+		for name, val := range profileParentConfigParams[(*server.Profiles)[0] {
```
ocket8888: Extra `[` here is causing compilation failure

ocket8888 (changes requested):

lib/go-atscfg/parentdotconfig.go
rimashah25: This was 4 commits ago

ocket8888:
These TP test failures are consistent and real. I [looks like](https://github.com/apache/trafficcontrol/runs/5412320747?check_suite_focus=true#step:11:265) it's failing to create servers because of the new format/requirements.

rimashah25:
> These TP test failures are consistent and real. I [looks like](https://github.com/apache/trafficcontrol/runs/5412320747?check_suite_focus=true#step:11:265) it's failing to create servers because of the new format/requirements.

Yes, I am aware and have been working on correcting that. The latest change should hopefully resolve the failure.

lib/go-tc/servers.go
```diff
@@ -700,6 +701,44 @@ type CommonServerProperties struct {
...
+	ILOPassword      *string              `json:"iloPassword" db:"ilo_password"`
+	ILOUsername      *string              `json:"iloUsername" db:"ilo_username"`
+	LastUpdated      *TimeNoMod           `json:"lastUpdated" db:"last_updated"`
+	MgmtIPAddress    *string              `json:"mgmtIpAddress" db:"mgmt_ip_address"`
+	MgmtIPGateway    *string              `json:"mgmtIpGateway" db:"mgmt_ip_gateway"`
+	MgmtIPNetmask    *string              `json:"mgmtIpNetmask" db:"mgmt_ip_netmask"`
+	OfflineReason    *string              `json:"offlineReason" db:"offline_reason"`
+	PhysLocation     *string              `json:"physLocation" db:"phys_location"`
+	PhysLocationID   *int                 `json:"physLocationId" db:"phys_location_id"`
+	Profiles         *pq.StringArray      `json:"profileNames" db:"profile_names"`
```
ocket8888, line 729: It's normal for the Go structure name and JSON tags to match except for case differences. IMO, the JSON tag ought to change to be simply `profiles`, but if you feel strongly that `profileNames` is better then the Go struct property name should just change to `ProfileNames`.

traffic_ops/app/db/migrations/2022011314075900_add_server_profile.up.sql
```diff
@@ -0,0 +1,29 @@
...
+                                                     profile_names text[] NOT NULL,
+                                                     priority bigint[] NOT NULL,
+    CONSTRAINT pk_server_profile PRIMARY KEY(profile_names, server),
+    CONSTRAINT fk_server_id FOREIGN KEY (server) REFERENCES public.server(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+INSERT into public.server_profile(server, profile_names, priority)
+    SELECT s.id, ARRAY [p.name], ARRAY [0]
+    FROM public.server AS s
+        JOIN public.profile p ON p.id=s.profile;
```
ocket8888, line 29: missing newline at EOF

traffic_ops/app/db/migrations/2022011314075900_add_server_profile.up.sql
```diff
@@ -0,0 +1,29 @@
...
+ * the License.
+ */
+
+CREATE TABLE IF NOT EXISTS public.server_profile (
+                                                     server bigint NOT NULL,
+                                                     profile_names text[] NOT NULL,
+                                                     priority bigint[] NOT NULL,
+    CONSTRAINT pk_server_profile PRIMARY KEY(profile_names, server),
+    CONSTRAINT fk_server_id FOREIGN KEY (server) REFERENCES public.server(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
```
ocket8888, line 24: Instead of
```
               Table "public.server_profile"
    Column     |   Type   | Collation | Nullable | Default 
---------------+----------+-----------+----------+---------
 server        | bigint   |           | not null | 
 profile_names | text[]   |           | not null | 
 priority      | bigint[] |           | not null | 
Indexes:
    "pk_server_profile" PRIMARY KEY, btree (profile_names, server)
Foreign-key constraints:
    "fk_server_id" FOREIGN KEY (server) REFERENCES server(id) ON UPDATE CASCADE ON DELETE CASCADE
```
Why not just 
```
           Table "public.server_profile"
  Column  |  Type  | Collation | Nullable | Default 
----------+--------+-----------+----------+---------
 server   | bigint |           | not null | 
 profile  | text   |           | not null | 
 priority | bigint |           | not null | 
Indexes:
    "pk_server_profile" PRIMARY KEY, btree (profile, server)
    "server_profile_unique_priority" UNIQUE CONSTRAINT, btree (server, priority)
Foreign-key constraints:
    "fk_server_profile_profile" FOREIGN KEY (profile) REFERENCES profile(name) ON UPDATE CASCADE ON DELETE RESTRICT
    "fk_server_profile_server" FOREIGN KEY (server) REFERENCES server(id) ON UPDATE CASCADE ON DELETE CASCADE
```
It's easier to work with, gives a slightly better guarantee of well-formed priorities (doesn't allow duplicates for the same server), and guarantees binding of Profile Names to actual Profiles - that is, under your current implementation, servers can have entries in their `profileNames` that aren't the names of any existing Profiles. This will happen if, for instance, the Name of a Profile is altered once it's been assigned to one or more Servers.

traffic_portal/app/src/common/modules/form/server/FormServerController.js
```diff
@@ -60,6 +60,14 @@ var FormServerController = function(server, $scope, $location, $state, $uibModal
             });
     };
 
+    $scope.getProfileID = function(profileName) {
+        for (let profile of $scope.profiles) {
+            if (profile.name == profileName) {
```
ocket8888, line 65: comparison should use `===`

traffic_portal/app/src/common/modules/form/server/FormServerController.js
```diff
@@ -60,6 +60,14 @@ var FormServerController = function(server, $scope, $location, $state, $uibModal
             });
     };
 
+    $scope.getProfileID = function(profileName) {
+        for (let profile of $scope.profiles) {
```
ocket8888, line 64: nit: `profile` is never reassigned - could be `const` instead.

traffic_portal/app/src/common/modules/form/server/edit/FormEditServerController.js
```diff
@@ -45,6 +45,7 @@ var FormEditServerController = function(server, $anchorScroll, $scope, $controll
     };
 
     $scope.save = function(server) {
+        console.log(server.profileNames = [server.profileName]);
```
ocket8888, line 48: This assignment has no return value, so this will only and always ever print `undefined` to the console. It shouldn't be logged.

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
```diff
@@ -112,11 +112,11 @@
 
             <label for="profile" ng-class="{'has-error': hasError(serverForm.profile)}">Profile *</label>
             <div ng-class="{'has-error': hasError(serverForm.profile), 'has-feedback': hasError(serverForm.profile)}">
-                <select id="profile" name="profile" class="form-control" ng-model="server.profileId" ng-options="profile.id as profile.name for profile in profiles" required>
+                <select id="profile" name="profile" class="form-control" ng-model="server.profileName" ng-options="profile.name as profile.name for profile in profiles" required>
```
ocket8888, line 115: we no longer need `as profile.name` here, since we're no longer valuing and labeling by different properties.

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
```diff
@@ -112,11 +112,11 @@
...
                     <option hidden selected disabled value="">Select...</option>
                 </select>
                 <small class="input-error" ng-show="hasPropertyError(serverForm.profile, 'required')">Required</small>
-                <small ng-show="server.profileId"><a ng-href="/#!/profiles/{{server.profileId}}" target="_blank">View Details&nbsp;&nbsp;<i class="fa fs-xs fa-external-link"></i></a></small>
+                <small ng-show="server.profileNames.length>0"><a ng-href="getProfileID(server.profileNames[0])" target="_blank">View Details&nbsp;&nbsp;<i class="fa fs-xs fa-external-link"></i></a></small>
```
ocket8888, line 119: I don't believe that `getProfileID(server.profileNames[0])` will return a valid relative or absolute URL.

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -365,6 +365,9 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
 			x.lastUpdated = x.lastUpdated ? new Date(x.lastUpdated.replace("+00", "Z")) : x.lastUpdated;
 			x.statusLastUpdated = x.statusLastUpdated ? new Date(x.statusLastUpdated): x.statusLastUpdated;
 			Object.assign(x, serverUtils.toLegacyIPInfo(x.interfaces));
+			if (x.profileNames != undefined) {
```
ocket8888, line 368: comparison should use `===`

ocket8888 (commented):
I think the TP code gets a lot simpler if you don't ever allow servers to have the `profileName` property and just always deal with the real data.

blueprints/layered-profile-server.md
```diff
@@ -93,7 +93,7 @@ With Layered Profiles, hundreds of profiles become a few dozen, each representin
 {
   "response": [{
     "id": 5,
-    "profiles": ["MID", "AMIGA_123", "CDN_FOO"]
+    "profileNames": ["MID", "AMIGA_123", "CDN_FOO"]
```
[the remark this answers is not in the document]
rimashah25: To make it more clear.. I felt profiles was generic.

lib/go-tc/servers.go
```diff
@@ -1005,6 +1005,47 @@ func (s ServerNullableV2) UpgradeToV40(cspV40 CommonServerPropertiesV40) (Server
 	return upgraded, nil
 }
 
+// UpdateCommonServerPropertiesV40 updates CommonServerProperties of V2 and V3 to CommonServerPropertiesV40
+func UpdateCommonServerPropertiesV40(profileName pq.StringArray, properties CommonServerProperties) CommonServerPropertiesV40 {
```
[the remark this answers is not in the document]
rimashah25: You are right, I can get away with it..will update the code accordingly.

lib/go-tc/servers.go
```diff
@@ -718,6 +757,15 @@ type ServerNullableV2 struct {
 	IP6IsService *bool `json:"ip6IsService" db:"ip6_address_is_service"`
 }
 
+type ServerNullableV40 struct {
```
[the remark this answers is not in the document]
rimashah25: I was updating all structs for V40 and hence created this one, even though I ended up not using it. This was a left over from earlier commits. It will be removed.

lib/go-tc/servers.go
```diff
@@ -718,6 +757,15 @@ type ServerNullableV2 struct {
...
+	LegacyInterfaceDetails
```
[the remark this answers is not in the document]
rimashah25: Okie. This struct will be deleted, so it is a moot point.

lib/go-tc/servers.go
```diff
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/lib/pq"
```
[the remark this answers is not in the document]
rimashah25: Ok

lib/go-tc/servers.go
rimashah25: Ok, will pick one option and change accordingly.

blueprints/layered-profile-server.md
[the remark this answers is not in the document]
ocket8888: Idk, I think the typing makes it clear. IDs are numbers, names are strings. Typically we use `profileID` and `profileName` because objects have _both_, so you need to be more specific.

I, personally, like `profiles` better, but it doesn't really matter. `profileNames` is fine; change it or don't, but it should be consistent everywhere (i.e. in the Go Struct - I think I have a comment about that later).

traffic_portal/app/src/common/modules/form/server/edit/FormEditServerController.js
rimashah25: I forgot to remove console.log after i finished with my checks.

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
rimashah25: we do.. it didn't work when `as profile.name` was removed.

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
rimashah25: Why not, the function returns `"/#!/profiles/"+profile.id`.. I did test it.

lib/go-tc/servers.go
```diff
@@ -718,6 +757,15 @@ type ServerNullableV2 struct {
 	IP6IsService *bool `json:"ip6IsService" db:"ip6_address_is_service"`
 }
 
+type ServerNullableV40 struct {
+	LegacyInterfaceDetails
+	CommonServerPropertiesV40
+	RouterHostName *string `json:"routerHostName" db:"router_host_name"`
+	RouterPortName *string `json:"routerPortName" db:"router_port_name"`
```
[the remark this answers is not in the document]
rimashah25: Since I removed the unused struct `ServerNullableV40`, this is not an issue.

lib/go-tc/servers.go
```diff
@@ -718,6 +757,15 @@ type ServerNullableV2 struct {
...
+	IPIsService    *bool   `json:"ipIsService" db:"ip_address_is_service"`
+	IP6IsService   *bool   `json:"ip6IsService" db:"ip6_address_is_service"`
```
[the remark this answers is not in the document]
rimashah25: Since I removed the unused struct `ServerNullableV40`, this is not an issue.

rimashah25:
> I think the TP code gets a lot simpler if you don't ever allow servers to have the `profileName` property and just always deal with the real data.

I am not sure I follow. If you are available for huddle, lets do it.

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
ocket8888: yeah, you're right. My bad.

lib/go-tc/servers.go
rimashah25: I chose `ProfileNames`.

traffic_ops/app/db/migrations/2022011314075900_add_server_profile.up.sql
ocket8888: You don't need to know what `btree` is to get that table, that's just an implementation detail (if you look above the current implementation's table also uses `btree` in its primary key, even though that term appears nowhere in your migration).

The priority and profile identity would still be arrays within the resulting Go structure, you'd just select them into an array from the multiple rows in the table. Even though you'd have to build an array in the read queries, I still think it's easier to use a table-based array because since the priority in the API is inherent in the list order, you have to use the priorities to sort the results which is much harder using two arrays instead of two columns.

Using your table, you select the arrays like this:
```sql
SELECT array_agg(profile_name)
FROM unnest(
	SELECT profile_names
	FROM public.server_profile
	WHERE "server" = $1
) WITH ORDINALITY AS profile_name
JOIN unnest(
	SELECT priority
	FROM public.server_profile
	WHERE "server" = $1
) WITH ORDINALITY AS "priority"
ON profile_name.ordinality = priority.ordinality
ORDER BY priority;
```

Using the table I suggested, you could just do:
```sql
SELECT array_agg(profile_name)
FROM public.server_profile
WHERE "server" = $1
ORDER BY "priority";
```

ocket8888: I ran those through `explain` and apparently they're more or less equal in terms of performance. The latter is slightly lower cost, but only by like 1/14th of the total (and idk those units but that's probably insignificant).

To see these in action, I did a sample of each table type and each query to show they give the same output:

<details><summary><h2>Array columns</h2></summary>

### Table Data
```
 server |                 profile_names                 | priority 
--------+-----------------------------------------------+----------
      1 | {ATS_MID_XP,ATS_MID_LINEAR,ATS_MID_DS145HACK} | {1,2,3}
      2 | {A,B}                                         | {2,1}
(2 rows)
```

### Queries and Results

```sql
SELECT array_agg(profile_name ORDER BY priority.priority) AS profile_names
FROM sample_A,
unnest(sample_A.profile_names) WITH ORDINALITY AS profile_name
JOIN
unnest(sample_A.priority) WITH ORDINALITY AS "priority"
ON profile_name.ordinality = priority.ordinality
WHERE sample_A.server = 1;

                profile_names                   
 {ATS_MID_XP,ATS_MID_LINEAR,ATS_MID_DS145HACK}
(1 row)
```

```sql
SELECT array_agg(profile_name ORDER BY priority.priority) AS profile_names
FROM sample_A,
unnest(sample_A.profile_names) WITH ORDINALITY AS profile_name
JOIN
unnest(sample_A.priority) WITH ORDINALITY AS "priority"
ON profile_name.ordinality = priority.ordinality
WHERE sample_A.server = 2;

 profile_names 
---------------
 {B,A}
(1 row)
```

</details>

<hr/>

<details><summary><h2>Simple columns</h2></summary>

### Table Data

```
traffic_ops_development=# select * from sample_B;
 server |   profile_name    | priority 
--------+-------------------+----------
      1 | ATS_MID_XP        |        1
      1 | ATS_MID_LINEAR    |        2
      1 | ATS_MID_DS145HACK |        3
      2 | A                 |        2
      2 | B                 |        1
(5 rows)
```

### Queries and Results

```sql
SELECT array_agg(profile_name ORDER BY "priority") AS profile_names
FROM sample_b
WHERE "server" = 1;

                 profile_names                 
 {ATS_MID_XP,ATS_MID_LINEAR,ATS_MID_DS145HACK}
```

```sql
SELECT array_agg(profile_name ORDER BY "priority") AS profile_names
FROM sample_b
WHERE "server" = 2;

 profile_names 
---------------
 {B,A}
(1 row)
```

</details>

rawlinp: I haven't been following this closely, but I believe @ocket8888 is correct -- we should generally avoid array-type columns (which are basically unique to PostgreSQL) in order to maintain referential integrity. This means we don't have referential integrity:
> under your current implementation, servers can have entries in their profileNames that aren't the names of any existing Profiles

lib/go-tc/servers.go
```diff
@@ -700,6 +701,44 @@ type CommonServerProperties struct {
...
+	ILOPassword      *string              `json:"iloPassword" db:"ilo_password"`
+	ILOUsername      *string              `json:"iloUsername" db:"ilo_username"`
+	LastUpdated      *TimeNoMod           `json:"lastUpdated" db:"last_updated"`
+	MgmtIPAddress    *string              `json:"mgmtIpAddress" db:"mgmt_ip_address"`
+	MgmtIPGateway    *string              `json:"mgmtIpGateway" db:"mgmt_ip_gateway"`
+	MgmtIPNetmask    *string              `json:"mgmtIpNetmask" db:"mgmt_ip_netmask"`
+	OfflineReason    *string              `json:"offlineReason" db:"offline_reason"`
+	PhysLocation     *string              `json:"physLocation" db:"phys_location"`
+	PhysLocationID   *int                 `json:"physLocationId" db:"phys_location_id"`
+	ProfileNames     *pq.StringArray      `json:"profileNames" db:"profile_names"`
```
rawlinp, line 729: I think we probably want this to be a plain `[]string` instead of `*pq.StringArray` -- otherwise, we will have to import `github.com/lib/pq` into any TO client that uses it. `pq.StringArray` should really only be used to read from the database, then once it's read it can be converted to `[]string` to be used from that point on.

rimashah25: yeah, I know.. I had already changed it before your comment came through.. but thanks.

lib/go-atscfg/astatsdotconfig.go
```diff
@@ -47,7 +47,7 @@ func MakeAStatsDotConfig(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 50: This `len` check will never be used in a non-tautological way. If the length is 0 but it's not nil, the first term is `false` so it `&&` anything can't be true and the second term isn't checked. If it's `nil`, the length is always zero because `len` on a `nil` slice returns `0`.

This should just check the length.

lib/go-atscfg/atsdotrules.go
```diff
@@ -46,7 +46,7 @@ func MakeATSDotRules(
 		opt = &ATSDotRulesOpts{}
 	}
 	warnings := []string{}
-	if server.Profile == nil {
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 49: same as above

lib/go-atscfg/cachedotconfig.go
```diff
@@ -67,22 +67,24 @@ func makeCacheDotConfigEdge(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
-		return Cfg{}, makeErr(warnings, "server missing profile")
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 70: same as above

lib/go-atscfg/cachedotconfig.go
```diff
@@ -67,22 +67,24 @@ func makeCacheDotConfigEdge(
...
+		return Cfg{}, makeErr(warnings, "server missing profiles")
 	}
 
 	profileServerIDsMap := map[int]struct{}{}
 	for _, sv := range servers {
-		if sv.Profile == nil {
+		if sv.ProfileNames == nil && len(sv.ProfileNames) == 0 {
```
ocket8888, line 76: same as above

lib/go-atscfg/cachedotconfig.go
```diff
@@ -67,22 +67,24 @@ func makeCacheDotConfigEdge(
...
 		if sv.ID == nil {
 			warnings = append(warnings, "servers had server with nil id, skipping!")
 			continue
 		}
-		if *sv.Profile != *server.Profile {
-			continue
+		for i, _ := range server.ProfileNames {
+			if sv.ProfileNames[i] != server.ProfileNames[i] {
+				continue
+			}
```
ocket8888, line 87: This doesn't replicate the logic it's replacing. The `continue` here is associated with your new `for`. So you'd need to do something like
```go
if len(sv.ProfileNames) != len(server.ProfileNames) {
    continue
}
profilesTheSame := true
for i, _ := range server.ProfileNames {
    if sv.ProfileNames[i] != server.ProfileNames[i] {
        profilesTheSame = false
        break
    }
}
if !profilesTheSame {
    continue
}
```
Note that you should also check the lengths before iterating, otherwise it could segfault.

lib/go-atscfg/dropqstringdotconfig.go
```diff
@@ -46,7 +46,7 @@ func MakeDropQStringDotConfig(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 49: Same as above RE: just check `len`

lib/go-atscfg/facts.go
```diff
@@ -39,13 +41,13 @@ func Make12MFacts(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 44: Same as above RE: just check `len`

lib/go-atscfg/facts.go
```diff
@@ -39,13 +41,13 @@ func Make12MFacts(
...
 		return Cfg{}, makeErr(warnings, "this server missing Profile")
 	}
 
 	hdr := makeHdrComment(opt.HdrComment)
 	txt := hdr
-	txt += "profile:" + *server.Profile + "\n"
+	txt += "profile:" + strings.Join(server.ProfileNames, ",") + "\n"
```
ocket8888, line 50: The header comment is meant to be human readable, so grammar and spacing count. `profile:` should probably be `profiles:` now that there are multiple, and the Profile names should be joined with `", "` to make reading the list easier.

lib/go-atscfg/loggingdotconfig.go
```diff
@@ -53,7 +53,7 @@ func MakeLoggingDotConfig(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 56: Same as above RE: just check `len`

lib/go-atscfg/loggingdotyaml.go
```diff
@@ -50,7 +50,7 @@ func MakeLoggingDotYAML(
 	warnings := []string{}
 	requiredIndent := 0
 
-	if server.Profile == nil {
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 53: Same as above RE: just check `len`

lib/go-atscfg/logsdotxml.go
```diff
@@ -49,7 +49,7 @@ func MakeLogsXMLDotConfig(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 52: Same as above RE: just check `len`

lib/go-atscfg/meta.go
```diff
@@ -69,7 +67,7 @@ func MakeConfigFilesList(
 		return nil, warnings, errors.New("server missing CDNName")
 	} else if server.ID == nil {
 		return nil, warnings, errors.New("server missing ID")
-	} else if server.Profile == nil {
+	} else if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 70: Same as above RE: just check `len`

lib/go-atscfg/parentdotconfig.go
```diff
@@ -162,7 +162,7 @@ func makeParentDotConfigData(
 		return nil, warnings, errors.New("server CDNName missing")
 	} else if server.Cachegroup == nil || *server.Cachegroup == "" {
 		return nil, warnings, errors.New("server Cachegroup missing")
-	} else if server.Profile == nil || *server.Profile == "" {
+	} else if server.ProfileNames == nil || len(server.ProfileNames) == 0 {
```
ocket8888, line 165: Same as above RE: just check `len`

lib/go-atscfg/parentdotconfig.go
```diff
@@ -1761,17 +1757,17 @@ func getParentConfigProfileParams(
 	warnings := []string{}
 	parentConfigServerCacheProfileParams := map[string]profileCache{} // map[profileName]ProfileCache
 	for _, cgServer := range cgServers {
-		if cgServer.Profile == nil {
-			warnings = append(warnings, "getting parent config profile params: server has nil profile, skipping!")
+		if cgServer.ProfileNames == nil {
```
ocket8888, line 1760: Same as above, RE: check `len` instead

lib/go-atscfg/parentdotconfig_test.go
```diff
@@ -3152,7 +3152,7 @@ func warningsContains(warnings []string, str string) bool {
 
 func makeTestParentServer() *Server {
 	server := &Server{}
-	server.ProfileID = util.IntPtr(42)
+	//server.ProfileID = util.IntPtr(42)
```
ocket8888, line 3155: This line should be removed instead of just commented out

lib/go-atscfg/parentdotconfig_test.go
```diff
@@ -3162,8 +3162,8 @@ func makeTestParentServer() *Server {
 	server.HTTPSPort = util.IntPtr(12443)
 	server.ID = util.IntPtr(44)
 	setIP(server, "192.168.2.1")
-	server.ProfileID = util.IntPtr(46)
-	server.Profile = util.StrPtr("serverprofile")
+	//server.ProfileID = util.IntPtr(46)
```
ocket8888, line 3165: This line should also be removed instead of just commented out

lib/go-atscfg/plugindotconfig.go
```diff
@@ -45,8 +45,8 @@ func MakePluginDotConfig(
 		opt = &PluginDotConfigOpts{}
 	}
 	warnings := []string{}
-	if server.Profile == nil {
-		return Cfg{}, makeErr(warnings, "server profile missing")
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 48: Same as above RE: just check `len`

lib/go-atscfg/recordsdotconfig.go
```diff
@@ -66,8 +66,8 @@ func MakeRecordsDotConfig(
 		opt = &RecordsConfigOpts{}
 	}
 	warnings := []string{}
-	if server.Profile == nil {
-		return Cfg{}, makeErr(warnings, "server profile missing")
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 69: Same as above RE: just check `len`

lib/go-atscfg/remapdotconfig_test.go
```diff
@@ -8075,13 +8073,13 @@ func TestMakeRemapDotConfigStrategies(t *testing.T) {
 			Name:       "serverpkgval",
 			ConfigFile: "package",
 			Value:      "serverpkgval __HOSTNAME__ foo",
-			Profiles:   []byte(*server.Profile),
+			Profiles:   []byte(`["` + server.ProfileNames[0] + `"]`),
```
ocket8888, line 8076: Previously, this was a string containing just the Profile name. It's now surrounded by square brackets and quotes - why?

lib/go-atscfg/sslservernamedotyaml.go
```diff
@@ -210,8 +210,8 @@ func GetServerSSLData(
 ) ([]SSLData, []string, error) {
 	warnings := []string{}
 
-	if server.Profile == nil {
-		return nil, warnings, errors.New("this server missing Profile")
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 213: Same as above RE: just check `len`

lib/go-atscfg/storagedotconfig.go
```diff
@@ -51,8 +51,8 @@ func MakeStorageDotConfig(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
-		return Cfg{}, makeErr(warnings, "server missing Profile")
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 54: Same as above RE: just check `len`

lib/go-atscfg/sysctldotconf.go
```diff
@@ -45,8 +45,8 @@ func MakeSysCtlDotConf(
 		opt = &SysCtlDotConfOpts{}
 	}
 	warnings := []string{}
-	if server.Profile == nil {
-		return Cfg{}, makeErr(warnings, "server missing Profile")
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 48: Same as above RE: just check `len`

lib/go-atscfg/urlsigconfig.go
```diff
@@ -49,8 +49,8 @@ func MakeURLSigConfig(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
-		return Cfg{}, makeErr(warnings, "server missing Profile")
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 52: Same as above RE: just check `len`

lib/go-atscfg/volumedotconfig.go
```diff
@@ -49,8 +49,8 @@ func MakeVolumeDotConfig(
 	}
 	warnings := []string{}
 
-	if server.Profile == nil {
-		return Cfg{}, makeErr(warnings, "server missing Profile")
+	if server.ProfileNames == nil && len(server.ProfileNames) == 0 {
```
ocket8888, line 52: Same as above RE: just check `len`

ocket8888 (commented):
I've only gone through the t3c and atstccfg code so far, but there are a few comments so I'm submitting that while I go through the rest.

lib/go-atscfg/remapdotconfig_test.go
rimashah25: Because the profile name is an array and resolving only the first element.

lib/go-atscfg/cachedotconfig.go
rimashah25: Hmm.. I see.

lib/go-atscfg/remapdotconfig_test.go
ocket8888: Profiles was already an array in [Parameter responses](https://traffic-control-cdn.readthedocs.io/en/latest/api/v2/parameters.html#response-structure)

rimashah25: ok, it takes json a RawMessage.. so i needn't do anything..copy.

zrhoffman:
Looks like all of the review comments are addressed, is anything needed before #6544 can be rechecked for mergeworthiness?

traffic_portal/app/src/common/modules/form/server/form.server.tpl.html
```diff
@@ -112,11 +112,11 @@
 
             <label for="profile" ng-class="{'has-error': hasError(serverForm.profile)}">Profile *</label>
             <div ng-class="{'has-error': hasError(serverForm.profile), 'has-feedback': hasError(serverForm.profile)}">
-                <select id="profile" name="profile" class="form-control" ng-model="server.profileId" ng-options="profile.id as profile.name for profile in profiles" required>
+                <select id="profile" name="profile" class="form-control" ng-model="server.profileNames" multiple ng-options="profile.name as profile.name for profile in profiles" required>
```
ocket8888, line 115: I'm _pretty_ sure that you can bind this to `server.profileNames[0]` if that property is properly initialized to an array in `traffic_portal/app/src/modules/private/servers/new/index.js`, but I still have to test that. I'll report back here once I have and either confirm or resolve the conversation.

ocket8888: Yup, I just checked, and with an initialized property you can bind to `server.profileNames[0]` without the `multiple` attribute.

shamrickus: If you do that server.profileNames will be:
```
[
  {
    "0": "selected value"
  }
]
```
instead of
```
[
  "selected value"
]
```

ocket8888: Nah. It'll do that if you don't get rid of the `multiple` attribute. I just tested it, it works fine, sends `["selected value"]` as expected. Check out https://github.com/rimashah25/trafficcontrol/pull/5

traffic_ops/app/db/migrations/2022032914235900_add_server_profile_table.up.sql
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+ * License for the specific language governing permissions and limitations undera
+ * the License.
+ */
+
+CREATE TABLE IF NOT EXISTS public.server_profile (
+                                                     server bigint NOT NULL,
+                                                     profile_name text NOT NULL,
+                                                     priority bigint NOT NULL,
```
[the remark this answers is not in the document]
rimashah25: That's fine, I am not stuck on `bigint`. We can go with `int`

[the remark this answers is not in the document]
rimashah25: better to be safe than sorry.

traffic_ops/testing/api/v4/tc-fixtures.json
```diff
@@ -2817,8 +2817,8 @@
                             "serviceAddress": false
                         },
                         {
-                            "address": "127.0.0.13/30",
-                            "gateway": "127.0.0.1",
+                            "address": "128.0.0.13/30",
+                            "gateway": "128.0.0.1",
```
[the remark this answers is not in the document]
rimashah25: Since there were issues in API v4 tests that were causing failures. Fix: https://github.com/apache/trafficcontrol/pull/6544/commits/e2b9c6d754f374a16a8d4779fcb63e56ebea9abb#diff-848f3f399dd9b0437c562368648f7acd64533910f7ee2740decb4ff00d92e380
Issue: https://github.com/apache/trafficcontrol/runs/5315051608?check_suite_focus=true

traffic_ops/testing/api/v4/tc-fixtures.json
```diff
@@ -3014,8 +3014,8 @@
                             "serviceAddress": false
                         },
                         {
-                            "address": "127.0.0.17/30",
-                            "gateway": "127.0.0.17",
+                            "address": "128.0.0.17/30",
+                            "gateway": "128.0.0.17",
```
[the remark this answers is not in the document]
rimashah25: Same as above.

traffic_ops/testing/api/v4/tc-fixtures.json
```diff
@@ -3352,8 +3352,8 @@
                 {
                     "ipAddresses": [
                         {
-                            "address": "127.0.0.1/30",
-                            "gateway": "127.0.0.1",
+                            "address": "128.0.0.1/30",
+                            "gateway": "128.0.0.1",
```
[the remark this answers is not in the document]
rimashah25: Same as above.

traffic_ops/testing/api/v4/tc-fixtures.json
```diff
@@ -4100,8 +4100,8 @@
                             "serviceAddress": false
                         },
                         {
-                            "address": "192.0.2.13/24",
-                            "gateway": "192.0.2.1",
+                            "address": "192.0.3.13/24",
+                            "gateway": "192.0.3.1",
```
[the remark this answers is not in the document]
rimashah25: Same as above.

zrhoffman:
There are conflicts now that #6569 is merged.

traffic_ops/testing/api/v4/tc-fixtures.json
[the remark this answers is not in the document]
ocket8888: That doesn't seem related - especially since those tests failed at all three API versions, but you only changed the v4 fixtures - which are run last, as well. The first problem I see in the APIv2 test logs is
```
    servers_test.go:39: Response:  edge2-cdn1-cg3   {[{server was created. success}]}
    servers_test.go:56: Failed to get at least one server by hostname 'edge2-cdn1-cg3'
```
and then it fails to delete that server, and fails to delete things associated with it, and then fails to create those things because they already exist. Each other API version failure looks similar, but it's hard to say. If a server with a hostname is created and then cannot later be filtered for in a GET, it seems like it caught a real bug in TO, and I can't understand why changing an IP address in a file not used by that test would fix that.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1788,3 +1788,102 @@ func GetRegionNameFromID(tx *sql.Tx, regionID int) (string, bool, error) {
...
+	if err == sql.ErrNoRows {
+		return fmt.Errorf("selecting server_profile by profile_name: " + err.Error())
+	}
+	return nil
+}
+
+// UpdateServerProfileTableForV2V3 updates CommonServerPropertiesV40 struct and server_profile table via Update (server) function for API v2/v3
+func UpdateServerProfileTableForV2V3(id *int, newProfile *string, origProfile string, tx *sql.Tx) ([]string, error) {
+	var profileName []string
+	query := `UPDATE server_profile SET profile_name=$1 WHERE server=$2 AND profile_name=$3`
```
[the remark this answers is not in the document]
rimashah25: Yes, the current array will go away with PUT APIv3 since it(profile) should be of length 1.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -468,9 +443,7 @@ RETURNING
 	offline_reason,
 	(SELECT name FROM phys_location WHERE phys_location.id=server.phys_location) AS phys_location,
 	phys_location AS phys_location_id,
-	profile AS profile_id,
-	(SELECT description FROM profile WHERE profile.id=server.profile) AS profile_desc,
-	(SELECT name FROM profile WHERE profile.id=server.profile) AS profile,
+	(SELECT ARRAY_AGG(profile_name) FROM server_profile WHERE server_profile.server=server.id) AS profile_name,
```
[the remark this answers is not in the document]
rimashah25: I'm ok either way.

traffic_ops/app/db/migrations/2022032914235900_add_server_profile_table.up.sql
```diff
@@ -0,0 +1,30 @@
...
+                                                     CONSTRAINT pk_server_profile PRIMARY KEY (profile_name, server),
+    CONSTRAINT fk_server_id FOREIGN KEY (server) REFERENCES public.server(id) ON DELETE CASCADE ON UPDATE CASCADE,
+    CONSTRAINT fk_server_profile_name_profile FOREIGN KEY (profile_name) REFERENCES public.profile(name) ON UPDATE CASCADE ON DELETE RESTRICT
+    );
+
+INSERT into public.server_profile(server, profile_name, priority)
+SELECT s.id, p.name, 0
+FROM public.server AS s
+         JOIN public.profile p ON p.id=s.profile;
```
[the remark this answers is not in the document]
rimashah25: Because we are not changing server table for this feature and to support backward compatibility.

traffic_portal/app/src/common/modules/form/server/FormServerController.js
```diff
@@ -60,6 +60,14 @@ var FormServerController = function(server, $scope, $location, $state, $uibModal
             });
     };
 
+    $scope.getProfileID = function(profileName) {
+        for (const profile of $scope.profiles) {
+            if (profile.name === profileName) {
+                return "/#!/profiles/"+profile.id
```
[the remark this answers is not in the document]
rimashah25: I'll let it be as I have coded.

traffic_ops/testing/api/v4/tc-fixtures.json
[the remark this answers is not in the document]
ocket8888: I just verified that those IP changes are the only things necessary to make the tests pass or fail. So you're right, but here's the thing. The error it's hitting (in v4 anyway, idk what's going in v2 and v2 isn't broken at the tip of your branch so I'm just gonna ignore that for now) is:
```
--- FAIL: TestCacheGroups (5.36s)
    servers_test.go:567: could not create server 'atlanta-org-1': error requesting Traffic Ops: path 'https://localhost:6443/api/4.0/servers' gave HTTP error 400 Bad Request - error-level alerts: there exists a server with id 6315 on the same profile that has the same service address 127.0.0.17/30 - alerts: [{Text:there exists a server with id 6315 on the same profile that has the same service address 127.0.0.17/30 Level:error}]
```
That's with the `tc-fixtures.json` from master just updated to use `profileNames` instead of `profile` on servers. There are 2 servers in the fixtures that use `127.0.0.17` - `atlanta-org-1` and `atlanta-mid-17` and both use it as their service address, that's true. But, `atlanta-mid-17` has `"profileNames": ["MID2"]` and `atlanta-org-1` has `"profileNames": ["EDGE1"]`.

So changing these IP addresses makes the tests pass, but only because it's hiding a bug. Two servers aren't allowed to share the same IP service address ever, because TO always thinks that they share a Profile. I believe I have a comment somewhere else asking about how we wanna go about handling that situation, but in any case what is happening right now is too restrictive; two different servers need to be able to share a service IP address as long as other conditions aren't met.

traffic_portal/app/src/common/modules/table/servers/TableServersController.js
```diff
@@ -162,7 +162,7 @@ var TableServersController = function(tableName, servers, filter, $scope, $state
 		},
 		{
 			headerName: "Profile",
-			field: "profile",
+			field: "profileName",
```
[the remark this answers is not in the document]
rimashah25: nope..

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1788,3 +1788,102 @@ func GetRegionNameFromID(tx *sql.Tx, regionID int) (string, bool, error) {
...
+		_, err := tx.Exec(query, *id, pName, i)
+		if err != nil {
+			return fmt.Errorf("error inserting into server_profile table, %v", err)
+		}
+	}
+
+	err = tx.QueryRow("SELECT ARRAY_AGG(profile_name) FROM server_profile WHERE server=$1", *id).Scan(pq.Array(&profileNames))
+	if err == sql.ErrNoRows {
+		return fmt.Errorf("selecting server_profile by profile_name: " + err.Error())
+	}
```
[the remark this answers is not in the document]
rimashah25: `But I do think reducing the number of queries to a constant is a good improvement to make.` Ok.. so nothing wrong with logic just need to reduce number of queries.

lib/go-tc/servers.go
```diff
@@ -700,6 +700,44 @@ type CommonServerProperties struct {
 	XMPPPasswd       *string              `json:"xmppPasswd" db:"xmpp_passwd"`
 }
 
+type CommonServerPropertiesV40 struct {
```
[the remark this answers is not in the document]
rimashah25: I like option `a` since it is simple and easy to understand. Unless @rawlinp is really hung on option `b`, I will not be changing the logic.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1634,6 +1787,36 @@ func Update(w http.ResponseWriter, r *http.Request) {
...
+
+	iSPQueryParts := make([]string, 0, len(pName))
+	iSPQueryValues := make([]interface{}, 0, len(pName))
+	for i, name := range pName {
+		valStart := i * 3
+		iSPQueryParts = append(iSPQueryParts, fmt.Sprintf("($%d, $%d, $%d)", valStart+1, valStart+2, valStart+3))
+		iSPQueryValues = append(iSPQueryValues, *id, name, i)
+	}
+
+	insertSPQuery += strings.Join(iSPQueryParts, ",")
```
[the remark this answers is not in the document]
rimashah25: We are doing the same for [interface](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/server/servers.go#L1168). Do we need to change that too? If not, I am going to stick with the logic I have currently.

lib/go-tc/servers.go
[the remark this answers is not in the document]
rawlinp: I leave the decision up to you, Rima -- I generally prefer option B, but I'm not hung up on it. Since the plan is to eventually cleanly remove API v3, the duplication is temporary (or, _should_ be temporary at least).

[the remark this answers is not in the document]
rimashah25: Ok, Thanks @rawlinp. I will leave it as is. And I updated my earlier comment. So, unless it is the right thing to do, I will not be changing the logic.

traffic_ops/testing/api/v4/tc-fixtures.json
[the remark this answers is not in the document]
rimashah25: Changed (dropped trigger) based on Brennan's and mine offline discussion.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1979,3 +1979,111 @@ WHERE server.id = $2;`
 
 	return nil
 }
+
+// GetCommonServerPropertiesFromV4 converts CommonServerPropertiesV40 to CommonServerProperties struct.
+func GetCommonServerPropertiesFromV4(s tc.ServerV40, tx *sql.Tx) (tc.CommonServerProperties, error) {
+	var id int
+	var desc string
+	rows, err := tx.Query("SELECT id, description from profile WHERE name=$1", (s.ProfileNames)[0])
```
zrhoffman, line 1987: This line panics (in API versions 2.0 and 3.0) if `s.ProfileNames` is nil/empty. In theory, this should not happen if the migration ran successfully, but since the DB can sometimes get in an inconsistent state if an operator tries to restore it, `dbhelpers.GetCommonServerPropertiesFromV4()` should make sure ProfileNames has a nonzero length before trying to get the first element.

I noticed this when testing the PR in the CDN in a Box for Developers (which seeds data straight from [`seed.psql`](https://github.com/apache/trafficcontrol/blob/e402b074a5/dev/traffic_ops/seed.psql), which has no `server_profile` table yet.

traffic_ops/app/db/migrations/2022032914235900_add_server_profile_table.up.sql
```diff
@@ -0,0 +1,38 @@
+/*
```
zrhoffman, line 1: `2022040623082100_server_config_update.down.sql` and `2022040623082100_server_config_update.up.sql` were added in #6730. so now the timestamp in the filename of the layered profile up/down migrations should be increased from `2022032914235900` to some value greater than `2022040623082100` so it is the last migration in the list when #6544 gets merged.

zrhoffman (changes requested):
The `server_profile` table should be populated in the Dev CiaB's [`seed.psql`](https://github.com/apache/trafficcontrol/blob/e402b074a5/dev/traffic_ops/seed.psql), especially since the TRUTH GHA depends on the Dev CiaB to be viable for Traffic Router testing.

An `INSERT` for the `server_profile` table could go after the `INSERT` for the `server` table:
https://github.com/apache/trafficcontrol/blob/e402b074a5b8e3af79b55d304993f6e34665a7ac/dev/traffic_ops/seed.psql#L199

It could probably just be a copy of the `INSERT` query from the `up` migration: https://github.com/apache/trafficcontrol/blob/e402b074a5b8e3af79b55d304993f6e34665a7ac/traffic_ops/app/db/migrations/2022032914235900_add_server_profile_table.up.sql#L27-L30

lib/go-atscfg/parentdotconfig.go
```diff
@@ -1694,9 +1694,6 @@ func getOriginServersAndProfileCaches(
 		} else if cgSv.TypeID == nil {
 			warnings = append(warnings, "getting origin servers: got server with nil TypeID, skipping!")
 			continue
-		} else if cgSv.ProfileID == nil {
-			warnings = append(warnings, "getting origin servers: got server with nil ProfileID, skipping!")
-			continue
```
ocket8888, line 1699: Shouldn't this be replaced with a check for `len(cgSv.ProfileNames) == 0`?

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -1979,3 +1979,114 @@ WHERE server.id = $2;`
...
+	if err != nil {
+		return fmt.Errorf("failed to insert/read into/from server_profile table, %w", err)
+	}
+	return nil
+}
+
+// UpdateServerProfileTableForV2V3 updates CommonServerPropertiesV40 struct and server_profile table via Update (server) function for API v2/v3.
+func UpdateServerProfileTableForV2V3(id *int, newProfile *string, origProfile string, tx *sql.Tx) ([]string, error) {
+	var profileName []string
+	query := `UPDATE server_profile SET profile_name=$1 WHERE server=$2 AND profile_name=$3`
```
ocket8888, line 2080: > Yes, the current array will go away with PUT APIv3 since it(profile) should be of length 1.

okay, so then shouldn't this be deleting all the other profiles for that server and putting in just one row for `priority = 0`? The way this currently works it looks like if I have a server with Profiles `["a", "b"]` and do a PUT in APIv3 to give it Profile `c`, then the resulting array if I query again would be `["c", "b"]`, but it sounds like you're saying it should be just `["c"]`

rimashah25: Ok, I should rephrase my earlier comment. 
So, if you make changes (PUT) via API v3/v2 to an existing server created via POST API v3/v2, then it will update the profile with priority `0` since v3/v2 allows 1 profile to a server. 
But if you create server via API v4 (multiple profile) and change it via API v2/v3, then the only change one shall see is to the profile with priority `0` but other profile will remain intact.

lib/go-atscfg/parentdotconfig.go
rimashah25: yup

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
ocket8888: okay, got it.

lib/go-tc/servers.go
ocket8888, line 703: This structure is superfluous and adds complexity we don't need

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -409,7 +442,53 @@ func validateCommon(s *tc.CommonServerProperties, tx *sql.Tx) []error {
...
+		errs = append(errs, fmt.Errorf("no profiles exists"))
+	}
+
+	var cdnID int
+	if err := tx.QueryRow("SELECT cdn from profile WHERE name=$1", s.ProfileNames[0]).Scan(&cdnID); err != nil {
+		log.Errorf("could not execute select cdnID from profile: %s\n", err)
+		if err == sql.ErrNoRows {
+			errs = append(errs, fmt.Errorf("no such profileName: '%v'", s.ProfileNames[0]))
+		} else {
+			errs = append(errs, tc.DBError)
```
ocket8888, line 483: From the GoDoc for `tc.DBError`:
> Deprecated: Since internal errors are not returned to users, there's no reason not to include more detail in an error message than this.

When something like this happens, the right response code is a `500 Internal Server Error`, but currently it would be a `400 Bad Request` - which means that the solution isn't to just append a different error. The only way I can think of to fix it is to change the return signature from `[]error` to `([]error, error)` so it can return a server-side error. It looks like that wouldn't be too big of a refactor, just add some `, nil`s and change the handling statement a little bit.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -409,7 +442,53 @@ func validateCommon(s *tc.CommonServerProperties, tx *sql.Tx) []error {
...
+	}
+
+	if len(s.ProfileNames) == 0 {
+		errs = append(errs, fmt.Errorf("no profiles exists"))
+	}
+
+	var cdnID int
+	if err := tx.QueryRow("SELECT cdn from profile WHERE name=$1", s.ProfileNames[0]).Scan(&cdnID); err != nil {
+		log.Errorf("could not execute select cdnID from profile: %s\n", err)
+		if err == sql.ErrNoRows {
```
ocket8888, line 480: error comparisons should use `errors.Is` rather than `==`

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -409,7 +442,53 @@ func validateCommon(s *tc.CommonServerProperties, tx *sql.Tx) []error {
...
+			errs = append(errs, fmt.Errorf("no such profileName: '%v'", s.ProfileNames[0]))
```
ocket8888, line 481: `%v` should be reserved for printing data structures and/or when the type of the thing being formatted isn't known until runtime. Strings should format with `%s`.

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -743,9 +822,14 @@ func Read(w http.ResponseWriter, r *http.Request) {
 	if version.Major >= 3 {
 		v3Servers := make([]tc.ServerV30, 0)
 		for _, server := range servers {
-			v3Server, err := server.ToServerV3FromV4()
+			csp, err := dbhelpers.GetCommonServerPropertiesFromV4(server, tx)
 			if err != nil {
-				api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("failed to convert servers to V3 format: %v", err))
+				api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("failed to get common server properties from V4 server struct: %v", err))
```
ocket8888, line 827: making an error from an error should wrap with `%w`

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -756,7 +840,12 @@ func Read(w http.ResponseWriter, r *http.Request) {
 
 	legacyServers := make([]tc.ServerNullableV2, 0, len(servers))
 	for _, server := range servers {
-		legacyServer, err := server.ToServerV2FromV4()
+		csp, err := dbhelpers.GetCommonServerPropertiesFromV4(server, tx)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("failed to get common server properties from V4 server struct: %v", err))
```
ocket8888, line 845: same as above RE: error wrapping

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1490,7 +1709,16 @@ func Update(w http.ResponseWriter, r *http.Request) {
 		}
 		api.WriteRespAlertObj(w, r, tc.SuccessLevel, "Server updated", srvr)
 	} else {
-		v2Server, err := srvr.ToServerV2FromV4()
+		csp, err := dbhelpers.GetCommonServerPropertiesFromV4(server, tx)
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusInternalServerError, nil, fmt.Errorf("failed to get common server properties from V4 server struct: %v", err))
```
ocket8888, line 1714: same as above RE: error wrapping

traffic_ops/traffic_ops_golang/server/servers.go
```diff
@@ -1490,7 +1709,16 @@ func Update(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		if err != nil {
+			api.HandleErr(w, r, tx, http.StatusBadRequest, nil, fmt.Errorf("failed to update server_profile: %v", err))
```
ocket8888, line 1718: same as above RE: error wrapping

ocket8888 (changes requested):
Functionally everything looks fine, everything from TP to T3C seems to be happily working with the new setup.

traffic_ops/traffic_ops_golang/server/servers.go
rimashah25: Why isn't it right [here](https://github.com/apache/trafficcontrol/blob/bafb05ef3088cf934bf135a118506956f11847af/traffic_ops/traffic_ops_golang/server/servers.go#L433), then?

ocket8888: Oh there are tons of places where it's not using `errors.Is`, and tons of places where error-wrapping isn't being done etc. But shallow comparisons like this will fail to work right if anything underlying the call that produces the error wraps it. [The shallow comparison only checks for literal equivalence while `errors.Is` unwraps errors to find out what the original error actually was](https://go.dev/play/p/3SM1ufNb6p-).

rimashah25: Then we should have tech debt and refactor all..

ocket8888 (approved):

ocket8888:
I'm assuming based on the fact that @zrhoffman's last comment was about a week ago and all those comments appear to have been addressed that he's cool with me merging this.

Issue #6551, opened 2022-02-09
Type of cache group in topology can be updated 


ericholguin:
## This Bug Report affects these Traffic Control components:

- Traffic Ops
- Traffic Portal

## Current behavior:

When updating a cache group that belongs to a topology a user should not be able to update the Type of that cache group, resulting in a Bad Request 400 from Traffic Ops as well as an alert in the response or alert in Traffic Portal. 

**However**, if you update the name field of the cache group, as well as the type this will work and result in an OK 200, with no alert on either TO or TP. This should not be allowed.

Traffic Ops has a separate issue where you can change the type, regardless if the name field is changed, as long as the id field is not present in your request, it will succeed. 

## Expected behavior:

Regardless of other fields being updated, type should not be updated and the request should return a Bad Request 400 with corresponding alert

## Steps to reproduce:

In TP/TO, create a cache group and a server that uses the cache group. Create a topology with the cache group associated to it. Finally try updating the cache group type, should fail. Try updating the type as well as the name field, should succeed.

PR #6552, opened 2022-02-09
Validate type in topology using current cache group name

ericholguin:
Closes: #6551 

Validates that cache group type is not being changed if it is used in a topology, using previous object name, instead of the requested name.

<hr/>

- Traffic Ops

Try updating the type of a cache group which belongs to a topology, should fail
Try updating the type and name of a cache group which belongs to a topology, should fail

- 5.1.x
- 6.0.x
- 6.1.x

## PR submission checklist

### DIFF
--- a/traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
+++ b/traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
@@ -189,13 +189,29 @@ func (cg *TOCacheGroup) ValidateTypeInTopology() error {
 	}
 
 	// language=SQL
+	const previousNameQuery = `
+	SELECT name
+	FROM cachegroup c
+	WHERE c.id = $1
+	`
+	var previousName string
+	err = cg.ReqInfo.Tx.QueryRow(previousNameQuery, *cg.ID).Scan(&previousName)
+	if err == sql.ErrNoRows {
+		return nil
+	}
+	if err != nil {
+		log.Errorf("%s: getting the previous name of cachegroup %s: %s", userErr.Error(), *cg.Name, err.Error())
+		return userErr
+	}
+
+	// language=SQL
 	const usedInTopologyQuery = `
 	SELECT EXISTS (SELECT
 	FROM topology_cachegroup tc
 	WHERE tc.cachegroup = $1)
 	`
 	var usedInTopology bool
-	err = cg.ReqInfo.Tx.QueryRow(usedInTopologyQuery, *cg.Name).Scan(&usedInTopology)
+	err = cg.ReqInfo.Tx.QueryRow(usedInTopologyQuery, previousName).Scan(&usedInTopology)
 	if err != nil {
 		log.Errorf("%s: querying topology_cachegroup by cachegroup name: %s", userErr.Error(), err.Error())
 		return userErr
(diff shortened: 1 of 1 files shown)


ericholguin:
This change will only fix the issue for Traffic Portal. The code performs a check against the requested body, so if in the request body there is a change to the name, the validation will perform a query against that new name and will not find that newly named cache group in any topology, hence allowing it to update and succeed. Changing the code to use the previous or current name, fixes this, as it will find the topology relationship.

However, in Traffic Ops, the type can be changed regardless of the name or any other field changes, if there is no id field in the request body. This is because the validation uses the cache group id in the query to get previous type and now this newly added query for previous name. The way validation is structured, is that it will only use the requested body, and in the case of this endpoint id is in the url path. If the user omits the id, the validation will be querying using an empty id value resulting in no real validation being made.

Not sure what the fix is, without changing the Validate functionality. Or unless id is required in the request body when updating a cache group.

zrhoffman (changes requested):
> [](https://github.com/ericholguin)@[ericholguin](https://github.com/ericholguin)
Contributor Author
ericholguin commented [2 days ago](https://github.com/apache/trafficcontrol/pull/6552#issuecomment-1034361900) •

This change will only fix the issue for Traffic Portal. The code performs a check against the requested body, so if in the request body there is a change to the name, the validation will perform a query against that new name and will not find that newly named cache group in any topology, hence allowing it to update and succeed. Changing the code to use the previous or current name, fixes this, as it will find the topology relationship.

However, in Traffic Ops, the type can be changed regardless of the name or any other field changes, if there is no id field in the request body. This is because the validation uses the cache group id in the query to get previous type and now this newly added query for previous name. The way validation is structured, is that it will only use the requested body, and in the case of this endpoint id is in the url path. If the user omits the id, the validation will be querying using an empty id value resulting in no real validation being made.

Sounds like `id` should be required in PUT requests

traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
```diff
@@ -188,14 +188,30 @@ func (cg *TOCacheGroup) ValidateTypeInTopology() error {
...
+	FROM cachegroup c
+	WHERE c.id = $1
+	`
+	var previousName string
+	err = cg.ReqInfo.Tx.QueryRow(previousNameQuery, *cg.ID).Scan(&previousName)
+	if err == sql.ErrNoRows {
+		return nil
+	}
+	if err != nil {
+		log.Errorf("%s: getting the previous name of cachegroup %s: %s", userErr.Error(), *cg.Name, err.Error())
```
zrhoffman, line 203: This line panics  if `userErr` is nil. Changing `userErr.Error()` to `userErr` would fix that

zrhoffman (changes requested):

traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go
ericholguin: Doesn't this line prevent that?
https://github.com/apache/trafficcontrol/blob/c0b747c18799952a4db7566c5a0f5dd9a5989c95/traffic_ops/traffic_ops_golang/cachegroup/cachegroups.go#L164

zrhoffman: oh, true

zrhoffman:
> I'm not sure if adding that requirement would be a breaking change to the API versions or other implications since the Update Handler simply uses the url path to set the id key in the object

Good point, it would be.

PR #6553, opened 2022-02-10
Fixed uncaught exception in Fetcher::getConnection()

brosler:
<hr/>

- Documentation
- Traffic Control Client 
- Traffic Monitor
- Traffic Ops
- Traffic Portal
- Traffic Router
- Traffic Stats
- Grove
- CDN in a Box
- Automation 
- unknown

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java
+++ b/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java
@@ -134,7 +134,16 @@ public class Fetcher {
 			connection.connect();
 
 		} catch (Exception e) {
-			LOGGER.error("Failed Http Request to " + http.getURL() + " Status " + http.getResponseCode());
+			String responseCode = "(none)";
+			try {
+				responseCode = String.valueOf(http.getResponseCode());
+
+			} catch (IOException e2) {
+				// Don't care
+				LOGGER.info("Exception during call attempt to http.getResponseCode()");
+			}
+
+			LOGGER.error("Failed Http Request to " + http.getURL() + " Status " + responseCode);
 			http.disconnect();
 		}
 


traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java
```diff
@@ -134,7 +134,16 @@ public boolean verify(final String arg0, final SSLSession arg1) {
...
 
 		} catch (Exception e) {
-			LOGGER.error("Failed Http Request to " + http.getURL() + " Status " + http.getResponseCode());
+			String responseCode = "(none)";
+			try {
+				responseCode = String.valueOf(http.getResponseCode());
+
+			} catch (IOException e2) {
+				// Don't care
+				LOGGER.info("Exception during call attempt to http.getResponseCode()");
```
rawlinp, line 143: This probably shouldn't be `info`, but if we don't care, do we really need to log anything at all? Also, there is a chance `http` could be null, so perhaps we need some `!= null` checks in this catch block as well?

brosler: yes - hadn't thought of that as a possibility. When http is assigned to connection it's being cast down to HttpURLConnection which can throw a ClassCastException which would land in this block.

I'll update.

I added the log message because PMD doesn't like empty exception blocks - it 's being pedantic but I figured I'd be thorough.

brosler: actually - I was chatting with Eric and decided to change course. I've got a small pile of changes queue'd up which deal with unit-test failures and other issues I've run into. I'm going to close this PR and spin up a larger one that includes the other changes as well. I'll incorporate your comment into the new changeset.

PR #6545, opened 2022-02-08
Add struct field names to Traffic Ops Routes

zrhoffman:
#5667 added struct field names to structs *inside* Routes but not to Routes themselves. #6545 adds struct field names to each Route.
<hr/>

## Which Traffic Control components are affected by this PR?

- Traffic Ops

## What is the best way to verify this PR?

Ensure all routes have field names now:
```shell
sed '/api\.Version{/!d; /{Version:/d' traffic_ops/traffic_ops_golang/routing/routes.go
```
(should return no lines)

## PR submission checklist
- [x] No tests necessary 
- [x] Behavior is unchanged, no documentation necessary 
- [x] No CHANGELOG.md entry necessary 
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://apache.org/security) for details)

PR #6554, opened 2022-02-10
Refactor cachegroup tests

ericholguin:
This PR refactors [cachegroup_test.go](https://github.com/apache/trafficcontrol/blob/b9fa07687597aee68db25a212dac0203962ec862/traffic_ops/testing/api/v4/cachegroups_test.go) in order to make tests more granular, providing transparency on what is being tested and reducing the need to dig through test code to confirm what was tested. In addition this new test structure should also make it easier for new tests to be added as well as reduce redundant code.

This refactor can be used as a POC of what the other _test files can be refactored to.

- Adds data driven testing 
  - Descriptive test names
  - Sub tests for organization
- Use GoLangs test runner 
- Use assert functionality
  - Streamlines code
  - Reduces nested conditionals 
- Adds expectation functions
  - test specific expectations in a clear concise way
  - Reusable expectations

The difference in output:

Before:
```go
--- PASS: TestCacheGroups (4.86s)
```

After:
```go
--- PASS: TestCacheGroups (3.84s)
    --- PASS: TestCacheGroups/GET (0.06s)
        --- PASS: TestCacheGroups/GET/NOT_MODIFIED_when_NO_CHANGES_made (0.00s)
        --- PASS: TestCacheGroups/GET/NOT_MODIFIED_when_VALID_NAME_parameter_when_NO_CHANGES_made (0.00s)
        --- PASS: TestCacheGroups/GET/OK_when_VALID_request (0.00s)
        --- PASS: TestCacheGroups/GET/OK_when_VALID_TOPOLOGY_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/FIRST_RESULT_when_LIMIT=1 (0.00s)
        --- PASS: TestCacheGroups/GET/BAD_REQUEST_when_INVALID_OFFSET_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/BAD_REQUEST_when_INVALID_PAGE_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/OK_when_VALID_NAME_parameter_AND_Lat/Long_are_0 (0.00s)
        --- PASS: TestCacheGroups/GET/NOT_MODIFIED_when_VALID_SHORTNAME_parameter_when_NO_CHANGES_made (0.00s)
        --- PASS: TestCacheGroups/GET/EMPTY_RESPONSE_when_INVALID_TYPE_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/OK_when_VALID_NAME_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/SECOND_RESULT_when_LIMIT=1_OFFSET=1 (0.01s)
        --- PASS: TestCacheGroups/GET/OK_when_VALID_SHORTNAME_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/OK_when_VALID_TYPE_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/EMPTY_RESPONSE_when_INVALID_ID_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/SECOND_RESULT_when_LIMIT=1_PAGE=2 (0.01s)
        --- PASS: TestCacheGroups/GET/BAD_REQUEST_when_INVALID_LIMIT_parameter (0.00s)
        --- PASS: TestCacheGroups/GET/UNAUTHORIZED_when_NOT_LOGGED_IN (0.01s)
    --- PASS: TestCacheGroups/POST (0.00s)
        --- PASS: TestCacheGroups/POST/UNAUTHORIZED_when_NOT_LOGGED_IN (0.00s)
    --- PASS: TestCacheGroups/PUT (0.05s)
        --- PASS: TestCacheGroups/PUT/BAD_REQUEST_when_updating_TYPE_of_CG_in_TOPOLOGY (0.01s)
        --- PASS: TestCacheGroups/PUT/PRECONDITION_FAILED_when_updating_with_IMS_&_IUS_Headers (0.01s)
        --- PASS: TestCacheGroups/PUT/PRECONDITION_FAILED_when_updating_with_IFMATCH_ETAG_Header (0.01s)
        --- PASS: TestCacheGroups/PUT/UNAUTHORIZED_when_NOT_LOGGED_IN (0.00s)
        --- PASS: TestCacheGroups/PUT/OK_when_VALID_request (0.01s)
        --- PASS: TestCacheGroups/PUT/OK_when_updating_CG_with_null_Lat/Long (0.01s)
    --- PASS: TestCacheGroups/DELETE (0.01s)
        --- PASS: TestCacheGroups/DELETE/NOT_FOUND_when_INVALID_ID_parameter (0.00s)
        --- PASS: TestCacheGroups/DELETE/UNAUTHORIZED_when_NOT_LOGGED_IN (0.00s)
    --- PASS: TestCacheGroups/GET_AFTER_CHANGES (0.00s)
        --- PASS: TestCacheGroups/GET_AFTER_CHANGES/OK_when_CHANGES_made (0.00s)
    --- PASS: TestCacheGroups/CDNLOCK (0.27s)
        --- PASS: TestCacheGroups/CDNLOCK/FORBIDDEN_when_updating_CG_when_CDN_LOCK_EXISTS (0.27s)
```

This new structure also allows a user to test specific methods, instead of running all the tests against all methods:
```go
go test -v -run TestCacheGroups/GET  ./v4 
go test -v -run TestCacheGroups/POST ./v4
go test -v -run TestCacheGroups/PUT ./v4
go test -v -run TestCacheGroups/DELETE ./v4
```
Or even run a single test case:
```go
go test -v -run TestCacheGroups/PUT/BAD_REQUEST_when_updating_type_of_cache_group_in_topology  ./v4
```

<hr/>

- Traffic Ops Testing

Run TO Integration tests

## PR submission checklist
- [x] This PR has tests 


traffic_ops/testing/api/utils/utils.go
```diff
@@ -73,3 +76,58 @@ func Compare(t *testing.T, expected []string, alertsStrs []string) {
 		}
 	}
 }
+
+type CkReqFunc func(*testing.T, toclientlib.ReqInf, interface{}, error)
```
ocket8888, line 80: GoDoc comment?

traffic_ops/testing/api/utils/utils.go
```diff
@@ -73,3 +76,58 @@ func Compare(t *testing.T, expected []string, alertsStrs []string) {
...
+
+func CkRequest(c ...CkReqFunc) []CkReqFunc {
```
ocket8888, line 82: GoDoc comment?

Also though this is just an identity function - what purpose does it serve?

traffic_ops/testing/api/utils/utils.go
```diff
@@ -73,3 +76,58 @@ func Compare(t *testing.T, expected []string, alertsStrs []string) {
...
+	return c
+}
+
+func NoError() CkReqFunc {
+	return func(t *testing.T, _ toclientlib.ReqInf, _ interface{}, err error) {
+		assert.NoError(t, err, "Expected no error, but got %v", err)
+	}
+}
+
+func HasError() CkReqFunc {
```
ocket8888, line 92: GoDoc comment?

traffic_ops/testing/api/utils/utils.go
```diff
@@ -73,3 +76,58 @@ func Compare(t *testing.T, expected []string, alertsStrs []string) {
...
+			assert.Error(t, err, "Expected error - but got: %+v %+v", resp.Response, resp.Alerts)
+		case tc.CacheGroupsNullableResponse:
+			assert.Error(t, err, "Expected error - but got: %+v %+v", resp.Response, resp.Alerts)
+		case tc.Alerts:
+			assert.Error(t, err, "Expected error - but got: %+v", resp.Alerts)
+		default:
+			assert.Fail(t, "Failed to find response type")
+		}
+	}
+}
```
ocket8888, line 105: If this switch has to encompass all types returned by the API it's going to get extremely long and annoying to maintain. Instead, I'd suggest just taking in the Response and the Alerts as separate arguments.

Ideally, though, the Alerts should just be sufficient to explain why a request failed, and in that case the Response is always supposed to be empty anyway (as far as I can remember right now). If they don't, then it's either a server-side error or a problem that needs to be fixed in the API, IMO. I think seeing response payloads might just make the output difficult to read without providing any real value.

traffic_ops/testing/api/utils/utils.go
```diff
@@ -73,3 +76,58 @@ func Compare(t *testing.T, expected []string, alertsStrs []string) {
...
+
+func HasStatus(expectedStatus int) CkReqFunc {
```
ocket8888, line 107: GoDoc comment?

traffic_ops/testing/api/utils/utils.go
```diff
@@ -73,3 +76,58 @@ func Compare(t *testing.T, expected []string, alertsStrs []string) {
...
+func ResponseHasLength(expected int) CkReqFunc {
+	return func(t *testing.T, _ toclientlib.ReqInf, resp interface{}, _ error) {
+		switch resp := resp.(type) {
+		case tc.CacheGroupsNullableResponse:
+			assert.Equal(t, expected, len(resp.Response), "Expected response object length %d, but got %d", expected, len(resp.Response))
+		default:
+			assert.Fail(t, "Failed to find response type")
+		}
+	}
+}
```
ocket8888, line 122: Same as above RE: how massive this will get if it has to switch through everything.

If this were to take in a `[]interface{}` that would make it much simpler, but it would cost the ability to to use it generically in structures and call signatures that call for a `CkReqFunc`. Which wouldn't be a terrible thing, in my opinion, but it would be a lot of work to change and it'd make the test cases a bit harder to write, not as nice as you have below that's for sure.

So instead if the Response and alerts are passed in separately in `CkReqFunc`s as I suggested above, you could narrow the type of the response to a slice by just doing
```go
func ResponseHasLength(expected int) CkReqFunc {
	return func(t *testing.T, _ toclientlib.ReqInf, resp interface{}, _ tc.Alerts, _ error) {
		if slice, ok := resp.([]interface{}); ok {
			assert.Equal(t, expected, len(slice), "Expected response object length %d, but got %d", expected, len(slice))
		} else {
			t.Failf("Expected response to be an array of length %d, but it wasn't an array at all: %+v", expected, resp)
		}
	}
}
```

ocket8888: Also same as above RE: error vs failure

traffic_ops/testing/api/utils/utils.go
```diff
@@ -73,3 +76,58 @@ func Compare(t *testing.T, expected []string, alertsStrs []string) {
...
+
+func ResponseLengthGreaterOrEqual(expected int) CkReqFunc {
```
ocket8888, line 124: GoDoc comment?

traffic_ops/testing/api/utils/utils.go
```diff
@@ -73,3 +76,58 @@ func Compare(t *testing.T, expected []string, alertsStrs []string) {
...
+	return func(t *testing.T, _ toclientlib.ReqInf, resp interface{}, _ error) {
+		switch resp := resp.(type) {
+		case tc.CacheGroupsNullableResponse:
+			assert.GreaterOrEqual(t, len(resp.Response), expected, "Expected response object length %d, but got %d", expected, len(resp.Response))
+		default:
+			assert.Fail(t, "Failed to find response type")
+		}
```
ocket8888, line 131: Same as above RE: the switch getting huge, also these are so similar the same suggestion(s) apply(/ies).

Also same as above RE: error vs failure

traffic_ops/testing/api/v4/cachegroups_test.go
```diff
@@ -16,10 +16,13 @@ package v4
...
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/utils"
+	"github.com/apache/trafficcontrol/traffic_ops/toclientlib"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"net/http"
 	"net/url"
-	"reflect"
 	"strconv"
 	"testing"
 	"time"
```
ocket8888, line 28: same as above RE: import order

traffic_ops/testing/api/v4/cachegroups_test.go
```diff
@@ -32,99 +35,393 @@ import (
...
-		GetTestCacheGroupsByType(t)
-		DeleteTestCacheGroupsByInvalidId(t)
-		UpdateCachegroupWithLocks(t)
+
+		methodTests := map[string]map[string]struct {
+			endpointId     func() int
+			clientSession  *client.Session
+			requestParams  map[string]string
+			requestBody    map[string]interface{}
+			requestHeaders http.Header
```
ocket8888, line 44: Can we collapse the `requestParams` and `requestHeaders` into the `client.RequestOptions` structure? Looks like you're just rebuilding that below.

traffic_ops/testing/api/v4/cachegroups_test.go
```diff
@@ -32,99 +35,393 @@ import (
...
+			expectations   []utils.CkReqFunc
+		}{
+			"GET": {
+				"OK when VALID NAME parameter AND Lat/Long are 0": {
+					nil, TOSession, map[string]string{"name": "nullLatLongCG"}, nil, nil,
```
ocket8888, line 49: This code would fail `go vet -composites` because it uses unkeyed fields. The rationale there being that if you add something or rearrange the structure of the type, the code will all fail to compile even if it still has exactly the information needed to work properly - because the fields will be improperly associated. We don't enforce that (yet, possibly, see #6545). That would also make it easier to read these without needing to scroll up to the type definition. Plus, with keyed fields we can rely on the zero values of unspecified properties, which allows us to omit  all these `nil`s I see.

traffic_ops/testing/api/v4/cachegroups_test.go
```diff
@@ -32,99 +35,393 @@ import (
...
+			"POST": {
+				"UNAUTHORIZED when NOT LOGGED IN": {
+					nil, NoAuthTOSession, nil, nil, nil,
+					utils.CkRequest(utils.HasError(), utils.HasStatus(http.StatusUnauthorized)),
+				},
+			},
+			"PUT": {
+				"OK when VALID request": {
+					GetCacheGroupId(t, "cachegroup1"), TOSession, nil,
+					map[string]interface{}{
```
ocket8888, line 137: Since these tests are specifically for the client, I think we should be using the structures used by the client - in this case `tc.CacheGroupNullable`

traffic_ops/testing/api/v4/cachegroups_test.go
```diff
@@ -32,99 +35,393 @@ import (
...
+							testCase.requestBody["id"] = testCase.endpointId()
+						}
+						if typeId, ok := testCase.requestBody["typeId"]; ok {
+							if typeId == -1 {
+								if typeName, ok := testCase.requestBody["typeName"]; ok {
+									testCase.requestBody["typeId"] = GetTypeId(t, typeName.(string))
+								}
+							}
+						}
+						dat, _ := json.Marshal(testCase.requestBody)
```
ocket8888, line 269: I think ignoring this error could cause debugging confusion when a test uses bad data

ocket8888 (commented):
Pulling in this testify library adds a new dependency which consists of 38 new files and somewhere in the neighborhood of 19,500 lines to the repository - I believe that the entire usage of that library can be implemented in the following 52 lines (which, to be fair, doesn't include GoDoc comments, which I'd appreciate being added)
```go
import (
	"reflect"
	"testing"
)

func Equal(t *testing.T, a, b interface{}, msg string) bool {
	if a != b {
		t.Error(msg)
		return false
	}
	return true
}

func RequireEqual(t *testing.T, a, b interface{}, msg string) {
	if !Equal(t, a, b, msg) {
		t.FailNow()
	}
}

func Error(t *testing.T, err error, msg string) bool {
	if err != nil {
		t.Error(msg)
		return false
	}
	return true
}

func Exactly(t *testing.T, a, b interface{}, msg string) bool {
	if !reflect.DeepEqual(a, b) {
		t.Error(msg)
		return false
	}
	return true
}

func Fail(t *testing.T, msg string) { t.Fail(msg) }

func GreaterOrEqual(t *testing.T, a, b interface{}, msg string) bool {
	if a > b {
		return true
	}
	return Equal(t, a, b, msg)
}

func RequireGreaterOrEqual(t *testing.T, a, b interface{}, msg string) {
	if !GreaterOrEqual(t, a, b, msg) {
		t.FailNow()
	}
}

func NoError(t *testing.T, err error, msg string) bool { return Equal(t, err, nil, msg) }

func RequireNoError(t *testing.T, err error, msg string) {
	if !NoError(t, err, msg) {
		t.FailNow()
	}
}

func NotNil(t *testing.T, a interface{}, msg string) bool {
	if a == nil {
		t.Error(msg)
		return false
	}
	return true
}

func RequireNotNil(t *testing.T, a interface{}, msg string) {
	if !NotNil(t, a, msg) {
		t.FailNow()
	}
}
```
(that's 55 lines, but I don't count the `import` open/close lines or the `testing` import line because those were already in there)

These functions are all extremely simple, why not do this instead of pulling in a dependency?

ocket8888:
This is only the APIv4 Cache Group tests - will these changes (or similar) be added to API versions 2 and 3

traffic_ops/testing/api/utils/utils.go
ericholguin: `assert.Error` and `assert.Fail` are not much different in implementation, at the core they call `t.Errorf`, which does not terminate the test execution like `FailNow`. Fail simply provides a way to fail without performing any checks, and in this case the checks are performed above. However, this will probably be changed since this library will probably get removed.

traffic_ops/testing/api/utils/utils.go
ericholguin: Type cast didn't work but reflection did

traffic_ops/testing/api/v4/cachegroups_test.go
ericholguin: Yeah, I've gone ahead and changed this per your recommendation, initially I had it this way but the lines were long and verbose so I separated them.

traffic_ops/testing/api/v4/cachegroups_test.go
ericholguin: The problem is with `tc.CacheGroupNullable` is you can't do inline pointer declarations.

ericholguin:
> Pulling in this testify library adds a new dependency which consists of 38 new files and somewhere in the neighborhood of 19,500 lines to the repository - I believe that the entire usage of that library can be implemented in the following 52 lines (which, to be fair, doesn't include GoDoc comments, which I'd appreciate being added)

> These functions are all extremely simple, why not do this instead of pulling in a dependency?

That's fair, I suppose we can expand on this and add functions as needed. For now I can't really think of anything special that testify might provide and I am okay with implementing our own. Do you think placing these in `utils.go` would be okay?

ericholguin:
> This is only the APIv4 Cache Group tests - will these changes (or similar) be added to API versions 2 and 3 as well?

These changes will be added to API v2 and v3, wanted to get feedback before doing so.

traffic_ops/testing/api/v4/cachegroups_test.go
zrhoffman: > We don't enforce that (yet, possibly, see #6545)

We do enforce it, but when 2 structs are on 1 line and one of them is keyed, `go vet` doesn't detect unkeyed fields in the second, which is a `go vet` limitation.

traffic_ops/testing/api/assert/assertions.go
```diff
@@ -0,0 +1,109 @@
...
+// It marks the test as failed and stops execution.
+func RequireGreaterOrEqual(t *testing.T, a, b int, msg string) {
+	t.Helper()
+	if !GreaterOrEqual(t, a, b, msg) {
+		t.FailNow()
+	}
+}
+
+// NoError asserts that a function returned no error (i.e. `nil`).
+func NoError(t *testing.T, err error, msg string) bool { t.Helper(); return Equal(t, err, nil, msg) }
```
zrhoffman, line 81: Should a small message about about expecting there not to be an error be prepended to the given message?

traffic_ops/testing/api/assert/assertions.go
```diff
@@ -0,0 +1,109 @@
...
+	if !NoError(t, err, msg) {
+		t.FailNow()
+	}
+}
+
+// NotNil asserts that the specified object is not nil.
+func NotNil(t *testing.T, a interface{}, msg string) bool {
+	t.Helper()
+	if a == nil {
+		t.Error(msg)
```
zrhoffman, line 96: Should a small message about expecting the operand not to be nil be prepended to the given message?

zrhoffman (changes requested):

traffic_ops/testing/api/utils/utils.go
ocket8888: that's strange, type casting worked for me in a playground

traffic_ops/testing/api/v4/cachegroups_test.go
ocket8888: Yes, that's a limitation with which many of us were frustrated, and there are some utility functions in [the `github.com/apache/trafficcontrol/lib/go-util` package like `IntPtr`](https://pkg.go.dev/github.com/apache/trafficcontrol/lib/go-util#IntPtr) that help mitigate that.

traffic_ops/testing/api/assert/assertions.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"reflect"
+	"testing"
+)
+
+// Equal asserts that two objects are equal.
+func Equal(t *testing.T, a, b interface{}, msg string) bool {
+	t.Helper()
+	if a != b {
+		t.Error(msg)
```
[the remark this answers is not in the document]
ocket8888: To be a drop-in replacement for the testify library, no, because that wouldn't. Wouldn't bother me if it did though, or if this actually supported variadic arguments for string formatting like `testify` does. It could save some `fmt.Sprintf`-ing in the calls.

ocket8888:
> Do you think placing these in utils.go would be okay?

Yes, that seems like the best place to put them.

traffic_ops/testing/api/v4/cachegroups_test.go
```diff
@@ -16,77 +16,367 @@ package v4
...
+						"typeName":  "MID_LOC",
+						"typeId":    -1,
+					},
+					expectations: utils.CkRequest(utils.HasError(), utils.HasStatus(http.StatusBadRequest)),
+				},
+				"PRECONDITION FAILED when updating with IMS & IUS Headers": {
+					endpointId: GetCacheGroupId(t, "parentCachegroup"), clientSession: TOSession,
+					requestOpts: client.RequestOptions{Header: http.Header{
+						rfc.IfModifiedSince:   {time.Now().UTC().Add(-5 * time.Second).Format(time.RFC1123)},
+						rfc.IfUnmodifiedSince: {time.Now().UTC().Add(-5 * time.Second).Format(time.RFC1123)},
```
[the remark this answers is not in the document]
ericholguin: I believe so, I matched what existed previously: https://github.com/apache/trafficcontrol/blob/1398c02a80dfd4090881f8f90961c11ff57feeaa/traffic_ops/testing/api/v4/cachegroups_test.go#L44-L49

traffic_ops/testing/api/assert/assertions.go
```diff
@@ -0,0 +1,144 @@
...
+		return false
+	}
+	return true
+}
+
+// Exactly asserts that two objects are equal in value and type.
+func Exactly(t *testing.T, a, b interface{}, msgAndArgs ...interface{}) bool {
+	t.Helper()
+	if !reflect.DeepEqual(a, b) {
+		msg := failureOutput(fmt.Sprintf("Error: Not equal: \n expected: %v\n actual  : %v", a, b), msgAndArgs...)
```
zrhoffman, line 79: Not: Maybe this should say *Not exactly equal* to distinguish it from the *Not equal* message?

zrhoffman (approved):
Looks good to me! It's a lot clearer now what passed or failed.

traffic_ops/testing/api/assert/assertions.go
```diff
@@ -16,14 +16,36 @@ package assert
...
+		if msgAsStr, ok := msg.(string); ok {
+			message = msgAsStr
+		}
+		message = fmt.Sprintf("%+v", msg)
+	}
+	if len(msgAndArgs) > 1 {
+		message = fmt.Sprintf(msgAndArgs[0].(string), msgAndArgs[1:]...)
+	}
+	if len(message) > 0 {
+		output = "\n" + output + "\nMessages: " + message + "\n"
```
ocket8888, line 39: I'm against putting newline characters in most kinds of output, including test failures. Line-based output is extremely common and easy to parse, so most tools work well with it, instead of with custom multi-line formats.

traffic_ops/testing/api/assert/assertions.go
```diff
@@ -0,0 +1,144 @@
...
+	if !NoError(t, err, msgAndArgs...) {
+		t.FailNow()
+	}
+}
+
+// NotNil asserts that the specified object is not nil.
+func NotNil(t *testing.T, a interface{}, msgAndArgs ...interface{}) bool {
+	t.Helper()
+	if a == nil {
+		msg := failureOutput("Error: Expected value not to be nil.", msgAndArgs...)
```
ocket8888, line 130: nit: you don't need to add `Error:`, `testing.T.Error` provides the context of being a test error already.

ocket8888 (commented):
Last thing

ocket8888 (approved):

PR #6559, opened 2022-02-11
Fix Security Issues in Traffic Portal Server

shamrickus:
This PR fixes `npm audit` vulnerabilities found in the Traffic Portal server. We currently use [`forever`](https://github.com/foreversd/forever) as a process manager for TP, but it is only community maintained and has security issues. As such, this replaces `forever` with a new process manager [`pm2`](https://pm2.keymetrics.io/).
<hr/>

- Traffic Portal
- CDN in a Box

Verify the TP Integration Tests GHA pass. Verify the TP rpm can be created as well as installed. Ensure CiaB properly starts the TP container still. Finally confirm `npm audit` returns 0 results.

 - master
 - 6.1.x
 - 5.1.x

## PR submission checklist
- [x] This PR has tests 


traffic_portal/build/etc/init.d/traffic_portal
```diff
@@ -43,86 +25,50 @@
...
+      --time \
+      --name "$NAME"
     RETVAL=$?
 }
- 
+
 stop() {
-    if [ -f $PIDFILE ]; then
+    echo `pm2 list` | grep -q "$NAME"
+    if [ "$?" -eq "0" ]; then
```
ocket8888, line 55: nit but <code>echo \`pm2 list\`</code> is equivalent to `pm2 list` and checking the exit code immediately can be more succinctly written as 
```bash
if pm2 list | grep -q "$NAME"; then
```
(add `>/dev/null` if the output to stdout is annoying, won't change the exit code)

shamrickus: `-q` on `grep` handles the output.

ocket8888 (approved):
TP still works

PR #6568, opened 2022-02-16
Various changes to restore integration tests in 'verify' target

brosler:
<hr/>
This PR includes changes for the following issues:

:: Race condition between certificate-file and CrConfig-file download and activation that can result in incorrect promotion of a CrConfig. The current logical flow has the two files arriving asynchronously from one another. When the ConfigHandler has ingested the new DeliveryServices it forwards the new information to the CertificatePublisher and then pends until the CP has OK'd the DS.  The certificates are downloaded periodically so the DS will be verified when the next certificate file download. Since the actions are asynchronous the verification process may start just before the ConfigHandler pushes the new set of DS and then pends - in this case the CP is verifying against the old/unupdated set of DS. When validation completes the CP will clear the pend-lock that clears the _new_ CrConfig to be accepted as a valid snapshot - the new DS were never validated. This was exposed during the integration test external/RouterTest.java in RouterTest::itRejectsCrConfigWithMissingCert. This (partial) fix was to introduce a flag in CertificatesPublisher which latches when the new DS has been received. If/When the CertificatesPublisher exits certificatesValidation it will not clear the pend-lock if the DS have been updated while validation was in progress - validation will be done at the next file download. There still exists a small window in which this failure can still occur but the vulnerability is greatly reduced - likely below a point in which it's likely to happen in the wild.
./core/src/main/java/org/apache/traffic_control/traffic_router/core/secure/CertificatesPublisher.java

:: Uncaught exception in Fetcher.java that results in a bogus exception being thrown upward. If/When the connection fails during Fetcher::getConnection() than the catch-block will be executed. During execution the LOGGER will attempt to call http;getResponseCode(), which will throw an IOException because the remote server cannot be reached. This second exception is then propagated upward, and is currently caught by the ultimate author of the original request (usually through Fetcher::getIfModifiedSince() or Fetcher::fetchIfModifiedSince()). The result is a semi-bogus exception being reported to the penultimate caller. A minor adjustment has been made to Fetcher.java to intercept this second exception, discard it, and re-throw the original exception. Only IOExceptions are allowed to be rethrown in order to avoid adding new/untested functionality to the logical blocks that depend on Fetcher.
./core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java

:: ConsoleAppender now requires a name in order to be constructed. A number of test-files depend on a log4j::ConsoleAppender, lacking a setName call construction aborts. This problem was hamstringing a number of the integration tests being run during the maven verify target run. ./core/src/test/java/org/apache/traffic_control/traffic_router/core/TestBase.java
./core/src/test/java/org/apache/traffic_control/traffic_router/core/TrafficRouterStart.java
./core/src/test/java/org/apache/traffic_control/traffic_router/core/external/ExternalTestSuite.java

:: During verify execution a number of pedantic stack traces are thrown when the Fake Server is started. The stack-track is thrown from TestBase itself when the platformMBeanServer (re)registers the DeliveryServiceCertificatesMBean. Added code to un-register this bean when the Fake Server is shut down.
./core/src/test/java/org/apache/traffic_control/traffic_router/core/TestBase.java

:: Changed the path for the czmap.json file. This file was located under ./core/src/test/resources/czmap.json alongside other files used during testing - but the path in the test file was ./core/src/test/db/czmap.json. It seems like this particular path was never updated when the file was moved.
./core/src/test/java/org/apache/traffic_control/traffic_router/core/dns/ZoneManagerTest.java

:: Monkeyed with HttpDataServer to generate ContentLength. Down in the bowels of the HttpURLConnection there's a check for Content-Length which is supposedly a "should" in the http spec. The HttpDataServer was leaving the x-fer length at 0, which was generating a chunked transfer. This was causing a little bit of churn in the logical flow during debugging. To clean this up the HttpDataServer now provides the Content-Length header during most transfers.
./core/src/test/java/org/apache/traffic_control/traffic_router/core/external/HttpDataServer.java

:: During debugging noted that the TrafficRouterTest.java was silently throwing some NPE's. Fixes have been added to more-completely populate the test setup to avoid these issues and provide better test coverage.
./core/src/test/java/org/apache/traffic_control/traffic_router/core/router/TrafficRouterTest.java

:: Added new tests to FetcherTest.java to verify coverage of exception cases
./core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java

:: During execution of the integration test RouterTest::itRejectsCrConfigWithMissingCert it was noted that execution would reliably hangup - further tests weren't executed. Additionally, the RouterTest was periodically failing. Debugging showed that something was happening when the dnsrecords were being refresher. It turns out that the dnschallengemapping.polling.url had not been changed in the CrConfig files used in integration tests, resulting in the dnsrecords fetch using an https address. Since the HttpDataServer is only setup for http service, these https connection attempts were failing after a timeout. Two things were happening. First, since the HttpDataServer only opened one-port/one-listener, the https attempt was tying up the test port, which was blocking the other fetches required for testing, throwing off test timing (what was happening in the RouterTest). Second, the connection failure was generating an uncaught exception in Fetcher::getConnection which was leaving the TCP connection in limbo, probably in a TIMEDWAIT state which would eventually give up - but long after the tests had failed. The earlier fix for the uncaught exception in Fetcher::getConnection fixes the second part, but since the dnsrecords aren't critical for the integration tests it was also helpful to just shove them out of the way. The CrConfig files have been updated to include adjustments for the dnsrecords url and fetch timeout to move them out of the way of the integration tests.
./core/src/test/resources/publish/CrConfig.json
./core/src/test/resources/publish/CrConfig2.json
./core/src/test/resources/publish/CrConfig3.json
./core/src/test/resources/publish/CrConfig4.json

- Traffic Router

Best way - run 'mvn verify'

This PR doesn't contain a specific bugfix as its bundles a number of changes related to smaller issues into one batch-commit.

## PR submission checklist
- [x] This PR has tests 


traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java
```diff
@@ -233,4 +239,23 @@ public void createStringBuilderFromResponse (final StringBuilder sb, final HttpU
...
+
+	private void connectionFailed(final HttpURLConnection http, final String url) {
+		String httpUrl = url;
+		String responseCode = "(none)";
+		try {
+			httpUrl = http.getURL().toString();
+			responseCode = String.valueOf(http.getResponseCode());
+		} catch (Exception e2) {
+			// Don't care
+			LOGGER.info("Exception during call attempt to retrieve url or responseCode from http");
```
rawlinp, line 251: Should this be `error` instead of `info`? Also, should we log the exception?

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java
```diff
@@ -233,4 +239,23 @@ public void createStringBuilderFromResponse (final StringBuilder sb, final HttpU
...
+		}
+
+		LOGGER.error("Failed Http Request to " + httpUrl + responseCode);
```
rawlinp, line 254: Can we change this to `LOGGER.error("Failed HTTP request to " + httpUrl + " status code: " + responseCode);`?

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/router/TrafficRouterTest.java
```diff
@@ -73,6 +73,7 @@ public void before() throws Exception {
         when(deliveryService.isCoverageZoneOnly()).thenReturn(false);
         when(deliveryService.getDispersion()).thenReturn(mock(Dispersion.class));
         when(deliveryService.isAcceptHttp()).thenReturn(true);
+		when(deliveryService.getId()).thenReturn("someDsName");
```
rawlinp, line 76: indentation is off

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/router/TrafficRouterTest.java
```diff
@@ -126,6 +127,9 @@ public void itCreatesHttpResults() throws Exception {
         HTTPRequest httpRequest = new HTTPRequest();
         httpRequest.setClientIP("192.168.10.11");
         httpRequest.setHostname("ccr.example.com");
+		Map<String, String> headers = new HashMap<String, String>();
+		headers.put("x-tc-steering-option", "itCreatesHttpResults");
+		httpRequest.setHeaders(headers);
```
rawlinp, line 132: indentation is off

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/router/TrafficRouterTest.java
```diff
@@ -163,6 +167,8 @@ public void itFiltersByIPAvailability() throws Exception {
 
         DeliveryService ds = mock(DeliveryService.class);
 
+		when(ds.getId()).thenReturn("itFiltersByIpAvailable");
```
rawlinp, line 170: indentation is off

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/router/TrafficRouterTest.java
```diff
@@ -280,6 +286,9 @@ public void itSetsResultToGeo() throws Exception {
         httpRequest.setClientIP("192.168.10.11");
         httpRequest.setHostname("ccr.example.com");
         httpRequest.setPath("/some/path");
+		Map<String, String> headers = new HashMap<String, String>();
+		headers.put("x-tc-steering-option", "itSetsResultToGeo");
+		httpRequest.setHeaders(headers);
```
rawlinp, line 291: indentation is off

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
```diff
@@ -62,4 +69,44 @@ public void itChecksIfDataHasChangedSinceLastFetch() throws Exception {
...
     }
-}
\ No newline at end of file
+
+    @Test
+    public void itChecksIfSocketTimeoutExceptionThrown() throws Exception {
+		// Test that an IOException (ex connection-failure) is caught, re-thrown, and
+		// that the embedded error is correctly encapsulated in the exception
+		//
+		final String mockedException = "Mocked Connection Failure";
```
rawlinp, line 78: indentation is off

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
```diff
@@ -62,4 +69,44 @@ public void itChecksIfDataHasChangedSinceLastFetch() throws Exception {
...
+        URL url = mock(URL.class);
+        HttpURLConnection httpURLConnection = mock(HttpURLConnection.class);
+
+        stub(method(URL.class, "openConnection")).toReturn(httpURLConnection);
+		doThrow(new SocketTimeoutException(mockedException)).when(httpURLConnection).connect();
```
rawlinp, line 83: indentation is off

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
```diff
@@ -62,4 +69,44 @@ public void itChecksIfDataHasChangedSinceLastFetch() throws Exception {
...
+        whenNew(URL.class).withArguments("http://www.example.com").thenReturn(url);
+
+        Fetcher fetcher = new Fetcher();
+		try {
+			fetcher.fetchIfModifiedSince("http://www.example.com", 123456L);
+			assertTrue(false);
+		} catch (SocketTimeoutException e) {
+			assertEquals(e.toString(), "java.net.SocketTimeoutException: " + mockedException);
+		}
```
rawlinp, line 92: indentation is off

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
```diff
@@ -62,4 +69,44 @@ public void itChecksIfDataHasChangedSinceLastFetch() throws Exception {
...
+		// is cast to type "HttpURLConnection". This will abort execution of
+		// Fetcher::getConnection() during the cast of "connection" to "http". This will
+		// result in the callstack unwinding normally with null return codes.
+		//
+        URLConnection urlConnection = mock(URLConnection.class);
+        stub(method(URL.class, "openConnection")).toReturn(urlConnection);
+
+        URL url = mock(URL.class);
+        Fetcher fetcher = new Fetcher();
+		assertEquals(fetcher.fetchIfModifiedSince("http://www.example.com", 123456L), null);
```
rawlinp, line 110: indentation is off

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
```diff
@@ -62,4 +69,44 @@ public void itChecksIfDataHasChangedSinceLastFetch() throws Exception {
...
+    @Test
+    public void itChecksIfOtherThrown() throws Exception {
+		// Test that other exceptions (ex CastClassException) is caught and
+		// squelched. This matches existing functionality.
+		//
+		// This test relies on the CastClassException which is thrown when "connection"
+		// is cast to type "HttpURLConnection". This will abort execution of
+		// Fetcher::getConnection() during the cast of "connection" to "http". This will
+		// result in the callstack unwinding normally with null return codes.
+		//
```
rawlinp, line 104: indentation is off

rawlinp (changes requested):
This looks good, thanks! I just have a minor logging formatting request and some indentation requests.

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java
brosler: We can definitely log the exception - I'll add that. I'll also add the exception to the main "Failed Http Request to "... error message.

Regarding this being an info vs error level. I'm of the opinion that generating an error message here (second/embedded catch block) would confuse diagnosis and debugging of the actual underlying issue that landed us in the main catch-block - which is likely to be related to a connection failure. If we generated an error while trying to generate the error message we'd just be adding noise to the problem space.

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java
brosler: Yeah - duh and whoops ... definitely

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/router/TrafficRouterTest.java
brosler: Yep - interesting that some files are tab'd and other's are spaced.

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/router/TrafficRouterTest.java
brosler: Yep

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
brosler: Yep

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
brosler: yep

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
brosler: yep

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
brosler: yep

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/util/FetcherTest.java
brosler: yep

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/util/Fetcher.java
rawlinp: Ok, maybe `debug` rather than `info`?

brosler: I like that ... makes sense

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/router/TrafficRouterTest.java
rawlinp: You have fine taste, then. Just don't tell @ocket8888

brosler: Should be good now. Check out the latest and ping me if you think of anything else.

traffic_router/core/src/test/java/org/apache/traffic_control/traffic_router/core/loc/AbstractServiceUpdaterTest.java
```diff
@@ -65,6 +69,9 @@ public void before() throws Exception {
 		databasesDirectory = mock(Path.class);
 		when(databasesDirectory.resolve((String) isNull())).thenReturn(databasePath);
 
+	    trafficRouterManager = mock(TrafficRouterManager.class);
```
rawlinp, line 72: this line mixes tabs and spaces

rawlinp (commented):
Just one inconsistent indentation; otherwise, I think this is good to go.

rawlinp (approved):

PR #6570, opened 2022-02-16
t3c uses mojolicious cookie for authentication. 

jpappa200:
This isn't related to an open issue. This PR will update t3c to use the mojolicious cookie for authentication and fall back to password auth if the cookie is expired. 

<hr/>

Build and test on admin_down cache. look at logs to see it's using a cookie for login.

This does not fix any bugs. 

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/cache-config/t3c-request/t3c-request.go
+++ b/cache-config/t3c-request/t3c-request.go
@@ -68,5 +68,5 @@ func main() {
 			os.Exit(3)
 		}
 	}
-	defer cfg.TCCfg.TOClient.WriteFsCookie(toreq.FsCookiePath)
+	cfg.TCCfg.TOClient.WriteFsCookie(toreq.FsCookiePath)
 }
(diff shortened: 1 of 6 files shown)


cache-config/t3c-request/t3c-request.go
```diff
@@ -68,4 +68,5 @@ func main() {
 			os.Exit(3)
 		}
 	}
+	defer cfg.TCCfg.TOClient.WriteFsCookie(toreq.FsCookiePath)
```
rob05c, line 71: This doesn't need to be a `defer`, since it's at the end of the function anyway

jpappa200: Removed defer

rob05c (approved):
LGTM. Code looks good. Tested, works as expected, login endpoint is not called if a valid cookie file exists on subsequent runs and requests in the same run, modified the cookie file to be both invalid JSON and valid JSON with a bad cookie, t3c correctly continued past the error with a normal login.

PR #6577, opened 2022-02-21
JWT Authorization

mattjackson220:
Added the ability to use a JWT for authorization and added to CDNi operations

This PR adds the ability to use a JWT for authorization. Its not a huge change from the mojolicious cookie we currently use since it uses the mojolicious cookie as a field in the JWT. However, use of a JWT is required for CDNi operations.  The existing CDNi endpoints have been updated to use this JWT.

<hr/>

- Documentation
- Traffic Ops
- Traffic Portal

Run TO and TP
POST to the `/api/4.0/user/login?token-type=jwt` endpoint with your username and password and get the `access-token` from the response
Use that access token instead of a mojo cookie to reach any normal endpoint
Use that access token and verify the expected functionality on the `/OC/FCI/advertisement`, `/OC/CI/configuration`,  and `/OC/CI/configuration/{host}` endpoints
Verify that the docs build and look good

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Traffic Monitor config option `distributed_polling` which enables the ability for Traffic Monitor to poll a subset of the CDN and divide into "local peer groups" and "distributed peer groups". Traffic Monitors in the same group are local peers, while Traffic Monitors in other groups are distibuted peers. Each TM group polls the same set of cachegroups and gets availability data for the other cachegroups from other TM groups. This allows each TM to be responsible for polling a subset of the CDN while still having a full view of CDN availability. In order to use this, `stat_polling` must be disabled.
 - Added support for a new Traffic Ops GLOBAL profile parameter -- `tm_query_status_override` -- to override which status of Traffic Monitors to query (default: ONLINE).
 - Traffic Router: Add support for `file`-protocol URLs for the `geolocation.polling.url` for the Geolocation database.
-- Added the ability to use a `JWT` for login in addition to a mojolicious cookie.
+- Added functionality for login to provide a Bearer token and for that token to be later used for authorization.
 
 ### Fixed
 - Update traffic\_portal dependencies to mitigate `npm audit` issues.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Traffic Monitor config option `distributed_polling` which enables the ability for Traffic Monitor to poll a subset of the CDN and divide into "local peer groups" and "distributed peer groups". Traffic Monitors in the same group are local peers, while Traffic Monitors in other groups are distibuted peers. Each TM group polls the same set of cachegroups and gets availability data for the other cachegroups from other TM groups. This allows each TM to be responsible for polling a subset of the CDN while still having a full view of CDN availability. In order to use this, `stat_polling` must be disabled.
 - Added support for a new Traffic Ops GLOBAL profile parameter -- `tm_query_status_override` -- to override which status of Traffic Monitors to query (default: ONLINE).
 - Traffic Router: Add support for `file`-protocol URLs for the `geolocation.polling.url` for the Geolocation database.
-- Added the ability to use a `JWT` for login instead of a mojolicious cookie.
+- Added the ability to use a `JWT` for login in addition to a mojolicious cookie.
 
 ### Fixed
 - Update traffic\_portal dependencies to mitigate `npm audit` issues.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Traffic Monitor config option `distributed_polling` which enables the ability for Traffic Monitor to poll a subset of the CDN and divide into "local peer groups" and "distributed peer groups". Traffic Monitors in the same group are local peers, while Traffic Monitors in other groups are distibuted peers. Each TM group polls the same set of cachegroups and gets availability data for the other cachegroups from other TM groups. This allows each TM to be responsible for polling a subset of the CDN while still having a full view of CDN availability. In order to use this, `stat_polling` must be disabled.
 - Added support for a new Traffic Ops GLOBAL profile parameter -- `tm_query_status_override` -- to override which status of Traffic Monitors to query (default: ONLINE).
 - Traffic Router: Add support for `file`-protocol URLs for the `geolocation.polling.url` for the Geolocation database.
+- Added the ability to use a `JWT` for login instead of a mojolicious cookie.
 
 ### Fixed
 - Update traffic\_portal dependencies to mitigate `npm audit` issues.
--- a/docs/source/admin/traffic_ops.rst
+++ b/docs/source/admin/traffic_ops.rst
@@ -500,7 +500,6 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 	.. versionadded:: 6.2
 
 	:dcdn_id: A string representing this :abbr:`CDN (Content Delivery Network)` to be used in the :abbr:`JWT (JSON Web Token)` and subsequently in :abbr:`CDNi (Content Delivery Network Interconnect)` operations.
-	:jwt_decoding_secret: A string used to decode the :abbr:`JWT (JSON Web Token)` to get information for :abbr:`CDNi (Content Delivery Network Interconnect)` operations.
 
 
 Example cdn.conf
--- a/docs/source/api/v4/oc_ci_configuration.rst
+++ b/docs/source/api/v4/oc_ci_configuration.rst
@@ -29,18 +29,20 @@ Triggers an asynchronous task to update the configuration for the :abbr:`uCDN (U
 :Roles Required: "admin" or "operations"
 :Permissions Required: CDNI:UPDATE
 :Response Type:  Object
-:Headers Required: "Authorization"
 
 Request Structure
 -----------------
-.. table:: Optional Request Headers
+This requires authorization using a :abbr:`JWT (JSON Web Token)` provided by the :abbr:`dCDN (Downstream Content Delivery Network)` to identify the :abbr:`uCDN (Upstream Content Delivery Network)`. This token must include the following claims:
 
-	+-----------------+------------------------------------------------------------------------------------------------------------------------------+
-	|    Name         | Description                                                                                                                  |
-	+=================+==============================================================================================================================+
-	|  Authorization  | A :abbr:`JWT (JSON Web Token)` provided by the :abbr:`dCDN (Downstream Content Delivery Network)` to identify the            |
-	|                 | :abbr:`uCDN (Upstream Content Delivery Network)`. This can be proivded as a header or a cookie.                              |
-	+-----------------+------------------------------------------------------------------------------------------------------------------------------+
+.. table:: Required JWT claims
+
+	+-----------------+--------------------------------------------------------------------------------------------------------------------+
+	|    Name         | Description                                                                                                        |
+	+=================+====================================================================================================================+
+	|      iss        | Issuer claim as a string key for the :abbr:`uCDN (Upstream Content Delivery Network)`                              |
+	|      aud        | Audience claim as a string key for the :abbr:`dCDN (Downstream Content Delivery Network)`                          |
+	|      exp        | Expiration claim as the expiration date as a Unix epoch timestamp (in seconds)                                     |
+	+-----------------+--------------------------------------------------------------------------------------------------------------------+
 
 :type: A string of the type of metadata to follow. See :rfc:`8006` for possible values. Only a selection of these are supported.
 :host: A string of the domain that the requested updates will change.
--- a/docs/source/api/v4/oc_ci_configuration.rst
+++ b/docs/source/api/v4/oc_ci_configuration.rst
@@ -23,6 +23,8 @@
 =======
 Triggers an asynchronous task to update the configuration for the :abbr:`uCDN (Upstream Content Delivery Network)` by adding the request to a queue to be reviewed later. This returns a 202 Accepted status and an endpoint to be used for status updates.
 
+.. note:: Users with an "admin" role will need to provide a "ucdn" query parameter to bypass the need for :abbr:`uCDN (Upstream Content Delivery Network)` information in the :abbr:`JWT (JSON Web Token)` and allow admins to view all :abbr:`CDNi (Content Delivery Network Interconnect)` information.
+
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
 :Permissions Required: CDNI:UPDATE
@@ -31,13 +33,13 @@ Triggers an asynchronous task to update the configuration for the :abbr:`uCDN (U
 
 Request Structure
 -----------------
-.. table:: Request Required Headers
+.. table:: Optional Request Headers
 
 	+-----------------+------------------------------------------------------------------------------------------------------------------------------+
 	|    Name         | Description                                                                                                                  |
 	+=================+==============================================================================================================================+
 	|  Authorization  | A :abbr:`JWT (JSON Web Token)` provided by the :abbr:`dCDN (Downstream Content Delivery Network)` to identify the            |
-	|                 | :abbr:`uCDN (Upstream Content Delivery Network)`                                                                             |
+	|                 | :abbr:`uCDN (Upstream Content Delivery Network)`. This can be proivded as a header or a cookie.                              |
 	+-----------------+------------------------------------------------------------------------------------------------------------------------------+
 
 :type: A string of the type of metadata to follow. See :rfc:`8006` for possible values. Only a selection of these are supported.
--- a/docs/source/api/v4/oc_ci_configuration.rst
+++ b/docs/source/api/v4/oc_ci_configuration.rst
@@ -23,7 +23,7 @@
 =======
 Triggers an asynchronous task to update the configuration for the :abbr:`uCDN (Upstream Content Delivery Network)` by adding the request to a queue to be reviewed later. This returns a 202 Accepted status and an endpoint to be used for status updates.
 
-.. note:: Users with an "admin" role will need to provide a "ucdn" query parameter to bypass the need for :abbr:`uCDN (Upstream Content Delivery Network)` information in the :abbr:`JWT (JSON Web Token)` and allow admins to view all :abbr:`CDNi (Content Delivery Network Interconnect)` information.
+.. note:: Users with the ``ICDN:UCDN-OVERRIDE`` permission will need to provide a "ucdn" query parameter to bypass the need for :abbr:`uCDN (Upstream Content Delivery Network)` information in the :abbr:`JWT (JSON Web Token)` and allow them to view all :abbr:`CDNi (Content Delivery Network Interconnect)` information.
 
 :Auth. Required: Yes
 :Roles Required: "admin" or "operations"
@@ -40,7 +40,9 @@ This requires authorization using a :abbr:`JWT (JSON Web Token)` provided by the
 	|    Name         | Description                                                                                                        |
 	+=================+====================================================================================================================+
 	|      iss        | Issuer claim as a string key for the :abbr:`uCDN (Upstream Content Delivery Network)`                              |
+	+-----------------+--------------------------------------------------------------------------------------------------------------------+
 	|      aud        | Audience claim as a string key for the :abbr:`dCDN (Downstream Content Delivery Network)`                          |
+	+-----------------+--------------------------------------------------------------------------------------------------------------------+
 	|      exp        | Expiration claim as the expiration date as a Unix epoch timestamp (in seconds)                                     |
 	+-----------------+--------------------------------------------------------------------------------------------------------------------+
 
--- a/docs/source/api/v4/users.rst
+++ b/docs/source/api/v4/users.rst
@@ -92,8 +92,6 @@ Response Structure
 :tenantId:          The integral, unique identifier of the tenant to which this user belongs
 :ucdn:              The name of the :abbr:`uCDN (Upstream Content Delivery Network)` to which the user belongs
 
-	.. versionadded:: 6.2
-
 	.. note:: This field is optional and only used if :abbr:`CDNi (Content Delivery Network Interconnect)` is in use.
 
 :uid:               A deprecated field only kept for legacy compatibility reasons that used to contain the UNIX user ID of the user - now it is always ``null``
@@ -175,8 +173,6 @@ Request Structure
 	.. note:: This field is optional if and only if tenancy is not enabled in Traffic Control
 
 :ucdn:              The name of the :abbr:`uCDN (Upstream Content Delivery Network)` to which the user belongs
-
-	.. versionadded:: 6.2
 
 	.. note:: This field is optional and only used if :abbr:`CDNi (Content Delivery Network Interconnect)` is in use.
 
--- a/docs/source/api/v4/users.rst
+++ b/docs/source/api/v4/users.rst
@@ -91,9 +91,6 @@ Response Structure
 :tenant:            The name of the tenant to which this user belongs
 :tenantId:          The integral, unique identifier of the tenant to which this user belongs
 :ucdn:              The name of the :abbr:`uCDN (Upstream Content Delivery Network)` to which the user belongs
-
-	.. note:: This field is optional and only used if :abbr:`CDNi (Content Delivery Network Interconnect)` is in use.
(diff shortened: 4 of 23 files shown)


traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -154,6 +154,42 @@ func LoginHandler(db *sqlx.DB, cfg config.Config) http.HandlerFunc {
 				httpCookie := tocookie.GetCookie(form.Username, defaultCookieDuration, cfg.Secrets[0])
 				http.SetCookie(w, httpCookie)
 
+				var jwtToken *jwt.Token
+				useJwt := r.URL.Query().Get("token-type") == "jwt"
+				var jwtSigned string
+				if useJwt {
```
ocket8888, line 160: why not just always also set this cookie?

traffic_portal/app/src/common/modules/form/user/form.user.tpl.html
```diff
@@ -82,6 +82,14 @@
                     <small ng-show="user.tenantId"><a href="/#!/tenants/{{user.tenantId}}" target="_blank">View Details&nbsp;&nbsp;<i class="fa fs-xs fa-external-link"></i></a></small>
                 </div>
             </div>
+            <div class="form-group" ng-class="{'has-error': hasError(userForm.uCDN), 'has-feedback': hasError(userForm.uCDN)}">
+                <label class="control-label col-md-2 col-sm-2 col-xs-12">Upstream CDN</label>
```
ocket8888, line 86: Label should have a `for` attribute with a value set to the ID of the labeled control

lib/go-tc/users.go
```diff
@@ -294,6 +294,7 @@ type UserV40 struct {
 	Tenant               *string    `json:"tenant"`
 	TenantID             int        `json:"tenantId" db:"tenant_id"`
 	Token                *string    `json:"-" db:"token"`
+	UCDN                 *string    `json:"ucdn"`
```
ocket8888, line 297: It looks like if this is `NULL` it'll be coerced to an empty string in JWTs - does it need to be "null-able"?

traffic_ops/traffic_ops_golang/login/login.go
mattjackson220: good call! might as well

traffic_portal/app/src/common/modules/form/user/form.user.tpl.html
```diff
@@ -82,6 +82,14 @@
                     <small ng-show="user.tenantId"><a href="/#!/tenants/{{user.tenantId}}" target="_blank">View Details&nbsp;&nbsp;<i class="fa fs-xs fa-external-link"></i></a></small>
                 </div>
             </div>
+            <div class="form-group" ng-class="{'has-error': hasError(userForm.uCDN), 'has-feedback': hasError(userForm.uCDN)}">
+                <label for="uCDN" class="control-label col-md-2 col-sm-2 col-xs-12">Upstream CDN</label>
+                <div id="uCDN" class="col-md-10 col-sm-10 col-xs-12">
```
ocket8888, line 87: `div` is not a [label-able element](https://html.spec.whatwg.org/multipage/forms.html#category-label), the `id` that corresponds to the `label`'s `for` attribute should be on the control it labels.

docs/source/api/v4/users.rst
```diff
@@ -90,6 +90,12 @@ Response Structure
 :stateOrProvince:   The name of the state or province where this user resides
 :tenant:            The name of the tenant to which this user belongs
 :tenantId:          The integral, unique identifier of the tenant to which this user belongs
+:ucdn:              The name of the :abbr:`uCDN (Upstream Content Delivery Network)` to which the user belongs
+
+	.. versionadded:: 6.2
```
ocket8888, line 95: APIv4 is still in development, so you don't need to call out in which ATC version this field (or any field) was added. As far as any client is concerned, it's new and native to APIv4.0. Notice, for example, the type of the `role` field changed after 4.0 had been "released" as an unstable API, but it doesn't have a `.. versionchanged::` directive.

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -518,7 +537,14 @@ func checkBearerToken(bearerToken string, inf *api.APIInfo) (string, error) {
 		return "", errors.New("invalid token - incorrect dcdn")
 	}
 	if ucdn == "" {
-		return "", errors.New("invalid token - empty ucdn field")
+		if inf.User.PrivLevel == auth.PrivLevelAdmin {
```
ocket8888, line 540: PrivLevel is deprecated and not even visible in APIv4 - does this correspond to some Permission?

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -472,12 +491,12 @@ func validateHostExists(host string, tx *sql.Tx) (int, error, error) {
 
 func checkBearerToken(bearerToken string, inf *api.APIInfo) (string, error) {
```
ocket8888, line 492: It seems this is extracting the `ucdn` claim from the token, but it doesn't appear to verify that the value matches the user's set `ucdn` value in the database - is that necessary?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a Traffic Ops endpoints to `PUT` a requested configuration change for a full configuration or per host and an endpoint to approve or deny the request.
 - Traffic Monitor config option `distributed_polling` which enables the ability for Traffic Monitor to poll a subset of the CDN and divide into "local peer groups" and "distributed peer groups". Traffic Monitors in the same group are local peers, while Traffic Monitors in other groups are distibuted peers. Each TM group polls the same set of cachegroups and gets availability data for the other cachegroups from other TM groups. This allows each TM to be responsible for polling a subset of the CDN while still having a full view of CDN availability. In order to use this, `stat_polling` must be disabled.
 - Traffic Router: Add support for `file`-protocol URLs for the `geolocation.polling.url` for the Geolocation database.
+- Added the ability to use a `JWT` for login in addition to a mojolicious cookie.
```
ocket8888, line 12: The JWT isn't _used for_ login, it's _provided by_ login. The fact that it's a JWT is also just an implementation detail, what it does is allow a Bearer token to be used for authorization - _after_ authentication.

docs/source/admin/traffic_ops.rst
```diff
@@ -500,7 +500,6 @@ This file deals with the configuration parameters of running Traffic Ops itself.
 	.. versionadded:: 6.2
 
 	:dcdn_id: A string representing this :abbr:`CDN (Content Delivery Network)` to be used in the :abbr:`JWT (JSON Web Token)` and subsequently in :abbr:`CDNi (Content Delivery Network Interconnect)` operations.
-	:jwt_decoding_secret: A string used to decode the :abbr:`JWT (JSON Web Token)` to get information for :abbr:`CDNi (Content Delivery Network Interconnect)` operations.
```
ocket8888, line 503: Why are we getting rid of this? Or, maybe it makes more sense to ask: if it's unnecessary/unwanted, why was it added in the first place? According to the above `.. versionadded::` directive, this wasn't in 6.1 so we can take it out if we want, no problems there, I'm just curious.

docs/source/api/v4/oc_ci_configuration.rst
```diff
@@ -31,13 +33,13 @@ Triggers an asynchronous task to update the configuration for the :abbr:`uCDN (U
...
 -----------------
-.. table:: Request Required Headers
+.. table:: Optional Request Headers
 
 	+-----------------+------------------------------------------------------------------------------------------------------------------------------+
 	|    Name         | Description                                                                                                                  |
 	+=================+==============================================================================================================================+
 	|  Authorization  | A :abbr:`JWT (JSON Web Token)` provided by the :abbr:`dCDN (Downstream Content Delivery Network)` to identify the            |
-	|                 | :abbr:`uCDN (Upstream Content Delivery Network)`                                                                             |
+	|                 | :abbr:`uCDN (Upstream Content Delivery Network)`. This can be proivded as a header or a cookie.                              |
```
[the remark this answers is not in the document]
mattjackson220: this was meant just so that it follows the SVA spec for CDNi endpoints but not have bigger impacts on our full system to minimize effort and make sure we dont break anything. but reading the spec again it says that the JWT is typically in the authorization header, not that it must be so im now debating just removing that functionality from these endpoints and leaving it as a cookie

docs/source/admin/traffic_ops.rst
mattjackson220: i added it originally with the intent to follow the CDNi spec without changing our full system but the more i read into it the more i realized it requires JWT Auth for our full system. so im removing this to just use the `secrets` field above in the cdn.conf like the mojo cookie does

docs/source/api/v4/oc_ci_configuration.rst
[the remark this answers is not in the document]
ocket8888: I think it's great to put the token in the Authorization header, I just think it's easier to support that as a general authorization method (i.e. instead of the `mojolicious` cookie) for the whole API, rather than only sending the header for specific endpoints.

docs/source/admin/traffic_ops.rst
ocket8888: Oh, if it's what the `mojolicious` cookie uses then it makes sense to me to use it as the JWT secret.

docs/source/api/v4/oc_ci_configuration.rst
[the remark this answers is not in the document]
mattjackson220: fair. ill remove it from this PR since it would be better as its own PR i think. with this JWT change we were trying to keep it as "break safe" as possible so im not sure if adding in Authorization header support would require a mailing list discussion or not. but i dont think it should be part of this PR either way

traffic_ops/traffic_ops_golang/cdni/shared.go
ocket8888: Right but just because it came from us doesn't mean it's up-to-date. If someone's `ucdn` changes since their token was issued, will that cause problems?

docs/source/api/v4/oc_ci_configuration.rst
```diff
@@ -23,22 +23,26 @@
...
+This requires authorization using a :abbr:`JWT (JSON Web Token)` provided by the :abbr:`dCDN (Downstream Content Delivery Network)` to identify the :abbr:`uCDN (Upstream Content Delivery Network)`. This token must include the following claims:
+
+.. table:: Required JWT claims
+
+	+-----------------+--------------------------------------------------------------------------------------------------------------------+
+	|    Name         | Description                                                                                                        |
+	+=================+====================================================================================================================+
+	|      iss        | Issuer claim as a string key for the :abbr:`uCDN (Upstream Content Delivery Network)`                              |
+	|      aud        | Audience claim as a string key for the :abbr:`dCDN (Downstream Content Delivery Network)`                          |
+	|      exp        | Expiration claim as the expiration date as a Unix epoch timestamp (in seconds)                                     |
```
ocket8888, line 44: I believe these are meant to be separate rows, but the way they are currently this is a one-row table. Well, two if you count the header.

docs/source/api/v4/oc_ci_configuration.rst
```diff
@@ -23,22 +23,26 @@
 =======
 Triggers an asynchronous task to update the configuration for the :abbr:`uCDN (Upstream Content Delivery Network)` by adding the request to a queue to be reviewed later. This returns a 202 Accepted status and an endpoint to be used for status updates.
 
+.. note:: Users with an "admin" role will need to provide a "ucdn" query parameter to bypass the need for :abbr:`uCDN (Upstream Content Delivery Network)` information in the :abbr:`JWT (JSON Web Token)` and allow admins to view all :abbr:`CDNi (Content Delivery Network Interconnect)` information.
```
ocket8888, line 26: There's only one "admin" Role. The way this is set up currently, there's no way to make any Role that has permission to bypass the need for uCDN info. Only users with exactly the "admin" Role, regardless of any Permission set or "Priv Level", will be able to use this functionality. Even the "admin" Role itself can't create a Role that's allowed to do this. If that's intentional, then this is just a semantic change in the docs ("an"->"the"). If you want to declare some set of behavior that requires special Permissions, you should use a Permission.

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -518,7 +531,14 @@ func checkBearerToken(bearerToken string, inf *api.APIInfo) (string, error) {
...
 	if ucdn == "" {
-		return "", errors.New("invalid token - empty ucdn field")
+		if inf.User.RoleName == tc.AdminRoleName {
+			ucdn = inf.Params["ucdn"]
+			if ucdn == "" {
+				return "", errors.New("admin level ucdn requests require a ucdn query parameter")
+			}
+		} else {
+			return "", errors.New("invalid token - empty ucdn field")
+		}
```
ocket8888, line 541: Do we need to check if this ucdn claim from the JWT cookie matches the user's stored `ucdn` value?

mattjackson220: i dont think we need to since we control all sides of the equation but i can add it in to be extra safe. people shouldnt be changing ucdns really and if they do then they would still see their old ucdn until they log out and log back in or the token expires in 6 hours. i was trying to avoid an extra DB call but might as well be extra safe

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -1029,9 +1030,42 @@ func ParseDBError(ierr error) (error, error, int) {
...
 func GetUserFromReq(w http.ResponseWriter, r *http.Request, secret string) (auth.CurrentUser, error, error, int) {
-	cookie, err := r.Cookie(tocookie.Name)
-	if err != nil {
-		return auth.CurrentUser{}, errors.New("Unauthorized, please log in."), errors.New("error getting cookie: " + err.Error()), http.StatusUnauthorized
+	var cookie *http.Cookie
+
+	if r.Header.Get("Authorization") != "" && strings.Contains(r.Header.Get("Authorization"), "Bearer") {
+		givenToken := r.Header.Get("Authorization")
+		if strings.Contains(givenToken, "access_token") {
+			tokenSplit := strings.Split(givenToken, "=")
```
ocket8888, line 1038: Do we need to support this? I think that would be pretty unusual to see in a Bearer token. That's usually just the token.

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -1029,9 +1030,35 @@ func ParseDBError(ierr error) (error, error, int) {
 // GetUserFromReq returns the current user, any user error, any system error, and an error code to be returned if either error was not nil.
 // This also uses the given ResponseWriter to refresh the cookie, if it was valid.
 func GetUserFromReq(w http.ResponseWriter, r *http.Request, secret string) (auth.CurrentUser, error, error, int) {
-	cookie, err := r.Cookie(tocookie.Name)
-	if err != nil {
-		return auth.CurrentUser{}, errors.New("Unauthorized, please log in."), errors.New("error getting cookie: " + err.Error()), http.StatusUnauthorized
+	var cookie *http.Cookie
+
+	if r.Header.Get("Authorization") != "" && strings.Contains(r.Header.Get("Authorization"), "Bearer") {
```
srijeet0406, line 1035: Could we define a couple of constants for these strings?

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -1029,9 +1030,35 @@ func ParseDBError(ierr error) (error, error, int) {
...
+		givenToken := r.Header.Get("Authorization")
+		tokenSplit := strings.Split(givenToken, " ")
+		if len(tokenSplit) > 1 {
+			givenToken = tokenSplit[1]
+		}
+		bearerCookie, err := getCookieFromAccessToken(givenToken, secret)
+		if err != nil {
+			return auth.CurrentUser{}, errors.New("Unauthorized, please log in."), err, http.StatusUnauthorized
```
srijeet0406, line 1043: nit: Errors should start with lowercase letters.

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -1029,9 +1030,35 @@ func ParseDBError(ierr error) (error, error, int) {
...
+	} else {
+		for _, givenCookie := range r.Cookies() {
+			if cookie != nil {
+				break
+			}
+			switch givenCookie.Name {
+			case "access_token":
+				bearerCookie, err := getCookieFromAccessToken(givenCookie.Value, secret)
+				if err != nil {
+					return auth.CurrentUser{}, errors.New("Unauthorized, please log in."), err, http.StatusUnauthorized
```
srijeet0406, line 1055: Same nit here re:error ^

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -1029,9 +1030,35 @@ func ParseDBError(ierr error) (error, error, int) {
...
+		if err != nil {
+			return auth.CurrentUser{}, errors.New("Unauthorized, please log in."), err, http.StatusUnauthorized
+		}
+		cookie = bearerCookie
+	} else {
+		for _, givenCookie := range r.Cookies() {
+			if cookie != nil {
+				break
+			}
+			switch givenCookie.Name {
```
srijeet0406, line 1051: Can there be a case where there is a (pointer to a)cookie in the array, but it is `nil`? If yes, this might panic in that case.

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -1029,9 +1030,35 @@ func ParseDBError(ierr error) (error, error, int) {
...
+			case "access_token":
```
srijeet0406, line 1052: Might be good to make this into a constant

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -1075,6 +1102,34 @@ func GetUserFromReq(w http.ResponseWriter, r *http.Request, secret string) (auth
...
+func getCookieFromAccessToken(bearerToken string, secret string) (*http.Cookie, error) {
+	var cookie *http.Cookie
+	claims := jwt.MapClaims{}
+	token, err := jwt.ParseWithClaims(bearerToken, claims, func(token *jwt.Token) (interface{}, error) {
+		return []byte(secret), nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("parsing claims: %w", err)
+	}
+	if !token.Valid {
```
srijeet0406, line 1114: I didn't look in the underlying functions, but can there be a case where there is no error, but the returned token pointer is `nil`? 
If so, this will panic in that case.
If not, then disregard this comment :)

traffic_ops/traffic_ops_golang/api/api.go
```diff
@@ -1075,6 +1102,34 @@ func GetUserFromReq(w http.ResponseWriter, r *http.Request, secret string) (auth
...
+	}
+
+	for key, val := range claims {
+		switch key {
+		case "mojoCookie":
+			if _, ok := val.(string); !ok {
+				return nil, errors.New("invalid token - mojoCookie must be a string")
+			}
+			cookie = &http.Cookie{
+				Value: val.(string),
```
srijeet0406, line 1125: Can we just use the value directly from line 1121, instead of type casting it again?
Also, `mojoCookie` could be a constant.

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -73,14 +72,16 @@ func GetCapabilities(w http.ResponseWriter, r *http.Request) {
 	}
 	defer inf.Close()
 
-	if inf.Config.Cdni == nil || inf.Config.Cdni.JwtDecodingSecret == "" || inf.Config.Cdni.DCdnId == "" {
+	if inf.Config.Cdni == nil || inf.Config.Secrets[0] == "" || inf.Config.Cdni.DCdnId == "" {
```
srijeet0406, line 75: Unless it is being done elsewhere, we should probably check to make sure that `inf.Config.Secrets` has length of `>=1` before getting the first item from there.

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -106,6 +107,25 @@ func GetCapabilities(w http.ResponseWriter, r *http.Request) {
 	api.WriteRespRaw(w, r, fciCaps)
 }
 
+func getBearerToken(r *http.Request) string {
+	if r.Header.Get("Authorization") != "" && strings.Contains(r.Header.Get("Authorization"), "Bearer") {
+		givenToken := r.Header.Get("Authorization")
+		if strings.Contains(givenToken, "access_token") {
+			givenToken = strings.Split(givenToken, "=")[1]
```
srijeet0406, line 114: We should probably do a length check before getting the second item from the array here.

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -106,6 +107,25 @@ func GetCapabilities(w http.ResponseWriter, r *http.Request) {
...
+		} else {
+			givenToken = strings.Split(givenToken, " ")[1]
```
srijeet0406, line 116: Same comment here ^

traffic_ops/traffic_ops_golang/api/api.go
mattjackson220: not that i can find but no reason not to check! ill put it in there

traffic_ops/traffic_ops_golang/cdni/shared.go
mattjackson220: this is handled at start up in config.go ParseConfig so if Secrets has a length of 0 it will fail at startup

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -120,14 +140,15 @@ func PutHostConfiguration(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if inf.Config.Cdni == nil || inf.Config.Cdni.JwtDecodingSecret == "" || inf.Config.Cdni.DCdnId == "" {
+	if inf.Config.Cdni == nil || inf.Config.Secrets[0] == "" || inf.Config.Cdni.DCdnId == "" {
```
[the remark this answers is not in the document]
mattjackson220: same response re: checked at startup

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -190,14 +211,15 @@ func PutConfiguration(w http.ResponseWriter, r *http.Request) {
 	}
 	defer inf.Close()
 
-	if inf.Config.Cdni == nil || inf.Config.Cdni.JwtDecodingSecret == "" || inf.Config.Cdni.DCdnId == "" {
+	if inf.Config.Cdni == nil || inf.Config.Secrets[0] == "" || inf.Config.Cdni.DCdnId == "" {
```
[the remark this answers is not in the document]
mattjackson220: same response re: checked at startup

traffic_ops/traffic_ops_golang/cdni/shared.go
mattjackson220: removed this one since having an "=" in there is uncommon per a comment above

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -517,8 +539,20 @@ func checkBearerToken(bearerToken string, inf *api.APIInfo) (string, error) {
...
+
 	if ucdn == "" {
-		return "", errors.New("invalid token - empty ucdn field")
+		if inf.User.Can("ICDN:UCDN-OVERRIDE") {
+			ucdn = inf.Params["ucdn"]
+			if ucdn == "" {
+				return "", errors.New("admin level ucdn requests require a ucdn query parameter")
+			}
+		} else {
+			return "", errors.New("invalid token - empty ucdn field")
```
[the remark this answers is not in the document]
mattjackson220: i dont think so. this error will get hit if the Bearer token has an empty string as the `ucdn` field. a valid Bearer token from our login would only have that if we havent assigned the user a `ucdn`. the `ICDN:UCDN-OVERRIDE` permission is a bypass for that requirement so that people who are admins or have that permission can see the response for any ucdn in order to know what people in those ucdns would get back

zrhoffman:
Let's try to merge #6434 first so that #6577 can use github.com/lestrrat-go/jwx/jwt.

zrhoffman:
Update: #6434 is merged, #6577 is unblocked

srijeet0406 (approved):
This looks great! All unit/ api tests pass.
Locally tested and made sure the new(and old) functionality works as expected.
The `CDNI-CAPACITY:READ` just needs to be a part of all roles with privilege level `>=10`.
@mattjackson220 said he will add that in a following PR.

PR #6578, opened 2022-02-21
Add support for File URLs to database loader

ocket8888:
This PR adds support for using `file://`-protocol URLs for the geolocation database polling URL.

<hr/>

- Traffic Router

Run Traffic Router with its `geolocation.polling.url` Parameter set to a `file://`-protocol URL.

## PR submission checklist
- [ ] This PR has tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a new Traffic Ops endpoint to `GET` capacity and telemetry data for CDNi integration.
 - Added a Traffic Ops endpoints to `PUT` a requested configuration change for a full configuration or per host and an endpoint to approve or deny the request.
 - Traffic Monitor config option `distributed_polling` which enables the ability for Traffic Monitor to poll a subset of the CDN and divide into "local peer groups" and "distributed peer groups". Traffic Monitors in the same group are local peers, while Traffic Monitors in other groups are distibuted peers. Each TM group polls the same set of cachegroups and gets availability data for the other cachegroups from other TM groups. This allows each TM to be responsible for polling a subset of the CDN while still having a full view of CDN availability. In order to use this, `stat_polling` must be disabled.
-- Traffic Router: Add support for `file://`-protocol URLs for the `geolocation.polling.url` for the Geolocation database.
+- Traffic Router: Add support for `file`-protocol URLs for the `geolocation.polling.url` for the Geolocation database.
 
 ### Fixed
 - Update traffic_portal dependencies to mitigate `npm audit` issues.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a new Traffic Ops endpoint to `GET` capacity and telemetry data for CDNi integration.
 - Added a Traffic Ops endpoints to `PUT` a requested configuration change for a full configuration or per host and an endpoint to approve or deny the request.
 - Traffic Monitor config option `distributed_polling` which enables the ability for Traffic Monitor to poll a subset of the CDN and divide into "local peer groups" and "distributed peer groups". Traffic Monitors in the same group are local peers, while Traffic Monitors in other groups are distibuted peers. Each TM group polls the same set of cachegroups and gets availability data for the other cachegroups from other TM groups. This allows each TM to be responsible for polling a subset of the CDN while still having a full view of CDN availability. In order to use this, `stat_polling` must be disabled.
+- Traffic Router: Add support for `file://`-protocol URLs for the `geolocation.polling.url` for the Geolocation database.
 
 ### Fixed
 - Update traffic_portal dependencies to mitigate `npm audit` issues.
--- a/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
+++ b/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
@@ -23,6 +23,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.net.URLConnection;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
@@ -42,6 +43,7 @@ import org.apache.traffic_control.traffic_router.core.router.TrafficRouterManage
 
 import static org.apache.commons.codec.digest.DigestUtils.md5Hex;
 
+@SuppressWarnings({"PMD.CyclomaticComplexity"})
 public abstract class AbstractServiceUpdater {
 	private static final Logger LOGGER = LogManager.getLogger(AbstractServiceUpdater.class);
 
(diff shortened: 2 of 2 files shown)


traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
```diff
@@ -366,13 +369,15 @@ protected File downloadDatabase(final String url, final File existingDb) throws
...
 			 OutputStream out = new FileOutputStream(outputFile)
 		) {
-			eTag = conn.getHeaderField("ETag");
-
-			if (conn.getResponseCode() == HttpURLConnection.HTTP_NOT_MODIFIED) {
-				LOGGER.info("[" + getClass().getSimpleName() + "] " + url + " not modified since our existing database's last update time of " + new Date(existingDb.lastModified()));
-				return existingDb;
+			if (conn instanceof HttpURLConnection) {
+				eTag = conn.getHeaderField("ETag");
+				if (((HttpURLConnection) conn).getResponseCode() == HttpURLConnection.HTTP_NOT_MODIFIED) {
```
zrhoffman, line 374: If `conn` is a `FileURLConnection`, it should check `conn.getLastModified()` against `existingDb.lastModified()` and return the existing DB if they are the same.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added a new Traffic Ops endpoint to `GET` capacity and telemetry data for CDNi integration.
 - Added a Traffic Ops endpoints to `PUT` a requested configuration change for a full configuration or per host and an endpoint to approve or deny the request.
 - Traffic Monitor config option `distributed_polling` which enables the ability for Traffic Monitor to poll a subset of the CDN and divide into "local peer groups" and "distributed peer groups". Traffic Monitors in the same group are local peers, while Traffic Monitors in other groups are distibuted peers. Each TM group polls the same set of cachegroups and gets availability data for the other cachegroups from other TM groups. This allows each TM to be responsible for polling a subset of the CDN while still having a full view of CDN availability. In order to use this, `stat_polling` must be disabled.
+- Traffic Router: Add support for `file://`-protocol URLs for the `geolocation.polling.url` for the Geolocation database.
```
zrhoffman, line 11: In Java, the `//` is just interpreted as redundant path components for absolute paths. `file:` can be used with a relative paths, so the `//` should be removed.

zrhoffman (changes requested):

CHANGELOG.md
ocket8888: > In Java, the `//` is just interpreted as redundant path components for absolute paths. `file:` can be used with a relative paths, so the `//` should be removed.

I think actually the first `//` is used to specify a host ("authority"), so `///` is a synonym for `//localhost/` because the file host is implicitly understood to be the local machine when the hostname is empty. According to [RFC8089](https://datatracker.ietf.org/doc/html/rfc8089) all paths are absolute, and the local path part of a file URI must begin with `/`, so `file:///my/file/path`, `file://localhost/my/file/path` and `file:/my/file/path` are all valid and point to the same file, but `file:my/file/path` is invalid.

I can drop the `//`, though, since although that's the most common usage it's not required and `file:` is sufficient to name the protocol.

zrhoffman: Works for me. They aren't totally the same, since using `file://` with a relative path makes the connection an `FtpURLConnection`, but that's an edge case, so keeping `//` makes sense.

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
ocket8888: Where's the definition of `FileURLConnection` - I've seen it in stack traces as `sun.net.www.protocol.file.FileURLConnection`, but trying to import that causes compilation to fail with:

```
error: package sun.net.www.protocol.file is not visible
```

Could that possibly be an implementation detail of the JRE used?

zrhoffman: Yeah, that class is not visible. Technically, it could be made visible by adding
```xml
					<compilerArgs>
						<arg>-XDignore.symbol.file</arg>
						<arg>--add-exports</arg>
						<arg>java.base/sun.net.www.protocol.file=ALL-UNNAMED</arg>
					</compilerArgs>
```
to the `maven-compiler-plugin` configuration options in place of https://github.com/apache/trafficcontrol/blob/739124d0318b184185ecf2f80cbbc12c6c36e0ba/traffic_router/core/pom.xml#L148

but doing that results in a PMD failure:
```java
[INFO] PMD Failure: org.apache.traffic_control.traffic_router.core.loc.AbstractServiceUpdater:36 Rule:DontImportSun Priority:4 Avoid importing anything from the sun.* packages.
```
Instead of checking for `instanceof FileURLConnection`, "not `instanceof HttpURLConnection`" should be good enough, IMO

ocket8888: I was gonna say I'll check the protocol string in comparison to `file` (or `file:` or whatever it is for `FileURLConnection`s), but since `getLastModified` is a method of `URLConnection` I can't imagine what difference it would make.

On the other hand, the docs for `URLConnection.getLastModified` say
> Returns the value of the `last-modified` header field.

(Although I think I was looking at Java 8 for that). Does the `FileURIConnection` sub class implement that with modification timestamps on supported filesystems. What about on filesystems that don't store that information?

zrhoffman: > I was gonna say I'll check the protocol string in comparison to `file` (or `file:` or whatever it is for `FileURLConnection`s)

That sounds right, checking the protocol is how the URL class decides which handler to use.

> (Although I think I was looking at Java 8 for that). Does the `FileURIConnection` sub class implement that with modification timestamps on supported filesystems. What about on filesystems that don't store that information?

Maybe it returns 0 in those cases? When we don't have that info, we can skip that check. We use `useModifiedTimestamp()` to check this case for HTTP URLs. We can use it for File URLs too, right?

https://github.com/apache/trafficcontrol/blob/de41a02eb357de3caebb692804213776884f1174/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java#L387-L390

ocket8888: > We can use it for File URLs too, right?

¯\\\_(ツ)\_/¯ as far as I know, but that seems to be up to the implementation of that URL protocol subclass.

traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/loc/AbstractServiceUpdater.java
```diff
@@ -372,12 +373,14 @@ protected File downloadDatabase(final String url, final File existingDb) throws
 			if (conn instanceof HttpURLConnection) {
 				eTag = conn.getHeaderField("ETag");
 				if (((HttpURLConnection) conn).getResponseCode() == HttpURLConnection.HTTP_NOT_MODIFIED) {
-					LOGGER.info("[" + getClass().getSimpleName() + "] " + url + " not modified since our existing database's last update time of " + new Date(existingDb.lastModified()));
+					LOGGER.info("[" + getClass().getSimpleName() + "] " + url + " not modified since our existing database's last update time of " + new Date(existingLastModified));
 					return existingDb;
 				}
+			} else if (dbURL.getProtocol().equals("file") && conn.getLastModified() <= existingLastModified) {
```
zrhoffman, line 379: Should this condition should only succeed if the *last modified* time is greater than 0?

ocket8888: Yeah, that should be fine as long as nobody makes any closed time-like loops

zrhoffman (approved):
Looks good! This will help users who get their geolocation db from a different service, like their package manager.

PR #6580, opened 2022-02-23
Fix cache config generation for MIDs with a cache parent and HTTPS origin

rob05c:
Fixes cache config remap.config generation when a server is MID type
with other caches as parents in a Topology, and HTTPS origins.

The bug was inserting the Origin's URI verbatim in the remap target,
without changing HTTPS schemes to HTTP for the internal CDN request.

This bug existed for ages, and was fixed in most other scenarios,
but not this exact one.

Note the corresponding parent.config line correctly expects http/80. 
The bug is only in remap.config generation.

Includes tests.
Includes changelog.
No docs, no interface change.

- Traffic Control Cache Config (t3c, formerly ORT)

Run tests. Create a Topology with a MID-type server, with other caches as parents, and assign it to a Delivery Service with an HTTPS origin. Verify generated remap.config has an `http` target, not `https`.

Bug has existed forever.

## PR submission checklist
- [x] This PR has tests 
~- [x] This PR has documentation ~ No docs, no interface change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Fixed
 - Update traffic_portal dependencies to mitigate `npm audit` issues.
 - Fixed a cdn-in-a-box build issue when using `RHEL_VERSION=7`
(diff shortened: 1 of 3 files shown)


traeak (approved):
Changes look fine.

PR #6581, opened 2022-02-23
TMs peer with other TMs of the same status, not just ONLINE

rawlinp:
Instead of every TM peering with every other ONLINE TM, make TMs peer
with other TMs of the same status. I.e. ONLINE TMs will peer with other
ONLINE TMs (like normal), but OFFLINE TMs will peer with other OFFLINE
TMs (instead of peering w/ ONLINE TMs). The same goes for TMs of any
other status.

<hr/>

- Traffic Monitor

In an environment with at least 2 TMs, change their statuses to `OFFLINE`, then verify that they're peered with each other. One way to check this is to look at `/publish/PeerStates`, and check the `"peers"` map keys. Each key will be the hostname of a peer.

## PR submission checklist
- [x] This PR has tests 
- [x] no docs necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 - Added Rocky Linux 8 support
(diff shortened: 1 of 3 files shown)


mattjackson220 (approved):
Looks good! code looks good and tested locally and works great!

PR #6582, opened 2022-02-24
Fix stored DSRs cannot be unmarshaled after changes to DS JSON representation

ocket8888:
Fixes a compatibility issue with #6441 and pre-existing Delivery Service Requests, causing them to be un-retrieve-able from the Traffic Ops API due to unmarshaling failure.

<hr/>

- Traffic Ops (database only)

Create a Delivery Service Request at a point before #6441 was merged (actually creation afterward might still cause the issue now that I'm thinking of it, I haven't checked if that works), then upgrade TO to a version/ including #6441 and try to GET `/deliveryservice_requests`. Observe failure. Then run the migration in this PR and observe that it now works. Finally, downgrade the database and verify that it's broken again, then downgrade TO to a version that once again _doesn't_ include #6441 and verify that the old representation works again.

- master

## PR submission checklist
- [ ] This PR has tests


srijeet0406 (approved):
Code looks good, local tests seem to pass.

PR #6583, opened 2022-02-24
Refactor v3 cachegroups test

ericholguin:
This PR refactors [cachegroups_test.go](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/testing/api/v3/cachegroups_test.go)

Following the new test structure format that uses table driven tests, sub tests, GoLang's test runner, and also leveraging the existing assert functionality and expectation functions.

With these changes tests will be more granular and transparent on what is being tested. Ideally reducing the need to dig through test code to confirm what was tested. In addition this new test structure should also make it easier for new tests to be added as well as reduce redundant code.

**NOTE:** The v3 client varies from the v4 client and as such the refactor is slightly different

In addition all deprecated functions that were being used have been replaced by the non-deprecated functions that match the same functionality.

<hr/>

- Traffic Ops Testing

Run TO Integration tests

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/testing/api/v3/cachegroups_test.go
+++ b/traffic_ops/testing/api/v3/cachegroups_test.go
@@ -16,144 +16,241 @@ package v3
 */
 
 import (
-	"fmt"
+	"encoding/json"
 	"net/http"
 	"net/url"
-	"reflect"
+	"strconv"
 	"testing"
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-rfc"
 	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/assert"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/utils"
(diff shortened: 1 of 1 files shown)


traffic_ops/testing/api/v3/cachegroups_test.go
```diff
@@ -16,144 +16,241 @@ package v3
...
+		currentTimeRFC := currentTime.Format(time.RFC1123)
+
+		methodTests := map[string]map[string]struct {
+			endpointId     func() int
+			clientSession  *client.Session
+			requestParams  url.Values
+			requestHeaders http.Header
+			requestBody    map[string]interface{}
+			expectations   []utils.CkReqFunc
+		}{
```
zrhoffman, line 48: This anonymous struct is now in API v3 and API v4. It's fine for now, but if you think it will be used more times, it may be useful to declare it somewhere for reuse.

zrhoffman (approved):
The improvements from #6554 seem to have all made it in this backport. Looks good!

PR #6584, opened 2022-02-25
Fix TR NPE when topology cachegroup doesn't exist in crconfig

rawlinp:
Also, fix TrafficRouterStart logging configuration so that it actually
logs at the INFO-level again (this was likely a regression from the
conversion to log4j2).

<hr/>

- Traffic Router

In an environment with a topology that is assigned to at least 1 delivery service in your CDN, start TR, let it process the current crconfig, set all the caches in a particular EDGE cachegroup of the topology to OFFLINE, snapshot the CDN, then observe TR. Before this fix, TR would reject the crconfig due to encountering a `NullPointerException`. With this fix, TR processes the snapshot successfully.

To verify the TrafficRouterStart logging, run TrafficRouterStart (typically in an IDE) and verify that it logs to the console at the INFO level.

- 5.x
- 6.x

## PR submission checklist
- [x] This PR does not have tests


zrhoffman (approved):
Looks good!

PR #6587, opened 2022-02-25
Fix user Permissions calculated incorrectly/inconsistently

ocket8888:
Permissions also updated to reflect those actually laid out in the documentation (although TPv2 only uses stable API versions).

<hr/>

- Traffic Portal

Make sure the included tests pass. Start TPv2, log in as an "admin"-Role user, make sure you can do everything. Log in as a read-only user, make sure that the "New Delivery Service" button has disappeared from the dashboard. Log in as a user who doesn't have Permissions to see users and/or servers, then verify that the associated menu items in the header vanish appropriately.

- master

## PR submission checklist
- [x] This PR has tests


shamrickus (approved):
LGTM

PR #6593, opened 2022-03-01
Layered Server Feature for `GET` API endpoint `servers/details`

rimashah25:
This PR (Layered Server Feature for `GET` API endpoint `servers/details`) is not related to any Issue but is a development for feature https://github.com/apache/trafficcontrol/pull/6095

<hr/>

- Documentation
- Traffic Ops

Using CURL for 
API v4
```
curl -X GET -k -H "Accept: application/json" --cookie "mojolicious=<cookie>" https://localhost:8443/api/4.0/servers/details?hostName=abc
```
Result as follows: Note: Field profileNames exists and profileDesc does not exist
```
{"limit":1000,"orderby":"hostName","response":[{"cachegroup":"mb","cdnName":"ALL","domainName":"abc.com","guid":null,"hardwareInfo":null,"hostName":"abc","httpsPort":443,"id":9,"iloIpAddress":null,"iloIpGateway":null,"iloIpNetmask":null,"iloPassword":null,"iloUsername":null,"mgmtIpAddress":null,"mgmtIpGateway":null,"mgmtIpNetmask":null,"offlineReason":null,"physLocation":"boulder","**profileNames":["INFLUXDB","RIAK_ALL"]**,"rack":null,"status":"OFFLINE","tcpPort":80,"type":"RIAK","xmppId":"b9e71a88-48a5-4d3c-bf80-4bc58f9b5d75","xmppPasswd":null,"interfaces":[{"ipAddresses":[{"address":"8.8.8.8","gateway":"8.0.0.1","serviceAddress":true}],"maxBandwidth":null,"monitor":true,"mtu":null,"name":"int1","routerHostName":"","routerPortName":""}]}],"size":1}
```

API v3/v2
```
curl -X GET -k -H "Accept: application/json" --cookie "mojolicious=<cookie>" https://localhost:8443/api/3.0/servers/details?hostName=abc
```
Result as follows: Note: Field profile and profileDesc exists
```
{"limit":1000,"orderby":"hostName","response":[{"cachegroup":"mb","cdnName":"ALL","domainName":"abc.com","guid":null,"hardwareInfo":null,"hostName":"abc","httpsPort":443,"id":9,"iloIpAddress":null,"iloIpGateway":null,"iloIpNetmask":null,"iloPassword":null,"iloUsername":null,"mgmtIpAddress":null,"mgmtIpGateway":null,"mgmtIpNetmask":null,"offlineReason":null,"physLocation":"boulder","profile":"INFLUXDB","profileDesc":"InfluxDb profile","rack":null,"status":"OFFLINE","tcpPort":80,"type":"","xmppId":"b9e71a88-48a5-4d3c-bf80-4bc58f9b5d75","xmppPasswd":null,"interfaces":[{"ipAddresses":[{"address":"8.8.8.8","gateway":"8.0.0.1","serviceAddress":true}],"maxBandwidth":null,"monitor":true,"mtu":null,"name":"int1"}],"routerHostName":"","routerPortName":""}],"size":1}
```

## PR submission checklist
- [x] This PR has tests 


ocket8888:
This has a lot of overlap with #6544  - I'll review it, but I want to finish that one first.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -2091,6 +2091,43 @@ func UpdateServerProfileTableForV2V3(id *int, newProfile *string, origProfile st
 	return profileName, nil
 }
 
+// GetServerDetailFromV4 function converts server details from V4 to V3/V2
+func GetServerDetailFromV4(sd tc.ServerDetailV40, tx *sql.Tx) (tc.ServerDetail, error) {
+	var profileDesc *string
+	if err := tx.QueryRow(`SELECT p.description FROM profile p WHERE p.name=$1`, sd.ProfileNames[0]).Scan(&profileDesc); err != nil {
+		return tc.ServerDetail{}, fmt.Errorf("querying profile description by profile name: " + err.Error())
```
ocket8888, line 2098: errors built from errors should wrap with `%w`

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -2091,6 +2091,43 @@ func UpdateServerProfileTableForV2V3(id *int, newProfile *string, origProfile st
...
+		MgmtIPGateway:      sd.MgmtIPGateway,
+		MgmtIPNetmask:      sd.MgmtIPNetmask,
+		OfflineReason:      sd.OfflineReason,
+		PhysLocation:       sd.PhysLocation,
+		Profile:            &sd.ProfileNames[0],
+		ProfileDesc:        profileDesc,
+		Rack:               sd.Rack,
+		Status:             sd.Status,
+		TCPPort:            sd.TCPPort,
+		Type:               "",
```
ocket8888, line 2125: shouldn't `Type` be populated from `sd.Type` like everything else?

traffic_ops/traffic_ops_golang/server/detail.go
```diff
@@ -91,7 +91,11 @@ func GetDetailParamHandler(w http.ResponseWriter, r *http.Request) {
 				routerPortName = interfaces[0].RouterPortName
 			}
 			v11server := tc.ServerDetailV11{}
-			v11server.ServerDetail = server.ServerDetail
+			v11server.ServerDetail, err = dbhelpers.GetServerDetailFromV4(server, inf.Tx.Tx)
+			if err != nil {
+				api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, nil, errors.New("failed to GetServerDetailFromV4: "+err.Error()))
```
ocket8888, line 96: This error is being passed up as an internal error and nothing is sent to the client - this should probably be returning a `500 Internal Server Error` response not `400 Bad Request`.

Also errors constructed from errors should wrap their source error using the `%w` formatting parameter of `fmt.Errorf`.

traffic_ops/traffic_ops_golang/server/detail.go
```diff
@@ -117,7 +121,11 @@ func GetDetailParamHandler(w http.ResponseWriter, r *http.Request) {
 				routerHostName = interfaces[0].RouterHostName
 				routerPortName = interfaces[0].RouterPortName
 			}
-			v3Server.ServerDetail = server.ServerDetail
+			v3Server.ServerDetail, err = dbhelpers.GetServerDetailFromV4(server, inf.Tx.Tx)
+			if err != nil {
+				api.HandleErr(w, r, inf.Tx.Tx, http.StatusBadRequest, nil, errors.New("failed to GetServerDetailFromV4: "+err.Error()))
```
ocket8888, line 126: same as above RE: response code and error wrapping

traffic_ops/traffic_ops_golang/server/detail_test.go
```diff
@@ -93,8 +94,6 @@ func TestGetDetailServers(t *testing.T) {
 	serviceNetmask := util.StrPtr("")
 	serviceInterface := util.StrPtr("")
 	serviceMtu := util.StrPtr("")
-	//routerHostName := util.StrPtr("")
-	//routerPort := util.StrPtr("")
```
ocket8888, line 97: why were these commented out?

lib/go-tc/servers.go
```diff
@@ -87,8 +87,33 @@ type ServerDetailV30 struct {
...
 // ServerDetailV40 is the details for a server for API v4.
 type ServerDetailV40 struct {
-	ServerDetail
-	ServerInterfaces []ServerInterfaceInfoV40 `json:"interfaces"`
+	CacheGroup         *string                  `json:"cachegroup" db:"cachegroup"`
+	CDNName            *string                  `json:"cdnName" db:"cdn_name"`
+	DeliveryServiceIDs []int64                  `json:"deliveryservices,omitempty"`
+	DomainName         *string                  `json:"domainName" db:"domain_name"`
+	GUID               *string                  `json:"guid" db:"guid"`
+	HardwareInfo       map[string]string        `json:"hardwareInfo"`
```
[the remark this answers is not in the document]
rimashah25: Sure.

traffic_ops/traffic_ops_golang/server/detail_test.go
rimashah25: Because both these fields are not part of [ServerDetail](https://github.com/apache/trafficcontrol/blob/master/lib/go-tc/servers.go#L1304){}

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rimashah25: ugh, missed it.

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rimashah25: It should be, wonder why I didn't use it. I went back in history (wrt this file) but seems I always had `Type: ""`. I have changed it now.

traffic_ops/traffic_ops_golang/server/detail.go
rimashah25: @ocket8888: I don't understand, why I need to use `fmt.Errorf` instead of `errors.New`. In my understanding fmt gives me a string format and errors gives me error type.
I agree on the status code change.

lib/go-tc/servers.go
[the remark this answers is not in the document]
ocket8888: actually that complicates things. I was only always seeing null because my TO instance isn't years old. If you created HWInfo structures back in APIv1, then APIv3's `servers/details` was still showing them. So that's probably something we'd need to take to the mailing list. For all I know that's the entire reason we still have this route.

So probably don't worry about that in this PR.

traffic_ops/traffic_ops_golang/server/detail.go
ocket8888: `fmt.Errorf` returns an `error` type, not a `string`. The reason is because when you make a new error by calling `err.Error()`, you're discarding the error and only carrying its string representation forward.

In this playground I constructed the "same" error 5 different ways: https://go.dev/play/p/oqjjeO4DcjA
When you run it, you can see that the identity of the underlying error is only preserved in the first case - using `fmt.Errorf` with `%w`.

traffic_ops/traffic_ops_golang/server/detail_test.go
ocket8888: Alright, that checks out. They should probably just be removed instead of merely commented-out, in that case.

lib/go-tc/servers.go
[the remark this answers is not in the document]
rimashah25: I see.. thanks for the explanation.

traffic_ops/traffic_ops_golang/server/detail.go
rimashah25: I see.. thanks for the explanation.

ocket8888 (approved):
LGTM

PR #6601, opened 2022-03-01
T3c cookie changes

jpappa200:
This PR fixes some nitpicks from the ssl checking PR and also does a better job of detecting an empty cookie and forcing a user/password login.

<hr/>

build and test on an admin_down cache. Edit JSON cookie file so it will not match the current structure but is still valid JSON and see that it skips the cookie and overwrites it. 

Not a bug fix

## PR submission checklist
- [x] This PR has tests 


rob05c (approved):
LGTM. Tested, cookie is still created and used as expected

Issue #6609, opened 2022-03-03
Missing Permissions in DB seed/migration


shamrickus:
## This Bug Report affects these Traffic Control components:

- Traffic Ops

## Current behavior:
The following permissions are not seeded in the database:
```
CDNI-CAPACITY:UPDATE (OC/CI/configuration/, OC/CI/configuration/{host})
CDNI-CAPACITY:ADMIN (OC/CI/configuration/request/{id}/{approved})

SSL-KEY-EXPIRATION:READ (sslkey_expirations/)

ACME:READ (acme_accounts/providers)
ACME:READ (deliveryservices/sslkeys/generate/acme/)
ACME:READ/CREATE/UPDATE (acme_accounts/)
ACME:READ/DELETE (acme_accounts/{provider}/{email})
ACME:READ (acme_autorenew/)

DBDUMP:READ (dbdump/?)

ROLE:UPDATE/CREATE/DELETE (roles/)

DS-SECURITY-KEY:CREATE/UPDATE/DELETE (deliveryservices/{xmlID}/urisignkeys$,
  deliveryservices/sslkeys/add$,
  deliveryservices/xmlId/{xmlid}/sslkeys$,
  deliveryservices/sslkeys/generate/,
  deliveryservices/xmlId/{name}/urlkeys/copyFromXmlId/{copy-name}/,
  deliveryservices/xmlId/{name}/urlkeys/generate/,
  deliveryservices/xmlId/{name}/urlkeys/,
  deliveryservices/{id}/urlkeys/,
  deliveryservices/sslkeys/generate/letsencrypt/)
  letsencrypt/autorenew/)

PLUGIN:READ (plugins/)
```

## Expected behavior:
At least the admin user should have all of these permissions, potentially other users too.

## Steps to reproduce:
Run the db setup and look in the `role_capability` table and see that these permissions do not exist in there.

ocket8888:
* `CDNI-CAPACITY:UPDATE`
* `CDNI-CAPACITY:ADMIN`
* `CDNI-CAPACITY:READ`

This didn't already exist when Permissions were updated. Since priv_level still exists and the routes requiring these Permissions ostensibly have a required priv_level that was intentionally set, we _can_ seed these, but that's not strictly necessary, IMO, since that's not something we'd do for a new concept like this once priv_level is removed.

[...]

Issue #6603, opened 2022-03-02
Missing Permissions in DB seed/migration


shamrickus:
## This Bug Report affects these Traffic Control components:

- Traffic Ops

## Current behavior:
Currently the `DNS-SEC:READ` and `DNS-SEC:DELETE` permissions are not seeded in the db. As a result non admin users will never have the permissions to query `GET/DELETE: cdns/name/{name}/dnsseckeys`

## Expected behavior:
Those permissions should be created and appropriately assigned to the correct existing roles.

## Steps to reproduce:
Create a user whose `role` has the following `role_capabiltity`: `CDN:READ, DELIVERY-SERVICE:READ, DELIVERY-SERVICE:UPDATE, CDN:UPDATE` and see that the user is not able to query the `dnsseckeys` endpoints.

PR #6610, opened 2022-03-03
Add Test to Catch Missing Permissions

shamrickus:
This PR adds a test that checks that at least one user (admin) has all the permissions for the TO API routes. It will catch #6603 and #6609 with a test failure.

Note: until #6609 is fixed, the TO API tests on this PR will fail.
<hr/>

- Traffic Ops

Run the TO API tests and verify they pass.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/testing/api/v4/roles_test.go
+++ b/traffic_ops/testing/api/v4/roles_test.go
@@ -16,6 +16,7 @@ package v4
 */
 
 import (
+	"fmt"
 	"net/http"
 	"net/url"
 	"reflect"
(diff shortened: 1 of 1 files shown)


ocket8888:
The definition of the "admin" Role is that it has all Permissions past, present, and future. [The set of Permissions afforded to the "admin" Role isn't even checked by Traffic Ops](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/traffic_ops_golang/auth/authorize.go#L53), so this is testing unintended behavior.

For example: The year is 2024, and I'm adding Foos as a new object to the Traffic Ops API. Foos are not related to any existing objects, and don't affect them in any way. Which users should be granted the `FOO:READ` and `FOO:CREATE` Permissions?

The intended answer is "none" and an "admin"-Role user will be responsible for doling out those new Permissions to whatever Roles need them. That's why we enforce an "admin" Role must exist, and why we treat it specially the way we do.

PR #6617, opened 2022-03-04
Fix fail log tests not being run

ocket8888:
This PR fixes the "fail log" tests for the behavior added in #6586 not being run, because the new file isn't a Go test file.

<hr/>

Run the "ORT tests", make sure the new test is run and passes. (You'll need to run with `-v` to ensure it passes and/or use `--run '^TestT3cApplyFailMsg$'`)

- master

## PR submission checklist
- [x] This PR has tests


ocket8888:
@rob05c I *think* the fail log tests are failing because it's actually not logging the failure line, not due to a problem with the tests. According to the failing test, the output of `t3c-apply` is empty, and looking at the code it seems like it ought to be logging the error message to stdout. Should I try to fix `t3c-apply` in this PR, or do you wanna do that and just keep this to being a PR that fixes the test not running? Or am I wrong and the problem is in the test?

rob05c (approved):
Looks good, fixes the test not being run, thanks!

PR #6618, opened 2022-03-04
Cleanup parent.config generation code

rob05c:
Various techdebt cleanup of parent.config generation, inherited from ancient transliterated Perl. 

Generated config should be byte-for-byte identical with or without this PR.

- combines `profileCache` and `originServers` into a single `serversWithParams` variable, since the params apply to and are only ever useful with their corresponding server
- renames `profileCache` struct to `parentServerParams`, which is what it is, and matches the corresponding `parentDSParams`.  The `profileCache` name was inherited from ancient Perl code
- removes unused `parentConfigDS*` symbols
- removes `cgServer` struct and replaces usage with real Servers; also inherited from ancient Perl code

Run tests. Generate config, verify config is unchanged.

Not a bug fix.

## PR submission checklist
- ~[x] This PR has tests~ tests exist for all affected code 
- ~[x] This PR has a CHANGELOG.md entry~ no interface change, no behavioral change 

### DIFF
--- a/lib/go-atscfg/parentdotconfig.go
+++ b/lib/go-atscfg/parentdotconfig.go
@@ -307,13 +307,13 @@ func makeParentDotConfigData(
 		parentServerDSes[dss.Server][dss.DeliveryService] = struct{}{}
 	}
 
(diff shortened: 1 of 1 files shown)


jpappa200 (approved):
👍  LGTM generated config and config was unchanged.

PR #6621, opened 2022-03-05
Refactor about, asns, and acme tests

ericholguin:
This PR refactors the **v3/about**, **v4/about**, **v3/asns**, **v4/asns** , and the **v4/acme** tests
Following the new test structure format that uses table driven tests, sub tests, GoLang's test runner, and also leveraging the existing assert functionality and expectation functions.

With these changes tests will be more granular and transparent on what is being tested. Ideally reducing the need to dig through test code to confirm what was tested. In addition this new test structure should also make it easier for new tests to be added as well as reduce redundant code.

<hr/>

- Traffic Ops Testing

Run TO Integration tests

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/testing/api/v3/about_test.go
+++ b/traffic_ops/testing/api/v3/about_test.go
@@ -16,18 +16,42 @@ package v3
 */
 
 import (
+	"net/http"
 	"testing"
+
+	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/utils"
+	client "github.com/apache/trafficcontrol/traffic_ops/v3-client"
 )
 
 func TestAbout(t *testing.T) {
-	m, _, err := TOSession.GetAbout()
-	if err != nil {
-		t.Errorf("error from GetAbout(): %v", err)
-	}
-	t.Logf("about: %v", m)
 
-	m, _, err = NoAuthTOSession.GetAbout()
-	if err == nil {
-		t.Error("expected error from GetAbout() when unauthenticated")
+	methodTests := map[string]map[string]struct {
+		clientSession *client.Session
+		expectations  []utils.CkReqFunc
+	}{
+		"GET": {
+			"OK when VALID request": {
+				clientSession: TOSession, expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK)),
+			},
+			"UNAUTHORIZED when NOT LOGGED IN": {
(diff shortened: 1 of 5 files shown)


zrhoffman (changes requested):
Between #6621, #6583, and #6554, there are 7 new anonymous structs added for testcases:

https://github.com/apache/trafficcontrol/blob/c9fa2b1dfa5fedd84cabe40ec94d2ac0178f8799/traffic_ops/testing/api/v3/about_test.go#L29-L32

https://github.com/apache/trafficcontrol/blob/c9fa2b1dfa5fedd84cabe40ec94d2ac0178f8799/traffic_ops/testing/api/v3/asns_test.go#L42-L49

https://github.com/apache/trafficcontrol/blob/8a6e744b086dc5d666707e302f3e89f712c39d42/traffic_ops/testing/api/v3/cachegroups_test.go#L41-L48

https://github.com/apache/trafficcontrol/blob/c9fa2b1dfa5fedd84cabe40ec94d2ac0178f8799/traffic_ops/testing/api/v4/about_test.go#L29-L33

https://github.com/apache/trafficcontrol/blob/c9fa2b1dfa5fedd84cabe40ec94d2ac0178f8799/traffic_ops/testing/api/v4/acme_test.go#L28-L32

https://github.com/apache/trafficcontrol/blob/c9fa2b1dfa5fedd84cabe40ec94d2ac0178f8799/traffic_ops/testing/api/v4/asns_test.go#L41-L47

https://github.com/apache/trafficcontrol/blob/c9fa2b1dfa5fedd84cabe40ec94d2ac0178f8799/traffic_ops/testing/api/v4/cachegroups_test.go#L42-L48

It looks like the `requestHeaders` and `questParams` fields could be expressed as a `requestOpts` field. Is there a new struct we could reuse in place of these anonymous structs?

ericholguin:
> It looks like the `requestHeaders` and `questParams` fields could be expressed as a `requestOpts` field. Is there a new struct we could reuse in place of these anonymous structs?

Unfortunately the v3 client doesn't have the `RequestOptions` struct. Do you think recreating it for v3 is the way to go?

zrhoffman:
> Unfortunately the v3 client doesn't have the `RequestOptions` struct. Do you think recreating it for v3 is the way to go?

One struct per API version would make sense, since it needs to refer to a specific client version. And good point, `requestOptions` cannot be used in API v3.

zrhoffman (approved):
Code looks good, tests pass, and test output is much cleaner!

PR #6622, opened 2022-03-07
Added Client Certificate Auth blueprint.

alficles:
This blueprint, if implemented, will allow users and systems to authenticate
with TLS certificates in addition to passwords. This will allow implementations
considerable flexibility in how they handle passwordless logins.

<hr/>

- Documentation

This is a blueprint.

No bug.

## PR submission checklist
- [ ] This PR has tests 

### DIFF
--- a/blueprints/client-certificate-auth.md
+++ b/blueprints/client-certificate-auth.md
@@ -1,0 +1,240 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+<Paste>"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Client Certificate Authentication
+
+## Problem Description
+
+Passwords currently serve as a powerful, but inflexible, way for clients to
+authenticate with Traffic Ops. However, an account can only have one password
+at a time, which means that service accounts tend to have a single shared
+password. This is undesirable from a security perspective.
+
+Client certificates are a flexible tool that allows operators to assign
+independent credentials to users of a service account. These credentials can
+have varying expirations and be issued in accordance with the operator's
+security policy.
+
+## Proposed Change
+
+When Traffic Portal or Traffic Ops receives a request, in addition to accepting
+a valid token, it will accept a valid TLS certificate chain. Traffic Ops will
+have a list of root certificates that it accepts.
+
+Issuance and management of these certificates will occur outside of Traffic
+Control and is the responsibility of the operator.
+
+### Traffic Portal Impact
+
+Traffic Portal will pull user information from requests before they are
+proxied. Connections between TP and TO will be signed with a client certificate
+to protect the connection.
+
+When TP proxies a connection to TO, it validates the TLS certificate against
+it's list of authorized roots. Iff it is valid, it includes the Subject of the
+certificate verbatim in a request header named `Client-Cert-Subject`.
+Otherwise, it strips any header named `Client-Cert-Subject` from the request.
+
+Whether the certificate is valid or not, it removes any existing
+`Client-Cert-Public-Key` header and adds a `Client-Cert-Public-Key` with the
+base64-encoded public key of the client certificate.
+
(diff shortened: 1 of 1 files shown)


blueprints/client-certificate-auth.md
```diff
@@ -0,0 +1,240 @@
...
+
+A certificate may not contain multiple UID RDNs. If it does, it is invalid.
+
+In all log messages that contain a user's identity, the CN of the Subject of
+the certificate and the IP address of connection should be logged as well. If
+the user id is matched in the `Client-Cert-Subject` on account of an empty UID
+in the Subject of the certificate, the CN of the Client-Cert-Subject and the IP
+address from the Forwarded header should be logged as well.
+
+An invalid certificate should be processed as though it were not present. So a
```
TaylorCFrey, line 96: Is there an order of precedence if both are provided? If authenticating with a valid and correct cerficate works, is that sufficient? If the client is providing user/pass credentials in addition to a certificate do we check both? And, if we check both, do we care if there is a conflict? Say the certificate works but the user/pass would fail?

blueprints/client-certificate-auth.md
```diff
@@ -0,0 +1,240 @@
...
+and include the UID in the cache key.
+
+### Developer Impact
+
+Developers with test systems may find it's easier to develop on test systems
+using certificates instead of passwords. It will present an additional option
+at least. Existing workflows will work just fine, though.
+
+### Alternatives
+
```
TaylorCFrey, line 221: I've only recently become aware of it, but is [SQRL](https://www.grc.com/sqrl/sqrl.htm) an alternative as well?

TaylorCFrey (commented):
Some initial questions

blueprints/client-certificate-auth.md
alficles: Any valid credential is valid. That is part of what I mean when I say "processed as though it were not present". If the credential is invalid, you act like it was never presented and do whatever you would have done if it weren't there.

blueprints/client-certificate-auth.md
alficles: SQRL relies on a consistent private key, similar to mTLS. While it could form the basis of a protocol, everything already supports mTLS and very little supports SQRL. Likewise, SQRL doesn't have protection for things like replay, expiration, third-party attestation and delegation the way mTLS does. This is fine, because SQRL is trying to solve a very different problem. It's not really a solid option here, though. I guess I could add a section on SQRL, but it doesn't really offer what we're looking for here.

blueprints/client-certificate-auth.md
```diff
@@ -0,0 +1,240 @@
...
+### Traffic Portal Impact
+
+Traffic Portal will pull user information from requests before they are
+proxied. Connections between TP and TO will be signed with a client certificate
+to protect the connection.
+
+When TP proxies a connection to TO, it validates the TLS certificate against
+it's list of authorized roots. Iff it is valid, it includes the Subject of the
+certificate verbatim in a request header named `Client-Cert-Subject`.
```
TaylorCFrey, line 51: Are these standard headers (`Client-Cert-Subject` `Client-Cert-Public-Key` etc)? My brief searching for `Client-Cert` headers shows the IETF draft https://www.ietf.org/id/draft-ietf-httpbis-client-cert-field-02.html.

zrhoffman (changes requested):
How you visualize this working or being used is difficult to conceptualize without a visualization. Would you please add a diagram?

blueprints/client-certificate-auth.md
alficles: These are not standard headers. I started with the actual Client-Cert draft, but a) we don't honestly need the whole cert in there and b) the draft isn't stable yet. When the draft stabilizes or is published, we can consider whether it makes sense to switch, but we don't really want to wait for that. In the mean time, this approach seems sensible and unlikely to conflict.

blueprints/client-certificate-auth.md
```diff
@@ -0,0 +1,240 @@
...
+issued an empty-UID client certificate in order to proxy authentication. It
+will also need to validate presented certs against the authorized roots, copy
+the Subject into the `Client-Cert-Subject` header, add the Forwarded header,
+and include the UID in the cache key.
+
+### Developer Impact
+
+Developers with test systems may find it's easier to develop on test systems
+using certificates instead of passwords. It will present an additional option
+at least. Existing workflows will work just fine, though.
```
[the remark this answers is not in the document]
alficles: Are you looking for a diagram regarding existing workflows or new workflows? This blueprint doesn't modify existing authentication methods.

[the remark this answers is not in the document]
zrhoffman: >>> Existing authentication workflows will work just fine, though.

>>or giving examples of existing authentication methods that will continue to work.

>Are you looking for a diagram regarding existing workflows or new workflows? This blueprint doesn't modify existing authentication methods.

This comment is not related to the diagram request from a separate comment. I gave an example of a way to reword it to make that sentence less vague

zrhoffman (approved):
Let's add the diagram later. Looks good!

Issue #6512, opened 2022-01-13
Reduce Traffic Monitor dependency on CRConfig data


rawlinp:
## This Improvement request (usability, performance, tech debt, etc.) affects these Traffic Control components:
- Traffic Monitor

## Current behavior:
As of https://github.com/apache/trafficcontrol/pull/6443, TM will no longer overwrite its monitoring.json snapshot data with corresponding CRConfig snapshot data if the monitoring.json contains IP addresses for the TMs, meaning it now has parity with the CRConfig data. However, it still needs the CRConfig in order to filter out delivery services from monitoring.json which aren't present in the CRConfig. That was fixed in #6443 in that the monitoring.json no longer contains delivery services from other CDNs to which the TM doesn't belong.

## New behavior:
Once ATC 6.1 is released with the changes from #6443, we should be able to remove this fallback behavior from TM (and possibly its reliance on the CRConfig altogether) in the next major/minor release. That is because monitoring.json snapshots as of TO 6.1 will contain all the necessary data and will no longer need to rely on the CRConfig snapshot.

mitchell852:
@rawlinp any idea what the impact and effort of this improvement is?

Issue #1739, opened 2018-01-09
Traffic Monitor CRConfig and monitoring.json polls aren't synchronized


rob05c:
The Monitor has two separate polls, to get new `CRConfig`, and to get new `monitoring.json`. 

Even if you make your changes and snapshot immediately, the Monitor might poll the CRConfig, get those updates, and it might be 5 or 30 seconds later before it polls `monitoring.json` and gets the updates from there. 

For example, if those changes include renaming a Profile, and the `monitoring.json` is polled first, the Monitor will be missing the old profile and be unable to poll caches with that profile, until it polls the CRConfig again and gets the change.

It’s a bug, the Monitor needs to not apply changes from one until it gets the other. Whoever fixes this also needs to determine which should be first (until #1738 is fixed, which will make them the same [I'd suggest fixing #1738 first, but it's likely considerably more work than this]). I _think_ the CRConfig should be first, but I may be wrong, and in fact, there may be no right answer, #1738 may have to be fixed first.

This is related to but distinct from #1738 . Symptoms will often be the result of some combination of both bugs.

rob05c:
I'm assigning myself, but I likely won't have time soon, if anyone else wants to fix this, feel free to take it from me.

mitchell852:
@rob05c - can you put a severity label on this?

rob05c:
**Update**

We have a PR to fix #1738 - PR #3104. But that PR does **not** fix this bug.
The two bugs are:
1. the `monitoring.json` data is not snapshotted
2. the Traffic Monitor doesn't ensure the versions of `monitoring.json` and `CRConfig` match.

This case is for number 2.

**How to fix**

The code needs changed to request both the CRConfig and monitoring.json immediately one after the other. But that doesn't completely fix it, because there's still a small race, if someone were to Snapshot in-between the Monitor's requests.

So, the requests actually need to happen in a loop, where each are requested, their timestamps are compared, and if one is older, it's fetched again, in a loop, until both timestamps match.

The pseudocode looks something like:
```
crConfig  = GetCRConfig()
monitoring = GetMonitoring()
for {
  if crConfig.SnapshotTime < monitoring.SnapshotTime {
    crConfig = GetCRConfig()
    continue
  }
  if monitoring.SnapshotTime < crConfig.SnapshotTime {
    monitoring = GetMonitoring()
    continue
  }
  break
}
```

mitchell852:
@rob05c - can this be closed?

[...]

PR #6625, opened 2022-03-09
Reduce TM dependency on CRConfig

rawlinp:
Do not use CRConfig data when the same data already exists in the
TMConfig. For data that TM needs but doesn't exist in TMConfig, add it
to TMConfig. This way, TM only uses data in the TMConfig and can become
simply a shuttle for the CRConfig without actually using any CRConfig
data.

Closes: #6512
Related: #1739
Related: #5367

<hr/>

- Documentation
- Traffic Monitor
- Traffic Ops

Ensure the GHA tests pass. Take a snapshot of a CDN before this change, then take a snapshot after this change and compare the TMConfig snapshot to see the newly added fields. Compare the newly added fields to the corresponding data in the CRConfig snapshot (they should be the same). Run TM with the new snapshot and ensure it still polls and functions normally.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/docs/source/api/v2/cdns_name_configs_monitoring.rst
+++ b/docs/source/api/v2/cdns_name_configs_monitoring.rst
@@ -58,9 +58,12 @@ Response Structure
 
 :deliveryServices: An array of objects representing each :term:`Delivery Service` provided by this CDN
 
+	:hostRegexes:        An array of strings which are the Delivery Service's HOST_REGEXP-type regexes
 	:status:             The :term:`Delivery Service`'s status
+	:topology:           A string that is the name of the Delivery Service's topology (if assigned one)
 	:totalKbpsThreshold: A threshold rate of data transfer this :term:`Delivery Service` is configured to handle, in Kilobits per second
 	:totalTpsThreshold:  A threshold amount of transactions per second that this :term:`Delivery Service` is configured to handle
+	:type:               A string that is the Delivery Service's type category (HTTP or DNS)
 	:xmlId:              A string that is the :ref:`Delivery Service's XMLID <ds-xmlid>`
 
 :profiles: An array of the :term:`Profiles` in use by the :term:`cache servers` and :term:`Delivery Services` belonging to this CDN
@@ -80,29 +83,33 @@ Response Structure
 
 	:type: A string that names the :ref:`Profile's Type <profile-type>`
 
+:topologies: A map of :term:`Topology` names to objects
+
+	:nodes: An array of strings which are the names of the EDGE_LOC-type cache groups in the topology
+
 :trafficMonitors: An array of objects representing each Traffic Monitor that monitors this CDN (this is used by Traffic Monitor's "peer polling" function)
 
-	:fqdn:     An :abbr:`FQDN (Fully Qualified Domain Name)` that resolves to the IPv4 (and/or IPv6) address of the server running this Traffic Monitor instance
-	:hostname: The hostname of the server running this Traffic Monitor instance
-	:ip6:      The IPv6 address of this Traffic Monitor - when applicable
-	:ip:       The IPv4 address of this Traffic Monitor
-	:port:     The port on which this Traffic Monitor listens for incoming connections
-	:profile:  A string that is the :ref:`profile-name` of the :term:`Profile` assigned to this Traffic Monitor
-	:status:   The status of the server running this Traffic Monitor instance
+	:cachegroup: The :term:`Cache Group` to which this Traffic Monitor belongs
+	:fqdn:       An :abbr:`FQDN (Fully Qualified Domain Name)` that resolves to the IPv4 (and/or IPv6) address of the server running this Traffic Monitor instance
+	:hostname:   The hostname of the server running this Traffic Monitor instance
+	:ip6:        The IPv6 address of this Traffic Monitor - when applicable
+	:ip:         The IPv4 address of this Traffic Monitor
+	:port:       The port on which this Traffic Monitor listens for incoming connections
+	:profile:    A string that is the :ref:`profile-name` of the :term:`Profile` assigned to this Traffic Monitor
+	:status:     The status of the server running this Traffic Monitor instance
 
 :trafficServers: An array of objects that represent the :term:`cache servers` being monitored within this CDN
 
-	:cacheGroup:    The :term:`Cache Group` to which this :term:`cache server` belongs
-	:fqdn:          An :abbr:`FQDN (Fully Qualified Domain Name)` that resolves to the :term:`cache server`'s IPv4 (or IPv6) address
-	:hashId:        The (short) hostname for the :term:`cache server` - named "hashId" for legacy reasons
-	:hostName:      The (short) hostname of the :term:`cache server`
-	:interfacename: The name of the network interface device being used by the :term:`cache server`'s HTTP proxy
-	:ip6:           The :term:`cache server`'s IPv6 address - when applicable
-	:ip:            The :term:`cache server`'s IPv4 address
-	:port:          The port on which the :term:`cache server` listens for incoming connections
-	:profile:       A string that is the :ref:`profile-name` of the :term:`Profile` assigned to this :term:`cache server`
-	:status:        The status of the :term:`cache server`
-	:type:          A string that names the :term:`Type` of the :term:`cache server` - should (ideally) be either ``EDGE`` or ``MID``
+	:cacheGroup:       The :term:`Cache Group` to which this :term:`cache server` belongs
+	:deliveryServices: An array of strings which are the XML IDs of the delivery services to which this cache server is assigned
+	:fqdn:             An :abbr:`FQDN (Fully Qualified Domain Name)` that resolves to the :term:`cache server`'s IPv4 (or IPv6) address
+	:hashId:           The (short) hostname for the :term:`cache server` - named "hashId" for legacy reasons
+	:hostName:         The (short) hostname of the :term:`cache server`
+	:port:             The port on which the :term:`cache server` listens for incoming connections
+	:profile:          A string that is the :ref:`profile-name` of the :term:`Profile` assigned to this :term:`cache server`
+	:status:           The status of the :term:`cache server`
+	:type:             A string that names the :term:`Type` of the :term:`cache server` - should (ideally) be either ``EDGE`` or ``MID``
+	:interfaces:       A set of the network interfaces in use by the server. In most scenarios, only one will be present, but it is illegal for this set to be an empty collection.
 
 .. code-block:: http
 	:caption: Response Example
@@ -120,6 +127,13 @@ Response Structure
 	Transfer-Encoding: chunked
 
 	{ "response": {
+	    "topologies": {
+	        "example-topology": {
+	            "nodes": [
+	                "CDN_in_a_Box_Edge"
+	            ]
+	        }
+	    },
 		"trafficServers": [
 			{
 				"profile": "ATS_EDGE_TIER_CACHE",
@@ -127,7 +141,7 @@ Response Structure
 				"ip": "172.16.239.100",
 				"ip6": "fc01:9400:1000:8::100",
 				"port": 80,
-				"cachegroup": "CDN_in_a_Box_Edge",
+				"cacheGroup": "CDN_in_a_Box_Edge",
 				"hostname": "edge",
 				"fqdn": "edge.infra.ciab.test",
 				"interfacename": "eth0",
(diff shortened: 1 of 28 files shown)


rawlinp:
Rebased to resolve changelog merge conflicts.

lib/go-tc/traffic_router.go
```diff
@@ -342,7 +249,6 @@ func (ts *TrafficServer) IPv6() string {
 	return *lid.IP6Address
 }
 
-type tsdeliveryService struct {
-	Xmlid  string   `json:"xmlId"`
-	Remaps []string `json:"remaps"`
+type TSDeliveryService struct {
```
ocket8888, line 252: GoDoc on this type?

traffic_monitor/todata/todata.go
```diff
@@ -162,44 +154,71 @@ func (d TODataThreadsafe) Update(to towrap.TrafficOpsSessionThreadsafe, cdn stri
...
-	}
+	// TODO: remove the fallback behavior on the CRConfig in ATC 8.0 (https://github.com/apache/trafficcontrol/issues/6627)
+	newTOData.DeliveryServiceServers, newTOData.ServerDeliveryServices = getDeliveryServiceServers(crConfig, mc)
 
-	newTOData.DeliveryServiceTypes, err = getDeliveryServiceTypes(crConfig)
+	// TODO: remove the fallback behavior on the CRConfig in ATC 8.0 (https://github.com/apache/trafficcontrol/issues/6627)
+	newTOData.DeliveryServiceTypes, err = getDeliveryServiceTypes(crConfig, mc)
 	if err != nil {
-		return fmt.Errorf("Error getting delivery service types from Traffic Ops: %v\n", err)
+		return fmt.Errorf("getting delivery service types from Traffic Ops: %v\n", err)
```
ocket8888, line 166: Can you remove the newlines in these, too? In TO I'd also suggest error-wrapping, but that can be very slow and TM needs all the performance it can get.

traffic_monitor/todata/todata.go
```diff
@@ -222,20 +241,34 @@ func getDeliveryServiceServers(crc CRConfig) (map[tc.DeliveryServiceName][]tc.Ca
...
 // getDeliveryServiceRegexes gets the regexes of each delivery service, for the given CDN, from Traffic Ops.
 // Returns a map[deliveryService][]regex.
-func getDeliveryServiceRegexes(crc CRConfig) (Regexes, error) {
+func getDeliveryServiceRegexes(crc CRConfig, mc tc.TrafficMonitorConfigMap) (Regexes, error) {
 	dsRegexes := map[tc.DeliveryServiceName][]string{}
 
+	if canUseMonitorConfig(mc) {
+		for dsName, dsData := range mc.DeliveryService {
+			if len(dsData.HostRegexes) < 1 {
+				log.Warnln("TMConfig missing regex for delivery service " + string(dsName))
```
ocket8888, line 255: Nit but as a `fmt.Stringer`, `DeliveryServiceName` values can be coerced automatically to strings - all you have to do is get rid of the manual cast and exchange ` " + ` for `, ` i.e.
```go
log.Warnln("TMConfig missing regex for delivery service " + string(dsName))
```
is equivalent to
```go
log.Warnln("TMConfig missing regex for Delivery Service", dsName)
```

traffic_ops/traffic_ops_golang/crconfig/servers.go
```diff
@@ -361,10 +322,10 @@ order by dsr.set_number asc
 	serverDSPatterns := map[tc.CacheName]map[string][]string{}
 	for server, dses := range serverDSNames {
 		for _, dsName := range dses {
-			dsInfList, ok := dsInfs[string(dsName)]
+			dsInfList, ok := dsInfs[dsName]
 			if !ok {
-				if !dsHasTopology[string(dsName)] {
-					log.Warnln("Creating CRConfig: deliveryservice " + string(dsName) + " has no regexes, skipping")
+				if !dsHasTopology[dsName] {
+					log.Warnln("creating CRConfig: deliveryservice " + dsName + " has no regexes, skipping")
```
ocket8888, line 328: why lowercase this?

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
```diff
@@ -543,6 +543,39 @@ func GetDSTenantIDFromXMLID(tx *sql.Tx, xmlid string) (int, bool, error) {
 	return id, true, nil
 }
 
+func GetServerDSNamesByCDN(tx *sql.Tx, cdn string) (map[tc.CacheName][]string, error) {
```
ocket8888, line 546: GoDoc?

traffic_ops/traffic_ops_golang/monitoring/monitoring.go
```diff
@@ -158,6 +164,10 @@ func GetMonitoringJSON(tx *sql.Tx, cdnName string) (*Monitoring, error) {
 	if err != nil {
 		return nil, fmt.Errorf("error getting config: %v", err)
 	}
+	topologies, err := topology.MakeTopologies(tx)
+	if err != nil {
+		return nil, fmt.Errorf("getting topologies: %s", err.Error())
```
ocket8888, line 169: formatting errors should wrap with `%w`

ocket8888 (commented):
I only made docs comments for the first version I encountered to avoid being overly repetitive. On that note: shouldn't these changes be made only to APIv4? I know the Snapshots have historically been unversioned, and "monitoring dot JSON" has always been a quasi-part of that snapshot, but it's used in TO Go client call signatures, so won't changing that be a breaking compile-time change?

traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
rawlinp: Done.

traffic_ops/traffic_ops_golang/monitoring/monitoring.go
rawlinp: Done

rawlinp (commented):
> shouldn't these changes be made only to APIv4?

I don't see why it can't be made to all versions, since this is non-breaking to the expected JSON formatting because it only adds new fields and doesn't change data types or rename existing fields.

As far as the TO client goes, this _does_ technically make a breaking change to the `TrafficServer` and `LegacyTrafficServer` structs by changing the type of `DeliveryServices` from `[]tsdeliveryService` to `[]TSDeliveryService`, but the odds that any external client is actually using that field, when it wasn't even populated in the monitoring.json snapshot in the first place, seem drastically low. Therefore, I felt that is an OK breaking change.

traffic_ops/traffic_ops_golang/crconfig/servers.go
ocket8888: It's a log line, so capitalization isn't a problem. From [the Golang wiki's Go Code Review Comments section](https://github.com/golang/go/wiki/CodeReviewComments#error-strings):

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use `fmt.Errorf("something bad")` not `fmt.Errorf("Something bad")`, so that `log.Printf("Reading %s: %v", filename, err)` formats without a spurious capital letter mid-message. **This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.**

(emphasis mine)

ocket8888: I don't really care; it's unnecessary, IMO, but not harmful to change it. Just thought I'd bring it up. Feel free to leave it as-is.

traffic_monitor/todata/todata.go
ocket8888: > Inserting a space between arguments is kind of surprising IMO.

I don't disagree, but it's Go's standard library doing it, technically, so not something we avoid without jumping through annoying hoops.

traffic_ops/traffic_ops_golang/crconfig/servers.go
rawlinp: Oh, good to know. I'll keep that in mind for next time 👍

docs/source/api/v2/cdns_name_configs_monitoring.rst
```diff
@@ -80,29 +83,36 @@ Response Structure
...
-	:ip6:           The :term:`cache server`'s IPv6 address - when applicable
-	:ip:            The :term:`cache server`'s IPv4 address
-	:port:          The port on which the :term:`cache server` listens for incoming connections
-	:profile:       A string that is the :ref:`profile-name` of the :term:`Profile` assigned to this :term:`cache server`
-	:status:        The status of the :term:`cache server`
-	:type:          A string that names the :term:`Type` of the :term:`cache server` - should (ideally) be either ``EDGE`` or ``MID``
+	:cacheGroup:       The :term:`Cache Group` to which this :term:`cache server` belongs
+	:deliveryServices: An array of objects which contain the XML IDs of the delivery services to which this cache server is assigned
+
+		:xmlId: A string which is the XML ID of the delivery service
```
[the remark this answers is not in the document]
ocket8888: Our API guidelines say

> If a property is missing from an object representation it indicates the absence of that property, and because of that there must be a very convincing argument if and when that is the case.

So being that it's unusual for properties to be undefined in our API, would you mind making note of that? People might reasonably expect `[]` - or even `null` - instead, since that's recommended. Omitting is consistent with the rest of that endpoint, unfortunately, though.

docs/source/api/v2/cdns_name_configs_monitoring.rst
```diff
@@ -58,9 +58,12 @@ Response Structure
 
 :deliveryServices: An array of objects representing each :term:`Delivery Service` provided by this CDN
 
+	:hostRegexes:        An array of strings which are the Delivery Service's HOST_REGEXP-type regexes
 	:status:             The :term:`Delivery Service`'s status
+	:topology:           A string that is the name of the Delivery Service's topology (if assigned one)
 	:totalKbpsThreshold: A threshold rate of data transfer this :term:`Delivery Service` is configured to handle, in Kilobits per second
 	:totalTpsThreshold:  A threshold amount of transactions per second that this :term:`Delivery Service` is configured to handle
+	:type:               A string that is the Delivery Service's type category (HTTP or DNS)
```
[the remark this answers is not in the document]
ocket8888: would you mind wrapping those in grave accents and double-quotation marks like
```rst
... (``"HTTP"`` or ``"DNS"``)
```
I just think that makes it very clear that those are literally the values you can expect.

ocket8888:
> > shouldn't these changes be made only to APIv4?
> 
> I don't see why it can't be made to all versions, since this is non-breaking to the expected JSON formatting because it only adds new fields and doesn't change data types or rename existing fields.
> 
> As far as the TO client goes, this _does_ technically make a breaking change to the `TrafficServer` and `LegacyTrafficServer` structs by changing the type of `DeliveryServices` from `[]tsdeliveryService` to `[]TSDeliveryService`, but the odds that any external client is actually using that field, when it wasn't even populated in the monitoring.json snapshot in the first place, seem drastically low. Therefore, I felt that is an OK breaking change.

That's fine with me, I suppose. It's not like we haven't broken that call signature in the recent past anyway - which I think was actually my doing. The docs pages for monitoring and Snapshots should really have a disclaimer about their instability. Right now monitoring has nothing and I think the Snapshots page has a "Caution" thing I put in saying how basically the guy writing the docs couldn't figure out a definite "shape" for that endpoint's data, so it needed updates from people who did. Should just be changed to "this, historically, just can't be relied upon to respect its versioning in the API". Not that I think you need to do that in this PR - although you're welcome to if you want.

And, of course, I'll then need you to tank aggro when @rob05c finds out about it.

docs/source/api/v2/cdns_name_configs_monitoring.rst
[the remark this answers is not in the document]
rawlinp: Done

docs/source/api/v2/cdns_name_configs_monitoring.rst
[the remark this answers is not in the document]
rawlinp: Done

rawlinp:
I'll tank if you heal 😄 

IMO all snapshots are unversioned because they should be handled as raw JSON data to which we can freely add new fields without making breaking changes. Plus, the fact that they're read-only means they're immune to round-trip data loss which could occur if a client were to drop unknown fields from a GET then make a PUT/POST without those fields.

ocket8888:
I think you need to rebase this on master to get the CiaB CI to pass

zrhoffman:
> ~I think you need to rebase this on master to get the CiaB CI to pass~ nvm - [that won't help yet](https://github.com/apache/trafficcontrol/pull/6673#issuecomment-1072656537)

fixed in #6678

rawlinp:
```
ERROR: for to_server  (<Service: to_server>, 'Build failed')
Service 'to_server' failed to build : Build failed
Error: Process completed with exit code 1.
```
However, this builds fine for me locally using docker... I can't really tell why it's failing here.

ocket8888 (approved):

PR #6631, opened 2022-03-10
update mmdb 304 date check

ezelkow1:
Currently the way we create a date string for an IMS request for the maxmind database is incorrect. Currently it is generated in AM/PM format. This changes it to 24hr

<hr/>

You can issue the current date string against a file such that it generates a 'PM' timestamp. Then try using that date string for an IMS request against a file modified in the PM but in 24hr format. In this case you can get back a 200 response which would trigger a maxmind db update and config reload

Then change to use the new string that has the proper 24hr format and you should be able to see a proper 304 response instead

## PR submission checklist
- [x] This PR has tests 


rob05c (approved):
LGTM. Tested that date format, produces what appears to be a proper `HTTP-Date`; existing `%r` does not, `PM` is not a `HTTP-Date`.

PR #6632, opened 2022-03-10
Only update tm_user.last_authenticated once per minute

rawlinp:
This is to avoid row-locking when the same user logs in frequently.

<hr/>

- Traffic Ops

As the same user, log in more than once a minute while checking the user's `last_authenticated` time. It should only update once per minute.

## PR submission checklist
- [x] ~This PR has tests~ 
- [x] ~This PR has documentation~ 

### DIFF
--- a/traffic_ops/traffic_ops_golang/login/login.go
+++ b/traffic_ops/traffic_ops_golang/login/login.go
@@ -63,7 +63,9 @@ WHERE name='tm.instance_name' AND
 `
 const userQueryByEmail = `SELECT EXISTS(SELECT * FROM tm_user WHERE email=$1)`
 const setTokenQuery = `UPDATE tm_user SET token=$1 WHERE email=$2`
-const UpdateLoginTimeQuery = `UPDATE tm_user SET last_authenticated = now() WHERE username=$1`
+
+// UpdateLoginTimeQuery is meant to only update the last_authenticated field once per minute in order to avoid row-locking when the same user logs in frequently.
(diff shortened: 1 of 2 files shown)


traffic_ops/traffic_ops_golang/login/login.go
```diff
@@ -63,7 +63,9 @@ WHERE name='tm.instance_name' AND
 `
 const userQueryByEmail = `SELECT EXISTS(SELECT * FROM tm_user WHERE email=$1)`
 const setTokenQuery = `UPDATE tm_user SET token=$1 WHERE email=$2`
-const UpdateLoginTimeQuery = `UPDATE tm_user SET last_authenticated = now() WHERE username=$1`
+
+// UpdateLoginTimeQuery is meant to only update the last_authenticated field once per minute in order to avoid row-locking when the same user logs in frequently.
+const UpdateLoginTimeQuery = `UPDATE tm_user SET last_authenticated = now() WHERE username=$1 AND last_authenticated < now() - INTERVAL '1 minute'`
```
zrhoffman, line 68: Since the other SQL keywords are uppercase, should `now()` be uppercase? Should `minute`?

zrhoffman, line 68: If `last_authenticated` is null, this condition will be false

rawlinp: Yeah, just found that too, will fix 👍

zrhoffman (approved):
Looks good! This def should help performance.

PR #6633, opened 2022-03-10
Fix TO to use DB port from config

rawlinp:
Allow ports other than the 5432 default to be used.

<hr/>

- Traffic Ops

Start postgres on a port other than 5432 for TODB, change your database.conf to use that port, and verify that TO works properly w/ TODB.

- 5.x
- 6.x

## PR submission checklist
- [x] ~This PR has tests~ 
- [x] ~This PR has documentation~ 

### DIFF
--- a/traffic_ops/traffic_ops_golang/config/config.go
+++ b/traffic_ops/traffic_ops_golang/config/config.go
@@ -28,6 +28,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/apache/trafficcontrol/lib/go-log"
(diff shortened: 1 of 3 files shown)


traffic_ops/traffic_ops_golang/config/config.go
```diff
@@ -317,6 +323,9 @@ func LoadConfig(cdnConfPath string, dbConfPath string, appVersion string) (Confi
 	if err != nil {
 		return Config{}, []error{fmt.Errorf("unmarshalling '%s': %v", dbConfPath, err)}, BlockStartup
 	}
+	if portNum, err := strconv.Atoi(cfg.DB.Port); err != nil || portNum < MinPort || MaxPort < portNum {
+		cfg.DB.Port = DefaultDBPort
```
srijeet0406, line 327: Do you think we should print an error here, saying that the port in the config file was invalid, so we're defaulting to 5432?

rawlinp: done

srijeet0406 (approved):
Code changes look good. Tested locally and made sure that TO could connect to the postgres instance specified at the port in `database.conf`, and not just `5432`.

PR #6636, opened 2022-03-11
Modify TP init script to be systemd compliant

shamrickus:
This PR modifies the init script to be systemd complaint, which it was before it was rewritten to support `pm2` in #6559
<hr/>

- Traffic Portal
- Automation (CI)

Attempt to run the `/etc/init.d/traffic_portal` script as a systemd service and verify that it works.

 - master

## PR submission checklist
- [ ] This PR has tests 

### DIFF
--- a/traffic_portal/build/etc/init.d/traffic_portal
+++ b/traffic_portal/build/etc/init.d/traffic_portal
@@ -52,6 +52,7 @@ RESTART_TIME="2000"
 
 PATH=$PM2_BIN_DIR:$NODE_BIN_DIR:$PATH
 export NODE_PATH=$NODE_PATH
+set -o pipefail
 
 start() {
     echo "Starting $NAME"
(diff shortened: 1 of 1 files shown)


traffic_portal/build/etc/init.d/traffic_portal
```diff
@@ -46,14 +62,14 @@ start() {
...
-      --name "$NAME"
+      --name "$NAME" 2>&1 > /dev/null
     RETVAL=$?
 }
 
 stop() {
     if pm2 list | grep -q "$NAME"; then
         echo "Shutting down $NAME"
-        pm2 stop $APPLICATION_PATH
+        pm2 stop $APPLICATION_PATH 2>&1 > /dev/null
```
zrhoffman, line 72: `2>&1` needs to go after `> /dev/null` or it doesn't redirect `/dev/stderr` to `/dev/null`

zrhoffman (changes requested):
https://github.com/apache/trafficcontrol/blob/1867d42ac8573826e49405115be0a71802068e3f/traffic_portal/build/etc/init.d/traffic_portal#L69-L70

https://github.com/apache/trafficcontrol/blob/1867d42ac8573826e49405115be0a71802068e3f/traffic_portal/build/etc/init.d/traffic_portal#L85-L86

Since the `pipefail` bash option is not enabled by default, this condition is always met.

traffic_portal/build/etc/init.d/traffic_portal
shamrickus: Will change, I was just copy & pasting what the script was doing previously.

zrhoffman (approved):
Looks good!

PR #6639, opened 2022-03-11
Added Traffic Portal page for CDNi config requests

mattjackson220:
This PR adds an endpoint to TO to GET all of the pending CDNi configuration update requests and a page in TP to view them and approve/deny

<hr/>

- Documentation
- Traffic Ops
- Traffic Portal

- Verify that all TO and TP tests pass
- Verify that all docs build and look good
- 1. Run TO and TP. (Be sure your [TO configuration file ](https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_ops.html#cdn-conf) sets the `cdni` property properly)
    2. Navigate to the new `cdni-config-requests` page using the side menu Configuration > CDNi Requests
    3. Manually put CDNi data in the database (test INSERTs for data below)
    4. Submit a configuration update request to `/api/4.0/OC/CI/configuration/`
    5. Verify that the data loads successfully
    6. Click on one of the rows and verify that it takes you to the corresponding request page by ID
    7. Click Approve and verify that the new data is implemented at the `/OC/FCI/advertisement` endpoint
    8. Submit another request, verify it shows up
    9. Deny the new request and verify that the data is NOT updated at the `/OC/FCI/advertisement` endpoint

```sql

INSERT INTO cdni_capabilities ("type", ucdn) VALUES ('FCI.CapacityLimits', 'ucdn1');
INSERT INTO cdni_capabilities ("type", ucdn) VALUES ('FCI.Telemetry', 'ucdn1');
INSERT INTO cdni_footprints (footprint_type, footprint_value, ucdn, capability_id) VALUES ('countrycode', ARRAY['us'], 'ucdn1', 1);
INSERT INTO cdni_footprints (footprint_type, footprint_value, ucdn, capability_id) VALUES ('ipv4cidr', ARRAY['127.0.0.1','127.0.0.2'], 'ucdn1', 1);
INSERT INTO cdni_telemetry (id, "type", capability_id) VALUES ('capacity_metrics', 'generic', 2);
INSERT INTO cdni_telemetry (id, "type", capability_id) VALUES ('request_metrics', 'generic', 2);
INSERT INTO cdni_telemetry_metrics ("name", time_granularity, data_percentile, latency, telemetry_id) VALUES ('capacity', 0, 50, 0, 'capacity_metrics');
INSERT INTO cdni_telemetry_metrics ("name", time_granularity, data_percentile, latency, telemetry_id) VALUES ('requests', 0, 50, 0, 'request_metrics');
INSERT INTO cdni_host_limits (limit_type, maximum_hard, maximum_soft, telemetry_id, capability_id, telemetry_metric, host) VALUES ('requests', 200, 10, 'request_metrics', 1, 'requests', 'example.com');
INSERT INTO cdni_host_limits (limit_type, maximum_hard, maximum_soft, telemetry_id, capability_id, telemetry_metric, host) VALUES ('requests', 20, 15, 'request_metrics', 1, 'requests', 'example2.com');
INSERT INTO cdni_total_limits (limit_type, maximum_hard, maximum_soft, telemetry_id, capability_id, telemetry_metric) VALUES ('egress', 3000, 500, 'capacity_metrics', 1, 'capacity');
```

## PR submission checklist
- [ ] This PR has tests


traffic_portal/test/integration/PageObjects/SideNavigationPage.po.ts
```diff
@@ -143,6 +144,10 @@ export class SideNavigationPage extends BasePage{
         await browser.actions().mouseMove(this.lnkStatuses).perform();
         await browser.actions().click(this.lnkStatuses).perform();
     }
+    /** NavigateToCdniRequestsPage verifies that the link to the CDNi Requests page is clickable. */
+    public async NavigateToCdniRequestsPage(): Promise<void>{
+        return this.lnkCdniConfigRequests.click();
+    }
```
ocket8888, line 150: This method is never used - were you going to write tests for this?

traffic_portal/app/src/common/modules/form/cdniConfigRequests/form.cdniConfigRequests.tpl.html
```diff
@@ -0,0 +1,53 @@
...
+        <div class="clearfix"></div>
+    </div>
+    <div class="x_content">
+        <br>
+        <form name="cdniRequestForm" class="form-horizontal form-label-left" novalidate>
+            <div class="form-group">
+                <label for="newData" class="control-label col-md-2 col-sm-2 col-xs-12">Updated Data</label>
+                <div class="col-md-10 col-sm-10 col-xs-12" style="padding-top: 10px">
+                    <div class="col-md-10 col-sm-10 col-xs-12">
+                        <textarea id="newData" name="newData" type="text" class="form-control" ng-model="cdniRequest.data" rows="25" readonly></textarea>
```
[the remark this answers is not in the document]
ocket8888: oh, I see

traffic_portal/app/src/modules/private/cdniConfigRequests/index.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+module.exports = angular.module('trafficPortal.private.cdniConfigRequests', [])
+    .config(function($stateProvider, $urlRouterProvider) {
+        $stateProvider
+            .state('trafficPortal.private.cdniConfigRequests', {
+                url: 'cdni-config-requests',
+                abstract: true,
+                views: {
+					privateContent: {
```
shamrickus, line 27: Inconsistent spacing

docs/source/api/v4/oc_fci_configuration_requests.rst
```diff
@@ -0,0 +1,96 @@
...
+``OC/FCI/configuration/requests``
+*********************************
+
+``GET``
+=======
+Returns the requested updates for :abbr:`CDNi (Content Delivery Network Interconnect)` configurations. An optional ``id`` parameter will return only information for a specific request.
+
+:Auth. Required: Yes
+:Roles Required: "admin"
+:Permissions Required: CDNI-CAPACITY:ADMIN
```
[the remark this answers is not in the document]
shamrickus: I think having it be it's own permission is preferable, also it should probably be `{name}:UPDATE` since this is a put.

ocket8888:
When I log into TP on this branch, run those queries, and create a few requests, I can see data in the TP table view, but when I click on a row it silently fails to navigate. There is a failed `GET` request being made to `OC/FCI/advertisement?ucdn=ucdn1`. The returned `error-level` Alert states "invalid token - ucdn must be defined in audience claim". So I can't view anything. (Note that the table in TP shows that the `ucdn` value for the entry I'm clicking on is "ucdn1", which is the same as that query parameter, and I am an `admin`-Role user).

The failure is a problem but also: the alert doesn't show up. Clicking just does nothing and you have to open the browser's developer tools to know an error occurred.

shamrickus:
@ocket8888 those fail because you didn't add
```
"cdni" : {
        "dcdn_id" : "CDN"
    }
``` 
to your cdn.conf

ocket8888:
My TO configuration file has
```json
"cdni": {
	"dcdn_id": "ucdn1"
}
```

Even if that's the problem, that means the alert message is wrong, the response code should probably be 500, and no matter what I should see the alert in TP so I know something failed.

ocket8888:
> https://github.com/apache/trafficcontrol/pull/6639#discussion_r840659246

I don't know why responses aren't allowed to the conversations in this PR anymore, so I'm responding here

>  also it should probably be {name}:UPDATE since this is a put.

that's actually a GET, though. The `.../approved` endpoint he mentioned is a PUT

> I can change it to be a READ but it would need to be different than the CDNI-CAPACITY:READ and makes more sense to me that it would mirror the approved endpoint. what do you think?

I think normally the way this is handled is the read-only operation has a <CODE><var>NAME</var>:READ</CODE> Permission and then the endpoint to actually do things with/to that data requires that Permission _and_ its own <code><var>NAME</var>:UPDATE</code> Permission (or `CREATE` or whatever's appropriate). So that way the data can be displayed to people who don't have Permission to change it.

These CDNI endpoints are highly incongruous with our existing API structure and practices, normally if you really only wanted people who could manipulate the data to see it I'd be puzzled why they are two different endpoints instead of different methods on a single endpoint. So that's part of the issue is it's just hard to think about how these things are "normally" handled. But if it's not a spec requirement that only people with the ability to change it be allowed to see it, I think it's better to just be more flexible and add the READ Permission here as well as an extra Permission required by the manipulation endpoint, since that's more in-line with how the regular API works.

traffic_ops/traffic_ops_golang/cdni/shared.go
```diff
@@ -738,32 +802,48 @@ const (
...
 }
+
+// ConfigurationUpdateRequest contains information about a requested CDNi configuration update request.
+type ConfigurationUpdateRequest struct {
+	ID            int             `json:"id"`
+	UCDN          string          `json:"ucdn"`
+	Data          json.RawMessage `json:"data"`
+	Host          string          `json:"host"`
+	RequestType   string          `json:"request-type" db:"request_type"`
+	AsyncStatusID int             `json:"async-status-id" db:"async_status_id"`
```
ocket8888, line 848: I don't know this for sure, but asynchronous job IDs doesn't seem like the kind of thing that would be in the CDI spec. If it isn't, then this should be camelCase in JSON as `asyncStatusID`

traffic_portal/app/src/common/api/CdniService.js
```diff
@@ -0,0 +1,75 @@
...
+ * under the License.
+ */
+
+var CdniService = function($http, ENV, messageModel) {
+
+	this.getCdniConfigRequests = function() {
+		return $http.get(ENV.api.unstable + 'OC/CI/configuration/requests').then(
+			function (result) {
+				return result.data.response;
+				},
```
ocket8888, line 26: this isn't something you need to change, but just for future reference our webpack version has been updated so we can use ES6 features in TP now. And in case that doesn't mean anything to you, it means we can use "fat arrow functions" to express these inline functions more succinctly with the syntax `(arg, list) => {some; statements;}`/`arg => {some; statements;}`/`(arg, list) => result`/`arg=>result`.

Like I said you don't have to change it, so I'm not gonna point out everywhere you could, but just to use this callback as an example, you could write the `getCdniConfigRequests` as such:
```javascript
this.getCdniConfigRequests = () => $http.get(`${ENV.api.unstable}OC/CI/configuration/requests`).then(
    result => result.data.response,
    err => {
        messageModel.setMessages(err.data.alerts, true);
        throw err;
    }
);
```
which I like a lot more than having to write out all the boilerplate of `function` and `return` and `{`/`}` pairs each time.

ocket8888 (commented):
I still need to retest again. Previously I had this in my TO config:
```json
"cdni": {
	"dcdn_id": "ucdn1"
}
```
and I'm wondering now, does `dcdn_id` need to be my user's ucdn, or is it supposed to be the name of some ATC CDN? Or some other third thing?

mattjackson220:
> I still need to retest again. Previously I had this in my TO config:
> 
> ```json
> "cdni": {
> 	"dcdn_id": "ucdn1"
> }
> ```
> 
> and I'm wondering now, does `dcdn_id` need to be my user's ucdn, or is it supposed to be the name of some ATC CDN? Or some other third thing?

yea so the `dcdn` is the downstream CDN which in our case would be like ATC_CDN or whatever you like. And the ucdn is the upstream so that would be an external CDN (i used `ucdn1` in the example). so the `dcdn_id` should be anything other than `ucdn1` really. otherwise it will think you are `ucdn1` talking to itself

traffic_portal/app/src/common/api/CdniService.js
mattjackson220: this is really good to know! thats a lot cleaner!

traffic_ops/traffic_ops_golang/routing/routes.go
```diff
@@ -135,8 +135,8 @@ func Routes(d ServerData) ([]Route, http.Handler, error) {
...
-		{Version: api.Version{Major: 4, Minor: 0}, Method: http.MethodPut, Path: `OC/CI/configuration/request/{id}/{approved}$`, Handler: cdni.PutConfigurationResponse, RequiredPrivLevel: auth.PrivLevelAdmin, RequiredPermissions: []string{"CDNI-CAPACITY:ADMIN"}, Authenticated: Authenticated, Middlewares: nil, ID: 541357729080},
-		{Version: api.Version{Major: 4, Minor: 0}, Method: http.MethodGet, Path: `OC/CI/configuration/requests/?$`, Handler: cdni.GetRequests, RequiredPrivLevel: auth.PrivLevelAdmin, RequiredPermissions: []string{"CDNI-CAPACITY:ADMIN"}, Authenticated: Authenticated, Middlewares: nil, ID: 541357729081},
+		{Version: api.Version{Major: 4, Minor: 0}, Method: http.MethodPut, Path: `OC/CI/configuration/request/{id}/{approved}$`, Handler: cdni.PutConfigurationResponse, RequiredPrivLevel: auth.PrivLevelAdmin, RequiredPermissions: []string{"CDNI-ADMIN:READ", "CDNI-ADMIN:UPDATE"}, Authenticated: Authenticated, Middlewares: nil, ID: 541357729080},
```
ocket8888, line 138: This change must also be reflected in the docs at `docs/source/api/v4/oc_ci_configuration_request_id_approved.rst`

ocket8888 (commented):
Works fine

ocket8888 (approved):

PR #6648, opened 2022-03-11
TR should support aggressive NSEC queries

srijeet0406:
This PR is not related to any issue. It fixes the way TR responds to aggressive NSEC queries.
<hr/>

- Traffic Router

Make sure all the unit/ integration tests pass/ run `mvn clean verify` succeeds.
Make the following DNS queries:
`
dig nsec speedtest-prod.xcr.comcast.net. +dnssec
`
`
dig nsec voice-guidance-tts.xcr.comcast.net +dnssec
`
Notice how the first one returns an incorrect `NSEC` response, saying that the next lexical record is at `cdn-tr-...`, whereas, the next record should be `cdn.speedtest-prod.xcr.comcast.net.`.
For the second one, it incorrectly returns two different `NSEC` responses:
`voice-guidance-tts.xcr.comcast.net. 30 IN NSEC cdn-ec-alb-301.voice-guidance-tts.xcr.comcast.net. NS SOA RRSIG NSEC DNSKEY
voice-guidance-tts.xcr.comcast.net. 30 IN NSEC ccr.voice-guidance-tts.xcr.comcast.net. NS SOA RRSIG NSEC DNSKEY`
The above two records cannot be true at the same time, because it implies that the next lexical record is `cdn-ec-alb...` and there is nothing in between them, while also saying that the next record is `ccr.voice...` so there is lexically nothing between them. Both those statements cannot be true at the same time.

Now, run the same two queries but by pointing it at your TR with the changes of this PR.
Notice the correct responses this time around.

- master

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/dns/ZoneManager.java
+++ b/traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/dns/ZoneManager.java
@@ -940,7 +940,7 @@ public class ZoneManager extends Resolver {
 			final DNSRouteResult result = trafficRouter.route(request, track);
 
 			if (result != null) {
-				final Zone dynamicZone = fillDynamicZone(dynamicZoneCache, staticZone, request, result);
+				final Zone dynamicZone = fillDynamicZone(getDynamicZoneCache(), staticZone, request, result);
 				track.setResultCode(dynamicZone, request.getName(), request.getQueryType());
 				if (result.getDeliveryService() == null) {
 					builder.deliveryServiceXmlIds(null);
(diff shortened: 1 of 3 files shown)


traffic_router/core/src/main/java/org/apache/traffic_control/traffic_router/core/dns/ZoneManager.java
```diff
@@ -1009,7 +1015,20 @@ private static Zone fillDynamicZone(final LoadingCache<ZoneKey, Zone> dzc, final
 						continue;
 					}
 
-					records.add(r);
+					if (r.getType() == Type.NSEC && r.getName().toString().equals(staticZone.getOrigin().toString())) {
```
rawlinp, line 1018: We might need an `&& dsRoutingName != null` check here just for good measure, although it would likely never be null in the first place.

PR #6651, opened 2022-03-14
Run Angular 12/13 migrations for TPv2

shamrickus:
This PR runs the migrations that were intended to be ran when upgrading from Angular 11. It also updates some packages.
<hr/>

- Traffic Portal 2

Verify tests pass, production build is correctly optimized/minimized/etc, dev build _is not_ and that theming still works within the app.

Only `package.json` and `.gitignore` were edited manually, `package` for the package updates and `.gitignore` to remove the ignoral of the entire `.angular` folder since it _can_ have project wide config files.

## PR submission checklist
- [x] This PR has tests 


experimental/traffic-portal/.gitignore
```diff
@@ -12,7 +12,6 @@
 
 # compiled output
 /dist
-/.angular
```
ocket8888, line 15: I don't think this should have been changed. As of right now, we don't want anything that's in `.angular`, not just the `cache` subdirectory.

ocket8888 (commented):
package-lock conflicts

experimental/traffic-portal/.gitignore
shamrickus: I wanted to match the behavior of a newly generated angular project and it doesn't ignore `.angular`. That having been said I don't mind adding it back since as you said, it doesn't currently do anything.

ocket8888: I can't fathom why they'd add it like that. `.angular` looks like a directory meant for use by angular, and configuration used by its users is already outside of there so I have no idea why anyone wouldn't just want to ignore the whole thing.

ocket8888 (approved):
LTS support for npm versions less than 6 ends this month, so we're fine to update to lockfile version 2 - but future PRs *must not* regress that to version 1.

PR #6664, opened 2022-03-16
Only `operations` and `admin` roles should have the `DELIVERY-SERVICE:UPDATE` permission

srijeet0406:
This PR is not related to any issue.

<hr/>

- Traffic Ops

Before the change: make sure a user with the `portal` role can update a delivery service.
After the change: make sure a user with the `portal` role can only update a DS via the `safe` update endpoint.

- master

## PR submission checklist
- [ ] This PR has tests 
- [ ] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://apache.org/security) for details)


ocket8888:
Allowing users with a `priv_level` below 20 to do "safe" updates is privilege escalation. The route definition in `traffic_ops/traffic_ops_golang/routing/routes.go` restricts that operation to users with a `priv_level` of _at least_ 20 - "operator" or higher.

srijeet0406:
> Allowing users with a `priv_level` below 20 to do "safe" updates is privilege escalation. The route definition in `traffic_ops/traffic_ops_golang/routing/routes.go` restricts that operation to users with a `priv_level` of _at least_ 20 - "operator" or higher.

@mitchell852 Do the roles below priv level 20 _need_ the "safe" update permissions?

ocket8888:
> @mitchell852 Do the roles below priv level 20 need the "safe" update permissions?
actually be able to do that
If we decide they do, the route definitions in `routes.go` will need to change. Otherwise, no matter what Permission you give their Role, users with a "priv level" that low using our current stable API version (3) will not actually be able to do it. And currently, they can't.

rawlinp:
I think requiring operations priv-level was a regression from TO-Perl: https://github.com/apache/trafficcontrol/blob/4.0.x/traffic_ops/app/lib/API/Deliveryservice.pm#L534

We only checked if they're operations priv-level if we weren't using tenancy (which we do now). Therefore, it should be usable by any role, and the update should be allowed as long as the user's tenant checks out.

ocket8888:
> I think requiring operations priv-level was a regression from TO-Perl: https://github.com/apache/trafficcontrol/blob/4.0.x/traffic_ops/app/lib/API/Deliveryservice.pm#L534
> 
> We only checked if they're operations priv-level if we weren't using tenancy (which we do now). Therefore, it should be usable by any role (other than `disallowed` obviously), and the update should be allowed as long as the user's tenant checks out.

That would be consistent with the documentation not mentioning any required Roles. If nobody's been relying on the behavior of that regression then never mind what I said.

rawlinp (approved):

PR #6666, opened 2022-03-16
When targets are ignored, only build delve if debug compose file is used

zrhoffman:
This PR fixes a hard-to-reproduce bug encountered when building the CDN in a Box images from within a Jenkins pipeline.
The bug is that Docker build targets are ignored, so even though `dlv` should not be installed in non-debug images, it ends up being installed in the `trafficops`, `trafficmonitor`, `trafficstats`, `enroller`, `edge`, `mid-01`, and `mid-02` images. As a result, if there is any issue building `dlv` in this environment, building the entire image will fail.

This PR adds, in the `optional/docker-compose.debugging.yml` file, a `DEBUG_ENABLE` build arg to each of the compose services listed above. That way, even if build targets are ignored, `dlv` will not be built unless `optional/docker-compose.debugging.yml` is included in the docker-compose command.
<hr/>

- CDN in a Box

```shell
docker-compose -f docker-compose.yml -f optional/docker-compose.debugging.yml build --build-arg DEBUG_ENABLE=false trafficops
```
then
```shell
docker-compose -f docker-compose.yml -f optional/docker-compose.debugging.yml build --build-arg DEBUG_ENABLE=false trafficops
```

Finally,
```shell
docker run --rm -it trafficops sh -c dlv
```
expected output:
```
sh: dlv: command not found
```

## PR submission checklist
- [ ] This PR has tests 

### DIFF
--- a/infrastructure/cdn-in-a-box/cache/Dockerfile
+++ b/infrastructure/cdn-in-a-box/cache/Dockerfile
@@ -120,8 +120,12 @@ RUN rpm -Uvh /$(basename $ORT_RPM) &&\
 CMD /run.sh
 
 FROM common-traffic-server-dependencies AS get-delve
-RUN dnf -y install golang && \
-    go install github.com/go-delve/delve/cmd/dlv@latest
+RUN if [[ "$DEBUG_ENABLE" == true ]]; then \
+        dnf -y install golang && \
+        go install github.com/go-delve/delve/cmd/dlv@latest; \
+    else \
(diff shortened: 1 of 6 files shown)


zrhoffman:
The target-ignoring bug described above is no longer encountered if you set `DOCKER_BUILDKIT` and `COMPOSE_DOCKER_CLI_BUILD=1` in the environment. Closing the PR as no-longer-necessary

PR #6668, opened 2022-03-17
Minor improvements to Cachegroup Tests

ericholguin:
This PR removes un-needed headers that were being used. Since `GET` requests only check the `If-Modified-Since` headers they don't need the `If-Unmodified-Since` headers and vice versa with `PUT` requests. 

The mod time was also changed from 5 to 15 seconds before current time to allow the tests time to perform necessary changes.

The cachegroup PUT request body was also simplified for the Precondition tests, to reduce additional lines.

<hr/>

- Traffic Ops Tests

Run the TO integration tests

## PR submission checklist


zrhoffman (approved):
Looks good! Running the TO API tests against CDN in a Box, all TestCachegroups tests now pass.

PR #6671, opened 2022-03-17
Add TO in-memory User cache option to improve performance

rawlinp:
Currently, logging in a user can take up to 4 different DB queries, and
most route handlers require at least 1 DB query to look up the user and
role of the requester.

By periodically reading users and roles from the database (using the
interval defined by `user_cache_refresh_interval_sec` in cdn.conf) and
storing them in memory, Traffic Ops can quickly check the user's data
and role without having to read from the database at request time.

This makes logins extremely efficient because they require no database
queries at request time. Additionally, overall database load is reduced
because most route handlers have reduced the number of queries they make
by 1.

From a security perspective, making logins extremely efficient is
advantageous because it reduces the effectiveness of denial of service
attacks targeting the login endpoint (which would normally make at least
1 database query to check that the user is allowed to log in). With this
feature enabled, a DoS attack against the login endpoint would not
exhaust limited database resources. Note: other endpoints are not as
susceptible to DoS attacks due to requiring a valid cookie before any
database queries are made.

The downside to enabling this feature is that updates to Users and/or
Roles are not enforced by Traffic Ops immediately when the updates are
made. If `user_cache_refresh_interval_sec` is N, it may take up to N
seconds before Traffic Ops will enforce updates to Users and/or Roles.
However, as long as this interval is sufficiently short, this should not
be a problem.

<hr/>

- Documentation
- Traffic Ops
- Ansible

Ensure the automated tests pass. Manual: configure `user_cache_refresh_interval_sec` in cdn.conf to 5, start TO, and change a user's role (or its role's permissions), while using that user to make requests for endpoints with the corresponding permissions that are being changed. Ensure that the changed permissions are enforced within 5 seconds of making the change via the API.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/traffic_ops_golang/auth/authorize.go
+++ b/traffic_ops/traffic_ops_golang/auth/authorize.go
@@ -254,8 +254,8 @@ func CheckLocalUserPassword(form PasswordForm, db *sqlx.DB, ctx context.Context)
 // maximum duration of timeout.
 func CheckLocalUserToken(token string, db *sqlx.DB, timeout time.Duration) (bool, string, error) {
 	if usersCacheIsEnabled() {
-		username := getUserNameFromCacheByToken(token)
-		return username != "", username, nil
+		username, matched := getUserNameFromCacheByToken(token)
+		return matched, username, nil
 	}
 	dbCtx, dbClose := context.WithTimeout(context.Background(), timeout)
 	defer dbClose()
--- a/traffic_ops/traffic_ops_golang/auth/authorize.go
+++ b/traffic_ops/traffic_ops_golang/auth/authorize.go
@@ -87,8 +87,6 @@ const PrivLevelUnauthenticated = 0
 
 const PrivLevelReadOnly = 10
 
-const PrivLevelORT = 11
-
 const PrivLevelSteering = 15
 
 const PrivLevelFederation = 15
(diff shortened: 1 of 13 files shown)


rawlinp:
```
ERROR: for to_server  (<Service: to_server>, 'Build failed')
Service 'to_server' failed to build : Build failed
Error: Process completed with exit code 1.
```
However, this builds fine for me locally using docker... I can't really tell why it's failing here.

traffic_ops/traffic_ops_golang/auth/authorize.go
```diff
@@ -197,15 +217,26 @@ func GetUserUcdn(form PasswordForm, db *sqlx.DB, ctx context.Context) (string, e
...
-	if err != nil {
-		if err == context.DeadlineExceeded || err == context.Canceled {
-			return false, nil, err
+	if usersCacheIsEnabled() {
+		u, exists := getUserFromCache(form.Username)
+		if !exists {
+			return false, fmt.Errorf("user '%s' not found in cache", form.Username), nil
+		}
+		if u.LocalPasswd == nil {
+			return false, nil, nil
```
srijeet0406, line 226: Should we return some kind of an error here, saying that the localpswd was `nil`?

traffic_ops/traffic_ops_golang/auth/usercache.go
```diff
@@ -0,0 +1,208 @@
...
+// getUserFromCache returns the user with the given username and a boolean indicating whether the user exists.
+func getUserFromCache(username string) (user, bool) {
+	usersCache.RLock()
+	defer usersCache.RUnlock()
+	u, exists := usersCache.userMap[username]
+	return u, exists
+}
+
+// getUserNameFromCacheByToken returns the username with the given token or an empty string if not found.
+func getUserNameFromCacheByToken(token string) string {
```
srijeet0406, line 98: just for uniformity sake, does it make sense to return a boolean here to denote whether or not the username was found, and whether or not it was = `""`?

srijeet0406: That way we don't have to check it everywhere this function gets called.

traffic_ops/traffic_ops_golang/auth/authorize.go
rawlinp: I don't think so, because this means the user doesn't have a local password, which is normal behavior if a user is LDAP-only. The original code (and non-usersCache path) will log this non-error, but I didn't feel like correcting that path since it means checking a different sql error and making that conditional more complex.

traffic_ops/traffic_ops_golang/auth/usercache.go
rawlinp: I actually considered that approach, but this is only getting called in one place and is only for the token login, so I just left it that way. That said, I'll go ahead and do it for consistency at least.

srijeet0406 (approved):
LGTM! Code looks good and all manual/ automated tests pass.

PR #6673, opened 2022-03-18
if no parents exist t3c will use secondary parents if configured 

jpappa200:
This PR will address CDN-15204 

<hr/>

In a staging environment offline the primary parent in an MSO and run t3c. you will see the error messages in the warning summary and check parent.config.  

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/lib/go-atscfg/parentdotconfig.go
+++ b/lib/go-atscfg/parentdotconfig.go
@@ -970,7 +970,13 @@ func getTopologyParentConfigLine(
 		return nil, warnings, errors.New("getting topology parents for '" + *ds.XMLID + "': skipping! " + err.Error())
 	}
 	if len(parents) == 0 {
-		return nil, warnings, errors.New("getting topology parents for '" + *ds.XMLID + "': no parents found! skipping! (Does your Topology have a CacheGroup with no servers in it?)")
+		if len(secondaryParents) > 0 {
+			warnings = append(warnings, "getting topology parents for '"+*ds.XMLID+"': no parents found! using secondary parents")
+			parents = secondaryParents
+			secondaryParents = nil
+		} else {
+			return nil, warnings, errors.New("getting topology parents for '" + *ds.XMLID + "': no parents found! skipping! (Does your Topology have a CacheGroup with no servers in it?)")
+		}
 	}
 
 	txt.Parents = parents


zrhoffman:
> This PR will address CDN-15204

Will you please file a GitHub issue? As far as the ATC community is concerned, there is no such thing as CDN-15204.

ocket8888:
CiaB failure may be addressed by rebasing on master

zrhoffman:
> CiaB failure may be addressed by rebasing on master

It won't be. The failure is coming from the fact that CDN in a Box expects an RPM named `trafficserver-8.1.3-10.965df952e.el8.x86_64.rpm`, but an RPM named `trafficserver-8.1.4_rc0-0.965df952e.el8.x86_64.rpm` is provided. So the [`fetch-github-branch-sha` GHA](https://github.com/apache/trafficcontrol/tree/9c7786bbfe/.github/actions/fetch-github-branch-sha) needs to have its latest-tag-getting logic updated to exclude tags that have characters other than digits and `.` in them.

zrhoffman:
> CiaB failure may be addressed by rebasing on master

@ocket8888 fixed in #6678

rob05c:
` This PR has tests` `This PR has documentation` `This PR has a CHANGELOG.md entry` are all checked, but it doesn't look like any of those are in the PR. @jpappa200 mind adding them?

I don't think we need docs, this behavior is incidental and the only right thing to do, we don't need to document that level of detail. But that should probably be noted, rather than checking the box that it has docs when it doesn't.

But it should probably have a Changelog entry, so people know how and when it changed.

And it should be straightforward to add a unit test, shouldn't need an Integration Test. In https://github.com/apache/trafficcontrol/blob/3b8c0cdc/lib/go-atscfg/parentdotconfig_test.go just needs to make a test with no Primaries and a Secondary via CacheGroups, which should pass with the code change and fail without it.

Other than that, this change looks good. The previous behavior was generating broken config, this is the right thing to do.

rob05c:
You can probably base the test on `TestMakeParentDotConfigHTTPSOrigin` in `parentdotconfig_test.go`, if it helps. The setup should be similar, just needs to set the `eCG.SecondaryParentCachegroupID = mid0.CachegroupID` and not `SecondaryParentCachegroupID`, I think.
`

ocket8888:
This is including a ton of changes already merged and not at all related to anything you actually did - I think you may want to rebase.

jpappa200:
Yea that's what got me into this mess. Something went wrong.

ocket8888:
TM tests are known to be flaky. I re-ran them

jpappa200:
> > This PR will address CDN-15204
> 
> Will you please file a GitHub issue? As far as the ATC community is concerned, there is no such thing as CDN-15204.

Created issue #6780

rob05c (approved):
LGTM, thanks!

PR #6675, opened 2022-03-18
Fix Geo Limit Countries in versions < 4

srijeet0406:
<hr/>

- Traffic Ops

Run TO and `GET` `servers/{id}/deliveryservices`.
Make sure the result contains an array of strings for `geoLimitCountries` in API v4, and a string in all older versions.
Make sure all the tests pass.

- master

## PR submission checklist
- [ ] This PR has tests 
- [ ] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://apache.org/security) for details)

### DIFF
--- a/lib/go-tc/deliveryservices.go
+++ b/lib/go-tc/deliveryservices.go
@@ -846,7 +846,29 @@ func (ds *DeliveryServiceV4) RemoveLD1AndLD2() DeliveryServiceV4 {
 	return *ds
 }
 
-// DowngradeToV3 converts the 4.x DS to a 3.x DS.
+// DowngradeToV2 converts a 4.x DS to a 2.x DS.
+func (ds *DeliveryServiceV4) DowngradeToV2() DeliveryServiceNullable {
+	nullableFields := ds.DeliveryServiceNullableFieldsV11
+	geoLimitCountries := ([]string)(ds.GeoLimitCountries)
+	geo := strings.Join(geoLimitCountries, ",")
+	nullableFields.GeoLimitCountries = &geo
+	return DeliveryServiceNullable{
+		DeliveryServiceNullableV14: DeliveryServiceNullableV14{
+			DeliveryServiceNullableV13: DeliveryServiceNullableV13{
+				DeliveryServiceNullableV12: DeliveryServiceNullableV12{
+					DeliveryServiceNullableV11: DeliveryServiceNullableV11{
+						DeliveryServiceNullableFieldsV11: nullableFields,
+					},
+				},
+				DeliveryServiceFieldsV13: ds.DeliveryServiceFieldsV13,
+			},
+			DeliveryServiceFieldsV14: ds.DeliveryServiceFieldsV14,
+		},
+		DeliveryServiceFieldsV15: ds.DeliveryServiceFieldsV15,
+	}
+}
+
+// DowngradeToV3 converts a 4.x DS to a 3.x DS.
 func (ds *DeliveryServiceV4) DowngradeToV3() DeliveryServiceNullableV30 {
 	nullableFields := ds.DeliveryServiceNullableFieldsV11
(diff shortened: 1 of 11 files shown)


zrhoffman (changes requested):
This looks good, but would you mind also removing this stray comment leftover from #6441?

https://github.com/apache/trafficcontrol/blob/4e8dd7adb0726ab30f59e16a98ccd8b1a7690db1/traffic_ops/testing/api/v4/deliveryservices_test.go#L121

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
```diff
@@ -956,8 +956,11 @@ func (dss *TODSSDeliveryService) Read(h http.Header, useIMS bool) ([]interface{}
 	for _, ds := range dses {
 		if version.Major > 3 && version.Minor >= 0 {
 			ds = ds.RemoveLD1AndLD2()
+			returnable = append(returnable, ds)
+		} else {
+			dsV3 := ds.DowngradeToV3()
+			returnable = append(returnable, dsV3)
```
ocket8888, line 962: all API versions use this handler, so you need to downgrade to v3.0 and v2.0 as well when appropriate

ocket8888: `DowngradeToV3` downgrades to an APIv3.1 structure - _even though it's named `DeliveryServiceNullableV30`!_

From that type's GoDoc:

> Again, this type is an alias that refers to the LATEST MINOR VERSION of API version 3 - NOT API version 3.0 as the name might imply.

zrhoffman (approved):
Stray comment is removed!

lib/go-tc/deliveryservices.go
```diff
@@ -846,7 +846,29 @@ func (ds *DeliveryServiceV4) RemoveLD1AndLD2() DeliveryServiceV4 {
 	return *ds
 }
 
-// DowngradeToV3 converts the 4.x DS to a 3.x DS.
+// DowngradeToV2 converts a 4.x DS to a 2.x DS.
+func (ds *DeliveryServiceV4) DowngradeToV2() DeliveryServiceNullable {
```
ocket8888, line 850: You don't need to write this whole method, I think, the v2 DS can be accessed directly through `DeliveryServiceNullableV30.DeliveryServiceNullableV15`. That's what the `/deliveryservices` handler does.

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
```diff
@@ -957,9 +957,12 @@ func (dss *TODSSDeliveryService) Read(h http.Header, useIMS bool) ([]interface{}
 		if version.Major > 3 && version.Minor >= 0 {
 			ds = ds.RemoveLD1AndLD2()
 			returnable = append(returnable, ds)
-		} else {
+		} else if version.Major > 2 {
 			dsV3 := ds.DowngradeToV3()
 			returnable = append(returnable, dsV3)
```
ocket8888, line 962: this will return APIv3.1 structures in calls to v3.0 - gotta check the minor version

ocket8888 (changes requested):

traffic_ops/traffic_ops_golang/deliveryservice/servers/servers.go
srijeet0406: yeah missed your previous comment, fixing this now

ocket8888 (approved):
LGTM, provided tests pass - which now look like they'd actually fail if this wasn't working.

T3C failures are systemic, I'll probably ignore them. CiaB failures are a known issue with ATS version determination.

PR #6676, opened 2022-03-18
Reduce over-verbose logging

ocket8888:
The "ORT tests" (`t3c` integration tests) print every file extracted from the Go distribution archive and the name of every Go module dependency the project has every time they run. That constitutes about 13,600 lines in a test that outputs 15,000 with the current failure (less on success, but that matters less since you'd probably never look at the logs in that case) which makes it about 90% of the test output. That information has never been helpful yet (afaik) and I'm skeptical it ever will be in the future, so this PR aims to make looking at the test output less annoying by removing it.

<hr/>

- Traffic Control Cache Config (tests)

Check the t3c test workflow action output on this PR, verify the logs from the "ORT Test" step are now roughly 2k lines instead of 15k (assuming the current failure continues).

## PR submission checklist
- [x] This PR has tests


zrhoffman (approved):
Making `tar` and `go mod vendor` non-verbose is fine

zrhoffman:
The Cache-config GHA was failing before this PR, but now it's easier to see what those failures are.

PR #6656, opened 2022-03-15
Update Go version to 1.18

zrhoffman:
## What does this PR (Pull Request) do?
- [x] This PR is not related to any Issue

This PR makes the Go components of Traffic Control build using Go version 1.18 and updates the `golang.org/x/` dependencies..

See the Go 1.18 [release notes](https://go.dev/doc/go1.18):

> <p> The latest Go release, version 1.18, is a significant release, including changes to the language, implementation of the toolchain, runtime, and libraries. Go 1.18 arrives seven months after <a href="https://go.dev/doc/go1.17">Go 1.17</a>. As always, the release maintains the Go 1 <a href="https://go.dev/doc/go1compat">promise of compatibility</a>. We expect almost all Go programs to continue to compile and run as before. </p>

## Which Traffic Control components are affected by this PR?
- CDN in a Box - Enroller
- Grove
- Traffic Control Client (Go)
- Traffic Monitor
- Traffic Ops
- Traffic Ops ORT
- Traffic Stats
- TC Health Client
- CI tests for Go components

## What is the best way to verify this PR?
Run unit tests and API tests. I have compiled our project from Go 1.18 for awhile now and have not run into issues, aside from bugs in the Go compiler that are now fixed.

## The following criteria are ALL met by this PR
- [x] Existing tests are sufficient, no additional tests necessary
- [x] The documentation pulls the Go version from the `GO_VERSION` file
- [x] This PR has a CHANGELOG entry
- [x] This PR includes any and all required license headers
- [x] This PR does not include a database migration
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://www.apache.org/security/) for details)

PR #6677, opened 2022-03-18
Fix build failures resulting from new bulid-time requirements in go1.18

ocket8888:
The developer environment fails to build Go binaries due to an external `git` binary being required for the Go compiler to work in Go 1.18, whenever the compiler can find a `.git` directory in the filesystem hierarchy that contains the working directory. At the time of this writing, there is ongoing discussion in golang/go#51748 that might result in changes being made that would fix this problem for us, but as a resolution hasn't yet been adopted and there's no timetable for when it might be implemented if one were adopted, this just fixes the issue by adding a new build flag explicitly disabling the new feature.

<hr/>

None.

`atc start && atc ready -w` should exit within a few minutes with exit code 0.

- master as of #6656.

## PR submission checklist
- [ ] This PR has tests


shamrickus (approved):
LGTM
* Fixes dev-ciab

PR #6690, opened 2022-03-24
Refactor DeliveryService Tests

ericholguin:
This PR refactors [v4/deliveryservices_test.go](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/testing/api/v4/deliveryservices_test.go) and [v3/deliveryservices_test.go](https://github.com/apache/trafficcontrol/blob/master/traffic_ops/testing/api/v3/deliveryservices_test.go) in order to make tests more granular, providing transparency on what is being tested and reducing the need to dig through test code to confirm what was tested. In addition this new test structure should also make it easier for new tests to be added as well as reduce redundant code.

**All the tests related to delivery service keys, i.e. ssl, url, or uri have been move to their own individual test file for better and cleaner organization.**

One test has been removed as I was unable to replicate it using the API. Can be discussed: [test here](https://github.com/apache/trafficcontrol/blob/7185a0220f26a82e7b7e8579d56527be6baf001b/traffic_ops/testing/api/v4/deliveryservices_test.go#L1490-L1606)

- Adds data driven testing
    - Descriptive test names
- Sub tests for organization
    - Use GoLangs test runner
- Use assert functionality
    - Streamlines code
    - Reduces nested conditionals
- Adds expectation functions
    - Reusable expectations

<hr/>

- Traffic Ops Testing

Run TO Integration tests

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/testing/api/v3/deliveryservices_test.go
+++ b/traffic_ops/testing/api/v3/deliveryservices_test.go
@@ -17,1180 +17,490 @@ package v3
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"net/url"
-	"reflect"
 	"strconv"
-	"strings"
 	"testing"
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-rfc"
 	"github.com/apache/trafficcontrol/lib/go-tc"
-	"github.com/apache/trafficcontrol/lib/go-util"
-	toclient "github.com/apache/trafficcontrol/traffic_ops/v3-client"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/assert"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/utils"
+	"github.com/apache/trafficcontrol/traffic_ops/toclientlib"
 )
 
 func TestDeliveryServices(t *testing.T) {
-	WithObjs(t, []TCObj{CDNs, Types, Tenants, Users, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, Topologies, ServerCapabilities, DeliveryServices}, func() {
-		currentTime := time.Now().UTC().Add(-5 * time.Second)
-		ti := currentTime.Format(time.RFC1123)
-		var header http.Header
-		header = make(map[string][]string)
-		header.Set(rfc.IfModifiedSince, ti)
-		header.Set(rfc.IfUnmodifiedSince, ti)
-		if includeSystemTests {
-			SSLDeliveryServiceCDNUpdateTest(t)
-			GetTestDeliveryServicesURLSigKeys(t)
+	WithObjs(t, []TCObj{CDNs, Types, Tenants, Users, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, Topologies, ServerCapabilities, DeliveryServices, DeliveryServicesRequiredCapabilities, DeliveryServiceServerAssignments}, func() {
+
+		tomorrow := time.Now().AddDate(0, 0, 1).Format(time.RFC1123)
+		currentTime := time.Now().UTC().Add(-15 * time.Second)
+		currentTimeRFC := currentTime.Format(time.RFC1123)
+
+		tenant4UserSession := utils.CreateV3Session(t, Config.TrafficOps.URL, "tenant4user", "pa$$word", Config.Default.Session.TimeoutInSecs)
+
+		methodTests := utils.V3TestCase{
+			"GET": {
+				"NOT MODIFIED when NO CHANGES made": {
+					ClientSession: TOSession, RequestHeaders: http.Header{rfc.IfModifiedSince: {tomorrow}},
+					Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusNotModified)),
+				},
+				"OK when VALID request": {
+					ClientSession: TOSession, Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK)),
+				},
+				"OK when ACTIVE=TRUE": {
+					ClientSession: TOSession, RequestParams: url.Values{"active": {"true"}},
+					Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK), utils.ResponseLengthGreaterOrEqual(1),
+						validateGetDSExpectedFields(map[string]interface{}{"Active": true})),
+				},
+				"OK when ACTIVE=FALSE": {
+					ClientSession: TOSession, RequestParams: url.Values{"active": {"false"}},
+					Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK), utils.ResponseLengthGreaterOrEqual(1),
+						validateGetDSExpectedFields(map[string]interface{}{"Active": false})),
+				},
+				"OK when VALID ACCESSIBLETO parameter": {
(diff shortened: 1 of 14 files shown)


TaylorCFrey (changes requested):
I'm not able to get the tests to pass locally.
```
cd traffic_ops/app
./db/admin --env=test reset
./db/admin -v --env=test reset
cd traffic_ops/testing/api
```

`go test -v ./v4` Fails with the following (edited for brevity)

```
--- FAIL: TestDeliveryServicesKeys (24.69s)
    --- FAIL: TestDeliveryServicesKeys/Create_URL_Signature_keys_for_a_Delivery_Service (0.00s)
    --- FAIL: TestDeliveryServicesKeys/Retrieve_URL_Signature_keys_for_a_Delivery_Service (0.00s)
    --- FAIL: TestDeliveryServicesKeys/Delete_URL_Signature_keys_for_a_Delivery_Service (0.00s)
    --- FAIL: TestDeliveryServicesKeys/Create_URI_Signing_Keys_for_a_Delivery_Service (0.00s)
    --- FAIL: TestDeliveryServicesKeys/Retrieve_URI_Signing_keys_for_a_Delivery_Service (0.00s)
    --- FAIL: TestDeliveryServicesKeys/Delete_URI_Signing_keys_for_a_Delivery_Service (0.00s)
--- FAIL: TestDeliveryServices (6.30s)
    --- FAIL: TestDeliveryServices/GET (0.14s)
        --- FAIL: TestDeliveryServices/GET/OK_when_VALID_PROFILE_parameter (0.00s)
```


`go test -v ./v3` Fails with the following (edited for brevity)

```
    deliveryservices_keys_test.go:164: Not equal. Expected: 1 Actual: 0 Messages: Expected 1 key, got 1
--- FAIL: TestDeliveryServicesKeys (6.79s)
```

traffic_ops/testing/api/v3/deliveryservices_test.go
```diff
@@ -17,1180 +17,499 @@ package v3
...
+	WithObjs(t, []TCObj{CDNs, Types, Tenants, Users, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, CacheGroups, Servers, Topologies, ServerCapabilities, DeliveryServices, DeliveryServicesRequiredCapabilities, DeliveryServiceServerAssignments}, func() {
+
+		tomorrow := time.Now().AddDate(0, 0, 1).Format(time.RFC1123)
+		currentTime := time.Now().UTC().Add(-15 * time.Second)
+		currentTimeRFC := currentTime.Format(time.RFC1123)
+
+		tenant4UserSession := utils.CreateV3Session(t, Config.TrafficOps.URL, "tenant4user", "pa$$word", Config.Default.Session.TimeoutInSecs)
+
+		methodTests := utils.V3TestCase{
+			"GET": {
```
TaylorCFrey, line 43: It might behove us to make these HTTP verbs and switch strings `const` somewhere
```go
const (
    Get = "GET"
    Post = "POST"
    GetAfterChanges = "GET AFTER CHANGES"
    // ...
)
```

Go's capitalization results in scope changes so the convention of ALL CAPS may not be appropriate, but these could be put somewhere so they are accessible everywhere? I will leave that decision up to you. Maybe the HTTP verbs could be someone globally accessible, but the delivery service specific (`DELIVERY SERVICES CAPACITY`) could be local?

I won't put this comment on every verb, but it applies for `GET` `POST` `PUT` `DELETE` etc.

traffic_ops/testing/api/v3/deliveryservices_test.go
```diff
@@ -17,1180 +17,499 @@ package v3
...
+				"OK when VALID ACCESSIBLETO parameter": {
+					ClientSession: TOSession, RequestParams: url.Values{"accessibleTo": {"1"}},
+					Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK), utils.ResponseLengthGreaterOrEqual(1)),
+				},
+				"EMPTY RESPONSE when TENANT attempts reading DS OUTSIDE TENANCY": {
+					ClientSession: tenant4UserSession, RequestParams: url.Values{"xmlId": {"ds3"}},
+					Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK), utils.ResponseHasLength(0)),
+				},
+			},
+			"POST": {
```
TaylorCFrey, line 70: This applies to both v3 and v4 test suites:

Does the order of these tests matter? Especially for endpoints that aren't idempotent (`CUD`, but not `R`)? Granted it wouldn't necessarily be prudent to call `DELETE` randomly, but I mean does the order matter in, say, all the `POST` or `PUT` requests?

traffic_ops/testing/api/v4/deliveryservices_test.go
```diff
@@ -17,2896 +17,772 @@ package v4
...
 	"github.com/apache/trafficcontrol/lib/go-tc"
-	"github.com/apache/trafficcontrol/lib/go-util"
-	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/deliveryservice"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/assert"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/utils"
+	"github.com/apache/trafficcontrol/traffic_ops/toclientlib"
 	client "github.com/apache/trafficcontrol/traffic_ops/v4-client"
 )
 
 func TestDeliveryServices(t *testing.T) {
```
TaylorCFrey, line 34: This is just a general observation and question, but not a requirement for any change.

I don't believe this PR changes how these tests are necessarily run. Even though the previous tests were written as individual tests, they still had a hard requirement and dependency on `WithObjs()`. It appeared like we could have run those tests individually, but with their dependency on the objects being initialized ahead of time, they couldn't actually be done that way.

Since these are now table tests, and the dependency on `WithObjs` is still there, we still don't really have the ability to run these individually, is that correct?

TaylorCFrey (changes requested):
Overall, this is a grand improvement on readability, clealiness, and I believe general mantainability. It seems like adding or modifying tests in the future will be easier. Nice work.

traffic_ops/testing/api/v3/deliveryservices_test.go
ericholguin: No order does not matter. These tests also run randomly by default, goLang doesn't run table driven tests sequentially. I also run these tests on my end with the argument `-shuffle on` in order to make sure they don't overlap or can run without any dependency on other tests.

traffic_ops/testing/api/v4/deliveryservices_test.go
ericholguin: Yeah with these changes `WithObjs`  is still run first. The `WithObjs` does contain all the prerequisites for the tests and is needed still. However you can run any of the test defined in the test file by its name. So for example if you just want to run the POST tests for Delivery Services all you need to do is: 

`go test -v  -run "^TestDeliveryServices$/POST" ./v4` 
(I add ^ and $ due to other tests having DeliveryServices in their test name and golang uses regex for finding what test needs to be ran)

You can also get more precise and run a specific test:
`go test -v  -run "^TestDeliveryServices$/GET/BAD_REQUEST_when_TOPOLOGY_DOESNT_EXIST" ./v4` 

So if you only want to run a specific test you can.

traffic_ops/testing/api/v3/deliveryservices_test.go
TaylorCFrey: This is only a suggestion. Perhaps if you find yourself using these value consistently through other tests it could be done at that time.

PR #6696, opened 2022-03-28
Add E2E testing to TPv2

shamrickus:
This PR implements E2E testing in TPv2 using `NightwatchJS`. Specifically it implements the tests for the login page.
<hr/>

- Traffic Portal v2
- Automation 

Verify GHA passes, ensure e2e tests pass locally. Also check out the README to make sure the instructions make sense

## PR submission checklist
- [x] This PR has tests 


experimental/traffic-portal/package.json
```diff
@@ -80,26 +82,30 @@
     "@angular-eslint/template-parser": "13.1.0",
     "@angular/cli": "^13.2.0",
     "@angular/compiler-cli": "^13.2.0",
+		"@nightwatch/schematics": "^0.1.9",
```
ocket8888, line 85: extra indent

experimental/traffic-portal/package.json
```diff
@@ -80,26 +82,30 @@
...
     "eslint-plugin-prefer-arrow": "^1.2.3",
+    "geckodriver": "3.0.1",
     "jasmine-core": "^3.10.1",
     "karma": "~6.3.2",
     "karma-chrome-launcher": "~3.1.0",
     "karma-coverage": "~2.2.0",
     "karma-jasmine": "~4.0.0",
     "karma-jasmine-html-reporter": "^1.5.0",
-    "protractor": "~7.0.0",
+    "nightwatch": "2.0.0-beta.3",
```
ocket8888, line 108: should we really be using a beta version?

experimental/traffic-portal/nightwatch/globals/globals.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+const config = {
+	chrome_headless: {},
+	default: {
+		adminPass: "twelve12",
+		adminUser: "admin",
+		trafficOpsURL: "https://localhost:6443"
+	}
+};
+config.chrome_headless = config.default;
```
ocket8888, line 33: idk if this is actually any better, but you could do this inline in the `config` literal as a `get` literal, fwiw. It's a little more convoluted, but it makes the typing correct - the exported `config` has the type:
```typescript
{
    chrome_headless: {},
    default: {
        adminPass: string,
        adminUser: string,
        trafficOpsUrl: string
    }
}
```
because `typeof config.default` is assignable to `{}` the assignment on this line succeeds, but anyone trying to access properties of `chrome_headless` won't be able to because its type has no properties. Although that doesn't appear to actually matter, for whatever reason.

ocket8888: I suppose you could also just put `as typeof this.default` on the initialization of `chrome_headless`. I just hate using `as` so I always try to figure out how to not use it.

.github/actions/tpv2-integration-tests/tpv2-integration-tests.ts
```diff
@@ -0,0 +1,31 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+"use strict";
+import child_process, { SpawnSyncOptions, SpawnSyncReturns } from "child_process";
```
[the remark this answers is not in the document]
shamrickus: This is literally the same file that all integration tests use, honestly not even sure why we're using JS runners seems like it's way more code and we do nothing with it.

experimental/traffic-portal/nightwatch/globals/globals.ts
shamrickus: It isn't, because you should _never_ reference this object yourself. It should be accessed through `browser.globals` since the type changes at runtime.

.github/actions/tpv2-integration-tests/tpv2-integration-tests.ts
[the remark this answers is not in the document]
ocket8888: Yeah, you know, I couldn't figure that out either, and at this point I'm just scared to ask since I already approved several personally (I think). I guess I just figured it's what everyone wants to do. I wouldn't bother.

experimental/traffic-portal/package.json
shamrickus: `2.1.0` breaks the tests and the [nightwatch schematic explicitly](https://github.com/nightwatchjs/nightwatch-schematics/blob/d69342bae71afb4f25918b004b88a540d685b845/src/schematics/nightwatch/utility/util.ts#L158) adds the [`next` version](https://www.npmjs.com/package/nightwatch?activeTab=versions) (2.0.0-beta.3) explicitly. Also `2.1.0` currently needs `chromedriver:99` _which is wrong_ (it's not even at a LTS release of chrome), leading me to believe these tags are not correct.

ocket8888: Wonderful. I figured it was something like that, I was just hoping I was wrong.

experimental/traffic-portal/nightwatch/globals/globals.ts
shamrickus: > I suppose you could also just put `as typeof this.default` on the initialization of `chrome_headless`. I just hate using `as` so I always try to figure out how to not use it.

I agree, however with how the typings for nightwatch are currently setup I don't think this is possible.

ocket8888: I suppose it doesn't matter since its an injected dependency and you can just import the type exported here.

ocket8888 (approved):

PR #6703, opened 2022-03-29
Don't check for locks on dequeue

srijeet0406:
This PR is not related to any issue. It removes the checking for CDN locks when trying to dequeue updates.

<hr/>

- Traffic Ops

Run TO and TP locally.
As user X, lock CDN A.
Queue updates on CDN A as user X.
Then as user Y, try clearing these updates on CDN A, while it is still locked by X.
This should succeed.

Make sure all unit/ API tests pass.

- master

## PR submission checklist
- [x] This PR has tests 


ocket8888:
Should this only ignore *soft* locks for dequeue? A hard lock is supposed to prevent editing servers, right? Queueing updates will change two fields (`lastUpdated` and `updPending` - once the changes in #6569 are merged it'll also change the update timestamp field)

srijeet0406:
> Should this only ignore _soft_ locks for dequeue? A hard lock is supposed to prevent editing servers, right? Queueing updates will change two fields (`lastUpdated` and `updPending` - once the changes in #6569 are merged it'll also change the update timestamp field) on affected servers.

Resolved offline.

traffic_portal/app/src/common/modules/dialog/select/lock/dialog.select.lock.tpl.html
```diff
@@ -51,7 +51,7 @@ <h4 class="modal-title">Create CDN Lock</h4>
                         </label>
                         <input id="soft" name="lockType" type="radio" ng-model="lock.soft" ng-value="true" required>
                         <label class="has-tooltip" for="hard">Hard<div class="helptooltip">
-                            <div class="helptext">A <dfn>hard</dfn> lock will prevent others from queueing server updates or snapshotting the locked CDN. It will also block others from making changes that would change the state of the locked CDN.
+                            <div class="helptext">A <dfn>hard</dfn> lock will prevent others from queueing server updates or snapshotting the locked CDN. It will also block others from making changes(except clearing updates) that would change the state of the locked CDN.
```
ocket8888, line 54: There should be a space between `changes` and `(except`

ocket8888 (changes requested):
Assuming this works - I'm about to test - it's pretty much good to go.

ocket8888 (approved):

PR #6708, opened 2022-03-31
adding strategies peering ring generation to t3c

jrushford:
Adds in to t3c strategies peering support when the DS profile parameter `use_peering` exists and is `true`

- Add's a peer group list in strategies.yaml
- adds the cache_peer_result flag to a strategy configured to use_peering
- sets the ring_mode on a strategy to `peering_ring` when configured to use_peering
- adds a DS parameter **use_peering**

<hr/>

Add the DS param **use_peering** to a Delivery service.  It should generate a strategy in strategies.yaml with a peer group set of hosts and the ring_mode should be set to **peering_ring**

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/lib/go-atscfg/parentdotconfig.go
+++ b/lib/go-atscfg/parentdotconfig.go
@@ -65,6 +65,7 @@ const ParentConfigCacheParamPort = "port"
 const ParentConfigCacheParamUseIP = "use_ip_address"
 const ParentConfigCacheParamRank = "rank"
 const ParentConfigCacheParamNotAParent = "not_a_parent"
+const StrategyConfigUsePeering = "use_peering"
 
 type OriginHost string
 type OriginFQDN string
@@ -237,6 +238,7 @@ func makeParentDotConfigData(
 
 	nameTopologies := makeTopologyNameMap(topologies)
 
+	cgPeers := map[int]Server{}   // map[serverID]server
 	cgServers := map[int]Server{} // map[serverID]server
 	for _, sv := range servers {
 		if sv.ID == nil {
@@ -258,6 +260,15 @@ func makeParentDotConfigData(
 		if *sv.CDNName != *server.CDNName {
 			continue
 		}
+		// save cachegroup peer servers
+		if *sv.CDNName == *server.CDNName && *sv.Cachegroup == *server.Cachegroup {
+			if *sv.Status == string(tc.CacheStatusReported) || *sv.Status == string(tc.CacheStatusOnline) {
+				if _, ok := cgPeers[*sv.ID]; !ok {
+					cgPeers[*sv.ID] = sv
+				}
+			}
+			continue
+		}
 		if _, ok := parentCacheGroups[*sv.Cachegroup]; !ok {
 			continue
 		}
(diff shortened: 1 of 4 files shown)


lib/go-atscfg/parentdotconfig.go
```diff
@@ -839,7 +865,11 @@ func getParentDSParams(ds DeliveryService, profileParentConfigParams map[string]
 	if !ok {
 		return params, warnings
 	}
-
+	if val, ok := dsParams[StrategyConfigUsePeering]; ok {
+		if val == "true" {
```
rob05c, line 869: Nitpick: this should probably ignore the value, and consider it always true if the Param exists, to be consistent with the behavior of other boolean params

lib/go-atscfg/strategiesdotconfig.go
```diff
@@ -162,11 +164,18 @@ func getStrategyErrorCodes(codes []int) string {
...
 func getStrategyGroups(svc *ParentAbstractionService) string {
 	txt := ""
-	if len(svc.Parents) != 0 {
-		txt += "\n" + `      - *group_parents_` + svc.Name
-	}
-	if len(svc.SecondaryParents) != 0 {
-		txt += "\n" + `      - *group_secondary_parents_` + svc.Name
+	if getStrategySecondaryMode(svc.SecondaryMode) == "peering_ring" {
+		txt += "\n" + `      - *peers_group`
+		if len(svc.Parents) != 0 {
```
rob05c, line 169: Nitpick: the `group_parents_` is duplicated in both cases, it could be de-duplicated and moved outside the if-statement

lib/go-atscfg/strategiesdotconfig.go
```diff
@@ -182,6 +191,9 @@ func getStrategyStrategiesSection(pa *ParentAbstraction) string {
 			txt += "\n" + `    hash_key: path`
 		}
 		txt += "\n" + `    go_direct: ` + strconv.FormatBool(svc.GoDirect)
+		if getStrategySecondaryMode(svc.SecondaryMode) == "peering_ring" {
```
rob05c, line 194: Nitpick: `"peering_ring"` should probably be put in a constant somewhere. Probably `lib/go-atscfg` so other things can use it

lib/go-atscfg/strategiesdotconfig.go
```diff
@@ -256,6 +275,22 @@ func getStrategyHostsSection(pa *ParentAbstraction) string {
 			txt += getStrategyHostsSectionHost(svc, parent)
 		}
 	}
+	if len(pa.Peers) != 0 {
+		count := 1
+		for _, peer := range pa.Peers {
```
rob05c, line 280: Nitpick: could use the loop index to avoid a separate variable here, ` for i, peer := range pa.Peers { txt += getStrategyHostsSectionPeer(peer, i+1)`

rob05c (approved):
LGTM, thanks!

All my comments are minor. 

It has docs. Could use a changelog and tests, but this feature should be considered "experimental" for a while anyway, so I think it's ok to add them later

PR #6721, opened 2022-04-05
t3c-generate fix for combining ds remap config params

traeak:
Fixes poor use of append() call in t3c-generate wrappers.go

<hr/>

6.2.0, etc

## PR submission checklist
- [x] This PR has tests 


rob05c (approved):
LGTM, thanks!

This isn't possible or reasonable to test, because append may or may not modify the existing slice, trying to test that would be contrived if possible at all, and testing for every slice append isn't reasonable.

PR #6724, opened 2022-04-06
updates to the trafficcontrol-health-client

jrushford:
updates to the trafficcontrol-health-client:

- updates the host status commands used by the health-client
  to support both ATS 9 and ATS 10 traffic_ctl commands.
- updates the parsing of host statuses output to support both
  ATS 9 and ATS 10 output formats.
- adds a **markup-poll-threshold** config variable that may be used to control 
   how many poll intervals must pass with a healthy cache status before the cache
   is marked up.  The idea here is to reduce flapping between marking up and down
   a cache

The health-client has been updated to process ATS version 9 and ATS version 10-Dev traffic_ctl host
commands.  Verify the health client can verify ATS host statuses and mark up or down caches
wether the cache is running ATS 9 or 10-Dev.  This functionality has been tested by me.

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/tc-health-client/README.md
+++ b/tc-health-client/README.md
@@ -71,24 +71,24 @@ Requires Apache TrafficServer 8.1.0 or later.
 
 -f, -\-config-file=config-file 
   
-  Specify the config file to use.  
-  Defaults to /etc/trafficcontro-health-client/tc-health-client.json
+Specify the config file to use.  
+Defaults to /etc/trafficcontro-health-client/tc-health-client.json
 
 -h, -\-help 
 
-  Prints command line usage and exits
+Prints command line usage and exits
 
 -l, -\-logging-dir=logging-directory
 
-  Specify the directory where log files are kept.  The default location
-  is **/var/log/trafficcontrol/**
+Specify the directory where log files are kept.  The default location
+is **/var/log/trafficcontrol/**
 
 -v, -\-verbose
 
-  Logging verbosity.  Errors are logged to the default log file 
-  **/var/log/trafficcontrol/tc-health-client.log**
-  To add Warnings, use -v.  To add Warnings and Informational 
-  logging, use -vv.  Finally you may add Debug logging using -vvv.
+Logging verbosity.  Errors are logged to the default log file 
+**/var/log/trafficcontrol/tc-health-client.log**
+To add Warnings, use -v.  To add Warnings and Informational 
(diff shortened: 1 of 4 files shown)


tc-health-client/README.md
```diff
@@ -176,6 +177,12 @@ Sample configuarion file:
   will be marked down when the number of consecutive polls reaches this threshold
   with the parent reported as unhealthy.  The default threshold is 2.
 
+### markup-poll-threshold
+
+  This controls when an healthy parent is marked up.  An healthy parent
```
jpappa200, line 182: Typo "An"

jpappa200 (changes requested):
One typo otherwise looks good to me.

tc-health-client/README.md
ocket8888: FWIW that also has some extra spaces at the beginning of the line. It doesn't affect rendering, they just get stripped so they're totally superfluous.

jrushford: @ocket8888 all the non mark lines with leading spaces?  Are you asking that I remove them?

jrushford: @ocket8888 I removed the leading spaces.  @jpappa200 I fixed the typo

jpappa200 (approved):
👍

tc-health-client/README.md
ocket8888: Oh, it would probably be most consistent to remove them everywhere then, but to be clear I'm not asking for anything, it's just something I noticed when I looked at the notification I got for an update on this PR.

tc-health-client/tmagent/tmagent.go
```diff
@@ -49,6 +49,14 @@ const (
 	StrategiesFile = "strategies.yaml"
 )
 
+// this global is used to auto select the
+// proper ATS traffic_ctl command to use
+// when querying host status. for ATS
+// version 10 and greater this will remain
+// at 0.  For ATS version 9, this will be
+// auto updated to 1
+var traffic_ctl_index = 0
```
rob05c, line 58: Nitpick: Go symbols should be consistentCased, i.e. `var trafficCtlIndex = 0`.
https://go.dev/doc/effective_go#mixed-caps

It'd also be ideal to avoid global variables, and pass it locally instead. But for what's effectively a singleton in a small app, not a huge deal

rob05c (approved):
LGTM, and I trust Joe's review

PR #6730, opened 2022-04-08
t3c race condition timestamp updates

rawlinp:
Remove the ability to set "update" times via the
/servers/:hostname/update endpoint (only "apply" times are writable via
this endpoint).

Fix docs to reflect the existence of the "updated" and "reval_updated"
query paramters in the /servers/:hostname/update endpoint (kept for
backwards-compatibility).

Correct the date on the DB migration so that it applies after migrations
that were merged earlier.

Related: #6569

<hr/>

- Documentation
- Traffic Ops

Verify that you cannot update the `config_update_time` or `revalidate_update_time` server fields via the `/servers/:hostname/update` endpoint.

## PR submission checklist
- [ ] This PR has tests 
- [x] Feature already contains a changelog entry


traffic_ops/traffic_ops_golang/server/update.go
```diff
@@ -155,83 +153,55 @@ func parseQueryParams(params map[string]string) (*updateValues, error) {
...
-	configUpdateTimeParam, hasConfigUpdateTimeParam := params["config_update_time"]
-	revalidateUpdateTimeParam, hasRevalidateUpdateTimeParam := params["revalidate_update_time"]
 	configApplyTimeParam, hasConfigApplyTimeParam := params["config_apply_time"]
 	revalidateApplyTimeParam, hasRevalidateApplyTimeParam := params["revalidate_apply_time"]
 
 	if !hasConfigApplyTimeParam && !hasRevalidateApplyTimeParam &&
-		!hasConfigUpdateTimeParam && !hasRevalidateUpdateTimeParam &&
 		!hasConfigUpdatedBoolParam && !hasRevalUpdatedBoolParam {
-		return nil, errors.New("must pass at least one query parameter: 'config_apply_time', 'revalidate_apply_time', 'config_update_time', 'revalidate_update_time' (may also pass bool `update` `reval_updated`)")
+		return nil, errors.New("must pass at least one of the following query parameters: 'config_apply_time', 'revalidate_apply_time' ,'update', 'reval_updated'")
```
mattjackson220, line 161: only the smallest comment. this should have `updated` instead of `update`

mattjackson220 (changes requested):
one small comment but other than that this all looks good to me! tested locally and works as expected

traffic_ops/traffic_ops_golang/server/update.go
rawlinp: DOH. Fixed.

PR #6733, opened 2022-04-08
T3C: parent.config fqdn dup check enhanced to include string compare

traeak:
This PR improves the logic of duplicate origin detection during parent.config generation.
Instead of just reporting a dest_domain name collision and picking the first, this PR does the following:

- Changes the duplicate check from dest_domain to dest_domain:port
- Does a string compare on dup detection:
--  If strings same, warn and continue.  
--  If strings different, abort parent.config generation

<hr/>

- Documentation

Unit test added in parentdotconfig_test.go

## PR submission checklist
- [x] This PR has tests 
- [] This PR has documentation - current behavior not documented?


traeak:
Probably too invasive and will break existing users.

PR #6744, opened 2022-04-11
Fix t3c nil reference on fallback

rob05c:
Fixes a nil reference in t3c on fallback.

Run t3c against a TO that doesn't serve the latest API, verify fallback does not crash.

Not in a release

## PR submission checklist
- ~[x] This PR has tests~ no tests, testing would require the Integration Test Framework support multiple TO versions, which would be ideal but is a large amount of work to add.
- ~[x] This PR has a CHANGELOG.md entry~ no changelog, bug is not in a release


jpappa200 (approved):
👍

PR #6748, opened 2022-04-13
Fix TM issue where CRConfig is never updated

rawlinp:
PR #6625 reduced the TM dependency on CRConfig data, but it accidentally
caused TM to no longer poll for new CRConfig snapshots. This adds the
CRConfig request back into the polling loop so that it updates its
in-memory copy for internal use (if necessary) and for serving to TR.

<hr/>

- Traffic Monitor

Start TM, then take a new snapshot, and ensure that TM updates its crconfig.backup file with the new snapshot and serves the new snapshot via its `/publish/CrConfig` endpoint.

- master

## PR submission checklist
- [ ] This PR has tests 


ocket8888 (approved):
Tests pass now that we have a workaround for the VCS issue

PR #6752, opened 2022-04-13
fixed multiple cookies

mattjackson220:
Currently, the mojolicious cookie gets set at login and refreshed when non-login endpoints are hit. However, there is a bug where the cookie is set multiple times depending on the version of the API used. So for 2.0, 3.0, 4.0 it only sets it once but for 3.1 and any future *.1 it sets it twice, and any *.2 it sets it 3 times, etc.

This PR moves the middleware addition so it is only added once per route instead of for each minor version.

<hr/>

- Traffic Ops

Run TO on master and hit a non-login endpoint on version 4.0 and 3.1
Verify that on 4.0 it returned only one `set-cookie` message for the mojo cookie and on 3.1 it set 2
Run TO off this branch and verify that all API versions only set one cookie
Verify that the rest of the middleware works as expected

## PR submission checklist
- [x] This PR has tests 


srijeet0406 (approved):
Code looks good. 
Unit tests and API tests pass.
Manually tested and verified that only one mojo cookie was being set.

PR #6754, opened 2022-04-13
Convert Traffic Ops into a Service Oriented architecture(SOA) product

srijeet0406:
This PR is not related to any issue.
It adds the ability in Traffic Ops, so that it can be run in conjunction with other backend services, wherein TO would just act as a reverse proxy and forward the requests coming in on the specified paths to the specified backends.
<hr/>

- Documentation
- Traffic Ops

 Set up a simple backend service that servers a couple of endpoints (say `foo` and `foos`) and run it on a couple of different ports(say, 8444 and 8445). Now, run Traffic Ops with the new config option and make sure that any request to Traffic Ops on the specified endpoints (in the new config) gets forwarded to the backend service. Also make sure that the user sees the response from the backend service.
 The new config option should be provided to traffic ops like this:
 `
-backendcfg <File name>`
 The new config file should look something like this:
 ```json
{
  "routes": [
    {
      "path": "^/api/4.0/foo/?$",
      "method": "GET",
      "hosts": [
        "localhost:8444"
      ],
      "insecure": true,
      "permissions": [
        "CDN:READ"
      ],
      "routeId": 123456,
      "opts": {
        "alg": "roundrobin"
      }
    },
    {
      "path": "^/api/4.0/foos/?$",
      "method": "GET",
      "hosts": [
        "localhost:8444"
      ],
      "insecure": true,
      "permissions": [
        "CDN:READ"
      ],
      "routeId": 123457,
      "opts": {
        "alg": "roundrobin"
      }
    }
  ]
}
```
Make sure all the tests pass.

- master

## PR submission checklist
- [ ] This PR has tests 

### CODE REVIEW

File: docs/source/admin/traffic_ops.rst [hunk 1]
```diff
@@ -600,9 +600,11 @@ This file deals with the configuration parameters of running Traffic Ops as a re
 
 	:path:              The endpoint that will be served by the backend, for example, `api/4.0/foo`.
 	:method:            The HTTP method for the above mentioned path, for example, `GET` or `PUT`.
-	:route_id:          The integral identifier for the new route being added.
+	:routeId:           The integral identifier for the new route being added.
 	:hosts:             An array of the hosts and ports where the request (if matched) needs to be forwarded to, for example, `cdn-foo-backend-service-host:9090`.
-	:insecure:          A boolean specifying whether or not to enable `InsecureSkipVerify`.
+	:insecure:          A boolean specifying whether or not to enable `InsecureSkipVerify`. This is an optional parameter, defaulting to `false` when not present.
+	:privLevel:         An integer designating the minimum privilege level required to use this API route.
```
ocket8888, line 606: privLevel is deprecated, we shouldn't use it moving forward.

traffic_ops/traffic_ops_golang/routing/routing.go
```diff
@@ -280,8 +307,88 @@ func Handler(
...
+				})
+				if backendRoute.Insecure {
+					rp.Transport = &http.Transport{
+						TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+					}
+				} else {
+					rp.Transport = &http.Transport{
+						TLSClientConfig: &tls.Config{},
+					}
+				}
```
rawlinp, line 332: This can be simplified to just 
```
rp.Transport = &http.Transport{
    TLSClientConfig: &tls.Config{InsecureSkipVerify: insecure}
}
```

traffic_ops/traffic_ops_golang/routing/routing.go
```diff
@@ -280,8 +307,88 @@ func Handler(
...
+	var errCode int
+	var user auth.CurrentUser
+	var inf *api.APIInfo
 
-	catchall.ServeHTTP(w, r)
+	user, userErr, sysErr, errCode = api.GetUserFromReq(w, r, cfg.Secrets[0])
+	if userErr != nil || sysErr != nil {
+		return userErr, sysErr, errCode
+	}
+	v := api.GetRequestedAPIVersion(r.URL.Path)
```
rawlinp, line 373: Why do we need to check the API version?

traffic_ops/traffic_ops_golang/routing/routing.go
```diff
@@ -280,8 +307,88 @@ func Handler(
...
+					}
+				}
+				rp.ErrorHandler = func(w http.ResponseWriter, r *http.Request, err error) {
+					api.HandleErr(w, r, nil, http.StatusInternalServerError, nil, err)
+					return
+				}
+				routeCtx := context.WithValue(ctx, api.DBContextKey, db)
+				routeCtx = context.WithValue(routeCtx, api.PathParamsKey, map[string]string{})
+				r = r.WithContext(routeCtx)
+				r.Header.Add(middleware.RouteID, strconv.Itoa(backendRoute.ID))
```
rawlinp, line 340: why do we need to add the routeID as a request header?

rawlinp (changes requested):

docs/source/admin/traffic_ops.rst
```diff
@@ -586,6 +590,29 @@ This file sets authentication options for connections to Riak when used as the T
...
+
+:routes: This is an array of options to configure Traffic Ops to forward requests of specified types to the appropriate backends.
+
+	:path:              The endpoint that will be served by the backend, for example, `api/4.0/foo`.
+	:method:            The HTTP method for the above mentioned path, for example, `GET` or `PUT`.
+	:routeId:           The integral identifier for the new route being added.
+	:hosts:             An array of the hosts and ports where the request (if matched) needs to be forwarded to, for example, `cdn-foo-backend-service-host:9090`.
+	:insecure:          A boolean specifying whether or not to enable `InsecureSkipVerify`. This is an optional parameter, defaulting to `false` when not present.
+	:permissions:       An array of permissions(strings) specifying the permissions required by the user to use this API route.
```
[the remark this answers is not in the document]
srijeet0406: Yeah, that's how the permissions model works on our normal routes as well.

traffic_ops/traffic_ops_golang/routing/routing.go
srijeet0406: We need it so that the middleware can log the route ID in the access logs.

srijeet0406: Yep, sure can. I've changed it in both the places this was being set as a header.

traffic_ops/traffic_ops_golang/traffic_ops_golang.go
```diff
@@ -234,6 +246,13 @@ func main() {
 
 	reloadProfilingConfig := func() {
 		setNewProfilingInfo(*configFileName, &profiling, &profilingLocation, cfg.Version)
+		backendConfig, err = setNewBackendConfig(backendConfigFileName)
+		if err != nil {
+			log.Errorf("could not reload backend config: %v", err)
+		} else {
+			d.BackendConfig = backendConfig
```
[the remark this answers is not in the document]
srijeet0406: Yeah I was trying to pass all the data in the `ServerData` object, which I now realize isn't necessary. I have modified this.

rawlinp (changes requested):

docs/source/admin/traffic_ops.rst
```diff
@@ -586,6 +590,34 @@ This file sets authentication options for connections to Riak when used as the T
...
 
+.. _backends.conf:
+
+backends.conf
+"""""""""""""
+This file deals with the configuration parameters of running Traffic Ops as a reverse proxy for certain endpoints that need to be served externally by other backend services. It is a JSON-format set of options and their respective values. `traffic_ops_golang`_ will use whatever file is specified (if any) by its :option:`--backendcfg` option. The keys of the file are described below.
+
+:routes: This is an array of options to configure Traffic Ops to forward requests of specified types to the appropriate backends.
+
+	:path:              The regex matching the endpoint that will be served by the backend, for example, `^api/4.0/foo?$`.
```
ocket8888, line 601: fyi there's a specific [Sphinx text role for regular expressions](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-regexp) - I think you meant for this to be monospace font e.g. using the `<code>` HTML element, but actually single grave accents (<kbd>\`</kbd>) uses the "default" role which is emphasis, so you're trying to render `^api/4.0/foo?$` but you're getting *^api/4.0/foo?$*

tl;dr: use `:regexp:` for regular expressions, use double grave accents (<kbd>\`</kbd>) for code.

docs/source/admin/traffic_ops.rst
```diff
@@ -586,6 +590,34 @@ This file sets authentication options for connections to Riak when used as the T
...
+	:method:            The HTTP method for the above mentioned path, for example, `GET` or `PUT`.
```
ocket8888, line 602: same as above RE: double vs single grave accents

docs/source/admin/traffic_ops.rst
```diff
@@ -586,6 +590,34 @@ This file sets authentication options for connections to Riak when used as the T
...
+	:routeId:           The integral identifier for the new route being added.
+	:hosts:             An array of the host object, which specifies the protocol, hostname and port where the request (if matched) needs to be forwarded to.
+
+		:protocol:     The protocol/scheme to be followed while forwarding the requests to the backend service.
+		:hostname:     The hostname of the server where the backend service is running.
+		:port:         The port (integer) on the backend server where the service is running.
+
+	:insecure:          A boolean specifying whether or not TO should verify the backend server's certificate chain and host name. This is not recommended for production use. This is an optional parameter, defaulting to `false` when not present.
```
ocket8888, line 610: same as above RE: double vs single grave accents

docs/source/admin/traffic_ops.rst
```diff
@@ -586,6 +590,34 @@ This file sets authentication options for connections to Riak when used as the T
...
+	:permissions:       An array of permissions (strings) specifying the permissions required by the user to use this API route.
+	:opts:              A collection of key value pairs to control how the requests should be forwarded/ handled, for example, `"alg": "roundrobin"`. Currently, only `roundrobin` is supported (which is also the default if nothing is specified) by Traffic Ops.
```
ocket8888, line 612: same as above RE: double vs single grave accents

docs/source/admin/traffic_ops.rst
```diff
@@ -586,6 +590,34 @@ This file sets authentication options for connections to Riak when used as the T
 
 .. impl-detail:: The name of this file is derived from the current database used in the implementation of Traffic Vault - `Riak KV <https://riak.com/products/riak-kv/index.html>`_.
 
+.. _backends.conf:
```
ocket8888, line 593: it's not hurting anything but this label is unused.
```rst
backends.conf_
```
is a link to a section titled `backends.conf` on the current page, not a link to a label. `:ref:` is used to refer to arbitrary labels from any page/section.

traffic_ops/app/conf/production/backends.conf
```diff
@@ -0,0 +1,48 @@
+{
+  "routes": [
```
ocket8888, line 2: Ah, it seems your editor somehow has become misconfigured to insert spaces instead of indentation. That happens sometimes, but it's usually easy to fix, so you can go back to indenting things instead of spacing them.

Also: shouldn't `routes` be empty in a default configuration? Based on the values below, I'm not sure exactly what I'd expect the configuration to do. In particular it's a bit concerning that the docs strongly recommend that `insecure` not be set to `true` in production, but this production configuration does exactly that.

traffic_ops/traffic_ops_golang/config/config.go
```diff
@@ -36,6 +36,31 @@ import (
...
+	Method      string   `json:"method"`
+	Hosts       []Host   `json:"hosts"`
+	Opts        Options  `json:"opts"`
+	ID          int      `json:"routeId"`
+	Insecure    bool     `json:"insecure"`
+	Permissions []string `json:"permissions"`
+	Index       int
+}
+
+type BackendConfig struct {
```
ocket8888, line 60: GoDocs pls

ocket8888 (commented):
I didn't mean to look at anything but the docs, but then I looked at the configuration that it was based on and wound up glancing at most everything. I'm not looking too hard at any of the logic though - I'm not trying to make a hostile takeover of the review.

rawlinp (approved):

PR #6756, opened 2022-04-14
Fix t3c TO API fallback

rob05c:
Fix t3c nil pointer crash on TO API fallback

Also fixes fallback with cookie login

Unfortunately, as before, these things can't be tested without an Integration Test Framework that supports multiple TO versions. Which we'd really like, but it's also a lot of work to add.

I manually tested against both a TO with the latest API (4.x), and a TO only serving the previous (3.x), both succeeded and generated config correctly, previous fell back as expected for both initial username and subsequent cookie login, latest did not fall back.

Run t3c, verify config is generated correctly against a latest TO, verify config is generated correctly against a TO not serving the latest major API.

Bug, but not in a release.

## PR submission checklist
- ~[x] This PR has tests~ framework doesn't exist, would be a large undertaking to add 
- ~[x] This PR has documentation~ no interface change 
- ~[x] This PR has a CHANGELOG.md entry~ not in a release 

### DIFF
--- a/cache-config/t3cutil/toreq/client.go
+++ b/cache-config/t3cutil/toreq/client.go
@@ -48,8 +48,22 @@ import (
 )
 
 type TOClient struct {
-	c          *toclient.Session
-	old        *toreqold.TOClient
+	// c is the "new" Traffic Ops client, for the latest major API.
+	//
+	// This MUST NOT be accessed without checking for nil.
+	// If the Traffic Ops server doesn't support the latest API, it will fall back,
+	// and c will be nil and old will not, and old must be used.
+	//
+	// **WARNING** This MUST NOT be accessed without checking for nil. See above.
+	c *toclient.Session
+
+	// old is the "old" Traffic Ops client, for the previous major API.
+	// This will be nil unless the Traffic Ops Server doesn't support the latest API,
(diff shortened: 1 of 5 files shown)


cache-config/t3cutil/toreq/client.go
```diff
@@ -48,8 +48,22 @@ import (
...
-	c          *toclient.Session
-	old        *toreqold.TOClient
+	// c is the "new" Traffic Ops client, for the latest major API.
+	//
+	// This MUST NOT be accessed without checking for nil.
+	// If the Traffic Ops server doesn't support the latest API, it will fall back,
+	// and c will be nil and old will not, and old must be used.
+	//
+	// **WARNING** This MUST NOT be accessed without checking for nil. See above.
+	c *toclient.Session
```
rob05c, line 58: This is something of a footgun. This being the second panic we've had recently. It'd be really nice to make it impossible to happen, but that'd require quite a bit of abstraction, it'd be a lot of code. So for now, I added some giant warning comments on the symbol. If it happens again, we can reconsider if the abstraction to make it impossible is worth it.

jpappa200 (approved):
👍

jpappa200 (approved):

PR #6757, opened 2022-04-15
updated cert expirations to only show Active DSes

mattjackson220:
This PR updates the certificate expirations endpoint to only show certs for Active delivery services.

<hr/>

- Traffic Ops

Run TO and TP
Check that the Cert Expirations page works
Set one of the DSes to Not Active
Verify that it no longer shows up on the Cert Expirations Page

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
+++ b/traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
@@ -180,6 +180,22 @@ func (p *Postgres) GetExpirationInformation(tx *sql.Tx, ctx context.Context, day
 		fedList = append(fedList, fedString)
 	}
 
+	inactiveQuery := "SELECT xml_id FROM deliveryservice WHERE NOT active"
+	iaRows, err := tx.Query(inactiveQuery)
+	if err != nil {
+		return []tc.SSLKeyExpirationInformation{}, err
+	}
+	defer iaRows.Close()
+
+	var inactiveList []string
+	for iaRows.Next() {
+		var inactiveXmlId string
+		if err = iaRows.Scan(&inactiveXmlId); err != nil {
+			return []tc.SSLKeyExpirationInformation{}, err
+		}
+		inactiveList = append(inactiveList, inactiveXmlId)
+	}
+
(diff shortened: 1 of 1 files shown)


traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -199,12 +215,10 @@ func (p *Postgres) GetExpirationInformation(tx *sql.Tx, ctx context.Context, day
 		if err = rows.Scan(&expirationInfo.DeliveryService, &expirationInfo.CDN, &expirationInfo.Provider, &expirationInfo.Expiration); err != nil {
 			return []tc.SSLKeyExpirationInformation{}, err
 		}
-		expirationInfo.Federated = false
-		for _, fed := range fedList {
-			if fed == expirationInfo.DeliveryService {
-				expirationInfo.Federated = true
-			}
+		if util.ContainsStr(inactiveList, expirationInfo.DeliveryService) {
```
srijeet0406, line 218: If we have a large number of inactive delivery services, this call might have to iterate over a number of items trying to find a match. Do you think a map could be better, in terms of reducing the lookup time, etc.?

traffic_ops/traffic_ops_golang/trafficvault/backends/postgres/postgres.go
```diff
@@ -199,12 +215,10 @@ func (p *Postgres) GetExpirationInformation(tx *sql.Tx, ctx context.Context, day
...
 		}
-		expirationInfo.Federated = false
-		for _, fed := range fedList {
-			if fed == expirationInfo.DeliveryService {
-				expirationInfo.Federated = true
-			}
+		if inactiveList[expirationInfo.DeliveryService] {
+			continue
 		}
+		expirationInfo.Federated = util.ContainsStr(fedList, expirationInfo.DeliveryService)
```
srijeet0406, line 221: I should have explained better in my previous comment, my bad. But, do you think converting the `fedList` into a map also makes sense, because of the same reason? That way we won't have to traverse the `fedList` array to look for the `expirationInfo.DeliveryService` key.

srijeet0406 (approved):
Code looks great! API, unit and manual testing passed. LGTM.

PR #6758, opened 2022-04-15
Refactor Delivery Service Requests

ericholguin:
**DEPENDS ON PR: https://github.com/apache/trafficcontrol/pull/6746** 

This PR refactors **v4/deliveryservice_requests_test.go**, **v3/deliveryservice_requests_test.go**, in order to make tests more granular, providing transparency on what is being tested and reducing the need to dig through test code to confirm what was tested. In addition this new test structure should also make it easier for new tests to be added as well as reduce redundant code.

- Adds data driven testing
    - Descriptive test names
- Sub tests for organization
    - Use GoLangs test runner
- Use assert functionality
    - Streamlines code
    - Reduces nested conditionals
- Adds expectation functions
     - Reusable expectations

<hr/>

- Traffic Ops Testing

Run TO Integration Tests

## PR submission checklist
- [x] This PR has tests 


traffic_ops/testing/api/v3/deliveryservice_requests_test.go
```diff
@@ -16,366 +16,317 @@ package v3
...
+							}
+						})
+					case "PUT":
+						t.Run(name, func(t *testing.T) {
+							alerts, reqInf, err := testCase.ClientSession.UpdateDeliveryServiceRequestByIDWithHdr(testCase.EndpointId(), dsReq, testCase.RequestHeaders)
+							for _, check := range testCase.Expectations {
+								check(t, reqInf, nil, alerts, err)
+							}
+						})
+					case "DELETE":
```
rimashah25, line 270: Why is there a case DELETE, when it is already covered in withobjs_test.go?

traffic_ops/testing/api/v4/deliveryservice_requests_test.go
```diff
@@ -16,672 +16,345 @@ package v4
...
+							"tenant":               "tenant1",
+							"type":                 "HTTP",
+							"xmlId":                "test-ds1",
+						},
+						"status": "draft",
+					},
+					Expectations: utils.CkRequest(utils.HasError(), utils.HasStatus(http.StatusBadRequest)),
+				},
+			},
+			"DELETE": {
```
rimashah25, line 231: Why is there a case DELETE, when it is already covered in withobjs_test.go?

traffic_ops/testing/api/v3/deliveryservice_requests_test.go
ericholguin: Both the create and delete functions that are used by withobjs are only used for the prerequisites, the only checks those functions perform is that no error occurred, but they do not check status codes or any other validation. And those also do not count as test cases so there is no transparency that post or delete were tested.

traffic_ops/testing/api/v4/deliveryservice_requests_test.go
ericholguin: See comment above

rimashah25 (approved):

traffic_ops/testing/api/v4/deliveryservice_requests_test.go
```diff
@@ -16,672 +16,356 @@ package v4
...
+					Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK), utils.ResponseLengthGreaterOrEqual(1),
+						validateGetDSRequestFields(map[string]interface{}{"XMLID": "test-ds1"})),
+				},
+			},
+			"PUT": {
+				"OK when VALID request": {
+					EndpointId:    GetDeliveryServiceRequestId(t, "test-ds1"),
+					ClientSession: TOSession,
+					RequestBody: map[string]interface{}{
+						"changeType": "create",
```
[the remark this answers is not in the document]
rimashah25: Per @ericholguin:
`Oh yeah so in that scenario the change type can be anything, we’re testing that we can update an existing delivery service request, not testing a delivery service request thats requesting to update a delivery service. If that makes sense`

Rima: confusing but I understand.

PR #6760, opened 2022-04-15
Fix missing input in execution environment for API tests GHA

ocket8888:
The GHA for our API tests isn't setting one of its required input parameters, this PR fixes that. It also wasn't exiting with a failure state on failure, which is why we didn't notice - I fixed that too.

<hr/>

- GitHub Actions

Make sure the actions _actually_ pass - you'll need to read the logs to be sure, even if they give out that green checkmark.

- master

## PR submission checklist
- [x] This PR has tests


shamrickus (approved):
LGTM
* TO Int tests now actually run

PR #6761, opened 2022-04-18
Add "sharing" ability to CDN locks

srijeet0406:
This PR is not related to any issue. It adds the ability to CDN locks, so that they can be shared amongst multiple users.
<hr/>

- Documentation
- Traffic Ops
- Traffic Portal

Run TO and TP locally.
1. As user A, acquire the lock on CDN X.
In the shared usernames field, type in the username of another user that has a tenancy of lesser than or equal to that of user A.
Make sure you can do all the activities on CDN X as user B, as you would have been able to as user A.

2. Make sure that user A cannot share their lock with a user that has a higher tenancy level than itself, or with a user that doesn't exist in the database.

3. Make sure all the tests pass.

- master

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
+++ b/traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
@@ -32,12 +32,32 @@ import (
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/api"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/auth"
 	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/dbhelpers"
+
+	"github.com/lib/pq"
 )
 
-const readQuery = `SELECT username, cdn, message, soft, last_updated FROM cdn_lock`
-const insertQuery = `INSERT INTO cdn_lock (username, cdn, message, soft) VALUES (:username, :cdn, :message, :soft) RETURNING username, cdn, message, soft, last_updated`
-const deleteQuery = `DELETE FROM cdn_lock WHERE cdn=$1 AND username=$2 RETURNING username, cdn, message, soft, last_updated`
-const deleteAdminQuery = `DELETE FROM cdn_lock WHERE cdn=$1 RETURNING username, cdn, message, soft, last_updated`
+const readQuery = `SELECT username, cdn, message, soft, shared_usernames, last_updated FROM cdn_lock`
+const insertQuery = `INSERT INTO cdn_lock (username, cdn, message, soft, shared_usernames) VALUES ($1, $2, $3, $4, $5) RETURNING username, cdn, message, soft, shared_usernames, last_updated`
+const deleteQuery = `DELETE FROM cdn_lock WHERE cdn=$1 AND username=$2 RETURNING username, cdn, message, soft, shared_usernames, last_updated`
+const deleteAdminQuery = `DELETE FROM cdn_lock WHERE cdn=$1 RETURNING username, cdn, message, soft, shared_usernames, last_updated`
+const checkSharedUsersValidityQuery = `select count(*) from tm_user u join role r on r.id = u.role join role_capability rc on rc.role_id = r.id where u.tenant_id in (WITH RECURSIVE
+user_tenant_id as (select (select tenant_id from tm_user where username=$1)::bigint as v),
+user_tenant_parents AS (
+  SELECT active, parent_id FROM tenant WHERE id = (select v from user_tenant_id)
+  UNION
+  SELECT t.active, t.parent_id FROM TENANT t JOIN user_tenant_parents ON user_tenant_parents.parent_id = t.id
+),
+user_tenant_active AS (
+  SELECT bool_and(active) as v FROM user_tenant_parents
+),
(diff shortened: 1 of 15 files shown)


ericholguin:
I am seeing an issue where I cannot always share a lock with an existing user. Is there something I am missing?

I am logged in as admin and have these users available (some I created):
Screen Shot 2022-04-21 at 11 03 56 AM

However when I try creating a lock with one of the users listed above, using my admin account, (on TO or TP)
Example:
```json
{
    "cdn": "CDN-in-a-Box",
    "message": "test",
    "soft": true,
    "sharedUserNames": ["tenant2"]
}
```

The response returned (or TP Alert):

```json
{
    "alerts": [
        {
            "text": "shared users must exist and be in the same/child tenants of the current user",
            "level": "error"
        }
    ]
}
```
The tenant2 user clearly exists and is a child tenant of root. This also happens with the "readonly" user which has the same tenancy as admin: root. Is there a certain rule or something else I might be missing?

srijeet0406:
> I am seeing an issue where I cannot always share a lock with an existing user. Is there something I am missing?
> 
> I am logged in as admin and have these users available (some I created): Screen Shot 2022-04-21 at 11 03 56 AM
> 
> However when I try creating a lock with one of the users listed above, using my admin account, (on TO or TP) Example:
> 
> ```json
> {
>     "cdn": "CDN-in-a-Box",
>     "message": "test",
>     "soft": true,
>     "sharedUserNames": ["tenant2"]
> }
> ```
> 
> The response returned (or TP Alert):
> 
> ```json
> {
>     "alerts": [
>         {
>             "text": "shared users must exist and be in the same/child tenants of the current user",
>             "level": "error"
>         }
>     ]
> }
> ```
> 
> The tenant2 user clearly exists and is a child tenant of root. This also happens with the "readonly" user which has the same tenancy as admin: root. Is there a certain rule or something else I might be missing?

Ok so the `read-only` role user will not be able to create locks because it doesn't have the `CDN-LOCK:CREATE` permission. I'll add a more informative error message for that use case. I'm looking into the other issue where a user cannot share their lock with a child tenant user.

srijeet0406:
Hmm, I'm able to create a lock with an `admin` owner and `root` tenant and share it with another user that has a tenant set to the child tenant of `root`. I'll sync up with you offline.

traffic_ops/traffic_ops_golang/cdn_lock/cdn_lock.go
```diff
@@ -32,12 +32,32 @@ import (
...
-const deleteQuery = `DELETE FROM cdn_lock WHERE cdn=$1 AND username=$2 RETURNING username, cdn, message, soft, last_updated`
-const deleteAdminQuery = `DELETE FROM cdn_lock WHERE cdn=$1 RETURNING username, cdn, message, soft, last_updated`
+const readQuery = `SELECT username, cdn, message, soft, shared_usernames, last_updated FROM cdn_lock`
+const insertQuery = `INSERT INTO cdn_lock (username, cdn, message, soft, shared_usernames) VALUES ($1, $2, $3, $4, $5) RETURNING username, cdn, message, soft, shared_usernames, last_updated`
+const deleteQuery = `DELETE FROM cdn_lock WHERE cdn=$1 AND username=$2 RETURNING username, cdn, message, soft, shared_usernames, last_updated`
+const deleteAdminQuery = `DELETE FROM cdn_lock WHERE cdn=$1 RETURNING username, cdn, message, soft, shared_usernames, last_updated`
+const checkSharedUsersValidityQuery = `select count(*) from tm_user u join role r on r.id = u.role join role_capability rc on rc.role_id = r.id where u.tenant_id in (WITH RECURSIVE
+user_tenant_id as (select (select tenant_id from tm_user where username=$1)::bigint as v),
```
rawlinp, line 44: Do we really need to check tenancy here? If someone has the ability to lock the CDN, shouldn't we be able to share the lock with them even if they're in a different or ancestral tenant?

traffic_portal/app/src/common/modules/dialog/select/lock/dialog.select.lock.tpl.html
```diff
@@ -39,6 +39,17 @@ <h4 class="modal-title">Create CDN Lock</h4>
                 </div>
             </div>
         </div>
+        <div class="row">
+            <div class="col-sm-12 col-md-12">
+                <div class="form-group">
+                    <label class="control-label" for="sharedUserNames">Usernames to share lock with</label>
+                    <input id="sharedUserNames" name="sharedUserNames" type="text" class="form-control" ng-list ng-model="lock.sharedUserNames" maxlength="256">
+                    <div class="helptext">
+                        A comma separated list of usernames that you want your lock to be shared with.
```
rawlinp, line 48: Comma-separated lists are kind of clunky -- it would be nice to mirror the list functionality of adding Consistent Hash Query Params in the DS form, where you can add/remove to the list by using buttons. We could even get the list of active users from the TO API and pre-populate the list of valid usernames to choose from.

rawlinp (changes requested):

rawlinp (approved):
We can probably make the UI options list for selecting shared users better, but it works for a first pass at it.

PR #6762, opened 2022-04-19
Refactor  cdn_domains, cdnnotifications, cookie, logs, ping, traffic_vault_ping tests

ericholguin:
This PR refactors 
- **v4/cdn_domains_test.go**
- **v3/cdn_domains_test.go**
- **v4/cdnnotifications_test.go**
- **v4/cookie_test.go**
- **v3/cookie_test.go**
- **v4/logs_test.go**
- **v3/logs_test.go**
- **v4/ping_test.go**
- **v3/ping_test.go**
- **v4/traffic_vault_ping_test.go**

in order to make tests more granular, providing transparency on what is being tested and reducing the need to dig through test code to confirm what was tested. In addition this new test structure should also make it easier for new tests to be added as well as reduce redundant code.

- Adds data driven testing
    - Descriptive test names
- Sub tests for organization
    - Use GoLangs test runner
- Use assert functionality
    - Streamlines code
    - Reduces nested conditionals
- Adds expectation functions
     - Reusable expectations

<hr/>

- Traffic Ops Tests

Run TO integration test

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/testing/api/v4/cdnnotifications_test.go
+++ b/traffic_ops/testing/api/v4/cdnnotifications_test.go
@@ -16,34 +16,61 @@ package v4
 */
 
 import (
+	"net/http"
+	"net/url"
+	"strconv"
 	"testing"
 
 	"github.com/apache/trafficcontrol/lib/go-tc"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/assert"
+	"github.com/apache/trafficcontrol/traffic_ops/testing/api/utils"
+	"github.com/apache/trafficcontrol/traffic_ops/toclientlib"
 	client "github.com/apache/trafficcontrol/traffic_ops/v4-client"
 )
 
 func TestCDNNotifications(t *testing.T) {
 	WithObjs(t, []TCObj{CDNs, CDNNotifications}, func() {
-		GetTestCDNotifications(t)
+		methodTests := utils.V4TestCase{
+			"GET": {
+				"OK when VALID request": {
+					ClientSession: TOSession, Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK),
+						utils.ResponseLengthGreaterOrEqual(1)),
+				},
+				"OK when VALID CDN parameter": {
+					ClientSession: TOSession, RequestOpts: client.RequestOptions{QueryParameters: url.Values{"cdn": {"cdn2"}}},
+					Expectations: utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK), utils.ResponseHasLength(1),
+						validateCDNNotificationFields(map[string]interface{}{"Notification": "test notification: cdn2"})),
+				},
+			},
+		}
+
+		for method, testCases := range methodTests {
+			t.Run(method, func(t *testing.T) {
+				for name, testCase := range testCases {
+					switch method {
+					case "GET":
+						t.Run(name, func(t *testing.T) {
+							resp, reqInf, err := testCase.ClientSession.GetCDNNotifications(testCase.RequestOpts)
+							for _, check := range testCase.Expectations {
+								check(t, reqInf, resp.Response, resp.Alerts, err)
+							}
+						})
+					}
+				}
+			})
+		}
 	})
 }
 
-// Note that this test will break if anyone adds a CDN notification to the test
-// data that isn't exactly `test notification: {{CDN Name}}` (where {{CDN Name}}
-// is the name of the associated CDN).
-func GetTestCDNotifications(t *testing.T) {
-	opts := client.NewRequestOptions()
-	for _, cdn := range testData.CDNs {
-		opts.QueryParameters.Set("cdn", cdn.Name)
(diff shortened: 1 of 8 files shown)


traffic_ops/testing/api/v4/cdnnotifications_test.go
```diff
@@ -53,26 +80,21 @@ func CreateTestCDNNotifications(t *testing.T) {
 	var opts client.RequestOptions
 	for _, cdn := range testData.CDNs {
 		resp, _, err := TOSession.CreateCDNNotification(tc.CDNNotificationRequest{CDN: cdn.Name, Notification: "test notification: " + cdn.Name}, opts)
```
ocket8888, line 82: This creates the new test notification using `cdn.Name`, but the above check is looking for a hard-coded CDN name. The test should either consistently use the same testing data CDN name, or just use a hard-coded notification text, preferably stored as a constant value so that it can be changed in one place if need be.

ericholguin: Since since is the pre-requisite function it'll create a cdn notification for all the cdns in the tc-fixtures file. The one above is the actual test and its hard-coded because we know what pre-requisite data we are using through tc-fixtures.json. Using testData.CDNs[0].Name would introduce some confusion in addition to having to be track of which specific cdn its referring to.

ericholguin: I think the confusion would be that someone would have to verify that the cdn at that index is the correct one for that specific test. (In this scenario it does not matter since they all have notifications) But this applies to all tests were we use a specific object for a specific test. Previously, before the refactor the index was being hardcoded, and I see that more confusing, since I would have to count in fixtures to see which object the test was referring to. 

> Alternatively, this could just check that a notification of the form test notification: CDN Name exists for every CDN instead of singling one out. In fact, I just checked and that's what the old tests used to do, so this is technically losing coverage by not replicating that.

It is bad testing practice to test the same thing multiple times, which a lot of tests currently do, we aren't losing coverage by removing duplicate tests.

ocket8888: On the other hand, since nothing actually depends on CDN notifications, and they're never used in a `WithObjs` anywhere, it may not be best to have them in the testing fixture data at all, and the tests could just make a couple with some randomly generated strings and then verify that the notifications are in fact created with that text. Random strings in tests do usually make me uncomfortable, but I can't really justify that feeling with any argument.

ericholguin: > That's not quite a duplicate test. If I change the notifications code so that no matter what you POST to it, it just stores the static string test notification: cdn2 it would have failed the old tests, but will pass the tests as of this rewrite

If someone is hardcoding core code to pass tests and that slips by a review, that falls outside the scope of what a test can cover.

> or if I change it to only store the first notification it gets and use it as the text content of all future notifications, or if a database change is introduced with a bad constraint such that it will store one notification just fine but any and all future ones erroneously violate a constraint and TO doesn't handle the resulting db error properly and just returns a 200. So there's a case not being covered anymore. Those are all pretty unlikely scenarios, but they used to be tested against and won't be anymore.

Those are technically covered by this test, the test performs a `GET` request against the second cdn notification `cdn2` not `cdn1` which is the first notification created. But again really unlikely scenarios where we wonder what actual value they provide.

> It probably suffices to check two before the tests become truly degenerate, but as it's a read-only operation I wouldn't see the harm in doing them all.

This is a common discussion when it comes to testing, where there isn't a good answer. Just because we can test everything doesn't mean we should. Those scenarios are rare and this endpoint isn't crucial. Although adding those tests wouldn't bring any harm, we are only increasing the amount of tests and test time but with not much of a gain.

> On the other hand, since nothing actually depends on CDN notifications, and they're never used in a WithObjs anywhere, it may not be best to have them in the testing fixture data at all, and the tests could just make a couple with some randomly generated strings and then verify that the notifications are in fact created with that text. Random strings in tests do usually make me uncomfortable, but I can't really justify that feeling with any argument.

Actually notifications themselves are not in the fixtures data, it's using the cdns that are defined in fixtures to create the notification.

I'm okay with adding two constant values that the prerequisite Create function will create, so no longer leverage testData.CDN, the cdnName will still be hardcoded against what exists in fixtures. And also adding a test for both - one for the valid request and for checking that the notification varies if you agree to that.

ocket8888: Hm. That does sound better, but it's not replicating the behavior of the existing tests, so I wouldn't ask you to do that. It just bothers me that something labeled a "refactor" changes the code's behavior and not merely its structure.

I'll leave it up to you. If you really think it's best to run fewer tests then that's fine.  If you think it adds value to make it do the two tests you just described, then do it. But if that doesn't sound worthwhile to you, don't waste time on it and I'll just approve this as-is.

ericholguin: I think this can be approved as is

ocket8888 (approved):

PR #6769, opened 2022-04-21
Document Traffic Monitor configuration files

ocket8888:
This PR adds/updates documentation for the configuration files used by Traffic Monitor (typically named `traffic_ops.cfg` and `traffic_monitor.cfg`).

This also fixes a bug in table rendering currently on master since #6569.

<hr/>

- Documentation

Make sure the docs build without errors or warnings, and are accurate, thorough, and clear. 

- master

## PR submission checklist
- [ ] This PR has tests

### DIFF
--- a/docs/source/admin/traffic_monitor.rst
+++ b/docs/source/admin/traffic_monitor.rst
@@ -80,7 +80,7 @@ traffic_monitor.cfg
 	.. Note:: ``both`` will poll IPv4 and IPv6 and report on availability based on if the respective IP addresses are defined on the server. So if only an IPv4 address is defined and the protocol is set to ``both`` then it will only show the availability over IPv4, but if both addresses are defined then it will show availability based on IPv4 and IPv6.
 
 :``crconfig_backup_file``:   The path to a file within which a backup of the most recently fetched CDN :term:`Snapshot` will be stored. Default is ``/opt/traffic_monitor/crconfig.backup``.
-:``crconfig_history_count``: The number of historical CDN Snapshots to store, which can then be retrieved through the :ref:`tm-api`. Default is 20,000.
+:``crconfig_history_count``: The number of historical CDN Snapshots to store, which can then be retrieved through the :ref:`tm-api`. Default is 100.
 :``distributed_polling``:    A boolean that controls whether `Distributed Polling`_ is enabled. Default is ``false``.
 
 	.. seealso:: The `Distributed Polling`_ section has more information on this setting.
--- a/docs/source/admin/traffic_monitor.rst
+++ b/docs/source/admin/traffic_monitor.rst
@@ -60,7 +60,7 @@ traffic_ops.cfg
 :``certFile``:      The path to an SSL certificate file that corresponds to ``keyFile`` which will be used for Traffic Monitor's HTTPS API server.
 :``httpListener``:  Sets the address and port on which Traffic Monitor will listen for HTTP requests in the format :samp:`{address}:{port}`. If ``address`` is omitted, Traffic Monitor will listen on all available addresses.
 :``httpsListener``: Sets the address and port on which Traffic Monitor will listen for HTTPS requests in the format :samp:`{address}:{port}`. If ``address`` is omitted, Traffic Monitor will listen on all available addresses. If not provided, ``null``, or the empty string, Traffic Monitor will only serve HTTP, and ``keyFile`` and ``certFile`` are not used. If this is provided, the ``httpListener`` address will be used only to redirect clients to use HTTPS.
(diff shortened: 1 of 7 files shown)


traffic_monitor/config/config.go
```diff
@@ -149,33 +191,33 @@ func GetAccessLogWriter(cfg Config) (io.WriteCloser, error) {
...
 var DefaultConfig = Config{
-	MonitorConfigPollingInterval: 5 * time.Second,
-	HTTPTimeout:                  2 * time.Second,
-	PeerOptimistic:               true,
-	PeerOptimisticQuorumMin:      0,
-	StatPolling:                  true,
-	MaxEvents:                    200,
+	CachePollingProtocol:         Both,
+	CRConfigBackupFile:           CRConfigBackupFile,
+	CRConfigHistoryCount:         20000,
```
rawlinp, line 196: we should make this like...100? I can't imagine needing more than like 10 in reality, but 20,000 seems ridiculous

traffic_monitor/config/config.go
```diff
@@ -97,37 +97,79 @@ func (t *PollingProtocol) UnmarshalJSON(b []byte) error {
...
+	LogLocationWarning string `json:"log_location_warning"`
+	// The maximum number of events to keep in TM's buffer to be served via its
+	// API.
+	MaxEvents uint64 `json:"max_events"`
+	// The interval on which to poll for this TM's CDN's "monitoring config".
+	MonitorConfigPollingInterval time.Duration `json:"-"`
+	// This appears to do nothing.
+	//
+	// Deprecated: This serves no purpose and will be removed in the future.
+	PeerOptimistic bool `json:"peer_optimistic"`
```
rawlinp, line 142: We should just remove it, no need to deprecate first. Existing configs that still add it won't be broken, it would just remain unused (as it already is).

rawlinp (changes requested):

traffic_monitor/config/config.go
ocket8888: ¯\\\_(ツ)\_/¯

I didn't change any of these values, just sorted the lines alphabetically. I can change the defaults if these don't make sense, though.

traffic_monitor/config/config.go
ocket8888: That's true. I don't mind taking it out at all, but the reason I deprecated it instead was just so that there were no changes to the logic/behavior of Traffic Monitor whatsoever. All I did was sort lines (which I guess technically affects structure packing) and add comments. I'll change things if you want, I just want to be clear that that's what you're asking.

traffic_monitor/config/config.go
rawlinp: Might as well while we're in here

traffic_monitor/config/config.go
rawlinp: Sure, let's just remove it then. I don't mind reviewing that

rawlinp (changes requested):

docs/source/admin/traffic_monitor.rst
```diff
@@ -48,17 +48,85 @@ These are the recommended hardware specifications for a production deployment of
...
+:``username``:      The username of the user as whom to authenticate with Traffic Ops.
+:``usingDummyTO``:  A boolean with no real effect. This value is used internally within the runtime of Traffic Monitor, and should never be set manually in its configuration file.
+
+	.. deprecated:: ATCv7
+		The dependency on this field being valid will be removed in the future. It already has no effect.
+
+
+traffic_monitor.cfg
+"""""""""""""""""""
+:file:`traffic_monitor.cfg` contains log file locations, as well as detailed application configuration variables such as processing flush times, initial poll intervals, and the polling protocols. Once started with the correct configuration, Traffic Monitor downloads its configuration from Traffic Ops which will override the options in this file and begins polling :term:`cache server` s. Once every :term:`cache server` has been polled, :ref:`health-proto` state is available via RESTful JSON endpoints and a web browser UI.
```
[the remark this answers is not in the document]
ocket8888: I only meant that the configuration present in Traffic Ops would override the file's configuration - so if there's no way to configure something in Traffic Ops it couldn't be overridden.

I can try to make that more clear, but is that base assumption true? Or are there things that can be configured in Traffic Ops that would be overridden by the TM config file?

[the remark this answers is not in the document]
rawlinp: Oh ok, yeah, if there are corresponding TM profile params, those will override options in the config file, but not the other way around. Ideally, we should notate the corresponding TM profile params which would override the config file options. I'm pretty sure that includes at least these ones:
```
tm.polling.interval
health.polling.interval
peers.polling.interval
heartbeat.polling.interval
```

rawlinp (approved):

PR #6773, opened 2022-04-22
Add e2e tests for TPv2 Servers table

ocket8888:
This PR adds end-to-end tests for the TPv2 "servers" page/table.

I also generalized some types to greatly simplify writing tests esp. for generic table pages.

<hr/>

- Traffic Portal (experimental v2)

Make sure the new tests pass.

## PR submission checklist
- [x] This PR has tests


.github/workflows/tpv2.yml
```diff
@@ -16,53 +16,206 @@
 # under the License.
 name: Lint and Test Experimental Traffic Portal
```
shamrickus, line 17: This should probably be renamed to just `Experimental Traffic Portal` or something of that sort.

.github/workflows/tpv2.yml
```diff
@@ -16,53 +16,206 @@
...
           node-version: 16.x
 
-      - name: Install latest Chrome
-        run: sudo apt-get update && sudo apt-get install google-chrome-stable
-
       - name: NPM install
         run: |
           cd experimental/traffic-portal/
-          npm ci
+          npm install
```
shamrickus, line 56: `npm ci` was correct

.github/workflows/tpv2.yml
```diff
@@ -16,53 +16,206 @@
...
+
+      - name: Node 16
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+
+      - name: NPM install
+        run: |
+          cd experimental/traffic-portal/
+          npm install
```
shamrickus, line 86: `npm ci`

.github/workflows/tpv2.yml
```diff
@@ -16,53 +16,206 @@
...
+        with:
+          node-version: 16.x
+
+      - name: Install latest Chrome
+        run: sudo apt-get update && sudo apt-get install google-chrome-stable
+
+      - name: NPM install
+        run: |
+          cd experimental/traffic-portal/
+          npm install
```
shamrickus, line 119: `npm ci`

.github/workflows/tpv2.yml
```diff
@@ -16,53 +16,206 @@
...
+        with:
+          go-version: ${{ steps.go-version.outputs.value }}
+      - name: Build Traffic Ops
+        run: |
+          cd "${GITHUB_WORKSPACE}/traffic_ops/traffic_ops_golang"
+          go build .
+
+      # Setup
+      - name: Install dependencies
+        run: sudo apt-get update && sudo apt-get install google-chrome-stable postgresql-client gettext-base
```
shamrickus, line 195: `google-chrome-stable` is not needed for the e2e tests

.github/workflows/tpv2.yml
```diff
@@ -16,53 +16,206 @@
...
+        run: |
+          psql -c "INSERT INTO tm_user (username, role, tenant_id, local_passwd) VALUES ('admin', 1, 1, 'SCRYPT:16384:8:1:p0Bppp/6IBeYxSwdLuYddsdMLBU/BNSlLY6fSIF7H1XW4eTbNVeMPVm7TuTEG4FM8PbqLlVwi8sPy8ZJznAlaQ==:sRcHWGe43mm/uEmXTIw37GcLEQZTlWAdf4vJqK8f0MDh8P+8gXoNx+nxWyb3r/0Bh+yyg0g/dUvti/ePZJL+Jw==');"
+      - name: Create SSL Certificates and AES key
+        run: |
+          openssl rand 32 | base64 | tee "${GITHUB_WORKSPACE}/aes.key"
+          openssl req -new -x509 -nodes -newkey rsa:4096 -out traffic_ops/traffic_ops_golang/localhost.crt -keyout traffic_ops/traffic_ops_golang/localhost.key -subj "/CN=tptests"
+      - name: NPM install
+        run: |
+          cd experimental/traffic-portal
+          npm install
```
shamrickus, line 206: `npm ci`

.github/workflows/tpv2.yml
```diff
@@ -16,53 +16,206 @@
...
+      - name: Upload Report
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: ${{ github.job }}
+          path: |
+            traffic_ops/traffic_ops_golang/out.log
+            experimental/traffic-portal/nightwatch/screens
+            experimental/traffic-portal/logs
+            experimental/traffic-portal/tests_output
```
shamrickus, line 218: `nightwatch/junit` is not uploaded. I don't think it actually provides any additional info but it's being generated so we might as well include it in the artifacts.

experimental/traffic-portal/package.json
```diff
@@ -40,7 +40,7 @@
     "lint": "ng lint",
     "e2e": "npm run e2e:build && ./node_modules/.bin/nightwatch --config './nightwatch/nightwatch.conf.js'",
     "e2e:build": "./node_modules/.bin/tsc -p ./nightwatch/tsconfig.e2e.json",
-    "e2e:ci": "npm run e2e:build && ./node_modules/.bin/nightwatch --env chrome_headless --config './nightwatch/nightwatch.conf.js'",
+    "e2e:ci": "npm run e2e:build && ./node_modules/.bin/nightwatch --env chrome_headless --config './nightwatch/nightwatch.conf.js' --window-size 1920,1080",
```
shamrickus, line 43: I think we should set this in the `nightwatch.conf.js` `chrome_headless` args instead.

shamrickus (changes requested):

.github/workflows/tpv2.yml
ocket8888: No, it's not. `npm ci` removes all of the `node_modules` and re-installs them fresh, effectively rendering our earlier caching pointless. Not loading the cached modules, then deleting them, then re-downloading them cuts the time it takes to lint and run unit tests down from like 5 minutes to about 1.5 minutes.

shamrickus: If thats the case then `npm install` shouldn't be necessary then? Assuming it isn't then you can just add `if: steps.cache-node-modules.outputs.cache-hit != 'true'` so that it doesn't overwrite in that case.

ocket8888: I did exactly that like three minutes before you made this comment.

shamrickus (approved):
LGTM

Issue #6654, opened 2022-03-15
TM should use TO API 4.0 (once stable) and fall back to 3.1


rawlinp:
## This Improvement request (usability, performance, tech debt, etc.) affects these Traffic Control components:

- Traffic Monitor

## Current behavior:

TM uses TO API 3.1 and will fall back to 2.0 in case 3.1 is unavailable.

## New behavior:

TM should use TO API 4.0 (once stable) and fall back to 3.1 in case 4.0 is unavailable.

mitchell852:
this is dependent on API 4.x being stabilized which is done with #6700

PR #6786, opened 2022-04-26
Update TM to use TO API 4.0/3.1

shamrickus:
Closes: #6654
<hr/>

- Traffic Monitor

Ensure the tests pass. Verify that TM is able to query TO correctly as well as other TMs

 - master

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_monitor/towrap/towrap.go
+++ b/traffic_monitor/towrap/towrap.go
@@ -406,6 +406,10 @@ func (s TrafficOpsSessionThreadsafe) CRConfigRaw(cdn string) ([]byte, error) {
 	} else {
 		crConfig = &response.Response
 		configBytes, err = json.Marshal(crConfig)
+		if err != nil {
+			crConfig = nil
+			log.Warnln("failed to marshal CRConfig using up-to-date client: " + err.Error())
+		}
 	}
 
 	if err == nil {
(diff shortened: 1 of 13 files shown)


traffic_monitor/towrap/towrap.go
```diff
@@ -431,23 +439,23 @@ func (s TrafficOpsSessionThreadsafe) CRConfigRaw(cdn string) ([]byte, error) {
 	}
 	defer s.crConfigHist.Add(hist)
 
-	crc := &tc.CRConfig{}
-	json := jsoniter.ConfigFastest
-	if err = json.Unmarshal(data, crc); err != nil {
-		err = errors.New("invalid JSON: " + err.Error())
-		hist.Err = err
-		return data, err
+	if crConfig == nil {
```
rawlinp, line 442: I think you mean `if crConfig != nil` here

rawlinp (changes requested):

traffic_monitor/towrap/towrap.go
```diff
@@ -431,23 +439,23 @@ func (s TrafficOpsSessionThreadsafe) CRConfigRaw(cdn string) ([]byte, error) {
...
 	defer s.crConfigHist.Add(hist)
 
-	crc := &tc.CRConfig{}
-	json := jsoniter.ConfigFastest
-	if err = json.Unmarshal(data, crc); err != nil {
-		err = errors.New("invalid JSON: " + err.Error())
-		hist.Err = err
-		return data, err
+	if crConfig != nil {
+		if err = json.Unmarshal(configBytes, *crConfig); err != nil {
```
rawlinp, line 443: Wait, do we even need to dereference `crConfig` here? Isn't the 2nd arg supposed to be a pointer? Should `crConfig` be initialized on L381 so that it's non-nil in either case?

rawlinp (approved):

PR #6803, opened 2022-04-29
Documentation fixes

villajo:
This PR is to fix random documentations issues. 

<hr/>

- Documentation

in the ./docs folder, run 'make' then view locally via webbrowser. 

## PR submission checklist
This is updates to documentation.


villajo:
The docs/source/api/v3/deliveryservice_requests_id_assign.rst is probably incorrect. I committed so I could work with you in finding the right solution. Not only does the JSON request seem like a small block, but it's probably incomplete.

docs/source/api/v3/deliveryservice_requests_id_assign.rst
```diff
@@ -29,7 +29,7 @@ Assign a :term:`Delivery Service Request` to a user.
 Request Structure
 -----------------
 :id:       The integral, unique identifier assigned to the :term:`DSR <Delivery Service Request>`
-:assignee: The username of the user to whom the :term:`Delivery Service Request` is assigned.
+:assigneeId: The username of the user to whom the :term:`Delivery Service Request` is assigned.
```
ocket8888, line 32: This description needs to be changed; the value is the user's *ID*, not the username.

docs/source/api/v3/deliveryservice_requests_id_assign.rst
```diff
@@ -43,13 +43,12 @@ Request Structure
...
 
 	{
-		"id": 1,
 		"assigneeId": 2
 	}
 
 Response Structure
 ------------------
-:assignee:        The username of the user to whom the :term:`Delivery Service Request` is assigned.
+:assigneeId:        The username of the user to whom the :term:`Delivery Service Request` is assigned.
```
ocket8888, line 51: `assigneeId` needs to be added here, but `assignee` should not be removed as it is actually present in the response.

ocket8888 (changes requested):

PR #6810, opened 2022-05-05
Blueprint: Cache Config Service

rawlinp:
Introduce a blueprint for the Cache Config Service.

<hr/>


rawlinp (commented):
Thanks for the feedback. This blueprint is temporarily on-hold as some issues are worked out offline w.r.t. its implementation and whether or not we can move forward with it.

PR #6812, opened 2022-05-06
Correct __HOSTNAME__ for short hostName unknown files generated by T3C

mikeV02:
This corrects a bug where using the placeholder `__HOSTNAME__` in "unknown" files (others than the defaults ones), was being replaced by the full FQDN instead of the shot hostname.
<hr/>

To trigger the issue create any non-default file for ATS in the parameters and associate with the ATS profile:
    - Create a location parameter to the path on ATS you want, name the file to something like test_hostname_unk
    - And create another parameter with the content of that file, something like `I want the hostname, but T3C gave the FQDN __HOSTNAME__`

- All

## PR submission checklist

### DIFF
--- a/lib/go-atscfg/serverunknown.go
+++ b/lib/go-atscfg/serverunknown.go
@@ -75,7 +75,7 @@ func MakeServerUnknown(
 		txt += pa.Val + "\n"
 	}
 
-	txt = strings.Replace(txt, `__HOSTNAME__`, *server.HostName+`.`+*server.DomainName, -1)
+	txt = strings.Replace(txt, `__HOSTNAME__`, *server.HostName, -1)
 	txt = strings.Replace(txt, `__RETURN__`, "\n", -1)
 
 	lineComment := getServerUnknownConfigCommentType(params)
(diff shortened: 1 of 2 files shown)


lib/go-atscfg/serverunknown.go
```diff
@@ -75,7 +75,7 @@ func MakeServerUnknown(
 		txt += pa.Val + "\n"
 	}
 
-	txt = strings.Replace(txt, `__HOSTNAME__`, *server.HostName+`.`+*server.DomainName, -1)
+	txt = strings.Replace(txt, `__HOSTNAME__`, *server.HostName, -1)
```
rob05c, line 78: Nitpick: This could be removed, since t3c-preprocess will do this. But not a huge deal, behaves correctly.

rob05c (approved):
Looks good, thanks! 

Code is obviously correct. Tests would be ideal, but this should really be part of larger t3c-preprocess tests, which don't appear to exist, so I wouldn't insist on it for this PR

PR #6815, opened 2022-05-06
Change TR to depend on java-11-openjdk-headless instead of java-11 

mikeV02:
While doing a test installation, noticed a lot of unnecessary packages were being installed along Traffic Router and tomcat. When checking why, I noticed Traffic Router has been marked to use java-11 as a dependency; however that package requires X11, and all the deps that come with X11.
<hr/>

- Traffic Router

Download trafficrouter RPM, run the installation for that RPM and see the dependencies required for java-11 and compare with java-11-openjdk-headless

## PR submission checklist

### DIFF
--- a/traffic_router/build/pom.xml
+++ b/traffic_router/build/pom.xml
@@ -236,7 +236,7 @@
 								</mapping>
 							</mappings>
 							<requires>
-								<require>java-11</require>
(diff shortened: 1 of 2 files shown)


rawlinp (approved):
Thanks for the PR!

PR #6817, opened 2022-05-09
Prevent modifying and/or deleting reserved statuses

### LINKED ISSUE #6291: Important Statuses can be deleted and modified
- Traffic Ops

## Current behavior:
The Statuses named `OFFLINE`, `ONLINE`, `REPORTED`, `PRE_PROD`, and `ADMIN_DOWN` are handled specially by different parts of Traffic Control, and different parts of the code even go so far as to assume that `OFFLINE`, `ONLINE`, and `REPORTED` exist, but in fact there is no guarantee that any of the five exist at any given time.

That's because you can rename them (and less importantly change their descriptions):
```http
GET /api/4.0/statuses?id=1 HTTP/1.1
Host: localhost:6443
Cookie: mojolicious=...
Content-Length: 0


HTTP/1.1 200 OK
Set-Cookie: mojolicious=...
Transfer-Encoding: chunked

{ "response": [{
	"description": "Server is Offline. Not active in any configuration.",
	"id": 1,
	"lastUpdated": "2021-10-14 08:04:39+00",
	"name": "OFFLINE"
}]}
```
```http
PUT /api/3.0/statuses/1 HTTP/1.1
Cookie: mojolicious=...
Content-Length: 88

{"name": "Not OFFLINE anymore", "description": "an unhelpful or inaccurate description"}


HTTP/1.1 200 OK
Permissions-Policy: interest-cohort=()
Set-Cookie: mojolicious=...
Content-Length: 188

{
	"alerts": [
		{
			"text": "status was updated.",
			"level": "success"
		}
	],
	"response": {
		"description": "an unhelpful or inaccurate description",
		"id": 1,
		"lastUpdated": "2021-10-15 08:41:42+00",
		"name": "Not OFFLINE anymore"
	}
}
```

and even delete them:
```http
DELETE /api/4.0/statuses/1 HTTP/1.1
Host: localhost:6443
Cookie: mojolicious=...
Content-Length: 0


HTTP/1.1 200 OK
Transfer-Encoding: chunked
Set-Cookie: mojolicious=...

{ "alerts": [{
	"text": "status was deleted.",
	"level": "success"
}]}
```

## Expected behavior:
You shouldn't be able to get rid of things that we assume should exist. That means either disallowing those operations on certain of those objects, or disallowing those operations entirely. I've always been hopeful that we could do away with custom statuses, eliminating a server field, two endpoints, hundreds of lines of code, two database tables, and a lot of confusion and edge cases like this. However, that behavior is depended upon by many, so we can't get rid of it without at least an alternative (like the one proposed by #4819), so if nothing else we should ensure that all valid (i.e. only manipulated through the API) Traffic Ops server states contain the Statuses that ATC requires to function. At a minimum that would be `REPORTED`, but to avoid unexpected behavior when creating new Statuses with reserved names we should also do the same with the others mentioned above.

srijeet0406:
This PR closes #6291 

<hr/>

- Traffic Ops

Run TO and TP locally.
Try to modify/ delete one of the following statuses. You should get an error while doing so.
Reserved status names:
`PRE_PROD, ONLINE, OFFLINE, ADMIN_DOWN, REPORTED`

- master

## PR submission checklist
- [x] This PR has tests 


mitchell852:
just curious. is "PRE_PROD" a reserved name? i.e. is there code that looks specifically for PRE_PROD and performs certain actions based on that status? is this status seeded? i was kind of under the impression Comcast added this status to support an internal process.

srijeet0406:
> just curious. is "PRE_PROD" a reserved status? i.e. is there code that looks specifically for PRE_PROD and performs certain actions based on that status? is this status seeded? i was kind of under the impression Comcast added this status to support an internal process.

I added it in there for two reasons:
1. The issue mentions it.
2. Our seeds file adds this status as part of the seeding process, so I assumed it was reserved.
That said, I can remove it if we think that it shouldn't be reserved. Not sure if/ when it got to be a reserved status.

ocket8888:
> just curious. is "PRE_PROD" a reserved status? i.e. is there code that looks specifically for PRE_PROD and performs certain actions based on that status? is this status seeded? i was kind of under the impression Comcast added this status to support an internal process.

Comcast may have done that, but it's seeded into TODB on installation. There is no code in ATC that checks for and acts explicitly on that status. It has no non-semantic meaning.

mitchell852:
> Comcast may have done that, but it's seeded into everyone's TODB on installation

ok well is guess since it's "seeded" it should be immutable

traffic_ops/traffic_ops_golang/status/statuses.go
```diff
@@ -40,6 +40,14 @@ type TOStatus struct {
...
 }
 
+var statusNameMap = map[tc.CacheStatus]bool{
+	tc.CacheStatusAdminDown: true,
+	tc.CacheStatusOnline:    true,
+	tc.CacheStatusOffline:   true,
+	tc.CacheStatusReported:  true,
+	tc.CacheStatusPreProd:   true,
+}
+
```
ocket8888, line 50: This map is not used

traffic_ops/testing/api/v4/statuses_test.go
```diff
@@ -142,13 +145,26 @@ func GetTestStatusesIMS(t *testing.T) {
...
-		if err != nil {
-			t.Errorf("could not create Status: %v - alerts: %+v", err, resp.Alerts)
+		if status.Name != nil {
+			if _, ok := statusNameMap[*status.Name]; !ok {
+				resp, _, err := TOSession.CreateStatus(status, client.NewRequestOptions())
+				t.Log("Response: ", resp)
+				if err != nil {
+					t.Errorf("could not create Status: %v - alerts: %+v", err, resp.Alerts)
+				}
+			}
```
ocket8888, line 165: You wouldn't need to build this map and do these checks if you just remove standard statuses from the testing fixture data.

lib/go-tc/statuses.go
```diff
@@ -70,3 +70,19 @@ type StatusNullable struct {
 	LastUpdated *TimeNoMod `json:"lastUpdated" db:"last_updated"`
 	Name        *string    `json:"name" db:"name"`
 }
+
+func IsReservedStatus(status string) bool {
```
ocket8888, line 74: This function could use a GoDoc.

ocket8888 (commented):
The changes to the tests are basically the same in all four of the places you're making them, so I just put the comment I had on v4 to avoid repeating myself a bunch of times.

Also, though, now that any cache status that t3c could possibly care about is guaranteed to always exist, they shouldn't need to have any fixture data or testing functions at all. Can probably even be removed from their `TCObj`/`WithObj` stuff.

cache-config/testing/ort-tests/tcdata/statuses.go
```diff
@@ -17,30 +17,14 @@ package tcdata
```
ocket8888, line 17: Are the two functions in this file ever called anymore with all of the removals you did? I wouldn't think so, but I could be wrong.

srijeet0406: You're right, I've removed the file now.

ocket8888 (approved):

Issue #6271, opened 2021-10-13
GET api/{version}/deliveryservices/{id}/health returns no info if delivery service uses a topology


mitchell852:
## This Bug Report affects these Traffic Control components:
- Traffic Ops

## Current behavior:
GET api/{{version}/deliveryservices/{id}/health returns no info if the delivery service uses a topology

```json
{ "response": {
  "totalOffline": 0,
  "totalOnline": 0,
  "cachegroups": [ ]
}}
```

## Expected behavior:
When a delivery service uses a topology, it still should have a total for all "ONLINE" and "OFFLINE" caches as well as totals for each Cache Group employed by the Topology. such as this:

```json
{ "response": {
  "totalOffline": 16,
  "totalOnline": 388,
  "cachegroups": [
    {
      "offline": 1,
      "online": 19,
      "name": "us-fl-sarasota"
    },
    {
      "offline": 2,
      "online": 39,
      "name": "us-il-chicago"
    }
  ]
}}
```

## Steps to reproduce:
1. Assign servers *directly* to a delivery service
1. `GET api/{{version}}/deliveryservices/{{id}}/health` and observe the online/offline counts (total and per cache group)
1. Assign a topology to the delivery service
1. `GET api/{{version}}/deliveryservices/{{id}}/health` and observe NO online/offline counts (total and per cache group)

PR #6827, opened 2022-05-12
`api/{{version}/deliveryservices/{id}/health` returns no info if the delivery service uses a topology

srijeet0406:
Closes: #6271 

<hr/>

- Traffic Ops

Run CDN-in-a-box.
Create two ds, one with and another without a valid topology.
Assign some "REPORTED" EDGE caches to the topology, and also do the same for the first DS (without topology).
Snap both the delivery services.
Make a call to GET `deliveryservices/{id}/health` for the both the DS's.
Make sure that both DS's show valid counts of online/ offline caches.

- master
## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/traffic_ops_golang/deliveryservice/health.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/health.go
@@ -122,6 +122,7 @@ func addHealth(ds tc.DeliveryServiceName, data map[tc.CacheGroupName]tc.HealthDa
 	var topology string
 	var cacheGroupNameMap = make(map[string]bool)
 	var cacheCapabilities = make(map[string]bool)
+	var skip bool
 
 	if deliveryService, ok = crConfig.DeliveryServices[string(ds)]; !ok {
 		log.Errorln("delivery service not found in CRConfig")
@@ -163,13 +164,17 @@ func addHealth(ds tc.DeliveryServiceName, data map[tc.CacheGroupName]tc.HealthDa
 			if _, ok := cacheGroupNameMap[*cache.CacheGroup]; !ok {
 				continue
 			}
-		}
-		cacheCapabilities = make(map[string]bool)
-		for _, cap := range cache.Capabilities {
-			cacheCapabilities[cap] = true
-		}
-		for _, rc := range deliveryService.RequiredCapabilities {
-			if _, ok = cacheCapabilities[rc]; !ok {
+			cacheCapabilities = make(map[string]bool)
+			for _, cap := range cache.Capabilities {
+				cacheCapabilities[cap] = true
+			}
+			for _, rc := range deliveryService.RequiredCapabilities {
+				if _, ok = cacheCapabilities[rc]; !ok {
+					skip = true
+					continue
+				}
+			}
+			if skip {
 				continue
 			}
 		}
--- a/traffic_ops/traffic_ops_golang/deliveryservice/health.go
+++ b/traffic_ops/traffic_ops_golang/deliveryservice/health.go
@@ -22,10 +22,10 @@ package deliveryservice
 import (
 	"database/sql"
 	"errors"
+	"fmt"
 	"net/http"
 	"strings"
 
-	"github.com/apache/trafficcontrol/lib/go-log"
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/lib/go-util"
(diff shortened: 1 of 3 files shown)


traffic_ops/traffic_ops_golang/deliveryservice/health.go
```diff
@@ -132,7 +160,24 @@ func addHealth(ds tc.DeliveryServiceName, data map[tc.CacheGroupName]tc.HealthDa
 		if cache.CacheGroup == nil {
 			continue // TODO warn?
 		}
-
+		if topology != "" {
+			if _, ok := cacheGroupNameMap[*cache.CacheGroup]; !ok {
+				continue
+			}
+			cacheCapabilities = make(map[string]bool)
```
ocket8888, line 167: This shadows a variable of the same name in an outer scope; it should be renamed to avoid collisions and confusion.

Also: nit but you know how big this needs to be. It's exactly the size of the cache server's `Capabilities` array, since it's not legal for those to contain duplicates. So you can construct with that in mind to cut down on re-allocations.

Also the value of each map key is unused, so to save an insignificant amount of space this could just be a `map[string]struct{}`

traffic_ops/traffic_ops_golang/deliveryservice/health.go
```diff
@@ -115,13 +116,40 @@ func getMonitorHealth(tx *sql.Tx, ds tc.DeliveryServiceName, monitorFQDNs []stri
...
+
+	var deliveryService tc.CRConfigDeliveryService
+	var ok bool
+	var topology string
+	var cacheGroupNameMap = make(map[string]bool)
+	var cacheCapabilities = make(map[string]bool)
+	var skip bool
+
+	if deliveryService, ok = crConfig.DeliveryServices[string(ds)]; !ok {
+		log.Errorln("delivery service not found in CRConfig")
```
ocket8888, line 128: Should this return an error instead of logging it and returning data as though the DS *were* found?

traffic_ops/traffic_ops_golang/deliveryservice/health.go
```diff
@@ -115,13 +116,40 @@ func getMonitorHealth(tx *sql.Tx, ds tc.DeliveryServiceName, monitorFQDNs []stri
...
+		return map[tc.CacheGroupName]tc.HealthDataCacheGroup{}, 0, 0
+	}
+
+	if deliveryService.Topology != nil {
+		topology = *deliveryService.Topology
+		if topology != "" {
+			if top, ok := crConfig.Topologies[topology]; !ok {
+				log.Errorf("CRConfig topologies does not contain DS topology: %s", topology)
```
ocket8888, line 136: Should this return an error instead of logging this and returning data as though the DS's topology *were* found?

traffic_ops/traffic_ops_golang/deliveryservice/health_test.go
```diff
@@ -0,0 +1,165 @@
...
+*/
+
+import (
+	"github.com/apache/trafficcontrol/lib/go-tc"
+
+	"encoding/json"
+	"testing"
+)
+
+const crConfigJson = `
```
ocket8888, line 25: Instead of making two JSON string constants and then doing nothing with them but unmarshalling into a real struct, why not just create that struct immediately, so you get compile-time type safety?

traffic_ops/traffic_ops_golang/deliveryservice/health.go
```diff
@@ -115,29 +118,26 @@ func getMonitorHealth(tx *sql.Tx, ds tc.DeliveryServiceName, monitorFQDNs []stri
 }
 
 // addHealth adds the given cache states to the given data and totals, and returns the new data and totals
-func addHealth(ds tc.DeliveryServiceName, data map[tc.CacheGroupName]tc.HealthDataCacheGroup, totalOnline uint64, totalOffline uint64, crStates tc.CRStates, crConfig tc.CRConfig) (map[tc.CacheGroupName]tc.HealthDataCacheGroup, uint64, uint64) {
+func addHealth(ds tc.DeliveryServiceName, data map[tc.CacheGroupName]tc.HealthDataCacheGroup, totalOnline uint64, totalOffline uint64, crStates tc.CRStates, crConfig tc.CRConfig) (error, map[tc.CacheGroupName]tc.HealthDataCacheGroup, uint64, uint64) {
```
ocket8888, line 121: nit: errors should be the last return value when there are multiple

ocket8888 (approved):

Issue #6828, opened 2022-05-12
Updating server without the id key in request body results in Internal Server Error 


ericholguin:
## This Bug Report affects these Traffic Control components:

- Traffic Ops

## Current behavior:

When updating a server without the `id` key in the request body it will result in an Internal Server Error. Previously a user could update a server without needing the `id` in the request body. This issue exists for both api v3.1 and v4.0. 

The same issue happens in api v3.1 when updating a server without the `profile` key. 

```http
HTTP/1.1 500 Internal Server Error
{
    "alerts": [
        {
            "text": "Internal Server Error",
            "level": "error"
        }
    ]
}
```

## Expected behavior:

Should update the server and return:
```http
HTTP/1.1 200 OK
{
    "alerts": [
        {
            "text": "Server updated",
            "level": "success"
        }
    ],
    "response": {...}
}
```

## Steps to reproduce:

Make a `PUT` request to `/api/4.0/servers` without the `id` in the request body this will result in an Internal Server Error. Do the same request but add the `id` key in the request body and it should return a 200 OK

## Comments:

Error.log: (v4)

```
ERROR: api.go:263: 2022-05-12T20:27:54.27204231Z: 172.20.0.11:57930 panic: (err: runtime error: invalid memory address or nil pointer dereference) stacktrace:
goroutine 39931 [running]:
github.com/apache/trafficcontrol/lib/go-util.Stacktrace()
	/tmp/go/src/github.com/apache/trafficcontrol/lib/go-util/util.go:32 +0x66
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware.WrapPanicRecover.func1.1()
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go:173 +0x66
panic({0xd164c0, 0x1506330})
	/usr/local/go/src/runtime/panic.go:838 +0x207
github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/server.Update({0xfa4850, 0xc0062629c0}, 0xc006240a00)
	/tmp/go/src/github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/server/servers.go:1496 +0xf57
```

Error.log (v3)
```
ERROR: api.go:263: 2022-05-12T20:56:58.578411246Z: 172.20.0.11:60012 panic: (err: runtime error: invalid memory address or nil pointer dereference) stacktrace:
[...]

PR #6829, opened 2022-05-12
Fix Internal Server Error when updating server without the id key

ericholguin:
Fixes: #6828 

Fixes issue where updating a server without an `id` key in the request body results in an Internal Server Error for api v3 and v4. This also fixes the issue when updating a server without a `profile` key in the request body results in an Internal Server Error for api v3 and v2.

In addition this fixes a previous bug in Traffic Ops that would not allow a user to update a server without the `id` key due to:

https://github.com/apache/trafficcontrol/blob/8c13de7054acf8554412c59fdf543ff30759b3ab/traffic_ops/traffic_ops_golang/server/servers.go#L653-L672

When the `id` key was omitted it resulted in that s.ID always being nil and resulting in the incorrect error message and incorrect 400 Bad Request.

<hr/>

- Traffic Ops

Make a `PUT` request to `/api/4.0/servers` without the `id` in the request body this should return a `200 OK`
Make a `PUT` request to `/api/3.1/servers` without the `id` in the request body this should return a `200 OK`

Make a `PUT` request to `/api/3.1/servers` without the `profile` in the request body this should return a `200 OK`
Make a `PUT` request to `/api/2.0/servers` without the `profile` in the request body this should return a `200 OK`

## PR submission checklist

### DIFF
--- a/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
+++ b/traffic_ops/traffic_ops_golang/dbhelpers/db_helpers.go
@@ -2124,10 +2124,17 @@ func UpdateServerProfilesForV4(id int, profile []string, tx *sql.Tx) error {
 }
 
 // UpdateServerProfileTableForV2V3 updates CommonServerPropertiesV40 struct and server_profile table via Update (server) function for API v2/v3.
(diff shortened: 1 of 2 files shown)


rimashah25:
LGTM

rimashah25 (approved):

rawlinp (approved):

PR #6838, opened 2022-05-17
Add TO option to cache server update status in memory

rawlinp:
To reduce load on the Traffic Ops database, add an option to
periodically read all server update status data from the database and
serve all requests to /servers/{hostname}/update_status from memory.

With this option enabled, `t3c` runs on cache servers that don't have
updates pending don't cause any TODB queries, which means TODB has more
capacity for actually propagating changes.

The tradeoff is that it can take up to
`server_update_status_cache_refresh_interval_sec` seconds for cache
servers to see updates/revalidations pending. However, this is better
because it allows a cache server to get updates sooner than it would
otherwise (due to being able to run `t3c` more frequently).

<hr/>

- Documentation
- Traffic Ops

Ensure the automated tests pass. Configure `server_update_status_cache_refresh_interval_sec` in cdn.conf to a number > 0, queue updates on various cache servers (edges and mids), and ensure that `GET /servers/{hostname}/update_status` returns the expected data.

## PR submission checklist
- [x] This PR has tests 


traffic_ops/traffic_ops_golang/server/servers_update_status.go
```diff
@@ -175,3 +180,248 @@ ORDER BY s.id
...
+	if err = topologyCachegroupRows.Err(); err != nil {
+		return nil, fmt.Errorf("iterating over topology cachegroup rows: %w", err)
+	}
+
+	serverUpdateStatuses := make(map[string][]tc.ServerUpdateStatusV40, len(serversByID))
+	for serverID, server := range serversByID {
+		updateStatus := tc.ServerUpdateStatusV40{
+			HostName:             server.hostname,
+			UpdatePending:        server.configUpdateTime.After(*server.configApplyTime),
+			RevalPending:         server.revalUpdateTime.After(*server.revalApplyTime),
```
srijeet0406, line 404: Can the `revalApplyTime` or `configApplyTime` be `nil`? In that case, this will panic.

rawlinp: No, not unless the DB schema is changed to make the fields nullable. Originally, I had these fields as non-pointers, because generally that's what we want when DB columns are non-null, but it was causing weird pointer issues in the unit tests that I couldn't really explain.

That said, if the DB columns ever _do_ become nullable and this code isn't changed, it might actually be better to crash than log an error. Fail fast, fail loud. Logged errors would be ignored, and we'd be sitting around wondering why none of our servers are able to queue updates.

srijeet0406 (approved):
Code looks great!
Manual and automated tests passed.

PR #6848, opened 2022-05-20
Change t3c non-topo non-parent line to origin

rob05c:
Previously, t3c non-topology DSes which didn't use parents would omit
the parent= from the parent.config line. The ATS docs seem to
indicate that should have worked, but it seems to only work for very
specific settings, e.g. parent_is_proxy=false made the line break
(along with potentially other settings).

This changes those DSes to set parent=origin, which appears to make
ATS behave correctly with parent_is_proxy, go_direct, and other
necessary settings.

Note Topology DSes were already doing this, this effectively
changes non-topo to match topo generation.

Run tests. Generate config for a non-topology DS that doesn't use parents, verify parent.config line includes parent=origin, verify ATS works as expected with the generated parent.config and remap.config lines.

Not in a release.

## PR submission checklist
- [x] This PR has tests 
- ~[x] This PR has a CHANGELOG.md entry~ no changelog, bug isn't in a release 


jpappa200 (approved):
👍  LGTM test runs as expected.

PR #6851, opened 2022-05-21
Refactor Servers Test

ericholguin:
This PR refactors **v4/servers_test.go**, **v3/servers_test.go**, in order to make tests more granular, providing transparency on what is being tested and reducing the need to dig through test code to confirm what was tested. In addition this new test structure should also make it easier for new tests to be added as well as reduce redundant code.

- Adds data driven testing
    - Descriptive test names
- Sub tests for organization
    - Use GoLangs test runner
- Use assert functionality
    - Streamlines code
    - Reduces nested conditionals
- Adds expectation functions
     - Reusable expectations

<hr/>

- Traffic Ops Testing

Run TO Integration tests

## PR submission checklist
- [x] This PR has tests 


traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -26,1017 +26,562 @@ import (
...
-		_, reqInf, err := TOSession.GetServersWithHdr(&params, header)
-		if err != nil {
-			t.Fatalf("Expected no error, but got %v", err.Error())
-		}
-		if reqInf.StatusCode != http.StatusNotModified {
-			t.Fatalf("Expected 304 status code, got %v", reqInf.StatusCode)
+func validateExpectedServers(expectedHostNames []string) utils.CkReqFunc {
+	return func(t *testing.T, _ toclientlib.ReqInf, resp interface{}, _ tc.Alerts, _ error) {
+		assert.RequireNotNil(t, resp, "Expected response to not be nil.")
+		serverResp := resp.([]tc.ServerV30)
```
ocket8888, line 403: Should use `ServerV3`.

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -26,1017 +26,562 @@ import (
...
+		var notInResponse []string
+		serverMap := make(map[string]struct{})
```
ocket8888, line 405: nit: you know exactly how big this map needs to be, so it could be allocated immediately instead of as-needed during iteration of the response.

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -26,1017 +26,562 @@ import (
...
+				case "Status":
+					assert.Equal(t, expected, *server.Status, "Expected Status to be %v, but got %v", expected, *server.Status)
+				case "TCPPort":
+					assert.Equal(t, expected, *server.TCPPort, "Expected TCPPort to be %v, but got %v", expected, *server.TCPPort)
+				case "Type":
+					assert.Equal(t, expected, server.Type, "Expected Type to be %v, but got %v", expected, server.Type)
+				case "TypeID":
+					assert.Equal(t, expected, *server.TypeID, "Expected Type to be %v, but got %v", expected, *server.TypeID)
+				default:
+					t.Errorf("Expected field: %v, does not exist in response", field)
```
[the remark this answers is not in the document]
ocket8888: Also there are a lot of dereferences here with no `nil` checks, which could crash the test if Traffic Ops gives bad responses.

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -26,1017 +26,562 @@ import (
...
-	if len(resp.Response) > 1 {
-		t.Errorf("Expected exactly one server to exist by hostname '%s' - actual: %d", hostName, len(resp.Response))
-		t.Logf("Testing will proceed with server: %+v", resp.Response[0])
-	}
-
-	respServer = resp.Response[0]
+		for method, testCases := range methodTests {
+			t.Run(method, func(t *testing.T) {
+				for name, testCase := range testCases {
+					server := tc.ServerV30{}
```
ocket8888, line 294: should use `ServerV3`

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -26,1017 +26,562 @@ import (
...
-		t.Log("Response: ", *server.HostName, " ", resp)
-		if err != nil {
-			t.Errorf("could not CREATE servers: %v", err)
+func validateServerTypeIsNotMid() utils.CkReqFunc {
+	return func(t *testing.T, _ toclientlib.ReqInf, resp interface{}, _ tc.Alerts, _ error) {
+		assert.RequireNotNil(t, resp, "Expected response to not be nil.")
+		serverResp := resp.([]tc.ServerV30)
+		for _, server := range serverResp {
+			if server.Type == tc.CacheTypeMid.String() {
+				t.Errorf("Expected to find no %s-typed servers but found server %s", tc.CacheTypeMid, *server.HostName)
```
ocket8888, line 424: this will segfault if Traffic Ops responds with a `null` or undefined `hostName`

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -26,1017 +26,562 @@ import (
...
-		resp, _, err := TOSession.GetServerDetailsByHostNameWithHdr(*server.HostName, nil)
-		if err != nil {
-			t.Errorf("cannot GET Server Details by name: %v - %v", err, resp)
-		}
+func GetPhysLocationId(t *testing.T, name string) func() int {
+	return func() int {
+		physLocResp, _, err := TOSession.GetPhysLocationByNameWithHdr(name, nil)
+		assert.NoError(t, err, "Get PhysLocation Request failed with error:", err)
+		assert.Equal(t, 1, len(physLocResp), "Expected response object length 1, but got %d", len(physLocResp))
+		assert.NotNil(t, physLocResp[0].ID, "Expected id to not be nil")
```
ocket8888, line 483: An `int` can never be `nil`; this assertion is unnecessary

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -26,1017 +26,562 @@ import (
...
+
+	for _, server := range servers.Response {
+		delResp, _, err := TOSession.DeleteServerByID(*server.ID)
+		assert.NoError(t, err, "Could not delete Server: %v - alerts: %+v", err, delResp.Alerts)
+		// Retrieve Server to see if it got deleted
+		params := url.Values{}
+		params.Set("id", strconv.Itoa(*server.ID))
+		getServer, _, err := TOSession.GetServersWithHdr(&params, nil)
+		assert.NoError(t, err, "Error deleting Server for '%s' : %v - alerts: %+v", *server.HostName, err, getServer.Alerts)
+		assert.Equal(t, 0, len(getServer.Response), "Expected Server '%s' to be deleted", *server.HostName)
```
ocket8888, line 585: these lines will segfault if Traffic Ops returns a server with a `null` or undefined `hostName`

traffic_ops/testing/api/v3/tc-fixtures.json
```diff
@@ -1543,6 +1543,10 @@
         {
             "xmlId": "test-ds-server-assignments",
             "serverNames": ["test-ds-server-assignments"]
+        },
+        {
+            "xmlId": "ds3",
+            "serverNames": ["atlanta-edge-14"]
```
ocket8888, line 1549: why'd you add this?

ocket8888 (commented):
These are some very large changes, so for now I just looked through v3 on the assumption that my comments will apply to probably both of them. There's v4-specific stuff to look at, but that'll be easier once v3 is out of the way and I can just skim over the parts that are the same.

traffic_ops/testing/api/v3/servers_test.go
```diff
@@ -26,1017 +26,562 @@ import (
...
-	_, reqInf, err := TOSession.UpdateServerByIDWithHdr(*remoteServer.ID, remoteServer, header)
-	if err == nil {
-		t.Errorf("Expected error about precondition failed, but got none")
-	}
-	if reqInf.StatusCode != http.StatusPreconditionFailed {
-		t.Errorf("Expected status code 412, got %v", reqInf.StatusCode)
+func validateServerFields(expectedResp map[string]interface{}) utils.CkReqFunc {
+	return func(t *testing.T, _ toclientlib.ReqInf, resp interface{}, _ tc.Alerts, _ error) {
+		assert.RequireNotNil(t, resp, "Expected response to not be nil.")
+		serverResp := resp.([]tc.ServerV30)
```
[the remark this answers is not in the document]
ericholguin: Client uses `ServerV30` and all the client functions return `ServerV30`, I actually don't think there is a `ServerV3`.

traffic_ops/testing/api/v3/tc-fixtures.json
ericholguin: This test: `UpdateDSGetServerDSID` needs it as a prerequisite.

ericholguin:
Tests will fail until this is fixed #6856

PR for fix here: #6857

traffic_ops/testing/api/v3/servers_test.go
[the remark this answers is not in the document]
ocket8888: You're right. That's not supposed to happen; each major version should have an alias for the latest minor version and that's what the client is supposed to be using. But that's not your PR's fault, so `ServerV30` is the right choice here.

traffic_ops/testing/api/v3/tc-fixtures.json
ocket8888: right, but that dependency was also added by this PR, so then my question becomes: "Why require a new Delivery Service Servers instead of using an existing one?"

ericholguin: The existing ones do not work for that test and they are also being used for other server tests.

ocket8888: > The existing ones do not work for that test

I can't see why at a glance, but I'll take your word for it.

ocket8888 (approved):

PR #6852, opened 2022-05-21
Move duplicate rand functionality to test rand

ericholguin:
This PR removes duplicate rand functions and has test files use the rand functions declared in `traffic_ops_golang/test/rand.go` removing the need to import `"math/rand"` in multiple files.

<hr/>

- Traffic Control Cache Config Unit Tests
- Traffic Monitor Unit Tests
- Traffic Ops Unit Tests
- Grove Unit Tests

Run the unit tests for the components above

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/traffic_ops_golang/crconfig/config_test.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/config_test.go
@@ -32,8 +32,8 @@ import (
 
 func ExpectedGetConfigParams(domain string) []CRConfigConfigParameter {
 	return []CRConfigConfigParameter{
-		{"tld.ttls.foo" + *test.RandStr(), *test.RandStr()},
-		{"tld.soa.bar" + *test.RandStr(), *test.RandStr()},
+		{"tld.ttls.foo" + test.RandStr(), test.RandStr()},
+		{"tld.soa.bar" + test.RandStr(), test.RandStr()},
 		{"domain_name", domain},
 	}
 }
--- a/traffic_ops/traffic_ops_golang/crconfig/config_test.go
+++ b/traffic_ops/traffic_ops_golang/crconfig/config_test.go
@@ -26,13 +26,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/test"
 	"gopkg.in/DATA-DOG/go-sqlmock.v1"
 )
 
 func ExpectedGetConfigParams(domain string) []CRConfigConfigParameter {
 	return []CRConfigConfigParameter{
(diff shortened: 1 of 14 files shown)


traffic_ops/traffic_ops_golang/test/rand.go
```diff
@@ -36,3 +38,68 @@ func RandStrArray() []string {
...
+		uint8(rand.Int31n(256)),
+		uint8(rand.Int31n(256)),
+		uint8(rand.Int31n(256)),
+		uint8(rand.Int31n(256)),
+		uint8(rand.Int31n(256)),
+		uint8(rand.Int31n(256)),
+		uint8(rand.Int31n(256)),
+	}).String()
+	return &ip
+}
```
ocket8888, line 105: GoDocs?

ocket8888 (commented):
Possible improvement: occurrences of `\*test\.Rand` are just a bit more common than those of `[^*]test\.Rand` - should these maybe return values instead of pointers?

ericholguin:
> Possible improvement: occurrences of `\*test\.Rand` are just a bit more common than those of `[^*]test\.Rand` - should these maybe return values instead of pointers?

We would have to declare variables wherever a pointer is needed because you can't get the address of a function value returned in go. So instead of just `test.RandStr()` it'd be: 

```go
randStr := test.RandStr()
...
Fqdn:            &randStr,
...
```

And we'd either be pointing to the same random string variable where needed or create multiple random string variables and point to those. Is this what you mean or do you know of another way?

ocket8888:
You could make use of [the `github.com/apache/trafficcontrol/lib/go-util` package](https://pkg.go.dev/github.com/apache/trafficcontrol/lib/go-util)'s functions `IntPtr`, `StrPtr`, etc. So it'd be
```go
...
Fqdn:   util.StrPtr(test.RandString())
...
```
no forward declaration required

traffic_ops/traffic_ops_golang/crconfig/config_test.go
```diff
@@ -26,13 +26,14 @@ import (
 	"testing"
 	"time"
 
+	"github.com/apache/trafficcontrol/traffic_ops/traffic_ops_golang/test"
 	"gopkg.in/DATA-DOG/go-sqlmock.v1"
```
ocket8888, line 30: By convention we generally leave a blank line between project-internal imports and external packages

traffic_ops/traffic_ops_golang/test/rand.go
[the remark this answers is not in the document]
ocket8888: I don't know how this got in here again. It's clearly already been in a review, and in fact already resolved.

Issue #4351, opened 2020-01-29
Provide a better message on DS create/update/delete to indicate if a CDN snapshot and/or queue updates is required


mitchell852:
## I'm submitting a ...

- [ ] bug report
- [ ] new feature / enhancement request
- [x] improvement request (usability, performance, tech debt, etc.)
- [ ] other 

## Traffic Control components affected ...

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [x] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:
Currently when you successfully create/edit/delete a delivery service you get a very simple message like "DS create/updated/deleted successfully". However, creating, updating and deleting a delivery service requires more than simply persisting the change to the database. In some cases it requires a cdn snapshot for the changes to take affect and/or an ATS config file update via a queue updates of all servers in the cdn.

## Expected / new behavior:
IMO it would be very helpful if the success message was something like:

- DS updated successfully. A CDN snapshot is required for the changes to take affect.
- DS updated successfully. Queue updates on all servers in the cdn for the changes to take affect.

- DS created successfully. Queue updates on all servers in the cdn then perform a CDN snapshot for the changes to take affect.

- DS deleted successfully. Perform a CDN snapshot then queue updates on all servers in the cdn for the changes to take affect.

Of course in the case of an update this would require inspecting which properties changed to determine the appropriate message. 

## Minimal reproduction of the problem with instructions:
create/update or delete a DS via the TO API or TP.

mitchell852:
it's very unclear what the user is supposed to do after making changes to a ds. should i queue updates? snapshot? both? i'm going to mark this as high impact as i believe it to be "high value" from a user perspective but i know that's very debatable.

PR #6855, opened 2022-05-24
Updated success message (curl and TP) for deleting a DS

rimashah25:
Related: #4351 

<hr/>

- Traffic Ops
- Traffic Portal

Create a DS or region or server
Delete a DS (see updated message), region and server (have same message as before i.e. `server/region was deleted.` on TO and `server/region deleted` on TP.

Updated message for DS
```
curl -X DELETE -k -H "Accept: application/json" --cookie "mojolicious=<mojolicious cookie>" https://localhost:8443/api/4.0/deliveryservices/<ds id>

{"alerts":[{
       "text":"ds was deleted. Perform a CDN snapshot then queue updates on all servers in the cdn for the changes to take affect.",
       "level":"success"
}]
```

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
+++ b/traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
@@ -301,7 +301,7 @@ var FormEditDeliveryServiceController = function(deliveryService, origin, topolo
 						function() {
 							messageModel.setMessages([
 								{ level: 'success', text: 'Delivery service [ ' + deliveryService.xmlId + ' ] deleted.' },
-								{ level: 'info', text: 'Perform a CDN snapshot then queue updates on all servers in the cdn for the changes to take affect.' }
+								{ level: 'info', text: 'Perform a CDN snapshot then queue updates on all servers in the cdn for the changes to take effect.' }
 							], true);
 							locationUtils.navigateToPath('/delivery-services');
 						},
--- a/traffic_ops/traffic_ops_golang/api/shared_handlers.go
+++ b/traffic_ops/traffic_ops_golang/api/shared_handlers.go
@@ -326,7 +326,7 @@ func DeleteHandler(deleter Deleter) http.HandlerFunc {
 		func(w http.ResponseWriter, r *http.Request, message string) {
 			if deleter.GetType() == "ds" {
 				alerts := tc.CreateAlerts(tc.SuccessLevel, message)
-				alerts.AddNewAlert(tc.InfoLevel, "Perform a CDN snapshot then queue updates on all servers in the cdn for the changes to take affect.")
+				alerts.AddNewAlert(tc.InfoLevel, "Perform a CDN snapshot then queue updates on all servers in the cdn for the changes to take effect.")
 				WriteAlerts(w, r, http.StatusOK, alerts)
 			} else {
 				WriteRespAlert(w, r, tc.SuccessLevel, message)


ocket8888:
Should this new message maybe go in an `info`-level alert instead of appending it to the success message?

traffic_portal/test/integration/Data/deliveryservices.ts
```diff
@@ -357,17 +357,17 @@ export const deliveryservices = {
 				{
 					description: "delete a delivery service",
 					Name: "tpdservice1",
-					validationMessage: "Delivery service [ tpdservice1 ] deleted"
+					validationMessage: "Delivery service [ tpdservice1 ] deleted."
```
ocket8888, line 360: Do you think it's better with the period? I think a fair few of the Alerts we send from the API have punctuation, but also a lot don't. To be fair, the actual Alert TO sends has one.

rimashah25: I think in this case a period makes sense since there are two messages, even though they are as two separate level.

ocket8888 (approved):
LGTM, TP tests pending

traffic_portal/app/src/common/modules/form/deliveryService/edit/FormEditDeliveryServiceController.js
```diff
@@ -299,7 +299,10 @@ var FormEditDeliveryServiceController = function(deliveryService, origin, topolo
 				deliveryServiceService.deleteDeliveryService(deliveryService)
 					.then(
 						function() {
-							messageModel.setMessages([ { level: 'success', text: 'Delivery service [ ' + deliveryService.xmlId + ' ] deleted' } ], true);
+							messageModel.setMessages([
```
mitchell852, line 302: why? TP can just show the message(s) sent from the API. This approach has essentially hard-coded the api message.

ocket8888: two main reasons

1. That's what it already was doing, so this PR merely isn't a fix for it - it's not introducing the unwanted behavior.
2. This hard-codes the response, because all Traffic Ops returns is "ds was deleted". The custom Alert puts the XMLID in the message. Fixing Traffic Ops to include that information is involved, because Delivery Services use the CRUDer, so it's difficult to change nearly anything about the endpoint without retooling the whole API. Which is out of scope for this change.

rimashah25: Exactly what Brennan said. He had the same question and I explained why it is difficult to change things on TP since cruder on TO gives a generic message, hence developers of TP set a different message than results.alerts.

mitchell852: but i see

`				alerts := tc.CreateAlerts(tc.SuccessLevel, message)
				alerts.AddNewAlert(tc.InfoLevel, "Perform a CDN snapshot then queue updates on all servers in the cdn for the changes to take effect.")
`

why not just show those. i think it's fine that the xmlId is lost in the UI.

ocket8888: I'm not sure either. It should be changed, but that can be done in a separate PR.

rimashah25: @ocket8888 @mitchell852 I am creating another to remove hardcoding and updating test to use the updated validationMessage.

PR #6857, opened 2022-05-25
Fix server not appearing after updating its profile

### LINKED ISSUE #6856: Server does not show up in GET response but exists in DB

- Traffic Ops

## Current behavior:

This is a very specific scenario, currently it works as such:

There must exist an active MSO Delivery Service with an origin server assigned (Must have the status ONLINE or REPORTED),
If you update the status of the origin to ADMIN_DOWN this should result in a 409, (Making sure all the other fields are the exact same):

```http
HTTP/1.1 409 Conflict
Transfer-Encoding: chunked
{
    "alerts": [
        {
            "text": "setting server status to 'ADMIN_DOWN' would leave Active Delivery Service #3 with no 'ONLINE' or 'REPORTED' ORG servers",
            "level": "error"
        }
    ]
}
```

However if you change the profileNames, replacing the previous profile to another you will get the same result but that server will no longer exist in any `GET` requests. However the server is still in the DB.

This only happens if you remove the previous profileName, adding additional profiles will not create this issue. 

The only `GET` request that still shows that the previous servers exist is `/api/4.0/deliveryservices/{dsId}/servers/eligible`

## Expected behavior:

Server should be appearing in any and all valid `GET` requests

## Steps to reproduce:

There must exist an active MSO Delivery Service with an origin server assigned (Must have the status ONLINE or REPORTED),
If you update the status of the origin to ADMIN_DOWN this should result in a 409, (Making sure all the other fields are the exact same) Update the profileNames to something other than the current profile.

ericholguin:
Fixes #6856 

<hr/>

- Traffic Ops

There must exist an active MSO Delivery Service with an origin server assigned (Must have the status ONLINE or REPORTED),
If you update the status of the origin to ADMIN_DOWN this should result in a 409, (Making sure all the other fields are the exact same) Update the profileNames to something other than the current profile. Perform a GET request using that same profile id, that server should be returned. Do a GET request with no parameters you should see the correct server count along with that server in the response.

## PR submission checklist

### DIFF
--- a/traffic_ops/traffic_ops_golang/server/servers.go
+++ b/traffic_ops/traffic_ops_golang/server/servers.go
@@ -1499,6 +1499,8 @@ func Update(w http.ResponseWriter, r *http.Request) {
 	originalStatusID := *original.StatusID
 
 	var server tc.ServerV40
+	var serverV3 tc.ServerV30
+	var legacyServer tc.ServerNullableV2
 	var statusLastUpdatedTime time.Time
 	server.ID = new(int)
 	*server.ID = inf.IntParams["id"]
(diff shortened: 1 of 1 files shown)


ocket8888 (approved):
Can't reproduce bug after these changes

Issue #6879, opened 2022-06-02
t3c should maintain onwer/group settings on config files


jpappa200:
## This Improvement request (usability, performance, tech debt, etc.) affects these Traffic Control components:

- Traffic Control Cache Config (`t3c`, formerly ORT)

## Current behavior:

Right now t3c will only enforce config file owner/group settings when the file is written because of a change. if for some reason the owner/group is changed it will not be set back until the file contents change and is updated. 

## New behavior:

t3c diff will flag the file for replacement if not owned by ats or not a member of the ats group.

PR #6880, opened 2022-06-02
t3c will ensure config files are owned by ats  

jpappa200:
Closes: #6879

<hr/>

Build and install rpm, change file owner and group on some config files, queue updates on the cache, run t3c apply in syncds or badass, and verify owner and group have been set back to ats.

## PR submission checklist
- [ ] This PR has tests 


cache-config/t3c-diff/t3c-diff.go
```diff
@@ -49,6 +49,8 @@ func main() {
 	version := getopt.BoolLong("version", 'V', "Print version information and exit")
 	lineComment := getopt.StringLong("line_comment", 'l', "#", "Comment symbol")
 	mode := getopt.IntLong("file-mode", 'm', 0644, "file mode default is 644")
+	uid := getopt.IntLong("file-uid", 'u', 0, "file uid default is 0")
+	gid := getopt.IntLong("file-gid", 'g', 0, "file gid default is 0")
```
ocket8888, line 53: The help text that is the final argument to `getopt.IntLong` should be

> ... the usage message for the option

The default values for options are already automatically printed in the help text output, so the help text needn't include them at all.

Speaking of the default values, though, is `0` really the best default for this? That seems prone to permissions errors when `t3c` isn't running as the root user. Plus, when I run a command-line utility, I don't typically expect it to create anything owned by any user or group besides the user and group that the process running it uses. Maybe the default should be the running process's uid/gid?

rob05c: The new flag also needs added to t3c-diff/README.md (which also gets transformed into the man page)

jpappa200: README.md has been updated.

cache-config/t3c-diff/README.md
```diff
@@ -58,12 +58,37 @@ If one file exists but the other doesn't, it will always be a diff.
...
 
+-a, -\-file-a
+
+    Path to first diff file, can also be stdin.
+
+-b, -\-file-b
+    Path to second diff file, can also be stdin.
+
+-g, -\-file-gid
+    Group id the file being checked should have.
```
ocket8888, line 77: This is a fine description - why not put it in the help text?

cache-config/t3c-diff/t3c-diff.go
ocket8888, line 53: > default is 0

is no longer accurate

cache-config/t3c-diff/README.md
```diff
@@ -38,7 +38,7 @@ t3c-diff - Traffic Control Cache Configuration contextual diff tool
 
 # SYNOPSIS
 
-t3c-diff \<file-a\> \<file-a\>
+t3c-diff \-a \<file-a\> \- b \<file-b\> \-l \<line_comment\> \-m \<file-mode\> \-u \<file-uid\> \-g \<file-gid\>
```
ocket8888, line 41: There's an extraneous space here between `-` and `b`

cache-config/t3c-diff/README.md
jpappa200: Updated help text

cache-config/t3c-diff/t3c-diff.go
jpappa200: Updated help text here too.

cache-config/t3c-diff/README.md
jpappa200: extraneous space removed

ocket8888:
To test this I made a copy of `t3c-diff`'s README, but it doesn't seem to detect that it doesn't have the right owner:
```console
$ stat ./README.md 
  File: ./README.md
  Size: 3471      	Blocks: 8          IO Block: 4096   regular file
Device: 10303h/66307d	Inode: 17697181    Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/ocket8888)   Gid: ( 1000/ocket8888)
Access: 2022-06-08 12:55:22.414520586 -0600
Modify: 2022-06-08 12:55:22.414520586 -0600
Change: 2022-06-08 12:55:22.414520586 -0600
 Birth: 2022-06-08 12:55:22.414520586 -0600
$ ./t3c-diff -l '<!--' -m 0664 -u 0 -g 0 -a ./README.md -b ./README.new.md
$ echo $?
0
```

jpappa200:
the default value if left blank is 0 
`uid := getopt.IntLong("file-uid", 'u', 0, "User id the file being checked should have, default is running process's uid")`
`gid := getopt.IntLong("file-gid", 'g', 0, "Group id the file being checked should have, default is running process's gid")`

if the value is 0 it will use the running process's uid/gid 
`if *uid == 0 {
		*uid = os.Geteuid()
	}`
`if *gid == 0 {
		*gid = os.Getgid()
	}`
If you ran it as root it would find the difference. I could change the default value but it seemed logical to say that only the root user could say it should be owned by root.

ocket8888 (approved):

PR #6881, opened 2022-06-02
Default routeID to 0, if it is not present in the request context

srijeet0406:
This PR is not related to any issue. It sets the routeID to 0 if it is absent in the request context.

<hr/>

- Traffic Ops

Make a request to a route that doesn't have a route ID
Tail the access logs and make sure that a route ID of `0` is logged

- master
## PR submission checklist
- [ ] This PR has tests 

### DIFF
--- a/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go
+++ b/traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go
@@ -212,7 +212,11 @@ func WrapAccessLog(secret string, h http.Handler) http.HandlerFunc {
 					imsType = IMSMISS
 				}
 			}
-			log.EventfRaw(`%s - %s [%s] "%v %v?%v %s" %v %v %v "%v" %v %s`, r.RemoteAddr, user, time.Now().Format(AccessLogTimeFormat), r.Method, r.URL.Path, r.URL.RawQuery, r.Proto, iw.Code, iw.ByteCount, int(time.Now().Sub(start)/time.Millisecond), r.UserAgent(), r.Context().Value(RouteID), imsType)
+			routeID := r.Context().Value(RouteID)
+			if routeID == nil {
+				routeID = 0
+			}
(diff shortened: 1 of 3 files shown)


traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go
```diff
@@ -212,7 +212,11 @@ func WrapAccessLog(secret string, h http.Handler) http.HandlerFunc {
 					imsType = IMSMISS
 				}
 			}
-			log.EventfRaw(`%s - %s [%s] "%v %v?%v %s" %v %v %v "%v" %v %s`, r.RemoteAddr, user, time.Now().Format(AccessLogTimeFormat), r.Method, r.URL.Path, r.URL.RawQuery, r.Proto, iw.Code, iw.ByteCount, int(time.Now().Sub(start)/time.Millisecond), r.UserAgent(), r.Context().Value(RouteID), imsType)
+			routeID := r.Context().Value(RouteID)
```
rawlinp, line 215: Can we safely type-assert this to an int instead of checking nil? It looks strange because routeID looks like a pointer, yet we're assigning an int to it.
```
routeID, _ := r.Context().Value(RouteID).(int)
log.EventfRaw(...)
```
if it was nil, this would make routeID the zero-value of type int, which is 0. If non-nil, we have a real routeID int. We could also change the log format from `%v` to `%d` for the route ID

srijeet0406: Done

traffic_ops/traffic_ops_golang/routing/middleware/wrappers.go
```diff
@@ -212,7 +212,8 @@ func WrapAccessLog(secret string, h http.Handler) http.HandlerFunc {
 					imsType = IMSMISS
 				}
 			}
-			log.EventfRaw(`%s - %s [%s] "%v %v?%v %s" %v %v %v "%v" %v %s`, r.RemoteAddr, user, time.Now().Format(AccessLogTimeFormat), r.Method, r.URL.Path, r.URL.RawQuery, r.Proto, iw.Code, iw.ByteCount, int(time.Now().Sub(start)/time.Millisecond), r.UserAgent(), r.Context().Value(RouteID), imsType)
+			routeID, _ := r.Context().Value(RouteID).(int)
```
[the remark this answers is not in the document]
srijeet0406: ugh, can't believe I pushed without testing, my bad. Should be done now.

rawlinp (approved):

PR #6883, opened 2022-06-03
Fix t3c cache to invalidate on version change

rob05c:
Previously, both upgrades and downgrades of `trafficcontrol-cache-config` could cause strange issues in config generation, from trying to re-use a cache with incompatibilities resulting in missing or different fields.

This fixes t3c to invalidate the cache if the running version is different than the version which created the cache.

Run a previous version of t3c, upgrade, verify from log messages and/or HTTP IMS requests that the cache is not used.

- 6.0.0
- 6.0.1
- 6.0.2
- 6.1.0

## PR submission checklist
- ~[x] This PR has tests~ no tests, testing requires HTTP requests and thus can't be unit tested, and integration tests don't support multiple versions yet.
- ~[x] This PR has documentation~ no docs, cache behavior is not documented.

### DIFF
--- a/cache-config/t3cutil/getdata.go
+++ b/cache-config/t3cutil/getdata.go
@@ -54,6 +54,10 @@ type TCCfg struct {
 
 	// OldCfg is the previously fetched ConfigData, for 'config' requests. May be nil.
 	OldCfg *ConfigData
+
+	// T3CVersion is the version of the t3c app ecosystem
+	// This value will be the same for any t3c app.
+	T3CVersion string
 }
 
 func GetDataFuncs() map[string]func(TCCfg, io.Writer) error {
@@ -254,7 +258,7 @@ func SetUpdateStatusCompat(cfg TCCfg, serverName tc.CacheName, configApply, reva
 
 // WriteConfig writes the Traffic Ops data necessary to generate config to output.
 func WriteConfig(cfg TCCfg, output io.Writer) error {
-	cfgData, err := GetConfigData(cfg.TOClient, cfg.TODisableProxy, cfg.CacheHostName, cfg.RevalOnly, cfg.OldCfg)
+	cfgData, err := GetConfigData(cfg.TOClient, cfg.TODisableProxy, cfg.CacheHostName, cfg.RevalOnly, cfg.OldCfg, cfg.T3CVersion)
 	if err != nil {
 		return errors.New("getting config data: " + err.Error())
 	}
--- a/cache-config/t3cutil/getdatacfg.go
+++ b/cache-config/t3cutil/getdatacfg.go
@@ -39,6 +39,10 @@ import (
 const TrafficOpsProxyParameterName = `tm.rev_proxy.url`
 
 type ConfigData struct {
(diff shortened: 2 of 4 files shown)


cache-config/t3cutil/getdata.go
```diff
@@ -54,6 +54,10 @@ type TCCfg struct {
 
 	// OldCfg is the previously fetched ConfigData, for 'config' requests. May be nil.
 	OldCfg *ConfigData
+
+	// T3cVersion is the version of the t3c app ecosystem
```
ocket8888, line 58: should be a capital "C" to match the field name

cache-config/t3cutil/getdata.go
```diff
@@ -54,6 +54,10 @@ type TCCfg struct {
...
+	// This value will be the same for any t3c app.
+	T3CVersion string
```
ocket8888, line 60: Should this be override-able through the configuration file? It just seems a little strange to tell a utility which could be wrong when the utility ostensibly always knows the correct value.

cache-config/t3cutil/getdatacfg.go
```diff
@@ -173,15 +177,27 @@ func MakeReqMetaData(respHdr http.Header) ReqMetaData {
...
 	start := time.Now()
 	defer func() { log.Infof("GetTOData took %v\n", time.Since(start)) }()
 
 	toIPs := &sync.Map{} // each Traffic Ops request could get a different IP, so track them all
 	toData := &ConfigData{}
+	toData.Version = version
 	toData.MetaData.CacheHostName = cacheHostName
 
+	if oldCfg != nil && oldCfg.Version != toData.Version {
+		log.Infof("old config version '%v' doesn't match current version '%v', old config will not be used!\n", oldCfg.Version, toData.Version)
```
ocket8888, line 197: For the best safety at compile-time, format-strings should use the most specific formatting parameter possible - `%s` in this case.

ocket8888 (changes requested):

cache-config/t3cutil/getdatacfg.go
rob05c: For the record, I don't agree, in my opinion it reduces readability without increasing safety. It also makes this code not idiomatic with the rest of t3c and lib/go-atscfg.

But it's more important to get this fix in. Changed.

cache-config/t3cutil/getdata.go
rob05c: > It just seems a little strange to tell a utility which could be wrong when the utility ostensibly always knows the correct value.

To clarify: it's in this struct, because this struct is what get passed from the app (`t3c-apply`) to the library function that needs it in `t3cutil`. It should never be changed. 

I could have made it impossible to change, by making it a function of the struct instead. But that isn't idiomatic Go, and it adds more complexity and increases the dev cost to read and write it. I decided the complexity and readability cost wasn't worth the very minor safety gain (especially considering Go standard libraries use non-const variables all over the place, that are equally "unsafe").

ocket8888 (approved):
Sorry, I forgot about this over my vacation

PR #6891, opened 2022-06-09
Traffic Portal: fix profiles servers table

ocket8888:
This fixes a bug where the view of a Profile's servers would show all existing servers without regard for what actually uses the Profile. It also eliminates some cruft that has been rendered obsolete by the generic table's ability to apply filters from query string parameters.

<hr/>

- Traffic Portal

- master

This is a regression because the latest version works, but if we cut a new release today it'd be broken.

## PR submission checklist
- [ ] This PR has tests - behavior should be largely unchanged, so existing tests should suffice


shamrickus (approved):
LGTM
* Properly filters profiles servers

PR #6895, opened 2022-06-10
Fix user update to raise exception when the username contains spaces

### LINKED ISSUE #4240: Can update a user's username to contain spaces
## I'm submitting a ...

- [X] bug report
- [ ] other 

- [ ] CDN in a Box
- [ ] Documentation
- [ ] Grove
- [ ] Traffic Control Client
- [ ] Traffic Monitor
- [ ] Traffic Ops
- [ ] Traffic Ops ORT
- [x] Traffic Portal
- [ ] Traffic Router
- [ ] Traffic Stats
- [ ] Traffic Vault
- [ ] unknown

## Current behavior:

When making a `PUT` request to `/user/current` with the following request body:
```
{
    "user": {
        "username": "space   test"
    }
}
```
The response returned is:
```
HTTP/1.1 200 OK
{
    "alerts": [
        {
            "text": "User profile was successfully updated",
            "level": "success"
        }
    ],
    "response": {
        "username": "space    test",
         .
         .
    }
}
```
## Expected / new behavior:

Should not be allowed to update a username that contains spaces. 
Should return:
```
HTTP/1.1 400 Bad Request
{
    "alerts": [
        {
            "text": "Couldn't parse request: username: cannot contain spaces",
            "level": "error"
        }
    ]
}
```

hritvikpatel4:
This PR fixes #4240 by adding checks to raise an exception when the username contains spaces.
<hr/>

- Traffic Ops

Run integration tests

- master

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/lib/go-tc/users.go
+++ b/lib/go-tc/users.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/apache/trafficcontrol/lib/go-rfc"
@@ -487,6 +488,8 @@ func (u *CurrentUserUpdateRequestUser) UnmarshalAndValidate(user *User) error {
 			errs = append(errs, fmt.Errorf("username: %w", err))
 		} else if user.Username == nil || *user.Username == "" {
 			errs = append(errs, errors.New("username: cannot be null or empty string"))
+		} else if strings.Contains(*user.Username, " ") {
+			errs = append(errs, errors.New("username: cannot contain spaces"))
 		}
 	}
 
--- a/traffic_ops/testing/api/v2/user_test.go
+++ b/traffic_ops/testing/api/v2/user_test.go
@@ -17,13 +17,13 @@ package v2
 import (
 	"bytes"
 	"fmt"
-	"github.com/apache/trafficcontrol/lib/go-rfc"
 	"net/http"
 	"net/mail"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/apache/trafficcontrol/lib/go-rfc"
 	"github.com/apache/trafficcontrol/lib/go-tc"
 	"github.com/apache/trafficcontrol/lib/go-util"
 	toclient "github.com/apache/trafficcontrol/traffic_ops/v2-client"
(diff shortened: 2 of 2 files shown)


lib/go-tc/users.go
```diff
@@ -487,6 +488,8 @@ func (u *CurrentUserUpdateRequestUser) UnmarshalAndValidate(user *User) error {
 			errs = append(errs, fmt.Errorf("username: %w", err))
 		} else if user.Username == nil || *user.Username == "" {
 			errs = append(errs, errors.New("username: cannot be null or empty string"))
+		} else if strings.Contains(*user.Username, " ") {
+			errs = append(errs, errors.New("username: cannot contain spaces"))
```
ocket8888, line 492: Since in older versions of Traffic Ops, it was possible to have users with spaces in their usernames, it's a breaking change to the API to no longer allow that. As such, that change cannot be made in APIv2 or v3.

However, I think a better solution than returning an error when the username contains spaces is to simply change the form validation in Traffic Portal to allow spaces, since the API already does. That also allows you to avoid the mess of API versioning.

traffic_ops/testing/api/v2/user_test.go
```diff
@@ -273,6 +273,30 @@ func UserSelfUpdateTest(t *testing.T) {
...
+
+	if len(resp2) < 1 {
+		t.Fatalf("no user returned when requesting user #%d", *user.ID)
+	}
+
+	if resp2[0].Username == nil {
+		t.Errorf("Username missing or null after update")
+	} else if *resp2[0].Username != currentUsername {
+		t.Errorf("Expected Username to still be '%s', but it was '%s'", currentUsername, *resp2[0].Username)
+	}
```
ocket8888, line 299: Tests must not be added only to APIv2; it's actually *more* important that they be in versions 3 and 4, since 2 is close to being removed.

ocket8888 (changes requested):

lib/go-tc/users.go
hritvikpatel4: So if the condition defined in line 491 is true, I would need to change the form validation in the Traffic Portal?

If yes, where can I find the part of form validation in Traffic Portal?

And would I need to make the change for the form validation from the Go code?

traffic_ops/testing/api/v2/user_test.go
hritvikpatel4: Added tests for APIv2, v3 and v4.

ocket8888: I re-ran them and they passed.

lib/go-tc/users.go
hritvikpatel4: > The controller for the user details page view can be found at `traffic_portal/app/src/common/modules/form/user/FormUserController.js`

Alright, I'll look into that file. Thanks!

PR #6899, opened 2022-06-14
TP v2 change Full Name to Name on the user page 

solomonbS:
this pr changes the full name in TP v2 to name on the user page 
<hr/>

- Traffic Portal v2

- Go to current user profile  to verify  that full name is changed to name.

## PR submission checklist
- [ ] This PR has tests 


experimental/traffic-portal/src/app/core/currentuser/currentuser.component.html
```diff
@@ -34,8 +34,8 @@ <h2>User Theme</h2>
 				<input matInput [(ngModel)]="editUser.username" name="username" type="text" required/>
 			</mat-form-field>
 			<mat-form-field style="grid-area: z;">
-				<mat-label>Full Name</mat-label>
-				<input matInput [(ngModel)]="editUser.fullName" name="fullName" type="text" required/>
+				<mat-label>Name</mat-label>
+				<input matInput [(ngModel)]="editUser.fullName" name="Name" type="text" required/>
```
ocket8888, line 38: The `name` attribute of an `input` is a programmatic name, not an accessibility or otherwise human-readable name. It's meant to be the key in a key-value pair submitted to the server in the request that results from the form's submission. It also adds a property to the containing `HTMLFormElement` in the DOM by the same name. So `document.forms[0].Name` with the value you have. if it were `fullName`, then the HTMLFormElement at `document.forms[0]` would actually fulfill the `User` interface. Instead this will make it have one property with a differing name.

For all these reasons, the `name` attribute here should remain `fullName`.

ocket8888 (changes requested):
This should also update the column of the Users page table.

ocket8888:
This also needs to get updated in the users table and the user details pages

ocket8888 (approved):

PR #6900, opened 2022-06-14
Fix Traffic Portal affected server count in Profile change confirmation dialogs

ocket8888:
Due to using a removed query string parameter to filter servers by Profile, there were several places in Traffic Portal that would incorrectly report that a Profile was used by all servers across all CDNs. This fixes that by updating the usage of the API to be consistent with the version of it that Traffic Portal uses.

<hr/>

- Traffic Portal

Make sure existing tests still pass, compare two profiles to see their usage, add and remove Parameters to verify that the confirmation dialogs accurately report the number of servers used by the Profile.

- master

## PR submission checklist
- [ ] This PR has tests


shamrickus (approved):
LGTM

PR #6908, opened 2022-06-17
Fix t3c panic on failed TO request

rob05c:
Fixes a t3c panic when the Traffic Ops /servers request fails. When that happens, the t3c run is going to fail anyway, but the panic causes it to not print the real error, the failed request.

Verify normal t3c run succeeds. Verifying the actual fix would be very difficult; you'd have to simulate a TO that succeeds on every request except /servers.

- `master`

## PR submission checklist
- ~[x] This PR has tests~ no tests, fix requires integration tests because it involves HTTP requests, and would require an integration test framework capable of simulating very specific server failures, which doesn't exist today and would be considerable work to create. 
- ~[x] This PR has a CHANGELOG.md entry~ no changelog, not in a release 

### DIFF
--- a/cache-config/t3cutil/getdatacfg.go
+++ b/cache-config/t3cutil/getdatacfg.go
@@ -794,10 +794,12 @@ func GetConfigData(toClient *toreq.TOClient, disableProxy bool, cacheHostName st
 		return true
 	})
 
-	err := error(nil)
-	toData.ServerParams, err = atscfg.GetServerParameters(toData.Server, combineParams(toData.ServerProfilesParams))
-	if err != nil {
-		errs = append(errs, err)
+	if len(errs) == 0 && toData.Server != nil {
+		err := error(nil)
+		toData.ServerParams, err = atscfg.GetServerParameters(toData.Server, combineParams(toData.ServerProfilesParams))
+		if err != nil {
+			errs = append(errs, err)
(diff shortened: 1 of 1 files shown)


ezelkow1 (approved):

cache-config/t3cutil/getdatacfg.go
```diff
@@ -794,10 +794,12 @@ func GetConfigData(toClient *toreq.TOClient, disableProxy bool, cacheHostName st
 		return true
 	})
 
-	err := error(nil)
-	toData.ServerParams, err = atscfg.GetServerParameters(toData.Server, combineParams(toData.ServerProfilesParams))
-	if err != nil {
-		errs = append(errs, err)
+	if len(errs) == 0 && toData.Server != nil {
```
zrhoffman, line 797: `errs` needs to be empty in order to append to it?

rob05c: No, but if there's an error, `toData.Server` may be nil, which is the panic here. 

The idiomatically-correct thing to do is to stop processing any data if there are any errors, i.e. `len(errs) == 0`.
Theoretically, if `len(errs) == 0` then `toData.Server` should never be nil here; the `toData.Server != nil` is just an additional safety, before dereferencing it.

PR #6918, opened 2022-06-22
Fix TR db file permissions

rawlinp:
By default, some files created by TR in the /opt/traffic_router/db
directory are world-writable. Make these files only writable by the
owner (typically root) when they're created.

<hr/>

- Traffic Router

Run TR, ensure that the files in the `/opt/traffic_router/db` directory are created as writable by the owner/user. Note: for existing files that remain unchanged, the existing file permissions will remain in place, so to test this you may need to delete existing files in the `db` directory or make a change in Traffic Ops (e.g. federations) which would change the resulting file.

all of the supported versions

## PR submission checklist
- [ ] This PR has tests 


zrhoffman (approved):
Verified that newly-created DB files are no longer world-writable. Existing files that were world-writable are not made non-world-writable, however.

PR #6905, opened 2022-06-15
Add t3c metadata file

rob05c:
Adds a t3c metadata file.

Currently, this is just to help operational debugging. No behavior changes, other than writing the file.

But a critical future feature this enables is deleting old files. Currently, t3c never deletes files, because it has no way of knowing what it dropped in the past, versus files in the ATS config directory places manually or by other tools. The metadata file tracks "owned" files, making it possible to safely delete old files. Though this PR doesn't add that.

## Which Traffic Control components are affected by this PR?
- Traffic Control Cache Config (`t3c`, formerly ORT)

## What is the best way to verify this PR?
Run t3c, verify a `t3c-apply-metadata.json` file is placed in the ATS config directory, with some useful information about the run.

## If this is a bugfix, which Traffic Control versions contained the bug?
Not a bug fix.

## PR submission checklist
- [x] This PR has tests 
- ~[x] This PR has documentation~ no docs, the metadata file is not stable or documented 
- ~[x] This PR has a CHANGELOG.md entry~ no changelog, no documented behavior changes 
- [x] This PR **DOES NOT FIX A SERIOUS SECURITY VULNERABILITY** (see [the Apache Software Foundation's security guidelines](https://apache.org/security) for details)

PR #6924, opened 2022-06-23
Add cache config action log

rob05c:
Adds an "action log" to t3c.

It can be difficult to figure out what "actions" t3c actually did, with how spammy the log is. But we need all that log data.
This makes it easy to see changes with e.g. `tail -F syncds.log | grep ACTION`.

It also adds the action data to the metadata file. 

This logs all changes to the local system. For example, fetching data from Traffic Ops isn't an "action," but updating config files and reloading ATS are.

Both are independently useful: the log makes it possible to easily see where an action occurred respective to other log lines, and e.g. to find details about why an action failed in the previous lines. While the metadata file makes it easy to see what actions happened in the latest run, as well as what actions occurred and succeeded or failed in the git config history.

This depends on #6905 - recommend merging that first.

Run t3c, verify log and metadata file contain ACTION data.

Not a bug fix.

## PR submission checklist
- [x] This PR has tests 
- ~[x] This PR has documentation~ no docs, log and metadata formats aren't documented or guaranteed to be consistent 
- ~[x] This PR has a CHANGELOG.md entry~ No changelog, log format isn't documented and metadata file isn't stable enough to be in a public changelog or feature list yet 

### DIFF
--- a/cache-config/t3c-apply/t3c-apply.go
+++ b/cache-config/t3c-apply/t3c-apply.go
@@ -125,13 +125,22 @@ func Main() int {
 
 	metaData.ServerHostName = cfg.CacheHostName
 
+	t3cutil.WriteActionLog(t3cutil.ActionLogActionApplyStart, t3cutil.ActionLogStatusSuccess, metaData)
+
 	if cfg.UseGit == config.UseGitYes {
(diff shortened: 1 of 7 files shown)


traeak (approved):
looks okay and works as expected under testing.

ocket8888 (approved):

PR #6933, opened 2022-06-28
Fix tc-health-client to handle credentials files with special characters.

### LINKED ISSUE #6931: tc-health-client cannot parse credential file when password contains equals sign

- Traffic Control Health Client (t3c-health-client)

## Current behavior:

When the tc-health-client's config file contains a TO credentials file name (`"to-credential-file"`) which contains a `TO_PASSWORD` that includes an equals sign (`=`), the tc-health-client does not parse the entire password to use. It splits on the `=` and only uses the 1st part of the password.

## Expected behavior:

The tc-health-client should be able to handle credentials files with passwords containing `=`.

## Steps to reproduce:

Have a password containing `=` for tc-health-client to use in its `"to-credential-file"`. Observe that it gets 401 when attempting to log in to Traffic Ops.

rob05c:
Currently, the tc-health-client does its own parsing of the credentials file, and doesn't correctly parse variables with special characters like `=` and ` ` in them.

This fixes it to use `sh` to do the parsing, and return the result of declared variables. This will make it handle all special characters, comments, etc exactly like sh/bash, and should prevent any future issues with parsing.

Fixes #6931

Run tc-health-client with a credentials file with ` ` or `=` in the user, password, or URL variables, verify it works correctly.

- 6.1.x (but the Health Client is considered experimental)

## PR submission checklist
- [x] This PR has tests 


jpappa200 (approved):
👍  LGTM

PR #6954, opened 2022-07-11
Refactor Static DNS Entries Tests

ericholguin:
This PR refactors **v4/staticdnsentries_test.go**, **v3/staticdnsentries_test.go**, in order to make tests more granular, providing transparency on what is being tested and reducing the need to dig through test code to confirm what was tested. In addition this new test structure should also make it easier for new tests to be added as well as reduce redundant code.

- Adds data driven testing
    - Descriptive test names
- Sub tests for organization
    - Use GoLangs test runner
- Use assert functionality
    - Streamlines code
    - Reduces nested conditionals
- Adds expectation functions
     - Reusable expectations

<hr/>

- Traffic Ops Tests

Run TO Integration Tests

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_ops/testing/api/v4/cdn_locks_test.go
+++ b/traffic_ops/testing/api/v4/cdn_locks_test.go
@@ -29,7 +29,7 @@ import (
 )
 
 func TestCDNLocks(t *testing.T) {
-	WithObjs(t, []TCObj{Types, CacheGroups, CDNs, Parameters, Profiles, Statuses, Divisions, Regions, PhysLocations, Servers, ServiceCategories, Topologies, Tenants, Roles, Users, DeliveryServices, CDNLocks}, func() {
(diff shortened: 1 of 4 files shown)


traffic_ops/testing/api/v4/cdn_locks_test.go
```diff
@@ -487,6 +555,33 @@ func TestCDNLocks(t *testing.T) {
...
+					case "STATIC DNS ENTRIES PUT":
+						{
+							t.Run(name, func(t *testing.T) {
+								alerts, reqInf, err := testCase.ClientSession.UpdateStaticDNSEntry(testCase.EndpointId(), staticDNSEntry, testCase.RequestOpts)
+								for _, check := range testCase.Expectations {
+									check(t, reqInf, nil, alerts, err)
+								}
+							})
+						}
+					case "STATIC DNS ENTRIES DELETE":
```
rawlinp, line 576: nit: would it make sense to include `default` switch cases that fail the test? That way we are less prone to typos sneaking in and causing the tests to not actually be run.

rawlinp (approved):

PR #6955, opened 2022-07-12
TP Use Correct Name in Role Put

### LINKED ISSUE #6925: Cannot update role name in Traffic Portal

- Traffic Portal

## Current behavior:

When attempting to change a role's name in Traffic Portal, it fails with "no such role".
Screen Shot 2022-06-24 at 8 56 04 AM
Screen Shot 2022-06-24 at 8 56 32 AM

## Expected behavior:

Traffic Portal should pass the _original_ role name in the request query parameters instead of the _new_ name, which would allow a user to change the role's name.

## Steps to reproduce:

Try to change the name of a role in Traffic Portal. This affects at least the `master` branch.

shamrickus:
This PR fixes #6925
<hr/>

- Traffic Portal

Verify that a user is able to update a role through TP.

- master

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/traffic_portal/app/src/common/api/RoleService.js
+++ b/traffic_portal/app/src/common/api/RoleService.js
@@ -42,8 +42,8 @@ var RoleService = function($http, messageModel, ENV) {
         );
     };
 
-    this.updateRole = function(role) {
-        return $http.put(ENV.api.unstable + "roles", role, {params: {name: role.name}}).then(
+    this.updateRole = function(role, oldRole) {
+        return $http.put(ENV.api.unstable + "roles", role, {params: {name: oldRole}}).then(
             function(result) {
                 return result.data;
             },
--- a/traffic_portal/app/src/common/modules/form/role/edit/FormEditRoleController.js
+++ b/traffic_portal/app/src/common/modules/form/role/edit/FormEditRoleController.js
@@ -31,7 +31,7 @@ var FormEditRoleController = function(roles, $scope, $controller, $uibModal, $an
 	};
 
 	var save = function(role) {
-		roleService.updateRole(role).
+		roleService.updateRole(role, $scope.roleName).
 			then(function(result) {
 				$scope.roleName = angular.copy(role.name);
 				messageModel.setMessages(result.alerts, false);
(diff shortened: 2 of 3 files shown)


ericholguin:
I was able to update a role's description, but a role's name cannot be updated. Is this normal?

ocket8888:
> I was able to update a role's description, but a role's name cannot be updated, you will get the TP alert `role was updated.` displayed but the name remains the same. Is this normal?

Yes and no. It's normal in that the endpoint as-is cannot change Role names. But that is itself a bug - but it's a bug with TO, and that's not the bug this PR claims to fix.

ericholguin:
> Yes and no. It's normal in that the endpoint as-is cannot change Role names. But that is itself a bug - but it's a bug with TO, and that's not the bug this PR claims to fix.

I see, so this fixes the TP issue. Is there an issue for TO. Looking at the code quickly the update query looks off

https://github.com/apache/trafficcontrol/blob/235256986342c411c515554d871a2269b0169d5d/traffic_ops/traffic_ops_golang/role/roles.go#L69-L74

mitchell852:
is it possible to just fix both with this PR? TP and TO?

ocket8888:
> is it possible to just fix both with this PR? TP and TO?

Yes it's certainly possible, but it's not strictly necessary. This PR fixes the linked bug, so there's no reason it can't be merged.

Also, based on that query, it's possible that we meant to make Role names immutable but simply forgot to document it and make the TP changes. I'd want to `git blame` it and ask the person who modified the query.

ericholguin (approved):
TP works as it should.

ocket8888:
re-running CiaB action

ericholguin:
> > I think names have been mutable, per this legacy query that exists above the new one
> 
> This makes me more sure the new behavior is intentional, otherwise the existing query language would've been re-used to a greater extent. I could still be wrong, but we'd need to hunt down the original proposal/PR to be sure. If that was the author's intent but was never communicated - or we just don't like that behavior -, we could reverse the decision even if it was intentional.

@srijeet0406

srijeet0406:
I _think_ the intention was to keep the role names immutable, but I'm totally on board with reversing that. As long as the `admin` role is immutable, which I believe it is, we should be fine.

ericholguin:
> I _think_ the intention was to keep the role names immutable, but I'm totally on board with reversing that. As long as the `admin` role is immutable, which I believe it is, we should be fine.

Thanks for clearing that up @srijeet0406 . I believe in the issue that @rawlinp created he was trying to change the name of a user. So this is probably something we would want. I don't see any harm in having mutable role names, but I may be wrong. Also this change would match our docs, which currently don't say they are immutable. If no one is opposed I can create the PR.

mitchell852:
> I _think_ the intention was to keep the role names immutable, but I'm totally on board with reversing that. As long as the `admin` role is immutable, which I believe it is, we should be fine.

that makes sense to me. admin name immutable, others mutable. if not, puts extra burden on operators of the system.

PR #6959, opened 2022-07-13
Make role name mutable

ericholguin:
Related: #6955 #6925

This PR:
- Makes a role's name mutable. 
- Fixes internal server error's that were occurring when a user would create or update a role with an already existing name. 
- Matches TP validation by not allowing a user to create a role with spaces
- Increases test coverage for roles to match these changes.

<hr/>

- Traffic Ops

Run TO Integration Tests

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/lib/go-tc/roles.go
+++ b/lib/go-tc/roles.go
@@ -59,8 +59,9 @@ type RoleV40 struct {
 
 // Validate will validate and make sure all that the fields in the supplied RoleV4 struct are semantically correct.
 func (role RoleV4) Validate() error {
+	noSpaces := validation.NewStringRule(tovalidate.NoSpaces, "cannot contain spaces")
 	errs := validation.Errors{
-		"name":        validation.Validate(role.Name, validation.Required),
+		"name":        validation.Validate(role.Name, validation.Required, noSpaces),
 		"description": validation.Validate(role.Description, validation.Required),
 	}
 	return util.JoinErrs(tovalidate.ToErrors(errs))
--- a/traffic_ops/testing/api/v4/roles_test.go
+++ b/traffic_ops/testing/api/v4/roles_test.go
@@ -61,12 +61,60 @@ func TestRoles(t *testing.T) {
 					Expectations:  utils.CkRequest(utils.NoError(), utils.HasStatus(http.StatusOK), validateRoleDescSort()),
 				},
 			},
+			"POST": {
+				"BAD REQUEST when NAME has SPACES": {
+					ClientSession: TOSession,
+					RequestBody: map[string]interface{}{
+						"name":        "role with spaces",
+						"description": "description",
+						"permissions": []string{
+							"all-read",
+							"all-write",
+						},
+					},
+					Expectations: utils.CkRequest(utils.HasError(), utils.HasStatus(http.StatusBadRequest)),
+				},
+				"BAD REQUEST when MISSING NAME": {
+					ClientSession: TOSession,
+					RequestBody: map[string]interface{}{
+						"description": "missing name",
+						"permissions": []string{
+							"all-read",
+							"all-write",
+						},
+					},
+					Expectations: utils.CkRequest(utils.HasError(), utils.HasStatus(http.StatusBadRequest)),
+				},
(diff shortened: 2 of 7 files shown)


lib/go-tc/roles.go
```diff
@@ -59,8 +59,9 @@ type RoleV40 struct {
 
 // Validate will validate and make sure all that the fields in the supplied RoleV4 struct are semantically correct.
 func (role RoleV4) Validate() error {
+	noSpaces := validation.NewStringRule(tovalidate.NoSpaces, "cannot contain spaces")
 	errs := validation.Errors{
-		"name":        validation.Validate(role.Name, validation.Required),
+		"name":        validation.Validate(role.Name, validation.Required, noSpaces),
```
ocket8888, line 64: Restricting names to not contain spaces is a breaking API change, and I don't think it's necessary to fix this bug.

ocket8888 (changes requested):

lib/go-tc/roles.go
ericholguin: Should that validation also be removed from TP?

ocket8888:
Since it affects 3.x I suppose so; intention doesn't matter because that version was supposed to be frozen.

traffic_ops/testing/api/v4/roles_test.go
```diff
@@ -239,14 +338,20 @@ func CreateTestRoles(t *testing.T) {
...
+		assert.NoError(t, err, "Expected no error while deleting role %s, but got %v", role.Name, err)
 		// Retrieve the Role to see if it got deleted
 		opts := client.NewRequestOptions()
-		opts.QueryParameters.Set("name", r.Name)
+		opts.QueryParameters.Set("name", role.Name)
 		getRole, _, err := TOSession.GetRoles(opts)
-		assert.NoError(t, err, "Error getting Role '%s' after deletion: %v - alerts: %+v", r.Name, err, getRole.Alerts)
-		assert.Equal(t, 0, len(getRole.Response), "Expected Role '%s' to be deleted, but it was found in Traffic Ops", r.Name)
+		assert.NoError(t, err, "Error getting Role '%s' after deletion: %v - alerts: %+v", role.Name, err, getRole.Alerts)
+		assert.Equal(t, 0, len(getRole.Response), "Expected Role '%s' to be deleted, but it was found in Traffic Ops", role.Name)
```
ocket8888, line 355: same comments apply here.

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -224,6 +217,13 @@ func (role *TORole) Read(h http.Header, useIMS bool) ([]interface{}, error, erro
 }
 
 func (role *TORole) Update(h http.Header) (error, error, int) {
+
+	if ok, err := dbhelpers.RoleExists(role.ReqInfo.Tx.Tx, *role.ID); err != nil {
+		return nil, errors.New("verifying Role exists: " + err.Error()), http.StatusInternalServerError
```
ocket8888, line 222: Creating an error from an underlying error should use `fmt.Errorf`

traffic_ops/traffic_ops_golang/role/roles.go
```diff
@@ -259,6 +259,13 @@ func (role *TORole) Update(h http.Header) (error, error, int) {
 }
 
 func (role *TORole) Delete() (error, error, int) {
+
+	if ok, err := dbhelpers.RoleExists(role.ReqInfo.Tx.Tx, *role.ID); err != nil {
+		return nil, errors.New("verifying Role exists: " + err.Error()), http.StatusInternalServerError
```
ocket8888, line 264: same as above

traffic_ops/testing/api/v4/roles_test.go
ocket8888: Ohhh. Good. I should really know that, since I'm quite sure I did that, and specifically so that you wouldn't have to do those checks I just asked for.

ocket8888 (approved):
I can change a Role's name now. This is good to go, pending the TP integration tests.

PR #6962, opened 2022-07-13
t3c to look for t3c-generate and other t3c- apps in executable directory

traeak:
Current t3c internall hard codes t3c-generate to /usr/bin/t3c-generate and does a path lookup when calling other t3c-* utilities.

This PR uses go 1.18+ os.Executable() to obtain the executable install directory.

<hr/>

Ensure t3c still runs as installed.

Make the t3c package, install t3c, copy out the binaries to another directory not in path, run t3c via path and ensure it works.

## PR submission checklist
- [ ] This PR has tests - already captured with normal t3c runs.
- [ ] This PR has documentation - minor under the hood change.


cache-config/t3cutil/t3cutil.go
```diff
@@ -228,3 +232,14 @@ func UserAgentStr(appName string, versionNum string, gitRevision string) string
 	}
 	return appName + "/" + versionNum + ".." + gitRevision
 }
+
+// InstallDirectory returns the location of the executable directory.
+// If error, returns "/usr/bin" as default.
+func InstallDirectory() string {
+	installdir, err := osext.ExecutableFolder()
```
rob05c, line 239: ATC is on Go >1.18 which has `os.Executable()`. Can you change this to `os.Executable`, and not use/vendor `osext`?

cache-config/t3cutil/t3cutil.go
```diff
@@ -228,3 +232,14 @@ func UserAgentStr(appName string, versionNum string, gitRevision string) string
...
+	if err != nil {
+		installdir = "/usr/bin"
```
rob05c, line 241: Mind putting this in a global const? Just so it's easier to see, change, etc.

cache-config/t3c-apply/torequest/cmd.go
```diff
@@ -85,7 +86,8 @@ func generate(cfg config.Cfg) ([]t3cutil.ATSConfigFile, error) {
 	args = append(args, "--disable-parent-config-comments="+strconv.FormatBool(cfg.DisableParentConfigComments))
 	args = append(args, "--use-strategies="+cfg.UseStrategies.String())
 
-	generatedFiles, stdErr, code := t3cutil.DoInput(configData, config.GenerateCmd, args...)
+	genpath := filepath.Join(t3cutil.InstallDirectory(), config.GenerateCmd)
```
rob05c, line 89: So, this is used in a bunch of places, basically every sub-app call.

If you changed them all to use the `t3c` wrapper, e.g. `t3c generate` instead of `t3c-generate`, you could avoid all the duplication, put this in a global var, `var t3cPath = filepath.Join(t3cutil.InstallDir() + "t3c")`, and then all the args would just start with the sub-app, e.g.
```
args := []string{"preprocess"}
args += ...
cmd := exec.Command(t3cPath, args...)
```

rob05c (changes requested):
Looks good, big +1 on this.

There's no perfect solution, in terms of security and usability. But this seems to be the most common compromise for Linux/UNIX apps. It lets users run it in any install dir or dev env, while still being reasonably secure, since if an attacker has compromised the exe dir they could almost always have compromised the existing app as well.

And this fixes the current issue of running in a dev environment being a huge pain.

Most of my requests are minor, only big one is using the standard library to avoid vendoring `osext`

cache-config/t3c-apply/torequest/cmd.go
traeak: did as suggested within t3c-apply/torequest/cmd.go

cache-config/t3c-apply/torequest/cmd.go
```diff
@@ -43,18 +44,31 @@ import (
 	"github.com/apache/trafficcontrol/lib/go-tc"
 )
 
+const (
+	t3cgen       = `t3c generate`
+	t3cupd       = `t3c update`
+	t3cdiff      = `t3c diff`
+	t3cchkrefs   = `t3c check refs`
+	t3cchkreload = `t3c check relaod`
```
rob05c, line 52: Typo "reload"

cache-config/t3c-apply/torequest/cmd.go
```diff
@@ -531,17 +551,18 @@ func requestConfig(cfg config.Cfg) ([]byte, error) {
...
 	if code != 0 {
-		logSubAppErr(`t3c-request stdout`, stdOut)
-		logSubAppErr(`t3c-request stderr`, stdErr)
-		return nil, fmt.Errorf("t3c-request returned non-zero exit code %v, see log for details", code)
+		logSubAppErr(`t3c stdout`, stdOut)
+		logSubAppErr(`t3c stderr`, stdErr)
+		return nil, fmt.Errorf("t3c returned non-zero exit code %v, see log for details", code)
 	}
-	logSubApp(`t3c-request`, stdErr)
+	logSubApp(`t3c`, stdErr)
```
rob05c, line 565: Looks like this should be `logSubApp(t3creq, stdErr)`

rob05c (approved):
LGTM. Code looks good, build RPM, installed, ran, generated files look correct.

Issue #6963, opened 2022-07-14
`POST /api/3.1/servers` cannot set `updPending`


zrhoffman:
## This Bug Report affects these Traffic Control components:

- Traffic Ops

## Current behavior:

In Traffic Ops API v3, you used to be able to include
```json
  "revalPending": false,
```
or
```json
  "updPending": true,
```

in the body of `POST /servers` or `PUT /servers` to queue updates. As of #6569, however, the `server.reval_pending` and `server.upd_pending` fields do not exist, and including those fields in the `POST` or `PUT` body does not have any effect.

Since this is undocumented behavior, it is not necessarily a bug. That said, for anyone depending on that behavior in stable TO API version 3, this is a breaking change.

ocket8888:
> ... this is undocumented behavior, ...

[The "Request Structure" section of the `PUT /servers/{{ID}}` documentation](https://traffic-control-cdn.readthedocs.io/en/latest/api/v3/servers_id.html#request-structure) lists `updPending` as a valid payload property, making no mention of it being immutable (because it isn't supposed to be), and in fact [the documentation for `/servers/{{ID}}/queue_update`](https://traffic-control-cdn.readthedocs.io/en/latest/api/v3/servers_id_queue_update.html#servers-id-queue-update) recommend using that method instead. So I'd say it's far from undocumented behavior.

rawlinp:
I don't necessarily agree with this note in the docs:
> In the vast majority of cases, it is advisable that the PUT method of the [servers/{{ID}}](https://traffic-control-cdn.readthedocs.io/en/latest/api/v3/servers_id.html#to-api-v3-servers-id) endpoint be used instead.

IMO queueing updates should be its own separate action -- being able to create a server in TO with updates already queued does not really buy us much. And for updates, it makes sense to queue updates separately without having to include all the unrelated server fields, because it's way more common to queue updates on a server than it is to actually change fields of a server.

So, while this is technically a regression, I actually think it's better this way. Just my 2 cents.

ocket8888:
I made that note a long, long time ago - it's much easier when you're learning to use an API to remember how to use one endpoint than it is to remember what second endpoint you need to use to modify some specific field of responses from that first endpoint. That's why the note is there, but that's probably not really a good enough reason for it to exist. Regardless, though, it's there.

zrhoffman:
[...]

PR #6965, opened 2022-07-14
Ansible Dataset Loader: Queue Caches after performing Snapshots

zrhoffman:
As noted in #6963, as of #6569, `POST /api/3.1/servers` and `PUT /api/3.1/servers` can no longer change `updPending`. In the Ansible Roles, this breaks deploying Grove, as Grove will not generate `/etc/grove/remap.json` if updates are not pending for that server.

This PR queues updates on Caches in all CDNs after performing snapshots at the end of the Dataset Loader.
<hr/>

- Automation - Ansible Roles

Test Dataset Loader, verify updates are queued for all Caches

## PR submission checklist
- [ ] This PR has tests 

### DIFF
--- a/infrastructure/ansible/roles/dataset_loader/tasks/dataset_loader.yml
+++ b/infrastructure/ansible/roles/dataset_loader/tasks/dataset_loader.yml
@@ -1043,3 +1043,18 @@
   delay: 10
   vars:
     error_query: "alerts[?level=='error'].text[?!contains(@,'already exists')]"
+
+- name: Queue Caches in all CDNs
+  uri:
+    url: "{{ dl_to_url }}/api/{{ dl_to_api_version }}/cdns/{{ item.id }}/queue_update"
+    method: POST
+    body:
+      action: "queue"
+  with_items: "{{ get_all_cdns.json.response[1:] }}"
+  register: snapshot_out
+  failed_when: (snapshot_out.status == 400 and snapshot_out.json | to_json | from_json | json_query(error_query) | length != 0) or (snapshot_out.status > 400 and snapshot_out.status < 600)
+  no_log: true
+  retries: 10
+  delay: 10
+  vars:
+    error_query: "alerts[?level=='error'].text[?!contains(@,'action must be')]"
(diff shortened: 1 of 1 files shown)


infrastructure/ansible/roles/dataset_loader/tasks/dataset_loader.yml
```diff
@@ -1043,3 +1043,18 @@
...
   vars:
     error_query: "alerts[?level=='error'].text[?!contains(@,'already exists')]"
+
+- name: Queue Caches in all CDNs
+  uri:
+    url: "{{ dl_to_url }}/api/{{ dl_to_api_version }}/cdns/{{ item.id }}/queue_update"
+    method: POST
+    body:
+      action: "queue"
+  with_items: "{{ get_all_cdns.json.response[1:] }}"
```
jhg03a, line 1053: This should use a different method like json_query to exclude the ALL CDN because the order of the json isn't guaranteed.

jhg03a (changes requested):

infrastructure/ansible/roles/dataset_loader/tasks/dataset_loader.yml
zrhoffman: The previous task, *Snapshot all CDNs*, uses this method to exclude the *ALL* CDN
https://github.com/apache/trafficcontrol/blob/a0fbd63083140e64d35d886e0cc87cc8b542aaa6/infrastructure/ansible/roles/dataset_loader/tasks/dataset_loader.yml#L1034-L1038

Maybe someone else can add `json_query` to both tasks in the future.

zrhoffman: I don't think that both tasks should use different methods for filtering out the *ALL* CDN, though, and changing the *Snapshot all CDNs* task is outside the scope of this PR.

zrhoffman: > because the order of the json isn't guaranteed.

Also, the sort order is always the same
https://github.com/apache/trafficcontrol/blob/19f3b2eafde573546221e767c77f283994fb2fa7/traffic_ops/traffic_ops_golang/cdn/cdns.go#L140

jhg03a: Ok, so even if json itself isn't we're coercing it to be.  It's still a bug laying in wait should you name a CDN something alphabetically before ALL.  If we want to merge this and open more issues to fix that separately that's fine.

jhg03a (approved):
Approved on the condition that a followup issue be created and worked to address proper queue/snap of CDNs when the CDN name begins alphabetically before `ALL`

PR #6967, opened 2022-07-15
TPv2 Add Change Logs Table

shamrickus:
This PR adds a change-logs table to TPv2. It adds functionality to the generic table to allow for adding more header buttons. It also enables the right click context menu on the users table to view change-logs for a given user.
<hr/>

- Traffic Portal v2

Verify the table works, can see a users change logs (currently just does this with a search param, should use param to filter the column but that is not implemented in the tpv2 version) and that the tests pass.

## PR submission checklist
- [x] This PR has tests 


experimental/traffic-portal/src/app/models/change-logs.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Represents the ChangeLog api object
+ */
+export interface ChangeLog {
+	id: number;
+	lastUpdated: Date;
+	level: string;
+	message: string;
+	ticketNum?: number | null;
+}
```
ocket8888, line 24: This definition is provided by the `trafficops-types` library; any reason you made your own?

experimental/traffic-portal/src/app/core/change-logs/last-days/last-days.component.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+*/
+import { ComponentFixture, TestBed } from "@angular/core/testing";
+import { MAT_DIALOG_DATA, MatDialogRef } from "@angular/material/dialog";
+
+import { LastDaysComponent } from "./last-days.component";
+
+describe("LastDaysComponent", () => {
+	let component: LastDaysComponent;
+	let fixture: ComponentFixture<LastDaysComponent>;
+	let mockMatDialog: jasmine.SpyObj<MatDialogRef<number>>;
```
ocket8888, line 22: why bother making a spy you're not using? Did you forget to write a test that you had planned to use this?

experimental/traffic-portal/src/app/api/change-logs.service.ts
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	/**
+	 * Calls api logs endpoint
+	 *
+	 * @param params Request paramteres to add
+	 * @returns Change logs
+	 */
+	public async getChangeLogs(params: Record<string, string>): Promise<Array<ChangeLog>> {
+		return this.get<[ChangeLog]>("logs", undefined, params).toPromise();
```
ocket8888, line 38: This is returning a tuple of one ChangeLog item, not an array. It's working because `[T]` is assignable to `T[]`, but it'd be more accurate if they were the same type.

ocket8888 (changes requested):

experimental/traffic-portal/src/app/models/change-logs.ts
shamrickus: Because the version we were using didn't have it.

ocket8888: It does, you probably just didn't realize it. It's called `Log`, not `ChangeLog`, named after the API endpoint - which is `/logs` not e.g. `/change_logs`

experimental/traffic-portal/src/app/api/testing/index.ts
```diff
@@ -17,7 +17,7 @@ import { NgModule } from "@angular/core";
 
 import {
 	CacheGroupService,
-	CDNService,
+	CDNService, ChangeLogsService,
```
ocket8888, line 20: nit: this should be on the next line

experimental/traffic-portal/src/app/core/users/users.component.ts
```diff
@@ -229,8 +229,8 @@ export class UsersComponent implements OnInit {
 			newTab: true
 		},
 		{
-			action: "viewChangelogs",
-			disabled: (): true =>true,
+			disabled: (us: GetResponseUser | Array<GetResponseUser>): boolean => Array.isArray(us),
```
ocket8888, line 232: if `multiRow` isn't `true`, the action is implicitly disabled when multiple rows are selected, so you shouldn't need to define this. If you do, that's a bug in the generic table component.

experimental/traffic-portal/src/app/utils/time.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @returns Formatted string in the form of 'N X Ago' where X is anything between Seconds and Years
+ */
+export function relativeTimeString(delta: number): string {
+	const SEC = 1000;
+	const MIN = SEC * 60;
+	const HOUR = MIN * 60;
+	const DAY = HOUR * 24;
+	const YEAR = 365 * DAY;
+	const MONTH = YEAR / 12;
+	const WEEK = MONTH / 4;
```
ocket8888, line 28: This probably doesn't really have any performance impact at all, but I'd define these outside the function so they aren't recalculated every time it's called. Do it or don't, just wanted to point it out.

ocket8888:
When I click on "View user changelogs" from the context menu on a user, I see changelogs for all users, not just that user

ocket8888 (approved):

PR #6971, opened 2022-07-18
Minor GHA changes

ericholguin:
Removes the TO testing and client path from TP integration GHA
Updates the action names to match the other actions, making them shorter.

<hr/>

- GitHub Actions

## PR submission checklist
- [ ] This PR has tests 

### DIFF
--- a/.github/workflows/cache-config-tests.yml
+++ b/.github/workflows/cache-config-tests.yml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: Traffic Control Cache Config integration tests
+name: Cache Config Integration Tests
 
 env:
   ATS_VERSION: 8.1.x
--- a/.github/workflows/health-client-tests.yml
+++ b/.github/workflows/health-client-tests.yml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: Traffic Control Health Client integration tests
+name: Health Client Integration Tests
 
 env:
   ATS_VERSION: 8.1.x
--- a/.github/workflows/traffic-ops.yml
+++ b/.github/workflows/traffic-ops.yml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: Traffic Ops Go client/API integration tests
+name: TO Integration Tests
 
 env:
   # alpine:3.13
--- a/.github/workflows/tp.integration.tests.yml
+++ b/.github/workflows/tp.integration.tests.yml
@@ -30,8 +30,6 @@ on:
       - .github/workflows/tp.integration.tests.yml
       - GO_VERSION
       - infrastructure/cdn-in-a-box/optional/traffic_vault/**
-      - traffic_ops/*client/**.go
-      - traffic_ops/testing/api/**.go
       - traffic_ops/traffic_ops_golang/**.go
       - traffic_portal/**
   create:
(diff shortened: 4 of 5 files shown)


.github/workflows/cache-config-tests.yml
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: Traffic Control Cache Config integration tests
+name: Cache Config Integration Tests
```
ocket8888, line 18: "Traffic Control Cache Config" is the name of the app being tested, so this was actually appropriate. If the name length bothers you, you could use the abbreviated name "t3c".

zrhoffman, line 18: Since the acronym is T3C (stylized `t3c`), *Traffic Control* should stay in the title IMO

ericholguin: True. Traffic Control Cache Config Integration Tests or T3C Integration Tests?

.github/workflows/health-client-tests.yml
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: Traffic Control Health Client integration tests
+name: Health Client Integration Tests
```
ocket8888, line 18: "Traffic Control Health Client" is the name of the app under test in this workflow, so that was actually appropriate - besides casing. The name can be slightly abbreviated to match the executable name: `tc-health-client`.

.github/workflows/traffic-ops.yml
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: Traffic Ops Go client/API integration tests
+name: TO Integration Tests
```
ocket8888, line 18: The cross-app integration being tested here is TO's API with its Go client. TO is basically just an API, so saying API may be redundant, and it can be abbreviated from Traffic Ops to TO, but imo this should still mention the Go client.

ericholguin: Is Go necessary? Do we have another client?

ocket8888: Yes, we have clients for the TO API in two other languages: Java and Python.

ocket8888 (approved):

.github/workflows/traffic-ops.yml
```diff
@@ -14,8 +14,8 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-
-name: Traffic Ops Go client/API integration tests
+  Traffic Ops Go client/API integration tests
```
zrhoffman, line 17: The name is on 2 lines

ericholguin: Oops good catch. Thanks

zrhoffman (approved):

zrhoffman:
The failure:

```go
--- FAIL: TestServerServerCapabilities (6.10s)
    serverservercapability_test.go:612: expected GET Server Server Capabilities with limit = 1 to return first result
    serverservercapability_test.go:623: expected GET server server capabilities with limit = 1, offset = 1 to return second result
    serverservercapability_test.go:634: expected GET Server Server Capabilities with limit = 1, page = 2 to return second result
FAIL
exit status 1
FAIL	github.com/apache/trafficcontrol/traffic_ops/testing/api/v4	272.169s
```

so, merging anyway

PR #6974, opened 2022-07-20
Fix Pkg Build Order to Mirror docker-compose.yml

shamrickus:
This changes `pkg` to (by default) build services in the order listed in `infrastructure/docker/build/docker-compose.yml` as opposed to alphabetical order. 

This issue was originally fixed in #5368, but was reintroduced in #6274 as `docker-compose config --services` lists services in alphabetical order.
<hr/>

- Build

Run `./pkg source` and then extract the produced source and run `./pkg -v` again. Weasel should not fail from build artifacts such as [these](https://lists.apache.org/thread/okndcgcxn3qr5nkqg1tokbf0w6p38h9n)

 - 6.1.0
 - 7.0.0 RC0
 - master

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/pkg
+++ b/pkg
@@ -155,6 +155,8 @@ while getopts h78abdf:lopqvsSL opt; do
 	esac
 done
 
+# Weasel must be built first otherwise it will fail from build artifacts/cache in an environment that doesn't have git
+# e.g. the environment produced by `./pkg source`.
 PROJECTS="weasel"$'\n'"$("${COMPOSECMD[@]}" -f $COMPOSE_FILE config --services | sed '/^weasel$/d')"
 HELP_TEXT="$(cat <<HELP_TEXT
 Usage: $SELF [options] [projects]
--- a/pkg
+++ b/pkg
@@ -155,7 +155,7 @@ while getopts h78abdf:lopqvsSL opt; do
 	esac
 done
 
-PROJECTS="$("${COMPOSECMD[@]}" -f $COMPOSE_FILE config --services)"
+PROJECTS="$(cat infrastructure/docker/build/docker-compose.yml | grep -E '^\s\s[a-zA-Z_-]+:$' | sed 's/[: ]//g')"
 HELP_TEXT="$(cat <<HELP_TEXT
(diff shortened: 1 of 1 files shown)


zrhoffman:
> but was reintroduced in #6274 as `docker-compose config --services` lists services in alphabetical order.

Upon closer inspection, `pkg` has used

```shell
docker-compose config --services
```

to list services since its creation. The Go implementation of `docker-compose config --services` alphabetizes the services, while the (now-unsupported) Python version,

```shell
python -m compose config --services
```
does not.

pkg
```diff
@@ -155,7 +155,7 @@ while getopts h78abdf:lopqvsSL opt; do
 	esac
 done
 
-PROJECTS="$("${COMPOSECMD[@]}" -f $COMPOSE_FILE config --services)"
+PROJECTS="$(cat infrastructure/docker/build/docker-compose.yml | grep -E '^\s\s[a-zA-Z_-]+:$' | sed 's/[: ]//g')"
```
zrhoffman, line 158: Parsing YAML with `grep` and `sed` does not seem so maintainable. What are your thoughts about rearranging the output of `docker-compose config --services` to have `weasel` first?

zrhoffman (changes requested):

pkg
shamrickus: That sounds like a better solution to me

pkg
```diff
@@ -155,7 +155,7 @@ while getopts h78abdf:lopqvsSL opt; do
 	esac
 done
 
-PROJECTS="$(cat infrastructure/docker/build/docker-compose.yml | grep -E '^\s\s[a-zA-Z_-]+:$' | sed 's/[: ]//g')"
+PROJECTS="weasel"$'\n'"$("${COMPOSECMD[@]}" -f $COMPOSE_FILE config --services | sed '/^weasel$/d')"
```
zrhoffman, line 158: This looks good, but would you mind including a comment that explains why it's like that?

zrhoffman (approved):
Looks good!

PR #6978, opened 2022-07-21
Add t3c support for __CACHEKEY_DIRECTIVE__ in raw remap text

traeak:
This PR allows the cachekey.so plugin and args to be moved into the raw remap text, similar to the __RANGE_DIRECTIVE__ hack.

This allows custom things like lua scripts to modify the uri path, args, etc before the cachekey is created and set.

<hr/>

- Documentation

## PR submission checklist
- [x] This PR has tests 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -598,6 +598,7 @@ will be returned indicating that overlap exists.
 - Changed Traffic Portal to use the more performant and powerful ag-grid for the delivery service request (DSR) table.
 - Traffic Ops: removed change log entry created during server update/revalidation unqueue
 - Updated CDN in a Box to CentOS 8 and added `RHEL_VERSION` Docker build arg so CDN in a Box can be built for CentOS 7, if desired
+- Added Delivery Service Raw Remap `__CACHEKEY_DIRECTIVE__` directive to allow inserting the cachekey directive into the Raw Remap text. This allows Raw Remaps which manipulate the cachekey.
 
 ### Deprecated
 - Deprecated the non-nullable `DeliveryService` Go struct and other structs that use it. `DeliveryServiceNullable` structs should be used instead.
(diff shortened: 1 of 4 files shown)


ocket8888 (changes requested):
This should be added to [the list of strings with special meaning to `t3c`](https://traffic-control-cdn.readthedocs.io/en/latest/admin/traffic_server.html?highlight=special%20meaning#strings-with-special-meaning-to-t3c).

rob05c:
@ocket8888 This does add docs for the particular DS field it affects. Is that sufficient?

Or if you want, I guess we could add some kind of line to the `the list of strings with special meaning to t3c` doc, "see individual Delivery Service and Server fields for replacements unique to data in those fields" or something?

rob05c:
Also removing the `low_impact` tag, if that's ok. This isn't a lot of code, but it's a pretty big impact, it makes it possible for tenants to do significant things that aren't possible without this.

ocket8888:
>  It isn't replaced in any other locations, like the doc says and like all those other directives are.

That's not strictly true. That section is also appropriate for replacements restricted to a single file, for example the `##OVERRIDE##` description reads:

> This string is only valid in the content of files named “remap.config”.

But looking at the code this isn't just restricted to a single configuration file it's, as you say:

> ... only replaced inside the DS Raw Remap Text...

which I didn't realize right away. So

> This does add docs for the particular DS field it affects. Is that sufficient?

Yes.

> Or if you want, I guess we could add some kind of line to the `the list of strings with special meaning to t3c` doc, "see individual Delivery Service and Server fields for replacements unique to data in those fields" or something?

No, I don't think that's necessary.

> Also removing the low_impact tag, if that's ok.

The point is just that it should have *some* impact level tag. I just took a guess at what that would be but you should absolutely replace it with whichever is more appropriate.

ocket8888 (approved):
Verified that the `@plugin=cachekey.so` part goes wherever the `__CACHEKEY_DIRECTIVE__` is found.

My only comment would be that `RemapConfigCachekeyDirective` could use a GoDoc comment, but most of the structures and functions in the `atscfg` package are missing those (most of the exported symbols under `lib/` for that matter) and it won't be hard to add it the next time I rebase #6099, which can be merged if/whenever we decide GoDocs are important.

rob05c (approved):
LGTM.

Not thrilled with these hacks, but it's something we need, certain configs we need are impossible without this, and a generic non-hacky solution will be drastically more work (and isn't immediately obvious what it would even look like).

PR #6991, opened 2022-07-28
t3c log run mode and log update flag changes

jpappa200:
Right now t3c-apply does not log what mode was run (syncds|badass|revalidate) this will log the run mode. At the end of the t3c-apply run TO is updated to clear the update/revalidate flags. this change logs the current value, what it was updated to, and how log it took to update.

<hr/>

Build RPM and install on a cache, queue updates and run t3c-apply in one of the 3 modes listed above. you should see the run mode logged at the beginning and the TO update change at the end of the run 

## PR submission checklist
- [ ] This PR has tests 

### DIFF
--- a/cache-config/t3c-apply/torequest/torequest.go
+++ b/cache-config/t3c-apply/torequest/torequest.go
@@ -1137,6 +1137,12 @@ func (r *TrafficOpsReq) StartServices(syncdsUpdate *UpdateStatus) error {
 	return nil
 }
 
+func (r *TrafficOpsReq) ShowUpdateStatus(flagType []string, start time.Time, curSetting, newSetting bool) {
+	for _, flag := range flagType {
+		log.Infof("%s flag currently set to %v, setting to %v took %v", flag, curSetting, newSetting, time.Since(start).Round(time.Millisecond))
+	}
+}
+
 func (r *TrafficOpsReq) UpdateTrafficOps(syncdsUpdate *UpdateStatus) error {
 	var performUpdate bool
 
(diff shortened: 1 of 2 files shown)


cache-config/t3c-apply/torequest/torequest.go
```diff
@@ -1169,17 +1175,25 @@ func (r *TrafficOpsReq) UpdateTrafficOps(syncdsUpdate *UpdateStatus) error {
...
 		if r.Cfg.Files == t3cutil.ApplyFilesFlagAll {
-			b := false
+			b = false
+			apply = append(apply, "update")
+			log.Infof("Update flag currently set to %v, setting to %v", serverStatus.UpdatePending, b)
 			err = sendUpdate(r.Cfg, serverStatus.ConfigUpdateTime, nil, &b, nil)
 		} else if r.Cfg.Files == t3cutil.ApplyFilesFlagReval {
-			b := false
+			b = false
+			apply = append(apply, t3cutil.ApplyFilesFlagReval.String())
```
rob05c, line 1188: This has the text value you want, but this enum is semantically something completely different.
The enum `ApplyFilesFlag` is the CLI flag for whether to apply all config files, or just the regex_revalidate.config file.
Whereas here, the word "reval" is being used to indicate revalidate pending flag in Traffic Ops.

That might confuse someone trying to understand the code in the future. Would you object to changing this to either a literal like update above, `"reval"`, or putting both in their own enum like `const TrafficOpsPendingFlag    TrafficOpsPendingFlagReval = "reval"`?

cache-config/t3c-apply/torequest/torequest.go
```diff
@@ -1137,6 +1137,12 @@ func (r *TrafficOpsReq) StartServices(syncdsUpdate *UpdateStatus) error {
 	return nil
 }
 
+func (r *TrafficOpsReq) ShowUpdateStatus(flagType []string, start time.Time, curSetting, newSetting bool) {
```
rob05c, line 1140: Nitpick, but it isn't clear from the name what "show" is doing, Usually "show" means displaying something to the user in a graphical UI. 
IMO `LogUpdateStatuses` would be clearer

cache-config/t3c-apply/torequest/torequest.go
```diff
@@ -1169,17 +1175,25 @@ func (r *TrafficOpsReq) UpdateTrafficOps(syncdsUpdate *UpdateStatus) error {
 
 	// TODO: The boolean flags/representation can be removed after ATC (v7.0+)
 	if !r.Cfg.ReportOnly && !r.Cfg.NoUnsetUpdateFlag {
+		start := time.Now()
+		apply := []string{}
+		var b bool
```
rob05c, line 1180: Another nitpick: the scope of this variable is a bit too long for a single character. Variable name lengths should correspond to their scope. 

Single-letter vars are ok for just a couple lines, but for a dozen or so like here, something more descriptive like "updateFlagVal" would be easier to read.

rob05c (approved):
LGTM. Ran t3c, observed new logs, everything else ran normally.

All my comments are minor, not worth holding this up

PR #6993, opened 2022-07-29
add support for __REGEX_REMAP_DIRECTIVE__ to remap.config processing

traeak:
Another remap.config Raw Remap Text "hack" which will allow users to modify the uri path and/or query string args that would affect regex_remap behavior.

A sample usage might be:

```
@plugin=tslua.so @pparam=uri-manipulation.lua __REGEX_REMAP_DIRECTIVE__
```

<hr/>

- Documentation

On the TP Delivery Service Page:

- Populate the regex remap text field
- Set raw remap text to: '@plugin=uri-manip.so __REGEX_REMAP_DIRECTIVE__'

Run t3c and observe the regex_remap plugin *after* the uri-manip.so plugin.

To negatively test, leave the regex remap text field blank, run again and ensure no regex_remap.so plugin directives exist.

## PR submission checklist
- [x] This PR has tests 


rob05c (approved):
LGTM. Builds, tests pass, follows existing pattern for replacements. It's a hack, but it's unfortunately something we need to make certain DS configurations possible.

